PR #549, opened 2023-02-21
[FLINK-30572] Make parquet as default data file format

SteNicholas:
- We have done some tests. Parquet is 30% faster.
- After [FLINK-30565](https://issues.apache.org/jira/browse/FLINK-30565), Parquet can support complex types and file systems such as OSS and s3 (decoupled from hadoop filesystem).
- After [FLINK-30569](https://issues.apache.org/jira/browse/FLINK-30569), the table can switch formats at will.

Therefore, if detailed and comprehensive tests have been carried out here, we can use Parquet as the default format. The default value of `file.format` should be `parquet`.

**The brief change log**

- Updates the default value of `file.format` to `parquet`.

### DIFF
--- a/docs/layouts/shortcodes/generated/core_configuration.html
+++ b/docs/layouts/shortcodes/generated/core_configuration.html
@@ -81,8 +81,14 @@
             <td>The discovery interval of continuous reading.</td>
         </tr>
         <tr>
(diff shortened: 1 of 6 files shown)


JingsongLi:
There is currently no strong reason to make parquet a default. (compaction performance is not better than orc)

PR #550, opened 2023-02-22
[FLINK-31178] Public Writer API

JingsongLi:
Based on https://github.com/apache/flink-table-store/pull/551


flink-table-store-core/src/main/java/org/apache/flink/table/store/table/sink/TableCommit.java
```diff
@@ -18,121 +18,47 @@
...
+        withOverwritten(Collections.emptyMap());
         return this;
     }
 
-    public TableCommit withLock(Lock lock) {
-        commit.withLock(lock);
-
-        if (expire != null) {
-            expire.withLock(lock);
+    default TableCommit withOverwritten(@Nullable Map<String, String> staticPartition) {
```
tsreaper, line 48: Why change method name? Specifically why use `withOverwritten` instead of `withOverwrite`? It is hard for users to understand what they'll overwrite.

flink-table-store-core/src/main/java/org/apache/flink/table/store/table/sink/TableCommit.java
```diff
@@ -18,121 +18,47 @@
...
-        }
-        return this;
-    }
+@Experimental
+public interface TableCommit extends AutoCloseable {
 
-    public TableCommit withOverwritePartitions(
-            @Nullable List<Map<String, String>> overwritePartitions) {
-        this.overwritePartitions = overwritePartitions;
+    default TableCommit withOverwritten() {
```
tsreaper, line 43: This method name is also confusing too. By calling this method nothing should be overwritten?

flink-table-store-core/src/main/java/org/apache/flink/table/store/file/operation/FileStoreWrite.java
```diff
@@ -38,6 +39,8 @@
 
     FileStoreWrite<T> withIOManager(IOManager ioManager);
 
+    FileStoreWrite<T> withMemoryPool(MemorySegmentPool memoryPool);
```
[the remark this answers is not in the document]
JingsongLi: Can share memory for multiple writes.
I will delete this method, it is wrong.

flink-table-store-core/src/main/java/org/apache/flink/table/store/table/AppendOnlyFileStoreTable.java
```diff
@@ -131,7 +130,7 @@ public RecordReader<InternalRow> createReader(Split split) throws IOException {
     }
 
     @Override
-    public TableWrite newWrite(String commitUser) {
+    public TableWriteImpl<?> newWrite(String commitUser) {
```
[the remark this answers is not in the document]
JingsongLi: OK

flink-table-store-core/src/main/java/org/apache/flink/table/store/table/sink/TableCommit.java
JingsongLi: No, it is same to SQL actually.
You can use:
`insert overwrite ...`
also you can use
`insert overwrite partition (static part...)`

flink-table-store-core/src/main/java/org/apache/flink/table/store/table/sink/TableCommit.java
JingsongLi: back to `overwrite`

flink-table-store-core/src/main/java/org/apache/flink/table/store/table/Table.java
```diff
@@ -84,6 +97,58 @@ public interface Table extends Serializable {
...
+
+    /**
+     * Create a {@link TableWrite} to write {@link InternalRow}s.
+     *
+     * <p>NOTE: This method will generate random commitUser. If you want to generate multiple
+     * commits, please use commitUser that is consistent with {@link #newCommit}.
+     */
+    default TableWrite newWrite() {
+        return newWrite(UUID.randomUUID().toString());
+    }
```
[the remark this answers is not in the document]
tsreaper: Considering that `commitUser` is needed by both `TableWrite` and `TableCommit`, and we also have a similar case that `predicate` is needed by both `TableScan` and `TableRead`, why not introduce a `ReadBuilder` and a `WriteBuilder`? We can set common variables in the builders and extract `TableWrite` or `TableCommit` from them.

flink-table-store-core/src/main/java/org/apache/flink/table/store/file/utils/RecordWriter.java
```diff
@@ -54,10 +54,10 @@
     /**
      * Prepare for a commit.
      *
-     * @param blocking if this method need to wait for current compaction to complete
+     * @param waitCompaction if this method need to wait for current compaction to complete
      * @return Incremental files in this snapshot cycle
      */
-    CommitIncrement prepareCommit(boolean blocking) throws Exception;
+    CommitIncrement prepareCommit(boolean waitCompaction) throws Exception;
```
SteNicholas, line 60: IMO, the `waitCompaction` has nothing to do with `prepareCommit`. On other words, the `compaction` operation should not be associated with the commit operation at the interface definition level.

flink-table-store-core/src/main/java/org/apache/flink/table/store/file/operation/AbstractFileStoreWrite.java
```diff
@@ -220,7 +227,9 @@ public void close() throws Exception {
             }
         }
         writers.clear();
-        compactExecutor.shutdownNow();
+        if (lazyCompactExecutor != null) {
+            lazyCompactExecutor.shutdownNow();
```
SteNicholas, line 231: `lazyCompactExecutor` sets to null?

flink-table-store-core/src/main/java/org/apache/flink/table/store/table/FileStoreTable.java
```diff
@@ -55,9 +56,10 @@ default List<String> partitionKeys() {
     FileStoreTable copy(Map<String, String> dynamicOptions);
 
     @Override
-    default BucketComputer bucketComputer() {
-        return new BucketComputer(schema());
-    }
+    TableWriteImpl<?> newWrite(String commitUser);
+
+    @Override
+    TableCommitImpl newCommit(String commitUser);
```
SteNicholas, line 62:
```suggestion
    TableCommitImpl<?> newCommit(String commitUser);
```

flink-table-store-core/src/main/java/org/apache/flink/table/store/table/ReadonlyTable.java
```diff
@@ -0,0 +1,42 @@
...
+import org.apache.flink.table.store.table.sink.TableWrite;
+
+/** Readonly table which only provide implementation for scan and read. */
+public interface ReadonlyTable extends Table {
+
+    @Override
+    default TableWrite newWrite(String commitUser) {
+        throw new UnsupportedOperationException(
+                String.format(
+                        "Readonly Table %s does not support newWrite.",
```
SteNicholas, line 31: No need to tell users the class name of `ReadonlyTable`? Or provide user with the type of table?

flink-table-store-core/src/main/java/org/apache/flink/table/store/table/ReadonlyTable.java
```diff
@@ -0,0 +1,42 @@
...
+                        this.getClass().getSimpleName()));
+    }
+
+    @Override
+    default TableCommit newCommit(String commitUser) {
+        throw new UnsupportedOperationException(
+                String.format(
+                        "Readonly Table %s does not support newCommit.",
```
SteNicholas, line 39: Ditto.

flink-table-store-core/src/main/java/org/apache/flink/table/store/table/sink/CommitMessage.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Commit message for partition and bucket.
+ *
+ * @since 0.4.0
+ */
+@Experimental
+public interface CommitMessage extends Serializable {
+
+    BinaryRow partition();
```
SteNicholas, line 34: Adds the comment of the interface methods.

flink-table-store-core/src/main/java/org/apache/flink/table/store/table/sink/TableCommitImpl.java
```diff
@@ -0,0 +1,137 @@
...
+    public void commit(List<ManifestCommittable> committables) {
+        if (overwritePartitions == null) {
+            for (ManifestCommittable committable : committables) {
+                commit.commit(committable, new HashMap<>());
+            }
+        } else {
+            ManifestCommittable committable;
+            if (committables.size() > 1) {
+                throw new RuntimeException(
+                        "Multiple committables appear in overwrite mode, this may be a bug, please report it: "
```
SteNicholas, line 109: How the users solve the problem that multiple committables appear in overwrite mode? This should provide users with the solution.

flink-table-store-core/src/main/java/org/apache/flink/table/store/table/sink/TableCommitImpl.java
```diff
@@ -0,0 +1,137 @@
...
+
+        if (partitionExpire != null) {
+            partitionExpire.expire();
+        }
+    }
+
+    @Override
+    public void close() throws Exception {
+        if (lock != null) {
+            lock.close();
```
SteNicholas, line 134: `lock` sets to null.

SteNicholas (changes requested):
@JingsongLi, I left the comments for Writer interface. PTAL.

flink-table-store-core/src/main/java/org/apache/flink/table/store/table/ReadonlyTable.java
JingsongLi: I think it is reasonable to provide class name.

flink-table-store-core/src/main/java/org/apache/flink/table/store/file/utils/RecordWriter.java
JingsongLi: In the batch mode, we should wait compaction for the writer.

flink-table-store-core/src/main/java/org/apache/flink/table/store/file/operation/AbstractFileStoreWrite.java
JingsongLi: A exception should be threw if the write is used again.

flink-table-store-core/src/main/java/org/apache/flink/table/store/table/FileStoreTable.java
JingsongLi: `TableCommitImpl` is not generic.

flink-table-store-core/src/main/java/org/apache/flink/table/store/table/sink/TableCommitImpl.java
JingsongLi: This is not user face api, this is internal api.

flink-table-store-core/src/main/java/org/apache/flink/table/store/table/sink/TableCommitImpl.java
JingsongLi: I think let it non-null to throw exception.

flink-table-store-core/src/main/java/org/apache/flink/table/store/table/sink/WriteBuilder.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.List;
+import java.util.Map;
+
+/**
+ * An interface for building the {@link TableWrite} and {@link TableCommit}.
+ *
+ * @since 0.4.0
+ */
+@Experimental
+public interface WriteBuilder extends Serializable {
```
tsreaper, line 38: There are detailed example in `ReadBuilder`. Also add example in `WriteBuilder`?

flink-table-store-core/src/main/java/org/apache/flink/table/store/table/source/ReadBuilder.java
```diff
@@ -0,0 +1,107 @@
...
+    /** A name to identify the table. */
+    String tableName();
+
+    /** Returns read row type, projected by {@link #withProjection}. */
+    RowType readType();
+
+    /**
+     * Push filters, the relationship between predicates is 'AND'. It will filter the data as much
+     * as possible, but it is not guaranteed that it is a complete filter.
+     */
```
tsreaper, line 72:
```suggestion
    /**
     * Apply filters to the readers to decrease the number of produced records.
     *
     * <p>This interface filters records as much as possible, however some produced records may not satisfy all predicates. Users need to recheck all records.
     */
```

flink-table-store-core/src/main/java/org/apache/flink/table/store/table/sink/TableCommit.java
```diff
@@ -18,121 +18,55 @@
...
+ * <ol>
+ *   <li>Before calling {@link TableCommit#commit}, if user cannot determine if this commit is done
+ *       before, user should first call {@link TableCommit#filterCommitted}.
+ *   <li>Before committing, it will first check for conflicts by checking if all files to be removed
+ *       currently exists, and if modified files have overlapping key ranges with existing files.
+ *   <li>After that it use the external {@link CatalogLock} (if provided) or the atomic rename of
+ *       the file system to ensure atomicity.
+ *   <li>If commit fails due to conflicts or exception it tries its best to clean up and aborts.
+ *   <li>If atomic rename fails it tries again after reading the latest snapshot from step 2.
+ * </ol>
```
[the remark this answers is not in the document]
JingsongLi: I think we can leave these comments:
```
 * <ol>
 *   <li>Before calling {@link TableCommit#commit}, if user cannot determine if this commit is done
 *       before, user should first call {@link TableCommit#filterCommitted}.
 *   <li>Commit may fail and throw an exception. Before committing, it will first check for conflicts by checking if all files to be removed currently exists.
 * </ol>
```

flink-table-store-core/src/main/java/org/apache/flink/table/store/table/sink/TableCommit.java
```diff
@@ -18,121 +18,55 @@
...
+ *
+ * <p>According to the options, the expiration of snapshots and partitions will be completed in
+ * commit.
```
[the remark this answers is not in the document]
JingsongLi: I still combining commit and expire together, but different methods, add methods in `TableCommit`: `expireSnapshots` and `expirePartitions`.

flink-table-store-core/src/main/java/org/apache/flink/table/store/table/sink/TableCommit.java
```diff
@@ -18,121 +18,76 @@
...
-        }
-    }
+    /** Filter committed commits. This method is used for failover cases. */
+    Set<Long> filterCommitted(Set<Long> commitIdentifiers);
+
+    /**
+     * Create a new commit. One commit may generate two snapshots, one for adding new files and the
+     * other for compaction.
+     */
+    void commit(long commitIdentifier, List<CommitMessage> commitMessages);
```
tsreaper, line 67: Explain what is a `commitIdentifier` and stress that it must become larger for later commits.

flink-table-store-core/src/main/java/org/apache/flink/table/store/table/sink/WriteBuilder.java
```diff
@@ -0,0 +1,108 @@
...
+    RowType rowType();
+
+    /** Get commit user, set by {@link #withCommitUser}. */
+    String commitUser();
+
+    /**
+     * Set commit user, the default value is UUID. The commit user used by {@link TableWrite} and
+     * {@link TableCommit} must be the same, otherwise there will be some conflicts.
+     */
+    WriteBuilder withCommitUser(String commitUser);
```
tsreaper, line 84: > Set commit user, the default value is UUID.

the default value is a random UUID.

> he commit user used by TableWrite and TableCommit must be the same, otherwise there will be some conflicts.

Users cannot set commitUser separately for `TableWrite` and `TableCommit` with public API. No need for this statement.

flink-table-store-core/src/main/java/org/apache/flink/table/store/table/sink/WriteBuilder.java
```diff
@@ -0,0 +1,108 @@
...
+ * // 3. Collect all CommitMessages to a global node
+ *
+ * // 3.1 commit
+ * TableCommit commit = builder.newCommit();
+ * if (commit.filterCommitted(0)) {
+ *     // It has already been committed. In order to avoid
+ *     // repeated committing, return directly.
+ *     return;
+ * }
+ * commit.commit(0, allCommitMessages());
```
tsreaper, line 58: What does 0 mean here? Add comments.

flink-table-store-core/src/main/java/org/apache/flink/table/store/table/sink/TableCommit.java
```diff
@@ -18,121 +18,76 @@
...
+    /**
+     * Default ignore empty commit, if this is set to false, when there is no new data, an empty
+     * commit will also be created.
+     *
+     * <p>NOTE: It is recommended to set 'ignoreEmptyCommit' to false in streaming write, in order
+     * to better remove duplicate commits (See {@link #filterCommitted}).
+     */
+    TableCommit ignoreEmptyCommit(boolean ignoreEmptyCommit);
+
+    /** Filter committed commit. This method is used for failover cases. */
```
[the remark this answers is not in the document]
JingsongLi: I will remove this method

flink-table-store-core/src/main/java/org/apache/flink/table/store/table/sink/WriteBuilder.java
JingsongLi: > Users cannot set commitUser separately for TableWrite and TableCommit with public API. No need for this statement.

User may use different `WriteBuilder`, I think we can add this setence.

flink-table-store-core/src/main/java/org/apache/flink/table/store/table/sink/TableWrite.java
```diff
@@ -18,39 +18,42 @@
...
     /**
-     * Notify that some new files are created at given snapshot in given bucket.
+     * Compact a bucket of a partition. By default, it will determine whether to perform the
+     * compaction according to the 'num-sorted-run.compaction-trigger' option. If fullCompaction is
+     * true, it will force a full compaction, which is expensive.
      *
-     * <p>Most probably, these files are created by another job. Currently this method is only used
-     * by the dedicated compact job to see files created by writer jobs.
+     * <p>NOTE: In Java API, full compaction is not automatically executed. If you open
+     * 'changelog-producer' of 'full-compaction', please execute this method regularly to produce
```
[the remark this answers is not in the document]
JingsongLi: set 'changelog-producer' to 'full-compaction'

tsreaper (approved):

PR #551, opened 2023-02-22
[FLINK-31179] Make data structures serializable

JingsongLi:
This can make our internal data structures easy to use.

### DIFF
--- a/flink-table-store-common/src/main/java/org/apache/flink/table/store/data/BinaryMap.java
+++ b/flink-table-store-common/src/main/java/org/apache/flink/table/store/data/BinaryMap.java
@@ -34,8 +34,10 @@ import static org.apache.flink.table.store.utils.Preconditions.checkArgument;
  */
 public final class BinaryMap extends BinarySection implements InternalMap {
 
-    private final BinaryArray keys;
-    private final BinaryArray values;
+    private static final long serialVersionUID = 1L;
+
+    private transient BinaryArray keys;
+    private transient BinaryArray values;
 
     public BinaryMap() {
         keys = new BinaryArray();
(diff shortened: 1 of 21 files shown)


flink-table-store-common/src/main/java/org/apache/flink/table/store/data/BinaryMap.java
```diff
@@ -54,7 +56,13 @@ public void pointTo(MemorySegment[] segments, int offset, int sizeInBytes) {
         final int valueArrayBytes = sizeInBytes - keyArrayBytes - 4;
         assert valueArrayBytes >= 0 : "valueArraySize (" + valueArrayBytes + ") should >= 0";
 
+        if (keys == null) {
+            keys = new BinaryArray();
+        }
         keys.pointTo(segments, offset + 4, keyArrayBytes);
+        if (values == null) {
+            values = new BinaryArray();
+        }
```
tsreaper, line 65: Add comments about why we need to check for `null`.

flink-table-store-common/src/main/java/org/apache/flink/table/store/data/BinaryRow.java
```diff
@@ -49,9 +50,14 @@
  * bigger pageSize of MemorySegment.
  *
  * <p>Variable-length part may fall into multiple MemorySegments.
+ *
+ * @since 0.4.0
```
tsreaper, line 54: Also add this to `BinaryArray`, `BinaryMap`, `ColumnarRow` and many other classes?

JingsongLi: Added. ColumnarRow is still internal.

tsreaper (approved):

PR #554, opened 2023-02-24
[FLINK-31201] Scan to generate splits should care about file order

JingsongLi:
Currently, it is out of order for partitions. Because HashMap is used, we may be able to use LinkedHashMap to make the order according to the creation time of the first file.

### DIFF
--- a/flink-table-store-connector/src/test/java/org/apache/flink/table/store/connector/AppendOnlyTableITCase.java
+++ b/flink-table-store-connector/src/test/java/org/apache/flink/table/store/connector/AppendOnlyTableITCase.java
@@ -19,6 +19,7 @@
 package org.apache.flink.table.store.connector;
 
 import org.apache.flink.table.api.TableException;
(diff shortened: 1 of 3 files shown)


docs/layouts/shortcodes/generated/core_configuration.html
```diff
@@ -218,6 +218,12 @@
...
         </tr>
+        <tr>
+            <td><h5>scan.plan-sort-partition</h5></td>
+            <td style="word-wrap: break-word;">false</td>
+            <td>Boolean</td>
+            <td>Whether to sort plan files by partition fields, this allows you to read according to the partition order, even if your partition writes are out of order.</td>
```
SteNicholas, line 225: It's better to explain the usage case of the `scan.plan-sort-partition` to avoid the user misunderstand the usage of this option and note that there is no difference between `scan.plan-sort-partition=true` and `order by [partition field]` in batch mode.

PR #560, opened 2023-02-27
[FLINK-31232] Parquet format supports MULTISET type for Table Store

SteNicholas:
Parquet format should support MULTISET type for Table Store. Meanwhile, the FileStats couldn't support the MULTISET type because the null count of ColumnStatistics for ARRAY, MAP and MULTISET type in Parquet isn't right.

**The brief change log**
- The reader, writer and converter of Parquet format supports MULTISET type.


JingsongLi (approved):
+1

PR #561, opened 2023-02-27
[FLINK-31236] Limit pushdown should not open useless RecordReader

JingsongLi:


flink-table-store-flink/flink-table-store-flink-common/src/main/java/org/apache/flink/table/store/connector/source/FileStoreSourceSplitReader.java
```diff
@@ -120,7 +122,9 @@ public void wakeUp() {}
     @Override
     public void close() throws Exception {
         if (currentReader != null) {
-            currentReader.close();
+            if (currentReader.lazyRecordReader != null) {
+                currentReader.lazyRecordReader.close();
```
SteNicholas, line 126: Should currentReader set to null?

flink-table-store-flink/flink-table-store-flink-common/src/main/java/org/apache/flink/table/store/connector/source/FileStoreSourceSplitReader.java
```diff
@@ -120,7 +122,9 @@ public void wakeUp() {}
     @Override
     public void close() throws Exception {
         if (currentReader != null) {
-            currentReader.close();
+            if (currentReader.lazyRecordReader != null) {
```
SteNicholas, line 125:
```suggestion
            if (currentReader.recordReader() != null) {
```

SteNicholas (approved):
LGTM.

PR #562, opened 2023-02-28
[FLINK-31248] Improve documentation for append-only table

JingsongLi:


docs/content/docs/features/primary-key-table.md
```diff
@@ -158,12 +154,21 @@ Full compaction changelog producer can produce complete changelog for any type o
 
 {{< /hint >}}
 
-## Changelog Tables without Primary Keys
+## Sequence Field
```
tsreaper, line 157: Change this as a sub-section of "Merge Engines", because it only affects the result of deduplicate merge engine.

docs/content/docs/features/primary-key-table.md
```diff
@@ -158,12 +154,21 @@ Full compaction changelog producer can produce complete changelog for any type o
...
 
-Changelog tables can also be used without primary keys. Users can only insert or delete a whole record from the table. No update is supported.
+By default, the primary key table determines the order according to the input order. However, in distributed computing,
```
tsreaper, line 159: By default, primary key tables determine the latest record according to the input order.

docs/content/docs/features/primary-key-table.md
```diff
@@ -158,12 +154,21 @@ Full compaction changelog producer can produce complete changelog for any type o
...
+there will be some cases that lead to data disorder. At this time, you can use a time field as `sequence.field`, so
+the storage engine will use this field to sort to ensure the consistency of data sequence. For example:
```
tsreaper, line 161: will use this field to determine the latest record instead. For example:

docs/content/docs/features/append-only-table.md
```diff
@@ -0,0 +1,98 @@
...
+You can only insert a whole record into the table. No delete or update is supported and you cannot define primary keys.
+This type of table is suitable for use cases that do not require updates (such as log data synchronization).
+
+## Distribution
+
+You also need to define bucket number for Append-only table, see [Bucket]({{< ref "docs/concepts/basic-concepts#bucket" >}}).
+
+It is recommended that you set the `bucket-key` field. Otherwise, the data will be hashed according to the whole row,
+and the performance will be poor.
+
```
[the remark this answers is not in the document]
JingsongLi: Bucketing is not only for performance, but also related to the order.
I will just change `Distribution` to `Bucketing`.

docs/content/docs/features/primary-key-table.md
JingsongLi: I think it is also related to the changelog, it will affect the changelog result.

docs/content/docs/features/primary-key-table.md
JingsongLi: Change to: `By default, the primary key table determines the merge order according to the input order`.

docs/content/docs/features/primary-key-table.md
JingsongLi: This sentence is repeated. I delete it.

docs/content/docs/features/append-only-table.md
```diff
@@ -0,0 +1,97 @@
...
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+# Append Only Table
```
SteNicholas, line 27: Could you add the configuration explanation for COW and MOR for append only table?

docs/content/docs/features/primary-key-table.md
```diff
@@ -158,12 +154,20 @@ Full compaction changelog producer can produce complete changelog for any type o
...
+    v1 DOUBLE,
+    v2 BIGINT,
+    dt TIMESTAMP
+) WITH (
+    'sequence.field' = 'dt'
+);
+```
 
-You can only insert a whole record into the table. No delete or update is supported and you cannot define primary keys. This type of table is suitable for use cases that do not require updates (such as log data synchronization).
+Regardless of the sequence of data input, the final correct result will always be obtained.
```
SteNicholas, line 173: Which is the final correct result? Do you mean that the output data of the table is correct no matter what the sequence of the data input is.

docs/content/docs/features/append-only-table.md
```diff
@@ -0,0 +1,97 @@
...
+
+By specifying `'write-mode' = 'append-only'` when creating the table, user creates an append-only table.
+
+You can only insert a complete record into the table. No delete or update is supported and you cannot define primary keys.
+This type of table is suitable for use cases that do not require updates (such as log data synchronization).
+
+```sql
```
SteNicholas, line 34: IMO, the DDL definition of the append only table is better to document at last.

docs/content/docs/features/primary-key-table.md
JingsongLi: I change this to: `Regardless of the order of data input, the final correct result will always be obtained`.

docs/content/docs/features/append-only-table.md
JingsongLi: I agree, it is good to put this after some concepts.

docs/content/docs/features/append-only-table.md
tsreaper: I disagree. Concepts is a category for general ideas, not for detailed documents like "how shall we create a table" or "how to tweak the performance of compaction". We might need a general "table types" page in concepts to introduce what is a primary-key table and what is a append-only table, but absolutely not for detailed examples.

SteNicholas: @tsreaper, I previously comment that refers to [Table & Query Types](https://hudi.apache.org/docs/table_types/) to introduce the append only table and primary key table. @JingsongLi, WDYT?

SteNicholas (approved):
LGTM.

docs/content/docs/concepts/primary-key-table.md
```diff
@@ -158,12 +154,28 @@ Full compaction changelog producer can produce complete changelog for any type o
 
 {{< /hint >}}
 
-## Changelog Tables without Primary Keys
+## Sequence Field
 
-Changelog tables can also be used without primary keys. Users can only insert or delete a whole record from the table. No update is supported.
+By default, the primary key table determines the merge order according to the input order. However, in distributed computing,
```
tsreaper, line 159:
```suggestion
By default, the primary key table determines the merge order according to the input order (the last input record will be the last to merge). However, in distributed computing,
```

docs/content/docs/concepts/primary-key-table.md
```diff
@@ -158,12 +154,28 @@ Full compaction changelog producer can produce complete changelog for any type o
...
+    'sequence.field' = 'dt'
+);
+```
+
+{{< /tab >}}
+
+{{< /tabs >}}
 
-You can only insert a whole record into the table. No delete or update is supported and you cannot define primary keys. This type of table is suitable for use cases that do not require updates (such as log data synchronization).
+Regardless of the order of data input, the final correct result will always be obtained.
```
tsreaper, line 181:
```suggestion
The record with the largest `sequence.field` value will be the last to merge, regardless of the input order.
```

docs/content/docs/concepts/append-only-table.md
```diff
@@ -0,0 +1,107 @@
...
+# Append Only Table
+
+By specifying `'write-mode' = 'append-only'` when creating the table, user creates an append-only table.
+
+You can only insert a complete record into the table. No delete or update is supported and you cannot define primary keys.
+This type of table is suitable for use cases that do not require updates (such as log data synchronization).
+
+## Bucketing
+
+You also need to define bucket number for Append-only table, see [Bucket]({{< ref "docs/concepts/basic-concepts#bucket" >}}).
```
tsreaper, line 36:
```suggestion
You can also define bucket number for Append-only table, see [Bucket]({{< ref "docs/concepts/basic-concepts#bucket" >}}).
```

Users don't "need to" define bucket number. They can always use the default bucket number, which is 1.

docs/content/docs/concepts/append-only-table.md
```diff
@@ -0,0 +1,107 @@
...
+
+* For any two records from two different partitions
+  * If `scan.plan-sort-partition` is set to true, the record with a smaller partition value will be produced first.
+  * Otherwise, the record with an earlier partition creation time will be produced first.
+* For any two records from the same partition and the same bucket, the first written record will be produced first.
+* For any two records from the same partition but two different buckets, different buckets are processed by different tasks, there is no order guarantee between them.
+
+## Compaction
+
+By default, the sink node will automatically perform compaction to solve the small file problem. The following options
```
tsreaper, line 53:
```suggestion
By default, the sink node will automatically perform compaction to control the number of files. The following options
```

tsreaper (approved):

docs/content/docs/features/append-only-table.md
```diff
@@ -0,0 +1,98 @@
...
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+# Append Only Table
+
+By specifying `'write-mode' = 'append-only'` when creating the table, user creates an append-only table.
+
+You can only insert a whole record into the table. No delete or update is supported and you cannot define primary keys.
```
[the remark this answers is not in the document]
Jackofmn12: A

PR #563, opened 2023-02-28
[FLINK-31252] Improve StaticFileStoreSplitEnumerator to assign batch splits

JingsongLi:
The following batch assignment operation is for two things:
1. It can be evenly distributed during batch reading to avoid scheduling problems (for example, the current resource can only schedule part of the tasks) that cause some tasks to fail to read data.
2. Read with limit, if split is assigned one by one, it may cause the task to repeatedly create SplitFetchers. After the task is created, it is found that it is idle and then closed. Then, new split coming, it will create a new SplitFetcher and repeatedly read the data of the limit number (because the limit status is in the SplitFetcher).

### DIFF
--- a/flink-table-store-flink/flink-table-store-flink-common/src/main/java/org/apache/flink/table/store/connector/source/StaticFileStoreSplitEnumerator.java
+++ b/flink-table-store-flink/flink-table-store-flink-common/src/main/java/org/apache/flink/table/store/connector/source/StaticFileStoreSplitEnumerator.java
@@ -75,14 +75,13 @@ public class StaticFileStoreSplitEnumerator
             return;
         }
 
-        // The following batch assignment operation is for two things:
-        // 1. It can be evenly distributed during batch reading to avoid scheduling problems (for
-        // example, the current resource can only schedule part of the tasks) that cause some tasks
-        // to fail to read data.
-        // 2. Read with limit, if split is assigned one by one, it may cause the task to repeatedly
-        // create SplitFetchers. After the task is created, it is found that it is idle and then
-        // closed. Then, new split coming, it will create SplitFetcher and repeatedly read the data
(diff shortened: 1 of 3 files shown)


flink-table-store-flink/flink-table-store-flink-common/src/main/java/org/apache/flink/table/store/connector/source/StaticFileStoreSplitEnumerator.java
```diff
@@ -61,17 +75,27 @@ public void handleSplitRequest(int subtask, @Nullable String hostname) {
...
+        // 1. It can be evenly distributed during batch reading to avoid scheduling problems (for
+        // example, the current resource can only schedule part of the tasks) that cause some tasks
+        // to fail to read data.
+        // 2. Read with limit, if split is assigned one by one, it may cause the task to repeatedly
+        // create SplitFetchers. After the task is created, it is found that it is idle and then
+        // closed. Then, new split coming, it will create SplitFetcher and repeatedly read the data
+        // of the limit number (the limit status is in the SplitFetcher).
+        List<FileStoreSourceSplit> splits = pendingSplitAssignment.remove(subtask);
+        if (splits != null && splits.size() > 0) {
+            context.assignSplits(new SplitsAssignment<>(Collections.singletonMap(subtask, splits)));
```
yuzelin, line 88: Comment:
The following batch assignment operation is for two purposes:
1. To distribute splits evenly when batch reading to avoid failure of some tasks to read data caused by scheduling problems (for example, the current resource can only schedule part of the tasks).
2. Optimize limit reading. In limit reading, the task will repeatedly create SplitFetcher to read the data of the limit number (the limit status is in the SplitFetcher). So if the splits are assigned one by one, it will cost more time.

yuzelin (approved):
LGTM. I think the comment can be clearer.

PR #564, opened 2023-02-28
[FLINK-31254] Improve the read performance for files table

JingsongLi:
At present, the reading performance of the Files table is very poor. Even every data read will read the schema file. We can optimize the reading performance.

### DIFF
--- a/flink-table-store-common/src/main/java/org/apache/flink/table/store/data/LazyGenericRow.java
+++ b/flink-table-store-common/src/main/java/org/apache/flink/table/store/data/LazyGenericRow.java
@@ -1,0 +1,171 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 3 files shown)


flink-table-store-core/src/main/java/org/apache/flink/table/store/file/stats/FieldStatsConverters.java
```diff
@@ -33,12 +33,14 @@
 public class FieldStatsConverters {
     private final Function<Long, List<DataField>> schemaFields;
     private final long tableSchemaId;
+    private final List<DataField> tableDataFields;
     private final ConcurrentMap<Long, FieldStatsArraySerializer> serializers;
     private final AtomicReference<List<DataField>> tableFields;
 
     public FieldStatsConverters(Function<Long, List<DataField>> schemaFields, long tableSchemaId) {
         this.schemaFields = schemaFields;
         this.tableSchemaId = tableSchemaId;
```
FangYongs, line 42: We can remove tableSchemaId here if we get table data fields in the constructor

flink-table-store-common/src/main/java/org/apache/flink/table/store/data/LazyGenericRow.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    @Override
+    public BinaryString getString(int pos) {
+        return (BinaryString) getField(pos);
+    }
+
+    @Override
+    public Decimal getDecimal(int pos, int precision, int scale) {
+        return (Decimal) getField(pos);
```
FangYongs, line 132: `scale` is missing here, how about throw exception here? Otherwise, it may return invalid result

flink-table-store-core/src/main/java/org/apache/flink/table/store/file/stats/FieldStatsConverters.java
JingsongLi: Still have check below: `if (tableSchemaId == id)`

flink-table-store-common/src/main/java/org/apache/flink/table/store/data/LazyGenericRow.java
JingsongLi: This is just like `GenericRow`, we trust the decimal returns by `suppliers` has correct precision and scale.

FangYongs:
LGTM +1

PR #565, opened 2023-02-28
[FLINK-31226] Bump FRocksDB version to 6.20.3-ververica-2.0

Myasuka:
Current FRocksDB version in flink-table-store is still 6.20.3-ververica-1.0, which cannot run with Apple's silicon chips.
Newly released 6.20.3-ververica-2.0 FRocksDB could run well with Apple's silicon chips and is compatible with the previous version.


JingsongLi (approved):
Looks good to me! Thanks @Myasuka

PR #568, opened 2023-03-02
[FLINK-31289] Default aggregate-function for field can be last_non_null_value

FangYongs:
If the field has no aggregate function for aggregation merge-engine, use last_non_null_value aggregation as default.


JingsongLi (approved):
Looks good to me!

PR #571, opened 2023-03-02
[FLINK-31291] Document "table.exec.sink.upsert-materialize" to "none" to avoid strange behavior

schnappi17:
## What is the purpose of the change

Document set `table.exec.sink.upsert-materialize` to `NONE` always to avoid strange behavior brought by FLINK SQL default sink materialize.


JingsongLi (approved):
+1

PR #576, opened 2023-03-03
[FLINK-31311] Supports Bounded Watermark streaming read

JingsongLi:
There are some bound stream scenarios that require that stream reading can be ended. Generally speaking, the end event time is the better.

So in this ticket, supports writing the watermark to the snapshot and can specify the ending watermark when reading the stream.

### DIFF
--- a/flink-table-store-core/src/main/java/org/apache/flink/table/store/table/source/snapshot/SnapshotEnumerator.java
+++ b/flink-table-store-core/src/main/java/org/apache/flink/table/store/table/source/snapshot/SnapshotEnumerator.java
@@ -29,11 +29,14 @@ import java.util.List;
 public interface SnapshotEnumerator {
 
     /**
-     * The first call to this method will produce a {@link DataFilePlan} containing the base files
-     * for the following incremental changes (or just return null if there are no base files).
+     * The first call to this method will produce a {@link PlanResult} containing the base files for
+     * the following incremental changes (or containing null if there are no base files).
      *
-     * <p>Following calls to this method will produce {@link DataFilePlan}s containing incremental
-     * changed files. If there is currently no newer snapshots, null will be returned instead.
+     * <p>Following calls to this method will produce {@link PlanResult}s containing incremental
+     * changed files. If there is currently no newer snapshots, {@link PlanResult} with null plan
+     * will be returned instead.
+     *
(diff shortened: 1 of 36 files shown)


flink-table-store-core/src/main/java/org/apache/flink/table/store/table/source/snapshot/SnapshotEnumerator.java
```diff
@@ -18,22 +18,78 @@
...
-     * the base files for the following incremental changes (or just return null if there are no
-     * base files).
+     * The first call to this method will produce a {@link DataFilePlan} containing the base files
+     * for the following incremental changes (or just return null if there are no base files).
      *
-     * <p>Following calls to this method will produce {@link DataTableScan.DataFilePlan}s containing
-     * incremental changed files. If there is currently no newer snapshots, null will be returned
-     * instead.
+     * <p>Following calls to this method will produce {@link DataFilePlan}s containing incremental
+     * changed files. If there is currently no newer snapshots, null will be returned instead.
```
yuzelin, line 36: Need update comment.
1. produce Result ?
2. return null -> return FinishedResult

flink-table-store-core/src/main/java/org/apache/flink/table/store/table/source/snapshot/SnapshotEnumerator.java
```diff
@@ -18,22 +18,78 @@
...
+
+    /** Result for finished. */
+    class FinishedResult implements Result {
+
+        @Nullable
+        @Override
+        public DataFilePlan plan() {
+            throw new RuntimeException("This is a finished result.");
+        }
+    }
```
yuzelin, line 94: I think maybe we can use `public static final FINISHED_RESULT = new Result{...}` instead.

yuzelin (approved):
LGTM.
I left some comments.

flink-table-store-core/src/main/java/org/apache/flink/table/store/table/source/snapshot/SnapshotEnumerator.java
JingsongLi: I still use `FinishedResult`, I think this is more clear to invokers.

PR #577, opened 2023-03-03
[FLINK-31287]Modify the default value of changelog-producer.compactio…

xuzhiwen1255:
…n-interval to 0s

### DIFF
--- a/docs/content/docs/concepts/primary-key-table.md
+++ b/docs/content/docs/concepts/primary-key-table.md
@@ -150,7 +150,7 @@ If your input can’t produce a complete changelog but you still want to get rid
 
 By specifying `'changelog-producer' = 'full-compaction'`, Table Store will compare the results between full compactions and produce the differences as changelog. The latency of changelog is affected by the frequency of full compactions.
 
-By specifying `changelog-producer.compaction-interval` table property (default value `30min`), users can define the maximum interval between two full compactions to ensure latency. This table property does not affect normal compactions and they may still be performed once in a while by writers to reduce reader costs.
+By specifying `changelog-producer.compaction-interval` table property (default value `0s`), users can define the maximum interval between two full compactions to ensure latency. This table property does not affect normal compactions and they may still be performed once in a while by writers to reduce reader costs.
 
 {{< img src="/img/changelog-producer-full-compaction.png">}}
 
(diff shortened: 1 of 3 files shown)


docs/content/docs/concepts/primary-key-table.md
```diff
@@ -150,7 +150,7 @@ If your input can’t produce a complete changelog but you still want to get rid
 
 By specifying `'changelog-producer' = 'full-compaction'`, Table Store will compare the results between full compactions and produce the differences as changelog. The latency of changelog is affected by the frequency of full compactions.
 
-By specifying `changelog-producer.compaction-interval` table property (default value `30min`), users can define the maximum interval between two full compactions to ensure latency. This table property does not affect normal compactions and they may still be performed once in a while by writers to reduce reader costs.
+By specifying `changelog-producer.compaction-interval` table property (default value `0s`), users can define the maximum interval between two full compactions to ensure latency. This table property does not affect normal compactions and they may still be performed once in a while by writers to reduce reader costs.
```
SteNicholas, line 153: Could this describe the default behavior for the default value of `changelog-producer.compaction-interval`, which default behavior is that each checkpoint have a full-compaction and generates a changelog?

xuzhiwen1255: @SteNicholas Thanks for the reminder, I was sloppy.

flink-table-store-flink/flink-table-store-flink-common/src/main/java/org/apache/flink/table/store/connector/FlinkConnectorOptions.java
```diff
@@ -102,7 +102,7 @@ public class FlinkConnectorOptions {
     public static final ConfigOption<Duration> CHANGELOG_PRODUCER_FULL_COMPACTION_TRIGGER_INTERVAL =
             key("changelog-producer.compaction-interval")
                     .durationType()
-                    .defaultValue(Duration.ofMinutes(30))
+                    .defaultValue(Duration.ofSeconds(0))
```
SteNicholas, line 105: Could `changelog-producer.compaction-interval` has no default value? If the value of `changelog-producer.compaction-interval` is null or 0, the default behavior is both that each checkpoint has a full compression and generates a change log.

xuzhiwen1255: Yes, change to default to a full compression for each ck.

SteNicholas: @xuzhiwen1255, therefore, this changes to `noDefaultValue()`?

xuzhiwen1255: If there's no default, isn't that a null pointer exception

xuzhiwen1255: noDefaultValue will cause NullPointerException,so the default is 0

xuzhiwen1255:
Compiling and deploying flink1.14 locally and executing test is no problem...

JingsongLi (approved):
+1

xuzhiwen1255:
Thanks @JingsongLi  and @SteNicholas .

PR #580, opened 2023-03-04
[FLINK-31323] Fix unstable E2E test for flink actions

yuzelin:
Use Table API to do insert job to ensure the batch configuration takes effect.


flink-table-store-flink/flink-table-store-flink-common/src/main/java/org/apache/flink/table/store/connector/action/ActionBase.java
```diff
@@ -125,12 +131,41 @@ protected boolean compatibleCheck(List<DataType> actualTypes, List<DataType> exp
...
+                                        Lock.factory(
+                                                catalog.lockFactory().orElse(null), identifier))
+                                .build()
+                                .getTransformation());
+
+        List<String> sinkIdentifierNames = Collections.singletonList(identifier.getFullName());
+
+        // invoke TableEnvironmentImpl#executeInternal through reflecting
+        try {
+            // for Flink 1.14
```
JingsongLi, line 150: This looks more complicated to maintain.
Can we just introduce a class `TableEnvironmentUtils` for this? 1.14 and common can have their own classes.
And you can add a simple ITCase for 1.14, 1.15 too.

JingsongLi (approved):
+1

PR #581, opened 2023-03-05
[FLINK-31190] Supports Spark call procedure command on Table Store

SteNicholas:
At present Hudi and Iceberg supports the Spark call procedure command to execute the table service action etc. Flink Table Store could also support Spark call procedure command to run compaction etc.

**The brief change log**
- Introduces `TableStoreSparkSessionExtensions` to extends the syntax and adds the rules of `SparkSessionExtensions`, including the CALL statement.


JingsongLi:
Wow~ Thanks @SteNicholas for your contribution!

JingsongLi:
Can we just implement a `CALL` procedure for validation?

PR #582, opened 2023-03-06
[FLINK-31329] Fix Parquet stats extractor

JingsongLi:
Some bugs in Parquet stats extractor:
- Decimal Supports
- Timestamp Supports
- Null nullCounts supports


SteNicholas (approved):
LGTM.

PR #583, opened 2023-03-06
[FLINK-31331] Flink 1.16 should implement new LookupFunction

JingsongLi:
Only implements new LookupFunction, retry lookup join can work.


yuzelin (approved):
LGTM!

PR #584, opened 2023-03-06
[FLINK-31338] support infer parallelism for flink table store

zhangjun0x01:


zhangjun0x01:
I run the e2e test `org.apache.flink.table.store.tests.FileStoreBatchE2eTest` of master branch in my computer ,it has the same error.I think it is may be unstable

```

[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 931.583 s <<< FAILURE! - in org.apache.flink.table.store.tests.FileStoreBatchE2eTest
[ERROR] testWithoutPk  Time elapsed: 930.599 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: 
Result is still unexpected after 60 retries.
Expected: {20211111, 08, Alice, Food, 90=1, 20211110, 08, Alice, Drink, 20=1, 20211111, 08, Alice, Drink, 100=1, 20211110, 08, Bob, Food, 30=1, 20211110, 09, Alice, Food, 50=1, 20211111, 09, Alice, Food, 130=1, 20211111, 08, Bob, Food, 110=1, 20211111, 09, Bob, Food, 150=1, 20211111, 09, Bob, Drink, 160=1, 20211111, 09, Alice, Drink, 140=1, 20211110, 09, Alice, Drink, 60=1, 20211110, 08, Alice, Food, 10=1, 20211110, 09, Bob, Drink, 80=1, 20211110, 08, Bob, Drink, 40=1, 20211111, 08, Bob, Drink, 120=1, 20211110, 09, Bob, Food, 70=1}
Actual: {}
        at org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:39)
        at org.junit.jupiter.api.Assertions.fail(Assertions.java:134)
        at org.apache.flink.table.store.tests.E2eTestBase.checkResult(E2eTestBase.java:261)
        at org.apache.flink.table.store.tests.FileStoreBatchE2eTest.testWithoutPk(FileStoreBatchE2eTest.java:112)



```

JingsongLi (commented):
I guess why there are test failures is because too much concurrency has been derived, resulting in scheduling failures.

JingsongLi:
@zhangjun0x01 Can we consider to use bucket number as default parallelism?

zhangjun0x01:
> I guess why there are test failures is because too much parallelism has been derived, resulting in scheduling failures.

Should we disable parallelism inference in default?

zhangjun0x01:
> @zhangjun0x01 Can we consider to use bucket number as streaming default parallelism? And use parallelism inference only for batch source.

ok

yuzelin:
The same problem I have met before. I've pushed some commits to try to solve the problem. You can rebase master.

zhangjun0x01:
> The same problem of E2E test I have met before. I've pushed some commits to try to solve the problem. You can rebase master.
yeah，I resubmitted

zhangjun0x01:
> @zhangjun0x01 Can we consider to use bucket number as streaming default parallelism? And use parallelism inference only for batch source.

I updated it, and disabled the `scan.infer-parallelism` in default . it is fine, It may be really caused by too much parallelism

flink-table-store-flink/flink-table-store-flink-common/src/main/java/org/apache/flink/table/store/connector/source/TableStoreSource.java
```diff
@@ -162,15 +164,49 @@ public ScanRuntimeProvider getScanRuntimeProvider(ScanContext scanContext) {
...
+        if (streaming) {
+            parallelism = options.get(CoreOptions.BUCKET);
+        }
+
+        if (options.containsKey(FlinkConnectorOptions.SCAN_PARALLELISM.key())) {
+            parallelism = options.get(FlinkConnectorOptions.SCAN_PARALLELISM);
+        }
+
+        // batch mode
+        if (options.get(FlinkConnectorOptions.INFER_SCAN_PARALLELISM) && !streaming) {
```
JingsongLi, line 189: How about:
```
if (options.get(FlinkConnectorOptions.INFER_SCAN_PARALLELISM)) {
   if (streaming) {
        parallelism = options.get(CoreOptions.BUCKET);
   } else {
        // infer parallelism for batch
   }
}
```

zhangjun0x01: I updated it

flink-table-store-flink/flink-table-store-flink-common/src/main/java/org/apache/flink/table/store/connector/source/TableStoreSource.java
```diff
@@ -162,15 +164,47 @@ public ScanRuntimeProvider getScanRuntimeProvider(ScanContext scanContext) {
...
+            FileStoreTable table, Predicate predicate, Long limitCount, boolean streaming) {
+        Options options = Options.fromMap(this.table.schema().options());
+        Integer parallelism = options.get(FlinkConnectorOptions.SCAN_PARALLELISM);
+
+        if (options.get(FlinkConnectorOptions.INFER_SCAN_PARALLELISM)) {
+            // for streaming mode, set the default parallelism to the bucket number.
+            if (streaming) {
+                parallelism = options.get(CoreOptions.BUCKET);
+            } else {
+                int splitSize = table.newScan().withFilter(predicate).plan().splits.size();
```
JingsongLi, line 184: I have concern here, plan is a heavy work, can we have a conservative approach?

JingsongLi: Maybe, without limit, just bucket number * 10?

zhangjun0x01: My idea is that `split` is related to the data size of table. If we use bucket size, I am worried that the data size of each bucket may be difficult to balance (if there is data skew, this situation will be more serious), so whether we can add a `totalDataSize` for snapshot,then we use `totalDataSize / 1G` to infer parallelism? Or the divisor can be specified by the user, depending on how much data the user wants each read operator to read?

In addition, the `totalDataSize` of snapshot is similar to `totalRecordCount`, which is also a metric of snapshot.

what do you think about?

JingsongLi: I think your thought is good~ But we should consider filter pushdown too, especially for partition&primary_key filter pushdown.

Another approach is doing some refactor. We can produce splits in compilation phase, and pass these splits to runtime source.

zhangjun0x01: > I think your thought is good~ But we should consider filter pushdown too, especially for partition&primary_key filter pushdown.


 yes, if there are query conditions, the `split` will be different

zhangjun0x01: > Another approach is doing some refactor. We can produce splits in compilation phase, and pass these splits to runtime source.

Let me investigate it

zhangjun0x01: > I think your thought is good~ But we should consider filter pushdown too, especially for partition&primary_key filter pushdown.
> 
> Another approach is doing some refactor. We can produce splits in compilation phase, and pass these splits to runtime source.

Hi,@JingsongLi, I generated the split in `FlinkTableSource`, do we need to add `snapshotId` so that we can refactor all the codes that use `scan` and `plan`,like here: https://github.com/apache/flink-table-store/blob/e715ac81a906e8f1b285d3847cc75879bb54d062/flink-table-store-flink/flink-table-store-flink-common/src/main/java/org/apache/flink/table/store/connector/source/StaticFileStoreSource.java#L86

flink-table-store-flink/flink-table-store-flink-common/src/main/java/org/apache/flink/table/store/connector/FlinkConnectorOptions.java
```diff
@@ -82,6 +82,20 @@ public class FlinkConnectorOptions {
                                     + "By default, if this option is not defined, the planner will derive the parallelism "
                                     + "for each statement individually by also considering the global configuration.");
 
+    public static final ConfigOption<Boolean> INFER_SCAN_PARALLELISM =
+            ConfigOptions.key("scan.infer-parallelism")
```
SteNicholas, line 86: IMO, user should not configure `scan.infer-parallelism` because inferring parallelism is default behavior that infers parallelism when `scan.parallelism` isn't set, and use `scan.parallelism` when `scan.parallelism` is set.

flink-table-store-flink/flink-table-store-flink-common/src/main/java/org/apache/flink/table/store/connector/FlinkConnectorOptions.java
```diff
@@ -82,6 +82,20 @@ public class FlinkConnectorOptions {
...
+                    .booleanType()
+                    .defaultValue(false)
+                    .withDescription(
+                            "If it is false, parallelism of source are set by scan.parallelism. "
+                                    + "If it is true, source parallelism is inferred according to splits number (batch mode) or bucket number(streaming mode).");
+
+    public static final ConfigOption<Integer> INFER_SCAN_PARALLELISM_MAX =
+            ConfigOptions.key("scan.infer-parallelism.max")
```
SteNicholas, line 94: Source parallelism is inferred according to splits number (batch mode) or bucket number(streaming mode), therefore when does the user need to config the max of inferred parallelism?

zhangjun0x01: In order to prevent too many datafiles from causing excessive parallelism, we also set a max infer parallelism. When the infer parallelism exceeds the setting, use the max parallelism.

SteNicholas: How does the user configure this config? Meanwhile, the essence of the situation that many datafiles from causing excessive parallelism should be controlled by splits number and bucket number, instead of this config.

zhangjun0x01: the  split number is affected by `source.split.target-size` and `source.split.open-file-cost`, for most cases, it is no problem. If the table data is large, the inferred number of splits is too large, but the user does not want this job to occupy too many resources, so set an upper limit, so that the resource utilization of most tasks can be met in a reasonable range, and at the same time, the cluster resources will not be filled up because the split of a job is too large.

refer to `https://github.com/apache/flink/blob/cf358d7d55ca48b9d25e5217006898e3070a85ad/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveOptions.java#L61`

flink-table-store-flink/flink-table-store-flink-common/src/main/java/org/apache/flink/table/store/connector/FlinkConnectorOptions.java
zhangjun0x01: Yes, your suggestion is also a reasonable idea. I guess that add the `scan.parallelism` option is a simple way to let users set the scan  parallelism , but it is not intelligent enough, so we added the parallelism inference.

So another solution is that we can also use infer-parallelism instead of `scan.parallelism`. If we enable parallelism inference, we use split or bucket number to as job parallelism. If not, we use the default parallelism of the flink job.

But now that the system has added the `scan parallelism` option, I am worried that some users who are using this option will be affected, so I do not delete this option  and  add the switch for inferring parallelism . and make its priority higher than `scan.parallelism`.

flink-table-store-flink/flink-table-store-flink-common/src/main/java/org/apache/flink/table/store/connector/FlinkConnectorOptions.java
SteNicholas: @zhangjun0x01, the implementation of the Hive reference makes sense because hive connector couldn't control the split number, but table store could control splits number and bucket number. I don't think this solution could refer to Hive for the different situation between table table and hive.

flink-table-store-flink/flink-table-store-flink-common/src/main/java/org/apache/flink/table/store/connector/FlinkConnectorOptions.java
SteNicholas: @zhangjun0x01, as you mentioned, could we update the default value description of `scan.parallelism`? IMO, the default value of `scan.parallelism` could use the inferred parallelism, and also custom to config via user.

zhangjun0x01: ok，I update it

JingsongLi (commented):
Thanks @zhangjun0x01 for your update!

I think we don't need to modify core codes (`AbstractFileStoreTable.java`).
If splits are generated in `DataTableSource`, we can pass this splits into `FlinkSourceBuilder` and finally into `StaticFileStoreSource`.

flink-table-store-flink/flink-table-store-flink-common/src/main/java/org/apache/flink/table/store/connector/FlinkConnectorOptions.java
SteNicholas: @zhangjun0x01, +1.

flink-table-store-flink/flink-table-store-flink-common/src/main/java/org/apache/flink/table/store/connector/FlinkConnectorOptions.java
SteNicholas: @zhangjun0x01, you should also remove the config option `scan.infer-parallelism`, not only update the description of `scan.parallelism`.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkConnectorOptions.java
```diff
@@ -80,7 +80,16 @@ public class FlinkConnectorOptions {
                     .withDescription(
                             "Define a custom parallelism for the scan source. "
                                     + "By default, if this option is not defined, the planner will derive the parallelism "
-                                    + "for each statement individually by also considering the global configuration.");
+                                    + "for each statement individually by also considering the global configuration. "
+                                    + "If user enable the scan.infer-parallelism, the planner will derive the parallelism by infer-parallelism.");
```
SteNicholas, line 84:
```suggestion
                                    + "If user enable the "+ INFER_SCAN_PARALLELISM.key() +", the planner will derive the parallelism by infer-parallelism.");
```

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkConnectorOptions.java
```diff
@@ -80,7 +80,16 @@ public class FlinkConnectorOptions {
...
+
+    public static final ConfigOption<Boolean> INFER_SCAN_PARALLELISM =
+            ConfigOptions.key("scan.infer-parallelism")
+                    .booleanType()
+                    .defaultValue(false)
+                    .withDescription(
+                            "If it is false, parallelism of source are set by scan.parallelism. "
```
SteNicholas, line 91:
```suggestion
                            "If it is false, source parallelism is determined by "+ SCAN_PARALLELISM.key() +". "
```

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkConnectorOptions.java
```diff
@@ -80,7 +80,16 @@ public class FlinkConnectorOptions {
...
+                                    + "If it is true, source parallelism is inferred according to splits number (batch mode) or bucket number(streaming mode).");
```
SteNicholas, line 92:
```suggestion
                                    + "Otherwise, source parallelism is inferred from splits number (batch mode) or bucket number(streaming mode).");
```

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/DataTableSource.java
```diff
@@ -163,15 +167,44 @@ public ScanRuntimeProvider getScanRuntimeProvider(ScanContext scanContext) {
...
-                                Options.fromMap(table.schema().options())
-                                        .get(FlinkConnectorOptions.SCAN_PARALLELISM))
+                        .withParallelism(inferParallelism())
                         .withWatermarkStrategy(watermarkStrategy);
 
         return new PaimonDataStreamScanProvider(
                 !streaming, env -> sourceBuilder.withEnv(env).build());
     }
 
+    private Integer inferParallelism() {
```
SteNicholas, line 177:
```suggestion
    private Integer scanParallelism() {
```

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/DataTableSource.java
```diff
@@ -163,15 +167,44 @@ public ScanRuntimeProvider getScanRuntimeProvider(ScanContext scanContext) {
...
+        Options options = Options.fromMap(this.table.schema().options());
+        Integer parallelism = options.get(FlinkConnectorOptions.SCAN_PARALLELISM);
+
+        if (options.get(FlinkConnectorOptions.INFER_SCAN_PARALLELISM)) {
+            // for streaming mode, set the default parallelism to the bucket number.
+            if (streaming) {
+                parallelism = options.get(CoreOptions.BUCKET);
+            } else {
+                int splitSize = 1;
```
SteNicholas, line 186: Why set `splitSize` to 1?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/DataTableSource.java
```diff
@@ -163,15 +167,44 @@ public ScanRuntimeProvider getScanRuntimeProvider(ScanContext scanContext) {
...
+                parallelism = splitSize;
+
+                if (null != limit && limit > 0) {
+                    int limitCount =
+                            limit >= Integer.MAX_VALUE ? Integer.MAX_VALUE : limit.intValue();
+                    parallelism = Math.min(parallelism, limitCount);
+                }
+
+                // parallelism must be positive.
+                parallelism = Math.max(1, parallelism);
```
SteNicholas, line 202: Why get the max between 1 and parallelism?

SteNicholas:
@zhangjun0x01, could you resolve the conflicts of above conflicting files?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkConnectorOptions.java
zhangjun0x01: the `INFER_SCAN_PARALLELISM` is defined after `SCAN_PARALLELISM`,so we can not add the `INFER_SCAN_PARALLELISM`  to the description of `SCAN_PARALLELISM`.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/DataTableSource.java
zhangjun0x01: for empty ,the split will be empty, my initial idea was init the `splitSize ` to 1 to guarantee inferred parallelism should be at least 1

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/DataTableSource.java
zhangjun0x01: To prevent the user set an incorrect value for `scan.parallelism`, for -1 or 0.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/DataTableSource.java
```diff
@@ -191,22 +195,54 @@ public ScanRuntimeProvider getScanRuntimeProvider(ScanContext scanContext) {
             }
         }
 
+        List<Split> splits = null;
+        DataTableScan.DataFilePlan dataFilePlan = table.newScan().plan();
+        // TODO: we will refactor this after https://github.com/apache/incubator-paimon/issues/638
+        // finished.
+        if (null != dataFilePlan) {
+            splits = dataFilePlan.splits();
```
JingsongLi, line 203: There is also a cost to generating splits at the compile stage, it causes the generated topology to become large if there are a lot of splits.
Can we just generate splits when necessary and keep the original process without inference?

zhangjun0x01: @JingsongLi I update it

JingsongLi:
You can rebase and update.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/DataTableSource.java
```diff
@@ -198,13 +202,45 @@ public ScanRuntimeProvider getScanRuntimeProvider(ScanContext scanContext) {
...
+                    Tuple2<Integer, List<Split>> scanContent = scanContent();
+                    return sourceBuilder
+                            .withParallelism(scanContent.f0)
+                            .withSplits(scanContent.f1)
+                            .withEnv(env)
+                            .build();
+                });
+    }
+
+    private Tuple2<Integer, List<Split>> scanContent() {
```
JingsongLi, line 219: maybe `void configureSource(SourceBuilder source)` here?

zhangjun0x01: I update it and fix the conflict

zhangjun0x01:
image

JingsongLi (approved):
+1

PR #585, opened 2023-03-06
[FLINK-31343] Remove JMH dependency in flink-table-store-micro-benchmark

JingsongLi:
JMH open source license is incompatible


FangYongs:
Thanks @JingsongLi  LGTM +1

PR #588, opened 2023-03-10
[FLINK-31389] Fix spark jar name in docs for table store

zhuangchong:
Fix spark jar name in docs for table store


JingsongLi (approved):
Thanks @zhuangchong for your contribution!
Looks good to me!

PR #591, opened 2023-03-10
[FLINK-31394] Fix spark jar name in the create release script for table store

zhuangchong:
Fix spark jar name in the create release script for table store


JingsongLi (approved):
Looks good to me!

PR #592, opened 2023-03-10
[FLINK-31292] Introduce HadoopUtils to get Configuration in CatalogContext

liyubin117:
At present, if HadoopConf is not passed in the CatalogContext, a new HadoopConf will be directly generated, which may not have the required parameters. we should correct it.

### DIFF
--- a/flink-table-store-common/src/main/java/org/apache/flink/table/store/utils/HadoopConfUtils.java
+++ b/flink-table-store-common/src/main/java/org/apache/flink/table/store/utils/HadoopConfUtils.java
@@ -1,0 +1,112 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.table.store.utils;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hdfs.HdfsConfiguration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.File;
+
+/**
+ * Utility class for working with Hadoop-related classes. This should only be used if Hadoop is on
+ * the classpath.
+ */
+public class HadoopConfUtils {
+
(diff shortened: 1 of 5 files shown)


flink-table-store-common/src/main/java/org/apache/flink/table/store/utils/HadoopConfUtils.java
```diff
@@ -0,0 +1,112 @@
...
+ * Utility class for working with Hadoop-related classes. This should only be used if Hadoop is on
+ * the classpath.
+ */
+public class HadoopConfUtils {
+
+    private static final Logger LOG = LoggerFactory.getLogger(HadoopConfUtils.class);
+    public static final String HADOOP_HOME_ENV = "HADOOP_HOME";
+    public static final String HADOOP_CONF_ENV = "HADOOP_CONF_DIR";
+
+    public static Configuration getHadoopConfiguration() {
```
JingsongLi, line 38: We can pass `Options` in here to support some config options like Flink HadoopUtils.

liyubin117: Thanks for your reminds, I wonder if it is reasonable to get extra options like Flink HadoopUtils, as of the feature in common module invoked by Spark/Hive should not rely on Fllink configuration.

JingsongLi: I think these should be our configuration, `Approach 2: Flink configuration (deprecated)` is deprecated, but approach 4 may looks good:
```
// Approach 4: catalog options
// add all configuration key with prefix 'hadoop.' in catalog options to hadoop conf
```
Maybe like this?

liyubin117: I get it~ to make it clear, do you mean make it like `create(options, HadoopConfUtils.getHadoopConfiguration(options))`

JingsongLi: Yes~

flink-table-store-common/src/main/java/org/apache/flink/table/store/utils/HadoopUtils.java
```diff
@@ -0,0 +1,133 @@
...
+import java.io.File;
+
+/**
+ * Utility class for working with Hadoop-related classes. This should only be used if Hadoop is on
+ * the classpath. Note: decoupled from specific engines.
+ */
+public class HadoopUtils {
+
+    private static final Logger LOG = LoggerFactory.getLogger(HadoopUtils.class);
+    private static final String[] FLINK_CONFIG_PREFIXES = {"flink.hadoop."};
```
JingsongLi, line 37: Don't use `flink.`.
`String[] FLINK_CONFIG_PREFIXES = {"flink.hadoop."}` -> `String[] CONFIG_PREFIXES = {"hadoop."}`.

liyubin117: done :)

JingsongLi (commented):
Can you add more tests, you can refer to Flink `HadoopConfigLoadingTest`.

liyubin117:
@JingsongLi I have added more tests, please take a look

JingsongLi (approved):
+1

PR #593, opened 2023-03-10
[FLINK-31397] Introduce write-once hash lookup store

JingsongLi:
Introduce interface for lookup changelog producer:
```
/**
 * A key-value store for lookup, key-value store should be single binary file written once and ready
 * to be used. This factory provide two interfaces:
 *
 * <ul>
 *   <li>Writer: written once to prepare binary file.
 *   <li>Reader: lookup value by key bytes.
 * </ul>
 */
public interface LookupStoreFactory {

    LookupStoreWriter createWriter(File file) throws IOException;

    LookupStoreReader createReader(File file) throws IOException;
}
```
We can convert remote columnar data to local lookup store, and ready to be used to lookup.

### DIFF
--- a/flink-table-store-common/src/main/java/org/apache/flink/table/store/lookup/LookupStoreReader.java
+++ b/flink-table-store-common/src/main/java/org/apache/flink/table/store/lookup/LookupStoreReader.java
@@ -24,5 +24,6 @@ import java.io.IOException;
 /** Reader, lookup value by key bytes. */
 public interface LookupStoreReader extends Closeable {
 
+    /** Lookup value by key. */
     byte[] lookup(byte[] key) throws IOException;
 }
--- a/flink-table-store-common/src/main/java/org/apache/flink/table/store/lookup/LookupStoreReader.java
+++ b/flink-table-store-common/src/main/java/org/apache/flink/table/store/lookup/LookupStoreReader.java
@@ -1,0 +1,28 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.table.store.lookup;
+
+import java.io.Closeable;
+import java.io.IOException;
+
+/** Reader, lookup value by key bytes. */
+public interface LookupStoreReader extends Closeable {
+
+    byte[] lookup(byte[] key) throws IOException;
+}
--- a/flink-table-store-common/src/main/java/org/apache/flink/table/store/lookup/LookupStoreWriter.java
+++ b/flink-table-store-common/src/main/java/org/apache/flink/table/store/lookup/LookupStoreWriter.java
@@ -24,5 +24,6 @@ import java.io.IOException;
 /** Writer to prepare binary file. */
 public interface LookupStoreWriter extends Closeable {
 
+    /** Put key value to store. */
     void put(byte[] key, byte[] value) throws IOException;
 }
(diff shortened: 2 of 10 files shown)


flink-table-store-common/src/main/java/org/apache/flink/table/store/lookup/LookupStoreReader.java
```diff
@@ -0,0 +1,28 @@
...
+
+package org.apache.flink.table.store.lookup;
+
+import java.io.Closeable;
+import java.io.IOException;
+
+/** Reader, lookup value by key bytes. */
+public interface LookupStoreReader extends Closeable {
+
+    byte[] lookup(byte[] key) throws IOException;
```
SteNicholas, line 27: Adds the java doc for interface methods.

flink-table-store-common/src/main/java/org/apache/flink/table/store/lookup/LookupStoreWriter.java
```diff
@@ -0,0 +1,28 @@
...
+
+package org.apache.flink.table.store.lookup;
+
+import java.io.Closeable;
+import java.io.IOException;
+
+/** Writer to prepare binary file. */
+public interface LookupStoreWriter extends Closeable {
+
+    void put(byte[] key, byte[] value) throws IOException;
```
SteNicholas, line 27: Ditto.

flink-table-store-common/src/main/java/org/apache/flink/table/store/lookup/hash/HashLookupStoreWriter.java
```diff
@@ -0,0 +1,486 @@
...
+        LOG.info("Number of keys: {}", keyCount);
+        LOG.info("Number of values: {}", valueCount);
+
+        // Prepare files to merge
+        List<File> filesToMerge = new ArrayList<>();
+
+        try {
+
+            // Write metadata file
+            File metadataFile = new File(tempFolder, "metadata.dat");
```
SteNicholas, line 178: Why do the process including writing metadata file, building index file, statsing collisions, adding data files, and merging and writing to output execute in close? Is the close operation heavy?

flink-table-store-common/src/main/java/org/apache/flink/table/store/lookup/hash/HashLookupStoreWriter.java
```diff
@@ -0,0 +1,486 @@
...
+                                + slotSize
+                                + " bytes";
+
+                LOG.info("Built index file {}\n" + msg, indexFile.getName());
+            } finally {
+                // Close input
+                tempIndexStream.close();
+
+                // Close index and make sure resources are liberated
+                indexChannel.close();
```
SteNicholas, line 349: Does the `byteBuffer` need to clean?

flink-table-store-common/src/main/java/org/apache/flink/table/store/lookup/hash/HashLookupStoreWriter.java
```diff
@@ -0,0 +1,486 @@
...
+            dos = new DataOutputStream(new BufferedOutputStream(new FileOutputStream(file)));
+            indexStreams[keyLength] = dos;
+
+            dataLengths[keyLength]++;
+        }
+        return dos;
+    }
+
+    private int getNumKeyCount() {
+        int res = 0;
```
SteNicholas, line 478:
```
return Arrays.stream(keyCounts).filter(c -> c != 0).count;
```

flink-table-store-common/src/main/java/org/apache/flink/table/store/utils/SimpleReadBuffer.java
```diff
@@ -0,0 +1,43 @@
...
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package org.apache.flink.table.store.utils;
+
+/** A simple read buffer provide {@code readUnsignedByte} and position. */
```
SteNicholas, line 17:
```suggestion
/** A simple read buffer provides {@code readUnsignedByte} and position. */
```

flink-table-store-common/src/main/java/org/apache/flink/table/store/lookup/hash/HashLookupStoreWriter.java
JingsongLi: Closing just copy files into one file. It is not so heavy.

flink-table-store-common/src/main/java/org/apache/flink/table/store/lookup/hash/HashLookupStoreWriter.java
JingsongLi: `FileChannel` is closed, there is no close in `byteBuffer`.

flink-table-store-common/src/main/java/org/apache/flink/table/store/lookup/hash/HashLookupStoreWriter.java
JingsongLi: In performance critical path, we don't use java stream.

SteNicholas (approved):
LGTM.

PR #594, opened 2023-03-12
[core] Introduce LookupMergeFunction and ForceUpLevel0Compaction

JingsongLi:
### Introduce `LookupMergeFunction`
When Lookup Changelog producer is enabled, each merge will query the old merged record, so the latest high level record should be the final merged value.

### Introduce `ForceUpLevel0Compaction`
Because changelog is generated when level 0 is merged to a higher level, we need a compaction policy to ensure that changelog is generated each time.

### Introduce `LookupChangelogMergeFunctionWrapper`
Changelog records are generated in the process of the level-0 file participating in the compaction, if during the compaction processing:
- Without level-0 records, no changelog.
- With level-0 record, with level-x (x > 0) record, level-x record should be BEFORE, level-0 should be AFTER.
- With level-0 record, without level-x record, need to lookup the history value of the upper level as BEFORE.

### DIFF
--- a/flink-table-store-core/src/main/java/org/apache/flink/table/store/file/mergetree/compact/ForceUpLevel0Compaction.java
+++ b/flink-table-store-core/src/main/java/org/apache/flink/table/store/file/mergetree/compact/ForceUpLevel0Compaction.java
@@ -1,0 +1,59 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.table.store.file.mergetree.compact;
+
+import org.apache.flink.table.store.file.compact.CompactUnit;
+import org.apache.flink.table.store.file.mergetree.LevelSortedRun;
+
+import java.util.List;
+import java.util.Optional;
+
+/** A {@link CompactStrategy} to force compacting level 0 files. */
+public class ForceUpLevel0Compaction implements CompactStrategy {
+
+    private final UniversalCompaction universalCompaction;
+
+    public ForceUpLevel0Compaction(UniversalCompaction universalCompaction) {
+        this.universalCompaction = universalCompaction;
+    }
+
+    @Override
+    public Optional<CompactUnit> pick(int numLevels, List<LevelSortedRun> runs) {
+        Optional<CompactUnit> pick = universalCompaction.pick(numLevels, runs);
+        if (pick.isPresent()) {
+            return pick;
+        }
+
+        if (runs.isEmpty() || runs.get(0).level() > 0) {
+            return Optional.empty();
+        }
+
+        // collect all level 0 files
+        int candidateCount = 1;
+        for (int i = candidateCount; i < runs.size(); i++) {
+            if (runs.get(i).level() > 0) {
+                break;
+            }
+            candidateCount++;
+        }
+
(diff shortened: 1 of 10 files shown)


flink-table-store-core/src/main/java/org/apache/flink/table/store/file/mergetree/compact/ForceUpLevel0Compaction.java
```diff
@@ -0,0 +1,59 @@
...
+        if (pick.isPresent()) {
+            return pick;
+        }
+
+        if (runs.isEmpty() || runs.get(0).level() > 0) {
+            return Optional.empty();
+        }
+
+        // collect all level 0 files
+        int candidateCount = 1;
```
SteNicholas, line 48: Why not `int candidateCount = runs.stream().filter(r->r.level()>0).count();`?

flink-table-store-core/src/main/java/org/apache/flink/table/store/file/mergetree/compact/LookupChangelogMergeFunctionWrapper.java
```diff
@@ -0,0 +1,142 @@
...
+    private final KeyValue reusedBefore = new KeyValue();
+    private final KeyValue reusedAfter = new KeyValue();
+
+    public LookupChangelogMergeFunctionWrapper(
+            MergeFunctionFactory<KeyValue> mergeFunctionFactory,
+            Function<InternalRow, KeyValue> lookup) {
+        MergeFunction<KeyValue> mergeFunction = mergeFunctionFactory.create();
+        checkArgument(
+                mergeFunction instanceof LookupMergeFunction,
+                "Merge function should be a  LookupMergeFunction, but is %s, there is a bug.",
```
SteNicholas, line 60:
```suggestion
                "Merge function should be a LookupMergeFunction, but is %s, there is a bug.",
```

flink-table-store-core/src/main/java/org/apache/flink/table/store/file/mergetree/compact/LookupMergeFunction.java
```diff
@@ -0,0 +1,99 @@
...
+
+import org.apache.flink.table.store.file.KeyValue;
+
+import javax.annotation.Nullable;
+
+import java.util.Iterator;
+import java.util.LinkedList;
+
+/**
+ * A {@link MergeFunction} for lookup, this wrapper only consider the latest high level record,
```
SteNicholas, line 29:
```suggestion
 * A {@link MergeFunction} for lookup, this wrapper only considers the latest high level record,
```

flink-table-store-core/src/main/java/org/apache/flink/table/store/file/mergetree/compact/aggregate/AggregateMergeFunction.java
```diff
@@ -172,10 +133,30 @@ public MergeFunction<KeyValue> create(@Nullable int[][] projection) {
...
+                                key(FIELDS + "." + fieldName + "." + AGG_FUNCTION)
+                                        .stringType()
+                                        .noDefaultValue()
+                                        .withDescription(
+                                                "Get " + fieldName + "'s aggregate function"));
+                boolean ignoreRetract =
+                        conf.get(
+                                key(FIELDS + "." + fieldName + "." + IGNORE_RETRACT)
+                                        .booleanType()
+                                        .defaultValue(false));
```
SteNicholas, line 153: Adds the description?

flink-table-store-core/src/main/java/org/apache/flink/table/store/file/mergetree/compact/ForceUpLevel0Compaction.java
JingsongLi: I think we can finish this in one loop:
```
// collect all level 0 files
        int candidateCount = 0;
        for (int i = candidateCount; i < runs.size(); i++) {
            if (runs.get(i).level() > 0) {
                break;
            }
            candidateCount++;
        }

        return candidateCount == 0 ?
                Optional.empty() :
                Optional.of(universal.pickForSizeRatio(numLevels - 1, runs, candidateCount, true));
```

flink-table-store-core/src/main/java/org/apache/flink/table/store/file/mergetree/compact/aggregate/AggregateMergeFunction.java
JingsongLi: I delete description of `AGG_FUNCTION`, here the description is useless, it can not be see by users.

SteNicholas (approved):
LGTM.

PR #598, opened 2023-03-14
[FLINK-31436] Remove schemaId from constructor of FileStoreCommitImpl and ManifestFile in Table Store

tsreaper:
As schema may change during a CTAS streaming job, the schema ID of snapshots and manifest files may also change. We should remove `schemaId` from their constructor and calculate the real `schemaId` on the fly.


JingsongLi (approved):
+1

PR #600, opened 2023-03-14
[FLINK-31431] Support copying a FileStoreTable with latest schema

tsreaper:
To capture schema changes, CDC sinks of Flink Table Store should be able to use the latest schema at any time. This requires us to copy a `FileStoreTable` with latest schema so that we can create `TableWrite` with latest schema.


JingsongLi (approved):
+1

PR #569, opened 2023-03-02
[FLINK-31269] Split hive connector to each module of each version

FangYongs:
This pr aims to split hive connector to each module of each version. 

Besides connector, there are different versions of hive catalog (hive 2 and 3). Hive3 metastore can be connected through the client of hive2.3.9. We do not need to create independent modules for catalog. If necessary, we can also divide the hive catalog into sub-modules in the future.

PR #601, opened 2023-03-14
[FLINK-31440] Split hive catalog to each module of each version

zhuangchong:
Split hive catalog to each module of each version


JingsongLi:
Hi @zhuangchong Thanks for the contribution.
In https://github.com/apache/flink-table-store/pull/569 , we found hive catalog with hive-2.3.9 can be used for every hive metastore versions, so we just keep one version for hive catalog.

zhuangchong:
> Hi @zhuangchong Thanks for the contribution. In #569 , we found hive catalog with hive-2.3.9 can be used for every hive metastore versions, so we just keep one version for hive catalog.

Thanks for your reminder.

PR #603, opened 2023-03-15
[FLINK-31459]add UPDATE COLUMN POSITION for flink table store

zhangjun0x01:
add UPDATE COLUMN POSITION for flink table store

### DIFF
--- a/flink-table-store-core/src/main/java/org/apache/flink/table/store/file/schema/SchemaManager.java
+++ b/flink-table-store-core/src/main/java/org/apache/flink/table/store/file/schema/SchemaManager.java
@@ -29,6 +29,7 @@ import org.apache.flink.table.store.file.schema.SchemaChange.RenameColumn;
 import org.apache.flink.table.store.file.schema.SchemaChange.SetOption;
 import org.apache.flink.table.store.file.schema.SchemaChange.UpdateColumnComment;
 import org.apache.flink.table.store.file.schema.SchemaChange.UpdateColumnNullability;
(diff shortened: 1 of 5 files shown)


flink-table-store-core/src/main/java/org/apache/flink/table/store/file/schema/SchemaManager.java
```diff
@@ -310,6 +311,27 @@ public TableSchema commitChanges(List<SchemaChange> changes) throws Exception {
...
+
+                    int fieldIndex = map.get(move.fieldName());
+                    int refIndex = 0;
+                    if (move.type().equals(SchemaChange.Move.MoveType.FIRST)) {
+                        checkMoveIndexEqual(move, fieldIndex, refIndex);
+                        newFields.add(refIndex, newFields.remove(fieldIndex));
+                    } else if (move.type().equals(SchemaChange.Move.MoveType.AFTER)) {
+                        refIndex = map.get(move.referenceFieldName());
+                        checkMoveIndexEqual(move, fieldIndex, refIndex);
+                        newFields.add(refIndex + 1, newFields.remove(fieldIndex));
```
JingsongLi, line 332: Maybe bug here?
Total has 4 fields.
fieldIndex = 1;
refIndex = 2;
f0, f1, f2, f3
after remove: f0 f2 f3
and add(3)? becoming f0 f2 f3 f1?

JingsongLi (approved):
Thanks for the update, looks good to me!

PR #604, opened 2023-03-15
[FLINK-31460] Fix hive catalog and connector jar name in the create release script for table store

zhuangchong:
Fix hive catalog and connector jar name in the create release script for table store


JingsongLi (approved):
Thanks @zhuangchong for the contribution, +1

PR #605, opened 2023-03-15
[FLINK-31430] Support migrating states between different instances of TableWriteImpl and AbstractFileStoreWrite

tsreaper:
Currently `Table` and `TableWrite` in Flink Table Store have a fixed schema. However to consume schema changes, Flink Table Store CDC sinks should have the ability to change its schema during a streaming job.

This require us to pause and store the states of a `TableWrite`, then create a `TableWrite` with newer schema and recover the states in the new `TableWrite`.

### DIFF
--- a/flink-table-store-core/src/main/java/org/apache/flink/table/store/file/operation/AbstractFileStoreWrite.java
+++ b/flink-table-store-core/src/main/java/org/apache/flink/table/store/file/operation/AbstractFileStoreWrite.java
@@ -24,6 +24,7 @@ import org.apache.flink.table.store.file.Snapshot;
 import org.apache.flink.table.store.file.disk.IOManager;
 import org.apache.flink.table.store.file.io.DataFileMeta;
 import org.apache.flink.table.store.file.manifest.ManifestEntry;
+import org.apache.flink.table.store.file.utils.CommitIncrement;
 import org.apache.flink.table.store.file.utils.ExecutorThreadFactory;
 import org.apache.flink.table.store.file.utils.RecordWriter;
 import org.apache.flink.table.store.file.utils.SnapshotManager;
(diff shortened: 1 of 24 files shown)


flink-table-store-core/src/main/java/org/apache/flink/table/store/file/operation/AbstractFileStoreWrite.java
```diff
@@ -232,50 +210,131 @@ public void close() throws Exception {
         }
     }
 
+    @Override
+    public List<State> extractStateAndClose() throws Exception {
```
JingsongLi, line 214: Can we just have two methods: `checkpoint` and `close`?

flink-table-store-core/src/main/java/org/apache/flink/table/store/file/operation/AbstractFileStoreWrite.java
```diff
@@ -232,50 +210,131 @@ public void close() throws Exception {
...
+        }
+
+        if (LOG.isDebugEnabled()) {
+            LOG.debug("Extracted state " + result.toString());
+        }
+        return result;
+    }
+
+    @Override
+    public void recoverFromState(List<State> states) {
```
JingsongLi, line 249: `restore`?

flink-table-store-core/src/main/java/org/apache/flink/table/store/file/utils/RecordWriter.java
```diff
@@ -51,6 +50,9 @@
      */
     void addNewFiles(List<DataFileMeta> files);
 
+    /** Get all files maintained by this writer. */
+    List<DataFileMeta> allFiles();
```
JingsongLi, line 54: allFiles ->  dataFiles?

flink-table-store-core/src/main/java/org/apache/flink/table/store/file/operation/AbstractFileStoreWrite.java
```diff
@@ -232,50 +210,131 @@ public void close() throws Exception {
...
+    public List<State> extractStateAndClose() throws Exception {
+        List<State> result = new ArrayList<>();
+        for (Map.Entry<BinaryRow, Map<Integer, WriterContainer<T>>> partitionEntry :
+                writers.entrySet()) {
+            BinaryRow partition = partitionEntry.getKey();
+            for (Map.Entry<Integer, WriterContainer<T>> bucketEntry :
+                    partitionEntry.getValue().entrySet()) {
+                int bucket = bucketEntry.getKey();
+                WriterContainer<T> writerContainer = bucketEntry.getValue();
+                CommitIncrement increment = writerContainer.writer.extractStateAndClose();
```
JingsongLi, line 223: We don't need to introduce `extractStateAndClose` to writer, we can just:
```
1. CommitIncrement increment = writer. prepareCommit...;
2. List<DataFileMeta> dataFiles = writer.dataFiles();
3. writer.close();
```
We should get data files before close.

JingsongLi (approved):
+1

PR #608, opened 2023-03-15
[FLINK-31467] Support time travel for Spark 3.3

yuzelin:
1. Add a new internal option `time-travel.snapshot-id` for time travel. When do table#copy, if this option is set, change schema to specified version and set `StartupMode` = `from-snapshot` to scan specified snapshot.
2. Implement new `loadTable` in Spark 3.3


docs/content/docs/how-to/querying-tables.md
```diff
@@ -91,6 +91,33 @@ Users can also adjust `changelog-producer` table property to specify the pattern
 Streaming Source can also be bounded, you can specify 'scan.bounded.watermark' to define the end condition for bounded streaming mode, stream reading will end until a larger watermark snapshot is encountered.
 {{< /hint >}}
 
+## Time Travel
+
+Currently, Table Store supports time travel for Flink and Spark 3.3.
+
+For flink, you can use [dynamic table options](https://nightlies.apache.org/flink/flink-docs-release-1.16/docs/dev/table/sql/queries/hints/#dynamic-table-options) to specify scan mode and from where to start:
```
JingsongLi, line 98: Use `tabs`?

flink-table-store-core/src/main/java/org/apache/flink/table/store/CoreOptions.java
```diff
@@ -524,6 +524,13 @@ public class CoreOptions implements Serializable {
                     .defaultValue(MemorySize.parse("256 mb"))
                     .withDescription("Max memory size for lookup cache.");
 
+    @ExcludeFromDocumentation("Internal use only")
+    public static final ConfigOption<Long> TIME_TRAVEL_SNAPSHOT_ID =
```
JingsongLi, line 528: Why not just use `scan.snapshot-id`?

flink-table-store-spark/flink-table-store-spark-common/src/main/java/org/apache/flink/table/store/spark/SparkCatalog.java
```diff
@@ -18,354 +18,7 @@
...
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.stream.Collectors;
-
-import static org.apache.flink.table.store.spark.SparkTypeUtils.toFlinkType;
-
-/** Spark {@link TableCatalog} for table store. */
-public class SparkCatalog implements TableCatalog, SupportsNamespaces {
```
JingsongLi, line 66: Can we make `SparkCatalog` abstract, and introduce a new `SparkCatalogImpl`?

flink-table-store-core/src/main/java/org/apache/flink/table/store/table/AbstractFileStoreTable.java
```diff
@@ -179,4 +188,25 @@ public TableCommitImpl newCommit(String commitUser) {
                 options().writeOnly() ? null : store().newExpire(),
                 options().writeOnly() ? null : store().newPartitionExpire(commitUser));
     }
+
+    private void tryTimeTravel(Options options) {
+        if (options.contains(CoreOptions.TIME_TRAVEL_SNAPSHOT_ID)) {
```
JingsongLi, line 193: In dynamicOptions, but not equals in old options, do time travel?

yuzelin: If we keep using `TIME_TRAVEL_SNAPSHOT_ID`, it's not a problem because when a table is created, this option won't appear, and if a time travel call pass this option, here it'll be cleaned.

If we use `scan.from-snapshot` and `scan.from-timstamp` as time travel mark, I think we should always use new value (new snapshotId or timestamp) as target. Furthermore, I think when the `TableScan` scan to next snapshot, the `TableSchema` should also be replaced to that snapshot's schema. 

What do you think?

flink-table-store-spark/flink-table-store-spark-common/src/main/java/org/apache/flink/table/store/spark/SparkCatalog.java
yuzelin: If I'm not mistaken, make `SparkCatalog` abstract, then we should specify `SparkCatalogImpl` when start spark env. Not seeing much difference between `abstract SparkCatalog`, `SparkCatalogImpl`  and `abstract SparkCatalogBase`, `SparkCatalog`.

JingsongLi: My bad, it is bad idea, `SparkCatalog` is the api for spark.

flink-table-store-core/src/main/java/org/apache/flink/table/store/table/AbstractFileStoreTable.java
JingsongLi: I may should know why we need to introduce `TIME_TRAVEL_SNAPSHOT_ID`.

PR #609, opened 2023-03-15
[core] Introduce RowCompactedSerializer to compact bytes in LookupLevels

JingsongLi:
Currently, there is no compression in `HashLookupStoreFactory`, we can introduce a compacted serializer to reduce disk usage.


yuzelin:
LGTM!

PR #616, opened 2023-03-16
[FLINK-31451] Introduce Presto Reader for table store

s7monk:
This PR aims to introduce presto reader for table store.The main change is to add the flink-table-store-presto module。

The Presto reader connector currently supports presto 0.236 and above.


s7monk:
@JingsongLi @SteNicholas PTAL

JingsongLi (commented):
Thanks @s7monk for the contribution!
Let's use new package name.

PR #619, opened 2023-03-17
[FLINK-31432] Introduce a special StoreWriteOperator to deal with schema changes

tsreaper:
Currently `StoreWriteOperator` is not able to deal with schema changes. We need to introduce a special `StoreWriteOperator` to deal with schema changes.


tsreaper:
I'll create a new pull request for this due to project name changes.

Issue #620, opened 2023-03-17
Move document from nightlies.apache.org/flink to paimon.apache.org/docs


tsreaper:
As Paimon now becomes an incubator project, we should move document out of nightlies.apache.org/flink and into paimon.apache.org/docs.

Task List:
- [ ] Disable document build in incubator-paimon.
- [ ] Build document in incubator-paimon-website instead.

PR #621, opened 2023-03-17
[doc] Disable document build in incubator-paimon and change document base url

tsreaper:
This PR is a subtask of #620 .

To move document out of nightlies.apache.org/flink and into paimon.apache.org/docs, we will first need to Disable document build in incubator-paimon and change document base url.


JingsongLi (approved):
+1

PR #624, opened 2023-03-17
[FLINK-31462] Supports full calculation from the specified snapshot in streaming mode

liming30:
### Purpose

When consuming multiple tables, it may be necessary to align data versions in different tables to ensure version consistency. However, the current stream mode only supports incremental processing from the specified snapshot, and does not support full calculation of the specified snapshot first, and then incremental calculation.

### Brief change

A new startup mode **FROM_SNAPSHOT_FULL** is added to support full calculation of the specified snapshot of data in streaming mode, and then incrementally process the newly added data to ensure the consistency of the data version when processing multiple tables.


JingsongLi (commented):
Thanks @liming30 for the contribution, can you add a test for this?

liming30:
@JingsongLi Sorry for replying now, I added a `StartupModeTest` to verify that the plan generated by each `StartupMode` is correct.

JingsongLi:
Hi @liming30 , test failed, you can find action in your repo: https://github.com/liming30/flink-table-store/actions

liming30:
@JingsongLi hi, I have fixed the test, can you please help to take a look?

JingsongLi (approved):
+1

PR #627, opened 2023-03-17
[docs] Move document pages from content/docs to content/

tsreaper:
We now change document base url to https://paimon.apache.org/docs/master. If the document pages still stays in `content/docs`, the url for a document page will become https://paimon.apache.org/docs/master/docs/xxxxx (note the two "docs" here) which is strange.


JingsongLi (approved):
+1

PR #644, opened 2023-03-19
[PAIMON-643] Change the import package to start with org.apache.paimon

### LINKED ISSUE #643: [Improvement] Change the import package to start with `org.apache.paimon`

zhuangchong:
this pr closes #643 

Changing the file is to modify the import order, without modifying the logic code.


JingsongLi (approved):
+1 Thanks for the contribution.

PR #645, opened 2023-03-19
minor: correct README.md

jackwener:


JingsongLi (commented):
Thanks for the contribution, this have been fixed in https://github.com/apache/incubator-paimon/pull/631

PR #646, opened 2023-03-19
[core] add read batch size option

zhangjun0x01:
related to https://github.com/apache/incubator-paimon/issues/641

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/format/FileFormat.java
+++ b/paimon-common/src/main/java/org/apache/paimon/format/FileFormat.java
@@ -18,7 +18,8 @@
 
 package org.apache.paimon.format;
 
-import org.apache.paimon.options.ConfigOption;
+import org.apache.paimon.annotation.VisibleForTesting;
+import org.apache.paimon.format.FileFormatFactory.FormatContext;
 import org.apache.paimon.options.Options;
 import org.apache.paimon.predicate.Predicate;
 import org.apache.paimon.types.RowType;
@@ -75,20 +76,18 @@ public abstract class FileFormat {
         return Optional.empty();
     }
 
-    /** Create a {@link FileFormat} from table options. */
-    public static FileFormat fromTableOptions(
-            Options tableOptions, ConfigOption<String> formatOption) {
-        String formatIdentifier = tableOptions.get(formatOption);
-        return fromIdentifier(formatIdentifier, tableOptions.removePrefix(formatIdentifier + "."));
+    @VisibleForTesting
+    public static FileFormat fromIdentifier(String identifier, Options options) {
+        return fromIdentifier(identifier, new FormatContext(options, 1024));
     }
 
     /** Create a {@link FileFormat} from format identifier and format options. */
-    public static FileFormat fromIdentifier(String identifier, Options options) {
+    public static FileFormat fromIdentifier(String identifier, FormatContext context) {
         Optional<FileFormat> format =
-                fromIdentifier(identifier, options, Thread.currentThread().getContextClassLoader());
+                fromIdentifier(identifier, context, Thread.currentThread().getContextClassLoader());
         return format.orElseGet(
                 () ->
-                        fromIdentifier(identifier, options, FileFormat.class.getClassLoader())
+                        fromIdentifier(identifier, context, FileFormat.class.getClassLoader())
                                 .orElseThrow(
                                         () ->
                                                 new RuntimeException(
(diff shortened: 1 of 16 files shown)


paimon-format/src/main/java/org/apache/paimon/format/orc/OrcFileFormat.java
```diff
@@ -62,13 +61,15 @@ public class OrcFileFormat extends FileFormat {
     private final Properties orcProperties;
     private final org.apache.hadoop.conf.Configuration readerConf;
     private final org.apache.hadoop.conf.Configuration writerConf;
+    private final int readBatchSize;
 
-    public OrcFileFormat(Options formatOptions) {
+    public OrcFileFormat(Options formatOptions, int readBatchSize) {
```
zhuangchong, line 66: I have a question? Why the `readBatchSize` attribute is not obtained from options.

zhangjun0x01: Yes, we can get `readBatchSize` by `option.get('read. batch-size')`. But I don't think we should do this way. We should use `CoreOptions#READ_BATCH_SIZE.key()`, but the `paimon-format` module does not reference the `paimon-core` module, so I pass this parameter from `paimon-core` to `paimon-format`

paimon-common/src/main/java/org/apache/paimon/format/FileFormatFactory.java
```diff
@@ -25,5 +25,5 @@ public interface FileFormatFactory {
 
     String identifier();
 
-    FileFormat create(Options formatOptions);
+    FileFormat create(Options formatOptions, int readBatchSize);
```
JingsongLi, line 28: Maybe just create a class `FormatContext`:
```
FormatContext {
    Options formatOptions();
    int readBatchSize();
}
```

paimon-common/src/main/java/org/apache/paimon/format/FileFormat.java
```diff
@@ -77,18 +78,35 @@ public Optional<FileStatsExtractor> createStatsExtractor(RowType type) {
 
     /** Create a {@link FileFormat} from table options. */
     public static FileFormat fromTableOptions(
-            Options tableOptions, ConfigOption<String> formatOption) {
+            Options tableOptions,
+            ConfigOption<String> formatOption,
+            ConfigOption<Integer> readBatchSizeOption) {
```
JingsongLi, line 83: Maybe we can remove this method.
Provide a util static method in `CoreOptions`.

paimon-common/src/main/java/org/apache/paimon/format/FileFormat.java
```diff
@@ -77,18 +78,35 @@ public Optional<FileStatsExtractor> createStatsExtractor(RowType type) {
...
     }
 
-    /** Create a {@link FileFormat} from format identifier and format options. */
+    @VisibleForTesting
     public static FileFormat fromIdentifier(String identifier, Options options) {
+        return fromIdentifier(identifier, 1000, options);
+    }
+
+    /** Create a {@link FileFormat} from format identifier and format options. */
+    public static FileFormat fromIdentifier(String identifier, int readBatchSize, Options options) {
```
JingsongLi, line 96: fromIdentifier(String identifier, FormatContext context)

JingsongLi (commented):
Thanks @zhangjun0x01 for the contribution, left some comments.

JingsongLi (commented):
Thanks for the update. Last comment: Can you add a test for format?
For example, you can create a format with identifier `test`, and validate options in create like:
```
public class TestFileFormatFactory implements FileFormatFactory {

    @Override
    public String identifier() {
        return "test";
    }

    @Override
    public FileFormat create(FormatContext context) {
          assertThat(context.options().get(myKey)).isEqualTo(myValue);
          assertThat(context.batchSize()).isEqualTo(myBatchSize);
          throw new SuccessException();
    }
}

public static class SuccessException extends RutimeException {}
```

JingsongLi:
I accidentally merged this PR.
@zhangjun0x01 You can add case in another PR, It is better to add case in `FileFormatTest`, test `CoreOptions.createFileFormat`.

zhangjun0x01:
> I accidentally merged this PR. @zhangjun0x01 You can add case in another PR, It is better to add case in `FileFormatTest`, test `CoreOptions.createFileFormat`.

ok,I will add later

PR #651, opened 2023-03-20
[docs] Fix stable version link 

leonardBang:


JingsongLi (approved):
+1

docs/layouts/partials/docs/inject/content-before.html
```diff
@@ -23,14 +23,14 @@
 {{ if $.Site.Params.ShowOutDatedWarning }}
 <article class="markdown">
     <blockquote style="border-color:#f66">
-        {{ markdownify "This documentation is for an out-of-date version of Apache Paimon. We recommend you use the latest [stable version](https://ci.apache.org/projects/flink/paimon-docs-stable/)."}}
+        {{ markdownify "This documentation is for an out-of-date version of Apache Paimon. We recommend you use the latest [stable version](https://ci.apache.org/projects/flink/flink-table-store-docs-stable/)."}}
```
klion26, line 26: Seems we change the url back to flink-table-store, Will we change the url here to paimon domain in the future?

leonardBang: I think we will change it back once paimon has its first release

PR #658, opened 2023-03-20
[spark] Extract SparkInputPartitionReader from paimon spark connector

TyrantLucifer:
Extract SparkInputPartitionReader from paimon spark connector

### Purpose

Optimize code structure

### Tests

### API and Format 

### Documentation

### DIFF
--- a/paimon-spark/paimon-spark-2/src/main/java/org/apache/paimon/spark/SparkInputPartition.java
+++ b/paimon-spark/paimon-spark-2/src/main/java/org/apache/paimon/spark/SparkInputPartition.java
@@ -54,31 +54,7 @@ public class SparkInputPartition
         }
         RecordReaderIterator<InternalRow> iterator = new RecordReaderIterator<>(recordReader);
         SparkInternalRow row = new SparkInternalRow(readBuilder.readType());
-        return new InputPartitionReader<org.apache.spark.sql.catalyst.InternalRow>() {
-
-            @Override
-            public boolean next() {
-                if (iterator.hasNext()) {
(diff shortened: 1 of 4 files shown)


JingsongLi:
Hi @TyrantLucifer , thanks for your contribution, can you explain the disadvantages of anonymous internal classes?

TyrantLucifer:
> Hi @TyrantLucifer , thanks for your contribution, can you explain the disadvantages of anonymous internal classes?

Thank you for your reply. There are two benefits to converting an internal anonymous class to a regular entity class:

1. Improved code readability and maintainability: Using an explicit entity class with a well-defined name and a clear purpose makes it easier for other developers to understand the code's purpose and functionality. This can also make it easier to maintain the code over time.

2. Enhanced code reusability: By separating the anonymous class into a separate entity class, the class can be reused in other parts of the codebase, potentially reducing code duplication and improving overall code quality.

Of course, from a performance perspective, there is no difference between the two approaches.

What do you think about it?

JingsongLi:
> 2\. t

Thanks for your reply and explanation. I will merge this. But this is not a bad mark for an anonymous internal class, and we would not recommend not using anonymous internal classes.
I think in most moments it's just a matter of personal preference.

TyrantLucifer:
> > 2. t
> 
> Thanks for your reply and explanation. I will merge this. But this is not a bad mark for an anonymous internal class, and we would not recommend not using anonymous internal classes. I think in most moments it's just a matter of personal preference.

Thank you for your good advice. I will avoid this problem in the future and try to contribute other useful codes.:wink:

PR #662, opened 2023-03-20
[ci] Add code style check step in all workflows

TyrantLucifer:

### Purpose

Add code style check step in all workflows

### Tests

### API and Format 

### Documentation

### DIFF
--- a/.github/workflows/e2e-tests-1.14.yml
+++ b/.github/workflows/e2e-tests-1.14.yml
@@ -3,9 +3,21 @@ name: End to End Tests Flink 1.14
 on: [push, pull_request]
 
 jobs:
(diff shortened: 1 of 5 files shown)


JingsongLi:
Thanks @TyrantLucifer for the contribution.
Why we need to style check? `mvn clean install` already did this.

TyrantLucifer:
> Thanks @TyrantLucifer for the contribution. Why we need to style check? `mvn clean install` already did this.

Thank you for your reply. If the code style of this pr not right, it will stop the build step. This can reduce the time and resource of all CI/CD process and tip the developer check their code style first. If you think this pr is uesless, I will close it.

PR #671, opened 2023-03-21
[FLINK-31433] Make SchemaChange serializable

tsreaper:
### Purpose

To avoid concurrent changes to table schema, CDC sinks for Paimon should send all `SchemaChange` to a special process function. This process function only has 1 parallelism and it is dedicated for schema changes.

To pass `SchemaChange` through network, `SchemaChange` must be serializable.

### Tests

N/A

### API and Format 

N/A

### Documentation

N/A

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/schema/SchemaChange.java
+++ b/paimon-core/src/main/java/org/apache/paimon/schema/SchemaChange.java
@@ -357,7 +357,8 @@ public interface SchemaChange extends Serializable {
     }
 
     /** Represents a requested column move in a struct. */
-    class Move {
+    class Move implements Serializable {
+
         public enum MoveType {
             FIRST,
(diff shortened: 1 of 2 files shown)


paimon-core/src/main/java/org/apache/paimon/schema/SchemaChange.java
```diff
@@ -306,6 +325,9 @@ public int hashCode() {
 
     /** A SchemaChange to update the field position. */
     final class UpdateColumnPosition implements SchemaChange {
+
+        private static final long serialVersionUID = 1L;
+
         private final Move move;
```
JingsongLi, line 331: `Move` is not serializable.
You can add test for all class instances.

JingsongLi (approved):
+1

PR #672, opened 2023-03-21
[refactor] simplify String operations and add JDK version into README

legendtkl:
### Purpose
1. some code refactor
2. add JDK version info into README


JingsongLi (approved):
Looks good to me!

PR #675, opened 2023-03-21
[spark] Upgrade spark version to 3.3.2 of spark-common

yuzelin:
### Purpose

To upgrade spark-common version.

### Tests

Original tests under spark-common.
Main fixing the return of `SHOW CREATE TABLE`
1. Column name doesn't escaped with '`' any more.
2. There is a blank space after `TABLEPROPERTIES`

When fixing tests, I made some changes:
1.  the `SparkReadTestBase` has set the catalog and namespace, so I remove the useless namespace in test SQLs;
2. most `SHOW CREATE TABLE` tests don't care about `TABLEPROPERTIES` so I remove them, and I add `SparkReadITCase#testShowCreateTable` to cover the case (USING, PARTITIONED, COMMENT, TBLPROPERTIES).

### API and Format 

No changes.

### Documentation

No.

### DIFF
--- a/paimon-spark/paimon-spark-common/pom.xml
+++ b/paimon-spark/paimon-spark-common/pom.xml
@@ -34,7 +34,7 @@ under the License.
     <name>Paimon : Spark : Common</name>
 
     <properties>
-        <spark.version>3.2.2</spark.version>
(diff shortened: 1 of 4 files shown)


JingsongLi (approved):
+1

PR #677, opened 2023-03-21
[bug] Bounded Stream should end directly if the full scan snapshot should be end

JingsongLi:
### Purpose

If the table has no input now, the bounded stream will never end.
Because there is no check for full scan snapshot.
We should fix this: Bounded Stream should end directly if the full scan snapshot should be end.

### Tests

- StreamDataTableScanTest
- BoundedWatermarkCheckerTest


yuzelin:
LGTM!

PR #680, opened 2023-03-22
[flink] Limit should not rely on assign all splits in StaticFileStoreSplitEnumerator

JingsongLi:
### Purpose

Limit should not rely on assign all splits in StaticFileStoreSplitEnumerator.

Assign all splits in one time will produce exceed `akka.framesize` to exceptions.

We can add a global limiter in `FileStoreSourceReader`, limiter is created in SourceReader, it can be shared in all split readers.

### Tests

Test Exists.


SteNicholas (approved):
LGTM.

PR #685, opened 2023-03-22
[doc] Download snapshot jars

JingsongLi:
### Purpose

Recommend downloading snapshot jars instead of building from source.


yuzelin:
All download source url (https://flink.apache.org/downloads.html) is linking to FLINK.

yuzelin:
LGTM!

PR #691, opened 2023-03-22
[core] Use enum instead of string for file.format

FangYongs:
### Purpose

Use enum instead of string for file.format and validate the format value in ddl

### Tests

1. Added the test `SchemaManagerTest.testInvalidFormatType` to validate the format value

### API and Format 

### Documentation

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-core/src/main/java/org/apache/paimon/CoreOptions.java
@@ -964,9 +964,9 @@ public class CoreOptions implements Serializable {
 
     /** Specifies the file format type for store. */
     public enum FileFormatType implements DescribedEnum {
-        ORC("orc", "ORC format for file store."),
-        PARQUET("parquet", "Parquet format for file store."),
-        AVRO("avro", "Avro format for file store.");
+        ORC("orc", "ORC file format."),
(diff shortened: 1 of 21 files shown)


paimon-core/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -956,6 +962,31 @@ public InlineElement getDescription() {
         }
     }
 
+    /** Specifies the file format type for store. */
+    public enum FileFormatType implements DescribedEnum {
+        ORC("orc", "ORC format for file store."),
```
JingsongLi, line 967: maybe remove `file store`? now log store is only in flink

paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
```diff
@@ -100,6 +101,9 @@ public static void validateTableSchema(TableSchema schema) {
             }
         }
 
+        // Valid the format type
+        checkNotNull(options.formatType());
```
JingsongLi, line 105: option with default value should never return null.
we don't need this checking.

FangYongs:
Thanks @JingsongLi  Updated

JingsongLi (approved):
+1

Issue #632, opened 2023-03-19
[Document] Provide jar download url in documentation


JingsongLi:
Now we have already published to snapshot jar to nexus repository.
We can just provide download in documentation, user don't need to build by themselve.

zhuangchong:
What is the address of the nexus repository you speak of?
I found no snapshot of `paimon` in the apache repository `https://www.apache.org/dyn/closer.lua/incubator/paimon`.

ai-smalleryu:
Let me do it

JingsongLi:
> What is the address of the nexus repository you speak of? I found no snapshot of `paimon` in the apache repository `https://www.apache.org/dyn/closer.lua/incubator/paimon`.

You can find them in here: https://repository.apache.org/#nexus-search;quick~paimon

JingsongLi:
> Let me do it

Welcome, you can create PR for this.

zhuangchong:
@ai-smalleryu Hi, is this PR in progress?

ai-smalleryu:
> @ai-smalleryu Hi, is this PR in progress?

I'm working on it, it's almost ready

PR #697, opened 2023-03-23
[doc] Provide jar download url in documentation

ai-smalleryu:
[doc] Provide jar download url in documentation #632 

> Now we have already published to snapshot jar to nexus repository.
> We can just provide download in documentation, user don't need to build by themselve.


JingsongLi (commented):
We can just introduce a download page in `/project/download.html`, including engines and filesystems.

And link it in `engines/overview` and `filesystems/overview`.

ai-smalleryu:
> And don't use `table store` now, use paimon.

PR #698, opened 2023-03-23
[core][bug] TableScan#plan should not return null

### LINKED ISSUE #638: [bug] TableScan.plan should never return null
We have already introduced `StreamTableScan`, the plan can not contain snapshot id, so empty splits can be returned.
The plan can never be null.

yuzelin:
### Purpose

Fix #638 

1. `#plan()` now return an empty list when scanner give an null result.
2. Remove snapshot id from `Plan`. In `StaticFileStoreSource`, it is always `null`; in `ContinuousFileSplitEnumerator`, it is from `#checkpoint`; In `Hybrid` case, a new Flink source using `StreamDataTableScan` is introduced to solve this.
3. Some codes depend on Plan's nullability, now fixed.

### Tests

Original tests.

### API and Format 

No.

### Documentation

No.


yuzelin:
D1A9FB16-CFC3-4594-8E0D-E030D11A54CF

Verified in local.

yuzelin:
Rebased master.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/StaticFileStoreSource.java
```diff
@@ -75,19 +75,16 @@ public SplitEnumerator<FileStoreSourceSplit, PendingSplitsCheckpoint> restoreEnu
...
             FileStoreSourceSplitGenerator splitGenerator = new FileStoreSourceSplitGenerator();
-
             // read all splits from scan
             DataTableScan.DataFilePlan plan =
                     scanFactory.create(table).withFilter(predicate).plan();
-            if (plan != null) {
-                snapshotId = plan.snapshotId;
-                splits.addAll(splitGenerator.createSplits(plan));
-            }
+            snapshotId =
```
JingsongLi, line 85: snapshotId only needs by `LogHybridSourceFactory`, we can just introduce a `FlinkSource` only for `LogHybridSourceFactory`:
```
public static FlinkSource buildHybridSource(
            Table table,
            @Nullable int[][] projectedFields,
            @Nullable Predicate predicate) {
        if (!(table instanceof DataTable)) {
            throw new UnsupportedOperationException("Only support DataTable, unsupported table: " + table.getClass());
        }

        DataTable dataTable = (DataTable) table;

        return new FlinkHybridSource(
                table.newReadBuilder().withProjection(projectedFields).withFilter(predicate),
                dataTable.snapshotManager(),
                dataTable.coreOptions().toConfiguration());
    }

    private static class FlinkHybridSource extends FlinkSource {

        private static final long serialVersionUID = 3L;

        private final SnapshotManager snapshotManager;
        private final Options options;

        public FlinkHybridSource(ReadBuilder readBuilder, SnapshotManager snapshotManager, Options options) {
            super(readBuilder, null);
            this.snapshotManager = snapshotManager;
            this.options = options;
        }

        @Override
        public Boundedness getBoundedness() {
            return Boundedness.BOUNDED;
        }

        @Override
        public SplitEnumerator<FileStoreSourceSplit, PendingSplitsCheckpoint> restoreEnumerator(
                SplitEnumeratorContext<FileStoreSourceSplit> context,
                PendingSplitsCheckpoint checkpoint) {
            Long snapshotId = null;
            Collection<FileStoreSourceSplit> splits;
            if (checkpoint == null) {
                splits = new ArrayList<>();
                FileStoreSourceSplitGenerator splitGenerator = new FileStoreSourceSplitGenerator();

                // read all splits from scan
                StreamTableScan scan = readBuilder.newStreamScan();
                TableScan.Plan plan = scan.plan();
                if (plan != null) {
                    snapshotId = scan.checkpoint();
                    splits.addAll(splitGenerator.createSplits(plan));
                }
            } else {
                // restore from checkpoint
                snapshotId = checkpoint.currentSnapshotId();
                splits = checkpoint.splits();
            }

            Snapshot snapshot = snapshotId == null ? null : snapshotManager.snapshot(snapshotId);
            return new StaticFileStoreSplitEnumerator(
                    context,
                    snapshot,
                    splits,
                    options.get(FlinkConnectorOptions.SCAN_SPLIT_ENUMERATOR_BATCH_SIZE));
        }
    }
```

yuzelin:
plan
Verified.

PR #702, opened 2023-03-23
[core] Return sorted snapshot rows by default

leo65535:
### Purpose

The system snapshots table return unordered snapshot rows by default, it's better to return sorted rows by default.

### Tests

_**Before**_
image

_**Enhance**_
image

### DIFF
--- a/paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/CatalogTableITCase.java
+++ b/paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/CatalogTableITCase.java
@@ -53,8 +53,9 @@ public class CatalogTableITCase extends CatalogITCaseBase {
         sql("INSERT INTO T VALUES (3, 4)");
 
         List<Row> result = sql("SELECT snapshot_id, schema_id, commit_kind FROM T$snapshots");
-        assertThat(result)
-                .containsExactlyInAnyOrder(Row.of(1L, 0L, "APPEND"), Row.of(2L, 0L, "APPEND"));
+
+        // check correctness and sequence snapshots.
+        assertThat(result).containsExactly(Row.of(1L, 0L, "APPEND"), Row.of(2L, 0L, "APPEND"));
     }
 
     @Test
--- a/paimon-core/src/main/java/org/apache/paimon/utils/SnapshotManager.java
+++ b/paimon-core/src/main/java/org/apache/paimon/utils/SnapshotManager.java
@@ -26,6 +26,7 @@ import javax.annotation.Nullable;
 
 import java.io.IOException;
 import java.io.Serializable;
+import java.util.Comparator;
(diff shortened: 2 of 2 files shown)


paimon-core/src/main/java/org/apache/paimon/table/system/SnapshotsTable.java
```diff
@@ -194,11 +195,18 @@ public RecordReader<InternalRow> createReader(Split split) throws IOException {
...
                 rows =
                         Iterators.transform(
                                 rows, row -> ProjectedRow.from(projection).replaceRow(row));
             }
+
+            rows =
+                    Arrays.stream(Iterators.toArray(rows, InternalRow.class))
+                            .sorted(Comparator.comparingLong(row -> row.getLong(0)))
+                            .iterator();
+
```
zhuangchong, line 209: Is it better to proactively add an order statement to the query sql?

leo65535: hi, it will be more complex if we do that, so let's keep it.

JingsongLi (commented):
Thanks @leo65535 for the contribution.

You can adjust test: `CatalogTableITCase.testSnapshotsTable`. Using `containsExactly` to validate this.

leo65535:
> `containsExactly`

Right, we can use `containsSequence`.

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/CatalogTableITCase.java
```diff
@@ -53,8 +53,9 @@ public void testSnapshotsTable() throws Exception {
         sql("INSERT INTO T VALUES (3, 4)");
 
         List<Row> result = sql("SELECT snapshot_id, schema_id, commit_kind FROM T$snapshots");
-        assertThat(result)
-                .containsExactlyInAnyOrder(Row.of(1L, 0L, "APPEND"), Row.of(2L, 0L, "APPEND"));
+
+        // check ordered snapshots.
+        assertThat(result).containsSequence(Row.of(1L, 0L, "APPEND"), Row.of(2L, 0L, "APPEND"));
```
JingsongLi, line 58: Should be `containsExactly`?
`containsSequence` verifies that the actual group contains the given sequence in the correct order:
```
   * Iterable&lt;Ring&gt; elvesRings = newArrayList(vilya, nenya, narya);
   *
   * // assertions will pass
   * assertThat(elvesRings).containsSequence(vilya, nenya)
   *                       .containsSequence(nenya, narya);
```

leo65535: Got, `containsExactly` check the correctness and order, thanks.

JingsongLi (approved):
+1

PR #704, opened 2023-03-24
[core] Improve the exception information when the changelog producer is `lookup` , the table has no primary key

zhuangchong:

### Purpose

Q1:
when the changelog producer is `lookup` , the table has no primary key, The exception information is: `Changelog table with full-compaction must have primary keys`

Improve:

`Changelog table with lookup must have primary keys`

Q2:
When the system field ["_VALUE_COUNT", "_SEQUENCE_NUMBER", "_VALUE_KIND"] is included in the schema, the exception information is: `Field name["_VALUE_COUNT"] in schema cannot be exist in [["_VALUE_COUNT", "_SEQUENCE_NUMBER", "_VALUE_KIND"]]`

Improve:
`Field name["_VALUE_COUNT"] in schema cannot be exist in ["_VALUE_COUNT", "_SEQUENCE_NUMBER", "_VALUE_KIND"]`

### Tests

### API and Format 

### Documentation

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
+++ b/paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
@@ -93,7 +93,9 @@ public class SchemaValidation {
                 case LOOKUP:
                     if (schema.primaryKeys().isEmpty()) {
                         throw new UnsupportedOperationException(
-                                "Changelog table with full compaction must have primary keys");
+                                "Changelog table with "
+                                        + options.changelogProducer()
+                                        + " must have primary keys");
                     }
                     break;
                 default:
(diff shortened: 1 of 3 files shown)


paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
```diff
@@ -107,7 +109,7 @@ public static void validateTableSchema(TableSchema schema) {
                             checkState(
                                     !SYSTEM_FIELD_NAMES.contains(f),
                                     String.format(
-                                            "Field name[%s] in schema cannot be exist in [%s]",
+                                            "Field name[%s] in schema cannot be exist in %s",
```
SteNicholas, line 112: Why need to change the info of `SYSTEM_FIELD_NAMES`? IMO, it's no need to change.

zhuangchong: I just modified the printed exception information, 
before modification: `Field name["_VALUE_COUNT"] in schema cannot be exist in [["_VALUE_COUNT", "_SEQUENCE_NUMBER", "_VALUE_KIND"]]` ,
 after modification: `Field name["_VALUE_COUNT"] in schema cannot be exist in ["_VALUE_COUNT", "_SEQUENCE_NUMBER", "_VALUE_KIND"]`

SteNicholas (approved):
LGTM. Thanks for @zhuangchong contribution.

PR #711, opened 2023-03-24
[core] Public API and add partitionKeys,primaryKeys,options to table

JingsongLi:
### Purpose

- Public API 
- add partitionKeys, primaryKeys, options to Table API

### Tests

No

### API and Format 

Yes

### Documentation

Will do later


paimon-core/src/main/java/org/apache/paimon/table/Table.java
```diff
@@ -41,6 +43,18 @@ public interface Table extends Serializable {
...
+
+    /** Primary keys of this table. */
+    List<String> primaryKeys();
+
+    /** Options of this table. */
+    Map<String, String> options();
+
+    /** Optional comment of this table. */
+    Optional<String> comment();
+
```
yuzelin, line 57: Is it necessary to public `fileIO()` ? I think it's useful for creating `SnapshotManager` and `SchemaManager`.

yuzelin (approved):

paimon-core/src/main/java/org/apache/paimon/table/system/AuditLogTable.java
```diff
@@ -105,6 +106,21 @@ public RowType rowType() {
...
+    }
+
+    @Override
+    public Map<String, String> options() {
+        return dataTable.options();
+    }
+
+    @Override
+    public List<String> primaryKeys() {
+        return Collections.emptyList();
```
SteNicholas, line 121:
```suggestion
        return Collections.singleList(dataTable.options().get("primary-key"));
```

SteNicholas (approved):
LGTM.

Issue #639, opened 2023-03-19
[improvement] Use new ReadBuilder and WriteBuilder API in tests


JingsongLi:
Stop using old interfaces.

TyrantLucifer:
I want to try it. Could you please assign it to me?

JingsongLi:
> I want to try it. Could you please assign it to me?

Welcome, you can just create PR for this. I think this may need multiple PRs.

SteNicholas:
@JingsongLi, IMO, it's no need to create multiple PRs to solve this issue. @TyrantLucifer, you could create multiple commits for different modules in one PR. Please keep one commit for initial PR.

JingsongLi:
Hi @TyrantLucifer , we don't just use the new ReadBuilder and WriteBuilder APIs either, we need to get rid of the FileStoreTable as well, our best practice is to use Catalog to create the Table and then follow through with the operations.

JingsongLi:
> @JingsongLi, IMO, it's no need to create multiple PRs to solve this issue. @TyrantLucifer, you could create multiple commits for different modules in one PR. Please keep one commit for initial PR.

One commit for initial PR? But create multiple commits for different modules in one PR? Can you explain more?

PR #712, opened 2023-03-24
[test] Use new ReadBuilder and WriteBuilder API in tests of paimon-core module  

TyrantLucifer:
### Purpose

Use new ReadBuilder and WriteBuilder API in tests of paimon-core module  

sub-task of #639


TyrantLucifer:
Hi @JingsongLi this is a sub task of #639 , PTAL

SteNicholas:
@TyrantLucifer, you could use new ReadBuilder and WriteBuilder API in tests of all modules in one PR and multiple commits. It's no need to create multiple PRs and multiple commits for one module. It's hard to review the initial PR with multiple commits.

JingsongLi (commented):
Not only for reader writer, we should get rid of `FileStoreTable` too.

PR #714, opened 2023-03-24
[core] support JDK 11

legendtkl:
### Purpose
Support JDK11
issue ref: https://github.com/apache/incubator-paimon/issues/693

### Tests

regression it/e2e case

### API and Format 

no

### Documentation

ye

### Note
all the java code change is because different behavior between jdk8 and jdk11:
1. `LocalDateTime.now()` return micro seconds in jdk11 (only return millisecond in jdk8), and the DataFileMeta serialize and deserialize both specify the TimeStamp(3). In jdk11, all the related cases will be broken.
2. Changes in `CachedRandomInputView.java` is to fix concurrency issues.

### DIFF
--- a/.github/workflows/e2e-tests-1.14.yml
+++ b/.github/workflows/e2e-tests-1.14.yml
@@ -25,6 +25,9 @@ on:
       - 'docs/**'
       - '**/*.md'
 
+env:
+  JDK_VERSION: 8
+
 jobs:
   build:
     runs-on: ubuntu-latest
@@ -32,10 +35,11 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
-      - name: Set up JDK 1.8
-        uses: actions/setup-java@v1
+      - name: Set up JDK ${{ env.JDK_VERSION }}
+        uses: actions/setup-java@v2
         with:
-          java-version: 1.8
+          java-version: ${{ env.JDK_VERSION }}
+          distribution: 'adopt'
       - name: Build Flink 1.14
         run: ./mvnw clean install -Dmaven.test.skip=true -Pflink-1.14
       - name: Test Flink 1.14
--- a/paimon-common/src/main/java/org/apache/paimon/io/cache/CachedRandomInputView.java
+++ b/paimon-common/src/main/java/org/apache/paimon/io/cache/CachedRandomInputView.java
@@ -73,9 +73,12 @@ public class CachedRandomInputView extends AbstractPagedInputView
     }
 
     private MemorySegment getCurrentPage() {
-        return segments.computeIfAbsent(
-                currentSegmentIndex,
-                key -> cacheManager.getPage(file, currentSegmentIndex, this::invalidPage));
+        MemorySegment segment = segments.get(currentSegmentIndex);
+        if (segment == null) {
+            segment = cacheManager.getPage(file, currentSegmentIndex, this::invalidPage);
+            segments.put(currentSegmentIndex, segment);
+        }
+        return segment;
     }
 
     @Override
--- a/paimon-spark/paimon-spark-2/pom.xml
+++ b/paimon-spark/paimon-spark-2/pom.xml
@@ -34,6 +34,7 @@ under the License.
     <properties>
         <spark2.version>2.4.8</spark2.version>
         <jackson.version>2.13.3</jackson.version>
+        <asm6.version>4.10</asm6.version>
     </properties>
 
     <dependencies>
@@ -75,6 +76,10 @@ under the License.
                     <groupId>org.slf4j</groupId>
                     <artifactId>slf4j-log4j12</artifactId>
                 </exclusion>
+                <exclusion>
+                    <groupId>org.apache.xbean</groupId>
+                    <artifactId>xbean-asm6-shaded</artifactId>
+                </exclusion>
             </exclusions>
         </dependency>
 
@@ -83,6 +88,11 @@ under the License.
             <artifactId>jackson-module-scala_2.11</artifactId>
             <version>${jackson.version}</version>
             <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.xbean</groupId>
+            <artifactId>xbean-asm6-shaded</artifactId>
+            <version>${asm6.version}</version>
         </dependency>
     </dependencies>
 
--- a/.github/workflows/e2e-tests-1.14-jdk11.yml
+++ b/.github/workflows/e2e-tests-1.14-jdk11.yml
@@ -16,20 +16,22 @@
 # limitations under the License.
 ################################################################################
 
-name: End to End Tests Flink 1.15
+name: End to End Tests Flink 1.14 on JDK 11
 
 on:
-  push:
-  pull_request:
-    paths-ignore:
-      - 'docs/**'
-      - '**/*.md'
+  issue_comment:
+    types: [created, edited, deleted]
+
+  # daily run
+  schedule:
+    - cron: "0 0 * * *"
 
 env:
-  JDK_VERSION: 8
+  JDK_VERSION: 11
 
 jobs:
   build:
+    if: contains(github.event.comment.html_url, '/pull/') && contains(github.event.comment.body, '/jdk11')
     runs-on: ubuntu-latest
 
     steps:
@@ -40,13 +42,12 @@ jobs:
         with:
           java-version: ${{ env.JDK_VERSION }}
           distribution: 'adopt'
-      - name: Build Flink 1.15
-        run: ./mvnw clean install -Dmaven.test.skip=true -Pflink-1.15
-      - name: Test Flink 1.15
-        timeout-minutes: 60
+      - name: Build Flink 1.14
+        run: ./mvnw clean install -Dmaven.test.skip=true -Pflink-1.14
+      - name: Test Flink 1.14
         run: |
           # run tests with random timezone to find out timezone related bugs
           . .github/workflows/utils.sh
           jvm_timezone=$(random_timezone)
           echo "JVM timezone is set to $jvm_timezone"
-          ./mvnw test -pl paimon-e2e-tests -Duser.timezone=$jvm_timezone -Pflink-1.15
+          ./mvnw test -pl paimon-e2e-tests -Duser.timezone=$jvm_timezone -Pflink-1.14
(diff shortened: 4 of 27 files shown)


.github/workflows/e2e-tests-1.14.yml
```diff
@@ -32,10 +32,11 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
-      - name: Set up JDK 1.8
-        uses: actions/setup-java@v1
+      - name: Set up JDK 11
+        uses: actions/setup-java@v2
         with:
-          java-version: 1.8
+          java-version: '11'
```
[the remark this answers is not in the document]
legendtkl: @TyrantLucifer  Yes, CI for both JDK8 and JDK11 will take a long time. 

I think we can adopt JDK11 only lately, and no need to support JDK8 in the future.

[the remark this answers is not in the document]
TyrantLucifer: Specifying multiple Java versions can result in two tasks running simultaneously, utilizing parallel processing. Moreover, in 2023, GitHub has increased action resources for Apache projects, so theoretically, this should not significantly increase the time required for continuous integration. Additionally, many users are still using Java 8. Therefore, completely abandoning Java 8 may be too drastic of a strategy. :joy:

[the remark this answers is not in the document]
legendtkl: @TyrantLucifer  OK, let's keep them both temporarily.

But from a long-term view, supporting JDK11 without deprecating JDK8 is not a real upgrading. We have to do much meaningless work for the compatibility, and give up the JDK11 specified programing feature.

I will open a new ISSUE to discuss the JDK8 Deprecation after this PR is merged.

s7monk:
I also think it is too radical to completely abandon JDK8, because many companies and individual developers are still using JDK8

.github/workflows/e2e-tests-1.14.yml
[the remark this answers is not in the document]
s7monk: I also think it is too radical to completely abandon JDK8, because many companies and individual developers are still using JDK8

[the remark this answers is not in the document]
legendtkl: @s7monk  ok, let's keep both now.

SteNicholas:
@legendtkl, `sun.misc.Unsafe` is deprecated since JDK11. You need to take a look at this class.

legendtkl:
hi, @SteNicholas 
1. CI failure is because of the unstable UT described in this issue: https://github.com/apache/incubator-paimon/issues/657. I am not sure the failure frequency of occurrence, would you help re-run the CI? thanks.
2. sun.misc.Unsafe is still available in JDK11, but removed to jdk.unsupported. We can still compile this with JDK11. There is also an issue to track this: https://github.com/apache/incubator-paimon/issues/692 . To enable the compiler error in IDEA, plz make the option un-tick. 
image

Thanks.

paimon-common/src/main/java/org/apache/paimon/io/cache/CachedRandomInputView.java
```diff
@@ -55,7 +55,7 @@ public CachedRandomInputView(File file, CacheManager cacheManager)
         this.file = new RandomAccessFile(file, "r");
         this.fileLength = file.length();
         this.cacheManager = cacheManager;
-        this.segments = new HashMap<>();
+        this.segments = new ConcurrentHashMap<>();
```
FangYongs, line 58: What's this change for? Maybe you should create an another issue and add test case for this change

legendtkl: @FangYongs  This is to fix ConcurrentModificationException threw by JDK11. 

Generally speaking, it's part of JDK11 supported work.

> What's this change for? Maybe you should create an another issue and add test case for this change

SteNicholas: @legendtkl, adds the test case to test this fix for `CachedRandomInputView` in JDK11?

legendtkl: @SteNicholas  ok

JingsongLi:
Thanks @legendtkl for the contribution!
I am very pleased to be able to officially announce support for JDK11!
But one thing is that still JDK8 is currently the main force, as mentioned by the partners of the review above.

Another thing is that I don't really want to cause too many tests, is it possible to provide the ability to test with JDK11, but by default it needs to be enabled manually?

legendtkl:
> Thanks @legendtkl for the contribution! I am very pleased to be able to officially announce support for JDK11! But one thing is that still JDK8 is currently the main force, as mentioned by the partners of the review above.
> 
> Another thing is that I don't really want to cause too many tests, is it possible to provide the ability to test with JDK11, but by default it needs to be enabled manually?

@JingsongLi  OK, let me have a check.

By the way, would you help review this PR either too?  I have changed some code written by you. Thanks.

legendtkl:
> Thanks @legendtkl for the contribution! I am very pleased to be able to officially announce support for JDK11! But one thing is that still JDK8 is currently the main force, as mentioned by the partners of the review above.
> 
> Another thing is that I don't really want to cause too many tests, is it possible to provide the ability to test with JDK11, but by default it needs to be enabled manually?

Hi, @JingsongLi , I have add the github action based PR's comment and update the workflows.

The JDK 11 workflows will not run by default. When we comment '/jdk11' in the PR, the workflow will start to run. ***But this feature will work only when the workflows files is on the default branch, namely master.*** That means, we can't verify this in this PR, and it will work only when this PR is merged.

I have tested it manually on my branch. When I comment '/jdk11', the jdk11 workflows start running, we can see it in the `Action` tab.

image


image

Ref: https://github.com/orgs/community/discussions/25389

paimon-common/src/main/java/org/apache/paimon/io/cache/CachedRandomInputView.java
```diff
@@ -50,12 +52,24 @@ public class CachedRandomInputView extends AbstractPagedInputView
...
+     * The thread pool is involved to solve computeIfAbsent issue. From JDK 9, computeIfAbsent will
+     * throw ConcurrentModificationException if it is detected that the mapping function modifies
+     * this map during computation. So we add the thread pool to remove segments expired page. To
+     * guarantee the concurrency safety, we need to replace the HashMap with ConcurrentHashMap.
+     *
+     * <p>Note: if the cache expires many pages in a while, the single thread executor can't process
+     * it. That means, some remove actions are blocked, this will cause dirty data read. But this
+     * occurs with low probability.
+     */
+    ExecutorService executor = Executors.newSingleThreadExecutor();
```
JingsongLi, line 65: We don't need to introduce this complex codes.
How about just modify `getCurrentPage`.
```
    private MemorySegment getCurrentPage() {
        MemorySegment segment = segments.get(currentSegmentIndex);
        if (segment == null) {
            segment = cacheManager.getPage(file, currentSegmentIndex, this::invalidPage);
            segments.put(currentSegmentIndex, segment);
        }
        return segment;
    }
```

paimon-spark/paimon-spark-2/pom.xml
```diff
@@ -84,6 +89,11 @@ under the License.
             <version>${jackson.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.xbean</groupId>
+            <artifactId>xbean-asm6-shaded</artifactId>
+            <version>${asm6.version}</version>
```
JingsongLi, line 95: which scope? test or provided?

.github/workflows/e2e-tests-1.15.yml
```diff
@@ -32,10 +32,11 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
-      - name: Set up JDK 1.8
-        uses: actions/setup-java@v1
+      - name: Set up JDK 8
```
JingsongLi, line 35: Can we just have a env here? if (is jdk 11) Set up JDK 11 else Set up JDK 8?

JingsongLi:
@legendtkl Thanks, left comments.

paimon-spark/paimon-spark-2/pom.xml
legendtkl: @JingsongLi  This package is to replace the package in `spark-sql_2.11`, so keep the scope same with `spark-sql_2.11`, namely default `compile`

paimon-common/src/main/java/org/apache/paimon/io/cache/CachedRandomInputView.java
legendtkl: ok. 

Actually, this is the first edition of this PR.

legendtkl:
Hi, @JingsongLi , I have addressed the review comments, would you help review again? thanksl

SteNicholas:
@legendtkl, I left one minor comment above. Please resolve the conflicts and address above comment. PTAL.

paimon-common/src/main/java/org/apache/paimon/io/cache/CachedRandomInputView.java
legendtkl: @SteNicholas  yes, once we replace computeIfAbsent log, it's ok to use HashMap. 

Let's keep the hashmap left.

legendtkl:
> @legendtkl, I left one minor comment above. Please resolve the conflicts and address above comment. PTAL.

hi, @SteNicholas  , I have updated the code, would you help review it again? cc @JingsongLi 

Thanks.

SteNicholas:
@legendtkl,  as the description mentioned, is the changes of `CachedRandomInputView` related to the support of JDK 11?

legendtkl:
> @legendtkl, as the description mentioned, is the changes of `CachedRandomInputView` related to the support of JDK 11? BTW, where is the change for `LocalDateTime.now()`?

@SteNicholas 

Two questions' answer:
1. yes. JDK11 will do more check in computeIfAbsent, now we replace it
2. you can refer paimon-core/src/main/java/org/apache/paimon/io/DataFileMeta.java  . The main reason is that serialization and deserialization have limit the precesion to TimeStamp(3). (Serialization specify this by schema)

SteNicholas (approved):
LGTM.

PR #719, opened 2023-03-26
[flink] Use new ReadBuilder and WriteBuilder API in tests of paimon-flink module

TyrantLucifer:
Use new ReadBuilder and WriteBuilder API in tests of paimon-flink module

sub-task of #639

### DIFF
--- a/paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/sink/CommitterOperatorTest.java
+++ b/paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/sink/CommitterOperatorTest.java
@@ -25,6 +25,7 @@ import org.apache.paimon.manifest.ManifestCommittableSerializer;
 import org.apache.paimon.table.FileStoreTable;
 import org.apache.paimon.table.sink.CommitMessage;
 import org.apache.paimon.table.sink.StreamTableWrite;
+import org.apache.paimon.table.sink.StreamWriteBuilder;
 import org.apache.paimon.utils.SnapshotManager;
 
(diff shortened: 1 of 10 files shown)


paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/sink/CommitterOperatorTest.java
```diff
@@ -246,7 +256,7 @@ private OneInputStreamOperatorTestHarness<Committable, Committable> createLossyT
                 new CommitterOperator(
                         true,
                         initialCommitUser,
-                        user -> new StoreCommitter(table.newCommit(user)),
+                        user -> new StoreCommitter(table.newStreamWriteBuilder().newCommit()),
```
tsreaper, line 259: You forgot `withCommitUser`. I'm thinking if we need to force users to specify a `commitUser` when using `StreamWriteBuilder`, instead of generating a random UUID for them. @JingsongLi

tsreaper (commented):
Please also delete the `commitUser` class member from the test classes.

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/sink/CommitterOperatorTest.java
TyrantLucifer: Thank you for you reply, I will fix this problem later.

JingsongLi (approved):
+1

JingsongLi:
It seems that test failed?

TyrantLucifer:
> It seems that test failed?

Could you please help me restart the failed task?

PR #718, opened 2023-03-26
[hive] Use new ReadBuilder and WriteBuilder API in tests of paimon-hive module

TyrantLucifer:
Use new ReadBuilder and WriteBuilder API in tests of paimon-hive module

sub-task of #639

PR #720, opened 2023-03-26
[ci] add timeout option for all test steps

TyrantLucifer:
### Purpose

Due to the instability of GitHub action resources, some tasks may remain in a dormant state indefinitely, resulting in false positives.

For example of #718 as the following shown:

image

It may be helpful to add a timeout delay parameter to mitigate this issue


SteNicholas:
@TyrantLucifer, I don't think we need to add timeout option for all test steps.

TyrantLucifer:
> @TyrantLucifer, I don't think we need to add timeout option for all test steps.

Can you provide some suggestions on which modules may benefit from this feature?

JingsongLi (approved):
Looks good to me, the current test is not too stable, indeed there will be hang situation

JingsongLi:
Thanks @TyrantLucifer , merged

PR #724, opened 2023-03-28
[hotfix] Minor fix to TableScan.plan non-null

JingsongLi:
Fix codes after https://github.com/apache/incubator-paimon/pull/698


yuzelin:
LGTM

PR #726, opened 2023-03-28
[flink][bug] Fix incorrect partitioning result from CdcBucketStreamPartitioner#selectChannel

tsreaper:
### Purpose

Current logic in `CdcBucketStreamPartitioner#selectChannel` is incorrect. This may lead to records from the same bucket being sent to different parallelism in downstream, and will cause the same bucket to be modified by multiple writers.

### Tests

FlinkCdcSinkITCase

### API and Format 

N/A

### Documentation

N/A


JingsongLi (commented):
+1

PR #755, opened 2023-03-29
[core] AppendOnly table full reading order is wrong

JingsongLi:
### Purpose

When latest-full streaming reading.
The records order of full reading is wrong. This is because:

New files may be created during the compaction process, then the results of the compaction may be put after the new files, and this order will be disrupted.

We need to ensure this order, so we force the order of files by sequence.

### Tests

`AppendOnlyFileStoreTableTest.testBatchOrderWithCompaction`

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/append/AppendOnlyCompactManager.java
+++ b/paimon-core/src/main/java/org/apache/paimon/append/AppendOnlyCompactManager.java
@@ -31,12 +31,14 @@ import org.apache.paimon.utils.Preconditions;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Comparator;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import java.util.TreeSet;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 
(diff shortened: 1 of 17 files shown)


paimon-core/src/main/java/org/apache/paimon/append/AppendOnlyCompactManager.java
```diff
@@ -301,4 +302,39 @@ public List<DataFileMeta> after() {
...
+
+    private static boolean isOverlap(DataFileMeta o1, DataFileMeta o2) {
+        if (o1.minSequenceNumber() <= o2.maxSequenceNumber()
+                && o1.maxSequenceNumber() >= o2.minSequenceNumber()) {
+            return true;
+        }
+
+        return o2.minSequenceNumber() <= o1.maxSequenceNumber()
+                && o2.maxSequenceNumber() >= o1.minSequenceNumber();
+    }
```
yuzelin, line 339: Maybe it's more clear to check false first:
```
if (o1.max() < o2.min || o1.min > o2.max) {
    return false;
}
return true;
```

paimon-core/src/main/java/org/apache/paimon/append/AppendOnlyCompactManager.java
```diff
@@ -301,4 +302,39 @@ public List<DataFileMeta> after() {
...
+                    return Long.compare(o1.minSequenceNumber(), o2.minSequenceNumber());
+                });
+        return files;
+    }
+
+    private static boolean isOverlap(DataFileMeta o1, DataFileMeta o2) {
+        if (o1.minSequenceNumber() <= o2.maxSequenceNumber()
+                && o1.maxSequenceNumber() >= o2.minSequenceNumber()) {
+            return true;
+        }
```
yuzelin, line 335: Duplicated.

yuzelin (approved):
LGTM

Issue #731, opened 2023-03-28
[Bug] When scan.mode is set to TimeTravel semantics, getTable from Catalog will fail


liming30:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/incubator-paimon/issues) and found nothing similar.

### Paimon version

0.4

### Compute Engine

flink-1.16

### Minimal reproduce step

Execute the following sql in sql client:
```
CREATE CATALOG ts_catalog WITH (
    'type' = 'paimon',
    'warehouse' = '/test-data/e1d13ddf-7423-4075-91b2-ca5f7a18b9fd.store'
);
USE CATALOG ts_catalog;
CREATE TABLE IF NOT EXISTS ts_table (
    k VARCHAR,
    v INT,
    PRIMARY KEY (k) NOT ENFORCED
) WITH (
    'bucket' = '2',
    'log.consistency' = 'eventual',
    'log.system' = 'kafka',
    'kafka.bootstrap.servers' = 'kafka:9092',
    'scan.mode' = 'from-snapshot',
    'scan.snapshot-id' = '1',
    'kafka.topic' = 'ts-topic-b6a8e1c8-003e-4ef9-a5cc-16979a0ce56b'
);

INSERT INTO result1 SELECT * FROM ts_table;
```

### What doesn't meet your expectations?

Creating a Table with timeTravel semantics should not report an error. But I get the following exception:
```
Could not execute SQL statement.
org.apache.flink.table.client.gateway.SqlExecutionException: Failed to parse statement: SELECT * FROM ts_table
;
	at org.apache.flink.table.client.gateway.local.LocalExecutor.parseStatement(LocalExecutor.java:174) ~[flink-sql-client-1.16.1.jar:1.16.1]
	at org.apache.flink.table.client.cli.SqlCommandParserImpl.parseCommand(SqlCommandParserImpl.java:45) ~[flink-sql-client-1.16.1.jar:1.16.1]
	at org.apache.flink.table.client.cli.SqlMultiLineParser.parse(SqlMultiLineParser.java:71) ~[flink-sql-client-1.16.1.jar:1.16.1]
	at org.jline.reader.impl.LineReaderImpl.acceptLine(LineReaderImpl.java:2964) ~[flink-sql-client-1.16.1.jar:1.16.1]
	at org.jline.reader.impl.LineReaderImpl$1.apply(LineReaderImpl.java:3778) ~[flink-sql-client-1.16.1.jar:1.16.1]
[...]

PR #761, opened 2023-03-29
[core][bug] fix the issue of failing to get table with timeTravel semantics from catalog (#731)

liming30:
### Purpose

[core][bug] fix the issue of failing to get table with timeTravel semantics from catalog (#731)

### Tests

The existing `FlinkCatalogTest` can already cover this change. I added the `scan.mode` configuration to `FlinkCatalogTest`.

### API and Format 

No changes.

### Documentation

No.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/FileStoreTableFactory.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/FileStoreTableFactory.java
@@ -72,6 +72,11 @@ public class FileStoreTableFactory {
     public static FileStoreTable create(
             FileIO fileIO, Path tablePath, TableSchema tableSchema, Options dynamicOptions) {
         FileStoreTable table;
+        tableSchema =
+                tableSchema.copy(
+                        Options.fromMap(tableSchema.options())
+                                .set(PATH, tablePath.toString())
+                                .toMap());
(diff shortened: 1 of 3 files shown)


paimon-core/src/main/java/org/apache/paimon/table/FileStoreTableFactory.java
```diff
@@ -72,6 +72,11 @@ public static FileStoreTable create(FileIO fileIO, Path tablePath, TableSchema t
     public static FileStoreTable create(
             FileIO fileIO, Path tablePath, TableSchema tableSchema, Options dynamicOptions) {
         FileStoreTable table;
+        tableSchema =
```
JingsongLi, line 75: Can we modify `AbstractFileStoreTable`?
```
    public AbstractFileStoreTable(FileIO fileIO, Path path, TableSchema tableSchema) {
        this.fileIO = fileIO;
        this.path = path;
        if (!tableSchema.options().containsKey(PATH.key())) {
            // make sure table is always available
            Map<String, String> newOptions = new HashMap<>(tableSchema.options());
            newOptions.put(PATH.key(), path.toString());
            tableSchema = tableSchema.copy(newOptions);
        }
        this.tableSchema = tableSchema;
    }
```

JingsongLi (approved):
+1

PR #762, opened 2023-03-29
[Paimon-682][Improvement]Support schema validation when invoke create sch…

hk-lrzy:
…ema.

https://github.com/apache/incubator-paimon/issues/682

### Purpose

### Tests

### API and Format 

### Documentation


FangYongs:
Thanks @hk-lrzy , we need to add tests for the validation

FangYongs:
The e2e test fails du to an exist issue
Thanks @hk-lrzy LGTM +1

Issue #660, opened 2023-03-20
[Bug] Hive catalog should not shade hive dependencies


JingsongLi:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/incubator-paimon/issues) and found nothing similar.

### Paimon version

0.4

### Compute Engine

flink,spark

### Minimal reproduce step

run flink with hive catalog with gateway with hive jdbc.

### What doesn't meet your expectations?

No conflicts.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #764, opened 2023-03-29
[hive] Remove bundled Hive classes from Paimon Hive catalog and bundle Hive catalog to distributed jars of all engines

tsreaper:
### Purpose

This PR solves #660 .

Currently, some classes from Hive 2.3 are bundled into the Paimon Hive catalog. However, this can lead to conflicts when users are using a different version of Hive.

Furthermore, the current Hive catalog only considers Hive 2.x, but the parameters of RetryingMetaStoreClient#getProxy differ between Hive 2.x and Hive 3.x. This can result in a NoSuchMethod exception for users of Hive 3.x.

Once this pull request is merged, the Paimon Hive catalog will be small and independent enough to be bundled into the distributed jars for all supported engines.

### Tests

* Hive23CatalogITCase
* Hive31CatalogITCase

### API and Format 

N/A

### Documentation

This pull request may only affect Flink users. Once it is merged, the Paimon Hive catalog for Flink will rely on the Flink Hive connector bundled jar, as indicated in the documentation.


docs/content/how-to/creating-catalogs.md
```diff
@@ -76,45 +76,11 @@ USE paimon.default;
...
-
 {{< tabs "hive-metastore-example" >}}
 
 {{< tab "Flink" >}}
 
-To enable Paimon Hive catalog support in Flink, you can pick one of the following two methods.
-
-* Copy Paimon Hive catalog jar file into the `lib` directory of your Flink installation directory. Note that this must be done before starting your Flink cluster.
-* If you're using Flink's SQL client, append `--jar /path/to/paimon-hive-catalog-{{< version >}}.jar` to the starting command of SQL client.
+Paimon Hive catalog in Flink relies on Flink Hive connector bundled jar. You should first download Flink Hive connector bundled jar and add it to classpath. See [here](https://nightlies.apache.org/flink/flink-docs-release-1.17/docs/connectors/table/hive/overview/#using-bundled-hive-jar) for more info.
```
SteNicholas, line 83: Is it better to use the master branch link https://nightlies.apache.org/flink/flink-docs-master/docs/connectors/table/hive/overview/#using-bundled-hive-jar?

paimon-e2e-tests/pom.xml
```diff
@@ -34,6 +34,7 @@ under the License.
     <properties>
         <flink.shaded.hadoop.version>2.8.3-10.0</flink.shaded.hadoop.version>
         <flink.sql.connector.kafka>flink-sql-connector-kafka</flink.sql.connector.kafka>
+        <flink.sql.connector.hive>flink-sql-connector-hive-2.3.9_${scala.binary.version}</flink.sql.connector.hive>
```
SteNicholas, line 37: Does the e2e tests need to add the test case for 3.x?

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
```diff
@@ -470,17 +472,53 @@ private Lock lock(Identifier identifier) {
...
-                            hiveConf,
-                            tbl -> null,
-                            new ConcurrentHashMap<>(),
-                            clientClassName,
-                            true);
-        } catch (MetaException e) {
+                    (IMetaStoreClient)
+                            getProxy.invoke(
+                                    null,
+                                    hiveConf,
```
SteNicholas, line 516: Why does the `Configuration` type parameter `conf` set to null?

paimon-hive/paimon-hive-connector-3.1/pom.xml
```diff
@@ -35,15 +35,455 @@ under the License.
...
+        -->
+        <dependency>
+            <groupId>org.apache.avro</groupId>
+            <artifactId>avro</artifactId>
+            <version>${avro.version}</version>
+            <scope>test</scope>
         </dependency>
 
+        <!--
+        hive-exec must stay after flink-connector-hive and avro to avoid conflicts
```
SteNicholas, line 485: Does the 3.1 version of `hive-exec` dependency exclude the calcite dependencies? As tested before, the version of calcite dependencies has conflict with flink-table-planner.

SteNicholas (commented):
@tsreaper, I left the minor comments for this pull request. PTAL.

paimon-e2e-tests/pom.xml
tsreaper: I'd love to. However currently I can't find any Hive 3.x docker image. Let's create an issue for adding Hive 3.x e2e tests.

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
tsreaper: The first parameter of `Method#invoke` is on which object you're calling this method. As this is a static method, this parameter should be `null`. Arguments for the original method starts from the second parameter.

JingsongLi (approved):
+1

PR #766, opened 2023-03-29
[core] BatchWriteBuilder#withOverwrite should not accept multiple partitions list

yuzelin:
### Purpose

`BatchWriteBuilder` is API for table, and currently there is no need to modify multiple partitions from table layer.
This PR removes the `BatchWriteBuilder#withOverwrite(List<Map<String, String>>)` method.

The abilities to drop multiple partitions is taken by `FileStoreCommit#dropPartitions(List<Map<String, String>>)`.

### Tests
`DropPartitionActionITCase`
`PartitionExpireTest`

Added:
`FileStoreCommitTest#testDropPartitions`

### API and Format 

Remove public API `BatchWriteBuilder#withOverwrite(List<Map<String, String>>)`.

### Documentation

No.

### DIFF
--- a/paimon-core/pom.xml
+++ b/paimon-core/pom.xml
@@ -145,6 +145,13 @@ under the License.
                 </exclusion>
             </exclusions>
         </dependency>
+
+        <dependency>
+            <groupId>org.apache.paimon</groupId>
+            <artifactId>paimon-test-utils</artifactId>
+            <version>${project.version}</version>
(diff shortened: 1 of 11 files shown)


yuzelin:
BE473832-FD08-4A08-9021-6D6D253D0F77
Verified.

JingsongLi (approved):
+1

Issue #722, opened 2023-03-27
[Feature] Assign batch splits for ContinuousFileSplitEnumerator


FangYongs:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/incubator-paimon/issues) and found nothing similar.

### Motivation

#687 introduces an option `scan.split-enumerator.batch-size` for assigning splits at one time to avoid exceed `akka.framesize` limit in `StaticFileStoreSplitEnumerator`. `ContinuousFileSplitEnumerator` should support this mechanism too.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

wxplovecc:
Hi @FangYongs  Assign this to me .

yuzelin:
This is solved, please close~

PR #770, opened 2023-03-30
[flink] Assign batch splits for ContinuousFileSplitEnumerator

wxplovecc:
### Purpose
This PR solves #722  .

introduces an option scan.split-enumerator.batch-size for assigning splits at one time to avoid exceed akka.framesize limit in StaticFileStoreSplitEnumerator. ContinuousFileSplitEnumerator should support this mechanism too.

### Tests

`ContinuousFileSplitEnumeratorTest.testSplitWithBatch`

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/ContinuousFileSplitEnumerator.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/ContinuousFileSplitEnumerator.java
@@ -64,6 +64,9 @@ public class ContinuousFileSplitEnumerator
 
     private final StreamTableScan scan;
 
+    /** Default batch splits size to avoid exceed `akka.framesize`. */
+    private final int splitBatchSize;
+
     @Nullable private Long nextSnapshotId;
 
     private boolean finished = false;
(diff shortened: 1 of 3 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/ContinuousFileSplitEnumerator.java
```diff
@@ -205,9 +210,13 @@ private Map<Integer, List<FileStoreSourceSplit>> createAssignment() {
                                 readersAwaitingSplit.remove(task);
                                 return;
                             }
+                            List<FileStoreSourceSplit> taskAssignment = new ArrayList<>();
+                            while (!splits.isEmpty() && taskAssignment.size() < splitBatchSize) {
```
JingsongLi, line 214: I think it is OK to just assign one split to a task for a bucket.
Maybe here we can just limit assignment batch.
```
List<FileStoreSourceSplit> taskAssignment = assignment.computeIfAbsent(task, i -> new ArrayList<>());
if (taskAssignment.size() < splitBatchSize) {
   taskAssignment.add(splits.poll());
}
```

This can be limited to a large number of buckets, and only a few tasks can work properly.
For example, buckets = 1000, but there is only one task.

JingsongLi (approved):
+1

PR #775, opened 2023-03-30
[flink] support tables inherit not built-in configurations from catalog

legendtkl:
### Purpose
issue ref: https://github.com/apache/incubator-paimon/issues/754

### Tests
Add test cases in ITCase.

Now we lack the UT for the catalog, but there's already an issue https://github.com/apache/incubator-paimon/issues/751 to track this.  So I add the test cases in ITCase temporarily.

### API and Format 

no

### Documentation

no

### DIFF
--- a/docs/content/how-to/creating-catalogs.md
+++ b/docs/content/how-to/creating-catalogs.md
@@ -66,12 +66,12 @@ The following shell command registers a paimon catalog named `paimon`. Metadata 
 spark-sql ... \
     --conf spark.sql.catalog.paimon=org.apache.paimon.spark.SparkCatalog \
     --conf spark.sql.catalog.paimon.warehouse=hdfs://path/to/warehouse \
-    --conf spark.sql.catalog.paimon.table_default.key=value
+    --conf spark.sql.catalog.paimon.table-default.key=value
 ```
 
-The conf `--conf spark.sql.catalog.paimon.table_default.key=value` specifies the default property `'key' = 'value'` for tables created in the catalog.
+The conf `--conf spark.sql.catalog.paimon.table-default.key=value` specifies the default property `'key' = 'value'` for tables created in the catalog.
 
-You can define any default properties with the prefix `spark.sql.catalog.paimon.table_default.` configuration. If the table defines the same properties with the default properties, the table's properties will be taken.
+You can define any default properties with the prefix `spark.sql.catalog.paimon.table-default.` configuration. If the table defines the same properties with the default properties, the table's properties will be taken.
 
 After `spark-sql` is started, you can switch to the `default` database of the `paimon` catalog with the following SQL.
 
@@ -126,7 +126,7 @@ spark-sql ... \
     --conf spark.sql.catalog.paimon.uri=thrift://<hive-metastore-host-name>:<port>
 ```
 
-You can also define any default properties with the prefix `spark.sql.catalog.paimon.table_default.` configuration for tables as above `Catalog with Filesystem Metastore`.
+You can also define any default properties with the prefix `spark.sql.catalog.paimon.table-default.` configuration for tables as above `Catalog with Filesystem Metastore`.
 
 After `spark-sql` is started, you can switch to the `default` database of the `paimon` catalog with the following SQL.
 
--- a/docs/content/how-to/creating-catalogs.md
+++ b/docs/content/how-to/creating-catalogs.md
@@ -44,11 +44,17 @@ The following Flink SQL registers and uses a Paimon catalog named `my_catalog`. 
 ```sql
 CREATE CATALOG my_catalog WITH (
     'type' = 'paimon',
-    'warehouse' = 'hdfs://path/to/warehouse'
+    'warehouse' = 'hdfs://path/to/warehouse',
+    'table-default.key' = 'value'
 );
 
 USE CATALOG my_catalog;
 ```
+
+The property `'table-default.key' = 'value'` specifies the default property `'key' = 'value'` for tables created in the catalog.
+
+You can define any default properties with the prefix `table-default.`. If the table defines the same properties with the default properties, the table's properties will be taken.
+
 
 {{< /tab >}}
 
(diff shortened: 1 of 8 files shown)


legendtkl:
hi, @JingsongLi  @FangYongs  , would you help review this?

Thanks.

paimon-common/src/main/java/org/apache/paimon/options/CatalogOptions.java
```diff
@@ -75,4 +77,18 @@ public class CatalogOptions {
...
+                {
+                    add(WAREHOUSE.key());
+                    add(METASTORE.key());
+                    add(URI.key());
+                    add(TABLE_TYPE.key());
+                    add(LOCK_ENABLED.key());
+                    add(LOCK_CHECK_MAX_SLEEP.key());
+                    add(LOCK_ACQUIRE_TIMEOUT.key());
+                    add(FS_ALLOW_HADOOP_FALLBACK.key());
+                }
```
FangYongs, line 92: Can we get all fields from `CatalogOptions` instead of add them manually? Such as `FlinkConnectorOptions#getOptions`, we can add a tool class for them

JingsongLi:
Thanks @legendtkl for the contribution.

I wonder if I need a prefix to identify this as the default parameter for table.
Is there a similar example in the industry?

FangYongs:
@JingsongLi @legendtkl Should we need specific prefix for different components? What do you think?

legendtkl:
> Thanks @legendtkl for the contribution.
> 
> I wonder if I need a prefix to identify this as the default parameter for table. Is there a similar example in the industry?

@JingsongLi 

Yes, this is a good idea to make the behavior more determinate. 

The flink kafka connector properties follow this convention to use `properties.` prefix for any kafka configuration. https://nightlies.apache.org/flink/flink-docs-release-1.17/docs/connectors/table/kafka/#properties. Let's do it in the same way. 

What about the prefix value? "table.properties.",  or "global.properties", or anything else? Do you have any good idea about this?

Thanks.

JingsongLi:
> > Thanks @legendtkl for the contribution.
> > I wonder if I need a prefix to identify this as the default parameter for table. Is there a similar example in the industry?
> 
> @JingsongLi
> 
> Yes, this is a good idea to make the behavior more determinate.
> 
> The flink kafka connector properties follow this convention to use `properties.` prefix for any kafka configuration. https://nightlies.apache.org/flink/flink-docs-release-1.17/docs/connectors/table/kafka/#properties. Let's do it in the same way.
> 
> What about the prefix value? "table.properties.", or "global.properties", or anything else? Do you have any good idea about this?
> 
> Thanks.

Thanks, let me take a look to others.

legendtkl:
> > > Thanks @legendtkl for the contribution.
> > > I wonder if I need a prefix to identify this as the default parameter for table. Is there a similar example in the industry?
> > 
> > 
> > @JingsongLi
> > Yes, this is a good idea to make the behavior more determinate.
> > The flink kafka connector properties follow this convention to use `properties.` prefix for any kafka configuration. https://nightlies.apache.org/flink/flink-docs-release-1.17/docs/connectors/table/kafka/#properties. Let's do it in the same way.
> > What about the prefix value? "table.properties.", or "global.properties", or anything else? Do you have any good idea about this?
> > Thanks.
> 
> Thanks, let me take a look to others.

Ping me if you have any good idea or update. @JingsongLi

JingsongLi:
Maybe Maybe we can introduce `table-default.` for this.

JingsongLi:
https://cloud.google.com/bigquery/docs/default-configuration also has default options.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkCatalog.java
```diff
@@ -210,11 +212,31 @@ public void createTable(ObjectPath tablePath, CatalogBaseTable table, boolean ig
...
 
         // remove table path
-        String specific = options.remove(PATH.key());
-        if (specific != null) {
-            catalogTable = catalogTable.copy(options);
+        options.remove(PATH.key());
+
+        // Put Paimon catalog global options into table options
+        // ref https://github.com/apache/incubator-paimon/issues/754
+        if (catalog instanceof org.apache.paimon.catalog.AbstractCatalog) {
```
JingsongLi, line 219: We should put this into `AbstractCatalog`. Other engines should have this capability too.
In `AbstractCatalog`:
```
protected AbstractCatalog(FileIO fileIO, Map<String, String> catalogOptions) {
        this.fileIO = fileIO;
        this.tableDefaultOptions = ...;
}
```

And copy Schema in `createTable` with table default options.

paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
```diff
@@ -95,6 +116,12 @@ protected void checkNotSystemTable(Identifier identifier, String method) {
         }
     }
 
+    protected void copyTableDefaultOptions(Map<String, String> options) {
+        tableDefaultOptions.keySet().stream()
+                .filter(key -> !options.containsKey(key))
```
JingsongLi, line 121: `putIfAbsent`?

JingsongLi:
Please rebase master, thanks!

paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
```diff
@@ -95,6 +116,12 @@ protected void checkNotSystemTable(Identifier identifier, String method) {
         }
     }
 
+    protected void copyTableDefaultOptions(Map<String, String> options) {
+        tableDefaultOptions
```
JingsongLi, line 120: NIT: `tableDefaultOptions.forEach`

JingsongLi (commented):
Hi @legendtkl , can you document this? Maybe in `create catalog` page?

legendtkl:
> Hi @legendtkl , can you document this? Maybe in `create catalog` page?

Yes, good catch.

docs/content/how-to/creating-catalogs.md
```diff
@@ -113,6 +126,8 @@ spark-sql ... \
     --conf spark.sql.catalog.paimon.uri=thrift://<hive-metastore-host-name>:<port>
 ```
 
+You can also define any default properties with the prefix `spark.sql.catalog.paimon.table_default.` configuration for tables as above `Catalog with Filesystem Metastore`.
```
JingsongLi, line 129: `table_default` works?

JingsongLi (approved):
+1

PR #777, opened 2023-03-30
[doc] Path style config for S3

wg1026688210:
Add path style config doc  for some  S3 compliant  object stores which  might not have virtual host style addressing enabled by default.

### DIFF
--- a/docs/content/filesystems/s3.md
+++ b/docs/content/filesystems/s3.md
@@ -138,3 +138,13 @@ The S3 Filesystem also support using S3 compliant object stores such as IBM’s 
 ```yaml
 s3.endpoint: your-endpoint-hostname
 ```
+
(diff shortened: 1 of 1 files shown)


docs/content/filesystems/s3.md
```diff
@@ -71,6 +71,13 @@ CREATE CATALOG my_catalog WITH (
 );
 ```
 
+Some S3 compliant object stores might not have virtual host style addressing enabled by default, for example when using Standalone MinIO for testing purpose.
+In such cases, you will have to provide the property to enable path style access in flink-conf.yaml.
+
+```sql
+'s3.path.style.access' = 'true'
```
JingsongLi, line 78: We can add this documentation at the bottom.

PR #790, opened 2023-03-31
[flink]add test for NaN and Infinity type

zhangjun0x01:

### Purpose

add test for NaN and Infinity type,

I test that ,and found it is fine  for issue https://github.com/apache/incubator-paimon/issues/736

### Tests

*org.apache.paimon.flink.ReadWriteTableITCase#testNaNType*

### API and Format 

### Documentation


JingsongLi (approved):
+1

PR #791, opened 2023-03-31
[core] Introduce full-compaction.delta-commits and read full-compacted mode

JingsongLi:
The PR is aim to provide a mode for Read Performance.

For Primary Key Table, it's a 'MergeOnRead' technology. When reading data, multiple layers of LSM data are merged,
and the number of parallelism will be limited by the number of buckets. Although Paimon's merge will be very efficient,
it still cannot catch up with the ordinary AppendOnly table.

If you want to query fast enough in certain scenarios, but can only find older data, you can:

1. Configure 'full-compaction.delta-commits', when writing data (currently only Flink), full checkpoint will be performed periodically.
2. Configure 'scan.mode' to 'compacted-full', when reading data, snapshot of full compaction is picked. Read performance is good.

You can flexibly balance query performance and data latency when reading.


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/GlobalFullCompactionSinkWrite.java
```diff
@@ -0,0 +1,254 @@
...
+        if (earliestId == null) {
+            return;
+        }
+
+        for (long id = latestId; id >= earliestId; id--) {
+            Snapshot snapshot = snapshotManager.snapshot(id);
+            if (snapshot.commitUser().equals(commitUser)
+                    && snapshot.commitKind() == Snapshot.CommitKind.COMPACT) {
+                long commitIdentifier = snapshot.commitIdentifier();
+                if (commitIdentifiersToCheck.contains(commitIdentifier)) {
```
yuzelin, line 195: Is `commitIdentifiersToCheck` only used for log purpose?

yuzelin (approved):
LGTM

paimon-core/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -542,6 +542,13 @@ public class CoreOptions implements Serializable {
                     .defaultValue(1024)
                     .withDescription("Read batch size for orc and parquet.");
 
+    public static final ConfigOption<Integer> FULL_COMPACTION_DELTA_COMMITS =
+            key("full-compaction.delta-commits")
+                    .intType()
+                    .noDefaultValue()
+                    .withDescription(
+                            "Full compaction will be constantly triggered after delta commits.");
+
```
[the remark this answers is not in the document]
JingsongLi: Only full-compaction changelog is enabled, the default value is 1.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/GlobalFullCompactionSinkWrite.java
JingsongLi: Take a look to `checkSuccessfulFullCompaction`.

Issue #794, opened 2023-03-31
[Feature] Display the specific table name in the error message when the table doesn't support lookup join


YesOrNo828:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/incubator-paimon/issues) and found nothing similar.

### Motivation

Display the specific table name in the error message when the table doesn't support lookup join, especially when there are multiple lookup tables joined in a Flink SQL.

### Solution

Throw the exception with the lookup table name.

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

PR #795, opened 2023-03-31
[Paimon-794][Improvement] Display the specific table name in the error message when the table doesn't support lookup join

YesOrNo828:
fix: #794 
module: flink

### Purpose

Show more detailed msg which include the table name when the table doesn't support the lookup join ability.

### Tests

*(List UT and IT cases to verify this change)* no

### API and Format 

### Documentation


JingsongLi:
+1 merged

PR #805, opened 2023-03-31
[core] Clean TableScan related codes

yuzelin:
### Purpose

Delete useless `DataTableScan` and `StreamDataTableScan`.

Add a internal option.

### Tests

Original Tests.

### API and Format 

No.

### Documentation

No.


yuzelin:
8787D4AE-E7E8-4703-A7C8-C272A567A4B2
Verified.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/CompactorSourceBuilder.java
```diff
@@ -92,14 +93,28 @@ public CompactorSourceBuilder withPartitions(List<Map<String, String>> partition
                                     .toArray(Predicate[]::new));
         }
 
-        ReadBuilder readBuilder = bucketsTable.newReadBuilder().withFilter(partitionPredicate);
         if (isContinuous) {
+            // set 'streaming-compact' and remove 'scan.bounded.watermark' if exists
+            Options options = bucketsTable.coreOptions().toConfiguration();
+            options.set(CoreOptions.STREAMING_COMPACT, true);
```
JingsongLi, line 99: create a new options to set.

yuzelin:
D9AE81F0-11AC-425D-9326-F5CF44A5E26C
Verified.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/CompactorSourceBuilder.java
```diff
@@ -92,14 +93,27 @@ public CompactorSourceBuilder withPartitions(List<Map<String, String>> partition
...
         }
 
-        ReadBuilder readBuilder = bucketsTable.newReadBuilder().withFilter(partitionPredicate);
         if (isContinuous) {
+            // set 'streaming-compact' and remove 'scan.bounded.watermark' if exists
+            Options options = Options.fromMap(bucketsTable.options());
+            options.set(CoreOptions.STREAMING_COMPACT, true);
+            Map<String, String> newOptions = options.toMap();
+            if (options.contains(CoreOptions.SCAN_BOUNDED_WATERMARK)) {
+                newOptions.remove(CoreOptions.SCAN_BOUNDED_WATERMARK.key());
```
JingsongLi, line 102: Do you have tests on this?
Dynamic options are new options, here can not just remove `SCAN_BOUNDED_WATERMARK`. Maybe we should pass a `Long.MAX`.

yuzelin: `SCAN_BOUNDED_WATERMARK`, `SCAN_TIMESTAMP_MILLIS`, `SCAN_SNAPSHOT_ID` may have conflict with new options when checking.

I think it's more reasonable to pass a map like `scan.bounded.watermark -> null`, then in `Table#copy`, we can remove the `ConfigOption` if passing value is null. WDYT?

JingsongLi: `null` looks good to me.

yuzelin:
684CCA12-31FF-470F-9131-34C40769266E
Verified.

JingsongLi (approved):
+1

PR #808, opened 2023-04-01
[hotfix] fix scheduled daily jdk11 workflow

legendtkl:
### Purpose
hotfix for jdk11 daily github workflow

### Test
Because the workflow works only on default branch (master), I test manually in personal repo.

image


JingsongLi (approved):
+1

PR #816, opened 2023-04-03
[Bug] Timestamp LTZ is unsupported for orc

JingsongLi:
Fix Timestamp LTZ support for orc and flink.

NOTE: hive spark trino, parquet still not support Timestamp LTZ.


yuzelin (approved):
LGTM!

PR #819, opened 2023-04-03
[pom] Rename paimon-shade to paimon-bundle

JingsongLi:
### Purpose

https://github.com/apache/incubator-paimon-shade contains a number of shaded dependencies.

But paimon-shade is in main repo. We should rename it to avoid conflicts with incubator-paimon-shade.


yuzelin (approved):
LGTM!

Issue #820, opened 2023-04-03
[Bug][Flink] Calculate splits with predicate in DataTableSource


FangYongs:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/incubator-paimon/issues) and found nothing similar.

### Paimon version

Calculate splits with predicate in DataTableSource

### Compute Engine

flink

### Minimal reproduce step

None

### What doesn't meet your expectations?

None

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #821, opened 2023-04-03
[Bug][Flink] Calculate splits with predicate in DataTableSource

FangYongs:
### Purpose

Calculate splits with predicate in DataTableSource for #820 

### Tests

Test case `testInferParallelism` in `ReadWriteTableITCase`

### API and Format 

### Documentation

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/DataTableSource.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/DataTableSource.java
@@ -34,10 +34,8 @@ import org.apache.paimon.predicate.Predicate;
 import org.apache.paimon.table.AppendOnlyFileStoreTable;
 import org.apache.paimon.table.ChangelogValueCountFileStoreTable;
 import org.apache.paimon.table.ChangelogWithKeyFileStoreTable;
-import org.apache.paimon.table.DataTable;
 import org.apache.paimon.table.Table;
 import org.apache.paimon.table.source.Split;
-import org.apache.paimon.utils.Preconditions;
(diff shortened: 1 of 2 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/DataTableSource.java
```diff
@@ -223,7 +224,11 @@ private DataStreamSource<RowData> configureSource(
 
                 Preconditions.checkState(table instanceof DataTable);
                 DataTable dataTable = (DataTable) table;
-                splits = dataTable.newScan().plan().splits();
+                BatchDataTableScan tableScan = dataTable.newScan();
```
JingsongLi, line 227: `TableScan tableScan = table.newReadBuilder().withFilter(predicate).newScan();`?

JingsongLi (commented):
Thanks @FangYongs for the contribution, left comment.

FangYongs:
Thanks @JingsongLi , updated

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/DataTableSource.java
```diff
@@ -222,8 +222,7 @@ private DataStreamSource<RowData> configureSource(
             } else {
 
                 Preconditions.checkState(table instanceof DataTable);
```
JingsongLi, line 224: we can remove this check?

FangYongs: Agree, I'll remove it

PR #824, opened 2023-04-04
[flink] MySqlSyncTableAction now supports flink run

tsreaper:
### Purpose

`MySqlSyncTableAction` now supports being used from `bin/flink run`. Users can use `MySqlSyncTableAction` without writing and compiling their own data stream jobs.

### Tests

* MySql57E2eTest
* MySql80E2eTest

### API and Format 

N/A

### Documentation

Document will be introduced in the next PR.

### DIFF
--- a/paimon-flink/paimon-flink-1.14/src/main/java/org/apache/paimon/flink/utils/SingleOutputStreamOperatorUtils.java
+++ b/paimon-flink/paimon-flink-1.14/src/main/java/org/apache/paimon/flink/utils/SingleOutputStreamOperatorUtils.java
@@ -1,0 +1,36 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.paimon.flink.utils;
+
+import org.apache.flink.streaming.api.datastream.DataStream;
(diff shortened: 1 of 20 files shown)


paimon-flink/paimon-flink-1.14/src/main/java/org/apache/paimon/flink/utils/SingleOutputStreamOperatorUtils.java
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Utility methods for {@link SingleOutputStreamOperator}.
+ *
+ * <p>TODO: merge all Flink version related utils into one class.
+ */
+public class SingleOutputStreamOperatorUtils {
+
+    public static <T> DataStream<T> getSideOutput(
+            SingleOutputStreamOperator<?> input, OutputTag<T> outputTag) {
+        return input.getSideOutput(outputTag);
```
JingsongLi, line 34: What is different between 1.14 and 1.17?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/cdc/mysql/MySqlSyncTableAction.java
```diff
@@ -118,13 +123,13 @@ public class MySqlSyncTableAction implements Action {
...
     }
 
-    @Override
-    public void run() throws Exception {
-        StreamExecutionEnvironment env = StreamExecutionEnvironment.getExecutionEnvironment();
-        build(env);
-        env.execute(String.format("MySQL CTAS: %s.%s", database, table));
+    private void setDefaultMySqlConfig() {
+        // https://ververica.github.io/flink-cdc-connectors/master/content/connectors/mysql-cdc.html#connector-options
+        mySqlConfig.putIfAbsent("port", "3306");
```
JingsongLi, line 132: Can we convert `mySqlConfig` to Flink `Configuration`?
Then we can get from it directly.

JingsongLi (commented):
Left two comments.

paimon-flink/paimon-flink-1.14/src/main/java/org/apache/paimon/flink/utils/SingleOutputStreamOperatorUtils.java
tsreaper: Their return types are different. In 1.14 and 1.15, return type of `SingleOutputStreamOperator#getSideOutput` is just `DataStream`, while in 1.16 and 1.17, return type is the newly introduced `SideOutputStream`. This `SideOutputStream` does not exist in 1.14 and 1.15.

JingsongLi (approved):
+1

PR #834, opened 2023-04-04
[core] Introduce SecurityContext

JingsongLi:
### Purpose

Introduce a SecurityContext, this can be used by trino, java api and etc...

### Tests

KerberosLoginProviderTest

### API and Format 

introduce `SecurityContext`.

### Documentation

hdfs.html


Dkbei:
+1

PR #844, opened 2023-04-06
[core][flink] Refactor BucketComputer and StreamPartitioner to gather bucketing / partitioning core logic in the same class

tsreaper:
### Purpose

Currently a lot of classes have bucketing / partitioning logic, such as `BucketComputer`, `SinkRecordConverter`, `BucketStreamPartitioner`, `CdcRecordStreamPartitioner`, etc. These bucketing / partitioning logic must be kept in sync.

By refactoring these logic into the same class, we can avoid problems like when creating a new stream partitioner we accidentally write the wrong partitioning or bucketing logic.

### Tests

Existing tests can cover the changes.

### API and Format 

N/A

### Documentation

N/A


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/BucketingStreamPartitioner.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    @Override
+    public boolean isPointwise() {
+        return false;
+    }
+
+    @Override
+    public String toString() {
+        return "bucketing-stream-partitioner";
```
JingsongLi, line 73: This is displayed in Flink UI. We should improve this.
maybe:
`HASH[bucket]` and `HASH[bucket, partition]`?

JingsongLi (approved):
+1

Issue #773, opened 2023-03-30
[Bug] When writing a partitioned table for Paimon using Flink SQL and storing the table on HDFS, the partition directory name on HDFS is displayed abnormally. For example, it shows as SCAN_DATE=19415 instead of the correct format of SCAN_DATE=2023-03-01, where SCAN_DATE is of type date.


yuanoOo:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/incubator-paimon/issues) and found nothing similar.

### Paimon version

paimon-flink-1.15-0.4-20230330.002117-12.jar

### Compute Engine

flink1.15.3

### Minimal reproduce step

When writing a partitioned table for Paimon using Flink SQL and storing the table on HDFS, the partition directory name on HDFS is displayed abnormally. For example, it shows as SCAN_DATE=19415 instead of the correct format of SCAN_DATE=2023-03-01, where SCAN_DATE is of type date.

### What doesn't meet your expectations?

When writing a partitioned table for Paimon using Flink SQL and storing the table on HDFS, the partition directory name on HDFS is displayed abnormally. For example, it shows as SCAN_DATE=19415 instead of the correct format of SCAN_DATE=2023-03-01, where SCAN_DATE is of type date.

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

JingsongLi:
Direct fixes may have compatibility issues, we should be careful.

yuanoOo:
> Direct fixes may have compatibility issues, we should be careful.

I’m trying to fix this bug, can you assign it to me?

My idea is to format the DATE type in the org.apache.paimon.utils.RowDataPartitionComputer#generatePartValues method. For example, converting the int value 19996 of the DATE type into the correct date string 2024-09-30.

dailai:
I think this issue need to fix. For example, we develop a paimon connector of seatunnel which use java api of paimon also has the same problem.The types in date,time,timestamp all have the same problem.

dailai:
@JingsongLi Hi, how about this problem?

PR #853, opened 2023-04-09
[core][bug] Fix the incorrect partition key format generated from the DATE type

yuanoOo:
fix issue： #773 

### Purpose

### Tests

### API and Format 

### Documentation


paimon-core/src/main/java/org/apache/paimon/utils/RowDataPartitionComputer.java
```diff
@@ -52,7 +57,18 @@ public LinkedHashMap<String, String> generatePartValues(InternalRow in) {
 
         for (int i = 0; i < partitionFieldGetters.length; i++) {
             Object field = partitionFieldGetters[i].getFieldOrNull(in);
-            String partitionValue = field != null ? field.toString() : null;
+            String partitionValue;
+
+            if (Objects.nonNull(field)) {
+                if (fieldTypes.get(i).is(DataTypeRoot.DATE)) {
```
JingsongLi, line 63: also check Time Type

JingsongLi:
I've thought about this fix, maybe we can make it more thorough. Like `org.apache.paimon.utils.TypeUtils.castFromString`, we can have a `castToString`. Then, the string will be SQL standard string.

But this will bring more serious issues, incompatibility, and we may need an option or an old class to fall back to the old path when we cannot find it. This is complicated.

Is this problem serious now? Can we keep it this way for now?

yuanoOo:
> I've thought about this fix, maybe we can make it more thorough. Like `org.apache.paimon.utils.TypeUtils.castFromString`, we can have a `castToString`. Then, the string will be SQL standard string.
> 
> But this will bring more serious issues, incompatibility, and we may need an option or an old class to fall back to the old path when we cannot find it. This is complicated.
> 
> Is this problem serious now? Can we keep it this way for now?

I think it's a very user experience affecting issue. 
In a stream writing paimon scenario, it is very common to cast a field of timestamp type to DATE type in order to write to the corresponding partition.
However, in paimon, when the user uses a partition of type DATE, the partition path becomes a strange unreadable string of numbers, which may be unacceptable. If you want to get the correct partition string, but need to convert timestamp type to date string, which is counterintuitive and sql is tedious.

I think the current addition of support for DATE type partitioning is sufficient. 

For the compatibility issue, it will be easy to solve if we can get the version information of paimon. For example, we can determine how to format DATE type partitions based on the version number. I think there are probably few users who use it for the strange performance of the current version of the DATE type partition.

SteNicholas:
@yuanoOo, the partition key format is user-defined that the format like `yyyyMMdd` could convert from the date or timestamp functions via users. It doesn't need to use fixed format in Paimon. Meanwhile, solving the compatibility issue via version doesn't mark sense to me because it's cumbersome to maintain the compatibility solution.

dailai:
I think this issue need to fix. For example, we develop a paimon connector of seatunnel which use java api of paimon also has the same problem.

PR #856, opened 2023-04-10
[flink] Assign splits evenly in StaticFileStoreSplitEnumerator

JingsongLi:
### Purpose

Assign splits evenly in StaticFileStoreSplitEnumerator, this can avoid the long tail problem caused by uneven distribution.

### Tests

FixBinPackingTest


paimon-common/src/main/java/org/apache/paimon/utils/FixBinPacking.java
```diff
@@ -0,0 +1,68 @@
...
+            long weight = weightFunc.apply(item);
+            Bin<T> bin = bins.size() < binNumber ? new Bin<>() : bins.poll();
+            bin.add(item, weight);
+            bins.add(bin);
+        }
+
+        // 3. output
+        bins.forEach(bin -> packed.add(bin.items));
+        return packed;
+    }
```
yuzelin, line 52: Better to move Line 40 to close to Line 50.

yuzelin (approved):
LGTM

PR #857, opened 2023-04-10
[core] Compacted StartingScanner is incorrect

### LINKED ISSUE #855: [Bug] Compacted StartingScanner is incorrect
### Search before asking

### Paimon version

0.4

### Compute Engine

flink

### Minimal reproduce step

- full-compaction.delta-commits = 10
- stop job and wait a long time.
- start job
- query table with `compacted-latest`, the result will be empty.

When Compacted StartingScanner can not find a compacted snapshot, it should return a normal snapshot.

### Anything else?

_No response_

JingsongLi:
fix #855


yuzelin (approved):
LGTM

PR #858, opened 2023-04-10
[core]Support Level Compaction

zhangjun0x01:

### Purpose

*(provide level compaction, minimizes space amplification at the cost of read and write amplification.)*

### Tests

*(org.apache.paimon.mergetree.compact.LevelCompactionTest)*

### API and Format 

### Documentation

### DIFF
--- a/docs/layouts/shortcodes/generated/core_configuration.html
+++ b/docs/layouts/shortcodes/generated/core_configuration.html
@@ -374,5 +374,29 @@
             <td>Boolean</td>
             <td>If set to true, compactions and snapshot expiration will be skipped. This option is used along with dedicated compact jobs.</td>
         </tr>
+        <tr>
+            <td><h5>compaction.style</h5></td>
+            <td style="word-wrap: break-word;">universal</td>
+            <td>Enum</td>
+            <td>Specify the compaction style for paimon.<br /><br />Possible values:<ul><li>"universal": </li><li>"level": </li></ul></td>
+        </tr>
+        <tr>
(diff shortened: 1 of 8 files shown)


wg1026688210:
hi~ @zhangjun0x01   ,the problem sonar reported has been resolved, you can rebase the master branch

zhangjun0x01:
> hi~ @zhangjun0x01 ,the problem sonar reported has been resolved, you can rebase the master branch

thanks,I update it

JingsongLi:
Why do we support level compaction? Is there any data result support on the benchmark?

JingsongLi:
I prefer not to consider supporting it at the moment, as analytical systems such as clickhouse and even HBase do not have a good application of level compaction.

If you have more input, please let me know.

zhangjun0x01:
> I prefer not to consider supporting it at the moment, as analytical systems such as clickhouse and even HBase do not have a good application of level compaction.
> 
> If you have more input, please let me know.

sorry for the late reply， my initial idea was，universal compaction use lower write amplification, trading off read amplification and space amplification. leveled compaction minimizes space amplification at the cost of read and write amplification. we should add multiple compaction styles so that users can choose based on their actual scenarios, just like other storage engines, such as Rocksdb and Cassandra

JingsongLi:
Thank you for your explanation, we should write code based on the user's needs, if it is a requirement we set out on our own, we should do sufficient testing to prove our conjecture, there is a lack of evidence that this PR is useful for the user scenario.

JingsongLi:
You can initiate a PIP if you have enough tests afterwards.

PR #861, opened 2023-04-10
[format] upgrate parquet  to 1.13.0

zhangjun0x01:

### Purpose

*parquet 1.13.0 release notes : https://github.com/apache/parquet-mr/blob/apache-parquet-1.13.0/CHANGES.md?plain=1#L22-L78*

upgrate parquet to 1.13.0

### Tests

### API and Format 

### Documentation

### DIFF
--- a/paimon-format/pom.xml
+++ b/paimon-format/pom.xml
@@ -32,7 +32,7 @@ under the License.
     <name>Paimon : Format</name>
 
     <properties>
-        <parquet.version>1.12.3</parquet.version>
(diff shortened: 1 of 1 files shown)


zhangjun0x01:
image

JingsongLi:
I feel that upgrading to 1.13 may involve certain risks and the possibility of incompatibility.
We don't need to upgrade it now. After the smaller version is fixed a bit, we will consider it again.

PR #862, opened 2023-04-10
[test] resolve the problem which Sonar reported that InterruptedException need to re-throw or re-interrupt

wg1026688210:
resolve the problem which Sonar reported that InterruptedException need to re-throw or re-interrupt


wg1026688210:
image

JingsongLi (approved):
+1 Thanks @wg1026688210

PR #865, opened 2023-04-10
[core][bug] BoundedChecker may check non-existing snapshot at fully read phase

yuzelin:
### Purpose

At fully read phase of streaming scan,  the snapshot may not existed in some start up modes. BoundedChecker should not check the snapshot.

### Tests

`StreamTableScanTest#testStartingFromNonExistingSnapshot`

### API and Format 

No
### Documentation

No

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/source/snapshot/StartingScanner.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/source/snapshot/StartingScanner.java
@@ -37,8 +37,8 @@ public interface StartingScanner {
         List<DataSplit> splits();
     }
 
-    /** A null Result. */
-    class NullResult implements Result {
+    /** Currently, there is no snapshot, need to wait for the snapshot to be generated. */
+    class NoSnapshot implements Result {
         @Override
         public long snapshotId() {
             throw new UnsupportedOperationException();
@@ -50,23 +50,19 @@ public interface StartingScanner {
         }
     }
 
-    /** A non null Result. */
-    abstract class NonNullResult implements Result {
-        private final long snapshotId;
+    /** Result with scanned snapshot. Next snapshot should be the current snapshot plus 1. */
+    class ScannedResult implements Result {
+        private final long currentSnapshotId;
         private final List<DataSplit> splits;
 
-        public NonNullResult(long snapshotId) {
-            this(snapshotId, Collections.emptyList());
-        }
-
-        public NonNullResult(long snapshotId, List<DataSplit> splits) {
-            this.snapshotId = snapshotId;
+        public ScannedResult(long currentSnapshotId, List<DataSplit> splits) {
+            this.currentSnapshotId = currentSnapshotId;
             this.splits = splits;
         }
 
         @Override
         public long snapshotId() {
-            return snapshotId;
+            return currentSnapshotId;
         }
 
         @Override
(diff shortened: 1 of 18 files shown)


paimon-core/src/main/java/org/apache/paimon/table/source/snapshot/StartingScanner.java
```diff
@@ -37,14 +37,17 @@ public interface StartingScanner {
     class Result {
```
JingsongLi, line 37: Can we just return a interface, and scanner returns different implement classes.

paimon-core/src/main/java/org/apache/paimon/table/source/snapshot/StartingScanner.java
```diff
@@ -22,38 +22,74 @@
...
     /** Scan result of {@link #scan}. */
-    class Result {
+    interface Result {
+        long snapshotId();
+
+        List<DataSplit> splits();
+    }
+
+    /** A null Result. */
+    class NullResult implements Result {
```
JingsongLi, line 41: `NoSnapshot`, and add comment: Currently, there is no snapshot, need to wait for the snapshot to be generated.

paimon-core/src/main/java/org/apache/paimon/table/source/snapshot/StartingScanner.java
```diff
@@ -22,38 +22,74 @@
...
+        }
+
+        @Override
+        public List<DataSplit> splits() {
+            throw new UnsupportedOperationException();
+        }
+    }
 
+    /** A non null Result. */
+    abstract class NonNullResult implements Result {
```
JingsongLi, line 54: No need to have this class.

paimon-core/src/main/java/org/apache/paimon/table/source/snapshot/StartingScanner.java
```diff
@@ -22,38 +22,74 @@
...
         }
 
+        @Override
         public List<DataSplit> splits() {
             return splits;
         }
     }
+
+    /** Contains an existing snapshot. */
+    class ExistingSnapshotResult extends NonNullResult {
```
JingsongLi, line 79: `ScannedResult`: Result with scanned snapshot. Next snapshot should be current snapshot plus 1.

JingsongLi, line 79:
```
private final long currentSnapshotId;
private final List<DataSplit> splits;
```

paimon-core/src/main/java/org/apache/paimon/table/source/snapshot/StartingScanner.java
```diff
@@ -22,38 +22,74 @@
...
+            super(snapshotId);
+        }
+
+        public ExistingSnapshotResult(long snapshotId, List<DataSplit> splits) {
+            super(snapshotId, splits);
+        }
+    }
+
+    /** Contains a non-existing snapshot. */
+    class NonExistingSnapshotResult extends NonNullResult {
```
JingsongLi, line 90: `NextSnapshot`: add some comments.
```
private final long nextSnapshotId;
```

yuzelin:
52B7C06A-9E06-40F7-B7C0-195F2FA50017

yuzelin:
B3A6F76E-D6CE-4E2C-B9BB-8DC61D0FEA2E

paimon-core/src/main/java/org/apache/paimon/table/source/snapshot/StartingScanner.java
```diff
@@ -22,38 +22,75 @@
...
 
 /** Helper class for the first planning of {@link TableScan}. */
 public interface StartingScanner {
 
-    @Nullable
     Result scan(SnapshotManager snapshotManager, SnapshotSplitReader snapshotSplitReader);
 
     /** Scan result of {@link #scan}. */
-    class Result {
+    interface Result {
```
JingsongLi, line 34: I think this `Result` class can be an empty interface.

yuzelin:
C54852D8-144B-4DFE-9A7D-FAE76835BEDA

JingsongLi (approved):
+1

PR #867, opened 2023-04-10
[core] Compare batch maxKey to reduce comparisons in SortMergeR…

### LINKED ISSUE #753: [Feature] Compare batch maxKey to reduce comparisons in SortMergeReader
### Search before asking

### Motivation

In SortMergeReader, each sub reader is batched reader.

When adding a new batch to the priority queue, we can look at the maximum key of the batch, and if its maximum key is smaller than the minimum key of other batches, then we can just output the whole batch.

### Solution

_No response_

### Anything else?

_No response_

schnappi17:
### Purpose

fix #753 

### Tests

`RecordReaderRegionTest`
`RecordReaderRegionManagerTest`

### Documentation

- Introduce `SortedRegionDataRecordReader` to read a file with `minKey` and `maxKey` in a `SortedRun`
- Introduce `RecordReaderSubRegion` includes a list of `SortedRegionDataRecordReader`, which is corresponding to a `SortedRun`
- Introduce `RecordReaderRegion` to merge multiple `RecordReaderSubRegion` which have intersection of keys. `RecordReaderRegion`s are splitted by introduced `RecordReaderRegionManager`, there are no intersection between `RecordReaderRegion`s
- If there is only one reader for a `RecordReaderRegion`, use this reader directly. Otherwise, create `SortMergeReader` for the `RecordReaderRegion`.

Will provide benchmark later to verify the performance improvement.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/mergetree/MergeTreeReaders.java
+++ b/paimon-core/src/main/java/org/apache/paimon/mergetree/MergeTreeReaders.java
@@ -27,6 +27,10 @@ import org.apache.paimon.mergetree.compact.MergeFunction;
 import org.apache.paimon.mergetree.compact.MergeFunctionWrapper;
 import org.apache.paimon.mergetree.compact.ReducerMergeFunctionWrapper;
 import org.apache.paimon.mergetree.compact.SortMergeReader;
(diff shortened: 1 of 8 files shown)


JingsongLi:
Lack benchmark?

PR #869, opened 2023-04-10
Imgrate MinioTestContainer from the module of spark to paimon-s3 

wg1026688210:
Imgrate MinioTestContainer from the module of spark to paimon-s3 ,in order that orther module such as flink,hive,presto can test the s3 FileSystem.


JingsongLi (approved):
+1

Issue #751, opened 2023-03-29
[Feature] Introduce CatalogTestBase


JingsongLi:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/incubator-paimon/issues) and found nothing similar.

### Motivation

Currently, only tests for FlinkCatalog or ITCase, we should add cases for catalogs.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

zhuangchong:
@JingsongLi I can try to do this, can you assign me.

PR #872, opened 2023-04-11
[test] Introduce CatalogTestBase

zhuangchong:

### Purpose

this pr closes: #751 

### Tests

### API and Format 

### Documentation

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
@@ -109,6 +109,14 @@ public interface Catalog extends AutoCloseable {
     List<String> listTables(String databaseName) throws DatabaseNotExistException;
 
     /**
+     * Check if a table exists in this catalog.
+     *
+     * @param identifier Path of the table
+     * @return true if the given table exists in the catalog false otherwise
+     */
+    boolean tableExists(Identifier identifier);
+
+    /**
      * Drop a table.
      *
      * <p>NOTE: System tables can not be dropped.
--- a/paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
@@ -114,7 +114,13 @@ public interface Catalog extends AutoCloseable {
      * @param identifier Path of the table
      * @return true if the given table exists in the catalog false otherwise
      */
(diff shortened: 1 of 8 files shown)


paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
```diff
@@ -108,6 +108,14 @@ void dropDatabase(String name, boolean ignoreIfNotExists, boolean cascade)
      */
     List<String> listTables(String databaseName) throws DatabaseNotExistException;
 
+    /**
+     * Check if a table exists in this catalog.
+     *
+     * @param identifier Path of the table
+     * @return true if the given table exists in the catalog false otherwise
+     */
+    boolean tableExists(Identifier identifier);
```
JingsongLi, line 117: I removed this method because it is very confused if we have system tables. You can find always return false for system tables in current implementations.

JingsongLi, line 117: If we just have a syntactic sugar:
```
  default boolean tableExists(Identifier ident) {
    try {
      return loadTable(ident) != null;
      return false;
    }
  }
```
OK, that is fine.

paimon-core/src/test/java/org/apache/paimon/catalog/CatalogTestBase.java
```diff
@@ -0,0 +1,504 @@
...
+import java.util.Collections;
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatCode;
+import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+/** Base test class of paimon catalog in {@link Catalog}. */
+public abstract class CatalogTestBase {
```
JingsongLi, line 47: Add tests for system tables.

paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
zhuangchong: Thanks, done.

paimon-core/src/test/java/org/apache/paimon/catalog/CatalogTestBase.java
zhuangchong: Thanks, done.

PR #877, opened 2023-04-11
[core] Refactor snapshot record history

### LINKED ISSUE #781: [Feature] Refactor the recording method of snapshot linear history
### Search before asking

### Motivation

Apache Paimon currently uses the file prefix and file system List command to manage the linear history of snapshots, but this method has the following drawbacks:

- It is not friendly with object storage, where operations such as List\Rename are very expensive.
- This method has a single linear relationship, which is not conducive to creating complex relationships, such as create a new branch\tag.

### Solution

Store the file path of the previous Snapshot directly in the Snapshot object.

### Anything else?

While we can easily manage snapshot linear history, we can implement many advanced features based on it, such as Branch.

hililiwei:
fix #781

### Purpose

Refactor the snapshot recording method of Paimon

Why do we need this change? 

First of all, a linear history can enable us to achieve some advanced features, such as branching. Secondly, relying on the file system management mode is unfriendly, especially for object storage. We should try to avoid using List\Rename and other operations that are very expensive on object storage. Finally, this is conducive to our transaction management.

### Documentation

Add a parentId field to the Snapshot object to record its ancestor relationship, and make the snapshot record a linear chain.

Snapshot:

| filed | type |      |
| ------------------- | ---- | ---- |
| parentId             |  Long    | The parent snapshot id     |

* `-1` it should be root snapshot. 
* `null` it is a old version,
* `> -1`  parent id of current snapshot

##### Compatibility

If parentId is null, it is a old version snapshot.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/Snapshot.java
+++ b/paimon-core/src/main/java/org/apache/paimon/Snapshot.java
@@ -70,6 +70,7 @@ public class Snapshot {
 
     private static final String FIELD_VERSION = "version";
     private static final String FIELD_ID = "id";
+    private static final String FIELD_PARENT_ID = "parentId";
     private static final String FIELD_SCHEMA_ID = "schemaId";
     private static final String FIELD_BASE_MANIFEST_LIST = "baseManifestList";
     private static final String FIELD_DELTA_MANIFEST_LIST = "deltaManifestList";
@@ -93,6 +94,10 @@ public class Snapshot {
 
     @JsonProperty(FIELD_ID)
     private final long id;
+
+    @JsonProperty(FIELD_PARENT_ID)
+    @Nullable
+    private final Long parentId;
 
     @JsonProperty(FIELD_SCHEMA_ID)
     private final long schemaId;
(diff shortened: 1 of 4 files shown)


JingsongLi:
Thanks @hililiwei !  I have two comments:

### Compatibility

Maybe we can have a compatible way to finish this. For example, a snapshot has parent id, parent id can be `exists`, `none` or `null`, three types. If it is exists, find its parent. If it is `none`, it should be root snapshot. If it is `null`, it is a old version, its parent is snapshot id - 1.

### When should we finish this?

There is currently no benefit to just modifying this behavior, we need to wait until we need it, such as when we need to support git-like features, and we decide to do it.

hililiwei:
@JingsongLi Sorry for the late reply.

>Maybe we can have a compatible way to finish this. For example, a snapshot has parent id, parent id can be exists, none or null, three types. If it is exists, find its parent. If it is none, it should be root snapshot. If it is null, it is a old version, its parent is snapshot id - 1.

This solution is indeed relatively small modification, compatibility is also very good. I also thought of using a similar method at first. But I chose a more complex method, because I wanted to avoid using traditional file system management methods such as list/rename to manage metadata. 
I will try to update the PR with this way.

hililiwei:
>When should we finish this?
There is currently no benefit to just modifying this behavior, we need to wait until we need it, such as when we need to support git-like features, and we decide to do it.

Yes, when we try to introduce git-like features, we need to finish it. But more than that, the new way of explicitly specifying parent snapshot id helps us better use paimon on object storage.

hililiwei:
@JingsongLi I added the ParentId field and modified the related methods.

JingsongLi:
@hililiwei Very thanks for your investigation.
Now that the scene is not yet clear, we will consider this change in the future.

JingsongLi (commented):
Currently, there are really lots places to assume incremental snapshots (1, 2, 3 and ...).
And there is PIP https://cwiki.apache.org/confluence/display/PAIMON/PIP-9%3A+Support+Branch
I am closing this one.

PR #879, opened 2023-04-11
[core][bug] PartitionExpire should use proper commit identifier

yuzelin:
### Purpose

The `PartitionExpire`  use `Long.MAX_VALUE` as commit identifier, may cause stream commit filter cannot work in some cases.

### Tests

`PartitionExpireTest#testFilterCommittedAfterExpiring`

### API and Format 

No

### Documentation

No


JingsongLi (approved):
Looks good to me!

PR #880, opened 2023-04-12
[docs] Document the right dropping table behaviour

### LINKED ISSUE #870: [Bug] Drop table will cause file leak with inserting jobs still running
### Search before asking

### Paimon version

master: fb9babbb8676d598e817704db4993604b0d3a9d6

### Compute Engine

Flink 1.16.1

### Minimal reproduce step

1. create table
2. insert into table
3. drop table

This will leave some data files and manifest folder there, and an abnormal insert job failover always. 

### Anything else?

I think we should reject dropping operation on table which is being written at that time, or at least we could document that stop the inserting jobs before you drop a paimon table.

schnappi17:
fix #870

### DIFF
--- a/docs/content/how-to/creating-tables.md
+++ b/docs/content/how-to/creating-tables.md
@@ -67,6 +67,10 @@ CREATE TABLE MyTable (
 
 {{< /tabs >}}
 
+{{< hint info >}}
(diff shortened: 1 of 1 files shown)


JingsongLi (approved):
+1

PR #883, opened 2023-04-12
[CI] Solve the problem of starting multiple actions at the same time

liugddx:

### Purpose

 Solve the problem of starting multiple actions at the same time.

### Tests

### API and Format 

### Documentation

### DIFF
--- a/.github/workflows/check-licensing.yml
+++ b/.github/workflows/check-licensing.yml
@@ -17,6 +17,10 @@ name: Check licensing
 
 on: [push, pull_request]
 
+concurrency:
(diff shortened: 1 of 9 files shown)


liugddx:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobs

liugddx:
> https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobs

I operated it and found that it didn't take more time than before to start the jdk11 test, and github supports running jobs in parallel

liugddx:
image

It is work.

JingsongLi (approved):
Looks good to me!

PR #885, opened 2023-04-12
[core] Add cache to manifest files

JingsongLi:
### Purpose

In the initialization of write, the writer of the bucket needs to read all historical files. If there is a bottleneck here (For example, writing a large number of partitions simultaneously), we can use manifest cache to cache the read manifest data to accelerate initialization.

### Tests

`FileStoreTableTestBase.testManifestCache`

### Documentation

Yes


yuzelin (approved):
LGTM

Issue #830, opened 2023-04-04
[Feature]  Avoid using hive's own FileSystem to access the location of paimon table


wg1026688210:
### Search before asking

- [x] I searched in the [[[issues](https://github.com/apache/incubator-paimon/issues)](https://github.com/apache/incubator-paimon/issues)](https://github.com/apache/incubator-paimon/issues) and found nothing similar.

### Motivation

When using a permission authentication policy based on `StorageBasedAuthorizationProvider`, creating an external table of  paimon will cause Hive to create its own FileSystem to access the  location of paimon table in object storage , and it will not read the configuration of paimon for accessing the object storage .

### Solution

- Extend  `StorageBasedAuthorizationProvider` for checking whether is paimon table through the `STORED BY` field of table

- Support multi strategy for  permission authentication  policy of paimon if the  table is stored by paimon :

  - strategy 1 : do nothing for permission authentication checking  

  - strategy 2 :using  paimon's filesystem for checking permission authentication , the following is what we need to do

    - get the configuration of paimon for accessing the object storage 

      support the catalog property of paimon's extern table like Iceberg , so that we can access the config of object storage .  For example  : 

     ```sql
     CREATE EXTERNAL TABLE my_table
     STORED BY 'org.apache.paimon.hive.PaimonStorageHandler'
     LOCATION 's3://realtime/wgcn/warehouse/default.db/word_count';
     TBLPROPERTIES ('paimon.catalog'='a1');
     ```

     

    ​	and then we can get the catalog info from the config of HMS like:

    

    ```xml
        <property>
            <name>paimon.catalog.a1.access.key</name>
            <value>...</value>
        </property>
        <property>
            <name>paimon.catalog.a1.secret.key</name>
            <value>...</value>
        </property>
    ```

    ​	 or somewhere user redefine

    - use  the paimon's filesystem and implement the  permission authentication  Refering to the implementation of `StorageBasedAuthorizationProvider`

- add the document for guiding users config the  extend  of StorageBasedAuthorizationProvider at HMS

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

wg1026688210:
I'm willing to submit a PR

JingsongLi:
@wg1026688210 Thanks, assigned to u~

zhuangchong:
The relevant PR has been merged and I will close this issue.

PR #886, opened 2023-04-12
[hive] Support Read&Write of paimon table  which location in the tbproperties of hive table 

wg1026688210:
related #887, 
This pr provide the ability of Read&Write in Hive for the table which location setting in tbproperties


wg1026688210:
@JingsongLi  PTAL

docs/content/engines/hive.md
```diff
@@ -211,6 +211,18 @@ OK
 3	Paimon
 */
 
+```
+## Setting of location Location 
+In addition to the way setting location above, you can also place the location setting in TBProperties 
+to avoid Hive accessing Paimon's location through its own file system when creating tables.
+This method is effective in scenarios using Object storage,such as s3.
```
SteNicholas, line 218: For the object storage, why does this need to configure the `paimon_meta_location`? IMO, using `location` is enough to satisfy the object storage need, otherwise the filesystem and the metastore of Hive will lead to inconsistencies.

SteNicholas:
@wg1026688210, could you create an issue for this?

wg1026688210:
> @wg1026688210, could you create an issue for this?

related  #830

docs/content/engines/hive.md
wg1026688210: Hi~ @SteNicholas  thanks for your concern.
When using `location`, Hive will use its own FileSystem for authentication and to create the table path. This bypasses Paimon's fileio configuration, causing authentication errors, especially in object storage scenarios.

docs/content/engines/hive.md
```diff
@@ -211,6 +211,18 @@ OK
...
+
+```sql
+CREATE EXTERNAL TABLE external_test_table
+STORED BY 'org.apache.paimon.hive.PaimonStorageHandler'
+TBLPROPERTIES (
+ 'paimon_meta_location' ='s3://xxxxx/path/to/table/store/warehouse/default.db/test_table'
```
JingsongLi, line 224: After looking at the api again, I feel we can adjust the key.
Maybe `paimon_location`?

paimon-hive/paimon-hive-common/src/main/java/org/apache/paimon/hive/LocationKeyExtractor.java
```diff
@@ -18,11 +18,20 @@
...
  * declaring the name of the key in the parameters of the Hive metastore table, which indicates
  * where the Paimon table is stored.
  */
 public class LocationKeyExtractor {
     // special at the tbproperties with the name paimon_meta_location.
     public static final String TBPROPERTIES_LOCATION_KEY = "paimon_meta_location";
+    public static final String STORAGE_LOCATION_KEY = "location";
+
+    public static String getLocation(Properties properties) {
+        String storageLocation = properties.getProperty(STORAGE_LOCATION_KEY);
```
JingsongLi, line 33: Can we just use `hive_metastoreConstants.META_TABLE_LOCATION`?

paimon-hive/paimon-hive-connector-common/src/main/java/org/apache/paimon/hive/PaimonJobConf.java
```diff
@@ -91,4 +91,13 @@ public static Options extractCatalogConfig(Configuration hiveConf) {
         }
         return Options.fromMap(configMap);
     }
+
+    public static void setLocationFromProperties(
```
JingsongLi, line 95: Why not using `configureOutputJobProperties` or `configureInputJobProperties`?

paimon-hive/paimon-hive-connector-common/src/main/java/org/apache/paimon/hive/mapred/PaimonInputFormat.java
```diff
@@ -48,8 +49,18 @@ public InputSplit[] getSplits(JobConf jobConf, int numSplits) {
         FileStoreTable table = createFileStoreTable(jobConf);
         InnerTableScan scan = table.newScan();
         createPredicate(table.schema(), jobConf).ifPresent(scan::withFilter);
+
+        final String location;
+
+        String propertiesLocation = jobConf.get(LocationKeyExtractor.STORAGE_LOCATION_KEY);
```
JingsongLi, line 55: Why do this? `table.location()` should be correct.

wg1026688210:
@SteNicholas @JingsongLi cc

docs/content/engines/hive.md
```diff
@@ -228,6 +228,18 @@ OK
 
 ```
 
+## Setting of Location
+In addition to the way setting location above, you can also place the location setting in TBProperties
+to avoid Hive accessing Paimon's location through its own file system when creating tables.
+This method is effective in scenarios using Object storage,such as s3.
+
+```sql
+CREATE EXTERNAL TABLE external_test_table
```
JingsongLi, line 237: You can just document this in `Hive SQL: access Paimon Tables by External Table`.

paimon-hive/paimon-hive-connector-common/src/main/java/org/apache/paimon/hive/mapred/PaimonInputFormat.java
```diff
@@ -48,8 +49,9 @@ public InputSplit[] getSplits(JobConf jobConf, int numSplits) {
         FileStoreTable table = createFileStoreTable(jobConf);
         InnerTableScan scan = table.newScan();
         createPredicate(table.schema(), jobConf).ifPresent(scan::withFilter);
+        String location = jobConf.get(hive_metastoreConstants.META_TABLE_LOCATION);
         return scan.plan().splits().stream()
-                .map(split -> new PaimonInputSplit(table.location().toString(), (DataSplit) split))
+                .map(split -> new PaimonInputSplit(location, (DataSplit) split))
```
JingsongLi, line 54: add comments here, why we need to use `hive_metastoreConstants.META_TABLE_LOCATION`.

JingsongLi (commented):
Left two minor comments.

JingsongLi (approved):
Looks good to me!
Thanks @wg1026688210 for your contribution!

PR #887, opened 2023-04-12
[hive] Avoid using hive own FileSystem to access the location of paimon table when using HiveCatalog

wg1026688210:
Currently, HiveCatalog specify the `warehouse` and the `location` of database/table  to HMS when creating database/table
,causing that HMS access the `location`  by filesystem of hive.
for resolve it, the pr make some changes 

-  do not specify the `warehouse`  config  and the `location ` of database and table 
-  and handoff the file access to fileio


wg1026688210:
a draft for discussions

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
```diff
@@ -150,12 +153,16 @@ public boolean databaseExists(String databaseName) {
     public void createDatabase(String name, boolean ignoreIfExists)
             throws DatabaseAlreadyExistException {
         try {
+            Path databasePath = databasePath(name);
```
JingsongLi, line 156: Move this after `client.createDatabase`?

JingsongLi (changes requested):
Thanks @wg1026688210 for the contribution.

I think this is a workable mode, but it may not be the default mode, the same location as Hive can make the experience much better and less misunderstanding.

We need an OPTION to switch it on and off.

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalogOptions.java
```diff
@@ -40,5 +40,12 @@ public final class HiveCatalogOptions {
                     .withDescription(
                             "File directory of the core-site.xml、hdfs-site.xml、yarn-site.xml、mapred-site.xml. Currently, only local file system paths are supported.");
 
+    public static final ConfigOption<Boolean> LOCATION_IN_PROPERTIES =
+            ConfigOptions.key("location-in-properties")
+                    .booleanType()
+                    .defaultValue(false)
+                    .withDescription(
+                            "setting the location in properties of hive table/database, you can try configing 'location-in-properties'= 'true'  if you don't want that accessing the location of hive table/database by the filesystem of hive,when using object store as the warehouse of hive catalog.");
```
JingsongLi, line 48: We should avoid single line that are too long

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalogFactory.java
```diff
@@ -69,7 +69,7 @@ public Catalog create(FileIO fileIO, Path warehouse, CatalogContext context) {
         // always using user-set parameters overwrite hive-site.xml parameters
         context.options().toMap().forEach(hiveConf::set);
         hiveConf.set(HiveConf.ConfVars.METASTOREURIS.varname, uri);
-        hiveConf.set(HiveConf.ConfVars.METASTOREWAREHOUSE.varname, warehouse.toUri().toString());
+        hiveConf.set(CatalogOptions.WAREHOUSE.key(), warehouse.toUri().toString());
```
JingsongLi, line 72: Can we delete this line?
We can just pass warehouse to `HiveCatalog`.
And just pass warehouse to `LoctionHelper`.

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
```diff
@@ -97,18 +100,31 @@ public class HiveCatalog extends AbstractCatalog {
...
+                hiveConf.getBoolean(
+                        LOCATION_IN_PROPERTIES.key(), LOCATION_IN_PROPERTIES.defaultValue());
+        if (needLocationInProperties) {
+            locationHelper = new TBPropertiesLocationHelper(hiveConf);
+        } else {
+            // set the warehouse location to the hiveConf
+            String warehouse = hiveConf.get(CatalogOptions.WAREHOUSE.key());
+            hiveConf.set(HiveConf.ConfVars.METASTOREWAREHOUSE.varname, warehouse);
+
+            locationHelper = new StorageLocationHelper(hiveConf);
```
JingsongLi, line 125: We can put these two lines to the constructor of `StorageLocationHelper`.
```
String warehouse = hiveConf.get(CatalogOptions.WAREHOUSE.key());
hiveConf.set(HiveConf.ConfVars.METASTOREWAREHOUSE.varname, warehouse);
```

paimon-common/src/main/java/org/apache/paimon/utils/LocationKeyExtractor.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+package org.apache.paimon.utils;
+
+import java.util.Properties;
+
+/**
+ * declaring the name of the key in the parameters of the Hive metastore table, which indicates
+ * where the Paimon table is stored.
+ */
+public class LocationKeyExtractor {
```
JingsongLi, line 26: Can we merge this class into `TBPropertiesLocationHelper`?

JingsongLi (changes requested):
Thanks @wg1026688210 ! It looks better!

Can we also update documentation for hive catalog? In `create-catalog` page.

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalogFactory.java
wg1026688210: It's better for passing warehouse to HiveCatalog for a better readable indeed.

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
wg1026688210: if we pass warehouse to `HiveCatalog` at the above conversation , we can specify the warehouse directly .So that  we don't need pass `hiveConf` for  specifying the warehouse at the both `LocationHelper` separately . Shall we retain these two lines  outside the `LocationHelper`.

paimon-common/src/main/java/org/apache/paimon/utils/LocationKeyExtractor.java
wg1026688210: Another pr #886  will use this  class  for  reading  hive extern table. Shall we  migrate it to `paimon-hive-common` for a smaller scope.

wg1026688210:
> Thanks @wg1026688210 ! It looks better!
> 
> Can we also update documentation for hive catalog? In `create-catalog` page.

done

docs/content/how-to/creating-catalogs.md
```diff
@@ -127,6 +132,11 @@ After `spark-sql` is started, you can switch to the `default` database of the `p
 USE paimon.default;
 ```
 
+If you are using a object storage , and you don't want that the location of paimon table/database is accessed by the filesystem of hive,
```
JingsongLi, line 135: Maybe we can just copy these sentence to after `tab`? So both Flink and Spark can contain it.

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/TBPropertiesLocationHelper.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @Override
+    public void specifyTableLocation(Table table, String location) {
+        table.getParameters().put(LocationKeyExtractor.TBPROPERTIES_LOCATION_KEY, location);
+    }
+
+    @Override
+    public void specifyDatabaseLocation(Path path, Database database) {
+        HashMap<String, String> properties = new HashMap<>();
```
JingsongLi, line 56: check if `database.getParameters != null`, `properties,putAll(database.getParameters)`?

paimon-hive/paimon-hive-common/src/main/java/org/apache/paimon/hive/LocationKeyExtractor.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.Properties;
+
+/**
+ * declaring the name of the key in the parameters of the Hive metastore table, which indicates
+ * where the Paimon table is stored.
+ */
+public class LocationKeyExtractor {
+    // special at the tbproperties with the name paimon_meta_location.
+    public static final String TBPROPERTIES_LOCATION_KEY = "paimon_meta_location";
+    public static final String STORAGE_LOCATION_KEY = "location";
```
JingsongLi, line 30: Can we remove this `STORAGE_LOCATION_KEY` and `getLocation`?
They should be introduced in next PR instead of this PR.

JingsongLi (approved):
Looks good to me!

Zouxxyy:
@wg1026688210 hi, can you explain what is `causing that HMS access the location by filesystem of hive`, I fell very confued about the role of `location-in-properties`

PR #895, opened 2023-04-13
[flink][bug] Partition and bucket should be reset when a new record is set in CdcRecordKeyAndBucketExtractor

tsreaper:
### Purpose

In `CdcRecordKeyAndBucketExtractor#setRecord` I forget to set `partition`, `bucketKey` and `bucket` to null. Thus a `CdcRecordKeyAndBucketExtractor` instance will only produce 1 constant type of `partition` and `bucket`, which is obviously incorrect.

This PR fixes the issue.

### Tests

* CdcRecordKeyAndBucketExtractorTest
* FlinkCdcSinkITCase

### API and Format 

N/A

### Documentation

N/A


JingsongLi (commented):
+1

PR #896, opened 2023-04-13
[flink] Introduce process functions and sinks for CDC database syncing

tsreaper:
### Purpose

This PR introduces `CdcMultiTableParsingProcessFunction` and `FlinkCdcSyncDatabaseSinkBuilder` which creates a separate Paimon sink for each table in the source database. With the new sink, we can support CDC database syncing.

The current implementation is not very efficient in resource saving. We may refine this implementation by merging all sinks into one instance in the future.

### Tests

* FlinkCdcSyncDatabaseSinkITCase

### API and Format 

N/A

### Documentation

N/A

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/cdc/FlinkCdcSyncDatabaseSinkBuilder.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/cdc/FlinkCdcSyncDatabaseSinkBuilder.java
@@ -1,0 +1,122 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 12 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/cdc/FlinkCdcSyncDatabaseSinkBuilder.java
```diff
@@ -0,0 +1,122 @@
...
+        for (FileStoreTable table : tables) {
+            DataStream<Void> schemaChangeProcessFunction =
+                    SingleOutputStreamOperatorUtils.getSideOutput(
+                                    parsed,
+                                    CdcMultiTableParsingProcessFunction.createSchemaChangeOutputTag(
+                                            table.name()))
+                            .process(
+                                    new SchemaChangeProcessFunction(
+                                            new SchemaManager(table.fileIO(), table.location())));
+            schemaChangeProcessFunction.getTransformation().setParallelism(1);
```
JingsongLi, line 102: `setMaxParallelism` too. Modify `FlinkCdcSinkBuilder` too.

JingsongLi (approved):
+1

PR #898, opened 2023-04-13
[core] Introduce TableCommit.abort

JingsongLi:
### Purpose

Introduce TableCommit.abort, somewhere needs this method to abort job commit, like Hive write.

### Tests

`FileStoreTableTestBase.testAbort`

### API and Format 

Add API

### Documentation

Not in normal path, don't need to document this.


yuzelin (approved):
LGTM

paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommit.java
```diff
@@ -72,4 +73,7 @@ void overwrite(
      * @param partitions A list of partition {@link Map}s. NOTE: cannot be empty!
      */
     void dropPartitions(List<Map<String, String>> partitions, long commitIdentifier);
+
+    /** Abort an unsuccessful commit. The data files will be deleted. */
+    void abort(List<CommitMessage> commitMessages);
```
SteNicholas, line 78: Could the param use `ManifestCommittable committable`?

JingsongLi: `ManifestCommittable` need an identifier. But here we don't need that.

SteNicholas (approved):
LGTM.

PR #914, opened 2023-04-14
[core] Introduce file.compression

### LINKED ISSUE #889: [Feature] Introduce file.compression
### Search before asking

### Motivation

Add a generic configuration: `file.compression`, which is used by default, replace orc.compress and parquet.compression. Then file.compression.per.level can override the compression of the specified level.

### Solution

_No response_

### Anything else?

_No response_

Tartarus0zm:
### Purpose

fix #889 

### Tests

### API and Format 

Format

### Documentation

introduce a new feature

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/format/FormatWriterFactory.java
+++ b/paimon-common/src/main/java/org/apache/paimon/format/FormatWriterFactory.java
@@ -20,8 +20,6 @@ package org.apache.paimon.format;
 
 import org.apache.paimon.fs.PositionOutputStream;
 
-import javax.annotation.Nullable;
-
 import java.io.IOException;
 
 /** A factory to create {@link FormatWriter} for file. */
@@ -35,9 +33,9 @@ public interface FormatWriterFactory {
      * @throws IOException Thrown if the writer cannot be opened, or if the output stream throws an
      *     exception.
      */
-    FormatWriter create(PositionOutputStream out, @Nullable String compression) throws IOException;
+    FormatWriter create(PositionOutputStream out, String compression) throws IOException;
 
     default FormatWriter create(PositionOutputStream out) throws IOException {
-        return create(out, null);
+        return create(out, "LZ4");
     }
 }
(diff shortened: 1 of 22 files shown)


paimon-common/src/main/java/org/apache/paimon/format/FormatWriterFactory.java
```diff
@@ -35,9 +33,9 @@ public interface FormatWriterFactory {
      * @throws IOException Thrown if the writer cannot be opened, or if the output stream throws an
      *     exception.
      */
-    FormatWriter create(PositionOutputStream out, @Nullable String compression) throws IOException;
+    FormatWriter create(PositionOutputStream out, String compression) throws IOException;
 
     default FormatWriter create(PositionOutputStream out) throws IOException {
```
JingsongLi, line 38: Can we remove this method?
I see that only manifestList uses this method. (excluding testing)

paimon-core/src/main/java/org/apache/paimon/manifest/ManifestFile.java
```diff
@@ -82,7 +85,9 @@ public long suggestedFileSize() {
     public List<ManifestFileMeta> write(List<ManifestEntry> entries) {
         RollingFileWriter<ManifestEntry, ManifestFileMeta> writer =
                 new RollingFileWriter<>(
-                        () -> new ManifestEntryWriter(writerFactory, pathFactory.newPath()),
+                        () ->
+                                new ManifestEntryWriter(
+                                        writerFactory, pathFactory.newPath(), fileCompression),
```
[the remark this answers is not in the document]
Tartarus0zm: use `CoreOptions.FILE_COMPRESSION.defaultValue()` instead of hardcord

paimon-common/src/main/java/org/apache/paimon/format/FormatWriterFactory.java
Tartarus0zm: I will remove this method, then use `CoreOptions.FILE_COMPRESSION.defaultValue()` in manifestList

Tartarus0zm:
@JingsongLi  hello, I have addressed your comment, please take a look again.

JingsongLi (approved):
Thanks @Tartarus0zm Looks good to me!

PR #917, opened 2023-04-15
[core & flink & hive] use new ReadBuilder and WriteBuilder API in modules

### LINKED ISSUE #715: [Improve] Use new ReadBuilder and WriteBuilder API in modules
### Search before asking

### Motivation

Now paimon has refactored the read/write api, so we should retire the old api interface in codes.

There are some examples:

image

image

### Solution

_No response_

### Anything else?

_No response_

TyrantLucifer:
### Purpose

use new ReadBuilder and WriteBuilder API in [core & hive & flink] modules

close #715

### Tests

already exists

### API and Format 

yes

### Documentation

no

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/system/AuditLogTable.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/system/AuditLogTable.java
@@ -38,6 +38,7 @@ import org.apache.paimon.table.source.DataSplit;
 import org.apache.paimon.table.source.InnerStreamTableScan;
 import org.apache.paimon.table.source.InnerTableRead;
 import org.apache.paimon.table.source.InnerTableScan;
+import org.apache.paimon.table.source.ReadBuilder;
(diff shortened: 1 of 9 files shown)


TyrantLucifer:
Hi, @JingsongLi PTAL

paimon-core/src/main/java/org/apache/paimon/table/system/AuditLogTable.java
```diff
@@ -84,8 +85,11 @@ public class AuditLogTable implements DataTable, ReadonlyTable {
 
     private final FileStoreTable dataTable;
 
+    private final ReadBuilder readBuilder;
+
     public AuditLogTable(FileStoreTable dataTable) {
         this.dataTable = dataTable;
+        this.readBuilder = dataTable.newReadBuilder();
```
JingsongLi, line 92: We don't need to modify these. They are inner table scans.

Issue #752, opened 2023-03-29
[Feature] Introduce Hive writer


JingsongLi:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/incubator-paimon/issues) and found nothing similar.

### Motivation

Introduce Hive writer

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

TyrantLucifer:
I'm interested in it. Could you please assign it to me?

JingsongLi:
@Alibaba-HZY Assigned to u, maybe we can have a write-only version for hive writer.

PR #918, opened 2023-04-15
[hive] Introduce Hive writer (#752)

Alibaba-HZY:
Purpose
Introduce Hive writer,user can use hive sql to write data to the paimon table, for example :"insert into paimon_table select * from hive_table".Supports complex nested type writes such as: array<row<id bigint>>,map<string,array<decimal(5,2)>>,row<id bigint,name string>.Also fixed an issue where paimonRead could not read complex nested types.
Note that currently only mr is supported, tez is not supported, and only  support single table writer, insert overwrite is also not supported.
https://github.com/apache/incubator-paimon/issues/752

Tests
TestPaimonStorageHandlerUseMR

API and Format
Documentation
Create a PaimonRecordWriter for each task in mapreduce, which contains a BatchTableWrite for writing data, It also contains a shared map that holds the mapping of task and PaimonRecordWriter, in order to obtain the precommit and commit data from BatchTableWrite in commitTask and commitJob,  or abortCommit if the Job failed.


JingsongLi:
Wow~ Thanks @Alibaba-HZY !

paimon-hive/paimon-hive-connector-common/src/main/java/org/apache/paimon/hive/mapred/PaimonOutputFormat.java
```diff
@@ -18,27 +18,63 @@
...
+            Properties properties,
+            Progressable progressable)
+            throws IOException {
+        return writer(jobConf);
+    }
+
+    private static PaimonRecordWriter writer(JobConf jobConf) {
+        TaskAttemptID taskAttemptID = TaskAttemptID.forName(jobConf.get(TASK_ATTEMPT_ID_KEY));
+        FileStoreTable table = createFileStoreTable(jobConf);
+        BatchWriteBuilder batchWriteBuilder = table.newBatchWriteBuilder();
```
JingsongLi, line 75: Here should we force write-only=true?

Alibaba-HZY: yes i have fixed

paimon-hive/paimon-hive-connector-common/src/test/java/org/apache/paimon/hive/TestPaimonStorageHandlerUseMR.java
```diff
@@ -0,0 +1,1033 @@
...
+import java.util.concurrent.ThreadLocalRandom;
+
+import static org.apache.paimon.hive.FileStoreTestUtils.DATABASE_NAME;
+import static org.apache.paimon.hive.FileStoreTestUtils.TABLE_NAME;
+import static org.apache.paimon.hive.RandomGenericRowDataGenerator.randomBigDecimal;
+import static org.assertj.core.api.Assertions.assertThat;
+
+/** IT cases for {@link PaimonStorageHandler} and {@link PaimonOutputFormat}. */
+@RunWith(PaimonEmbeddedHiveRunner.class)
+public class TestPaimonStorageHandlerUseMR {
```
JingsongLi, line 70: `HiveReadWriteITCase`?

paimon-hive/paimon-hive-connector-common/src/main/java/org/apache/paimon/hive/utils/HiveUtils.java
```diff
@@ -0,0 +1,65 @@
...
+/** Utils for create {@link FileStoreTable} and {@link Predicate}. */
+public class HiveUtils {
+
+    private static final Logger LOG = LoggerFactory.getLogger(HiveUtils.class);
+
+    public static FileStoreTable createFileStoreTable(JobConf jobConf) {
+        PaimonJobConf wrapper = new PaimonJobConf(jobConf);
+        Options options = PaimonJobConf.extractCatalogConfig(jobConf);
+        options.set(CoreOptions.PATH, wrapper.getLocation());
+        options.set(CoreOptions.WRITE_ONLY, Boolean.TRUE);
```
JingsongLi, line 48: add comments: why need to add this.

paimon-hive/paimon-hive-connector-common/src/main/java/org/apache/paimon/hive/mapred/PaimonOutputCommitter.java
```diff
@@ -0,0 +1,284 @@
...
+        long startTime = System.currentTimeMillis();
+        LOG.info("CommitJob {} has started", jobContext.getJobID());
+        FileStoreTable table = createFileStoreTable(jobConf);
+
+        if (table != null) {
+            BatchWriteBuilder batchWriteBuilder = table.newBatchWriteBuilder();
+            List<List<CommitMessage>> commitMessagesList =
+                    getAllPreCommitMessage(table.location().getPath(), jobContext, table.fileIO());
+            for (List<CommitMessage> commitTables : commitMessagesList) {
+                try (BatchTableCommit batchTableCommit = batchWriteBuilder.newCommit()) {
```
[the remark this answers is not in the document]
Alibaba-HZY: yes We only need one commit

paimon-hive/paimon-hive-connector-common/src/main/java/org/apache/paimon/hive/utils/HiveUtils.java
JingsongLi: my bad~
I mean, if you add a test, please reply here which test was added

paimon-hive/paimon-hive-connector-common/src/test/java/org/apache/paimon/hive/TestPaimonStorageHandlerUseMR.java
JingsongLi: OK

paimon-hive/paimon-hive-connector-common/src/main/java/org/apache/paimon/hive/utils/HiveUtils.java
Alibaba-HZY: > my bad~ I mean, if you add a test, please reply here which test was added

ok i removed this option in here,i use FileStoreTable.copy to add write-only

paimon-hive/paimon-hive-connector-common/src/main/java/org/apache/paimon/hive/DataTypeDefaultVisitor.java
```diff
@@ -0,0 +1,117 @@
...
+import org.apache.paimon.types.MapType;
+import org.apache.paimon.types.RowType;
+
+import java.util.List;
+
+/**
+ * DataTypeDefaultVisitor to visitor the sourceInspector and writerInspector to get the
+ * FieldDeserializer.
+ */
+public abstract class DataTypeDefaultVisitor<P, R> {
```
JingsongLi, line 35: I mean using `org.apache.paimon.types.DataTypeDefaultVisitor.`

paimon-hive/paimon-hive-connector-common/src/main/java/org/apache/paimon/hive/PaimonSerDe.java
```diff
@@ -39,11 +42,18 @@
 public class PaimonSerDe extends AbstractSerDe {
 
     private PaimonInternalRowObjectInspector inspector;
+    private HiveSchema tableSchema;
+
+    private RowDataContainer rowData = new RowDataContainer();
+
+    private Map<ObjectInspector, HiveDeserializer> deserializers =
```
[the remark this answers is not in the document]
JingsongLi: Why not `final`?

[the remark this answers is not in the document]
Alibaba-HZY: ok my bad ,deserializers is also can be final

paimon-hive/paimon-hive-connector-common/src/test/java/org/apache/paimon/hive/HiveWriteITCase.java
```diff
@@ -0,0 +1,1033 @@
...
+                                    DataTypes.INT(),
+                                    DataTypes.BIGINT(),
+                                    DataTypes.STRING()
+                                },
+                                new String[] {"pt", "a", "b", "c"}),
+                        Collections.singletonList("pt"),
+                        emptyData,
+                        "hive_test_table_output");
+
+        hiveShell.execute(
```
JingsongLi, line 206: We can add a case: insert 5 times, and see there is no compaction. (See no compact snapshots)
To validate `write-only` option.

paimon-hive/paimon-hive-connector-common/src/main/java/org/apache/paimon/hive/mapred/PaimonOutputFormat.java
```diff
@@ -18,27 +18,70 @@
...
+            Properties properties,
+            Progressable progressable)
+            throws IOException {
+        return writer(jobConf);
+    }
+
+    private static PaimonRecordWriter writer(JobConf jobConf) {
+        TaskAttemptID taskAttemptID = TaskAttemptID.forName(jobConf.get(TASK_ATTEMPT_ID_KEY));
+        FileStoreTable table = createFileStoreTable(jobConf);
+        Map<String, String> newOptions = Maps.newHashMapWithExpectedSize(1);
```
JingsongLi, line 78: Collections.singleMap

paimon-hive/paimon-hive-connector-common/src/main/java/org/apache/paimon/hive/DataTypeDefaultVisitor.java
Alibaba-HZY: DataTypeDefaultVisitor or DataTypeVisitor may not apply in this case,they are have only one parameter,in hive we need tree param to deserialize value

JingsongLi: Maybe return a custom class to contains tree param?

paimon-hive/paimon-hive-connector-common/src/test/java/org/apache/paimon/hive/HiveWriteITCase.java
Alibaba-HZY: it may insert 50 times, because eacge file is too small to write 5 times without a compaction

JingsongLi: If records in a bucket, 5 times are enough.

Alibaba-HZY: I have added a test for write-only option

Issue #921, opened 2023-04-17
[Bug] OrcWriterFactoryTest.testNotOverrideInMemoryManager throws NullPointerException in unit test


liming30:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/incubator-paimon/issues) and found nothing similar.

### Paimon version

0.4-snapshot

### Compute Engine

flink-1.16

### Minimal reproduce step

mvn test -pl paimon-format -Dtest=org.apache.paimon.format.orc.OrcWriterFactoryTest

### What doesn't meet your expectations?

We should assign a non-null value to the `writerProperties` of the `OrcWriterFactory` and check that it is not null.

### Anything else?

![image](https://user-images.githubusercontent.com/69444450/232415697-9a35afd7-5eec-4710-b272-53fa580e8959.png)

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

PR #922, opened 2023-04-17
[format] Fix the issue of OrcWriterFactoryTest#testNotOverrideInMemoryManager NullPointerException in unit test.

liming30:
### Purpose

[format] Fix the issue of OrcWriterFactoryTest#testNotOverrideInMemoryManager NullPointerException in unit test.  #921 

### Tests

No.

### API and Format 

No.

### Documentation

No.


liming30:
> lz4 for parquet requires native libraries. I would suggest changing lz4 compressions in tests to snappy or something else.

@tsreaper I also initiated a new issue about lz4（https://github.com/apache/incubator-paimon/issues/923）.  Should I fix both issues in this one issue?

liming30:
@tsreaper Is this mr worthy of being merged? We're using `writerProperties` [here](https://github.com/apache/incubator-paimon/blob/master/paimon-format/src/main/java/org/apache/paimon/format/orc/OrcWriterFactory.java#L102), but we never check if `writerProperties` is `null`, which makes it easy to make a mistake.

JingsongLi (approved):
+1

PR #925, opened 2023-04-17
[Improve] Use FlinkCatalogFactory to create FlinkCatalog

liugddx:

### Purpose

### Tests

### API and Format 

### Documentation

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/ActionBase.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/ActionBase.java
@@ -81,9 +81,8 @@ public abstract class ActionBase implements Action {
 
     ActionBase(String warehouse, String databaseName, String tableName, Options options) {
         identifier = new Identifier(databaseName, tableName);
-        CatalogContext catalogContext =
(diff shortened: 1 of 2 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/ActionBase.java
```diff
@@ -80,11 +79,15 @@ public abstract class ActionBase implements Action {
...
     ActionBase(String warehouse, String databaseName, String tableName, Options options) {
         identifier = new Identifier(databaseName, tableName);
-        catalog =
-                CatalogFactory.createCatalog(
-                        CatalogContext.create(
-                                new Options().set(CatalogOptions.WAREHOUSE, warehouse)));
-        flinkCatalog = new FlinkCatalog(catalog, catalogName, DEFAULT_DATABASE);
+        CatalogContext catalogContext =
+                CatalogContext.create(new Options().set(CatalogOptions.WAREHOUSE, warehouse));
+        catalog = CatalogFactory.createCatalog(catalogContext);
```
JingsongLi, line 84: catalog can assign from `FlinkCatalog`

JingsongLi:
The version is looks same to previous version.
I think there is no improvement here.

PR #926, opened 2023-04-17
[Improve] remove checkstyle and scalastyle check

liugddx:

### Purpose

We already have spotless to format the code, so remove the check style plugin.

### Tests

### API and Format 

### Documentation


zhuangchong (changes requested):
checkstyle should not be removed. Currently, the project uses spotless format and checkstyle together.
For example: `import *`, `shade use`, etc. in checkstyle

liugddx:
> checkstyle should not be removed. Currently, the project uses spotless format and checkstyle together. For example: `import *`, `shade use`, etc. in checkstyle

https://www.libhunt.com/compare-spotless-vs-checkstyle?ref=compare

I think spotless and checkstyle are duplicate checks, we should not use them at the same time, If spotless doesn't support these checks, what we should do is make it support these checks.

zhuangchong:
> > checkstyle should not be removed. Currently, the project uses spotless format and checkstyle together. For example: `import *`, `shade use`, etc. in checkstyle
> 
> https://www.libhunt.com/compare-spotless-vs-checkstyle?ref=compare
> 
> I think spotless and checkstyle are duplicate checks, we should not use them at the same time, If spotless doesn't support these checks, what we should do is make it support these checks.

When removing checkstyle, can you add the project checkstyle special configuration item to the spotless format?

liugddx:
> > > checkstyle should not be removed. Currently, the project uses spotless format and checkstyle together. For example: `import *`, `shade use`, etc. in checkstyle
> > 
> > 
> > https://www.libhunt.com/compare-spotless-vs-checkstyle?ref=compare
> > I think spotless and checkstyle are duplicate checks, we should not use them at the same time, If spotless doesn't support these checks, what we should do is make it support these checks.
> 
> When removing checkstyle, can you add the project checkstyle special configuration item to the spotless format?

I would love to try.  Thank you so much for your suggestion.

liugddx:
> > > > checkstyle should not be removed. Currently, the project uses spotless format and checkstyle together. For example: `import *`, `shade use`, etc. in checkstyle
> > > 
> > > 
> > > https://www.libhunt.com/compare-spotless-vs-checkstyle?ref=compare
> > > I think spotless and checkstyle are duplicate checks, we should not use them at the same time, If spotless doesn't support these checks, what we should do is make it support these checks.
> > 
> > 
> > When removing checkstyle, can you add the project checkstyle special configuration item to the spotless format?
> 
> I would love to try. Thank you so much for your suggestion.

I added import wildcard rules, PTAL thanks  @zhuangchong

pom.xml
```diff
@@ -798,7 +763,11 @@ under the License.
                             <importOrder>
                                 <order>org.apache.paimon,org.apache.paimon.shaded,,javax,java,scala,\#</order>
                             </importOrder>
-
+                            <replaceRegex>
+                                <name>Remove wildcard imports</name>
+                                <searchRegex>import\s+(static)*\s*[^\*\s]+\*;(\r\n|\r|\n)</searchRegex>
+                                <replacement>$1</replacement>
+                            </replaceRegex>
```
zhuangchong, line 770: This is not a very good solution, the part is to remove the import * if I understand correctly. If this is the case, after the contributor executes `mvn spotless:apply`, he submits the PR, and only when he executes the CI will he find that some classes are not imported.

zhuangchong (commented):
Let's see what everyone thinks about this change.

pom.xml
liugddx: I'm afraid there is no good solution yet.
https://github.com/diffplug/spotless/issues/649.  Perhaps it is more appropriate now, thanks for help and i will close this pr . @zhuangchong

PR #927, opened 2023-04-17
[pom] Comment the repositories tag in pom.xml to speed up maven build

tsreaper:
### Purpose

After #884, project building time increases from 5 minutes to 15 minutes in a github action. This is because downloading dependencies from apache repositories is very slow.

This PR comments the repositories tag in pom.xml to speed up maven build. This PR also adds a notice in the README file for users who do not have a snapshot repositories in their maven settings.

### Tests

N/A

### API and Format 

N/A

### Documentation

N/A


JingsongLi (approved):
+1

PR #928, opened 2023-04-17
[doc] set taskmanager.numberOfTaskSlots: 10

liugddx:

### Purpose

flink test optimization.

`taskmanager.numberOfTaskSlots: 10`

### Tests

### API and Format 

### Documentation


JingsongLi:
Hi @liugddx , thanks for the contribution, Where is the enhancement?

liugddx:
> Hi @liugddx , thanks for the contribution, Where is the enhancement?

I ran the test case, but the task could not run due to insufficient slots

JingsongLi:
> > Hi @liugddx , thanks for the contribution, Where is the enhancement?
> 
> I ran the test case, but the task could not run due to insufficient slots

I think our tests work fine now, we don't need to increase slots.

PR #929, opened 2023-04-17
[core] Filter on Manifest Cache

### LINKED ISSUE #913: [Bug] Performance issues with appendOnly table with a large number of partitions
### Search before asking

### Paimon version

0.4

### Compute Engine

flink

### Minimal reproduce step

I use Paimon to build an AppendOnly table to complete the writing of user logs to hdfs. 
This table has a large number of fields, about 300+,  and it defines two level partitions, the first for `date` and the second for `event`. There are about 400 events, that is, we can see 400+ directories under the event partition directory.

When I first started the job, the writing went smoothly without any end-to-end delay, but when the job ran for about 30 minutes, the back pressure began to be generated, and it could not be eliminated at all. I suspect that checkpointing took too long. The long-term impact on the decline in data throughput, my checkpoint interval is 1 minute, and each checkpointing takes almost tens of seconds, but when backpressure occurs, the time-consuming will increase, and the end-to-end delay will begin to increase. Normal consumption levels cannot be restored. 

During this period, I tried to increase the sink parallelism and memory resources, but it didn't work. I also increased the configuration of `manifest.cache-size`, but it didn't work either.

How can I go about solving the above performance issues?

Writing to appendOnly table with a large number of partitions has performance issues

### Anything else?

_No response_

JingsongLi:
Fix #913

### Purpose

- Use Filter to speed up cache access and filter directly when unconverted.
- Also, advance filtering when loading cache to improve cache utilization


wxplovecc:
+1 for fix

tsreaper (approved):

Issue #943, opened 2023-04-19
[Bug] MysqlCdcAction can not convert decimal type correct 


zhangjun0x01:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/incubator-paimon/issues) and found nothing similar.

### Paimon version

0.4

### Compute Engine

flink

### Minimal reproduce step

1. start a flink cdc action , sync the mysql data to paimon.
   the mysql table has a column with `decimal` type.
 ```
mysql> SELECT * FROM products;
+----+----------+-------+
| ID | CREATEBY | a     |
+----+----------+-------+
| 1  | aa       | 1.230 |
| 2  | bb       | 1.230 |
+----+----------+-------+
2 rows in set (0.00 sec)

mysql> desc products;
+----------+---------------+------+-----+---------+-------+
| Field    | Type          | Null | Key | Default | Extra |
+----------+---------------+------+-----+---------+-------+
| ID       | varchar(100)  | NO   | PRI | NULL    |       |
| CREATEBY | varchar(100)  | YES  |     | NULL    |       |
| a        | decimal(10,3) | YES  |     | NULL    |       |
+----------+---------------+------+-----+---------+-------+
3 rows in set (0.01 sec)

``` 

2. stop the flink job
3. start the flink job again, the paimon will throw an exception.

### What doesn't meet your expectations?

java.lang.IllegalArgumentException: Paimon schema and MySQL schema are not compatible.
Paimon fields are: [`ID` VARCHAR(100) NOT NULL, `CREATEBY` VARCHAR(100), `a` DECIMAL(10, 3)].
MySQL fields are: {ID=VARCHAR(100), CREATEBY=VARCHAR(100), a=DECIMAL(10, 3)}

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

JingsongLi:
longtext also not work?

zhangjun0x01:
> longtext also not work?

I'll check the other types again

zhangjun0x01:
> longtext also not work?

I check it ,found that `longtext` throws an exception `Don't support MySQL type` ,I add the `longtext` and `longblob` in `MySqlTypeUtils#toDataType`.

`time` datatype is also don't support, but I think it may be complex, It will involve issues of timezone and precision，I will fix it in another pr。

I also add the date type to check compatibility.

JingsongLi:
Thanks @zhangjun0x01

PR #944, opened 2023-04-19
[flink] check compatibility for some data type

zhangjun0x01:
fix https://github.com/apache/incubator-paimon/issues/943

### Purpose

* check compatibility for some data type : decimal , date ，timestamp, longtext, longblob *

### Tests

### API and Format 

### Documentation

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/cdc/mysql/MySqlTypeUtils.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/cdc/mysql/MySqlTypeUtils.java
@@ -184,12 +184,14 @@ public class MySqlTypeUtils {
             case VARCHAR:
                 return DataTypes.VARCHAR(Preconditions.checkNotNull(length));
             case TEXT:
+            case LONGTEXT:
                 return DataTypes.STRING();
             case BINARY:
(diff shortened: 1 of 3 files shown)


zhangjun0x01:
fix #943

zhangjun0x01:
image

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/cdc/UpdatedDataFieldsProcessFunction.java
```diff
@@ -200,6 +213,29 @@ public static ConvertAction canConvert(DataType oldType, DataType newType) {
             return oldIdx <= newIdx ? ConvertAction.CONVERT : ConvertAction.IGNORE;
         }
 
+        oldIdx = DECIMAL_TYPES.indexOf(oldType.getTypeRoot());
```
JingsongLi, line 216: We don't need to support the decimal modification, we just need to determine if it is equal here.
```
if (oldType.equalsIgnoreNullable(newType)) {
    return ConvertAction.CONVERT;
}
```

JingsongLi (changes requested):
we should add tests in `paimon-flink-common/src/test/resources/mysql/setup.sql` to test longtext and longblob

JingsongLi (approved):
+1

PR #957, opened 2023-04-20
[release] Fix the NOTICE years of the META-INFO/NOTICE file nested in the generated jars are not correct

zhuangchong:

### Purpose

 the NOTICE years of the META-INFO/NOTICE file nested in the generated jars are not correct(This problem exists in all modules that automatically generate NOTICE files.)
iShot_2023-04-20_15 04 00
image

The cause of the problem is that the parent class `apache-23` has a maven attribute `<project.build.outputTimestamp>2020-01-22T15:10:15Z</project.build.outputTimestamp>`

https://github.com/apache/maven-apache-parent/blob/99592e229c20f079c1998875e5c24556d217b491/pom.xml#L95

Modified:
iShot_2023-04-20_15 15 33
image

### Tests

### API and Format 

### Documentation

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -69,6 +69,7 @@ under the License.
     </modules>
 
     <properties>
+        <project.build.outputTimestamp>${maven.build.timestamp}</project.build.outputTimestamp>
         <paimon.shade.jackson.version>2.14.2</paimon.shade.jackson.version>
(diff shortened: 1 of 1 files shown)


JingsongLi (approved):
+1

zhuangchong:
Thanks. @JingsongLi

PR #958, opened 2023-04-20
[core & hive] Introduce renameDatabase method for catalog

s7monk:
fix https://github.com/apache/incubator-paimon/issues/716

### Purpose

 Introduce renameDatabase method for catalog

### Tests

FileSystemCatalogTest

### API and Format 

no

### Documentation

no

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
@@ -90,6 +90,10 @@ public interface Catalog extends AutoCloseable {
     /**
      * Rename a database.
      *
+     * <p>NOTE: If you use object storage, such as S3 or OSS, please use this syntax carefully,
+     * because the renaming of object storage is not atomic, and only partial files may be moved in
+     * case of failure.
+     *
      * @param fromDatabase the name of the database which need to rename
      * @param toDatabase the new database
      * @param ignoreIfNotExists Flag to specify behavior when the database does not exist: if set to
--- a/paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
@@ -88,6 +88,19 @@ public interface Catalog extends AutoCloseable {
             throws DatabaseNotExistException, DatabaseNotEmptyException;
 
     /**
+     * Rename a database.
+     *
+     * @param fromDatabase the name of the database which need to rename
+     * @param toDatabase the new database
+     * @param ignoreIfNotExists Flag to specify behavior when the database does not exist: if set to
+     *     false, throw an exception, if set to true, do nothing.
+     * @throws DatabaseNotExistException if the fromDatabase does not exist
+     * @throws DatabaseAlreadyExistException if the toDatabase already exists
+     */
+    void renameDatabase(String fromDatabase, String toDatabase, boolean ignoreIfNotExists)
+            throws DatabaseNotExistException, DatabaseAlreadyExistException;
+
+    /**
      * Return a {@link Table} identified by the given {@link Identifier}.
      *
      * <p>System tables can be got by '$' splitter.
--- a/paimon-core/src/main/java/org/apache/paimon/catalog/FileSystemCatalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/catalog/FileSystemCatalog.java
@@ -100,6 +100,24 @@ public class FileSystemCatalog extends AbstractCatalog {
     }
 
     @Override
+    public void renameDatabase(String fromDatabase, String toDatabase, boolean ignoreIfNotExists)
+            throws DatabaseNotExistException, DatabaseAlreadyExistException {
+        if (!databaseExists(fromDatabase)) {
+            if (ignoreIfNotExists) {
+                return;
+            }
+
+            throw new DatabaseNotExistException(fromDatabase);
+        }
+
+        if (databaseExists(toDatabase)) {
+            throw new DatabaseAlreadyExistException(toDatabase);
+        }
+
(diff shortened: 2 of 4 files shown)


paimon-core/src/main/java/org/apache/paimon/catalog/FileSystemCatalog.java
```diff
@@ -99,6 +99,24 @@ public void dropDatabase(String name, boolean ignoreIfNotExists, boolean cascade
...
+            }
+
+            throw new DatabaseNotExistException(fromDatabase);
+        }
+
+        if (databaseExists(toDatabase)) {
+            throw new DatabaseAlreadyExistException(toDatabase);
+        }
+
+        uncheck(() -> fileIO.rename(databasePath(fromDatabase), databasePath(toDatabase)));
```
SteNicholas, line 117: If there are several task that write the table data of this database, what's the behavior of these tasks?

s7monk: I think that if there are multiple tasks writing to this database, after modifying the database name, all tasks need to re-designate the database. Is there any way to make the writing task aware of the database name change?

SteNicholas: IMO, there is no way to aware of database name change, therefore the `renameDatabase` depend on the write path listener which we could introduce the listener to update the path of writer.

s7monk: ok, let me think about it.

paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
```diff
@@ -87,6 +87,19 @@ public interface Catalog extends AutoCloseable {
     void dropDatabase(String name, boolean ignoreIfNotExists, boolean cascade)
             throws DatabaseNotExistException, DatabaseNotEmptyException;
 
+    /**
+     * Rename a database.
```
SteNicholas, line 91: Adds the blew note for renaming database:

```
     * <p>NOTE: If you use object storage, such as S3 or OSS, please use this syntax carefully,
     * because the renaming of object storage is not atomic, and only partial files may be moved in
     * case of failure.
```

paimon-core/src/main/java/org/apache/paimon/catalog/FileSystemCatalog.java
SteNicholas: @s7monk, above problem could be solved in another pull request. This pull request could keep the same behavior of `renameTable`.

paimon-core/src/test/java/org/apache/paimon/catalog/FileSystemCatalogTest.java
```diff
@@ -44,4 +45,11 @@ public void testListDatabasesWhenNoDatabases() {
         List<String> databases = catalog.listDatabases();
         assertThat(databases).isEmpty();
     }
+
+    @Test
+    public void testRenameDatabase() throws Exception {
+        catalog.createDatabase("db1", false);
+        catalog.renameDatabase("db1", "db2", false);
+        assertThat(catalog.listDatabases()).isEqualTo(Lists.newArrayList("db2"));
```
SteNicholas, line 53: Could we add the test case to rename to the same name of other database?

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
```diff
@@ -177,6 +177,12 @@ public void dropDatabase(String name, boolean ignoreIfNotExists, boolean cascade
         }
     }
 
+    @Override
+    public void renameDatabase(String fromDatabase, String toDatabase, boolean ignoreIfNotExists)
+            throws DatabaseNotExistException, DatabaseAlreadyExistException {
+        throw new UnsupportedOperationException("Hive catalog does not support rename database.");
```
SteNicholas, line 183: Why not support rename database of `HiveCatalog`? Just uses the `IMetaStoreClient#alterDatabase`?

SteNicholas (changes requested):
@s7monk, I have left some comments for rename database. PTAL.

paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
s7monk: ok,I will do it.

paimon-core/src/test/java/org/apache/paimon/catalog/FileSystemCatalogTest.java
s7monk: I will do it.

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
s7monk: Hive does not support rename database.

s7monk:
I have updated this PR. PTAL. @SteNicholas thx.

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
SteNicholas: @s7monk, could you take a look at above method `IMetaStoreClient#alterDatabase`?

JingsongLi:
@s7monk could you take a look at above method `IMetaStoreClient#alterDatabase` (from @SteNicholas )?

s7monk:
> @s7monk could you take a look at above method `IMetaStoreClient#alterDatabase` (from @SteNicholas )?

ok.

PR #962, opened 2023-04-20
[flink][bug] Datetime parsing should not be affected by server time zone in MySqlDebeziumJsonEventParser

### LINKED ISSUE #945: [Bug] Mysql cdc action can not sync mysql datetime type correct
### Search before asking

### Paimon version

0.4

### Compute Engine

flink

### Minimal reproduce step

1. we have a mysql table with `datetime` and `timestamp` columns.
2. start the paimon mysql  cdc action to sync the mysql data to paimon
3. result: the timestamp type is correct, and the datetime type is not correct.
4. the time zone of my mysql and computer are all cst-8 
5. I use flink cdc (2.3 version) to sync the table ,it is correct.

1.mysql table 
```

mysql> select * from datatime_test;
+----+---------------------+---------------------+
| id | a                   | b                   |
+----+---------------------+---------------------+
|  1 | 2023-04-19 13:45:12 | 2023-04-19 13:45:12 |
+----+---------------------+---------------------+
1 row in set (0.00 sec)

mysql> show global variables like '%time_zone%';
+------------------+--------+
| Variable_name    | Value  |
+------------------+--------+
| system_time_zone | CST    |
| time_zone        | SYSTEM |
+------------------+--------+
2 rows in set (0.01 sec)

mysql> desc datatime_test;
+-------+-----------+------+-----+---------+-------+
| Field | Type      | Null | Key | Default | Extra |
+-------+-----------+------+-----+---------+-------+
| id    | int       | NO   | PRI | NULL    |       |
| a     | datetime  | YES  |     | NULL    |       |
| b     | timestamp | YES  |     | NULL    |       |
+-------+-----------+------+-----+---------+-------+
3 rows in set (0.01 sec)
```

2. paimon table 
```
Flink SQL> select * from datatime_test;
+----+---------------------+---------------------+
| id |                   a |                   b |
+----+---------------------+---------------------+
|  1 | 2023-04-19 21:45:12 | 2023-04-19 13:45:12 |
+----+---------------------+---------------------+
```

### Anything else?

_No response_

tsreaper:
### Purpose

This PR fixes #945.

According to [MySQL standard](https://dev.mysql.com/doc/refman/8.0/en/datetime.html), display value of datetime is not affected by server time zone. So current implementation of `MySqlDebeziumJsonEventParser` is not correct when parsing datetime.

### Tests

* MySqlSyncTableActionITCase#testAllTypes

### API and Format 

N/A

### Documentation

N/A

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/cdc/mysql/MySqlDebeziumJsonEventParser.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/cdc/mysql/MySqlDebeziumJsonEventParser.java
@@ -37,6 +37,7 @@ import java.math.BigDecimal;
 import java.time.Instant;
 import java.time.LocalDateTime;
 import java.time.ZoneId;
+import java.time.ZoneOffset;
 import java.util.ArrayList;
 import java.util.Base64;
 import java.util.HashMap;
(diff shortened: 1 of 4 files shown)


JingsongLi (approved):
+1

PR #963, opened 2023-04-20
[doc] fix flink api doc error

SYG-ATONE:
### Purpose

fix flink api error. 'USE paimon'should be 'USE CATALOG paimon'

### Tests

no

### API and Format 

no

### Documentation

api#flink-api.md


JingsongLi (approved):
+1 thanks @SYG-ATONE

PR #968, opened 2023-04-21
[flink]Modify the prompt when creating a flink table using FlinkCatalog.

wg1026688210:
Modify the  prompt when creating a flink table using FlinkCatalog.


JingsongLi (approved):
Looks good to me!

PR #969, opened 2023-04-21
[test] FileStoreTableTestBases should use TraceableFileIO

JingsongLi:
### Purpose

TraceableFileIO is for testing connection leak


yuzelin (approved):
+1

PR #976, opened 2023-04-21
[core] use append-only mode as the default write mode

### LINKED ISSUE #973: [Feature] Disable ChangelogValueCountFileStoreTable by default
### Search before asking

### Motivation

We can disable `ChangelogValueCountFileStoreTable` by default.
Most of users want to use append-only mode.

### Solution

_No response_

### Anything else?

_No response_

zhuangchong:

### Purpose

close #973 

### Tests

### API and Format 

### Documentation

### DIFF
--- a/docs/layouts/shortcodes/generated/core_configuration.html
+++ b/docs/layouts/shortcodes/generated/core_configuration.html
@@ -382,9 +382,9 @@
         </tr>
         <tr>
             <td><h5>write-mode</h5></td>
-            <td style="word-wrap: break-word;">change-log</td>
+            <td style="word-wrap: break-word;">auto</td>
             <td><p>Enum</p></td>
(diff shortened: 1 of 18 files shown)


docs/layouts/shortcodes/generated/core_configuration.html
```diff
@@ -376,7 +376,7 @@
         </tr>
         <tr>
             <td><h5>write-mode</h5></td>
-            <td style="word-wrap: break-word;">change-log</td>
+            <td style="word-wrap: break-word;">append-only</td>
```
JingsongLi, line 379: We may need a `auto` write mode...
Append-only for non-primary-key.
Change-log for primary key.

zhuangchong: @JingsongLi Thanks. done.

PR #986, opened 2023-04-22
[core] Implove SchemaChange class

s7monk:
fix https://github.com/apache/incubator-paimon/issues/985

### Purpose

Add updateColumnComment and updateColumnNullability overloaded methods to the SchemaChange class

### Tests

CatalogTestBase

### API and Format 

no

### Documentation

no


JingsongLi:
We should support ROW type for hive catalog first.

JingsongLi:
See tast failure.

PR #992, opened 2023-04-23
[hotfix] Add IssueNavigationLink for IDEA git log

Zouxxyy:
### Purpose

Add IssueNavigationLink for IDEA, so we would click the link to jira and PR from IDEA

image

### Tests

None

### API and Format 

None

### Documentation

None


JingsongLi (approved):
+1

PR #994, opened 2023-04-23
Hive catalog, name should be lower case

### LINKED ISSUE #990: [Bug] 【doc】Table name[MyTable] cannot contain upper case
### Search before asking

### Paimon version

0.4

### Compute Engine

flink

### Minimal reproduce step

CREATE CATALOG my_hive WITH (
>     'type' = 'paimon',
>     'metastore' = 'hive',
>     'uri' = 'thrift://***:9083',
>     'warehouse' = 'hdfs://***/user/zym/paimon'
> );

Flink SQL> USE CATALOG my_hive;

CREATE TABLE MyTable (
>     user_id BIGINT,
>     item_id BIGINT,
>     behavior STRING,
>     dt STRING,
>     hh STRING,
>     PRIMARY KEY (dt, hh, user_id) NOT ENFORCED
> );
java.lang.IllegalStateException: Table name[MyTable] cannot contain upper case

hive catalog org.apache.paimon.hive.HiveCatalog#checkIdentifierUpperCase  ;

do we need to change the table name "MyTable" to lowercase "my_table"?

### Anything else?

_No response_

ymZhao1001:
close #990

### DIFF
--- a/docs/content/how-to/creating-catalogs.md
+++ b/docs/content/how-to/creating-catalogs.md
@@ -80,6 +80,8 @@ USE paimon.default;
 
 By using Paimon Hive catalog, changes to the catalog will directly affect the corresponding Hive metastore. Tables created in such catalog can also be accessed directly from Hive.
 
+To use Hive catalog, Database name、Table name and Field names should be lower case.
+
 {{< tabs "hive-metastore-example" >}}
 
 {{< tab "Flink" >}}
--- a/docs/content/how-to/creating-catalogs.md
+++ b/docs/content/how-to/creating-catalogs.md
@@ -80,7 +80,7 @@ USE paimon.default;
 
 By using Paimon Hive catalog, changes to the catalog will directly affect the corresponding Hive metastore. Tables created in such catalog can also be accessed directly from Hive.
 
-To use Hive catalog, Database name、Table name and Field names should be lower case.
+To use Hive catalog, Database name, Table name and Field names should be lower case.
 
 {{< tabs "hive-metastore-example" >}}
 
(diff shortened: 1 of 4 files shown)


docs/content/how-to/creating-catalogs.md
```diff
@@ -80,6 +80,8 @@ USE paimon.default;
 
 By using Paimon Hive catalog, changes to the catalog will directly affect the corresponding Hive metastore. Tables created in such catalog can also be accessed directly from Hive.
 
+To use Hive catalog, Database name、Table name and Field names should be lower case.
```
JingsongLi, line 83: 、 ->  ,

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
```diff
@@ -334,6 +335,8 @@ public void alterTable(
             // first commit changes to underlying files
             TableSchema schema = schemaManager.commitChanges(changes);
```
JingsongLi, line 336: validate in `schemaManager.commitChanges`?
assert in `AddColumn` and `RenameColumn`.

Otherwise, the upper case fields will be commit into file system.

ymZhao1001: Will it affect other types of catalogs

ymZhao1001: checkFieldNamesUpperCaseInSchemaChange before commit to underlying files . please take a look

JingsongLi (approved):
+1

PR #1006, opened 2023-04-24
[flink] Support skipping incompatible tables in MySqlSyncDatabaseAction

### LINKED ISSUE #996: [Feature] Skipping synchronizing table of which schema is incompatible with Paimon's instead of throwing exception in MySqlSyncDatabaseAction
### Search before asking

### Motivation

In `MySqlSyncDatabaseAction`, if some table's schema is not compatible with Paimon, the process will fail because an exception will be thrown.
I think it's better to skip the incompatible table and put the error message to log.

### Solution

`assertSchemaCompatible` return a flag to indicates whether compatible. If not, put message to log and continue.

### Anything else?

_No response_

yuzelin:
### Purpose

Fix #996 

### Tests

`MySqlSyncDatabaseActionITCase#testIgnoreIncompatibleTables`

### API and Format 

mysql-sync-database new arg:
--ignore-incompatible (default false)

### Documentation
CDC Ingestion # Synchronizing Databases


JingsongLi (approved):
+1

PR #1018, opened 2023-04-25
[oss] Do not relocate hadoop classes in paimon-oss

JingsongLi:
image

Some configurations will depend on class paths and relocation will cause class not to be found.


leaves12138:
+1

PR #1022, opened 2023-04-25
[core] Introduce consume-id for retore consuming and reserve snapshot

### LINKED ISSUE #881: [Feature] Support consume group id like kafka
### Search before asking

### Motivation

Consume group id in kafka is useful for restarting without state.
We can use files to record consume group id too.

### Solution

_No response_

### Anything else?

_No response_

JingsongLi:
fix #881

### Purpose

This PR is intended to introduce a consume mode, when a `consume-id` is specified, scan will writes the progress of the consume to the file system, which has several advantages:
1. When the job is killed, the newly started job can continue to consume from the previous progress without resuming from the state.
2. When deleting expired snapshots, it ensures that the snapshot being consumed will not be deleted, so that the consumer will not read the expired snapshot and get hung.

This PR also introduces a new Source implementation:
- MonitorFunction
- ReadOperator

This is to avoid Flink's Source interface forcing a checkpoint on Splits, which may only read part of the snapshot data when consuming, the progress of consuming cannot be recorded.

### Documentation

Added.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/source/AbstractInnerTableScan.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/source/AbstractInnerTableScan.java
@@ -21,6 +21,8 @@ package org.apache.paimon.table.source;
 import org.apache.paimon.CoreOptions;
 import org.apache.paimon.CoreOptions.ChangelogProducer;
 import org.apache.paimon.annotation.VisibleForTesting;
+import org.apache.paimon.consumer.Consumer;
+import org.apache.paimon.consumer.ConsumerManager;
 import org.apache.paimon.operation.FileStoreScan;
 import org.apache.paimon.table.source.snapshot.CompactedStartingScanner;
 import org.apache.paimon.table.source.snapshot.ContinuousCompactorStartingScanner;
@@ -34,6 +36,8 @@ import org.apache.paimon.table.source.snapshot.StartingScanner;
 import org.apache.paimon.table.source.snapshot.StaticFromSnapshotStartingScanner;
 import org.apache.paimon.table.source.snapshot.StaticFromTimestampStartingScanner;
 import org.apache.paimon.utils.Preconditions;
+
+import java.util.Optional;
 
 import static org.apache.paimon.CoreOptions.FULL_COMPACTION_DELTA_COMMITS;
 
@@ -63,6 +67,16 @@ public abstract class AbstractInnerTableScan implements InnerTableScan {
             Preconditions.checkArgument(
                     isStreaming, "Set 'streaming-compact' in batch mode. This is unexpected.");
             return new ContinuousCompactorStartingScanner();
+        }
+
+        // read from consumer id
+        String consumerId = options.consumerId();
+        if (consumerId != null) {
+            ConsumerManager consumerManager = snapshotSplitReader.consumerManager();
+            Optional<Consumer> consumer = consumerManager.consumer(consumerId);
+            if (consumer.isPresent()) {
+                return new ContinuousFromSnapshotStartingScanner(consumer.get().nextSnapshot());
+            }
         }
 
         CoreOptions.StartupMode startupMode = options.startupMode();
--- a/paimon-core/src/main/java/org/apache/paimon/table/source/InnerStreamTableScanImpl.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/source/InnerStreamTableScanImpl.java
@@ -20,6 +20,7 @@ package org.apache.paimon.table.source;
 
 import org.apache.paimon.CoreOptions;
 import org.apache.paimon.Snapshot;
+import org.apache.paimon.consumer.Consumer;
 import org.apache.paimon.predicate.Predicate;
 import org.apache.paimon.table.source.snapshot.BoundedChecker;
 import org.apache.paimon.table.source.snapshot.CompactionChangelogFollowUpScanner;
@@ -202,5 +203,19 @@ public class InnerStreamTableScanImpl extends AbstractInnerTableScan
     @Override
     public void restore(@Nullable Long nextSnapshotId) {
         this.nextSnapshotId = nextSnapshotId;
+    }
+
+    @Override
+    public void notifyCheckpointComplete(@Nullable Long nextSnapshot) {
+        if (nextSnapshot == null) {
+            return;
+        }
+
+        String consumerId = options.consumerId();
+        if (consumerId != null) {
+            snapshotSplitReader
+                    .consumerManager()
+                    .recordConsumer(consumerId, new Consumer(nextSnapshot));
+        }
     }
 }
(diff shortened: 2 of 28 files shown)


paimon-core/src/main/java/org/apache/paimon/table/source/AbstractInnerTableScan.java
```diff
@@ -65,6 +69,16 @@ protected StartingScanner createStartingScanner(boolean isStreaming) {
             return new ContinuousCompactorStartingScanner();
         }
 
+        // read from consumer id
+        String consumerId = options.consumerId();
```
lilyzhoupeijie, line 73: do we need to compare the snapshot from consumer with snapshot from state?

paimon-core/src/main/java/org/apache/paimon/table/source/InnerStreamTableScanImpl.java
```diff
@@ -194,4 +195,18 @@ public Long checkpoint() {
...
+    public void notifyCheckpointComplete(@Nullable Long nextSnapshot) {
+        if (nextSnapshot == null) {
+            return;
+        }
+
+        String consumerId = options.consumerId();
+        if (consumerId != null) {
+            snapshotSplitReader
+                    .consumerManager()
+                    .recordConsumer(consumerId, new Consumer(nextSnapshot));
```
lilyzhoupeijie, line 209: If two flink programs with the same consumer-id exist simultaneously, what semantics do we guarantee? or  it is necessary we try to  guarantee? cause there is no lock on the file,and anyone with the consumer-id can change it.

paimon-core/src/main/java/org/apache/paimon/table/source/AbstractInnerTableScan.java
JingsongLi: I think if the relevant content is recorded inside the state, it should not be read from the consumer file and should be trusted in the state.
Do you think something bad will happen?

paimon-core/src/main/java/org/apache/paimon/table/source/InnerStreamTableScanImpl.java
JingsongLi: Good point!
I'm not sure what kind of behavior it should be, so I'll let them each play it out here, not sure who will record success.

JingsongLi:
Thanks @lilyzhoupeijie for the review, do you have any thoughts on the solution? Does the current solution work to your satisfaction?

lilyzhoupeijie:
maybe each consumer has an inner UUID store in state?and use UUID +consumer-id to update the record file and then we can decide which semantics  to use.
or compare the snapshot in record file with current snapshot -1,and then to decide whether to update? if not equals,it means that some other consumer has finished reading the snapshot.something like java CAS to avoid ABA ?
or use the feature in storage,such as hdfs,it provide lease to keep single writer,keep only one consumer writing the record file

JingsongLi:
> maybe each consumer has an inner UUID store in state?and use UUID +consumer-id to update the record file and then we can decide which semantics to use. or compare the snapshot in record file with current snapshot -1,and then to decide whether to update? if not equals,it means that some other consumer has finished reading the snapshot.something like java CAS to avoid ABA ? or use the feature in storage,such as hdfs,it provide lease to keep single writer,keep only one consumer writing the record file

The discussion is based on the premise that we need to support a third requirement:

3. the same consumer-id supports multiple simultaneous reads, and the same snapshot needs to be guaranteed to be read by only one reader. It should behave just like the groupId of the message queue.

I think it's an interesting topic, but I'm not sure if the current users have such a need, and I propose that I can keep it for now.
The current information is enough, for example I know which snapshot I read and I know how I need to update the consumer file, similar to CAS (compare and swap), but the hard part is the subsequent processing, really to avoid it, I guess we can only have pre-commit and commit phases.

lilyzhoupeijie:
> > maybe each consumer has an inner UUID store in state?and use UUID +consumer-id to update the record file and then we can decide which semantics to use. or compare the snapshot in record file with current snapshot -1,and then to decide whether to update? if not equals,it means that some other consumer has finished reading the snapshot.something like java CAS to avoid ABA ? or use the feature in storage,such as hdfs,it provide lease to keep single writer,keep only one consumer writing the record file
> 
> The discussion is based on the premise that we need to support a third requirement:
> 
> 3. the same consumer-id supports multiple simultaneous reads, and the same snapshot needs to be guaranteed to be read by only one reader. It should behave just like the groupId of the message queue.
> 
> I think it's an interesting topic, but I'm not sure if the current users have such a need, and I propose that I can keep it for now. The current information is enough, for example I know which snapshot I read and I know how I need to update the consumer file, similar to CAS (compare and swap), but the hard part is the subsequent processing, really to avoid it, I guess we can only have pre-commit and commit phases.

Unless we have a global lock, maybe we cannot prevent data from being consumed multiple times.But is there some way to avoid sending overconsumed data to downstream operators?so that we still can keep an end-to-end eos.
similar with:
https://github.com/apache/incubator-paimon/issues/881
is this still needed?maybe global lock or pre-allocation based on snapshot.if not needed,this one can be closed.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/operator/MonitorFunction.java
```diff
@@ -0,0 +1,222 @@
...
+            }
+
+            // given that the parallelism of the function is 1, we can only have 1 retrieved items.
+
+            Preconditions.checkArgument(
+                    retrievedStates.size() <= 1,
+                    getClass().getSimpleName() + " retrieved invalid state.");
+
+            if (retrievedStates.size() == 1) {
+                this.scan.restore(retrievedStates.get(0));
```
FangYongs, line 130: Should we need to restore scan with the snapshot id in state or in consume-id file here?

JingsongLi:
> > > maybe each consumer has an inner UUID store in state?and use UUID +consumer-id to update the record file and then we can decide which semantics to use. or compare the snapshot in record file with current snapshot -1,and then to decide whether to update? if not equals,it means that some other consumer has finished reading the snapshot.something like java CAS to avoid ABA ? or use the feature in storage,such as hdfs,it provide lease to keep single writer,keep only one consumer writing the record file
> > 
> > 
> > The discussion is based on the premise that we need to support a third requirement:
> > 
> > 3. the same consumer-id supports multiple simultaneous reads, and the same snapshot needs to be guaranteed to be read by only one reader. It should behave just like the groupId of the message queue.
> > 
> > I think it's an interesting topic, but I'm not sure if the current users have such a need, and I propose that I can keep it for now. The current information is enough, for example I know which snapshot I read and I know how I need to update the consumer file, similar to CAS (compare and swap), but the hard part is the subsequent processing, really to avoid it, I guess we can only have pre-commit and commit phases.
> 
> Unless we have a global lock, maybe we cannot prevent data from being consumed multiple times.But is there some way to avoid sending overconsumed data to downstream operators?so that we still can keep an end-to-end eos. similar with: #881 is this still needed?maybe global lock or pre-allocation based on snapshot.if not needed,this one can be closed.

We can disregard it for now. In fact, a simple case oriented to multiple reads is to report an error.

JingsongLi:
Thanks @lilyzhoupeijie @yuzelin for your review, merged...

PR #1023, opened 2023-04-25
[core] Introduce file.compression

### LINKED ISSUE #889: [Feature] Introduce file.compression

Tartarus0zm:
### Purpose

fix #889 

### Tests

### API and Format 

Format

### Documentation

introduce a new feature

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/io/KeyValueFileWriterFactory.java
+++ b/paimon-core/src/main/java/org/apache/paimon/io/KeyValueFileWriterFactory.java
@@ -18,6 +18,7 @@
 
 package org.apache.paimon.io;
 
+import org.apache.paimon.CoreOptions;
 import org.apache.paimon.KeyValue;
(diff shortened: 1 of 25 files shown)


Tartarus0zm:
@JingsongLi  please take a look, if you have time, thanks

paimon-core/src/main/java/org/apache/paimon/io/KeyValueFileWriterFactory.java
```diff
@@ -160,6 +168,18 @@ private Builder(
 
         public KeyValueFileWriterFactory build(
```
JingsongLi, line 169: Do not provide this method, it just for testing?

PR #1028, opened 2023-04-26
[flink][bug] MySQL datetime and timestamp default precision should be 0

### LINKED ISSUE #1020: [Bug]  CDC  schema change  error :  mysql  type DATETIME where  in  UpdatedDataFields tranform  is  different from  mysql schema.
### Search before asking

### Paimon version

paimon 0.4-snapshot
flink 1.16.1
mysql 8.0.27

### Compute Engine

flink

### Minimal reproduce step

I have  a   column  c9   , type  is  DATETIME  (without precision) .
 When I had  add column :  alter table cdc_test  add c12  INT,
It   had  made error:   ERROR 
Caused by: java.lang.UnsupportedOperationException: Cannot convert field c9 from type TIMESTAMP(0) to TIMESTAMP(6)
	at org.apache.paimon.flink.sink.cdc.UpdatedDataFieldsProcessFunction.applySchemaChange(UpdatedDataFieldsProcessFunction.java:146)
	at org.apache.paimon.flink.sink.cdc.UpdatedDataFieldsProcessFunction.processElement(UpdatedDataFieldsProcessFunction.java:66)
	at org.apache.paimon.flink.sink.cdc.UpdatedDataFieldsProcessFunction.processElement(UpdatedDataFieldsProcessFunction.java:50)

I had locationed  that  column c9  in  table schema  is  TIMESTAMP(0)  ,but in  updatedDataFields  is  TIMESTAMP(6)

1111

### Anything else?

MySqlTypeUtils.toDataType:
private static final int JDBC_TIMESTAMP_BASE_LENGTH = 19;
 case DATETIME:
            case TIMESTAMP:
                if (length == null) {
                    return DataTypes.TIMESTAMP();
                } else if (length >= JDBC_TIMESTAMP_BASE_LENGTH) {
                    if (length > JDBC_TIMESTAMP_BASE_LENGTH + 1) {
                        // Timestamp with a fraction of seconds.
                        // For example "2023-03-23 17:20:00.01".
                        // The decimal point will occupy 1 character.
                        return DataTypes.TIMESTAMP(length - JDBC_TIMESTAMP_BASE_LENGTH - 1);
                    } else {
                        return DataTypes.TIMESTAMP(0);
                    }
in mysql schema   length is 19   ,  in   updatedDataFields   length  is null .

tsreaper:
### Purpose

This PR fixes #1020.

In MySQL, `datetime` and `timestamp` default precision is 0 (see https://dev.mysql.com/doc/refman/8.0/en/date-and-time-type-syntax.html). However in Paimon, `timestamp` default precision is 6.

This PR fixes the conversion between MySQL `datetime` / `timestamp` types with default precision and Paimon `timestamp` type.

### Tests

* `MySqlSyncTableActionITCase#testAllTypes`

### API and Format 

N/A

### Documentation

N/A

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/cdc/mysql/MySqlTypeUtils.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/cdc/mysql/MySqlTypeUtils.java
@@ -171,7 +171,9 @@ public class MySqlTypeUtils {
             case DATETIME:
             case TIMESTAMP:
                 if (length == null) {
-                    return DataTypes.TIMESTAMP();
+                    // default precision is 0
+                    // see https://dev.mysql.com/doc/refman/8.0/en/date-and-time-type-syntax.html
+                    return DataTypes.TIMESTAMP(0);
                 } else if (length >= JDBC_TIMESTAMP_BASE_LENGTH) {
                     if (length > JDBC_TIMESTAMP_BASE_LENGTH + 1) {
                         // Timestamp with a fraction of seconds.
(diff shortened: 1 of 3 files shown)


JingsongLi (approved):
+1

PR #1030, opened 2023-04-26
[core] Add validation in Schema#Builder

yuzelin:
### Purpose

Currently, Schema#Builder.build hasn't check the arguments. This PR add validation to it.

### Tests

No.

### API and Format 

No.

### Documentation

No.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/schema/Schema.java
+++ b/paimon-core/src/main/java/org/apache/paimon/schema/Schema.java
@@ -35,7 +35,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
-import java.util.function.Function;
 import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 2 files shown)


paimon-core/src/main/java/org/apache/paimon/schema/Schema.java
```diff
@@ -94,7 +95,22 @@ public String comment() {
     private static List<DataField> normalizeFields(
             List<DataField> fields, List<String> primaryKeys, List<String> partitionKeys) {
         List<String> fieldNames = fields.stream().map(DataField::name).collect(Collectors.toList());
+
+        Set<String> duplicateColumns = duplicate(fieldNames);
```
JingsongLi, line 99: Add a test for `Schema.newBuilder`?

paimon-core/src/main/java/org/apache/paimon/schema/Schema.java
```diff
@@ -133,6 +155,12 @@ private static List<DataField> normalizeFields(
         return newFields;
     }
 
+    private static Set<String> duplicate(List<String> names) {
```
SteNicholas, line 158: Could we return the result that uses `anyMatch(name -> Collections.frequency(names, name) > 1)` that represents whether the `names` is duplicate? Meanwhile uses the return value to invoke `Preconditions.checkState`.

yuzelin: I collect duplicate columns to put them in error message:
`Found: {duplicatedColumns}`. 
Maybe it is not necessary. What's your suggestion?

JingsongLi (approved):
+1

PR #1031, opened 2023-04-26
[flink]validateMySqlConfig

ymZhao1001:
[flink]

### Purpose

validateMySqlConfig

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/cdc/mysql/MySqlActionUtils.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/cdc/mysql/MySqlActionUtils.java
@@ -208,34 +208,29 @@ class MySqlActionUtils {
 
     private static void validateMySqlConfig(Configuration mySqlConfig) {
         checkArgument(
-                mySqlConfig.get(MySqlSourceOptions.HOSTNAME) != null
-                        && !mySqlConfig.get(MySqlSourceOptions.HOSTNAME).isEmpty(),
+                mySqlConfig.get(MySqlSourceOptions.HOSTNAME) != null,
                 String.format(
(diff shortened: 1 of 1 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/cdc/mysql/MySqlActionUtils.java
```diff
@@ -202,4 +205,37 @@ static MySqlSource<String> buildMySqlSource(Configuration mySqlConfig) {
                 new JsonDebeziumDeserializationSchema(true, customConverterConfigs);
         return sourceBuilder.deserializer(schema).includeSchemaChanges(true).build();
     }
+
+    private static void validateMySqlConfig(Configuration mySqlConfig) {
```
JingsongLi, line 209: Can we invoke `org.apache.flink.table.factories.FactoryUtil.validateFactoryOptions`?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/cdc/mysql/MySqlActionUtils.java
```diff
@@ -202,4 +205,37 @@ static MySqlSource<String> buildMySqlSource(Configuration mySqlConfig) {
...
+        checkArgument(
+                mySqlConfig.get(MySqlSourceOptions.HOSTNAME) != null
```
JingsongLi, line 211: `mySqlConfig.contains(MySqlSourceOptions.HOSTNAME)`
Just check `contains` is OK, if it is a empty string, detailed message will be reported in mysql client.

ymZhao1001: remove the check whether it is empty

JingsongLi (approved):
+1

PR #1033, opened 2023-04-26
[hotfix][docs]Add tips for using the generate-docs maven plugin

Tartarus0zm:
### Purpose

when use generate-docs plugin, we should install jar first.

### Tests

no

### API and Format 

no

### Documentation

no


JingsongLi (approved):
+1

Issue #1036, opened 2023-04-26
[Bug] readers in the cached list not close correctly


leaves12138:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/incubator-paimon/issues) and found nothing similar.

### Paimon version

0.4

### Compute Engine

trino

### Minimal reproduce step

1. produce a mount of data in s3 filesystem
2. config a few connections
3. query by trino

### What doesn't meet your expectations?

the world no more bug

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

leaves12138:
merged

PR #1037, opened 2023-04-26
[core][bug] readers don't close while io exception happening

leaves12138:
readers don't close while io exception happening, the readers cached in the list will lose control (#1036)

### Purpose

### Tests

### API and Format 

### Documentation

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/mergetree/MergeTreeReaders.java
+++ b/paimon-core/src/main/java/org/apache/paimon/mergetree/MergeTreeReaders.java
@@ -28,6 +28,7 @@ import org.apache.paimon.mergetree.compact.MergeFunctionWrapper;
 import org.apache.paimon.mergetree.compact.ReducerMergeFunctionWrapper;
 import org.apache.paimon.mergetree.compact.SortMergeReader;
 import org.apache.paimon.reader.RecordReader;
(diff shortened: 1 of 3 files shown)


paimon-core/src/main/java/org/apache/paimon/mergetree/MergeTreeReaders.java
```diff
@@ -70,8 +71,14 @@ public static RecordReader<KeyValue> readerForSection(
             MergeFunctionWrapper<KeyValue> mergeFunctionWrapper)
             throws IOException {
         List<RecordReader<KeyValue>> readers = new ArrayList<>();
-        for (SortedRun run : section) {
-            readers.add(readerForRun(run, readerFactory));
+        // if one of the readers creating failed, we need to close them all.
+        try {
+            for (SortedRun run : section) {
+                readers.add(readerForRun(run, readerFactory));
```
tsreaper, line 77: `readerForRun` is also used in `ChangelogMergeTreeRewriter#rewriteChangelogCompaction`. Consider creating a method `List<RecordReader<KeyValue>> readerListForSection(List<SortedRun> section, KeyValueFileReaderFactory readerFactory)`.

JingsongLi (approved):
+1

PR #1040, opened 2023-04-26
[core] add range check of snapshot id during scan to prevent endless waiting

### LINKED ISSUE #807: [Bug] restored from checkpoint, ContinuousDataFileSnapshotEnumerator keeps waiting an old snapshot which is expired
### Search before asking

### Paimon version

0.4-SNAPSHOT

### Compute Engine

flink 1.14

### Minimal reproduce step

Stop a flink countinues reading from paimon with savepoint, and after the latest snapshot of the source paimon table is expired, restore the flink from savepoint.

Discover the target snapshot restored from checkpoint/savepoint is outdated and throw exception.

### Anything else?

_No response_

liming30:
### Purpose

[core] add range check of snapshot id during scan to prevent endless waiting.  This will fix #807 .

### Tests

No.

### API and Format 

No.

### Documentation

No.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/source/InnerStreamTableScanImpl.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/source/InnerStreamTableScanImpl.java
@@ -114,6 +114,13 @@ public class InnerStreamTableScanImpl extends AbstractInnerTableScan
             }
 
             if (!snapshotManager.snapshotExists(nextSnapshotId)) {
+                Long earliestSnapshotId = snapshotManager.earliestSnapshotId();
+                if (earliestSnapshotId != null && earliestSnapshotId > nextSnapshotId) {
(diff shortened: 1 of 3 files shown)


JingsongLi:
In the exception message, we should remind the user what to do at this point to avoid this situation, e.g:
1. increase the snapshot expiration time.
2. Use consumer-id: https://github.com/apache/incubator-paimon/pull/1022

liming30:
> In the exception message, we should remind the user what to do at this point to avoid this situation, e.g:
> 
> 1. increase the snapshot expiration time.
> 2. Use consumer-id: [[core] Introduce consume-id for retore consuming and reserve snapshot #1022](https://github.com/apache/incubator-paimon/pull/1022)

@JingsongLi resolved

JingsongLi (approved):
+1

PR #1052, opened 2023-04-28
[doc] Fix document mistake of append only table

ymZhao1001:
fix doc mistake


SteNicholas (approved):
LGTM. Thanks for @ymZhao1001 contribution.

PR #1064, opened 2023-05-04
[CI]remove benchmark code check

liugddx:

### Purpose

Benchmarking frameworks do not require sonar checks.

image

### Tests

### API and Format 

### Documentation


JingsongLi:
Maybe we can just fix them?

liugddx:
> Maybe we can just fix them?

OK.

PR #1080, opened 2023-05-05
[CI] Add a GHA for adding labels based on modified files

jackwener:

### Purpose

Add a GHA for adding labels based on modified files


JingsongLi (approved):
+1
thanks @jackwener

JingsongLi:
reverted, this will result in PRs that do not pass the test and show a pattern ✅ that passes the test.

jackwener:
> reverted, this will result in PRs that do not pass the test and show a pattern ✅ that passes the test.

Is there a example to see this condition? I can take a look it.

jackwener:
I can't see the error log.
Just guess, If it failed due to this GitHub Action, it may be caused by Paimon don't have those `label`, we just need to add it.

PR #1083, opened 2023-05-06
[filnk] support alter constraint for flink

zhangjun0x01:
### Purpose

Related to https://github.com/apache/incubator-paimon/issues/785

### Tests

org.apache.paimon.flink.SchemaChangeITCase


JingsongLi (commented):
Hi @zhangjun0x01 , we don't have plan to support change pk, change pk will break the data, we can not support this now.

PR #1084, opened 2023-05-06
[hotfix][hive] Improve hive create table test case.

zhuangchong:
### Purpose

Improve hive create table test case

### Tests

CreateTableITCase


JingsongLi (approved):
+1

PR #1099, opened 2023-05-08
[core]check conflict for option

zhangjun0x01:
### Purpose

fix https://github.com/apache/incubator-paimon/issues/1097

### Tests

org.apache.paimon.flink.CatalogTableITCase#testConflictOption
org.apache.paimon.spark.SparkReadITCase#testConflictOption


paimon-core/src/main/java/org/apache/paimon/schema/SchemaManager.java
```diff
@@ -177,6 +182,17 @@ public TableSchema createTable(Schema schema) throws Exception {
         }
     }
 
+    private void checkOptionConflict(Map<String, String> options) {
```
JingsongLi, line 185: Add in `SchemaValidation.validateTableSchema`?

paimon-core/src/main/java/org/apache/paimon/schema/SchemaManager.java
```diff
@@ -170,6 +170,8 @@ public TableSchema createTable(Schema schema) throws Exception {
                             options,
                             schema.comment());
 
+            SchemaValidation.validateTableSchema(newSchema);
```
JingsongLi, line 173: Do we need this? Finally, `validateTableSchema` will be invoked.

paimon-core/src/main/java/org/apache/paimon/schema/SchemaManager.java
```diff
@@ -372,6 +374,8 @@ public TableSchema commitChanges(List<SchemaChange> changes) throws Exception {
                             newOptions,
                             schema.comment());
 
+            SchemaValidation.validateTableSchema(schema);
```
JingsongLi, line 377: Do we need this? Finally, `validateTableSchema` will be invoked.

Issue #634, opened 2023-03-19
[feature] Provide changelog-producer.row-deduplicate to deduplicate same change


JingsongLi:
In `lookup` and `full-compaction` changelog-producer, same record will produce -U and +U, but they are the same.
We can provide a option to avoid producing changelog for the same record.

wenlybi:
I want to do it. Could you assign it to me?

Aitozi:
@JingsongLi Can this option default be true?  Are there use case will rely on the same value -U and +U value ?

PR #1100, opened 2023-05-08
[core] provide changelog-producer.row-deduplicate to deduplicate same change

liming30:
### Purpose

 Provide `changelog-producer.row-deduplicate` to deduplicate same change.  #634 

### Tests

Modify the following unit test to verify:
- `org.apache.paimon.mergetree.compact.FullChangelogMergeFunctionWrapperTestBase`
- `org.apache.paimon.mergetree.compact.LookupChangelogMergeFunctionWrapperTest`

Modify the following integration tests to verify:

- `org.apache.paimon.flink.FullCompactionFileStoreITCase#testUpdate`
- `org.apache.paimon.flink.LookupChangelogWithAggITCase#testMultipleCompaction`

### API and Format

No

### Documentation

No

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-core/src/main/java/org/apache/paimon/CoreOptions.java
@@ -348,6 +348,13 @@ public class CoreOptions implements Serializable {
                                     + "This changelog file keeps the details of data changes, "
                                     + "it can be read directly during stream reads.");
 
+    public static final ConfigOption<Boolean> CHANGELOG_PRODUCER_ROW_DEDUPLICATE =
+            key("changelog-producer.row-deduplicate")
+                    .booleanType()
+                    .defaultValue(false)
+                    .withDescription(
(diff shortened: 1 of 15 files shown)


paimon-core/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -807,6 +814,10 @@ public ChangelogProducer changelogProducer() {
         return options.get(CHANGELOG_PRODUCER);
     }
 
+    public boolean rowDeduplicate() {
```
FangYongs, line 817: Use changelogRowDeduplicate()?

JingsongLi (approved):
Thanks @liming30 for the contribution!
The PR looks pretty good. +1

We can create an Issue for introducing `RecordEqualiser`, it is bit faster than comparator.

paimon-core/src/main/java/org/apache/paimon/CoreOptions.java
liming30: resolved.

PR #1115, opened 2023-05-10
[flink] update document for flink cdc action

zhangjun0x01:
### Purpose

Currently, the framework can not deal `DROP`, `RENAME` ,but it wouldn't throw an exception either, so we should update the doc.


JingsongLi (approved):
+1

PR #833, opened 2023-04-04
 [Feature] Optimize SortMergeReader: use loser tree to reduce comparisons

liming30:
### Purpose

 [Feature] Optimize SortMergeReader: use loser tree to reduce comparisons (#741)

### Tests

- Unit Tests: `SortMergeReaderTestBase` already exists to verify the correctness of `SortMergeReader`.
- Benchmark：I created a jmh benchmark to compare the performance of different implementations，Here is the [benchmark code](https://github.com/liming30/paimon-benchmark/blob/master/src/main/java/com/bytedance/MergeReaderBenchmark.java).

From the benchmark results, there is almost no difference between `loser-tree` and `min-heap` when `RecordReader`/`data volume` is small. With the increase of `RecordReader`/`data volume`, there can be about 10% performance improvement.

```
Benchmark                               (readersNum)  (recordNum)  Mode  Cnt     Score    Error  Units
MergeReaderBenchmark.group:min-heap            2         1000  avgt   10     0.172 ±  0.017  ms/op
MergeReaderBenchmark.group:loser-tree          2         1000  avgt   10     0.172 ±  0.015  ms/op
MergeReaderBenchmark.group:min-heap            2        10000  avgt   10     1.829 ±  0.247  ms/op
MergeReaderBenchmark.group:loser-tree          2        10000  avgt   10     1.633 ±  0.217  ms/op
MergeReaderBenchmark.group:min-heap            2       100000  avgt   10    17.664 ±  1.678  ms/op
MergeReaderBenchmark.group:loser-tree          2       100000  avgt   10    17.355 ±  1.238  ms/op
MergeReaderBenchmark.group:min-heap            5         1000  avgt   10     0.615 ±  0.095  ms/op
MergeReaderBenchmark.group:loser-tree          5         1000  avgt   10     0.616 ±  0.092  ms/op
MergeReaderBenchmark.group:min-heap            5        10000  avgt   10     6.207 ±  0.261  ms/op
MergeReaderBenchmark.group:loser-tree          5        10000  avgt   10     6.169 ±  0.221  ms/op
MergeReaderBenchmark.group:min-heap            5       100000  avgt   10    76.777 ±  6.675  ms/op
MergeReaderBenchmark.group:loser-tree          5       100000  avgt   10    61.951 ±  5.007  ms/op
MergeReaderBenchmark.group:min-heap           10         1000  avgt   10     1.719 ±  0.127  ms/op
[...]

PR #1116, opened 2023-05-10
[core][fix] reset the iterator in loserTree to null after releaseBatch.

liming30:
Fix #833 

### Purpose

When an exception occurs in the `readBatch` of the `RecordReader`,  the `close` of `LoserTree` may be the second time to close the `iterator`. 

solution: reset the `iterator` in `loserTree` to `null` after `releaseBatch`.

### Tests

Added unit test `org.apache.paimon.mergetree.compact#testLoserTreeCloseNormally` to verify that the `close` of `LoserTree` is normal.

### API and Format

No

### Documentation

No


JingsongLi (approved):
+1

PR #1117, opened 2023-05-10
[doc] Fix trino doc error

s7monk:
### Purpose

Fix trino doc error

### Tests

NO.

### API and Format

NO.

### Documentation

engines#trino.md


JingsongLi (approved):
+1

PR #1121, opened 2023-05-10
[core] The default value of continuous.discovery-interval can be 10 s

JingsongLi:
### Purpose

At present, the default value for `continuous.discovery-interval` is 1 second, which is too short and can cause frequent access to the file system, causing significant pressure.

We can modify its default value to 10 seconds, as our current recommended stream read delay is 30 seconds to 1 minute, which does not have a significant impact.

### Tests

No

### API and Format

No

### Documentation

Yes


zhuangchong (approved):
+1

Issue #1004, opened 2023-04-24
[Feature] add listPartitions and dropPartitions in paimon Catalog


leaves12138:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/incubator-paimon/issues) and found nothing similar.

### Motivation

1、Paimon data lake support partition function but Catalog does not support partition actions

2、Flink Catalog has listPartitions function

### Solution

1、add listPartitions function in Catalog

2、add dropPartitions function in Catalog

### Anything else?

listPartition function could finish by Projections, pass the avro schema to the avro reader, get rid of redundant bytes decoding.

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

zhuangchong:
Related issues: #747

SteNicholas:
@leaves12138, duplicated with #747. Closed.

PR #1122, opened 2023-05-10
[core] paimon Catalog listPartition support 

leaves12138:
### Purpose

paimon Catalog listPartition support (#1004)


paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
```diff
@@ -187,6 +187,20 @@ default boolean caseSensitive() {
...
+     * List partitions by the given identifier
+     *
+     * <p>NOTE: cause paimon use dynamic partition all the time, Exists Partitions is not a table
+     * properties cached in memory, the method will cause io read every time it called.
+     *
+     * @param identifier path of the table
+     * @return a list of partitions of the specified table in identifier
+     * @throws TableNotExistException if the table does not exist
+     */
+    default List<Partition> listPartitions(Identifier identifier) throws TableNotExistException {
```
JingsongLi, line 200: Maybe we don't this method now, we can just expose `List<BinaryRow> TableScan.listPartitions`, the interface can `BinaryRow` here, because we just use `BinaryRow` in `TableWrite` too.

paimon-core/src/test/java/org/apache/paimon/catalog/FileSystemCatalogTest.java
```diff
@@ -1,47 +0,0 @@
...
-
-import org.junit.jupiter.api.BeforeEach;
-import org.junit.jupiter.api.Test;
-
-import java.util.List;
-
-import static org.assertj.core.api.Assertions.assertThat;
-
-/** Tests for {@link FileSystemCatalog}. */
-public class FileSystemCatalogTest extends CatalogTestBase {
```
JingsongLi, line 31: Why remove this?

paimon-core/src/main/java/org/apache/paimon/table/system/AuditLogTable.java
```diff
@@ -263,6 +274,11 @@ public Plan plan() {
             return streamScan.plan();
         }
 
+        @Override
+        public List<BinaryRow> partitions() {
```
JingsongLi, line 278: listPartitions?

paimon-core/src/test/java/org/apache/paimon/catalog/FileSystemCatalogTest.java
leaves12138: I moved it to Table Test. Cause we don't need the interface here, so "CatalogTest" seems not suitable. Something like TablePartitionTest seems more reasonable.

leaves12138: https://github.com/apache/incubator-paimon/pull/1122/files#diff-6a15f320a8c13b04de320f37b24def5c89ab55415d33777529b2e64a7ef368ce

JingsongLi (approved):
+1

PR #1124, opened 2023-05-10
[bug][core] Fix loser tree test

qidian99:
### Purpose
Fixed loser tree test case is not actually doing validation.
Fixed duplicate keys in readeres will cause misordering in loser tree.
### Tests
LoserTreeTest

### API and Format

No.

### Documentation

No.


qidian99:
Looks like it's already been fixed in #1116 . Will close this pr.

PR #1149, opened 2023-05-12
[core] support kryo serialization

YannByron:
### Purpose

to fix https://github.com/apache/incubator-paimon/issues/635


JingsongLi (commented):
Thanks @YannByron for the contribution, can you add test for this?

JingsongLi (approved):
+1

PR #1150, opened 2023-05-13
[doc][maintenance] - Updated use case bullet point

mans2singh:
[core]

### Purpose

[Updated use case section of maintenance](https://paimon.apache.org/docs/master/maintenance/rescale-bucket/#use-case)

### Tests

None

### API and Format

No

### Documentation

[maintenance](https://paimon.apache.org/docs/master/maintenance/rescale-bucket/#use-case)


JingsongLi (approved):
Thanks @mans2singh LGTM

mans2singh:
Thanks @JingsongLi for your time.

PR #1153, opened 2023-05-15
[test] Add test for ADD/DROP/MODIFY columns for Flink 1.17

### LINKED ISSUE #1152: [Feature] Add unit test for ALTER TABLE in Flink 1.17
### Search before asking

### Motivation

1. I ran Flink 1.17 cluster and tested out Paimon catalog DDL scenarios, and add/drop column statements suceeded without exceptions
```sql
ALTER TABLE my_table ADD (
    ts AS TO_TIMESTAMP(log_ts) AFTER log_ts
);
```
 
image

2. I noticed that SchemaChangeITCase missed the altering columns case, so I tried to add corresponding test cases to SchemaChangeITCase, and got an exception:

image

3. I noticed that in Flink 1.15 and 1.16，SqlToOperationConverter::convertAlterTable does not support add/drop columns

Altering table add/drop columns is a new feature introduced in Flink 1.17. Therefore, it's better to add tests in paimon-flink-1.17

### Solution

Add the corresponding tests that cover add column, drop column and modify column.

### Anything else?

_No response_

qidian99:
### Purpose

This PR resolves #1152 

### Tests

SchemaChangeITCase under paimon-flink-1.17 

### API and Format

No

### Documentation

In a future issue.

### DIFF
--- a/paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/SchemaChangeITCase.java
+++ b/paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/SchemaChangeITCase.java
@@ -428,4 +428,65 @@ public class SchemaChangeITCase extends CatalogITCaseBase {
                 .isInstanceOf(UnsupportedOperationException.class)
                 .hasMessage("Change 'sequence.field' is not supported yet.");
     }
+
+    @Test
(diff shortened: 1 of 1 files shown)


JingsongLi (approved):
+1

PR #1159, opened 2023-05-15
[test] Enable JUnit 5 Automatic Extension Detection

reswqa:
### Purpose

We can use Junit5's `SPI` mechanism to automatically introduce `TestLoggerExtension`.

### Tests

*Manually verified.*

### API and Format

*No change.*

### Documentation

*No need.*

### DIFF
--- a/paimon-benchmark/paimon-micro-benchmarks/pom.xml
+++ b/paimon-benchmark/paimon-micro-benchmarks/pom.xml
@@ -131,6 +131,15 @@ under the License.
             <artifactId>paimon-bundle</artifactId>
             <version>${project.version}</version>
         </dependency>
+
+        <!-- Test -->
+
+        <dependency>
+            <groupId>org.apache.paimon</groupId>
+            <artifactId>paimon-test-utils</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <profiles>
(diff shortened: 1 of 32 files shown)


reswqa:
Hi @yuzelin, would you mind taking a look at this? Thanks!

JingsongLi (commented):
I am not a expert of junit5... Why to modify the version of maven-surefire-plugin?

reswqa:
Thanks for the reply! 

> Why to modify the version of maven-surefire-plugin?
As mentioned in the first commit message: maven surefire version `3.x` provides better support regarding the `JUnit Platform`. Otherwise, Junt5 related configurations such as `<junit.jupiter.extensions.autodetection.enabled>` will not take effect. BTW, flink has also updated its version to `3.x` for this reason.

JingsongLi:
> Thanks for the reply @JingsongLi !
> 
> > Why to modify the version of maven-surefire-plugin?
> 
> As mentioned in the first commit message: maven surefire version `3.x` provides better support regarding the `JUnit Platform`. Otherwise, Junt5 related configurations such as `<junit.jupiter.extensions.autodetection.enabled>` will not take effect. BTW, flink has also updated its version to `3.x` for this reason.

Thanks for the explanation, can you also mention the PR in Flink?

reswqa:
https://junit.org/junit5/docs/current/user-guide/#running-tests-build-maven

reswqa:
Junit5's [doc](https://junit.org/junit5/docs/current/user-guide/#running-tests-build-maven) recommended to use version `3.0.0-M4` or later to avoid interoperability issues. I have also tested `2.22.1(The version we are currently using)`, which supports running junit5 tests but does not support `junit.jupiter.extensions.autodetection.enabled`.

> BTW, flink has also updated its version to 3.x for this reason.

Correction: I made a mistake here, flink updated the version of `maven-surefire-plugin` to resolve `Corrupted STDOUT` issues(FLINK-24983), not to support `Junit5`.

> can you also link the PR in Flink?
But it also enables JUnit 5 automatic extension detection(FLINK-26218). to avoid mark too many test classes with `TestLoggerExtension`.

JingsongLi (approved):
Thanks @reswqa ! Looks good to me!

PR #1160, opened 2023-05-16
[flink] Default enable scan.infer-parallelism

JingsongLi:
### Purpose

Considering that splits have already been scanned during the statistical information inference process, we can default to enabling parallelism inference.

Additionally, we do not need to pass the splits to the runtime, as there may be a risk of OOM when there are many splits.

### DIFF
--- a/docs/layouts/shortcodes/generated/flink_connector_configuration.html
+++ b/docs/layouts/shortcodes/generated/flink_connector_configuration.html
@@ -22,7 +22,7 @@
         </tr>
         <tr>
             <td><h5>scan.infer-parallelism</h5></td>
-            <td style="word-wrap: break-word;">false</td>
+            <td style="word-wrap: break-word;">true</td>
             <td>Boolean</td>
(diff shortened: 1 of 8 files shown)


paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/ReadWriteTableITCase.java
```diff
@@ -1146,7 +1146,12 @@ public void testSourceParallelism() throws Exception {
                 createTable(
                         Arrays.asList("currency STRING", "rate BIGINT"),
                         Collections.emptyList(),
-                        Collections.emptyList());
+                        Collections.emptyList(),
+                        new HashMap<String, String>() {
+                            {
+                                put(INFER_SCAN_PARALLELISM.key(), "false");
+                            }
+                        });
```
yuzelin, line 1154: `Collections.singletonMap`

yuzelin (approved):

PR #1165, opened 2023-05-17
[Improve] Support of tencent clound chdfs schema

hiliuxg:
### Purpose
Cloud HDFS (CHDFS) is a distributed file system of Tencent Cloud that provides standard HDFS access protocols, excellent performance, and hierarchical namespaces. CHDFS mainly solves massive data storage and data analysis in big data scenarios.

It's provides standard HDFS access protocols but file shema is ofs, so this pr offer a ChdfsFileIOLoader to support it.

chdfs product link : https://cloud.tencent.com/document/product/1105

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/fs/hadoop/ChdfsFileIOLoader.java
+++ b/paimon-common/src/main/java/org/apache/paimon/fs/hadoop/ChdfsFileIOLoader.java
@@ -22,7 +22,7 @@ import org.apache.paimon.fs.FileIOLoader;
 import org.apache.paimon.fs.Path;
 
 /**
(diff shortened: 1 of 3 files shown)


JingsongLi (commented):
Thanks @hiliuxg for the contribution.
See `FileIO.get`.
At present, if we cannot find the corresponding FileIO, we will use HadoopFileIO. Have you tested it? What are the current issues?

JingsongLi (commented):
Thanks @hiliuxg for the contribution.
See `FileIO.get`.
At present, if we cannot find the corresponding FileIO, we will use HadoopFileIO. Have you tested it? What are the current issues?

PR #1168, opened 2023-05-17
[bug] Enforce snapshot expire will not expire all snapshots

### LINKED ISSUE #1167: [Bug] Snapshot expire may have unexpected behavior when snapshot.num-retained.min = 0
### Search before asking

### Paimon version

0.5-SNAPSHOT

### Compute Engine

Flink

### Minimal reproduce step

Setting snapshot.num-retained.min=0 in table property will cause all snapshots being expired after the maximum retain time of completed snapshots.

snapshot.num-retained.min should be greater than or equal to 1, so that we always has at least 1 snapshot.

### Anything else?

_No response_

qidian99:
### Purpose

This PR closes #1167 
### Tests

No.
### API and Format

No.
### Documentation

No.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/FileStoreExpireImpl.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/FileStoreExpireImpl.java
@@ -22,6 +22,7 @@ import org.apache.paimon.Snapshot;
 import org.apache.paimon.annotation.VisibleForTesting;
 import org.apache.paimon.consumer.ConsumerManager;
 import org.apache.paimon.data.BinaryRow;
+import org.apache.paimon.utils.Preconditions;
 import org.apache.paimon.utils.SnapshotManager;
 
 import org.slf4j.Logger;
@@ -66,6 +67,12 @@ public class FileStoreExpireImpl implements FileStoreExpire {
             SnapshotManager snapshotManager,
             SnapshotDeletion snapshotDeletion,
             TagFileKeeper tagFileKeeper) {
+        Preconditions.checkArgument(
+                numRetainedMin >= 1,
+                "The minimum number of completed snapshots to retain should be >= 1.");
+        Preconditions.checkArgument(
+                numRetainedMax > numRetainedMin,
+                "The maximum number of snapshots to retain should be > the minimum number.");
         this.numRetainedMin = numRetainedMin;
         this.numRetainedMax = numRetainedMax;
         this.millisRetained = millisRetained;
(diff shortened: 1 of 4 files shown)


JingsongLi (commented):
See paimon-docs/README to rebuild docs.

paimon-core/src/main/java/org/apache/paimon/operation/FileStoreExpireImpl.java
```diff
@@ -88,6 +89,12 @@ public FileStoreExpireImpl(
             SnapshotManager snapshotManager,
             ManifestFile.Factory manifestFileFactory,
             ManifestList.Factory manifestListFactory) {
+        Preconditions.checkArgument(
+                numRetainedMin >= 1,
+                "The minimum number of completed snapshots to retain should be >= 1.");
+        Preconditions.checkArgument(
+                numRetainedMax > numRetainedMin,
+                "The maximum number of snapshots to retain should be > the minimum number.");
```
yuzelin, line 97: I think the limit should be `numRetainedMax >= numRetainedMin` because we can set `numRetainedMax == numRetainedMin` to control exactly how many snapshots are retained.

qidian99: Yes you are correct. I was confused by the following code in snapshot expiration.

```java
        // find the earliest snapshot to retain
        for (long id = Math.max(latestSnapshotId - numRetainedMax + 1, earliest);
                id <= latestSnapshotId - numRetainedMin;
                id++) {
            if (snapshotManager.snapshotExists(id)
                    && currentMillis - snapshotManager.snapshot(id).timeMillis()
                            <= millisRetained) {
                // within time threshold, can assume that all snapshots after it are also within
                // the threshold
                expireUntil(earliest, id);
                return;
            }
        }
```

This segment tries to locate the first snapshot within a range of snapshots, specifically between the numRetainedMax th and numRetainedMin th latest snapshots. This snapshot needs to be preserved because it doesn't fulfill the time threshold condition for expiration.

I've updated the comment to be more clear.

yuzelin (approved):

JingsongLi:
Test failed due to: 
```
ConfigOptionsDocsCompletenessITCase.testCompleteness:65->compareDocumentedAndExistingOptions:209 Documentation is outdated, please regenerate it according to the instructions in paimon-docs/README.md.
```

qidian99:
> Test failed due to:
> 
> ```
> ConfigOptionsDocsCompletenessITCase.testCompleteness:65->compareDocumentedAndExistingOptions:209 Documentation is outdated, please regenerate it according to the instructions in paimon-docs/README.md.
> ```

@JingsongLi Done generating docs.

JingsongLi:
Can you rebase latest master?

JingsongLi (approved):
+1

Issue #1174, opened 2023-05-19
[doc] explain  the `read.batch-size` option on the impact of memory consumption of  compaction in the Memory section of the write-performance page


wg1026688210:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/incubator-paimon/issues) and found nothing similar.

### Motivation

If compaction is enabled for writing and each row is very large, reading too many lines of data at once can consume a lot of memory. Reducing the `read.batch-size` option  can alleviate the impact of this issue.
We can introduce `read.batch-size`  in the write-performance page ,so that Users can easily understand the impact of this option during compaction.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #1175, opened 2023-05-19
Introduce how the read.batch-size option can impact memory consumption when compaction in write-performance page

wg1026688210:
related #1174


JingsongLi (approved):
+1

PR #1183, opened 2023-05-20
[flink] Replace org.apache.flink.util.function.SerializableFunction by paimon's implementation.

reswqa:
### Purpose

Replace `org.apache.flink.util.function.SerializableFunction` by paimon's implementation.

### Tests

Should be covered by existing tests already.

### API and Format

No changes.

### Documentation

No need.


JingsongLi (approved):
+1 Thanks @reswqa

PR #1190, opened 2023-05-22
[license] Add header to test java files

JingsongLi:
### Purpose

Even java files in resources, we need to add header to these files.

From https://lists.apache.org/thread/2fdsps5ggllo56oy61cp0gwb8spov1vf

- These files are missing ASF headers [3][4][5] and other files in that directory, and these directories [6][7][8][9][10] and other directories under [11]. Are they ASF files and how are they licensed?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -407,8 +407,6 @@ under the License.
                         <exclude>release/**</exclude>
                         <!-- antlr grammar files -->
                         <exclude>paimon-common/src/main/antlr4/**</exclude>
-                        <!-- Test files -->
-                        <exclude>paimon-common/src/test/resources/codesplit/**</exclude>
                     </excludes>
                 </configuration>
             </plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -398,7 +398,6 @@ under the License.
                         <exclude>docs/themes/book/**</exclude>
                         <exclude>docs/assets/github.css</exclude>
                         <exclude>docs/static/js/anchor.min.js</exclude>
-                        <exclude>docs/layouts/shortcodes/generated/**</exclude>
                         <exclude>**/*.svg</exclude>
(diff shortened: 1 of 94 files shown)


carp84 (changes requested):
It seems to me that in our [apache-rat plugin configuration](https://github.com/apache/incubator-paimon/blob/master/pom.xml#L411) (in the pom file) we have excluded all codesplit resources, where we are missing license headers and trying to fix in this PR. I suggest we carefully review our apache-rat configuration and do necessary change to make sure no more unexpected exclusions.

Note: files such as `paimon-common/src/test/resources/codesplit/groups` should be excluded indeed, and we need to carefully check and make sure that `mvn verify` has the precise coverage (no more no less).

pom.xml
carp84, line 1: The current changes LGTM, and I have confirmed `mvn verify` could pass.

Could you further make sure whether the below exclusions in the rat check configuration are as expected and already with necessary notice/license declaration?:
1. docs/assets/github.css (L399)
2. docs/static/js/anchor.min.js (L400)
3. docs/layouts/shortcodes/generated/** (L401)

And please make sure the exclusion of font-awesome (L395) is removed in PR#1192 if the dependency is removed there. Thanks.

JingsongLi: >1. docs/assets/github.css (L399)

Add license in https://github.com/apache/incubator-paimon/pull/1192

>2.docs/static/js/anchor.min.js (L400)

Already noticed in NOTICE and LICENSE

>3.docs/layouts/shortcodes/generated/** (L401)

Addressed, I generated in `paimon-doc`.

Very thanks Yu.

carp84 (approved):
The latest changes LGTM, and I've confirmed `mvn verify` could pass locally. Thanks for the efforts @JingsongLi

PR #1192, opened 2023-05-22
[license] Fix license problems

JingsongLi:
From https://lists.apache.org/thread/2fdsps5ggllo56oy61cp0gwb8spov1vf

- LICENSE is missing license for [1] (see [2]) and other files and the mavin wrapper. It would also be good if the LICENSE included more information i.e. what products and versions the 3rd party code is from.
- NOTICE is not correct. For instance, these NOTICE files [12][13] haven't been included correctly. It is unclear why Apache Flink and Apache Hadoop are mentioned; this information is missing from LICENSE.


NOTICE
```diff
@@ -10,18 +10,35 @@ Copyright 2014-2023 The Apache Software Foundation
 Apache Hadoop
 Copyright 2006 and onwards The Apache Software Foundation.
 
-Apache Hadoop
-Copyright 2006 and onwards The Apache Software Foundation.
+Flink Connector for Apache Doris
+Copyright 2018-2022 The Apache Software Foundation
```
SteNicholas, line 14:
```suggestion
Copyright 2018-2023 The Apache Software Foundation
```

carp84 (approved):
LGTM, +1

PR #1196, opened 2023-05-22
[core] Add Batch Table write-mode

leaves12138:
### Purpose

Add "Table" write mode


paimon-core/src/main/java/org/apache/paimon/WriteMode.java
```diff
@@ -32,7 +32,11 @@ public enum WriteMode implements DescribedEnum {
             "append-only",
             "The table can only accept append-only insert operations. Neither data deduplication nor any "
                     + "primary key constraints will be done when inserting rows into paimon."),
-    CHANGE_LOG("change-log", "The table can accept insert/delete/update operations.");
+    CHANGE_LOG("change-log", "The table can accept insert/delete/update operations."),
+    TABLE(
+            "table",
+            "The table without bucket.  Don't guarantee the order in the table. "
```
JingsongLi, line 38: We can have streaming read too. But just no order guarantee.
`This table is used in the traditional sense of "table". It can act like a Hive or Iceberg table. It can be written and read in a distributed manner without the concept of a bucket. It can be streaming read but no order guarantee.`

paimon-core/src/main/java/org/apache/paimon/table/BatchFileStoreTable.java
```diff
@@ -0,0 +1,73 @@
...
+
+/** {@link FileStoreTable} for {@link WriteMode#TABLE} write mode. */
+public class BatchFileStoreTable extends AppendOnlyFileStoreTable {
+
+    @VisibleForTesting private static final int batchTableBucket = 0;
+
+    BatchFileStoreTable(FileIO fileIO, Path path, TableSchema tableSchema) {
+        super(fileIO, path, tableSchema);
+        Preconditions.checkState(
+                tableSchema.options().get(CoreOptions.BUCKET_KEY.key()) == null,
```
JingsongLi, line 43: also check `bucket`?

paimon-core/src/main/java/org/apache/paimon/table/BatchFileStoreTable.java
```diff
@@ -0,0 +1,73 @@
...
+                "\"Table mode could not set properties 'bucket_key' and 'bucket'");
+    }
+
+    @Override
+    public TableWriteImpl<InternalRow> newWrite(
+            String commitUser, ManifestCacheFilter manifestFilter) {
+        AppendOnlyFileStoreWrite writer = store().newWrite(commitUser);
+        writer.withOverwrite(true);
```
JingsongLi, line 51: Why with overwrite?

paimon-core/src/main/java/org/apache/paimon/table/BatchFileStoreTable.java
```diff
@@ -0,0 +1,73 @@
...
+import org.apache.paimon.schema.TableSchema;
+import org.apache.paimon.table.sink.InternalRowKeyAndBucketExtractor;
+import org.apache.paimon.table.sink.TableWriteImpl;
+import org.apache.paimon.types.RowKind;
+import org.apache.paimon.utils.Preconditions;
+
+/** {@link FileStoreTable} for {@link WriteMode#TABLE} write mode. */
+public class BatchFileStoreTable extends AppendOnlyFileStoreTable {
+
+    @VisibleForTesting private static final int batchTableBucket = 0;
```
JingsongLi, line 38: checkstyle.
Please `mvn clean install -DskipTests` in your local machine.

paimon-core/src/main/java/org/apache/paimon/table/FileStoreTableFactory.java
```diff
@@ -83,6 +83,10 @@ public static FileStoreTable create(
         }
         if (writeMode == WriteMode.APPEND_ONLY) {
             table = new AppendOnlyFileStoreTable(fileIO, tablePath, tableSchema);
+        } else if (writeMode == WriteMode.TABLE) {
+            // BatchFileStoreTable need set write_only option to disable compaction
+            dynamicOptions.set(CoreOptions.WRITE_ONLY.key(), "true");
```
JingsongLi, line 88: I prefer disable compaction for `AppendOnlyFileStoreWrite`. Or just create a `BatchFileStoreWrite`.

paimon-core/src/main/java/org/apache/paimon/table/BatchFileStoreTable.java
leaves12138: private static final seems work.
but i changed it to:
public static final int BATCH_TABLE_BUCKET = 0;

paimon-core/src/main/java/org/apache/paimon/operation/BatchTableFileStoreWrite.java
```diff
@@ -0,0 +1,67 @@
...
+import org.apache.paimon.table.BatchFileStoreTable;
+import org.apache.paimon.types.RowType;
+import org.apache.paimon.utils.FileStorePathFactory;
+import org.apache.paimon.utils.SnapshotManager;
+
+import java.util.List;
+import java.util.concurrent.ExecutorService;
+
+/** Store write for {@link BatchFileStoreTable}. */
+public class BatchTableFileStoreWrite extends AppendOnlyFileStoreWrite {
```
JingsongLi, line 36: If just extends `AppendOnlyFileStoreWrite`, I think we can just use `AppendOnlyFileStoreWrite`.
```
AppendOnlyFileStoreWrite write = new AppendOnlyFileStoreWrite(...);
write.skipCompaction();
write.fromEmptyWriter() // renaming withOverwrite
return write;
```

paimon-core/src/main/java/org/apache/paimon/table/BatchFileStoreTable.java
JingsongLi: comment above.

leaves12138:
https://github.com/apache/incubator-paimon/pull/1256

PR #1206, opened 2023-05-23
[ci] Remove sonar check

JingsongLi:
### Purpose

One of the biggest benefits of sonar check is the ability to see test coverage, but our integrations have not been able to produce test coverage.

Another problem is the occasional sonar test error.

Remove this now.
If we are sure to add it, we will add the soar integration later.


SteNicholas (approved):
LGTM.

Issue #1207, opened 2023-05-23
[Bug] doc have a sql syntax question【Concepts - Append Only Table】


www2388258980:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/incubator-paimon/issues) and found nothing similar.

### Paimon version

paimon

### Compute Engine

flink

### Minimal reproduce step

Official website documentation:
 ```sql
CREATE TABLE T (
    `user` BIGINT,
    product STRING,
    order_time TIMESTAMP(3),
    WATERMARK FOR order_time AS order_time - INTERVAL '5' SECOND
) WITH (...);

-- launch a bounded streaming job to read paimon_table
SELECT window_start, window_end, SUM(f0) FROM
 TUMBLE(TABLE T, DESCRIPTOR(order_time), INTERVAL '10' MINUTES)) GROUP BY window_start, window_end;
```

it should be (https://nightlies.apache.org/flink/flink-docs-release-1.16/docs/dev/table/sql/queries/window-tvf/):
 ```sql
CREATE TABLE T (
    `user` BIGINT,
    product STRING,
    order_time TIMESTAMP(3),
    WATERMARK FOR order_time AS order_time - INTERVAL '5' SECOND
) WITH (...);

-- launch a bounded streaming job to read paimon_table
SELECT window_start, window_end, count(user) FROM TABLE(
 TUMBLE(TABLE T, DESCRIPTOR(order_time), INTERVAL '10' MINUTES)) GROUP BY window_start, window_end;
```

### What doesn't meet your expectations?

no

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

PR #1209, opened 2023-05-23
fix a sql syntax in doc

www2388258980:
**Purpose** 

Fix [#1207](https://github.com/apache/incubator-paimon/issues/1207)


JingsongLi (approved):
+1 Thanks @www2388258980

Issue #1181, opened 2023-05-19
[Feature] Support sync newly added table in CDC ingestion


qidian99:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/incubator-paimon/issues) and found nothing similar.

### Motivation

After CdcSyncDatabaseAction job start (for now we only sync a single database), if new table is created in MySQL, then action should sync the new mysql table to paimon. Specifically,

1. The new Paimon table is created upon receiving the create table DDL event in CDC source. Need to respect prefix/suffix.
2. CDC changelog of the newly added table should be successfully written to Paimon
3. Support multiple newly added tables.

### Solution

The draft design is as follows:
1. The new Paimon table is created upon receiving the create table DDL event in CDC source. Need to respect prefix/suffix.

Therefore we need to parse create table statement and utilize Paimon catalog to create the physical table

2. CDC changelog of the newly added table should be successfully written to Paimon

So we need another type of Paimon sink designed to write changelog from newly added tables to their corresponding Paimon tables. 

3. Support multiple newly added tables.

Specifically, the new Paimon sink needs to write the data from newly added tables discovered during runtime by managing multiple Paimon tables at the same time.

### Anything else?

To accomplish this feature, it's better to go by several steps:

1. PoC: implement a first draft to sync newly added table. 
2. Implement a multiplexed Paimon sink to support multi-table write
3. Implement a multiplexed Paimon committer to support multi-table commit
4. End-to-end test against newly added table scenario

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

qidian99:
There is a real-world use case as mentioned in #1106, so prioritize working on this issue.

PR #1212, opened 2023-05-23
[feature] Support multi-table write operator

qidian99:
### Purpose

This PR implement a multiplexed Paimon sink writer and committer to support multi-table write as mentioned in #1181.

### Tests

1. CdcRecordStoreMultiWriteOperatorTest: test blocking behaviors under schema updates for multiple tables
2. MultiTableCommittableSerializerTest: test serde of multi-table committables

### API and Format

Yes.

### Documentation

Yes.


paimon-core/src/main/java/org/apache/paimon/catalog/Identifier.java
```diff
@@ -109,4 +109,8 @@ public int hashCode() {
     public String toString() {
         return "Identifier{" + "database='" + database + '\'' + ", table='" + table + '\'' + '}';
     }
+
+    public String getTableName() {
```
JingsongLi, line 113: Put this together with `getObjectName`.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/MultiTableCommittable.java
```diff
@@ -0,0 +1,69 @@
...
+public class MultiTableCommittable extends Committable {
+
+    private final String database;
+    private final String table;
+    private final String commitUser;
+
+    public MultiTableCommittable(
+            String database,
+            String table,
+            String commitUser,
```
JingsongLi, line 37: Do we need this one? We can set same UUID user `StreamWriteBuilder.withCommitUser` to writer and committer.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/MultiTableCommittable.java
```diff
@@ -0,0 +1,69 @@
...
+package org.apache.paimon.flink.sink;
+
+import org.apache.paimon.catalog.Identifier;
+
+/**
+ * MultiTableCommittable produced by {@link PrepareCommitOperator}. This type of Committable will
+ * only be produced by multiplexed operators that handles data from multiple tables. Thus, the
+ * database, table, and commit user of each table is included in the committable.
+ */
+public class MultiTableCommittable extends Committable {
```
JingsongLi, line 28: Can we use just create a field `Committable` instead of extending?
Extending will make the caller think it can be used as a Committable.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/cdc/MultiplexCdcRecord.java
```diff
@@ -0,0 +1,53 @@
...
+
+import org.apache.paimon.types.RowKind;
+
+import java.util.Map;
+
+/**
+ * A data change message from the CDC source. Compared to {@link CdcRecord}, MultiplexCdcRecord
+ * contains database and table information.
+ */
+public class MultiplexCdcRecord extends CdcRecord {
```
JingsongLi, line 29: Can we use just create a field `CdcRecord ` instead of extending?
Extending will make the caller think it can be used as a CdcRecord.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/cdc/CdcRecordStoreMultiWriteOperator.java
```diff
@@ -0,0 +1,176 @@
...
+                        key,
+                        id -> {
+                            try {
+                                return (FileStoreTable) catalog.getTable(id);
+                            } catch (Catalog.TableNotExistException e) {
+                                return null;
+                            }
+                        });
+
+        if (table == null) {
```
JingsongLi, line 106: here should be a `while (true) {...}` to wait table creation by `UpdatedDataFieldsProcessFunction`.

JingsongLi (commented):
Thanks @qidian99 ! Good job. Left some comments.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/MultiTableCommittable.java
qidian99: The extending class is necessary here becase in the base case `PrepareCommitOperator`, we have the following signature:

```java
    protected abstract List<Committable> prepareCommit(boolean doCompaction, long checkpointId)
            throws IOException;
```

Therefore, the new operator should also produce a subclass of committable and use another custom serializer.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/cdc/MultiplexCdcRecord.java
qidian99: Yes a field is better than extending here.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/cdc/CdcRecordStoreMultiWriteOperator.java
qidian99: For MySQL CDC:

1. In snapshot phase, any DDL changes (including new table added) will result in Source table schema inconsistency and the job will fail, so we only consider ddl in incremental phase.

2. In incremental phase, it will only use 1 parallelism (source reader) to subscribe binlog, so the process function that create the new table in a blocking manner strictly `happen-before` this code segment.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/MultiTableCommittable.java
qidian99: Yes, it's possible. There are several places where commit user is used.

1. When the write operator is assembly into job graph, a random uid is set to be the commit user id.

2. When the job is deployed and running, it will try to recover the previous commit user from Flink state to make the commit user consistent accross different Flink job attempts.

3. When the CommitterOperator do commit, it will use create the committer with commit user.


So in this case we user the same commit user for all tables.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/MultiTableCommittable.java
JingsongLi: I think it is OK to bring generic type to `PrepareCommitOperator`.
`PrepareCommitOperator<IN, OUT>`

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/cdc/CdcRecordStoreMultiWriteOperator.java
JingsongLi: The process function is in a separate task, it is not a blocking manner.

qidian99: That makes sense...

Changed the code and added async table create test.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/MultiTableCommittable.java
qidian99: Done refactoring to PrepareCommitOperator<IN, OUT>. Please take a look.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/CommittableType.java
```diff
@@ -0,0 +1,57 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.paimon.flink.sink;
+
+/** CommittableType produced by {@link PrepareCommitOperator}. */
+public interface CommittableType {
```
JingsongLi, line 22: Do we need to introduce this class?

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/sink/cdc/CdcRecordStoreMultiWriteOperatorTest.java
```diff
@@ -0,0 +1,635 @@
...
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Predicate;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/** Tests for {@link CdcRecordStoreMultiWriteOperator}. */
+public class CdcRecordStoreMultiWriteOperatorTest {
```
JingsongLi, line 70: Can you add tests for `initializeState`? Just like `CommitterOperatorTest`.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/CommittableType.java
qidian99: No we don't :) Deleted

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/NoopCommittableStateManager.java
```diff
@@ -31,7 +31,7 @@
  * <p>Useful for committing optional snapshots. For example COMPACT snapshots produced by a separate
  * compact job.
  */
-public class NoopCommittableStateManager implements CommittableStateManager {
+public class NoopCommittableStateManager implements CommittableStateManager<ManifestCommittable> {
 
     @Override
     public void initializeState(StateInitializationContext context, Committer committer)
```
JingsongLi, line 37: Committer > Committer<....>

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/RestoreAndFailCommittableStateManager.java
```diff
@@ -41,7 +41,8 @@
  * <p>Useful for committing snapshots containing records. For example snapshots produced by table
  * store writers.
  */
-public class RestoreAndFailCommittableStateManager implements CommittableStateManager {
+public class RestoreAndFailCommittableStateManager
+        implements CommittableStateManager<ManifestCommittable> {
 
     private static final long serialVersionUID = 1L;
```
JingsongLi, line 47: Committer > Committer<....> all in this class.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkSink.java
```diff
@@ -142,7 +145,7 @@ public DataStreamSink<?> sinkFrom(
                 written.transform(
                                 GLOBAL_COMMITTER_NAME + " -> " + table.name(),
                                 typeInfo,
-                                new CommitterOperator(
+                                new CommitterOperator<CommitT, GlobalCommitT>(
```
JingsongLi, line 148: <> is ok

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/cdc/CdcRecordStoreMultiWriteOperator.java
```diff
@@ -0,0 +1,181 @@
...
+        catalog = catalogLoader.load();
+
+        // Each job can only have one user name and this name must be consistent across restarts.
+        // We cannot use job id as commit user name here because user may change job id by creating
+        // a savepoint, stop the job and then resume from savepoint.
+        commitUser =
+                StateUtils.getSingleValueFromState(
+                        context, "commit_user_state", String.class, initialCommitUser);
+
+        state = new StoreSinkWriteState(context, (tableName, partition, bucket) -> true);
```
JingsongLi, line 88: Here add a TODO, we should use `CdcRecordMultiChannelComputer` to filter.

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/sink/cdc/CdcRecordStoreMultiWriteOperatorTest.java
```diff
@@ -0,0 +1,696 @@
...
+        assertThat(actual).isEqualTo(expected);
+
+        runner.stop();
+        t.join();
+        harness.close();
+    }
+
+    @Test
+    @Timeout(30)
+    public void testInitializeState() throws Exception {
```
JingsongLi, line 215: I mean using `OneInputStreamOperatorTestHarness.initializeState` to test state recover for the operator.

JingsongLi (approved):
Looks good to me!

PR #1218, opened 2023-05-24
[flink] Should pass FlinkFileIOLoader when creating catalog in flink actions

yuzelin:
### Purpose

As tiltle.
This can fix problems like cannot access s3 in flink actions.

### Tests

orignin.
### API and Format

no
### Documentation

no


JingsongLi (approved):
+1

PR #1227, opened 2023-05-25
[doc] Introduce a page which documents small file number

### LINKED ISSUE #1132: [Feature] Introduce a page which documents small file number for paimon table
### Search before asking

### Motivation

At present, there may be checkpoint and snapshot expiration mechanisms for the paimon table. Users do not understand how the number of small files comes from. We need to add a page to explain and provide suggestions.

### Solution

_No response_

### Anything else?

_No response_

JingsongLi:
### Purpose

close #1132

### DIFF
--- a/docs/content/maintenance/manage-files.md
+++ b/docs/content/maintenance/manage-files.md
@@ -1,0 +1,97 @@
+---
+title: "Manage Files"
+weight: 5
+type: docs
+aliases:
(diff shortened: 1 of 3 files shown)


yuzelin (approved):
+1

leaves12138:
Turning up the checkpoint interval seems could reduce small files too. Do we need to add the compute engine level desmonstration?

PR #1230, opened 2023-05-25
[flink] Respect scan.parallelism whatever the case

JingsongLi:
### Purpose

Respect scan.parallelism, this is user-specified and should be the first priority

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/AbstractDataTableSource.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/AbstractDataTableSource.java
@@ -214,7 +214,7 @@ public abstract class AbstractDataTableSource extends FlinkTableSource
             FlinkSourceBuilder sourceBuilder, StreamExecutionEnvironment env) {
         Options options = Options.fromMap(this.table.options());
         Integer parallelism = options.get(FlinkConnectorOptions.SCAN_PARALLELISM);
-        if (options.get(FlinkConnectorOptions.INFER_SCAN_PARALLELISM)) {
+        if (parallelism == null && options.get(FlinkConnectorOptions.INFER_SCAN_PARALLELISM)) {
             if (streaming) {
                 parallelism = options.get(CoreOptions.BUCKET);
             } else {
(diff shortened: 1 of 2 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/AbstractDataTableSource.java
```diff
@@ -214,7 +214,7 @@ private DataStream<RowData> configureSource(
             FlinkSourceBuilder sourceBuilder, StreamExecutionEnvironment env) {
         Options options = Options.fromMap(this.table.options());
         Integer parallelism = options.get(FlinkConnectorOptions.SCAN_PARALLELISM);
-        if (options.get(FlinkConnectorOptions.INFER_SCAN_PARALLELISM)) {
+        if (parallelism == null && options.get(FlinkConnectorOptions.INFER_SCAN_PARALLELISM)) {
```
reswqa, line 217: It seems that we did not perform boundary check(e.g. `>0`) on `parallelism`. If `SCAN_PARALLELISM` is set to `-1`, `OperatorValidationUtils.validateParallelism` will not reject it because it is equal to `ExecutionConfig.PARALLELISM_DEFAULT`, which may cause some confusion. But it can always be argued that this is the behavior supported by `flink` (-1 is considered the default parallelism). Just a comment, maybe not do this check here is exactly the right one. 🤔

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/ReadWriteTableITCase.java
```diff
@@ -1213,10 +1213,10 @@ public void testInferParallelism() throws Exception {
                                         new HashMap<String, String>() {
                                             {
                                                 put(INFER_SCAN_PARALLELISM.key(), "true");
-                                                put(SCAN_PARALLELISM.key(), "-1");
```
reswqa, line 1216: Perhaps we should keep this case and assert that it will throw an `IllegalArgumentException`. But `-1` should be changed to another illegal value (such as `-2`), as explained in the previous comment.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/AbstractDataTableSource.java
JingsongLi: Yes, let this behavior explained by Flink.

reswqa (approved):
Thanks for the update @JingsongLi, +1 for the.

PR #1235, opened 2023-05-25
[flink] make sync action construct method to public

zhangjun0x01:
### Purpose

the[ paimon doc](https://paimon.apache.org/docs/master/how-to/cdc-ingestion/#synchronizing-tables) desc that we can use DataStream to use `MySqlSyncTableAction`,  but when we use the action in DataStream, it throw an exception 

`MySqlSyncTableAction(java.util.Map<java.lang.String,java.lang.String>, java.lang.String, java.lang.String, java.lang.String, java.util.List<java.lang.String>, java.util.List<java.lang.String>, java.util.Map<java.lang.String,java.lang.String>, ...)' is not public in 'org.apache.paimon.flink.action.cdc.mysql.MySqlSyncTableAction'. Cannot be accessed from outside package`

so I think we may need to make the construct method to public.


JingsongLi (approved):
+1

PR #1236, opened 2023-05-25
[doc] Remind users that beeline does not support hive write

Alibaba-HZY:
### Purpose

Remind users that beeline does not support hive write

### Tests

no

### API and Format

no

### Documentation

no


JingsongLi (approved):
+1

PR #1244, opened 2023-05-26
[flink] Use pipeline.name to action job name

JingsongLi:
### Purpose

Action job should respect user define `pipeline.name`.


yuzelin (approved):
+1

Issue #1191, opened 2023-05-22
[BUG]-Using hive to write paimon table resources is quite high


Dkbei:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/incubator-paimon/issues) and found nothing similar.

### Paimon version

The paimon hive-connector  version is[paimon-hive-connector-2.1-cdh-6.3-0.5-20230520.001837-16.jar](https://repository.apache.org/content/groups/snapshots/org/apache/paimon/paimon-hive-connector-2.1-cdh-6.3/0.5-SNAPSHOT/paimon-hive-connector-2.1-cdh-6.3-0.5-20230520.001837-16.jar)

### Compute Engine

hive-2.1.1-cdh-6.3.2

### Minimal reproduce step

1.Data from the hive table is read to the paimon table
2.The hive table data totals 3.8 million
3.The paimon table is partitioned according to create_time and contains data within three years. If data needs to be written to the paimon table successfully, set the memory size to 8 GB. In the same scenario, only 1.5 GB memory needs to be allocated for hive table writing
4.If the paimon table increases the number of buckets, more memory resources will be required.
5.The EXCLUSIVE lock is added when hive writes the paimon table. As a result, any query is blocked.

### What doesn't meet your expectations?

1. Reduce resource usage
2. Check whether hive writes data to the paimon table without blocking the paimon table query.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

Alibaba-HZY:
@JingsongLi  Please assign this issue to me.

PR #1246, opened 2023-05-27
[hive] Fix hive writer memory leak

Alibaba-HZY:
### Purpose

Fix hive writer memory leak #1191 
every map task  are not close writer ,it causes more memory required.
 

### Tests

no

### API and Format

no

### Documentation

no


JingsongLi (approved):
+1

PR #1247, opened 2023-05-27
[core] Support to make up the sequence number to nanosecond precision

### LINKED ISSUE #1050: [Feature] When records have same sequence number , the latter one is used as the final result.
### Search before asking

### Motivation

When consuming CDC  data, we use sequence.field to avoid disorder. However, since MySQL's execution time is at the level of seconds, there may be issues with the same ID and same execution time.

### Solution

add an auto-incrementing "inner-sequence-number" field in the SystemColumns. When the sequence numbers are the same, can compare the inner-sequence-numbers to determine the order.

### Anything else?

_No response_

schnappi17:
### Purpose

fix #1050 

Taking `System.nanotime()` as the time source with high precision, make up to the provided `sequence.field` to nanosecond if it's in`Timestamp` data type and with the precision of second or millisecond.

### Tests

ChangelogWithKeyFileStoreTableTest#testNanosSequenceNumberOnTimestampSecond
ChangelogWithKeyFileStoreTableTest#testNanosSequenceNumberOnTimestampMilliSecond
ChangelogWithKeyFileStoreTableTest#testNanosSequenceNumberOnTimestampMicroSecond
ChangelogWithKeyFileStoreTableTest#testNanosSequenceNumberOnNonTimestampField

### DIFF
--- a/docs/content/concepts/primary-key-table.md
+++ b/docs/content/concepts/primary-key-table.md
@@ -260,6 +260,8 @@ there will be some cases that lead to data disorder. At this time, you can use a
 {{< hint info >}}
 When the record is updated or deleted, the `sequence.field` must become larger and cannot remain unchanged. For example,
 you can use [Mysql Binlog operation time](https://ververica.github.io/flink-cdc-connectors/master/content/connectors/mysql-cdc.html#available-metadata) as `sequence.field`.
+If the provided `sequence.field` doesn't meet the precision, you can set `sequence.nanos` as `true` so that the precision of sequence number will be made up to nanosecond by system. 
+Note that only fields of data type `Timestamp` with precision of `second` and `millisecond` can be made up by nanoseconds.
 {{< /hint >}}
 
 {{< tabs "sequence.field" >}}
--- a/docs/content/concepts/primary-key-table.md
+++ b/docs/content/concepts/primary-key-table.md
@@ -260,8 +260,8 @@ there will be some cases that lead to data disorder. At this time, you can use a
 {{< hint info >}}
 When the record is updated or deleted, the `sequence.field` must become larger and cannot remain unchanged. For example,
 you can use [Mysql Binlog operation time](https://ververica.github.io/flink-cdc-connectors/master/content/connectors/mysql-cdc.html#available-metadata) as `sequence.field`.
-If the provided `sequence.field` doesn't meet the precision, you can set `sequence.nanos` as `true` so that the precision of sequence number will be made up to nanosecond by system. 
-Note that only fields of data type `Timestamp` with precision of `second` and `millisecond` can be made up by nanoseconds.
+If the provided `sequence.field` doesn't meet the precision, like a rough second or millisecond, you can set `sequence.auto-padding` to `second-to-micro` or `millis-to-micro` so that the precision of sequence number will be made up to microsecond by system. 
+Note that only fields of data type `Timestamp`, `Integer` and `BigInt` which indicates a `second` or `millisecond` can be padded by microseconds.
 {{< /hint >}}
 
 {{< tabs "sequence.field" >}}
--- a/docs/content/concepts/primary-key-table.md
+++ b/docs/content/concepts/primary-key-table.md
@@ -261,7 +261,6 @@ there will be some cases that lead to data disorder. At this time, you can use a
 When the record is updated or deleted, the `sequence.field` must become larger and cannot remain unchanged. For example,
 you can use [Mysql Binlog operation time](https://ververica.github.io/flink-cdc-connectors/master/content/connectors/mysql-cdc.html#available-metadata) as `sequence.field`.
 If the provided `sequence.field` doesn't meet the precision, like a rough second or millisecond, you can set `sequence.auto-padding` to `second-to-micro` or `millis-to-micro` so that the precision of sequence number will be made up to microsecond by system. 
-Note that only fields of data type `Timestamp`, `Integer` and `BigInt` which indicates a `second` or `millisecond` can be padded by microseconds.
 {{< /hint >}}
 
 {{< tabs "sequence.field" >}}
--- a/paimon-core/src/main/java/org/apache/paimon/table/sink/SequenceGenerator.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/sink/SequenceGenerator.java
@@ -19,6 +19,7 @@
 package org.apache.paimon.table.sink;
 
 import org.apache.paimon.data.InternalRow;
+import org.apache.paimon.data.Timestamp;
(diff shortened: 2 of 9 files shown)


JingsongLi (commented):
Can we just introduce something like:
1. sequence.auto-gen.second-micro: the user provides a long with precision to the second, and we give the complement to the micro.
2. sequence.auto-gen.mills-micro: the user provides a long with precision to the mills, and we give the complement to the micro.

This has the advantage that
1. it can be applied to all types
2. a long value can only be saved up to micro, not to nanos

JingsongLi (commented):
And please add some ut cases for `SequenceGenerator`.

JingsongLi:
Or we can introduce a 'sequence.auto-padding' key, valid values are 'second-to-micro' and 'mills-to-micro' and 'none'. Default is none.

schnappi17:
@JingsongLi Updated, thanks for you suggestions~

paimon-core/src/main/java/org/apache/paimon/table/sink/SequenceGenerator.java
```diff
@@ -68,6 +73,32 @@ public long generate(InternalRow row) {
         return generator.generate(row, index);
     }
 
+    public long generateWithPadding(InternalRow row, CoreOptions.SequenceAutoPadding autoPadding) {
```
JingsongLi, line 76: I think we can try to finish this in a simpler way:
```
public long generateWithPadding(InternalRow row, CoreOptions.SequenceAutoPadding autoPadding) {
    switch (autoPadding) {
            case SECOND_TO_MICRO:
                long value = generate(row); 
                // timestamp returns mills
                int second = fieldType.is(DataTypeFamily.TIMESTAMP) ? value / 1000 : value;
                return secondToMicro(second); 
            case MILLIS_TO_MICRO:
                return millsToMicro(generate(row)); 
            default:
                        "Unknown sequence padding mode " + autoPadding.name());
        }
}
```

paimon-core/src/main/java/org/apache/paimon/table/sink/TableWriteImpl.java
```diff
@@ -102,6 +103,20 @@ public SinkRecord writeAndReturn(InternalRow row) throws Exception {
         return record;
     }
 
+    @VisibleForTesting
+    public T writeAndReturnData(InternalRow row) throws Exception {
```
JingsongLi, line 107: why not using `writeAndReturn` and get row from `SinkRecord`?

docs/content/concepts/primary-key-table.md
```diff
@@ -260,6 +260,8 @@ there will be some cases that lead to data disorder. At this time, you can use a
 {{< hint info >}}
 When the record is updated or deleted, the `sequence.field` must become larger and cannot remain unchanged. For example,
 you can use [Mysql Binlog operation time](https://ververica.github.io/flink-cdc-connectors/master/content/connectors/mysql-cdc.html#available-metadata) as `sequence.field`.
+If the provided `sequence.field` doesn't meet the precision, like a rough second or millisecond, you can set `sequence.auto-padding` to `second-to-micro` or `millis-to-micro` so that the precision of sequence number will be made up to microsecond by system. 
+Note that only fields of data type `Timestamp`, `Integer` and `BigInt` which indicates a `second` or `millisecond` can be padded by microseconds.
```
JingsongLi, line 264: We can apply this padding to al types.

JingsongLi:
@schnappi17 Please ensure test passing first.

paimon-core/src/main/java/org/apache/paimon/table/sink/TableWriteImpl.java
```diff
@@ -102,6 +103,20 @@ public SinkRecord writeAndReturn(InternalRow row) throws Exception {
...
+        keyAndBucketExtractor.setRecord(row);
+        SinkRecord record =
```
JingsongLi, line 109: you can create a method: `toSinkRecord` to reuse code.

schnappi17:
@JingsongLi Updated, and I didn't see any test failed, maybe it's confused by another pr  of mine that UT failed, please help to approve the workflows and let's check it, thanks a lot!

paimon-core/src/test/java/org/apache/paimon/table/ChangelogWithKeyFileStoreTableTest.java
```diff
@@ -154,6 +161,219 @@ public void testSequenceNumber() throws Exception {
...
+                ts.getMillisecond(),
+                TimeUnit.MILLISECONDS.convert(sequenceNumber2, TimeUnit.MICROSECONDS));
+        commit.commit(0, write.prepareCommit(true, 0));
+        write.close();
+        String expectedResult;
+        if (sequenceNumber2 > sequenceNumber1) {
+            expectedResult =
+                    "1|10|101|1685530987|1685530987123|2023-05-23T11:22:33|2023-05-23T11:22:33.123|2023-05-23T11:22:33.123456|a2";
+        } else {
+            expectedResult =
```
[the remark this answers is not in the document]
schnappi17: > Based on the situation mentioned in #1050, a2 should always be the final result.

@lppsuixn In this situation, if you just set the `sequence.field` without auto-padding, the last record is the final result as expected.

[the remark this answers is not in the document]
JingsongLi: @lppsuixn In real world, the current implementation is possible because it is difficult to have two data with the same key in one microsecond at the same time.

JingsongLi (approved):
+1

hzjhjjyy:
In the current implementation, it's possible for the "padding" of later data's sequence to be smaller than that of the earlier data,resulting in data errors.
for example:
seconds->micros
currentNanoTime1:1062154,766334,200
currentSecondsTime1:1062154
currentNanoTime2:1062155,676231,100
currentSecondsTime2:1062155
sourceMillis:1691482327000
result_sequence:1691482327,766334 and 1691482327,676231(smaller)

Possible solutions?
set an incremental sequence for the sequence.field, incrementing by one for each incoming data?

lppsuixn:
> In the current implementation, it's possible for the "padding" of later data's sequence to be smaller than that of the earlier data,resulting in data errors. for example: seconds->micros currentNanoTime1:1062154,766334,200 currentSecondsTime1:1062154 currentNanoTime2:1062155,676231,100 currentSecondsTime2:1062155 sourceMillis:1691482327000 result_sequence:1691482327,766334 and 1691482327,676231(smaller)
> 
> Possible solutions? set an incremental sequence for the sequence.field, incrementing by one for each incoming data?

You are right. This PR does not solve the problem.

PR #1250, opened 2023-05-29
[flink] rm VisibleForTesting annotation from FlinkCatalog#catalog method

zhangjun0x01:
### Purpose

When we use flink api to develop some program, we need the `org.apache.paimon.table.Table` object, and we can use the follow code to get the `Table`  directly， it is very convenient，so I think we may remove the `VisibleForTesting` annotation from `FlinkCatalog#catalog` method

```
FlinkCatalog catalog = (FlinkCatalog) tenv.getCatalog(catalogName).get();
Table table = catalog.catalog().getTable(new Identifier(databaseName, tableName));
```


JingsongLi (approved):
+1

PR #1252, opened 2023-05-29
[docs] add docs: how to insert into a non-null column with nullable column

leaves12138:
### Purpose

With issue: how to insert into a non-null column with nullable column. One work around method.

see: https://github.com/apache/incubator-paimon/issues/1048


JingsongLi (approved):
+1

PR #1258, opened 2023-05-29
[flink] Support sync multiple database

s7monk:
fix https://github.com/apache/incubator-paimon/issues/1161

### Purpose

Support sync multiple database

### Tests

YES.

### API and Format

NO.

### Documentation

YES.


paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/action/cdc/mysql/MySqlSyncDatabaseActionITCase.java
```diff
@@ -479,6 +479,70 @@ private void testTableAffixImpl(Statement statement) throws Exception {
         waitForResult(expected, table2, rowType2, primaryKeys2);
     }
```
yuzelin, line 481: We need to test if the schema can be merged correctly.

yuzelin (commented):
Thanks for your contributing. Left some comments.

PR #1263, opened 2023-05-30
[flink] Fix kafka hybrid restore enumerator

JingsongLi:
### Purpose

Restore kafka source should not throw NPE exception.


yuzelin (approved):
LGTM.

PR #1264, opened 2023-05-31
[hive] Parse calcite to convert Chinese comments into unicode values whose encoding is UTF-16LE

### LINKED ISSUE #1262: [Bug] [hive] Field Chinese comments cannot be displayed normally when using hive to query table metadata
### Search before asking

### Paimon version

0.5-SNAPSHOT

### Compute Engine

hive

### Minimal reproduce step

1. Flink SQL uses hive metastore to create catalog.

2.Flink SQL creates a table with Chinese comments.

3.Hive SQL query this table `desc formatted table_name`

Field Chinese comments cannot be displayed normally when using hive to query table metadata

image

image

### Anything else?

_No response_

zhuangchong:
…

### Purpose

fix #1262 

### Tests

iShot_2023-05-30_19 03 51

paimon schema:

image

flink sql query;

image

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkCatalog.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkCatalog.java
@@ -28,6 +28,7 @@ import org.apache.paimon.table.Table;
 import org.apache.paimon.types.DataField;
 import org.apache.paimon.utils.Preconditions;
 
+import org.apache.commons.text.StringEscapeUtils;
 import org.apache.flink.table.api.Schema.UnresolvedColumn;
 import org.apache.flink.table.api.TableColumn;
 import org.apache.flink.table.api.TableSchema;
@@ -575,7 +576,17 @@ public class FlinkCatalog extends AbstractCatalog {
     private static Map<String, String> getColumnComments(CatalogTable catalogTable) {
         return catalogTable.getUnresolvedSchema().getColumns().stream()
                 .filter(c -> c.getComment().isPresent())
-                .collect(Collectors.toMap(UnresolvedColumn::getName, c -> c.getComment().get()));
+                .collect(
+                        Collectors.toMap(
+                                UnresolvedColumn::getName,
+                                c -> {
+                                    String comment = c.getComment().get();
+                                    if (comment.toUpperCase().startsWith("U&'")) {
+                                        comment = comment.substring(3).replaceAll("\\\\", "\\\\u");
+                                        comment = StringEscapeUtils.unescapeJava(comment);
+                                    }
+                                    return comment;
+                                }));
     }
 
     private static List<DataField> addColumnComments(
--- a/paimon-hive/paimon-hive-connector-common/src/test/java/org/apache/paimon/hive/HiveCatalogITCaseBase.java
+++ b/paimon-hive/paimon-hive-connector-common/src/test/java/org/apache/paimon/hive/HiveCatalogITCaseBase.java
@@ -50,11 +50,10 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.concurrent.Callable;
 import java.util.concurrent.atomic.AtomicInteger;
-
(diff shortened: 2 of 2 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkCatalog.java
```diff
@@ -575,7 +582,26 @@ public static Schema fromCatalogTable(CatalogTable catalogTable) {
...
                 .filter(c -> c.getComment().isPresent())
-                .collect(Collectors.toMap(UnresolvedColumn::getName, c -> c.getComment().get()));
+                .collect(
+                        Collectors.toMap(
+                                UnresolvedColumn::getName,
+                                c -> {
+                                    String comment = c.getComment().get();
+                                    if (comment.toUpperCase().startsWith("U&'")) {
+                                        try {
+                                            SqlParser parser = SqlParser.create(comment + "'");
```
JingsongLi, line 592: I am more worried about whether there is another way? The introduction of the calcite dependency looks a bit heavy, is it possible to just simply deal with it?

JingsongLi, line 592: We need test to cover chinese comments, for Flink, for Hive

wxplovecc: we use ``update_time` 	 timestamp 	 COMMENT _utf8'更新时间'` this type to define comment and replace `_utf8`

zhuangchong: @wxplovecc This method is good, but we can't force all users to write like this `... COMMENT _utf8'更新时间'`

wxplovecc: Right

zhuangchong: https://github.com/apache/incubator-paimon/blob/fbff232ac43c6e19079c0d0c373601a3a54871f2/paimon-hive/paimon-hive-connector-common/src/test/java/org/apache/paimon/hive/HiveCatalogITCaseBase.java#L247-L263

JingsongLi:
@wxplovecc raise a very good point, the memories I once had came back to me... I'm not sure if we have side effects from converting in `FlinkCatalog` like this. But `_utf8'更新时间'` looks a good solve method to this problem.

zhuangchong:
> @wxplovecc raise a very good point, the memories I once had came back to me... I'm not sure if we have side effects from converting in `FlinkCatalog` like this. But `_utf8'更新时间'` has problem too, "utf8" prefix...

I didn't think of a very good solution. At present, this solution has to judge whether to start with `U&'` to convert to minimize the impact.

JingsongLi:
involving @lincoln-lil , this seems a Flink bug.
We can fix it in paimon, but should be consistent with flink.

zhuangchong:
> 

I understand that this is not a bug of flink, because other engines (hive, spark...)do not support the analysis of calcite's UTF-16LE unicode

zhuangchong:
> involving @lincoln-lil , this seems a Flink bug. We can fix it in paimon, but should be consistent with flink.


I understand that this is not a bug of flink, because other engines (hive, spark...)do not support the analysis of calcite's UTF-16LE unicode.

lincoln-lil:
@JingsongLi thanks for pointing it out! I'll investigate it and try to get the solution.

lincoln-lil:
At first glance, when Flink interacts with CatalogTable, it directly passes the Calcite's NlsString comment as a string to the comment attribute of the schema and column. In theory, a Java string should be passed here. I will spend some time verifying the specific impact of the changes in Flink.

lincoln-lil:
> #1262
flink issue created https://issues.apache.org/jira/browse/FLINK-32249

zhuangchong:
Thanks @JingsongLi @lincoln-lil 

I will close this PR and wait for Flink to fix this issue.

Issue #1245, opened 2023-05-26
[Feature] support the full compaction for manifest file


wg1026688210:
## Search before asking

- [X] I searched in the [issues](https://github.com/apache/incubator-paimon/issues) and found nothing similar.

## Motivation

Due to the current limitation of manifest only supporting partial merging,  Delete type  ManifestEntrys cannot be applied to the ADD type  ManifestEntrys, resulting in the problem of data inflation in the ManifestEntry . 

This issue will  discuss the introduction of fullCompaction mechanism for  manifest to solve this problem.

## Solution

We divide the Manifest File into two layers: Base and Delta .
### The logic of Dividing layers
![image](https://github.com/apache/incubator-paimon/assets/14267759/d15c8f7c-c99e-4a35-b41d-ad21f75af897)

Assuming the target size for merging is 8MB.Firstly, we sequentially traverse the Manifest file List. When we find the first Manifest that contains a Delete Entry or has a file size smaller than the target size, this Manifest and all the following Manifests belong to the Delta layer, while the Manifests before this one belong to the Base layer.

![image](https://github.com/apache/incubator-paimon/assets/14267759/2dc67c2f-3d4e-437b-9abf-bc7df96d436e)

In this picture, we will see that :

- Base  (indicated by the orange box in the figure):  Include the Manifest file (hereinafter  `target ADD` ) only contains ADD type ManifestEntry, and the size of file is larger than the target size  , which don't need to be merged if there is no DELETE Entry to apply .

- Delta (indicated by the red box in the figure):  Include the Manifest file only contains ADD type ManifestEntry which file size is smaller than the target size or the file contain DELETE  type  ManifestEntry ,also possibly include the the `target ADD`.

### Apply  DELETE  To The Base Lay

![image](https://github.com/apache/incubator-paimon/assets/14267759/8b1ff349-6ac1-4eb9-b1be-c94de49b62cf)

Firstly, we merge the Delta layer to reduce the file size of the Delta layer. The remaining DELETE Entries after merging are the effective DELETE. Then, we calculate which partitions are affected by the DELETE. Next, we traverse the Base layer and find the first Manifest file that contains the affected partition. All the following Manifest files are split into a new layer called Base', which needs to be merged with the Delta layer to eliminate all DELETE files. The Base files other than the file of Base' will not participate in this FullCompaction process.

![image](https://github.com/apache/incubator-paimon/assets/14267759/9b5a8291-4ff1-4e84-8803-f47f3843b891)

[...]

PR #1268, opened 2023-05-31
[core] Skip more unwanted merges for Manifest full compaction

JingsongLi:
### Purpose

We can check the deleteEntries to skip base file which has not deletion.

This is useful for non-partition tables.

This is related to #1245

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/manifest/ManifestFileMeta.java
+++ b/paimon-core/src/main/java/org/apache/paimon/manifest/ManifestFileMeta.java
@@ -19,6 +19,7 @@
 package org.apache.paimon.manifest;
 
 import org.apache.paimon.data.BinaryRow;
+import org.apache.paimon.manifest.ManifestEntry.Identifier;
 import org.apache.paimon.predicate.Predicate;
 import org.apache.paimon.predicate.PredicateBuilder;
 import org.apache.paimon.stats.BinaryTableStats;
@@ -44,7 +45,8 @@ import java.util.stream.Collectors;
 
 /** Metadata of a manifest file. */
 public class ManifestFileMeta {
-    private static final Logger LOGGER = LoggerFactory.getLogger(ManifestFileMeta.class);
+
+    private static final Logger LOG = LoggerFactory.getLogger(ManifestFileMeta.class);
 
     private final String fileName;
     private final long fileSize;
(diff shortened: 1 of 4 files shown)


paimon-core/src/main/java/org/apache/paimon/manifest/ManifestFileMeta.java
```diff
@@ -280,31 +299,65 @@ public static List<ManifestFileMeta> tryFullCompaction(
...
-                // There is no DELETE Entry in Delta,Base don't need compaction
+                // There is no DELETE Entry in Delta, Base don't need compaction
                 j = base.size();
                 result.addAll(base);
             }
         }
 
-        Map<ManifestEntry.Identifier, ManifestEntry> fullMerged = new LinkedHashMap<>();
+        // 2.2. try to skip base files by reading entries
+
```
wg1026688210, line 309: Do we need to add a condition if (partitionType.getFieldCount() ==0)

paimon-core/src/main/java/org/apache/paimon/manifest/ManifestFileMeta.java
```diff
@@ -137,53 +139,68 @@ public String toString() {
...
+                // reach suggested file size, perform merging and produce new file
+                mergeCandidates(candidates, manifestFile, result, newMetas);
+                candidates.clear();
+                totalSize = 0;
+            }
+        }
+
+        // merge the last bit of manifests if there are too many
+        if (candidates.size() >= suggestedMinMetaCount) {
+            mergeCandidates(candidates, manifestFile, result, newMetas);
```
wg1026688210, line 200: Do we need to keep this logic after full compaction

paimon-core/src/main/java/org/apache/paimon/manifest/ManifestFileMeta.java
JingsongLi: I feel also do not need, now this to check the cost should not be high, just hashcode to compare

paimon-core/src/main/java/org/apache/paimon/manifest/ManifestFileMeta.java
JingsongLi: If we've done a full compaction, there shouldn't be any small files that need to be merged

PR #1275, opened 2023-06-01
[flink][bug] MySqlSyncDatabaseActionITCase doesn't need to handle incompatible table

Alibaba-HZY:
### Purpose
In `MySqlSyncDatabaseAction`, if user has specified `ignoreIncompatible = true`, and the table is not compatible, we should ignore it. However, current implementation still add the ignored table to `fileStoreTables`, then in `FlinkCdcSyncDatabaseSinkBuilder` a `DataStream<Void> schemaChangeProcessFunction` will be created, which is unnecessary.
So this PR remove the ignored table.

### Tests

Original tests.
### API and Format

no

### Documentation

no


yuzelin (approved):
+1

yuzelin:
Can you test that excluding table will not be created?

JingsongLi:
This should be a bug, but I cannot understand from title or PR description.

yuzelin:
Modify commit message and PR information for better history tracking.

PR #1279, opened 2023-06-01
[refactor][test] Extract duplicate code to util method in FileStoreExpireDeleteDirTest

yuzelin:
### Purpose

Refactor test class.


JingsongLi (approved):
+1

PR #1282, opened 2023-06-02
[flink] support sync multiple mysql database to paimon

### LINKED ISSUE #1238: [Feature] support multiple database sync for flink cdc sync action
### Search before asking

### Motivation

flink cdc support multiple database sync,I think we also support it.

for some databases with a large number but a small amount of data, we can use one job to complete data synchronization and reduce resources.

### Solution

_No response_

### Anything else?

_No response_

zhangjun0x01:
### Purpose

Linked issue: close #1238

Users can synchronize one or multiple MySQL database into one or multiple Paimon database,  It is default false, in this case, all the MySQL database will be synchronized to one paimon database, and the table with same name in different database will be merged, it is suitable for database sharding scenarios. if it is true, the parameter "--database" will be ignored, all the MySQL database will be synchronize to multiple paimon database with the same schema as MySQL before, it is suitable for scenarios with a large number of databases under a database instance, which can save resources.

for new added table in the sync database runtime , I need to get the `databaseName` and `tableName` ,otherwise the test case will be failed,  but the `druid` framework can not get the  `databaseName`, so I merge PR https://github.com/apache/incubator-paimon/pull/1621 to this PR, so we should review PR https://github.com/apache/incubator-paimon/pull/1621  first , when it is merged to master , I will rebase from the master .

### Tests

org.apache.paimon.flink.action.cdc.mysql.MySqlSyncDatabaseActionITCase

### DIFF
--- a/docs/content/how-to/cdc-ingestion.md
+++ b/docs/content/how-to/cdc-ingestion.md
@@ -101,7 +101,7 @@ the regular expressions.
 
 ### Synchronizing Databases
 
-By using [MySqlSyncDatabaseAction](/docs/{{< param Branch >}}/api/java/org/apache/paimon/flink/action/cdc/mysql/MySqlSyncDatabaseAction) in a Flink DataStream job or directly through `flink run`, users can synchronize the whole MySQL database into one Paimon database.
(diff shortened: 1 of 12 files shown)


JingsongLi:
Can you rebase latest master?

zhuangchong:

Issue #1217, opened 2023-05-24
[Bug] Batch write report error  when the second time


pongandnoon:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/incubator-paimon/issues) and found nothing similar.

### Paimon version

0.5-SNAPSHOT

### Compute Engine

Flink-1.17.0

### Minimal reproduce step

1. new a flink batch write task with the following parameters:

   parallelism.default=200
   execution.batch.adaptive.auto-parallelism.enabled=true ( the default value  of Flink-17 is true)

2. The first time ，the task runs successfully。the operator execution graph is：
<img width="833" alt="image" src="https://github.com/apache/incubator-paimon/assets/25953674/0de6f645-e060-44cd-90cc-1596b22c8df1">

3. The second time，the task report error：This is a bug! The partition org.apache.paimon.data.BinaryRow@23546dee and bucket 0 is filtered!

4. when we change  execution.batch.adaptive.auto-parallelism.enabled to false or remove  parallelism.default parameter, the task runs successfully

### What doesn't meet your expectations?

Caused by: java.lang.IllegalArgumentException: This is a bug! The partition org.apache.paimon.data.BinaryRow@23546dee and bucket 0 is filtered!
	at org.apache.paimon.utils.Preconditions.checkArgument(Preconditions.java:127)
	at org.apache.paimon.operation.AbstractFileStoreScan.withPartitionBucket(AbstractFileStoreScan.java:156)
	at org.apache.paimon.operation.AbstractFileStoreWrite.scanExistingFileMetas(AbstractFileStoreWrite.java:332)
	at org.apache.paimon.operation.AbstractFileStoreWrite.createWriterContainer(AbstractFileStoreWrite.java:314)
	at org.apache.paimon.operation.AbstractFileStoreWrite.lambda$getWriterWrapper$2(AbstractFileStoreWrite.java:289)
	at java.util.HashMap.computeIfAbsent(HashMap.java:1127)
	at org.apache.paimon.operation.AbstractFileStoreWrite.getWriterWrapper(AbstractFileStoreWrite.java:288)
	at org.apache.paimon.operation.AbstractFileStoreWrite.write(AbstractFileStoreWrite.java:98)
	at org.apache.paimon.table.sink.TableWriteImpl.writeAndReturn(TableWriteImpl.java:105)
	at org.apache.paimon.flink.sink.StoreSinkWriteImpl.write(StoreSinkWriteImpl.java:83)
[...]

PR #1199, opened 2023-05-22
[core] add FullCompaction for Manifest File

wg1026688210:
add FullCompaction for Manifest File

PR #1286, opened 2023-06-02
[core] Remove manifest cache load filter

JingsongLi:
### Purpose

Linked issue: related #1217

The load filter is designed to make the cache more efficient, which allows the initialization of the writer to make better use of the cache.

But this feature is rather hacky and code unfriendly, and it leads to the problem #1217

I'm thinking of deleting this feature, because currently the full-compaction of manifest has been merged, manifest will be much smaller. see #1199 and #1268

We can later consider using other means to improve cache utilization, such as sorting manifest by partition, so as to avoid reading all the manifest files when initializing the writer.


JingsongLi:
Please DO NOT merge this PR, this PR is waiting for production validation.

PR #1300, opened 2023-06-04
[core] Fix consuming from-snapshot-full on expired snapshot

### LINKED ISSUE #1299: [Bug] Specifying a expired snapshot with 'from-snapshot-full' mode will cause the job hanging
### Search before asking

### Paimon version

3094806bc5e3c7e28cab66fb4491188e955ae58e

### Compute Engine

flink 1.17

### Minimal reproduce step

1. create table with 'snapshot.time-retained'='2min'
2. insert into table
3. after 2min, scan the table in stremaing mode, specify the snapshot-id=1, and scan.mode='from-snapshot-full'

The scanning will hang, no scan result returned and no exceptions thrown.

However, switching to scan.mode='from-snapshot' works normally, the continuous scanned result returned.

### Anything else?

_No response_

schnappi17:
fix #1299 

Ceil up to earliest snapshot id if the specified snapshot id is expired in 'from-snapshot-full' scan mode, which is sync to the behavior as scan mode 'from-snapshot'.


schnappi17:
@JingsongLi Please help to review this pr if you have time, all workflows passed, thanks~

JingsongLi (approved):
+1

PR #1305, opened 2023-06-05
[bug] bugfix: packing return one empty list

leaves12138:
### Purpose

The packing method will return one empty list while binItems is empty and the next file weight is bigger than targetWeight.
e.g   The first item weight is 500, the target is 400, then this method will pack a redundant empty list.
We need to check binItems' size

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/utils/BinPacking.java
+++ b/paimon-common/src/main/java/org/apache/paimon/utils/BinPacking.java
@@ -39,7 +39,9 @@ public class BinPacking {
 
         for (T item : items) {
             long weight = weightFunc.apply(item);
-            if (binWeight + weight > targetWeight) {
+            // when get a much big item or total weight enough, we check the binItems size. If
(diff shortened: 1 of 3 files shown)


paimon-common/src/main/java/org/apache/paimon/utils/OrderedPacking.java
```diff
@@ -35,7 +35,9 @@ public static <T> List<List<T>> pack(
 
         for (T item : items) {
             long weight = weightFunc.apply(item);
-            if (binWeight + weight > targetWeight) {
+            // when get a much big item or total weight enough, we check the binItems size. If
+            // greater than zero, we pack it
+            if (binWeight + weight > targetWeight && binItems.size() > 0) {
```
JingsongLi, line 40: You can add a test for this.

leaves12138:
rebase master and create a new pull request, see https://github.com/apache/incubator-paimon/pull/1306

PR #1306, opened 2023-06-05
[bug] bugfix: packing return one empty list

leaves12138:
### Purpose

The packing method will return one empty list while binItems is empty and the next file weight is bigger than targetWeight.
e.g The first item weight is 500, the target is 400, then this method will pack a redundant empty list.
We need to check binItems' size


leaves12138:
rebase to master and close old one https://github.com/apache/incubator-paimon/pull/1305

PR #1307, opened 2023-06-05
[core] rename withOverwrite to fromEmptyWriter

leaves12138:
In writers, the overwrite flag does not overwrite actually. It just don't scan restored files, so empty writer is a more suitable name maybe. 

This pull request shift name withOverWrite to fromEmptyWriter.


JingsongLi (approved):
+1

PR #1308, opened 2023-06-05
[flink] Avoid FileNotFoundException in snapshot loop

### LINKED ISSUE #1302: [Feature] Avoid FileNotFoundException in snapshot loop
### Search before asking

### Motivation

In `GlobalFullCompactionSinkWrite.checkSuccessfulFullCompaction`, we have a loop to look at all snapshots.
But if there is committer to expire snapshots, there may be `FileNotFoundException`.

We can avoid this exception, if not found a snapshot file, check earliest snapshot in the loop.

### Solution

_No response_

### Anything else?

_No response_

JingsongLi:
### Purpose

Linked issue: close #1302

After modifying this to safely loop, we can remove limitation in documentation, users can set a free full compaction commits delta.

### Tests

`SnapshotManagerTest.testTraversalSnapshotsFromLatestSafely`.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/utils/SnapshotManager.java
+++ b/paimon-core/src/main/java/org/apache/paimon/utils/SnapshotManager.java
@@ -41,6 +41,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.function.BinaryOperator;
+import java.util.function.Function;
 import java.util.function.Predicate;
 
 import static org.apache.paimon.utils.FileUtils.listVersionedFiles;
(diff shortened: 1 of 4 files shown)


paimon-core/src/main/java/org/apache/paimon/utils/SnapshotManager.java
```diff
@@ -222,6 +223,46 @@ public Optional<Snapshot> latestSnapshotOfUser(String user) {
         return Optional.empty();
     }
 
+    /**
+     * Traversal snapshots from latest to earliest safely, this is applied on the writer side
+     * because the committer may delete obsolete snapshots, which may cause the writer to encounter
+     * unreadable snapshots.
+     */
+    public void traversalSnapshotsFromLatestSafely(Function<Snapshot, Boolean> consumer) {
```
yuzelin, line 231: How about using `java.util.function.Predicate` ?

JingsongLi: It should not be a predicate, it is a consumer, and just return a early break boolean.

yuzelin: Ok, make sense.

yuzelin (approved):
LGTM

PR #1309, opened 2023-06-05
[core] expose compaction flag out and generate single file rewriter for compaction

leaves12138:
1、Non-bucket compaction worker need single row data writer to execute compaction task
2、Need to expose skipCompaction out, because non-bucket compaction append-only table need set this while found itself is a non-bucket table


paimon-core/src/main/java/org/apache/paimon/operation/AppendOnlyFileStoreWrite.java
```diff
@@ -158,4 +159,39 @@ private AppendOnlyCompactManager.CompactRewriter compactRewriter(
             return rewriter.result();
         };
     }
+
+    // AppendOnlyCompactWorker need rewriter to execute compaction task
+    public AppendOnlyCompactManager.CompactRewriter singleCompactRewriter(
+            BinaryRow partition, int bucket) {
+        return toCompact -> {
```
JingsongLi, line 166: You can refactor `compactRewriter` to provide a `allowRolling` option.

JingsongLi:
We don't need to provide a non-rolling rewriter now.

PR #1315, opened 2023-06-06
[doc] Fix the sql statement that changes the position of the column in Flink SQL

zhuangchong:
### Purpose

Fix the sql statement that changes the position of the column in Flink SQL.
see https://nightlies.apache.org/flink/flink-docs-release-1.17/docs/dev/table/sql/alter/#modify
image


JingsongLi (approved):
+1

PR #1323, opened 2023-06-06
[spark] add TimestampNTZType for spark

zhangjun0x01:
### Purpose

add TimestampNTZType for spark

### Tests

org.apache.paimon.spark.SparkTypeTest

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/SparkTypeUtils.java
+++ b/paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/SparkTypeUtils.java
@@ -285,7 +285,8 @@ public class SparkTypeUtils {
                 return new CharType(((org.apache.spark.sql.types.CharType) atomic).length());
             } else if (atomic instanceof org.apache.spark.sql.types.DateType) {
                 return new DateType();
-            } else if (atomic instanceof org.apache.spark.sql.types.TimestampType) {
+            } else if (atomic instanceof org.apache.spark.sql.types.TimestampType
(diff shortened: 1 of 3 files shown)


paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/SparkTypeUtils.java
```diff
@@ -285,7 +285,8 @@ public org.apache.paimon.types.DataType atomic(DataType atomic) {
                 return new CharType(((org.apache.spark.sql.types.CharType) atomic).length());
             } else if (atomic instanceof org.apache.spark.sql.types.DateType) {
                 return new DateType();
-            } else if (atomic instanceof org.apache.spark.sql.types.TimestampType) {
+            } else if (atomic instanceof org.apache.spark.sql.types.TimestampType
+                    || atomic instanceof org.apache.spark.sql.types.TimestampNTZType) {
```
JingsongLi, line 289: `TimestampNTZType` is introduced in spark 3.4.
Here spark 3.1 3.2 3.3 will fail

zhangjun0x01: if we define a column with type `TIMESTAMP_NTZ` in spark 3.3 (or 3.2, 3.1) ,it will throw an exception by spark directly , It will not enter the `SparkTypeUtils.SparkToPaimonTypeVisitor#atomic` method,  if we define a column with `Timestamp` in spark 3.3 ,I test it , the method could deal it correctly.

PR #1327, opened 2023-06-07
[common] Make it more accurate to locate exception when loading FileIoLoader.

LsomeYeah:
### Purpose

Linked issue: close #xxx

[common] Make it more accurate to locate exception when loading FileIoLoader.

In FileIO#get, if an exception occurs when loading FileIOLoader, the cause of the exception should be traced back. The exception may be caused by LocalFileIO or HadoopFileIO.


paimon-common/src/main/java/org/apache/paimon/fs/FileIO.java
```diff
@@ -267,11 +279,20 @@ static FileIO get(Path path, CatalogContext config) throws IOException {
...
+                                    "Could not find a file io implementation for scheme '%s' in the classpath."
+                                            + "%s Hadoop FileSystem also cannot access this path '%s'.",
+                                    uri.getScheme(), fallbackMsg, path));
+
+            if (localFileIOException != null) {
+                ex.addSuppressed(localFileIOException);
+            }
+            if (hadoopIOException != null) {
+                ex.addSuppressed(hadoopIOException);
+            }
```
tsreaper, line 294: Save all suppressed exceptions in a `List`, so we don't have to check if each possible exception is null.

PR #1328, opened 2023-06-07
[bug][refactor][flink] All flink actions should parse catalog config

yuzelin:
### Purpose

Some flink actions hasn't parse catalog config, which cause S3 config cannot work. 
In addition, refactor some codes.

### Documentation

Add catalog-conf for the flink actions


JingsongLi (approved):
+1

PR #1329, opened 2023-06-07
[core] add scan type for non-bucket compaction

leaves12138:
Add Scan Type "non-bucket" which is used for non-bucket compaction.

When we assigned "non-bucket", we will use 'FullStartingUpScanner' and 'ContinuousAppendAndCompactScanner' to invoke streaming scan.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/source/AbstractInnerTableScan.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/source/AbstractInnerTableScan.java
@@ -65,14 +65,20 @@ public abstract class AbstractInnerTableScan implements InnerTableScan {
     }
 
     protected StartingScanner createStartingScanner(boolean isStreaming) {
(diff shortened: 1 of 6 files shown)


paimon-core/src/main/java/org/apache/paimon/table/source/AbstractInnerTableScan.java
```diff
@@ -65,10 +65,14 @@ public CoreOptions options() {
     }
 
     protected StartingScanner createStartingScanner(boolean isStreaming) {
-        if (options.toConfiguration().get(CoreOptions.STREAMING_COMPACT)) {
+        if (options.toConfiguration().get(CoreOptions.STREAMING_COMPACT)
```
JingsongLi, line 68: extract `options.toConfiguration().get(CoreOptions.STREAMING_COMPACT)`  to a field

JingsongLi (approved):
+1

PR #1331, opened 2023-06-08
[core] Introduce bucket mode to table and store

JingsongLi:
### Purpose

Introduce bucket mode:
```
/**
 * Bucket mode of the table, it affects the writing process and also affects the data skipping in
 * reading.
 */
public enum BucketMode {

    /**
     * The fixed number of buckets configured by the user can only be modified through offline
     * commands. The data is distributed to the corresponding buckets according to bucket key
     * (default is primary key), and the reading end can perform data skipping based on the
     * filtering conditions of the bucket key.
     */
    FIXED,

    /**
     * The Dynamic bucket mode records which bucket the key corresponds to through the index files.
     * This mode cannot support multiple concurrent writes or data skipping for reading filter
     * conditions. This mode only works for changelog table.
     */
    DYNAMIC,

    /**
     * Ignoring buckets can be equivalent to understanding that all data enters the global bucket,
     * and data is randomly written to the table. The data in the bucket has no order relationship
     * at all. This mode only works for append-only table.
     */
    UNAWARE
}
```


leaves12138:
+1

PR #1335, opened 2023-06-08
[core] Introduce IntHashSet and Int2ShortHashMap

JingsongLi:
### Purpose

They are used for dynamic bucket tables.


leaves12138:
+1

PR #1337, opened 2023-06-08
[core] add unaware-bucket process for 'AppendOnlyFileStoreTable', add test for unaware-bucket table check and compaction.

leaves12138:
1. Check schema when create non-bucket table.
2. Disable compaction when append-only table has unaware bucket
3. Add more tests for scan and schema check

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-core/src/main/java/org/apache/paimon/CoreOptions.java
@@ -1203,9 +1203,10 @@ public class CoreOptions implements Serializable {
 
     /** Compaction type when trigger a compaction action. */
     public enum StreamingCompactionType implements DescribedEnum {
-        NONE("none", "not a streaming compaction."),
-        NORMAL("normal", "Traditional bucket table compaction."),
-        NON_BUCKET("non-bucket", "Compaction for non-bucket table.");
+        NONE("none", "Not a streaming compaction."),
+        FIX("fix", "Compaction for fix bucket table."),
+        DYNAMIC("dynamic", "Compaction for dynamic bucket table."),
+        UNAWARE("unaware", "Compaction for unaware bucket table.");
 
         private final String value;
         private final String description;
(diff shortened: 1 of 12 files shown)


paimon-core/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -1203,9 +1203,10 @@ public static StreamingReadMode fromValue(String value) {
 
     /** Compaction type when trigger a compaction action. */
     public enum StreamingCompactionType implements DescribedEnum {
-        NONE("none", "not a streaming compaction."),
-        NORMAL("normal", "Traditional bucket table compaction."),
-        NON_BUCKET("non-bucket", "Compaction for non-bucket table.");
```
JingsongLi, line 1208: NONE("none", "not a streaming compaction."),
NORMAL("normal", "Traditional bucket table compaction."),
BUCKET_UNAWARE("non-bucket", "Compaction for non-bucket table.");

paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
```diff
@@ -144,6 +145,13 @@ public static void validateTableSchema(TableSchema schema) {
                             + "still want to keep the primary key definition.");
         }
 
+        if (options.bucket() == -1
+                && (options.toMap().get(BUCKET_KEY.key()) != null
+                        || !schema.primaryKeys().isEmpty())) {
```
JingsongLi, line 150: primaryKeys can be set for dynamic bucket mode.

paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
```diff
@@ -144,6 +145,13 @@ public static void validateTableSchema(TableSchema schema) {
...
+            throw new RuntimeException(
+                    "Cannot define 'bucket-key' or 'primary-key' in non-bucket mode.");
```
JingsongLi, line 152: improve this.

paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
```diff
@@ -144,6 +145,12 @@ public static void validateTableSchema(TableSchema schema) {
                             + "still want to keep the primary key definition.");
         }
 
+        if (options.bucket() == -1
+                && schema.primaryKeys().isEmpty()
+                && options.toMap().get(BUCKET_KEY.key()) != null) {
+            throw new RuntimeException("Cannot define 'bucket-key' in unaware bucket mode.");
```
JingsongLi, line 151: remove  `schema.primaryKeys().isEmpty()` and here `in unaware or dynamic bucket mode`.

PR #1340, opened 2023-06-08
[core] Extract RowKeyExtractor

JingsongLi:
### Purpose

For `BucketMode`, we need multiple bucket extractor implementations.


leaves12138:
+1

PR #1344, opened 2023-06-08
[core] Introduce index metadata

JingsongLi:
### Purpose

The index file is partition + bucket granularity.

This is the pre work of dynamic buckets, which allows for saving which keys are contained in a bucket in dynamic bucket mode, in order to locate the bucket of the key.


leaves12138:
+1

PR #1348, opened 2023-06-09
[core] Introduce consumer expiration

yuzelin:
### Purpose

Consumer file will prevent snapshot expiration. This PR introduce consumer expiration options to manage lifetime of consumer files.

### Tests

`ConsumerManagerTest#testExpire`
`ConsumerExpireTest`

### API and Format

new core options:
`consumer.expiration-time`
`consumer.expiration-check-interval`

### Documentation
`Consumer ID`


docs/layouts/shortcodes/generated/core_configuration.html
```diff
@@ -98,6 +98,18 @@
             <td>String</td>
             <td>Consumer id for recording the offset of consumption in the storage.</td>
         </tr>
+        <tr>
+            <td><h5>consumer.expiration-check-interval</h5></td>
```
JingsongLi, line 102: we can check every time. remove this option.

paimon-core/src/main/java/org/apache/paimon/consumer/ConsumerManager.java
```diff
@@ -90,6 +92,34 @@ public OptionalLong minNextSnapshot() {
...
+
+            if (statuses == null) {
+                throw new RuntimeException(
+                        String.format(
+                                "The return value is null of the listStatus for the '%s' directory.",
+                                directory));
+            }
+
+            for (FileStatus status : statuses) {
+                LocalDateTime modificationTime =
```
JingsongLi, line 112: if (isConsumerFile(status.getPath().name()))

paimon-core/src/main/java/org/apache/paimon/operation/ConsumerExpire.java
```diff
@@ -0,0 +1,59 @@
...
+
+import org.apache.paimon.annotation.VisibleForTesting;
+import org.apache.paimon.consumer.Consumer;
+import org.apache.paimon.consumer.ConsumerManager;
+
+import java.time.Duration;
+import java.time.LocalDateTime;
+
+/** Expire {@link Consumer}s. */
+public class ConsumerExpire {
```
JingsongLi, line 29: we can remove this class.

JingsongLi (approved):
+1

PR #1350, opened 2023-06-09
[format] Supports estimation of data size in memory for better control of file size.

### LINKED ISSUE #1210: HdfsFileSizeLimitation_Not_effective
### Search before asking

### Paimon version

image
Flink1.16Version——IcebergData BatchProcessing in Paimon

image
Flink1.17Version——Add Limit file size parameter
image

### Compute Engine

Flink

### Minimal reproduce step

After adding the file size limit parameter，The data file still exceeds 128mb。
image

image

Because the block size set by my HDFS is 128mb，If it exceeds 128mb, hdfs query data is not disk friendly, so the maximum file size should not exceed 128mb。

### Anything else?

_No response_

liming30:
### Purpose

Supports estimation of data size in memory for better control of file size.

**NOTE**: The calculation result is an estimated value, so there is no strong limit on the file size.

Linked issue: close #1210 

### Tests

Extended the format type in `org.apache.paimon.io.RollingFileWriterTest` to support testing for `avro/orc/parquet`.

### API and Format

No.

### Documentation

No.


paimon-format/src/main/java/org/apache/paimon/format/orc/writer/OrcBulkWriter.java
```diff
@@ -67,4 +89,43 @@ public void finish() throws IOException {
         flush();
         writer.close();
     }
+
+    @Override
+    public long length() throws IOException {
+        long estimateMemory = treeWriter.estimateMemory();
+
+        long dataLength = 0;
```
JingsongLi, line 97: Do we really need to get this from stripes? Can we just get pos from output stream?
All stripes should have be flushed? right?

paimon-format/src/main/java/org/apache/paimon/format/orc/writer/OrcBulkWriter.java
```diff
@@ -67,4 +89,43 @@ public void finish() throws IOException {
...
+                    stripeInformation != null
+                            ? stripeInformation.getOffset() + stripeInformation.getLength()
+                            : 0;
+        }
+
+        // This value is estimated, not actual.
+        return (long)
+                Math.ceil(
+                        dataLength
+                                + (estimateMemory + (long) rowBatch.size * avgRowByteSize) * 0.2);
```
JingsongLi, line 112: By default, rowBatch is 1000, and the rolling is required every 1000 records, my question is: can we check rolling during flushBatch?

So the `avgRowByteSize` can be removed.

paimon-format/src/main/java/org/apache/paimon/format/orc/writer/OrcBulkWriter.java
liming30: This is mainly the in-memory incompletely generated stripe size, the default value is 64MB or 256MB.

JingsongLi: Thanks, I get it.

paimon-format/src/main/java/org/apache/paimon/format/orc/writer/OrcBulkWriter.java
JingsongLi: It is OK to let it be 1024.

liming30: @JingsongLi  I have removed `avgRowByteSize`  and make the check interval equal with the size of `rowBatch`. Please help to review again.

paimon-common/src/main/java/org/apache/paimon/format/FormatWriter.java
```diff
@@ -60,4 +60,27 @@ public interface FormatWriter {
...
+    /**
+     * Check if the writer has reached the <code>targetSize</code>.
+     *
+     * @param suggestedCheck Whether it needs to be checked, but subclasses can also decide whether
+     *     to check it themselves.
+     * @param targetSize The size of the target.
+     * @return true if the target size was reached, otherwise false.
+     * @throws IOException Thrown if calculating the length fails.
+     */
+    default boolean reachTargetSize(boolean suggestedCheck, long targetSize) throws IOException {
```
JingsongLi, line 83: It is good to not introduce a default method which is invoked by per record.
I remember the default method's performance wasn't that good.

paimon-core/src/main/java/org/apache/paimon/io/SingleFileWriter.java
```diff
@@ -114,12 +114,16 @@ public long recordCount() {
         return recordCount;
     }
 
+    public boolean reachTargetSize(boolean suggestedCheck, long targetSize) throws IOException {
+        return writer.reachTargetSize(suggestedCheck, targetSize);
+    }
+
     @Override
     public long length() throws IOException {
```
JingsongLi, line 122: Maybe we can remove this method.
Only rolling will invoke this method.

paimon-common/src/main/java/org/apache/paimon/format/FormatWriter.java
```diff
@@ -60,4 +60,27 @@ public interface FormatWriter {
...
+
+    /**
+     * Returns the number of bytes that were currently written by this writer.
+     *
+     * <p>NOTE: This is an estimated value.
+     *
+     * @return the number of written bytes
+     * @throws IOException Thrown if calculating the length fails.
+     */
+    long length() throws IOException;
```
JingsongLi, line 72: We don't need to introduce this method.

paimon-format/src/main/java/org/apache/paimon/format/orc/writer/OrcBulkWriter.java
```diff
@@ -67,4 +80,35 @@ public void finish() throws IOException {
...
         writer.close();
     }
+
+    @Override
+    public long length() throws IOException {
+        long estimateMemory = treeWriter.estimateMemory();
+        long fileLength = underlyingStream.getPos();
+
+        // This value is estimated, not actual.
+        return (long) Math.ceil(fileLength + estimateMemory * 0.2);
```
JingsongLi, line 90: We don't need to read `stripes` now?

paimon-common/src/main/java/org/apache/paimon/format/FormatWriter.java
liming30: Removed the default method.

paimon-core/src/main/java/org/apache/paimon/io/SingleFileWriter.java
liming30: removed.

paimon-common/src/main/java/org/apache/paimon/format/FormatWriter.java
liming30: Removed and moved implementation into reachTargetSize.

paimon-format/src/main/java/org/apache/paimon/format/orc/writer/OrcBulkWriter.java
liming30: For `stripes` that have been flushed, the size can be obtained through `underlyingStream.getPos()`, and for the stripe that is not fully formed, the size can be obtained through `estimateMemory()`.

JingsongLi: I see

JingsongLi (commented):
The PR looks good to me now~ Thanks @liming30 

But I'm not entirely sure if it's so perfect. Can you test it? How large can a file be restricted now. The previous result was 180MB to 220MB (target-file-size is 128MB).

liming30:
>  How large can a file be restricted now. The previous result was 180MB to 220MB (target-file-size is 128MB).

I wrote a simple test: use `RollingFileWriter` to continuously write 10 million data of type `<Int, Varchar(32), Binary(16), BigInt>`.  

For `parquet` and `orc` format, both the old and new `rollingFile` methods are very close to the `targetFileSize`.  But for the `ORC` format, since the size of the `stripe` in the memory will greatly affect the final size of the file, this is very dependent on the `compression` effect, and since our test data is random, the compression effect is not good.

| format | old | new |
|--------|--------|--------|
| ORC | 181.4 MB | 152.4 MB |
| PARQUET | 128.03 MB | 128.03MB |
| AVRO | 128.1 MB | 128.1 MB |

JingsongLi:
@liming30 Thanks!

JingsongLi (approved):
+1

PR #1351, opened 2023-06-10
[Flink] Avoid unexpeted Num of  L0 File

wg1026688210:
When writing Paimon in Flink, Even if compaction fails, there will still be new L0 files commited successfully. The motivation behind this PR is to reduce the number of L0 files .

Waiting for the merge to complete during preCommit if there are too many Sorted Runs.

### DIFF
--- a/paimon-core/pom.xml
+++ b/paimon-core/pom.xml
@@ -162,6 +162,14 @@ under the License.
         </dependency>
 
         <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-inline</artifactId>
+            <version>${mockito.version}</version>
+            <type>jar</type>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
             <groupId>org.apache.hive</groupId>
             <artifactId>hive-storage-api</artifactId>
             <version>2.6.0</version>
--- a/paimon-core/pom.xml
+++ b/paimon-core/pom.xml
@@ -162,14 +162,6 @@ under the License.
         </dependency>
 
         <dependency>
-            <groupId>org.mockito</groupId>
-            <artifactId>mockito-inline</artifactId>
-            <version>${mockito.version}</version>
-            <type>jar</type>
-            <scope>test</scope>
-        </dependency>
-
-        <dependency>
             <groupId>org.apache.hive</groupId>
             <artifactId>hive-storage-api</artifactId>
             <version>2.6.0</version>
--- a/paimon-core/src/main/java/org/apache/paimon/compact/CompactFutureManager.java
+++ b/paimon-core/src/main/java/org/apache/paimon/compact/CompactFutureManager.java
@@ -18,6 +18,9 @@
 
 package org.apache.paimon.compact;
 
+import org.apache.paimon.annotation.VisibleForTesting;
+import org.apache.paimon.utils.ExceptionUtils;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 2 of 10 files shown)


paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreWrite.java
```diff
@@ -272,6 +272,13 @@ public void restore(List<State> states) {
                             state.commitIncrement,
                             compactExecutor());
             notifyNewWriter(writer);
+
+            try {
```
JingsongLi, line 276: revert here, restore is just for schema change, we don't need to create a compaction.

paimon-core/src/main/java/org/apache/paimon/mergetree/MergeTreeWriter.java
```diff
@@ -227,7 +227,8 @@ private void flushWriteBuffer(boolean waitForLatestCompaction, boolean forcedFul
     @Override
     public CommitIncrement prepareCommit(boolean waitCompaction) throws Exception {
         flushWriteBuffer(waitCompaction, false);
-        trySyncLatestCompaction(waitCompaction || commitForceCompact);
+        trySyncLatestCompaction(
+                waitCompaction || commitForceCompact || compactManager.shouldWaitCompaction());
```
JingsongLi, line 231: Good point!

wg1026688210:
fix #1391

wg1026688210:
@JingsongLi cc

paimon-core/src/main/java/org/apache/paimon/compact/CompactFutureManager.java
```diff
@@ -48,15 +51,24 @@ protected final Optional<CompactResult> innerGetCompactionResult(boolean blockin
             if (blocking || taskFuture.isDone()) {
                 CompactResult result;
                 try {
-                    result = taskFuture.get();
+                    result = obtainCompactResult();
                 } catch (CancellationException e) {
                     taskFuture = null;
                     return Optional.empty();
+                } catch (ExecutionException e) {
```
JingsongLi, line 58: We don't need to copy exception here.
We can just:
```
             try {
                  return Optional.of(taskFuture.get());
              } catch (CancellationException e) {
                  return Optional.empty();
              } finally {
                  taskFuture = null;
              }
```

paimon-core/pom.xml
```diff
@@ -161,6 +161,14 @@ under the License.
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.mockito</groupId>
```
JingsongLi, line 165: It is better to not introduce mockito. Mockito is not good test framework, we should not introduce it until it is absolutely necessary, we introduce our own abstraction.

paimon-core/src/main/java/org/apache/paimon/compact/CompactManager.java
```diff
@@ -31,6 +31,8 @@ public interface CompactManager extends Closeable {
     /** Should wait compaction finish. */
     boolean shouldWaitCompaction();
```
JingsongLi, line 32: `boolean shouldWaitForLatestCompaction`

paimon-core/src/main/java/org/apache/paimon/compact/CompactManager.java
```diff
@@ -31,6 +31,8 @@ public interface CompactManager extends Closeable {
...
 
+    boolean shouldWaitWhenPreparingCheckpoint();
```
JingsongLi, line 34: `boolean shouldWaitForPrepareCommit`

paimon-core/src/main/java/org/apache/paimon/mergetree/compact/MergeTreeCompactManager.java
```diff
@@ -75,6 +75,15 @@ public boolean shouldWaitCompaction() {
         return levels.numberOfSortedRuns() > numSortedRunStopTrigger;
     }
 
+    @Override
+    public boolean shouldWaitWhenPreparingCheckpoint() {
+        return levels.numberOfSortedRuns() > numSortedRunStopTrigger + 1;
+    }
+
+    private boolean checkSortedRuns(int extra) {
```
JingsongLi, line 83: remove this.

JingsongLi (approved):
+1

wg1026688210:
related #1375

PR #1353, opened 2023-06-11
[pulsar][Ecosystem] Support Apache Pulsar in External Log System

golden-yang:
### Purpose

Linked issue: [issue 765](https://github.com/apache/incubator-paimon/issues/765)

### DIFF
--- a/paimon-flink/paimon-flink-common/pom.xml
+++ b/paimon-flink/paimon-flink-common/pom.xml
@@ -35,6 +35,7 @@ under the License.
 
     <properties>
         <flink.version>1.17.1</flink.version>
+        <flink.pulsar.version>4.0.0-1.17</flink.pulsar.version>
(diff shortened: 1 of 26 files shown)


JingsongLi:
Sorry for the late reply, I am not very confident in the current LogSystem's role. It seems that not many people use it.

Can Pulsar integrate in a different way, such as integration with CDC data, just like Kafka? This scenario is clear.
https://paimon.apache.org/docs/master/how-to/cdc-ingestion/#kafka

golden-yang:
> Sorry for the late reply, I am not very confident in the current LogSystem's role. It seems that not many people use it.
> 
> Can Pulsar integrate in a different way, such as integration with CDC data, just like Kafka? This scenario is clear. https://paimon.apache.org/docs/master/how-to/cdc-ingestion/#kafka

OK. That's a good suggestion!

PR #1354, opened 2023-06-12
[core] Introduce IndexMaintainer to FileStoreWrite and HashBucketAssigner

JingsongLi:
### Purpose

Prepare classes for dynamic bucket for paimon-core


leaves12138:
+1

PR #1357, opened 2023-06-12
[flink] Introduce Dynamic bucket sink to Flink

JingsongLi:
### Purpose

A bucket is the smallest storage unit for reads and writes, each bucket directory contains an [LSM tree].

Primary Key Table supports two bucket mode:
1. Fixed Bucket mode: configure a bucket greater than 0, rescaling buckets can only be done through offline processes, 
   see [Rescale Bucket]. A too large number of buckets leads to too many
   small files, and a too small number of buckets leads to poor write performance.
2. Dynamic Bucket mode: configure `'bucket' = '-1'`, Paimon dynamically maintains the index, keeping the data volume
   in the bucket below `'dynamic-bucket.target-row-num'`. (This is an experimental feature)


leaves12138:
+1

PR #1360, opened 2023-06-13
[format] add orc push down test case 

zhangjun0x01:
### Purpose

add test for orc push down 

### Tests

- org.apache.paimon.format.orc.filter.OrcFilterConverterTest
- OrcFormatFilterPushdownITCase


JingsongLi (changes requested):
Thanks @zhangjun0x01 

I'm not sure if ITCase is valid, because SQL always adds the corresponding filter after source. How do you determine if SQL filters out or our filter pushdown filters out the data?

zhangjun0x01:
> Thanks @zhangjun0x01
> 
> I'm not sure if ITCase is valid, because SQL always adds the corresponding filter after source. How do you determine if SQL filters out or our filter pushdown filters out the data?


A perfect test case should first check the correctness of `OrcReaderFactory#connunctPredicates` during SQL execution, but I couldn't find a method to get `OrcReaderFactory#connunctPredicates` during program execution. So I divided this process and verified it by different test cases, such as  `flink filter ->paimon predicate`, `paimon predicate ->orc predicate`, and  to verify whether the query results are correct.

Do you have any better suggestions?thanks

JingsongLi:
Maybe we can just create an base class `FormatFilterTest` for `OrcFilterConverterTest`.

zhangjun0x01:
> Maybe we can just create an base class `FormatFilterTest` for `OrcFilterConverterTest`.

I remove the test case in flink module.
and  there is no common logic in `OrcFilterConverterTest` now, it is all for orc format, so I think there's  no much need to create a base class `FormatFilterTest `,what do you think about?

JingsongLi (approved):
Looks good to me!

Issue #924, opened 2023-04-17
[Bug] SdkClientException: Unable to execute HTTP request: Timeout waiting for connection from pool


www2388258980:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/incubator-paimon/issues) and found nothing similar.

### Paimon version

paimon0.4

### Compute Engine

flink1.16

### Minimal reproduce step

**描述**    
/bin/yarn-session.sh --detached \    
-Dtaskmanager.memory.process.size=5000m \    
-Dtaskmanager.memory.managed.size=0m \   
-Dtaskmanager.memory.network.min=80m \   
-Dtaskmanager.memory.network.max=80m \   
-Dtaskmanager.numberOfTaskSlots=4   

**flink on yarn**    

文件系统使用s3，使用paimon构建实时数仓分层,比如会查几张ods的paimon表，写入到一张'merge-engine' = 'partial-update'的大宽表.    
运行一段时间，半小时或者1小时以上。    
其他任务(flink cdc)插入到s3://xxxxxxx/hadoop/warehouse/ods_medatc_fts.db/src_public_comments/schema/schema-0任务正常。

ava.io.UncheckedIOException: java.io.InterruptedIOException: getFileStatus on s3://xxxxxxx/hadoop/warehouse/ods_medatc_fts.db/src_public_comments/schema/schema-0: com.amazonaws.SdkClientException: Unable to execute HTTP request: Timeout waiting for connection from pool
	at org.apache.paimon.schema.SchemaManager.schema(SchemaManager.java:460)
	at org.apache.paimon.operation.KeyValueFileStoreRead.<init>(KeyValueFileStoreRead.java:88)
	at org.apache.paimon.KeyValueFileStore.newRead(KeyValueFileStore.java:84)
	at org.apache.paimon.table.ChangelogWithKeyFileStoreTable.newRead(ChangelogWithKeyFileStoreTable.java:193)
	at org.apache.paimon.table.source.ReadBuilderImpl.newRead(ReadBuilderImpl.java:81)
	at org.apache.paimon.flink.source.FlinkSource.createReader(FlinkSource.java:50)
	at org.apache.flink.streaming.api.operators.SourceOperator.initReader(SourceOperator.java:286)
	at org.apache.flink.streaming.runtime.tasks.SourceOperatorStreamTask.init(SourceOperatorStreamTask.java:94)
[...]

PR #1361, opened 2023-06-13
[bug] fix connection leak while lookup join paimon table #924

leaves12138:
This is a pull request to fix connection leak while do look-up join with paimon table.
close https://github.com/apache/incubator-paimon/issues/924

The leak is covert.  We need to figure out some method to find out all open connections which should be closed.


JingsongLi (approved):
+1

PR #1364, opened 2023-06-13
[bug] bugfix: we should do continuously refresh while invoking look-up join

leaves12138:
A bug fix about bug fix: https://github.com/apache/incubator-paimon/pull/1361

Continuously refresh while look-up scan

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/FileStoreLookupFunction.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/FileStoreLookupFunction.java
@@ -194,9 +194,9 @@ public class FileStoreLookupFunction implements Serializable, Closeable {
     }
 
     private void refresh() throws Exception {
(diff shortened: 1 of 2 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/FileStoreLookupFunction.java
```diff
@@ -194,9 +194,9 @@ private void checkRefresh() throws Exception {
     }
 
     private void refresh() throws Exception {
-        try (RecordReaderIterator<InternalRow> batch =
-                new RecordReaderIterator<>(streamingReader.getRecordReader())) {
-            while (true) {
+        while (true) {
+            try (RecordReaderIterator<InternalRow> batch =
+                    new RecordReaderIterator<>(streamingReader.getRecordReader())) {
```
JingsongLi, line 199: `streamingReader.getRecordReader()` should be `streamingReader.nextBatch()`

JingsongLi:
+1

PR #1365, opened 2023-06-13
[Core]Add skip collect statistics columns config

wg1026688210:
When collecting statistics for a large column in Paimon, it can consume a large amount of memory, especially when dealing with a large number of entries. Therefore, a configuration was added to the PR for collecting statistics on a specific column.

Add a draft for  discusstion


JingsongLi:
https://iceberg.apache.org/docs/latest/configuration/

See write.metadata.metrics.default , this is what we want

PR #1366, opened 2023-06-13
[flink] unaware-bucket table insert topology in flink engine

leaves12138:
Flink insert topology for unaware-bucket table. We can assign compaction while inserting.

Topology will be like this:

compactor coordinator      ------->  compaction worker   -------> committer   <--------   writer  <-------- other input

the reason to introduce this topology:
we can't put the compaction coordinator between writer and compaction worker like this:

writer ------->  compaction coordinator  -------> compaction worker  --------> committer

cause, this topology, compaction coordinator need to do two things:
1.  accept the committable passed from the writer
2. scan the latest snapshot for the result of the worker generates

The coordinator doesn't know anything about the files workers generate, so it have to scan the snapshot again and again for the result. The scanned result and the files passed from the writer maybe in conflict. The whole job will failover frequently. Though we have method to solve this question, but it is costly(need remember files and do checkpoint).

So the first topology, the coordinator only scan the snapshot fully and continuously will simplify the topology, and solve this question.

link: https://github.com/apache/incubator-paimon/issues/1224

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/append/AppendOnlyCompactionTask.java
+++ b/paimon-core/src/main/java/org/apache/paimon/append/AppendOnlyCompactionTask.java
@@ -69,7 +69,8 @@ public class AppendOnlyCompactionTask {
                 new CompactIncrement(compactBefore, compactAfter, Collections.emptyList());
         return new CommitMessageImpl(
                 partition,
-                0, // bucket 0 is temp bucket for unaware-bucket table
+                0, // bucket 0 is bucket for unaware-bucket table for compatibility with the old
+                // design
                 NewFilesIncrement.emptyIncrement(),
                 compactIncrement);
     }
--- a/paimon-core/src/main/java/org/apache/paimon/append/AppendOnlyCompactionTask.java
+++ b/paimon-core/src/main/java/org/apache/paimon/append/AppendOnlyCompactionTask.java
@@ -69,7 +69,7 @@ public class AppendOnlyCompactionTask {
                 new CompactIncrement(compactBefore, compactAfter, Collections.emptyList());
         return new CommitMessageImpl(
                 partition,
-                0, // bucket 0 is temp bucket for non-bucket table
+                0, // bucket 0 is temp bucket for unaware-bucket table
                 NewFilesIncrement.emptyIncrement(),
                 compactIncrement);
     }
--- a/paimon-core/src/main/java/org/apache/paimon/operation/AppendOnlyFileStoreWrite.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/AppendOnlyFileStoreWrite.java
@@ -32,6 +32,7 @@ import org.apache.paimon.io.DataFileMeta;
 import org.apache.paimon.io.DataFilePathFactory;
 import org.apache.paimon.io.RowDataRollingFileWriter;
 import org.apache.paimon.reader.RecordReaderIterator;
+import org.apache.paimon.table.BucketMode;
 import org.apache.paimon.table.source.DataSplit;
 import org.apache.paimon.types.RowType;
 import org.apache.paimon.utils.CommitIncrement;
@@ -62,9 +63,10 @@ public class AppendOnlyFileStoreWrite extends AbstractFileStoreWrite<InternalRow
     private final int compactionMinFileNum;
     private final int compactionMaxFileNum;
     private final boolean commitForceCompact;
-    private boolean skipCompaction;
     private final boolean assertDisorder;
     private final String fileCompression;
+    private boolean skipCompaction;
+    private BucketMode bucketMode = BucketMode.FIXED;
 
     public AppendOnlyFileStoreWrite(
             FileIO fileIO,
@@ -159,7 +161,20 @@ public class AppendOnlyFileStoreWrite extends AbstractFileStoreWrite<InternalRow
         };
     }
 
-    public void skipCompaction() {
-        skipCompaction = true;
+    public AppendOnlyFileStoreWrite withBucketMode(BucketMode bucketMode) {
+        // AppendOnlyFileStoreWrite is sensitive with bucket mode. It will act difference in
+        // unaware-bucket mode (no compaction and force empty-writer).
+        this.bucketMode = bucketMode;
+        if (bucketMode == BucketMode.UNAWARE) {
+            super.withIgnorePreviousFiles(true);
+            skipCompaction = true;
+        }
+        return this;
+    }
+
+    @Override
+    public void withIgnorePreviousFiles(boolean emptyWriter) {
+        // in unaware bucket mode, we need all writers to be empty
+        super.withIgnorePreviousFiles(emptyWriter || bucketMode == BucketMode.UNAWARE);
     }
 }
--- a/paimon-core/src/main/java/org/apache/paimon/operation/AppendOnlyFileStoreWrite.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/AppendOnlyFileStoreWrite.java
@@ -173,8 +173,8 @@ public class AppendOnlyFileStoreWrite extends AbstractFileStoreWrite<InternalRow
     }
 
     @Override
-    public void withIgnorePreviousFiles(boolean emptyWriter) {
+    public void withIgnorePreviousFiles(boolean ignorePrevious) {
         // in unaware bucket mode, we need all writers to be empty
-        super.withIgnorePreviousFiles(emptyWriter || bucketMode == BucketMode.UNAWARE);
+        super.withIgnorePreviousFiles(ignorePrevious || bucketMode == BucketMode.UNAWARE);
     }
 }
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/compact/UnawareBucketCompactionTopoBuilder.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/compact/UnawareBucketCompactionTopoBuilder.java
@@ -22,7 +22,7 @@ import org.apache.paimon.append.AppendOnlyCompactionTask;
 import org.apache.paimon.flink.FlinkConnectorOptions;
 import org.apache.paimon.flink.sink.Committable;
 import org.apache.paimon.flink.sink.UnawareBucketCompactionSink;
-import org.apache.paimon.flink.source.UnawareBucketSourceFunction;
+import org.apache.paimon.flink.source.UnawareBucketCompactorCoordinationSourceFunction;
 import org.apache.paimon.options.Options;
 import org.apache.paimon.predicate.Predicate;
 import org.apache.paimon.predicate.PredicateBuilder;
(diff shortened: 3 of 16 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkSinkBuilder.java
```diff
@@ -102,4 +104,13 @@ private DataStreamSink<?> buildForFixedBucket() {
                 new FileStoreSink(table, lockFactory, overwritePartition, logSinkFunction);
         return sink.sinkFrom(partitioned);
     }
+
+    private DataStreamSink<?> buildUnawareBucketSink() {
+        return new UnawareBucketWriteSink(
```
JingsongLi, line 109: Just inline `UnawareBucketWriteSink` here?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/UnawareBucketCompactionSink.java
```diff
@@ -42,6 +42,13 @@ public static DataStreamSink<?> sink(
         return new UnawareBucketCompactionSink(table).sinkFrom(input);
     }
 
+    public static DataStream<Committable> uncommitted(
```
JingsongLi, line 45: this method is useless... remove it. (it just wrap one line)

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkWriteSink.java
```diff
@@ -41,7 +41,15 @@ public FlinkWriteSink(
             FileStoreTable table,
             @Nullable Map<String, String> overwritePartition,
             Lock.Factory lockFactory) {
-        super(table, overwritePartition != null);
+        this(table, overwritePartition, lockFactory, false);
+    }
+
+    public FlinkWriteSink(
```
JingsongLi, line 47: useless?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/UnawareBucketWriteSink.java
```diff
@@ -0,0 +1,95 @@
...
+        String commitUser = UUID.randomUUID().toString();
+        // we set partitioner equals null, if the parallelism we set equals to input, we could chain
+        // sink operator and input in one stage
+        PartitionTransformation<RowData> partitioned =
+                new PartitionTransformation<>(input.getTransformation(), null);
+        if (parallelism != null) {
+            partitioned.setParallelism(parallelism);
+        }
+        FileStoreSink sink =
+                new FileStoreSink(table, lockFactory, overwritePartitions, logSinkFunction, true);
```
JingsongLi, line 73: Why just true?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FileStoreSink.java
```diff
@@ -40,7 +40,16 @@ public FileStoreSink(
             Lock.Factory lockFactory,
             @Nullable Map<String, String> overwritePartition,
             @Nullable LogSinkFunction logSinkFunction) {
-        super(table, overwritePartition, lockFactory);
+        this(table, lockFactory, overwritePartition, logSinkFunction, false);
+    }
+
+    public FileStoreSink(
```
JingsongLi, line 46: should introduce this?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/UnawareBucketWriteSink.java
JingsongLi: It did confuse me...
Actually, you already force setting emptyWriter to true in TableWrite...

You could make UnawareBucketWriteSink  extends FlinkWriteSink, and override `sinkFrom(DataStream<T> input, String commitUser, StoreSinkWrite.Provider sinkProvider)`.

Then just plug in the nodes related to the compaction.

leaves12138: I found it will set emptyWrite to false back in `StoreSinkWriteImpl`
When overwritePartitions is null, `StoreSinkWriteImpl.newTableWrite` will call `fromEmptyWriter(false) ` set the flag back to false

leaves12138:
Fixed comment and find out a way to disable fromEmptyWriter call from outside without add new method.
please review and check if the solution meet requests.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkSinkBuilder.java
leaves12138: add checks for append-only mode

leaves12138:
fixed expected

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/UnawareBucketWriteSink.java
```diff
@@ -0,0 +1,95 @@
...
+        super(table, lock, overwritePartitions, logSinkFunction);
+        this.table = table;
+        this.enableCompaction = !table.coreOptions().writeOnly();
+    }
+
+    public DataStreamSink<?> build(DataStream<RowData> input, Integer parallelism) {
+        DataStream<RowData> partitioned = partition(input, parallelism);
+
+        return sinkFrom(partitioned);
+    }
```
tsreaper, line 59: This `build` method looks strange to me. This is not a builder class. Why not just merge all these into `sinkFrom`?

tsreaper (commented):
Could you also add some random tests and failure tests, just like `ChangelogWithKeyFileStoreTableITCase`?

paimon-core/src/main/java/org/apache/paimon/operation/AppendOnlyFileStoreWrite.java
```diff
@@ -159,7 +161,20 @@ public AppendOnlyCompactManager.CompactRewriter compactRewriter(
...
+            super.withIgnorePreviousFiles(true);
+            skipCompaction = true;
+        }
+        return this;
+    }
+
+    @Override
+    public void withIgnorePreviousFiles(boolean emptyWriter) {
+        // in unaware bucket mode, we need all writers to be empty
+        super.withIgnorePreviousFiles(emptyWriter || bucketMode == BucketMode.UNAWARE);
```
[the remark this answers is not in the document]
leaves12138: In unaware-bucket mode, AppendOnlyFileStoreWrite no longer compact any files. The coordinator will scan the whole.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/compact/UnawareBucketCompactionTopoBuilder.java
```diff
@@ -73,14 +74,22 @@ public void withPartitions(List<Map<String, String>> partitions) {
...
+
+        // rebalance input to default or assigned parallelism
+        DataStream<AppendOnlyCompactionTask> rebalanced = rebalanceInput(source);
+
+        return new UnawareBucketCompactionSink(table).doWrite(rebalanced, commitUser);
+    }
+
     private DataStreamSource<AppendOnlyCompactionTask> buildSource() {
         long scanInterval = table.coreOptions().continuousDiscoveryInterval().toMillis();
         UnawareBucketSourceFunction source =
```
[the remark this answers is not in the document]
leaves12138: ok

paimon-core/src/main/java/org/apache/paimon/append/AppendOnlyCompactionTask.java
```diff
@@ -69,7 +69,7 @@ public CommitMessage doCompact(AppendOnlyCompactManager.CompactRewriter rewriter
                 new CompactIncrement(compactBefore, compactAfter, Collections.emptyList());
         return new CommitMessageImpl(
                 partition,
-                0, // bucket 0 is temp bucket for non-bucket table
+                0, // bucket 0 is temp bucket for unaware-bucket table
```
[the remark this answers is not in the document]
leaves12138: To be compatible with previous design, we put these record in 'bucket = 0'. It's the temp solution for this. Maybe unaware of bucket only need one directory 'data'. Because there is no bucket concept.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/UnawareBucketWriteSink.java
leaves12138: Learned from `RowDynamicBucketSink`.

leaves12138: changed this from build to sinkFrom

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkSink.java
```diff
@@ -126,33 +126,29 @@ public DataStreamSink<?> sinkFrom(DataStream<T> input) {
...
+        return doCommit(written, initialCommitUser);
     }
 
-    public DataStreamSink<?> sinkFrom(
-            DataStream<T> input, String commitUser, StoreSinkWrite.Provider sinkProvider) {
-        StreamExecutionEnvironment env = input.getExecutionEnvironment();
-        ReadableConfig conf = StreamExecutionEnvironmentUtils.getConfiguration(env);
-        CheckpointConfig checkpointConfig = env.getCheckpointConfig();
-
+    public SingleOutputStreamOperator<Committable> doWrite(DataStream<T> input, String commitUser) {
```
JingsongLi, line 136: `Integer parallelism` to set writer parallelism.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/UnawareBucketWriteSink.java
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        // commit the committable to generate a new snapshot
+        return doCommit(written, initialCommitUser);
+    }
+
+    private DataStream<RowData> partition(DataStream<RowData> input, Integer parallelism) {
+        // we set partitioner equals null, if the parallelism we set equals to input, we could chain
+        // sink operator and input in one stage
+        PartitionTransformation<RowData> partitionTransformation =
```
JingsongLi, line 88: You should remove this to set writer parallelism.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/UnawareBucketCompactorCoordinationSourceFunction.java
```diff
@@ -38,10 +38,20 @@
...
+/**
+ * Source Function for unaware-bucket Compaction.
+ *
+ * <p>Note: The function is the source function of unaware-bucket compactor coordinator. It will
+ * read the latest snapshot continuously by compactionCoordinator, and generate new compaction
+ * tasks. The source function is used in unaware-bucket compaction job (both stand-alone and
+ * write-combined). Besides, we don't need to save state in this function, it will invoke a full
+ * scan when starting up, and scan continuously for the following snapshot.
+ */
+public class UnawareBucketCompactorCoordinationSourceFunction
```
JingsongLi, line 50: `BucketUnawareCompactSource`

JingsongLi (approved):
+1

PR #1370, opened 2023-06-14
[flink] Remove per record RecordsWithSplitIds

JingsongLi:
### Purpose

At the moment, we don't have to go line by line to send data, checkpoint from a particular data insert makes little sense, which gives us the complexity of the architecture.


yuzelin (approved):
+1

PR #1371, opened 2023-06-14
[bug] Fix: make hive load the correct FileIO set by conf

LsomeYeah:
### Purpose

Linked issue: close #xxx

When creating table in hive, if the scheme of the table location is not consistent with the fs.defaultFs in HiveConf, hive may load false FileIOLoader which will cause IOException.

So, before loading FileIOLoader, it should get the correct path according to the fs.defaultFs in HiveConf.

### Tests

There is an e2e-test named 'testHiveWrite' in HiveE2eTest for fixing this bug.


paimon-hive/paimon-hive-connector-common/src/main/java/org/apache/paimon/hive/PaimonMetaHook.java
```diff
@@ -76,10 +77,12 @@ public void preCreateTable(Table table) throws MetaException {
...
-        String location = LocationKeyExtractor.getLocation(table);
+        String location = LocationKeyExtractor.getLocation(table, conf);
         if (location == null) {
             String warehouse = conf.get(HiveConf.ConfVars.METASTOREWAREHOUSE.varname);
+            org.apache.hadoop.fs.Path hadoopPath =
+                    getDnsPath(new org.apache.hadoop.fs.Path(warehouse), conf);
+            warehouse = hadoopPath.toUri().toString();
             Identifier identifier = Identifier.create(table.getDbName(), table.getTableName());
             location = AbstractCatalog.dataTableLocation(warehouse, identifier).toUri().toString();
             table.getSd().setLocation(location);
```
tsreaper, line 88: Why not merge these implementation into `LocationKeyExtractor.getLocation`?

LsomeYeah: `Identifier.class` and `AbstractCatalog.class` are used here. The two classes are defined in '`paimon-core`' module, if merge these implementation into `LocationKeyExtractor.getLocation`, we need add dependency to the `pom.xml` of the module '`paimon-hive : paimon-hive-common`' in which `LocationKeyExtractor.getLocation` is defined.

tsreaper (approved):

PR #1373, opened 2023-06-14
[flink] Supports watermark transmitting for consumer-id mode

JingsongLi:
### Purpose

For consumer-id mode, when there is no watermark definition, the Paimon table will pass the watermark in the snapshot to the downstream Paimon table, which means you can track the progress of the watermark for the entire pipeline.


yuzelin (approved):
+1

PR #1378, opened 2023-06-14
[hotfix] Fix that Maven 3.2.5 cannot build paimon-core

yuzelin:
### Purpose

Use Maven 3.2.5 to build paimon-core will throw error: cannot find org.apache.hadoop.hive.ql.exec.vector.
This PR fix it.


JingsongLi:
+1

Issue #1375, opened 2023-06-14
[Feature] Optimize memory for big column


JingsongLi:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/incubator-paimon/issues) and found nothing similar.

### Motivation

Currently Paimon has some problems with memory usage if there is a column with a very large field, more than 2MB.

Problem:

1. Batch size (read.batch-size) not work for sink, currently, Orc writer also create a ColumnarBatch, we should limit its memory too.
2. Orc Dictionary needs much memory, we can introduce a common option and document this to close Dictionary.
3. In the scenario of frequent failover, we should restrict the continuation of new file committing, see #1351 
4. `RollingFileWriter` should not refer all writers, or we should set writer and maxKey to null in `SingleFileWriter`s.
5. We should avoid adding statistics to very large columns, which can cause our Manifest file to become very large.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

wg1026688210:
I'm willing to submit a PR!

zhangjun0x01:
hi,@JingsongLi , do you think is it necessary to separate key and value  for big value? like this : https://rocksdb.org/blog/2021/05/26/integrated-blob-db.html

JingsongLi:
> hi,@JingsongLi , do you think is it necessary to separate key and value for big value? like this : https://rocksdb.org/blog/2021/05/26/integrated-blob-db.html

This is a huge topic....

zhangjun0x01:
> > hi,@JingsongLi , do you think is it necessary to separate key and value for big value? like this : https://rocksdb.org/blog/2021/05/26/integrated-blob-db.html
> 
> This is a huge topic....

Doris also has similar function, and I think paimon currently may does not have strong user needs. We can discuss this topic when users have needs later.

JingsongLi:
> Doris also has similar function, and I think paimon currently may does not have strong user needs. We can discuss this topic when users have needs later.

Hi @zhangjun0x01 , if just for performance improvement, yes Paimon also has the requirement. Can you just create a new issue?

zhangjun0x01:
> > Doris also has similar function, and I think paimon currently may does not have strong user needs. We can discuss this topic when users have needs later.
> 
> Hi @zhangjun0x01 , if just for performance improvement, yes Paimon also has the requirement. Can you just create a new issue?

I create a new issue https://github.com/apache/incubator-paimon/issues/1428

PR #1382, opened 2023-06-15
[core] RollingFileWriter should not refer all writers

JingsongLi:
### Purpose

RollingFileWriter should not refer all writers, or we should set writer and maxKey to null in SingleFileWriters.

part of #1375


leaves12138:
+1

PR #1384, opened 2023-06-15
 [spark] Support to write to dynamic bucket table

YannByron:
### Purpose

Supports to `insert into` Paimon KVTable with dynamic bucket by Spark.

### Tests

SparkWriteITCase.testWriteDynamicBucketPartitionedTable

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/index/PartitionIndex.java
+++ b/paimon-core/src/main/java/org/apache/paimon/index/PartitionIndex.java
@@ -33,9 +33,7 @@ import java.util.function.Function;
 
 import static org.apache.paimon.index.HashIndexFile.HASH_INDEX;
 
-/**
- * Bucket Index Per Partition.
- */
+/** Bucket Index Per Partition. */
 public class PartitionIndex {
 
     public final Int2ShortHashMap hash2Bucket;
--- a/paimon-core/src/main/java/org/apache/paimon/index/PartitionIndex.java
+++ b/paimon-core/src/main/java/org/apache/paimon/index/PartitionIndex.java
@@ -1,0 +1,136 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.paimon.index;
+
+import org.apache.paimon.data.BinaryRow;
+import org.apache.paimon.manifest.IndexManifestEntry;
+import org.apache.paimon.utils.Int2ShortHashMap;
+import org.apache.paimon.utils.IntIterator;
+
+import java.io.EOFException;
+import java.io.IOException;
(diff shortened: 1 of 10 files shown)


paimon-core/src/main/java/org/apache/paimon/index/PartitionIndex.java
```diff
@@ -0,0 +1,137 @@
...
+            Map<Integer, Long> bucketInformation,
+            long targetBucketRowNumber) {
+        this.hash2Bucket = hash2Bucket;
+        this.bucketInformation = bucketInformation;
+        this.targetBucketRowNumber = targetBucketRowNumber;
+        this.lastAccessedCommitIdentifier = Long.MIN_VALUE;
+        this.accessed = true;
+    }
+
+    public int assign(int hash) {
```
JingsongLi, line 60: It is better to remove this method, let invoker to define a filter function.

paimon-core/src/main/java/org/apache/paimon/index/PartitionIndex.java
```diff
@@ -0,0 +1,137 @@
...
+        @SuppressWarnings("OptionalGetWithoutIsPresent")
+        int maxBucket =
+                bucketInformation.keySet().stream().mapToInt(Integer::intValue).max().getAsInt();
+        throw new RuntimeException(
+                String.format(
+                        "To more bucket %s, you should increase target bucket row number %s.",
+                        maxBucket, targetBucketRowNumber));
+    }
+
+    public static PartitionIndex loadIndex(
```
JingsongLi, line 103: It is better to remove this method, let invoker to define a filter function.

paimon-core/src/main/java/org/apache/paimon/table/Table.java
```diff
@@ -58,6 +58,8 @@ public interface Table extends Serializable {
     /** Optional comment of this table. */
     Optional<String> comment();
 
+    boolean isFileStoreTable();
```
JingsongLi, line 61: `Table` is the top level API. Regarding the write path, maybe we need a new design that doesn't need to be exposed here yet.
I think we can not expose this method now. We can just try cast to `FileStoreTable`.

paimon-core/src/main/java/org/apache/paimon/table/sink/DynamicBucketRow.java
```diff
@@ -26,8 +26,10 @@
 import org.apache.paimon.data.Timestamp;
 import org.apache.paimon.types.RowKind;
 
+import java.io.Serializable;
+
 /** An {@link InternalRow} wraps another {@link InternalRow} with bucket. */
-public class DynamicBucketRow implements InternalRow {
+public class DynamicBucketRow implements InternalRow, Serializable {
```
JingsongLi, line 32: I am concerns about kryo serialization here...

Maybe we can just use `Tuple2<Row, Integer>` in `SparkWrite`?

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/SparkWrite.java
```diff
@@ -75,6 +114,34 @@ public InsertableRelation toInsertableRelation() {
...
 
+    private boolean isDynamicBucketTable() {
+        return table.isFileStoreTable()
+                && ((FileStoreTable) table).bucketMode() == BucketMode.DYNAMIC;
+    }
+
+    private static Iterator<InternalRow> assignBucket(
+            FileStoreTable fileStoreTable, BatchWriteBuilder writeBuilder, Iterator<Row> rows) {
+        long targetBucketRowNumber = fileStoreTable.coreOptions().dynamicBucketTargetRowNum();
+        List<InternalRow> dynamicBucketRows = new ArrayList<>();
```
JingsongLi, line 125: I think it is better to return a lazy Iterator for Spark `RDD.mapPartitions`, otherwise there is a risk of OOM

JingsongLi:
Thanks @YannByron for the contribution! Left comments.

YannByron:
@JingsongLi all of comments have been solved. pls review again.

JingsongLi (approved):
+1

PR #1394, opened 2023-06-19
[core] Do not truncate string and binary for casting

JingsongLi:
### Purpose

DO NOT Truncate or fill, let the data remain in its original form, even it is not compatible with the precision.


zhuangchong:
Add a description to the official website?

JingsongLi:
> Add a description to the official website?

Sounds good.

JingsongLi:
orc will truncate string too... we can not just ignore precision..

PR #1399, opened 2023-06-20
Support for query manifests

### LINKED ISSUE #1154: [Feature] Support for query manifests
### Search before asking

### Motivation

Paimon currently supports query the table  snapshots, schemas, options,  and so on , but does not support query table manifests.

### Solution

_No response_

### Anything else?

_No response_

xwmr-max:
Linked issue: close #1154

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/system/ManifestsTable.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/system/ManifestsTable.java
@@ -1,0 +1,252 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.paimon.table.system;
+
+import org.apache.paimon.CoreOptions;
+import org.apache.paimon.Snapshot;
+import org.apache.paimon.data.BinaryString;
+import org.apache.paimon.data.GenericRow;
+import org.apache.paimon.data.InternalRow;
+import org.apache.paimon.format.FileFormat;
+import org.apache.paimon.fs.FileIO;
+import org.apache.paimon.fs.Path;
+import org.apache.paimon.manifest.ManifestFileMeta;
+import org.apache.paimon.manifest.ManifestList;
+import org.apache.paimon.predicate.Predicate;
+import org.apache.paimon.reader.RecordReader;
+import org.apache.paimon.table.ReadonlyTable;
+import org.apache.paimon.table.Table;
+import org.apache.paimon.table.source.InnerTableRead;
+import org.apache.paimon.table.source.InnerTableScan;
+import org.apache.paimon.table.source.ReadOnceTableScan;
+import org.apache.paimon.table.source.Split;
+import org.apache.paimon.types.BigIntType;
+import org.apache.paimon.types.DataField;
+import org.apache.paimon.types.RowType;
+import org.apache.paimon.utils.FileStorePathFactory;
+import org.apache.paimon.utils.IteratorRecordReader;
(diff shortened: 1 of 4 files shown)


JingsongLi (commented):
Thanks @xwmr-max for the contribution, please fix checkstyle, you can run `mvn spotless:apply`

xwmr-max:
> Thanks @xwmr-max for the contribution, please fix checkstyle, you can run `mvn spotless:apply`
The code has been recommitted, but I can't see my workflows.

JingsongLi (commented):
@xwmr-max Can you document this?

xwmr-max:
> document

xwmr-max:
> @xwmr-max Can you document this?

Yes, I will write the documentation after the code has been reviewed and there are no problems.

paimon-core/src/main/java/org/apache/paimon/table/system/ManifestsTable.java
```diff
@@ -0,0 +1,256 @@
...
+/** A {@link Table} for showing committing snapshots of table. */
+public class ManifestsTable implements ReadonlyTable {
+    private static final long serialVersionUID = 1L;
+
+    public static final String MANIFESTS = "manifests";
+
+    public static final RowType TABLE_TYPE =
+            new RowType(
+                    Arrays.asList(
+                            new DataField(0, "snapshot_id", new BigIntType(false)),
```
JingsongLi, line 69: snapshot_id looks weird here, what is the snapshot_id? The snapshot_id to generate manifest file?

Is this needed?

xwmr-max: I think ManifestsTable mainly provides all file names and file sizes that manifest file under the latest snapshot, and also provides the number of files added and deleted. Or do I think it's necessary to list all the manifests files under all the snapshots? Therefore, the snapshot_id mainly reflects the latest snapshot_id and what manifests file in the latest snapshot. That's my understanding. What else do you suggest? @JingsongLi

JingsongLi: > I think ManifestsTable mainly provides all file names and file sizes that manifest file under the latest snapshot, and also provides the number of files added and deleted. Or do I think it's necessary to list all the manifests files under all the snapshots?

Yes, I agree with you.

> Therefore, the snapshot_id mainly reflects the latest snapshot_id and what manifests file in the latest snapshot. 

Yes, I mean it may bring some confuses. 
1. The snapshot that gave birth to the manifest file
2. or the current snapshot

xwmr-max: hi，Sorry for the delay in replying to you due to internal work.
The snapshot_id of the current design is generated by the latest snapshot, which generates all the manifest files. Now that I think about it, this can be confusing. I came up with two options:
1) Change the name of snapshot_id to lastest_snaphot_id to make it clear that this is the latest snapshot
2) List all manifest files, marking which snapshot generated each manifest file, so the snapshot_id here indicates which manifest files were generated.
Which one do you think? @JingsongLi

JingsongLi: Why does it have to be added? What do you need for it?

JingsongLi: I don't quite understand why snapshot_id are necessary

xwmr-max: Refer to option 2 I gave and add snapshot_id to ensure that each manifest file can clearly know which snapshot it belongs to. Therefore, in a table, you can query the manifest files based on the given snapshot_id.

JingsongLi: It is hard to implement option 2.

JingsongLi: I repeat, I don't understand why we need to return the same snapshot for each row.

JingsongLi: Please do not reply simply, it is clearer. This means nothing to me.

xwmr-max: Among the two options I proposed, option 1 does return the same value in each row, which does not seem to make sense; option 2 is indeed more difficult to implement. Sorry for my misunderstanding

xwmr-max: I will delete the snapshot_id

JingsongLi: Thanks~

JingsongLi:
Compile failed.

xwmr-max:
>

xwmr-max:
> Compile failed.

done

JingsongLi (approved):
+1

PR #1402, opened 2023-06-20
[core] Refactor Lock into FileStoreTable and hide it from connectors

tsreaper:
### Purpose

Currently `Lock` is exposed to all connectors such as Flink, Spark, etc. When creating writers in the connectors, we have to keep in mind that we should use a lock if needed. This is sort of cumbersome for connector creators.

This PR refactors `Lock` into `FileStoreTable` and hide it from connectors. Connectors now do not need to concern about locks. They can create `TableCommit` as they want and don't have to check if there is a lock.

### Tests

Existing UT/IT cases should cover this change.

### API and Format

N/A

### Documentation

N/A


JingsongLi (approved):
+1

Issue #739, opened 2023-03-29
[Feature] Introduce generic mode for paimon Flink&Spark catalog


JingsongLi:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/incubator-paimon/issues) and found nothing similar.

### Motivation

- Spark: Like iceberg `SparkSessionCatalog`.
- Flink: Merge Flink `HiveCatalog`.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

legendtkl:
I would like to take this ticket if no one take it.

PR #1403, opened 2023-06-20
[spark] Introduce SparkGenericCatalog

JingsongLi:
### Purpose

When starting `spark-sql`, use the following command to register Paimon’s Spark Generic catalog to replace Spark
default catalog `spark_catalog`. (default warehouse is Spark `spark.sql.warehouse.dir`)

```bash
spark-sql ... \
    --conf spark.sql.catalog.spark_catalog=org.apache.paimon.spark.SparkGenericCatalog
```

Using `SparkGenericCatalog`, you can use Paimon tables in this Catalog or non-Paimon tables such as Spark's csv,
parquet, Hive tables, etc.

part of #739

### Tests

`SparkGenericCatalogTest`

### API and Format

`SparkGenericCatalog`

### Documentation

yes

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -232,7 +232,7 @@ paimon-common/src/main/java/org/apache/paimon/types/DataType.java
 paimon-common/src/main/java/org/apache/paimon/options/ConfigOption.java
 from http://flink.apache.org/ version 1.17.0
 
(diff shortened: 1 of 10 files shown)


yuzelin (approved):
+1

PR #1411, opened 2023-06-20
[spark] Remove SparkCatalogBase

JingsongLi:
### Purpose

The fact that there are more methods doesn't affect older versions of Spark, we don't need to have SparkCatalogBase


yuzelin (approved):
+1

PR #1414, opened 2023-06-21
[core] Introduce filterAndCommit method for StreamTableCommit, which merges filterCommitted and commit

tsreaper:
### Purpose

Currently to retry the commit process after failure, the user has to first call `StreamTableCommit#filterCommitted` and then call `StreamTableCommit#commit`. This is kind of cumbersome.

This PR merges the two method calls into one `StreamTableCommit#filterAndCommit`. Users just need to pass all `CommitMessage`s in question into the method and it will automatically choose what to commit.

What's more, we are going to support adding partition into Hive metastore in the future. By adding the `StreamTableCommit#filterAndCommit` method we can check and create our desired partition in this method during retry to ensure consistency.

### Tests

Existing UT/IT should cover this change.

### API and Format

Yes. It changes the public Java API. `StreamTableCommit#filterCommitted` is now deprecated.

### Documentation

No.

### DIFF
--- a/docs/content/api/java-api.md
+++ b/docs/content/api/java-api.md
@@ -560,7 +560,7 @@ Key points to achieve exactly-once consistency:
 - Different applications need to use different commitUsers.
 - The commitIdentifier of `StreamTableWrite` and `StreamTableCommit` needs to be consistent, and the
   id needs to be incremented for the next committing.
-- When a failure occurs, if you still have uncommitted `CommitMessage`s, please use `StreamTableCommit#filterCommitted`
+- When a failure occurs, if you still have uncommitted `CommitMessage`s, please use `StreamTableCommit#filterAndCommit`
   to exclude the committed messages by commitIdentifier.
 
 ```java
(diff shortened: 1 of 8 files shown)


docs/content/api/java-api.md
```diff
@@ -592,10 +592,22 @@ public class StreamWriteTable {
 
         // 3. Collect all CommitMessages to a global node and commit
         StreamTableCommit commit = writeBuilder.newCommit();
-        commit.commit(commitIdentifier, messages);
-
-        // 4. When failover, you can use 'filterCommitted' to filter committed commits.
-        commit.filterCommitted(committedIdentifiers);
+        boolean firstTry = true;
```
JingsongLi, line 595: Here is very confused...
Actually, when failure occurs, just `filterAndCommit` messages in state, and use commit for new messages.

JingsongLi (approved):
+1

JingsongLi (approved):
+1

PR #1417, opened 2023-06-21
[core] Introduce tags metadata table

yuzelin:
### Purpose

Add system table for tags.

### API and Format
new system table `<table>$tags`

### Documentation

TODO


paimon-core/src/main/java/org/apache/paimon/table/system/TagsTable.java
```diff
@@ -0,0 +1,209 @@
...
+
+    private static final long serialVersionUID = 1L;
+
+    public static final String TAGS = "tags";
+
+    public static final RowType TABLE_TYPE =
+            new RowType(
+                    Arrays.asList(
+                            new DataField(0, "tag_name", SerializationUtils.newStringType(false)),
+                            new DataField(1, "tagged_snapshot_id", new BigIntType(false)),
```
JingsongLi, line 69: just snapshot_id?

JingsongLi (commented):
Please document this.

PR #1422, opened 2023-06-21
[flink] Introduce create-tag and delete-tag actions

yuzelin:
### Purpose

Allow user to create and delete tag by flink.
### Tests
`TagActionITCase`
`FlinkActionsE2eTest#testCreateAndDeleteTag`

### API and Format
new flink actions:
create-tag
delete-tag

### Documentation

todo

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/utils/TagManager.java
+++ b/paimon-core/src/main/java/org/apache/paimon/utils/TagManager.java
@@ -25,6 +25,7 @@ import org.apache.paimon.fs.Path;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Comparator;
 import java.util.List;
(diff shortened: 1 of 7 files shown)


paimon-core/src/main/java/org/apache/paimon/utils/TagManager.java
```diff
@@ -113,7 +113,9 @@ public SortedMap<Snapshot, String> tags() {
         for (FileStatus status : statuses) {
             Path path = status.getPath();
             if (path.getName().startsWith(TAG_PREFIX)) {
-                tags.put(Snapshot.fromPath(fileIO, path), path.getName());
+                tags.put(
```
JingsongLi, line 116: Add a test in `CatalogTableITCase`.

JingsongLi (approved):
+1

PR #1424, opened 2023-06-21
[flink] Add supports of lookup join on append-only table

### LINKED ISSUE #1410: [Feature] Append-only supports lookup join
### Search before asking

### Motivation

Append-only also can support Flink lookup join

### Solution

_No response_

### Anything else?

_No response_

Aitozi:
### Purpose

Linked issue: close #1410

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/NoPrimaryKeyLookupTable.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/NoPrimaryKeyLookupTable.java
@@ -1,0 +1,84 @@
+/*
+ *
+ *  * Licensed to the Apache Software Foundation (ASF) under one
+ *  * or more contributor license agreements.  See the NOTICE file
+ *  * distributed with this work for additional information
+ *  * regarding copyright ownership.  The ASF licenses this file
+ *  * to you under the Apache License, Version 2.0 (the
+ *  * "License"); you may not use this file except in compliance
+ *  * with the License.  You may obtain a copy of the License at
+ *  *
+ *  *     http://www.apache.org/licenses/LICENSE-2.0
+ *  *
+ *  * Unless required by applicable law or agreed to in writing, software
+ *  * distributed under the License is distributed on an "AS IS" BASIS,
+ *  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  * See the License for the specific language governing permissions and
+ *  * limitations under the License.
+ *
+ */
+
+package org.apache.paimon.flink.lookup;
+
+import org.apache.paimon.data.InternalRow;
+import org.apache.paimon.data.serializer.InternalSerializers;
+import org.apache.paimon.types.RowKind;
+import org.apache.paimon.types.RowType;
+import org.apache.paimon.utils.KeyProjectedRow;
+import org.apache.paimon.utils.TypeUtils;
+
+import java.io.IOException;
+import java.util.Iterator;
+import java.util.List;
+import java.util.function.Predicate;
+
+/** A {@link LookupTable} for table without primary key. */
+public class NoPrimaryKeyLookupTable implements LookupTable {
+
+    private final RocksDBSetState state;
+
+    private final Predicate<InternalRow> recordFilter;
+
+    private final KeyProjectedRow joinKeyRow;
+
+    public NoPrimaryKeyLookupTable(
+            RocksDBStateFactory stateFactory,
+            RowType rowType,
+            List<String> joinKeys,
+            Predicate<InternalRow> recordFilter,
+            long lruCacheSize)
+            throws IOException {
+        List<String> fieldNames = rowType.getFieldNames();
+        int[] joinKeyMapping = joinKeys.stream().mapToInt(fieldNames::indexOf).toArray();
+        this.joinKeyRow = new KeyProjectedRow(joinKeyMapping);
+        this.state =
+                stateFactory.setState(
+                        "join-key-index",
+                        InternalSerializers.create(TypeUtils.project(rowType, joinKeyMapping)),
+                        InternalSerializers.create(rowType),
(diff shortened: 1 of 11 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/NoPrimaryKeyLookupTable.java
```diff
@@ -0,0 +1,84 @@
...
+
+import java.io.IOException;
+import java.util.Iterator;
+import java.util.List;
+import java.util.function.Predicate;
+
+/** A {@link LookupTable} for table without primary key. */
+public class NoPrimaryKeyLookupTable implements LookupTable {
+
+    private final RocksDBSetState state;
```
JingsongLi, line 38: Maybe we should introduce `RocksDBMapState`?  Duplicate Rows should be output multiple times.

You should add test for this.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/NoPrimaryKeyLookupTable.java
```diff
@@ -0,0 +1,84 @@
...
+    @Override
+    public void refresh(Iterator<InternalRow> incremental) throws IOException {
+        while (incremental.hasNext()) {
+            InternalRow row = incremental.next();
+            joinKeyRow.replaceRow(row);
+            if (row.getRowKind() == RowKind.INSERT || row.getRowKind() == RowKind.UPDATE_AFTER) {
+                if (recordFilter.test(row)) {
+                    state.add(joinKeyRow, row);
+                } else {
+                    state.retract(joinKeyRow, row);
```
JingsongLi, line 77: Why we need to retract?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/NoPrimaryKeyLookupTable.java
Aitozi: You are right, will fix this.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/NoPrimaryKeyLookupTable.java
Aitozi: No need, will remove this.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/NoPrimaryKeyLookupTable.java
Aitozi: I use the RocksDBListState here.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/NoPrimaryKeyLookupTable.java
```diff
@@ -0,0 +1,84 @@
...
+    public void refresh(Iterator<InternalRow> incremental) throws IOException {
+        while (incremental.hasNext()) {
+            InternalRow row = incremental.next();
+            joinKeyRow.replaceRow(row);
+            if (row.getRowKind() == RowKind.INSERT || row.getRowKind() == RowKind.UPDATE_AFTER) {
+                if (recordFilter.test(row)) {
+                    state.add(joinKeyRow, row);
+                }
+            } else {
+                state.retract(joinKeyRow, row);
```
JingsongLi, line 80: here we can just throw exception? We can assume all records should be insert or ua.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/RocksDBSetState.java
```diff
@@ -89,7 +96,8 @@ public void retract(InternalRow key, InternalRow value) throws IOException {
     public void add(InternalRow key, InternalRow value) throws IOException {
         try {
             byte[] bytes = invalidKeyAndGetKVBytes(key, value);
-            db.put(columnFamily, writeOptions, bytes, EMPTY);
+            long count = toLong(db.get(columnFamily, bytes));
```
JingsongLi, line 99: Why need to modify this class?

JingsongLi:
Thanks @Aitozi for the update, left two comments.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/NoPrimaryKeyLookupTable.java
Aitozi: as I  know, we can have delete record here, if we execute the `DELETE FROM` on a changlog table without primary key

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/RocksDBSetState.java
Aitozi: Reverted

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/NoPrimaryKeyLookupTable.java
JingsongLi: we can just not support `changlog table without pk`.

actually, I am considering to remove this mode....

Aitozi: Got it. Updated

JingsongLi (approved):
Looks good to me! Thanks @Aitozi

PR #1426, opened 2023-06-24
[core]Fix inconsistent format option between read and write

wg1026688210:
The format options ,such as 'orc.xx','parquet.xx' currently don't work for reading. 
It work for writing


JingsongLi:
Why don't work for reading? You can see `OrcFileFormat.getOrcProperties`. It will add prefix "orc.".

JingsongLi (commented):
You can add a test in `FileFormatTest`.

JingsongLi (approved):
+1

PR #1427, opened 2023-06-24
fix workflow fail

wg1026688210:
the jdk distrbution "Temurin" seem to fail setting up, we can change the  distribution to adopt to work around it.
image


JingsongLi (approved):
+1 thanks @wg1026688210

PR #1431, opened 2023-06-25
[core] Introduce sort-spill-threshold

JingsongLi:
### Purpose

If the maximum number of sort readers exceeds this value, a spill will be attempted.
This prevents too many readers from consuming too much memory and causing OOM.


leaves12138:
+1

PR #1433, opened 2023-06-25
[hdfs] Introduce hadoop-conf-loader

JingsongLi:
### Purpose

Introduce `hadoop-conf-loader`:
Specifies the way of loading hadoop config.


yuzelin (approved):
+1

PR #1434, opened 2023-06-25
[flink] integrate DELETE filter pushdown for flink-1.17

legendtkl:
### Purpose
Linked issue: close https://github.com/apache/incubator-paimon/issues/989

### Brief change
Supported filters:
    1. where primary key = x
    2. where primary key = x or key = y
    3. where primary key in (x, y, z)
    4. where partition key = x

Note: primary key between filter now is not supported in Paimon, ignore it now and let's start a new ticket for it later.

### Tests
UT:
* ReadWriteTableITCase#testDeletePushDownWithPrimaryKey
* ReadWriteTableITCase#testDeletePushDownWithPartitionKey

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/predicate/DeletePushDownFunctionVisitor.java
+++ b/paimon-common/src/main/java/org/apache/paimon/predicate/DeletePushDownFunctionVisitor.java
@@ -105,7 +105,7 @@ public class DeletePushDownFunctionVisitor implements FunctionVisitor<Boolean> {
     @Override
     public Boolean visitAnd(List<Boolean> children) {
         if (children.size() != 2) {
-            throw new RuntimeException("Illegal children size: " + children.size());
+            return false;
         }
         if (children.get(0) && children.get(1)) {
             return true;
@@ -116,7 +116,7 @@ public class DeletePushDownFunctionVisitor implements FunctionVisitor<Boolean> {
     @Override
     public Boolean visitOr(List<Boolean> children) {
         if (children.size() != 2) {
-            throw new RuntimeException("Illegal children size: " + children.size());
+            return false;
         }
 
         if (children.get(0) || children.get(1)) {
--- a/paimon-common/src/main/java/org/apache/paimon/predicate/DeletePushDownFunctionVisitor.java
+++ b/paimon-common/src/main/java/org/apache/paimon/predicate/DeletePushDownFunctionVisitor.java
@@ -19,10 +19,9 @@
 package org.apache.paimon.predicate;
 
 import java.util.List;
-import java.util.Optional;
 
 /** DeletePushDownFunctionVisitor visit the predicate and check if it can be push down. */
-public class DeletePushDownFunctionVisitor implements FunctionVisitor<Optional<Long>> {
+public class DeletePushDownFunctionVisitor implements FunctionVisitor<Boolean> {
 
     private final List<String> primaryKeys;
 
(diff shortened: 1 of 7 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkTableSink.java
```diff
@@ -112,17 +132,95 @@ public RowLevelUpdateInfo applyRowLevelUpdate(
     @Override
     public RowLevelDeleteInfo applyRowLevelDelete(
             @Nullable RowLevelModificationScanContext rowLevelModificationScanContext) {
+        if (supportDelete()) {
```
JingsongLi, line 135: `void checkDeletable`?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkTableSink.java
```diff
@@ -112,17 +132,95 @@ public RowLevelUpdateInfo applyRowLevelUpdate(
...
+    4. where partition key = x
+    */
+    @Override
+    public boolean applyDeleteFilters(List<ResolvedExpression> list) {
+        if (supportDelete()) {
+            if (list.size() == 0) {
+                return false;
+            }
+
+            predicates = new ArrayList<>();
```
JingsongLi, line 157: local variable, don't assign to class field when return false.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkTableSink.java
```diff
@@ -135,4 +233,50 @@ public RowLevelDeleteInfo applyRowLevelDelete(
                             table.getClass().getName()));
         }
     }
+
+    private boolean shouldPushdownDeleteFilter(Predicate predicate, int predicateSize) {
```
JingsongLi, line 237: `canPushdownDeleteFilter`

JingsongLi, line 237: Use `FunctionVisitor` to visit predicate?

klion26:
Hi, can this pr fix the problem `Could not execute SQL statement. Reason:
org.apache.flink.table.api.TableException: DELETE statement is not supported for streaming mode now.`?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkTableSink.java
legendtkl: hi, @JingsongLi , is the visitor too complicated for this to implement all the method visitXXX of interface FunctionVisitor?

When the filter is `where id=1 or id=2 or id=3` or `where id in (1,2,3)`, it will be converted to 
```sql
               compound(OR)
                   /     \
         compound(OR)     leaf(Equal, id, 1)
         /        \
leaf(Equal,id,2)   leaf(Equal,id,3)
```

We just need to check all the compoundFilter's func is OR and all the leafFilter's func is Equal and field is in primary keys.

JingsongLi: We can just throw exception if encounter an unsupported predicate.
And catch the specific exception outer.

This is a common code style in flink-calcite.

JingsongLi: And we can support `IN` too.

legendtkl:
> Hi, can this pr fix the problem `Could not execute SQL statement. Reason: org.apache.flink.table.api.TableException: DELETE statement is not supported for streaming mode now.`?

@klion26 , no, the exception is threw by Flink TableEnv.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkTableSink.java
```diff
@@ -112,17 +132,95 @@ public RowLevelUpdateInfo applyRowLevelUpdate(
...
+                        String.format("%s=%s", leaf.fieldName(), leaf.literals().get(0))
+                    };
+
+            Optional<Action> action = DropPartitionAction.create(args);
+            try {
+                action.get().run();
+            } catch (Exception e) {
+                throw new RuntimeException(
+                        String.format("DropPartitionAction %s failed.", Arrays.toString(args)), e);
+            }
```
yuzelin, line 204: I think you can directly use `commit.dropPartitions` instead of building an action.

paimon-core/src/main/java/org/apache/paimon/table/TableUtils.java
```diff
@@ -40,10 +40,11 @@ public class TableUtils {
      *
      * <p>NOTE: This method is only suitable for deletion of small amount of data.
      */
-    public static void deleteWhere(Table table, List<Predicate> filters) {
+    public static long deleteWhere(Table table, List<Predicate> filters) {
```
yuzelin, line 43: Add `@return`.

yuzelin (commented):
Thanks for contributing, left some comments.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkTableSink.java
legendtkl: hi, @yuzelin , do you mean 
```code
action = DropPartitionAction(xx)
action.run()
```

But in this case, we need to parse the parameters for DropPartitionAction construct method. This might cause some duplicated code and some complicated cases.

legendtkl:
Thanks for the review, @JingsongLi @yuzelin.

I have addressed the comment and leave some comments from my view.

Would you help review it again when you're available? thanks.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkTableSink.java
yuzelin: I mean you can:
1. get `FileStoreCommit` from table;
2. call `FileStoreCommit#dropPartitions` directly.

legendtkl: > I mean you can:
> 
> 1. get `FileStoreCommit` from table;
> 2. call `FileStoreCommit#dropPartitions` directly.

@yuzelin  got it. Resolved.

yuzelin (approved):
+1

JingsongLi (commented):
Regarding the implementation of `DeletePushDownFunctionVisitor`, I feel that it is OK to return bool. For `AND`, as long as there is a true, it can return true; for `OR`, as long as there is a false, it returns false.

JingsongLi (changes requested):
-1, we can improve more.

paimon-common/src/main/java/org/apache/paimon/predicate/DeletePushDownFunctionVisitor.java
```diff
@@ -0,0 +1,128 @@
...
+
+    @Override
+    public Boolean visitLessOrEqual(FieldRef fieldRef, Object literal) {
+        return false;
+    }
+
+    @Override
+    public Boolean visitEqual(FieldRef fieldRef, Object literal) {
+        if (partitionKeys.contains(fieldRef.name())) {
+            if (predicatesSize == 1) {
```
JingsongLi, line 77: Why?

paimon-common/src/main/java/org/apache/paimon/predicate/DeletePushDownFunctionVisitor.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    @Override
+    public Boolean visitNotIn(FieldRef fieldRef, List<Object> literals) {
+        return false;
+    }
+
+    @Override
+    public Boolean visitAnd(List<Boolean> children) {
+        if (children.size() != 2) {
```
JingsongLi, line 107: Why check 2?

paimon-common/src/main/java/org/apache/paimon/predicate/DeletePushDownFunctionVisitor.java
```diff
@@ -0,0 +1,128 @@
...
+        }
+        if (children.get(0) && children.get(1)) {
+            return true;
+        }
+        return false;
+    }
+
+    @Override
+    public Boolean visitOr(List<Boolean> children) {
+        if (children.size() != 2) {
```
JingsongLi, line 118: Why check 2?

paimon-common/src/main/java/org/apache/paimon/predicate/DeletePushDownFunctionVisitor.java
legendtkl: @JingsongLi  partition key filter could be push down when there is only one filter, such as `when partition key = xx`. When the partition key filter is mixed with others, do not push down.

paimon-common/src/main/java/org/apache/paimon/predicate/DeletePushDownFunctionVisitor.java
legendtkl: follow legacy code, can remove it

paimon-common/src/main/java/org/apache/paimon/predicate/DeletePushDownFunctionVisitor.java
legendtkl: follow legacy code, can remove it

JingsongLi:
I feel like there are still some issues:
- For the second visitor, what you need to determine is whether all pk (pk including partition keys) equivalence filters are included, others filters can be ignore.
- Does ITCase seem to be of little use for testing? What do you need is UT to verify if it has been pushed down?

JingsongLi (approved):
+1

PR #1436, opened 2023-06-26
[flink] Run flink continuous file store IT Cases with parameterized changelog option

schnappi17:
### Purpose
Run flink continuous file store IT Cases with parameterized changelog option, remove the double coded method with `changelog = true` and `changelog = false`


schnappi17:
@JingsongLi Re-write the  ContinuousFileStoreITCase for flink 1.14 / 1.15, run the tests with parameterized changelog options instead of double writing test method codes. Please help to review this pr, thanks a lot~

PR #1439, opened 2023-06-27
[doc] Add a description explaining that DELETE needs to be run in the batch mode

klion26:
### Purpose

Add a description to explaining that DELETE needs to be run in the batch mode

### Tests

no code change

### API and Format

no API/Format change

### Documentation

Change the documentation

### DIFF
--- a/docs/content/concepts/file-operations.md
+++ b/docs/content/concepts/file-operations.md
@@ -43,10 +43,6 @@ following sections:
 2. [File Layouts]({{< ref "concepts/file-layouts" >}}) and 
 3. How to use Paimon in [Flink]({{< ref "engines/flink" >}}).
 
-```
-Please execute the code in batch mode if there is a {{< label Batch >}} label above it, and you can add an entry `execution.runtime-mode: batch` in flink-conf.yaml to enable batch mode.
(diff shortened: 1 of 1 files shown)


klion26:
Hi @JingsongLi , Please help to have a look when you're free, thanks.

Add the label above the code, because that it seems strange after adding label in the code block(such as below)

image

docs/content/concepts/file-operations.md
```diff
@@ -43,6 +43,10 @@ following sections:
 2. [File Layouts]({{< ref "concepts/file-layouts" >}}) and 
 3. How to use Paimon in [Flink]({{< ref "engines/flink" >}}).
 
+```
+Please execute the code in batch mode if there is a {{< label Batch >}} label above it, and you can add an entry `execution.runtime-mode: batch` in flink-conf.yaml to enable batch mode.
```
JingsongLi, line 47: I think we don't need this.

klion26: Ok, I can revert this.

klion26:
@JingsongLi thanks for the review and merging!

PR #1440, opened 2023-06-27
[cdc] Introduce RichCdcSinkBuilder API

JingsongLi:
### Purpose

The previous CDC API, which required the user to implement an EventParser.Factory, was a bit too complicated for the user.

This PR introduces a RichCdcSinkBuilder, the user only needs to provide a DataStream<RichCdcRecord>, which can extract Schema changes from the data.


yuzelin (approved):
+1

PR #1442, opened 2023-06-28
[core] Query audit_log with incremental between should see DELETE rec…

JingsongLi:
…ords

### Purpose

For increment-between:
In Batch SQL, the `DELETE` records are not allowed to be returned, so records of `-D` will be dropped.
If you want see `DELETE` records, you can use audit_log table:

```sql
SELECT * FROM t$audit_log /*+ OPTIONS('incremental-between' = '12,20') */;
```


yuzelin (approved):
+1

PR #1443, opened 2023-06-28
[core] Support rollback to a tag

yuzelin:
### Purpose

Add roll back to tag methods to table.

### Tests

`FileStoreTableTestBase#testRollbackToTag`
`RollbackToActionITCase#rollbackToTagTest`

### API and Format

`Table.rollbackTo(String tagName)`


paimon-core/src/main/java/org/apache/paimon/operation/DeletionUtils.java
```diff
@@ -144,6 +144,10 @@ public static Set<String> collectManifestSkippingSet(
                 snapshot.dataManifests(manifestList).stream()
                         .map(ManifestFileMeta::fileName)
                         .collect(Collectors.toCollection(HashSet::new));
+
+        skip.add(snapshot.baseManifestList());
```
JingsongLi, line 148: You should remove FileStoreExpireImpl:
```
skipManifestFiles.add(snapshot.baseManifestList());
skipManifestFiles.add(snapshot.deltaManifestList());
```

paimon-core/src/main/java/org/apache/paimon/operation/TagDeletion.java
```diff
@@ -29,22 +29,16 @@
...
 import java.util.Set;
 
 import static org.apache.paimon.operation.DeletionUtils.addMergedDataFiles;
-import static org.apache.paimon.operation.DeletionUtils.collectManifestSkippingSet;
 import static org.apache.paimon.operation.DeletionUtils.containsDataFile;
 import static org.apache.paimon.operation.DeletionUtils.readEntries;
-import static org.apache.paimon.operation.DeletionUtils.tryDeleteDirectories;
 
 /**
  * Delete tag files. This class doesn't check changelog files because they are handled by {@link
```
JingsongLi, line 44: Maybe we can try to create a abstract class for TagDeletion and SnapshotDeletion.

paimon-core/src/main/java/org/apache/paimon/utils/TagManager.java
```diff
@@ -119,6 +157,65 @@ public SortedMap<Snapshot, String> tags() {
         return tags;
     }
 
+    /** Rollback to tag. */
+    public void rollbackTo(String tagName, Snapshot latestSnapshot, TagDeletion tagDeletion) {
```
JingsongLi, line 161: Introduce a real `rollbackTo(long snapshotId)`, delete all tags and all snapshots bigger than it.

paimon-core/src/main/java/org/apache/paimon/table/AbstractFileStoreTable.java
```diff
@@ -290,19 +297,151 @@ public void createTag(String tagName, long fromSnapshotId) {
...
+            throw new UncheckedIOException(e);
+        }
+    }
+
+    /**
+     * Clean snapshots and tags whose id is larger than given snapshot's in the reverse direction.
+     *
+     * <p>Snapshots are cleaned first because it is easier to collect their deletion skipping set.
+     */
+    private void cleanLargerThan(Snapshot snapshot) {
```
JingsongLi, line 338: Create a separate class to contain `cleanLargerThan` method. For example, a `RollbackHelper` class.

paimon-core/src/main/java/org/apache/paimon/table/AbstractFileStoreTable.java
```diff
@@ -290,19 +297,151 @@ public void createTag(String tagName, long fromSnapshotId) {
...
+
+        SortedMap<Snapshot, String> tags = tagManager.tags();
+        if (tags.isEmpty()) {
+            return;
+        }
+
+        // delete tag files and collect skipping set
+        skippedSnapshots.clear();
+        toBeCleaned.clear();
+        for (int i = taggedSnapshots.size() - 1; i >= 0; i--) {
```
JingsongLi, line 402: `skippedSnapshots` should just be a snapshot.

paimon-core/src/main/java/org/apache/paimon/operation/SnapshotDeletion.java
```diff
@@ -224,46 +92,53 @@ void doDeleteExpiredDataFiles(
...
 
-    public Set<String> collectManifestSkippingSet(Snapshot snapshot) {
-        return DeletionUtils.collectManifestSkippingSet(snapshot, manifestList, indexFileHandler);
+    /** Used to record which tag is cached in tagged snapshots list. */
+    private int cachedTagIndex = -1;
+
+    /** Used to cache data files used by current tag. */
+    private final Map<BinaryRow, Map<Integer, Set<String>>> cachedTagDataFiles = new HashMap<>();
+
+    public Predicate<ManifestEntry> dataFileSkipper(
```
JingsongLi, line 132: You can create a separate class `TagFileSkipper`.

paimon-core/src/main/java/org/apache/paimon/operation/FileStoreExpireImpl.java
```diff
@@ -174,32 +169,26 @@ public void expireUntil(long earliestId, long endExclusiveId) {
...
-        Set<String> skipManifestFiles =
-                snapshotDeletion.collectManifestSkippingSet(
-                        snapshotManager.snapshot(endExclusiveId));
-        for (Snapshot snapshot :
-                tagFileKeeper.findOverlappedSnapshots(beginInclusiveId, endExclusiveId)) {
-            skipManifestFiles.add(snapshot.baseManifestList());
-            skipManifestFiles.add(snapshot.deltaManifestList());
-            skipManifestFiles.addAll(snapshotDeletion.collectManifestSkippingSet(snapshot));
-        }
+        List<Snapshot> skippingSnapshots =
```
JingsongLi, line 181: Change `Set<String>` to `List<Snapshot>` has performance issue.

paimon-core/src/main/java/org/apache/paimon/table/AbstractFileStoreTable.java
```diff
@@ -290,19 +297,151 @@ public void createTag(String tagName, long fromSnapshotId) {
...
+            snapshotDeletion.deleteAddedDataFiles(s.deltaManifestList());
+            snapshotDeletion.deleteAddedDataFiles(s.changelogManifestList());
+        }
 
+        // delete directories
+        snapshotDeletion.cleanDataDirectories();
+
+        // delete manifest files
+        skippedSnapshots.add(snapshot);
+        // NOTE: must skip tag manifests because tag deletion will read them
```
JingsongLi, line 379: Why? You can have a unified loop.

paimon-core/src/main/java/org/apache/paimon/utils/TagManager.java
```diff
@@ -77,6 +80,40 @@ public void createTag(Snapshot snapshot, String tagName) {
         }
     }
 
+    public void deleteTag(
+            String tagName, TagDeletion tagDeletion, SnapshotManager snapshotManager) {
+        checkArgument(!StringUtils.isBlank(tagName), "Tag name '%s' is blank.", tagName);
+        checkArgument(tagExists(tagName), "Tag '%s' doesn't exist.", tagName);
+
+        Snapshot taggedSnapshot = taggedSnapshot(tagName);
+        List<Snapshot> taggedSnapshots = taggedSnapshots();
```
JingsongLi, line 89: move `taggedSnapshots` down after `if (snapshotManager.snapshotExists)`.

paimon-core/src/main/java/org/apache/paimon/utils/TagManager.java
```diff
@@ -77,6 +80,40 @@ public void createTag(Snapshot snapshot, String tagName) {
...
+        fileIO.deleteQuietly(tagPath(tagName));
+
+        // skip file deletion if snapshot exists
+        if (snapshotManager.snapshotExists(taggedSnapshot.id())) {
+            return;
+        }
+
+        // collect skipping sets from the earliest snapshot and neighbor tags
+        List<Snapshot> skippedSnapshots = new ArrayList<>();
+        skippedSnapshots.add(snapshotManager.earliestSnapshot());
```
JingsongLi, line 99: Here we just need to add one or two snapshots. Never three snapshots.

paimon-core/src/main/java/org/apache/paimon/operation/SnapshotDeletion.java
yuzelin: It will use some methods in FileDeletionBase, so I think it's not necessary to move them out.

paimon-core/src/main/java/org/apache/paimon/utils/TagManager.java
yuzelin: Cannot do it because delete the tag file will make the taggedSnapshots return wrong result.

paimon-core/src/main/java/org/apache/paimon/table/RollbackHelper.java
```diff
@@ -0,0 +1,191 @@
...
+            // store tag data manifest entries to avoid redundant reading when cleaning tag
+            List<ManifestEntry> base = new ArrayList<>();
+            snapshotDeletion
+                    .tryReadManifestEntries(snapshot.baseManifestList())
+                    .iterator()
+                    .forEachRemaining(base::add);
+
+            List<ManifestEntry> data = new ArrayList<>(base);
+            data.addAll(delta);
+            tagManifestEntries.put(snapshot, data);
```
JingsongLi, line 136: Why store delta too? We just need base files.

paimon-core/src/main/java/org/apache/paimon/table/RollbackHelper.java
```diff
@@ -0,0 +1,191 @@
...
+
+    // don't concern about tag data files because file deletion methods won't throw exception
+    // when deleting non-existing data files
+    private void cleanSnapshotDataFiles(Snapshot snapshot) {
+        List<ManifestEntry> delta = new ArrayList<>();
+        snapshotDeletion
+                .tryReadManifestEntries(snapshot.deltaManifestList())
+                .iterator()
+                .forEachRemaining(delta::add);
+        if (tags.containsKey(snapshot)) {
```
JingsongLi, line 126: Maybe we can just let it be a good tag, we don't need to delete manifest and datafiles.

paimon-core/src/main/java/org/apache/paimon/utils/TagManager.java
JingsongLi: You can optimize it too:
```
if (snapshotManager.snapshotExists(taggedSnapshot.id())) {
    fileIO.deleteQuietly(tagPath(tagName)); 
    return;
} else {
    // document why this should be first.
    List<Snapshot> taggedSnapshots = taggedSnapshots();
    fileIO.deleteQuietly(tagPath(tagName)); 
}
```

JingsongLi (approved):
+1

Issue #864, opened 2023-04-10
[Bug] Paimon Table produce more reduce task for same data volume


JingsongLi:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/incubator-paimon/issues) and found nothing similar.

### Paimon version

0.4

### Compute Engine

hive

### Minimal reproduce step

Same data volume.
- read Hive table: 50 reducers.
- read paimon table: 300 reducers.

### What doesn't meet your expectations?

read paimon table should have lower reducer number.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

calvinjiang:
I'd like to fix this issue.

JingsongLi:
> I'd like to fix this issue.

How to?

wg1026688210:
hi~ @JingsongLi   Is it because there are more files in the Paimon table.  Is it  effective  to reduce the number of reduce tasks, If we set number of  the upstream map by `mapred.map.tasks` ,which reduce the number of shuffle files.

JingsongLi:
> hi~ @JingsongLi Is it because there are more files in the Paimon table. Is it effective to reduce the number of reduce tasks, If we set number of the upstream map by `mapred.map.tasks` ,which reduce the number of shuffle files.

We should figure out what is the mechanism of the task number inference in Hive, and try to work around.

PR #1445, opened 2023-06-28
[hive] Support syncing partition into Hive metastore when using Hive catalog

tsreaper:
### Purpose

Currently we cannot create partitioned Paimon table in Hive catalog. In Hive, all fields are considered as normal fields and there is no partition key. It brings us at least two downsides:

1. Users cannot list all partitions through the `SHOW PARTITIONS` command.
2. Hive's native partition pushdown cannot be performed. Although currently we approach partition pushdown through Hive's filter pushdown, some Hive compatible system may only support partition pushdown and not filter pushdown, making Paimon nearly unusable in those systems.

This PR introduces a feature to synchronize partitions into Hive metastore when using Hive catalog. In this PR, users can create partitioned Hive table through Flink and Spark using Hive catalog. Creating partitioned table directly in Hive is currently not supported and will be introduced in the next PR.

### Tests

* `TableCommitTest`
* `HiveCatalogITCaseBase#testAddPartitionsToMetastore`

### API and Format

No.

### Documentation

Yes. Document will also be introduced in the next PR.

### DIFF
--- a/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/AddPartitionCommitCallback.java
+++ b/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/AddPartitionCommitCallback.java
@@ -1,0 +1,141 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.paimon.hive;
+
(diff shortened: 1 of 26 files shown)


Alibaba-HZY:
Very useful pr, yesterday I was discussing with colleagues, how to synchronize partition information to hms

wg1026688210:
nice done , I also have thought about the solution of inferring the reduce quantity ，this may fix  #864

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/AddPartitionCommitCallback.java
```diff
@@ -0,0 +1,141 @@
...
+            this.sd =
+                    client.getTable(identifier.getDatabaseName(), identifier.getObjectName())
+                            .getSd();
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    @Override
+    public void call(List<ManifestCommittable> committables) {
```
wg1026688210, line 72: How about adding  options that  add partition  by comparing watermark and parition time.

JingsongLi:
We need a catalog to register for callbacks here, but we also have a requirement to add separate callbacks to the table. We need to reconsider all these requirements and summarize the modifications that should be made to the `paimon-core` in the end.

JingsongLi:
Some ideas about callback:
1. Callback should belong to the table, and it should be parsed from options
2. Callbacks can be parsed from options and can have a mode that includes some built-in callbacks, as well as custom class methods. Additionally, we can provide parameter configurations for custom classes.
3. The catalog should provide a metastore API abstraction for tables, which can complete the adding partition from the corresponding callback.

tsreaper:
I'm going to split this PR into two: introduce (custom) callback and introduce metastore client for table. Closing this PR.

PR #1446, opened 2023-06-28
[core] Added validation of bucket keys when creating a table.

zhuangchong:
### Purpose

Added validation of bucket keys when creating a table.


JingsongLi (approved):
+1

PR #1450, opened 2023-06-28
 [flink] supports alignment with flink checkpoint when consuming paimon snapshots.

### LINKED ISSUE #1486: [Feature] Align snapshot and checkpoint in split enumerator and source
### Search before asking

### Motivation

subtask for #1105 , @liming30 I create this issue for alignment between split enumerator and source reader, please feel free to create subtask for it.

### Solution

_No response_

### Anything else?

_No response_

liming30:
### Purpose
 [flink] supports alignment with flink checkpoint when consuming paimon snapshots. closes #1486 

As a subtask of [PIP-5](https://cwiki.apache.org/confluence/x/jRE0Dw).

### Tests

- Added `AlignedEnumeratorTest` to verify that `AlignedEnumerator` will send `CheckpointEvent` during checkpoint.
- Added `AlignedSourceReaderTest` to verify that `AlignedSourceReader` can correctly send split, create checkpoint, and restore in different `CheckpointAlignMode`.

### API and Format

No.

### Documentation

https://cwiki.apache.org/confluence/x/jRE0Dw

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/align/AlignedContinuousFileSource.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/align/AlignedContinuousFileSource.java
@@ -1,0 +1,99 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.paimon.flink.source.align;
+
+import org.apache.paimon.CoreOptions;
+import org.apache.paimon.flink.FlinkConnectorOptions;
+import org.apache.paimon.options.Options;
+import org.apache.paimon.table.source.ReadBuilder;
+import org.apache.paimon.table.source.Split;
+
+import org.apache.flink.api.connector.source.Boundedness;
+import org.apache.flink.api.connector.source.Source;
+import org.apache.flink.api.connector.source.SourceReader;
+import org.apache.flink.api.connector.source.SourceReaderContext;
+import org.apache.flink.api.connector.source.SplitEnumerator;
+import org.apache.flink.api.connector.source.SplitEnumeratorContext;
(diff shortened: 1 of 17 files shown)


liming30:
@JingsongLi @FangYongs Based on Flip-27, I implemented a draft that aligns with flink checkpoint when consuming paimon snapshots. Please help review it when you have time.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/align/AlignedContinuousFileSource.java
```diff
@@ -0,0 +1,99 @@
...
+ * <p>The difference with {@link org.apache.paimon.flink.source.ContinuousFileStoreSource} is that
+ * it will coordinate flink's checkpoint and paimon snapshot.
+ */
+public class AlignedContinuousFileSource implements Source<Split, AlignedSourceSplit, Void> {
+
+    private static final long serialVersionUID = 1L;
+
+    protected final ReadBuilder readBuilder;
+
+    private final boolean emitSnapshotWatermark;
```
JingsongLi, line 49: Wow, there is `emit watermark` too. 👍🏻

JingsongLi (commented):
Can other Aligned classes just like `AlignedEnumerator`? Just to wrap another class.

liming30:
> Can other Aligned classes just like `AlignedEnumerator`? Just to wrap another class.

Sorry, I didn't find some other code that can be reused. The logic implementation of this code is quite different from other implementation.  The closest implementation is `MonitorFunction`, but the two classes use completely different parent interfaces.

liming30:
@JingsongLi @FangYongs I have added related tests, please review it if you have time.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/align/AlignedSourceReader.java
```diff
@@ -0,0 +1,298 @@
...
+        this.emitSnapshotWatermark = emitSnapshotWatermark;
+        this.executors = executors;
+        this.snapshotIdByNextCheckpoint = null;
+        this.blocking = false;
+        this.endOfScan = false;
+    }
+
+    @Override
+    public void start() {
+        this.scan = readBuilder.newStreamScan();
```
JingsongLi, line 114: Can splits just be send to here? Distributed scan I feel is still a little bit not so good from the design point of view

liming30: If the splits are still scanned and assigned by the `Enumerator`, there is a problem: in the `strictly-align` mode, if the source table has not produced the corresponding snapshot at the time of the checkpoint, then the checkpoint will wait until the timeout.

JingsongLi: But we can also send a event to reader to show that there is no split.

liming30: So what should the reader do for this checkpoint? directly abort? This may cause unnecessary checkpoint failures.

PR #1452, opened 2023-06-29
[core] add support of stats mode

Aitozi:
Related issue: https://github.com/apache/incubator-paimon/issues/1375

> We should avoid adding statistics to very large columns, which can cause our Manifest file to become very large.

Related PR: https://github.com/apache/incubator-paimon/pull/1365

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/statistics/Stats.java
+++ b/paimon-common/src/main/java/org/apache/paimon/statistics/Stats.java
@@ -1,0 +1,70 @@
+/*
+ *
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+
+package org.apache.paimon.statistics;
+
+import org.apache.paimon.data.serializer.Serializer;
+import org.apache.paimon.format.FieldStats;
+
+import java.util.regex.Matcher;
+
+import static org.apache.paimon.statistics.TruncateStats.TRUNCATE_PATTERN;
+
+/** The mode of the Stats */
+public interface Stats {
+
+    /**
+     * collect stats from the field.
+     *
+     * @param field The target field object.
+     * @param fieldSerializer The serializer of the field object.
+     */
+    void collect(Object field, Serializer<Object> fieldSerializer);
+
+    /** @return The collected field stats. */
+    FieldStats result();
+
+    /**
+     * Convert the field stats according to the strategy.
+     *
+     * @param source The source field stats, extracted from the file.
+     * @return The converted field stats.
+     */
+    FieldStats convert(FieldStats source);
+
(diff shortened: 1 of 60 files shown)


paimon-common/src/main/java/org/apache/paimon/statistics/TruncateStats.java
wg1026688210, line 1: We may need consider the filter pushdown when reading.

Aitozi: I'm not sure why this will affect the filter down? IMO, this only cut the stats when writing to the metadata file.

wg1026688210: Paimon will pick up the datafile by the stats of manifest entry when scanning table.  You can see  `AbstractFileStoreScan#filterManifestFileMeta`.

Aitozi: Get your meaning now, will revisit this part.

Aitozi: If we truncate a min value, the new value is <= the old value. If we truncate a max value, it's also <= the old value. We should ensure the truncated new value is >= the old value by increase one to the value.

In this way, the new stats's scope is larger than the old stats's scope, it may lead to a broader scan list but will not affect the result now.

paimon-common/src/main/java/org/apache/paimon/statistics/TruncateStats.java
```diff
@@ -0,0 +1,127 @@
...
+                return field;
+            }
+
+            StringBuilder truncatedStringBuilder = new StringBuilder(truncated.toString());
+
+            // Try incrementing the code points from the end
+            for (int i = length - 1; i >= 0; i--) {
+                // Get the offset in the truncated string buffer where the number of unicode
+                // characters = i
+                int offsetByCodePoint = truncatedStringBuilder.offsetByCodePoints(0, i);
```
JingsongLi, line 112: Is there some ut cases for this?

Aitozi: StatsTest#testTruncateTwoChar

paimon-core/src/main/java/org/apache/paimon/utils/StatsUtils.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.List;
+
+import static org.apache.paimon.CoreOptions.FIELD_STATS_MODE_PREFIX;
+import static org.apache.paimon.CoreOptions.FIELD_STATS_MODE_SUFFIX;
+import static org.apache.paimon.options.ConfigOptions.key;
+
+/** The stats utils. */
+public class StatsUtils {
+
+    public static Stats[] getFieldsStatsMode(CoreOptions options, List<String> fields) {
```
JingsongLi, line 36: Can this method just in `Stats` class?

JingsongLi (commented):
Can we document this? maybe in `CREATE TABLE`?

paimon-common/src/main/java/org/apache/paimon/statistics/Stats.java
```diff
@@ -0,0 +1,70 @@
...
+
+import org.apache.paimon.data.serializer.Serializer;
+import org.apache.paimon.format.FieldStats;
+
+import java.util.regex.Matcher;
+
+import static org.apache.paimon.statistics.TruncateStats.TRUNCATE_PATTERN;
+
+/** The mode of the Stats */
+public interface Stats {
```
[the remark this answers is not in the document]
Aitozi: done

paimon-core/src/main/java/org/apache/paimon/utils/StatsUtils.java
Aitozi: It's mainly for the access of `CoreOptions` and `Options`

Aitozi:
> Can we document this? maybe in `CREATE TABLE`?

Added

JingsongLi (approved):
Thanks @Aitozi for the contribution!

Looks good to me!

PR #1453, opened 2023-06-29
[fix] rename it case tot ITCase

leaves12138:
`KafkaSchemaTest` is a IT Test


JingsongLi (approved):
+1

JingsongLi:
Titile should be `[flink] rename KafkaSchemaTest case to ITCase`.

PR #1455, opened 2023-06-29
[fix] disable compaction when unaware-bucket table writes in batch mode

leaves12138:
Unaware-bucket table inserting in batch mode does not need compaction.

Add Tests for this.


JingsongLi (approved):
+1

PR #1459, opened 2023-06-29
[spark] SQL: supports INSERT OVERWRITE syntax

YannByron:
### Purpose

Supports `INSERT OVERWRITE table (PARTITION (...)) ...` in Spark SQL

Linked issue: close #xxx

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/commands/PaimonCommand.scala
+++ b/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/commands/PaimonCommand.scala
@@ -17,8 +17,11 @@
  */
 package org.apache.paimon.spark.commands
 
+import org.apache.paimon.predicate.PredicateBuilder
+import org.apache.paimon.spark.SparkFilterConverter
 import org.apache.paimon.table.{BucketMode, FileStoreTable, Table}
 import org.apache.paimon.table.sink.{CommitMessage, CommitMessageSerializer}
+import org.apache.paimon.types.RowType
 
 import org.apache.spark.sql.catalyst.analysis.Resolver
 import org.apache.spark.sql.sources.{AlwaysTrue, And, EqualNullSafe, Filter, Not, Or}
@@ -61,19 +64,15 @@ trait PaimonCommand {
    * For the 'INSERT OVERWRITE T PARTITION (partitionVal, ...)' semantics of SQL, Spark will
    * transform `partitionVal`s to EqualNullSafe Filters.
    */
-  def convertFilterToMap(
-      filter: Filter,
-      partitionColumns: Seq[String],
-      resolver: Resolver): Map[String, String] = {
+  def convertFilterToMap(filter: Filter, partitionRowType: RowType): Map[String, String] = {
+    val converter = new SparkFilterConverter(partitionRowType)
     splitConjunctiveFilters(filter).map {
       case EqualNullSafe(attribute, value) =>
         if (isNestedFilterInValue(value)) {
           throw new RuntimeException(
             s"Not support the complex partition value in EqualNullSafe when run `INSERT OVERWRITE`.")
         } else {
-          partitionColumns.find(resolver(_, attribute)).map((_, value.toString)).getOrElse {
-            throw new RuntimeException(s"this $attribute is not recognized in partition columns.")
-          }
+          (attribute, converter.convertLiteral(attribute, value).toString)
         }
       case _ =>
         throw new RuntimeException(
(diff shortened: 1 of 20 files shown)


paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/SparkTypeUtils.java
```diff
@@ -173,8 +173,7 @@ public DataType visit(RowType rowType) {
             List<StructField> fields = new ArrayList<>(rowType.getFieldCount());
             for (DataField field : rowType.getFields()) {
                 StructField structField =
-                        DataTypes.createStructField(
-                                field.name(), field.type().accept(this), field.type().isNullable());
+                        DataTypes.createStructField(field.name(), field.type().accept(this), true);
```
YannByron, line 176: This change is for https://github.com/apache/incubator-paimon/issues/1048. Should submit another pr for this, just leave it for the pr's tests.

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/commands/PaimonCommand.scala
```diff
@@ -44,4 +47,50 @@ trait PaimonCommand {
...
+      filter: Filter,
+      partitionColumns: Seq[String],
+      resolver: Resolver): Map[String, String] = {
+    splitConjunctiveFilters(filter).map {
+      case EqualNullSafe(attribute, value) =>
+        if (isNestedFilterInValue(value)) {
+          throw new RuntimeException(
+            s"Not support the complex partition value in EqualNullSafe when run `INSERT OVERWRITE`.")
+        } else {
+          partitionColumns.find(resolver(_, attribute)).map((_, value.toString)).getOrElse {
```
JingsongLi, line 74: We should use `PredicateBuilder.convertJavaObject` to convert spark java object to our internal data structure.

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/commands/WriteIntoPaimonTable.scala
```diff
@@ -113,7 +120,9 @@ case class WriteIntoPaimonTable(table: FileStoreTable, overwrite: Boolean, data:
         .map(deserializeCommitMessage(serializer, _))
 
     try {
-      val tableCommit = writeBuilder.newCommit().asInstanceOf[InnerTableCommit]
+      val tableCommit = writeBuilder
+        .withOverwrite(overwritePartition.asJava)
```
JingsongLi, line 124: if there is no overwrite, we should not invoke `withOverwrite(emptyMap)` here?
I don't understand why tests passing.

paimon-spark/paimon-spark-common/src/test/scala/org/apache/paimon/spark/sql/InsertOverwriteSuite.scala
```diff
@@ -0,0 +1,173 @@
...
+ * limitations under the License.
+ */
+package org.apache.paimon.spark.sql
+
+import org.apache.paimon.WriteMode.{APPEND_ONLY, CHANGE_LOG}
+
+import org.apache.spark.sql.Row
+import org.scalactic.source.Position
+
+class InsertOverwriteSuite extends PaimonSparkTestBase {
```
JingsongLi, line 25: in our code style, we use xxTest instead of Suite

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/commands/WriteIntoPaimonTable.scala
YannByron: if no overwrite here, `overwritePartition` is actually `null`, that is consistent with the behavior of `insert into`.

paimon-spark/paimon-spark-common/src/test/scala/org/apache/paimon/spark/sql/InsertOverwriteSuite.scala
YannByron: OK

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/commands/PaimonCommand.scala
YannByron: OK, and i'll add more UTs which use the different data types as the partition columns.

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/commands/WriteIntoPaimonTable.scala
JingsongLi: I get it, you are right.

JingsongLi (approved):
Thanks @YannByron , +1 for this.

YannByron:
This pr just supports `static insert overwrite`, `dynamic overwrite` is under planning.

PR #1463, opened 2023-06-30
[spark] add warning message when using InMemoryCatalog in SparkGenericCatalog

LsomeYeah:
### Purpose

If the implementation of `sessionCatalog` in `SparkGenericCatalog` is `InMemoryCatalog`, it may create the table which has the same name with the table created by `paimonCatalog `.  So we apply a distinct operation for the result of `listTables` in `SparkGenericCatalog`.

Linked issue: close #xxx


JingsongLi (approved):
+1

PR #1464, opened 2023-06-30
[core] Change default sort engine back to min heap

tsreaper:
### Purpose

We've heard from some users that they experience some issues when using loser tree as the sort engine. This PR changes the default sort engine back to `min-heap`, which is more stable.

### Tests

N/A

### API and Format

No.

### Documentation

Document for `sort-engine` option is also changed.


JingsongLi (approved):
+1

PR #1465, opened 2023-06-30
[flink] Prohibit adaptive parallelism

JingsongLi:
### Purpose

Paimon Sink does not support Flink's Adaptive Parallelism mode. Please manually turn it off or set Paimon `sink.parallelism` manually.


leaves12138:
+1

PR #1466, opened 2023-06-30
[common] Remove HdfsConfiguration dependency

yuzelin:
### Purpose

This class is in different package of hadoop 2.8.5 and 3.3.4 which may cause exception in different enviroment.


JingsongLi:
I found that we don't need to change this, there's no problem with adding this dependency

PR #1469, opened 2023-06-30
[Flink]update union list state to avoid  inflation

wg1026688210:
### Purpose
fix the list state which lead that jm heap oom


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/StateUtils.java
```diff
@@ -58,10 +60,15 @@ public class StateUtils {
...
-            state.add(defaultValue);
+        }
+
+        if (values.isEmpty()){
             values.add(defaultValue);
         }
 
-        return values.get(0);
+        T value = values.get(0);
+        state.update(Lists.newArrayList(value));
```
JingsongLi, line 70: we need to an ut case to cover this.

JingsongLi (approved):
+1

PR #1477, opened 2023-07-02
[WIP][Flink]add dry run arg for dropping parition

wg1026688210:
add dry run for dropping parition to meet two conditions:

1. Users are unsure if their dropped partition meets expectations
2. Whether it is a valid drop (the partition has already been dropped, but some residual files have not been completely deleted , as task restart when delete file )


JingsongLi:
First, there is a rollback for regret.
And, you can SELECT * FROM T$files WHERE xxx to query files before dropping partitions.

PR #1478, opened 2023-07-03
[doc] Modify query system tag table name error

zhuangchong:
### Purpose

Modify query system tag table name error


JingsongLi (approved):
+1

PR #1487, opened 2023-07-03
[core] Add base commit user and cause to commit conflict messages, also add back pressure as a possible cause for commit conflicts

tsreaper:
### Purpose

We heard from some users that they're experiencing commit conflicts although they're sure that there is only one job writing into the Paimon table. To better diagnose the problem, this PR adds base commit user and cause to commit conflict messages.

Update: after testing, we found that back pressuring is also a cause for commit conflicts. Conflicts will be automatically resolved and the exception won't be repeated, so it is harmless. This PR also adds back pressure as a possible cause for commit conflicts in the exception message.

### Tests

### API and Format

No.

### Documentation

No.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
@@ -867,10 +867,18 @@ public class FileStoreCommitImpl implements FileStoreCommit {
                         "\n",
                         "Conflicts during commits are normal and this failure is intended to resolve the conflicts.",
                         "Conflicts are mainly caused by the following scenarios:",
-                        "1. Multiple jobs are writing into the same partition at the same time, you can use "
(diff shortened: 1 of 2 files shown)


paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
```diff
@@ -842,27 +847,46 @@ private void noConflictsOrFail(List<ManifestEntry> baseEntries, List<ManifestEnt
...
-            String message, List<ManifestEntry> baseEntries, List<ManifestEntry> changes) {
+            String message,
+            String baseCommitUser,
+            List<ManifestEntry> baseEntries,
+            List<ManifestEntry> changes,
+            Throwable cause) {
         String possibleCauses =
                 String.join(
                         "\n",
                         "Conflicts during commits are normal and this failure is intended to resolve the conflicts.",
```
JingsongLi, line 868: In the first line, we should let user be relax.

JingsongLi (approved):
+1

PR #1492, opened 2023-07-04
[core] Introduce commit callback for tables

tsreaper:
### Purpose

As discussed in #1445, in this PR we will first introduce `CommitCallback` for tables. Users can also create their own `CommitCallback` and use it through `commit.callback` option.

### Tests

* `TableCommitTest`.

### API and Format

No.

### Documentation

Yes. This PR comes with changes in document.


JingsongLi (approved):
+1

PR #1498, opened 2023-07-04
[core]the creation_time of FilesTable should be local date time

zhangjun0x01:
### Purpose

now，for the `creation_time` of `FilesTable` ，when writing, it is local time ,but we read it as UTC time, we should read it as local time.

the snapshots create time is correct.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/io/DataFileMeta.java
+++ b/paimon-core/src/main/java/org/apache/paimon/io/DataFileMeta.java
@@ -30,6 +30,7 @@ import org.apache.paimon.types.DataTypes;
 import org.apache.paimon.types.IntType;
 import org.apache.paimon.types.RowType;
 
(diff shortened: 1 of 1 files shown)


JingsongLi (commented):
The correct fixing should in `DataFileMeta`.
```
Timestamp.fromEpochMillis(System.currentTimeMillis())
```
Should be 
```
Timestamp.fromLocalDateTime(LocalDateTime.now())
```

zhangjun0x01:
> The correct fixing should in `DataFileMeta`.
> 
> ```
> Timestamp.fromEpochMillis(System.currentTimeMillis())
> ```
> 
> Should be
> 
> ```
> Timestamp.fromLocalDateTime(LocalDateTime.now())
> ```

I updated it follow your suggestion.

JingsongLi (approved):
+1

PR #1499, opened 2023-07-05
[core] Stats mode should not work for partition stats

JingsongLi:
### Purpose

Partition stats should always contains all stats.


leaves12138:
+1

PR #1500, opened 2023-07-05
[core] add file-format-per-level for paimon

zhangjun0x01:
### Purpose

Linked issue:  https://github.com/apache/incubator-paimon/issues/1474

### Tests

org.apache.paimon.flink.CatalogTableITCase#testFileFormatPerLevel


paimon-core/src/main/java/org/apache/paimon/io/KeyValueFileWriterFactory.java
```diff
@@ -90,11 +97,21 @@ public DataFilePathFactory pathFactory() {
...
+        updateFormatByLevel(level);
         return new RollingFileWriter<>(
                 () -> createDataFileWriter(pathFactory.newPath(), level, getCompression(level)),
                 suggestedFileSize);
     }
 
+    private void updateFormatByLevel(int level) {
+        String fileFormat = levelFormats.get(level);
+        if (null != fileFormat) {
+            pathFactory.setFormatIdentifier(fileFormat);
```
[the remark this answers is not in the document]
zhangjun0x01: I udpate it by construct a new `DataFilePathFactory`

paimon-core/src/main/java/org/apache/paimon/io/KeyValueFileWriterFactory.java
```diff
@@ -90,11 +97,21 @@ public DataFilePathFactory pathFactory() {
     }
 
     public RollingFileWriter<KeyValue, DataFileMeta> createRollingMergeTreeFileWriter(int level) {
+        updateFormatByLevel(level);
         return new RollingFileWriter<>(
                 () -> createDataFileWriter(pathFactory.newPath(), level, getCompression(level)),
                 suggestedFileSize);
     }
 
+    private void updateFormatByLevel(int level) {
```
JingsongLi, line 106: This style is not good because it dynamically decontaminates class member variables, which can lead to all sorts of problems.

In general, we recommend a functional style, which means that class members should ideally be immutable, with at most some cache stuff.

zhangjun0x01: I updated it.

JingsongLi (commented):
I understand that there is still some work needed here to make the code clearer, such as putting format related things into one class in write and introducing a LazyMap class to avoid creating all the structures.

zhangjun0x01:
> 

ok，I will deal it later

JingsongLi:
I added a commit [ef1af77](https://github.com/apache/incubator-paimon/pull/1500/commits/ef1af775db342df2d3376da55ec35af05e2f941e) to improve code structure

JingsongLi (approved):
+1

zhangjun0x01:
> I added a commit [ef1af77](https://github.com/apache/incubator-paimon/pull/1500/commits/ef1af775db342df2d3376da55ec35af05e2f941e) to improve code structure

thank you very much .

PR #1501, opened 2023-07-05
[core] Improve Avro read write performance

JingsongLi:
### Purpose

- Improve Avro read write performance by custom reader and writer
- Quick path for NONE stats mode


leaves12138:
+1

PR #1502, opened 2023-07-05
[Bug] Fix the parquet.compression do no work in parquet format

### LINKED ISSUE #1494: [Bug] file.compression do no work for parquet format
### Search before asking

### Paimon version

master

### Compute Engine

parquet format

### Minimal reproduce step

image

In parquet format, it will use `compress` key to extract from the option, and it has the higher preference than from the `file.compression`. So the actual work key is `compress`

the actual compression algorithm

### Anything else?

_No response_

Aitozi:
### Purpose

Linked issue: close #1494

### DIFF
--- a/paimon-format/src/main/java/org/apache/paimon/format/parquet/writer/RowDataParquetBuilder.java
+++ b/paimon-format/src/main/java/org/apache/paimon/format/parquet/writer/RowDataParquetBuilder.java
@@ -28,6 +28,8 @@ import org.apache.parquet.hadoop.ParquetWriter;
 import org.apache.parquet.hadoop.metadata.CompressionCodecName;
 import org.apache.parquet.io.OutputFile;
 
+import javax.annotation.Nullable;
+
 import java.io.IOException;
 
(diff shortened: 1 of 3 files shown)


paimon-format/src/main/java/org/apache/paimon/format/parquet/writer/RowDataParquetBuilder.java
```diff
@@ -45,15 +47,8 @@ public RowDataParquetBuilder(RowType rowType, Options conf) {
...
             throws IOException {
 
-        String compressName = CompressionCodecName.SNAPPY.name();
-        if (null != compression) {
-            compressName = compression;
-        }
-
         return new ParquetRowDataBuilder(out, rowType)
-                .withCompressionCodec(
-                        CompressionCodecName.fromConf(
```
Aitozi, line 55: In `org.apache.paimon.format.parquet.ParquetFileFormatFactory#supplyDefaultOptions` it will always set the `parquet.compression` to snappy. So the effective compression option is always snappy.

PR #1503, opened 2023-07-05
[Bug] Add the zstd to paimon-format artifact

### LINKED ISSUE #1496: [Bug] zstd compress is not available
### Search before asking

### Paimon version

master

### Compute Engine

format

### Minimal reproduce step

use compress zstd
 
image

do not work

### Anything else?

_No response_

Aitozi:
### Purpose

Linked issue: close #1496


paimon-format/pom.xml
```diff
@@ -298,6 +298,10 @@ under the License.
                                     <include>org.apache.parquet:parquet-format-structures</include>
                                     <include>org.apache.parquet:parquet-jackson</include>
                                     <include>commons-pool:commons-pool</include>
+
+                                    <!-- compress -->
+                                    <include>com.github.luben:zstd-jni</include>
+                                    <include>org.xerial.snappy:snappy-java</include>
```
Aitozi, line 304: The JNI class should not be relocated: https://github.com/luben/zstd-jni#limitations

paimon-format/pom.xml
```diff
@@ -298,6 +298,10 @@ under the License.
                                     <include>org.apache.parquet:parquet-format-structures</include>
                                     <include>org.apache.parquet:parquet-jackson</include>
                                     <include>commons-pool:commons-pool</include>
+
+                                    <!-- compress -->
+                                    <include>com.github.luben:zstd-jni</include>
```
JingsongLi, line 303: please update NOTICE file too.

paimon-format/pom.xml
JingsongLi, line 304: Do we need snappy? I think hadoop jars already have snappy?

Aitozi: Yes, I just think we should not rely on the hadoop jars. Removed now, we can add back if necessary.

paimon-format/pom.xml
Aitozi: Updated

paimon-format/src/main/resources/META-INF/NOTICE
```diff
@@ -26,6 +26,7 @@ This project bundles the following dependencies under the Apache Software Licens
 - org.apache.parquet:parquet-format-structures:1.12.3
 - org.apache.parquet:parquet-jackson:1.12.3
 - commons-pool:commons-pool:1.6
+- com.github.luben:zstd-jni:1.15.0-1
```
JingsongLi, line 29: This is not apache license: https://github.com/luben/zstd-jni/blob/master/LICENSE

Aitozi: fixed

JingsongLi (approved):
+1

PR #1507, opened 2023-07-06
[flink] ContinuousFileSplitEnumerator.assignSplits should be thread safe

JingsongLi:
### Purpose

Method should be synchronized because {@link #handleSplitRequest} and {@link #processDiscoveredSplits} have thread conflicts.


leaves12138:
+1

PR #1509, opened 2023-07-06
[fix] fix bug: raised ConcurrentModificationException while readers removed in SplitEnumeratorContext

leaves12138:
Raised ConcurrentModificationException in SplitEnumeratorContext if registered reader has been removed.


JingsongLi (approved):
+1

PR #1511, opened 2023-07-06
[fix] bugfix: use the same class load of object need to be copy in `CodeGeneratorContext`

leaves12138:
use the same class load of object need to be copy in `CodeGeneratorContext`


JingsongLi (commented):
Thanks @leaves12138 , but test is not good to contains binary bytes.

JingsongLi (approved):
+1

PR #1516, opened 2023-07-07
[flink] Provide scan.push-down to disable source pushdown

### LINKED ISSUE #1515: [Feature] reuse paimon source in flink job with statement set
### Search before asking

### Motivation

We expect to insert two mysql tables sequentially in a flink job with statement set, and need to reuse the paimon source in multiple insert SQL statements.

```
EXECUTE STATEMENT SET BEGIN
insert into svc1_mysql_test.gem_svc1_vpn.object select
  cus_id id,
  cus_id container,
  'B_CUSTOMER' class,
  case when isNew='新增' then localtimestamp
       else existing_cus_crtime
       end crtime,
  localtimestamp updtime,
  0 archived,
  0 seq,
  concat('B_CUSTOMER:', code, ' - ', name, ' - ', isNew) memo,
  'cq-ABSCustInfoServ_paimon.yaml' creator
from svc1_paimon_prod.cq.b_customer;
insert into svc1_mysql_test.gem_svc1_vpn.b_customer select
  cus_id id,
  name,
  code,
  customerlevel,
  prov
from svc1_paimon_prod.cq.b_customer;
END;
```

Actually if we use kafka as source, we can do this. The execution plan is like this
image

Currently using paimon as source, the execution plan is like this
image

### Solution

_No response_

### Anything else?

_No response_

JingsongLi:
### Purpose

close #1515


yuzelin (approved):
+1

PR #1517, opened 2023-07-07
[core] compacted_full should never return null

JingsongLi:
### Purpose

Scan mode should just try its best to find a snapshot, and if it cannot be found, it should return the closest snapshot

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/source/snapshot/CompactedStartingScanner.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/source/snapshot/CompactedStartingScanner.java
@@ -19,6 +19,7 @@
 package org.apache.paimon.table.source.snapshot;
 
 import org.apache.paimon.CoreOptions;
+import org.apache.paimon.Snapshot;
 import org.apache.paimon.operation.ScanKind;
 import org.apache.paimon.utils.SnapshotManager;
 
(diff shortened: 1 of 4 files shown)


paimon-core/src/main/java/org/apache/paimon/table/source/snapshot/CompactedStartingScanner.java
```diff
@@ -53,6 +54,6 @@ public Result scan(SnapshotManager snapshotManager, SnapshotReader snapshotReade
 
     @Nullable
```
yuzelin, line 55: remove annotation.

paimon-core/src/main/java/org/apache/paimon/table/source/snapshot/FullCompactedStartingScanner.java
```diff
@@ -40,7 +40,7 @@ public FullCompactedStartingScanner(int deltaCommits) {
     @Override
     @Nullable
     protected Long pick(SnapshotManager snapshotManager) {
-        return snapshotManager.pickFromLatest(this::picked);
+        return snapshotManager.pickOrLatest(this::picked);
```
yuzelin, line 43: remove annotation.

yuzelin (approved):
+1

paimon-core/src/main/java/org/apache/paimon/table/source/snapshot/CompactedStartingScanner.java
JingsongLi: Latest also can be nullable

paimon-core/src/main/java/org/apache/paimon/table/source/snapshot/FullCompactedStartingScanner.java
JingsongLi: Latest also can be nullable

PR #1521, opened 2023-07-07
[bug][flink] CDC schema evolution should pass catalog information

yuzelin:
### Purpose

For example, if user use Hive catalog in CDC, currently the schema change information is only recorded into file system, but 
cannot be seen in Hive meta store. This PR fix it.
 

### Tests

`TestAlterTableCatalogFactory`

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/cdc/UpdatedDataFieldsProcessFunction.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/cdc/UpdatedDataFieldsProcessFunction.java
@@ -19,6 +19,7 @@
 package org.apache.paimon.flink.sink.cdc;
 
 import org.apache.paimon.catalog.Catalog;
+import org.apache.paimon.catalog.Identifier;
 import org.apache.paimon.schema.SchemaChange;
 import org.apache.paimon.schema.SchemaManager;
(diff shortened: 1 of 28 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/cdc/UpdatedDataFieldsProcessFunction.java
```diff
@@ -107,17 +122,22 @@ private List<SchemaChange> extractSchemaChanges(List<DataField> updatedDataField
...
-                // handle duplicated column adds, so we just catch the exception and log it.
-                if (LOG.isDebugEnabled()) {
-                    LOG.debug(
-                            "Failed to perform SchemaChange.AddColumn {}, "
-                                    + "possibly due to duplicated column name",
-                            schemaChange,
-                            e);
+                catalog.alterTable(identifier, schemaChange, false);
+            } catch (Exception e) {
+                if (e.getCause() instanceof Catalog.ColumnAlreadyExistException) {
```
JingsongLi, line 127: Consider to modify alterTable interface:
```
void alterTable(Identifier identifier, List<SchemaChange> changes, boolean ignoreIfNotExists)
            throws TableNotExistException, ColumnAlreadyExistException, ColumnNotExistException;
```

JingsongLi (approved):
+1

PR #1522, opened 2023-07-10
[core] Do not create new Index file if there is no change

JingsongLi:
### Purpose

We should check the hash index, not new hash, do not create new index file.


yuzelin (approved):
+1

PR #1523, opened 2023-07-10
[hotfix] Clean up redundant code in Manifests Table.

zhuangchong:
### Purpose

Clean up redundant code in Manifests Table.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/system/ManifestsTable.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/system/ManifestsTable.java
@@ -142,7 +142,7 @@ public class ManifestsTable implements ReadonlyTable {
 
         @Override
         public long rowCount() {
-            return StatsManifestsGetter.manifestFileMetas(fileIO, location, dataTable).size();
+            return allManifests(fileIO, location, dataTable).size();
         }
 
         @Override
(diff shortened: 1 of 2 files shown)


paimon-core/src/main/java/org/apache/paimon/table/system/ManifestsTable.java
```diff
@@ -233,6 +232,10 @@ private StatsManifestsGetter(FileIO fileIO, Path location, Table dataTable) {
 
         private void initialize() {
```
JingsongLi, line 233: Just convert `StatsManifestsGetter` to a static method?

JingsongLi: I mean we can do a bit of refactoring, here the class doesn't seem to be of much use, just a static method will do for our needs.

zhuangchong: Sorry, I misunderstood, I have added a static method.

paimon-core/src/main/java/org/apache/paimon/table/system/ManifestsTable.java
```diff
@@ -219,34 +218,20 @@ private InternalRow toRow(ManifestFileMeta manifestFileMeta) {
     }
 
     private static class StatsManifestsGetter {
```
JingsongLi, line 220: You can just remove this class.

zhuangchong: OK, Thanks.

JingsongLi (approved):
+1

PR #1524, opened 2023-07-10
[spark] Spark Show create table should show primary key

JingsongLi:
### Purpose

Spark dose not have primary key syntax, we should add primary key option for its SHOW CREATE TABLE.


yuzelin (approved):
+1

PR #1527, opened 2023-07-10
[core] Refactor DataSplit to contain before files

JingsongLi:
### Purpose

- Refactor DataSplit to contain before files
- Overwrite use before files in DataSplit

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/source/snapshot/SnapshotReaderImpl.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/source/snapshot/SnapshotReaderImpl.java
@@ -238,19 +238,41 @@ public class SnapshotReaderImpl implements SnapshotReader {
 
         List<DataSplit> splits = new ArrayList<>();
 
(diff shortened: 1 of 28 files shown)


paimon-core/src/main/java/org/apache/paimon/table/source/snapshot/SnapshotReaderImpl.java
```diff
@@ -234,21 +238,44 @@ public Plan readOverwrittenChanges() {
...
-                        FileStoreScan.Plan.groupByPartFiles(plan.files(FileKind.DELETE))));
-
-        splits.addAll(
-                generateSplits(
-                        snapshotId,
-                        true,
-                        false,
-                        splitGenerator,
-                        FileStoreScan.Plan.groupByPartFiles(plan.files(FileKind.ADD))));
+        Map<BinaryRow, Map<Integer, List<DataFileMeta>>> beforeFiles =
```
yuzelin, line 241: Comments above should be updated.

yuzelin (approved):

PR #1531, opened 2023-07-11
[core][lineage] Introduce lineage meta interfaces

### LINKED ISSUE #1479: [Feature] Add lineage meta related interfaces
### Search before asking

### Motivation

Add lineage meta related interfaces, subtask of #684

### Solution

_No response_

### Anything else?

_No response_

FangYongs:
### Purpose

Linked issue: close #1479 

This PR aims to introduce lineage interfaces for Paimon, includes
1. Introduce lineage interfaces
2. Store source lineage when flink creates source table for job
3. Store sink lineage when flink creates sink table for job

### Tests

1. Added FlinkLineageITCase for source and sink table linage

### API and Format

no

### Documentation

no

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/catalog/CatalogContext.java
+++ b/paimon-common/src/main/java/org/apache/paimon/catalog/CatalogContext.java
@@ -19,18 +19,14 @@
 package org.apache.paimon.catalog;
 
 import org.apache.paimon.annotation.Public;
-import org.apache.paimon.factories.FactoryUtil;
 import org.apache.paimon.fs.FileIOLoader;
 import org.apache.paimon.fs.Path;
-import org.apache.paimon.lineage.LineageMeta;
-import org.apache.paimon.lineage.LineageMetaFactory;
 import org.apache.paimon.options.Options;
 
 import org.apache.hadoop.conf.Configuration;
 
 import javax.annotation.Nullable;
 
-import static org.apache.paimon.options.CatalogOptions.LINEAGE_META;
 import static org.apache.paimon.options.CatalogOptions.WAREHOUSE;
 import static org.apache.paimon.utils.HadoopUtils.getHadoopConfiguration;
 import static org.apache.paimon.utils.Preconditions.checkNotNull;
@@ -46,17 +42,14 @@ public class CatalogContext {
     private final Options options;
     private final Configuration hadoopConf;
     @Nullable private final FileIOLoader fallbackIOLoader;
-    @Nullable private final LineageMeta lineageMeta;
 
     private CatalogContext(
             Options options,
             @Nullable Configuration hadoopConf,
-            @Nullable FileIOLoader fallbackIOLoader,
-            @Nullable LineageMeta lineageMeta) {
+            @Nullable FileIOLoader fallbackIOLoader) {
         this.options = checkNotNull(options);
         this.hadoopConf = hadoopConf == null ? getHadoopConfiguration(options) : hadoopConf;
         this.fallbackIOLoader = fallbackIOLoader;
-        this.lineageMeta = lineageMeta;
     }
 
     public static CatalogContext create(Path warehouse) {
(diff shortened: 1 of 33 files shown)


JingsongLi (commented):
Can `LineageMeta` passed from catalog to table just like `CatalogLock`?

FangYongs:
@JingsongLi Nice! My previous thought was to put it in the options of the table, but I don't think it's a good way. Add it in Paimon table is much better and I like it 👍

JingsongLi (commented):
Can we just put it into `Table`?
`AbstractFileStoreTable` already has `lockFactory` and `metastoreClientFactory`, maybe we can just create `CatalogEnvirment` to contain all catalog classes.

FangYongs:
@JingsongLi That's what I want to do, but I found that if we put it into `Table`, we will updated some `Public` interfaces which are not mentioned in PIP. Do you think it's a problem? Or we can just do it directly 😄

JingsongLi:
> @JingsongLi That's what I want to do, but I found that if we put it into `Table`, we will updated some `Public` interfaces which are not mentioned in PIP. Do you think it's a problem? Or we can just do it directly 😄

You mean then we need to public some methods like getXX for Table?
Maybe it is OK to add these getXX for `FileStoreTable` or `DataTable` first.

FangYongs:
@JingsongLi I have redefined the interfaces for lineage according our discussion above, please help to review it again, thanks

paimon-common/src/main/java/org/apache/paimon/catalog/CatalogContext.java
```diff
@@ -42,14 +46,17 @@ public class CatalogContext {
...
     private final Configuration hadoopConf;
     @Nullable private final FileIOLoader fallbackIOLoader;
+    @Nullable private final LineageMeta lineageMeta;
 
     private CatalogContext(
             Options options,
             @Nullable Configuration hadoopConf,
-            @Nullable FileIOLoader fallbackIOLoader) {
+            @Nullable FileIOLoader fallbackIOLoader,
+            @Nullable LineageMeta lineageMeta) {
```
JingsongLi, line 55: It looks like `LineageMeta` should be together with `LockFactory`. They are created by catalog. Not to create catalog? (CatalogContext is for creating catalog)

FangYongs: @JingsongLi I have updated this PR to create lineage meta in AbstractCatalog

paimon-common/src/main/java/org/apache/paimon/lineage/LineageMeta.java
```diff
@@ -0,0 +1,103 @@
...
+     *
+     * @param entity the data lineage entity
+     */
+    void storeSinkDataLineage(DataLineageEntity entity);
+
+    /**
+     * Get sink data and job lineages.
+     *
+     * @param predicate the predicate for the table lineages
+     * @return the iterator for source table and job lineages
```
liming30, line 100:
```suggestion
     * @return the iterator for sink table and job lineages
```

paimon-common/src/main/java/org/apache/paimon/lineage/LineageMeta.java
```diff
@@ -0,0 +1,103 @@
...
+     *
+     * @param entity the table lineage entity
+     */
+    void storeSinkTableLineage(TableLineageEntity entity);
+
+    /**
+     * Get sink table and job lineages.
+     *
+     * @param predicate the predicate for the table lineages
+     * @return the iterator for source table and job lineages
```
liming30, line 63:
```suggestion
     * @return the iterator for sink table and job lineages
```

paimon-common/src/main/java/org/apache/paimon/lineage/LineageMeta.java
```diff
@@ -0,0 +1,103 @@
...
+    /**
+     * Get source data and job lineages.
+     *
+     * @param predicate the predicate for the table lineages
+     * @return the iterator for source table and job lineages
+     */
+    Iterator<DataLineageEntity> sourceDataLineages(@Nullable Predicate predicate);
+
+    /**
+     * Store the source table and job lineage.
```
liming30, line 90:
```suggestion
     * Store the sink table and job lineage.
```

paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
```diff
@@ -43,13 +52,17 @@ public abstract class AbstractCatalog implements Catalog {
...
+    @Nullable protected final LineageMeta lineageMeta;
+
     protected AbstractCatalog(FileIO fileIO) {
         this.fileIO = fileIO;
+        this.lineageMeta = null;
         this.tableDefaultOptions = new HashMap<>();
     }
 
-    protected AbstractCatalog(FileIO fileIO, Map<String, String> options) {
+    protected AbstractCatalog(FileIO fileIO, Map<String, String> options, ClassLoader classLoader) {
```
JingsongLi, line 63: Can we just use AbstractCatalog.class.getClassLoader? Simple, and the implementation should be togeter with paimon jar.

paimon-common/src/main/java/org/apache/paimon/lineage/LineageMeta.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.Iterator;
+
+/** Metadata store will manage table lineage and data lineage information for the catalog. */
+public interface LineageMeta extends Serializable {
+    /**
+     * Store the source table and job lineage.
+     *
+     * @param entity the table lineage entity
+     */
+    void storeSourceTableLineage(TableLineageEntity entity);
```
JingsongLi, line 35: maybe we can rename `storexxx` to `savexxx`.

I just think that the store is easy to confuse with our storage itself.  Paimon used to be called Flink Table Store.

JingsongLi (approved):
+1 after test passing

PR #1535, opened 2023-07-11
[core] Support incremental read on tags

JingsongLi:
### Purpose

Process:
- First get the differences in the files and delete the same files directly
- Then compare the before and after files to get the incremental data.

Table types:
- Primary Key table: Compare diff with two tags, perfect.
- Append table: Ignore before files, this may cause some more data to be read due to compaction


paimon-core/src/main/java/org/apache/paimon/table/source/snapshot/IncrementalTagStartingScanner.java
```diff
@@ -0,0 +1,43 @@
...
+        this.start = start;
+        this.end = end;
+    }
+
+    @Override
+    public Result scan(SnapshotManager manager, SnapshotReader reader) {
+        TagManager tagManager = new TagManager(manager.fileIO(), manager.tablePath());
+        Snapshot tag1 = tagManager.taggedSnapshot(start);
+        Snapshot tag2 = tagManager.taggedSnapshot(end);
+        return StartingScanner.fromPlan(reader.withSnapshot(tag2).readIncrementalDiff(tag1));
```
yuzelin, line 41: check tag1 is earliest than tag2

yuzelin (approved):
+1

PR #1538, opened 2023-07-11
[core] Support commit metrics

### LINKED ISSUE #1468: [Feature] Support CommitMetrics
### Search before asking

### Motivation

Sub task of #742

### Solution

_No response_

### Anything else?

_No response_

schnappi17:
fix #1468 

1. Introduce `MetricGroup` and `AbstractMetricGroup` to register metrics.
2. `CommitMetrics` is used to register the metrics for commit, using the generic metric group to register metrics untagged, and the bucketed / partitioned metric group to register metrics tagged with buckets and partitions.
3. The untagged metrics are registered at the instantiation from `AbstractFileStore`, the tagged metrics are registered through the commit operations.
4. Introduce `CommitStats` to calc the statistics of commits.


schnappi17:
@tsreaper Could you help to review this pr? Thanks a lot!

paimon-core/src/main/java/org/apache/paimon/metrics/commit/CommitStats.java
```diff
@@ -0,0 +1,376 @@
...
+
+    /** Return a map group by bucket. */
+    @VisibleForTesting
+    protected static Map<Integer, List<DataFileMeta>> groupByBucket(List<ManifestEntry> files) {
+        Map<Integer, List<DataFileMeta>> groupByBucket = new LinkedHashMap<>();
+        for (ManifestEntry entry : files) {
+            groupByBucket.computeIfAbsent(entry.bucket(), k -> new ArrayList<>()).add(entry.file());
+        }
+        return groupByBucket;
+    }
```
tsreaper, line 213: It is not very useful to only group by bucket and ignores partition. Bucket 1 in partition 1 is different from bucket 1 in partition 2, although they have the same bucket number.

paimon-core/src/main/java/org/apache/paimon/metrics/commit/CommitStats.java
```diff
@@ -0,0 +1,376 @@
...
+
+    @VisibleForTesting
+    protected static List<Integer> changedBuckets(List<ManifestEntry>... changes) {
+        return Arrays.stream(changes)
+                .flatMap(Collection::stream)
+                .map(ManifestEntry::bucket)
+                .distinct()
+                .collect(Collectors.toList());
+    }
```
tsreaper, line 251: Same as above. Even with the same bucket number, if the partitions are different, they are considered to be different buckets.

paimon-core/src/main/java/org/apache/paimon/metrics/commit/CommitMetrics.java
```diff
@@ -0,0 +1,280 @@
...
+            if (commit != null) {
+                return commit.getNumBucketsWritten();
+            } else {
+                return 0L;
+            }
+        }
+    }
+
+    public void reportCommit(CommitStats commitStats) {
+        commitMetricsReadWriteLock.lock();
```
tsreaper, line 266: `public synchronized void`. No need for a `ReentrantLock`.

Also, `reportCommit` is only called in `FileStoreCommitImpl#commit`. This method should only be called in a single thread, so there is no need to lock this method.

paimon-core/src/main/java/org/apache/paimon/metrics/Metrics.java
```diff
@@ -0,0 +1,53 @@
...
+    public static Metrics getInstance() {
+        if (instance == null) {
+            synchronized (Metrics.class) {
+                if (instance == null) {
+                    instance = new Metrics();
+                }
+            }
+        }
+        return instance;
+    }
```
tsreaper, line 39: Why lazy initialization? Creating a `Metrics` object has no cost.

paimon-core/src/main/java/org/apache/paimon/metrics/Metrics.java
```diff
@@ -0,0 +1,53 @@
...
+
+    /** Add a metric group. */
+    public void addGroup(AbstractMetricGroup group) {
+        metricGroups.add(group);
+    }
+
+    /** Get metric groups. */
+    public List<MetricGroup> getMetricGroups() {
+        return metricGroups;
+    }
```
tsreaper, line 52: You use `synchronized` to protect `getInstance` from multi-threading. What about `metricGroups`? Will it be visited by multiple threads?

paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
```diff
@@ -230,15 +239,17 @@ public void commit(ManifestCommittable committable, Map<String, String> properti
...
-            tryCommit(
-                    appendTableFiles,
-                    appendChangelog,
-                    appendIndexFiles,
-                    committable.identifier(),
-                    committable.watermark(),
-                    committable.logOffsets(),
-                    Snapshot.CommitKind.APPEND,
-                    safeLatestSnapshotId);
+            generatedSnapshot += 1;
```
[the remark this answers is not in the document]
schnappi17: Thanks, fixed it.

paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
```diff
@@ -254,17 +265,29 @@ public void commit(ManifestCommittable committable, Map<String, String> properti
...
+        CommitStats commitStats =
+                new CommitStats(
+                        appendTableFiles,
+                        appendChangelog,
+                        compactTableFiles,
+                        compactChangelog,
+                        commitDuration,
+                        generatedSnapshot,
+                        attempts);
+        commitMetrics.reportCommit(commitStats);
```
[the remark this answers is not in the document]
schnappi17: I'm confused here, if one of the snapshot failed, then it could be regarded as a complete commit? And if the `APPEND` snapshot failed due to conflicts, should the statistics been reported?

paimon-core/src/main/java/org/apache/paimon/metrics/commit/CommitStats.java
schnappi17: Yes sure, I understand the scenario. Updated the group design.

paimon-core/src/main/java/org/apache/paimon/metrics/Metrics.java
schnappi17: Thanks a lot for reminding, it's mistake here, I've fixed it.

paimon-core/src/main/java/org/apache/paimon/metrics/AbstractMetricGroup.java
```diff
@@ -0,0 +1,228 @@
...
+
+    /**
+     * Concatenates the given component names separated by the delimiter character. Additionally the
+     * character filter is applied to all component names.
+     *
+     * @param delimiter Delimiter to separate component names
+     * @param components Array of component names
+     * @return The concatenated component name
+     */
+    public String scopeConcat(Character delimiter, String... components) {
```
tsreaper, line 109: `String.join(delimiter, components)`

paimon-core/src/main/java/org/apache/paimon/metrics/MetricGroup.java
```diff
@@ -0,0 +1,141 @@
...
+ */
+@Public
+public interface MetricGroup {
+    /**
+     * Creates and registers a new {@link org.apache.paimon.metrics.Counter} with Paimon.
+     *
+     * @param name name of the counter
+     * @return the created counter
+     */
+    default Counter counter(int name) {
```
tsreaper, line 38: Why do we use an integer as the name? It seems meaningless.

paimon-core/src/main/java/org/apache/paimon/metrics/MetricGroup.java
```diff
@@ -0,0 +1,141 @@
...
+    /**
+     * Returns the fully qualified metric name, for example {@code "myTable.bucket-1.metricName"}.
+     *
+     * @param metricName metric name
+     * @return fully qualified metric name
+     */
+    String getMetricIdentifier(String metricName, char delimiter);
+
+    /** Returns a map of all variables and their associated value. */
+    Map<String, String> getAllVariables();
```
tsreaper, line 131: What the difference between `scope` and `variable` (or `tag`)? It seems that with scope, the same metric for different buckets have different names. This makes aggregating the metrics difficult.

paimon-core/src/main/java/org/apache/paimon/metrics/AbstractMetricGroup.java
```diff
@@ -0,0 +1,228 @@
...
+        if (metric == null) {
+            LOG.warn(
+                    "Ignoring attempted registration of a metric due to being null for name {}.",
+                    metricName);
+            return;
+        }
+        // add the metric only if the group is still open
+        synchronized (this) {
+            if (!closed) {
+                synchronized (this) {
```
tsreaper, line 185: Why two `synchornized`?

paimon-core/src/main/java/org/apache/paimon/metrics/AbstractMetricGroup.java
```diff
@@ -0,0 +1,228 @@
...
+                    }
+                }
+            }
+        }
+    }
+
+    @Override
+    public Map<String, Metric> getMetrics() {
+        return metrics;
+    }
```
tsreaper, line 206: No `synchronized` here?

paimon-core/src/main/java/org/apache/paimon/metrics/AbstractMetricGroup.java
```diff
@@ -0,0 +1,228 @@
...
+     */
+    public abstract String getGroupName();
+
+    public void close() {
+        synchronized (this) {
+            if (!closed) {
+                closed = true;
+                metrics.clear();
+            }
+        }
```
tsreaper, line 222: There is no need to protect `closed` and `metrics` with `synchronized`. Try `AtomicBoolean` and `ConcurrentHashMap`.

paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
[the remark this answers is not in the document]
tsreaper: If `APPEND` succeeds and `COMPACT` fails it is still a successful commit. If `APPEND` fails I think it should still count as an attempt, though not a successful attempt.

paimon-core/src/main/java/org/apache/paimon/metrics/groups/TaggedMetricGroup.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+
+package org.apache.paimon.metrics.groups;
+
+import org.apache.paimon.metrics.AbstractMetricGroup;
+
+import java.util.Map;
+
+/** Special {@link org.apache.paimon.metrics.MetricGroup} representing a bucket. */
+public class TaggedMetricGroup extends AbstractMetricGroup {
```
tsreaper, line 26: This class name is too generic. What about `BucketMetricGroup`?

PR #1541, opened 2023-07-11
[spark] Supports paimon_incremental_query Table Valued Function

YannByron:
- Builds Paimon SparkSession Extensions Arch;
- Supports paimon_incremental_query TVF to query the incr data between the two versions provided.

Linked issue: close #xxx

### DIFF
--- a/docs/content/how-to/querying-tables.md
+++ b/docs/content/how-to/querying-tables.md
@@ -138,7 +138,7 @@ To enable this needs these configs below:
 
 ```text
 --conf spark.sql.catalog.spark_catalog=org.apache.paimon.spark.SparkGenericCatalog
---conf spark.sql.extensions=org.apache.paimon.spark.PaimonSparkSessionExtensions
+--conf spark.sql.extensions=org.apache.paimon.spark.PaimonSparkSessionExtension
 ```
 
 you can use `paimon_incremental_query` in query to extract the incremental data:
--- a/docs/content/how-to/querying-tables.md
+++ b/docs/content/how-to/querying-tables.md
@@ -129,7 +129,20 @@ SELECT * FROM t /*+ OPTIONS('incremental-between' = '12,20') */;
 ```
 {{< /tab >}}
 
-{{< tab "Spark" >}}
+{{< tab "Spark3" >}}
+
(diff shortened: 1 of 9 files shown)


JingsongLi (commented):
Thanks! We can also modify documentation too.
In `/how-to/querying-tables`.

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonTableValuedFunctions.scala
```diff
@@ -0,0 +1,112 @@
...
+import org.apache.spark.sql.catalyst.plans.logical.{LeafNode, LogicalPlan}
+import org.apache.spark.sql.connector.catalog.Identifier
+import org.apache.spark.sql.execution.datasources.v2.DataSourceV2Relation
+import org.apache.spark.sql.util.CaseInsensitiveStringMap
+
+import scala.collection.JavaConverters._
+
+object PaimonTableValuedFunctions {
+
+  val INCREMENTAL_QUERY = "paimon_incremental_query"
```
JingsongLi, line 37: Dose this need to be `paimon_` prefix? Dose `incremental_query` work?
How about Delta and Hudi?

YannByron: Delta doesn't use the prefix, but Hudi does. It's better to have, to distinguish when more then one SparkSessionExtensions are used.

docs/content/how-to/querying-tables.md
```diff
@@ -129,7 +129,20 @@ SELECT * FROM t /*+ OPTIONS('incremental-between' = '12,20') */;
 ```
 {{< /tab >}}
 
-{{< tab "Spark" >}}
+{{< tab "Spark3" >}}
+
+Requires Spark 3.2+.
+
+you can use `paimon_incremental_query` in query to extract the incremental data:
```
JingsongLi, line 136: Should also add Extension and use SparkGenericCatalog?

JingsongLi (approved):
+1 Thanks @YannByron

PR #1545, opened 2023-07-12
[core][hive] Support syncing partition into Hive metastore when using Hive catalog

tsreaper:
### Purpose

Currently we cannot create partitioned Paimon table in Hive catalog. In Hive, all fields are considered as normal fields and there is no partition key. It brings us at least two downsides:

1. Users cannot list all partitions through the SHOW PARTITIONS command.
2. Hive's native partition pushdown cannot be performed. Although currently we approach partition pushdown through Hive's filter pushdown, some Hive compatible system may only support partition pushdown and not filter pushdown, making Paimon nearly unusable in those systems.

This PR introduces a feature to synchronize partitions into Hive metastore when using Hive catalog. In this PR, users can create partitioned Hive table through Flink and Spark using Hive catalog. Creating partitioned table directly in Hive is currently not supported and will be introduced in the next PR.

### Tests

* `HiveCatalogITCaseBase#testAddPartitionsToMetastore`

### API and Format

No.

### Documentation

Yes.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-core/src/main/java/org/apache/paimon/CoreOptions.java
@@ -755,14 +755,15 @@ public class CoreOptions implements Serializable {
                             "Parameter string for the constructor of class #. "
                                     + "Callback class should parse the parameter by itself.");
 
-    public static final ConfigOption<Boolean> PARTITION_ADD_TO_METASTORE =
-            key("partition.add-to-metastore")
+    public static final ConfigOption<Boolean> METASTORE_PARTITIONED_TABLE =
+            key("metastore.partitioned-table")
                     .booleanType()
                     .defaultValue(false)
                     .withDescription(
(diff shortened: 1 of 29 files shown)


paimon-core/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -755,6 +755,15 @@ public class CoreOptions implements Serializable {
                             "Parameter string for the constructor of class #. "
                                     + "Callback class should parse the parameter by itself.");
 
+    public static final ConfigOption<Boolean> PARTITION_ADD_TO_METASTORE =
+            key("partition.add-to-metastore")
```
JingsongLi, line 759: I have a feeling that what is actually being said here is that we will synchronize paimon's partition table into Hive's partition table, and after synchronization, of course, we need to addPartition.

So this key better say create hive partitioned table?

PR #1547, opened 2023-07-12
[common] Fix can not create table by java api with ROW data type

### LINKED ISSUE #1530: [Bug]  can not create table by java api with ROW data type
### Search before asking

### Paimon version

master

### Compute Engine

no

### Minimal reproduce step

create table with  java api

```
        Schema.Builder schemaBuilder = Schema.newBuilder();
        schemaBuilder.column("f0", DataTypes.INT());
        schemaBuilder.column("f1", DataTypes.STRING());
        schemaBuilder.column("f2", DataTypes.ROW(DataTypes.STRING()));
        Schema schema = schemaBuilder.build();

        Identifier identifier = Identifier.create("test", "my_table1");
        try {
            catalog.createTable(identifier, schema, false);
        } catch (Catalog.TableAlreadyExistException e) {
            // do something
        } catch (Catalog.DatabaseNotExistException e) {
            // do something
        }

```

```
java.lang.RuntimeException: java.lang.RuntimeException: Broken schema, field id 0 is duplicated.

	at org.apache.paimon.catalog.FileSystemCatalog.uncheck(FileSystemCatalog.java:207)
	at org.apache.paimon.catalog.FileSystemCatalog.createTable(FileSystemCatalog.java:164)


```

### Anything else?

_No response_

zhuangchong:
### Purpose

Linked issue: close #1530


JingsongLi:
Field Id should be unique. It is correct.

Maybe we can enhance `Schema.Builder` to nested Row field.

FangYongs:
I think the root cause is because when a nested type such as row type/map/list is added with java api, the column id starts from 0. We should update the id in `Schema.Builder.column()` for nested type instead of `RowType.collectFieldIds`

JingsongLi:
> I think the root cause is because when a nested type such as row type/map/list is added with java api, the column id starts from 0. We should update the id in `Schema.Builder.column()` for nested type instead of `RowType.collectFieldIds`

Yes, we can just update ids if there are duplicates.

zhangjun0x01:
maybe we can have users specify IDs, refer to iceberg : https://iceberg.apache.org/docs/latest/java-api-quickstart/#create-a-schema

paimon-common/src/main/java/org/apache/paimon/types/RowType.java
```diff
@@ -170,7 +170,6 @@ public void collectFieldIds(Set<Integer> fieldIds) {
                         String.format("Broken schema, field id %s is duplicated.", field.id()));
             }
             fieldIds.add(field.id());
-            field.type().collectFieldIds(fieldIds);
```
JingsongLi, line 173: Can we keep here as it is?

paimon-core/src/main/java/org/apache/paimon/schema/Schema.java
```diff
@@ -243,6 +243,10 @@ public Builder column(String columnName, DataType dataType, @Nullable String des
             Preconditions.checkNotNull(columnName, "Column name must not be null.");
             Preconditions.checkNotNull(dataType, "Data type must not be null.");
             columns.add(new DataField(++highestFieldId, columnName, dataType, description));
+
+            Set<Integer> structuredFieldIds = new HashSet<>();
+            dataType.collectFieldIds(structuredFieldIds);
+            highestFieldId += structuredFieldIds.size();
```
FangYongs, line 249: Should we need to update the id in the nested data type?

zhuangchong: I understand it is needed and I am modifying the `RowType.collectFieldIds` method.

JingsongLi (changes requested):
Perhaps we should discuss clearly why not just modify `Schema.Builder`.

paimon-core/src/main/java/org/apache/paimon/schema/Schema.java
JingsongLi: Why?

zhuangchong:
> Perhaps we should discuss clearly why not just modify `Schema.Builder`.

Because the Map type contains key type and value type, I cannot get all nested subtype fields in Schema.Builder to reset ID processing, I added the copy id method.

JingsongLi (changes requested):
You need to take a look to `ReassignFieldId`.
Actually, the final modification just be:
In `Schema.Builder.column`, just use `ReassignFieldId` to create a new DataType.

zhuangchong:
> You need to take a look to `ReassignFieldId`. Actually, the final modification just be: In `Schema.Builder.column`, just use `ReassignFieldId` to create a new DataType.

Thank you very much @JingsongLi .

JingsongLi (approved):
+1

PR #1553, opened 2023-07-12
[hive] Added  the test case for Flink Write And Hive Read scenario.

### LINKED ISSUE #1552: [Bug] paimon table hive side map type field parsing value is null
### Search before asking

### Paimon version

master branch

### Compute Engine

flink 1.16.2 + paimon0.5 + hive 3.1.0

### Minimal reproduce step

Created a paimon table with the map field type
CREATE TABLE IF NOT EXISTS paimon_heaven_eye_even_f_d (
  properties MAP<STRING,STRING>,
   login_id STRING,
  lib MAP<STRING,STRING>   dt STRING
) PARTITIONED BY (dt) WITH (
    'bucket' = '10',
    'bucket-key' = 'login_id',
    'file.format' = 'orc'
);
The flink client and hive client query the same table sql as follows:
select properties['bjx_nginx_time'],properties from paimon_heaven_eye_even_f_d where dt='20230712' limit 10;

Querying this paimon table on the flink side resolves the value of the map field properties['bjx_nginx_time'] normally, but on the hive side properties['bjx_nginx_time'] resolves to null values. For example：

14141689157731_ pic

 
14151689157731_ pic

Wish hive query paimon table map field parsed value not null

### Anything else?

_No response_

zhuangchong:
### Purpose

Linked issue: related #1552 , but not solve it.

### DIFF
--- a/paimon-hive/paimon-hive-connector-common/src/test/java/org/apache/paimon/hive/HiveCatalogITCaseBase.java
+++ b/paimon-hive/paimon-hive-connector-common/src/test/java/org/apache/paimon/hive/HiveCatalogITCaseBase.java
@@ -251,12 +251,18 @@ public abstract class HiveCatalogITCaseBase {
 
     @Test
     public void testFlinkWriteAndHiveRead() throws Exception {
-        tEnv.executeSql("CREATE TABLE t ( a INT, b STRING ) WITH ( 'file.format' = 'avro' )")
+        tEnv.executeSql(
+                        "CREATE TABLE t ( a INT, b Map<STRING, STRING> ) WITH ( 'file.format' = 'avro' )")
                 .await();
-        tEnv.executeSql("INSERT INTO t VALUES (1, 'Hi'), (2, 'Hello')").await();
+        tEnv.executeSql(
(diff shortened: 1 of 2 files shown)


paimon-hive/paimon-hive-connector-common/src/test/java/org/apache/paimon/hive/HiveCatalogITCaseBase.java
```diff
@@ -251,12 +251,18 @@ public void testCreateExternalTable() throws Exception {
 
     @Test
     public void testFlinkWriteAndHiveRead() throws Exception {
```
JingsongLi, line 253: I just run this test, it is OK without your fixing.

zhuangchong: When I verified for the first time, I could reproduce the problem submitted by the user, but now when I verify again, the query data is normal, and I have begun to doubt my life.

zhuangchong:
@JingsongLi 
I perfected the test case for Flink Write And Hive Read scenario.

JingsongLi:
Can you modify PR title too?

zhuangchong:
> Can you modify PR title too?

done.

JingsongLi (approved):
+1

PR #1557, opened 2023-07-13
[core] Add support of first_row merge engine

### LINKED ISSUE #1473: [Feature] Add support of first_row merge engine
### Search before asking

### Motivation

Introduce first_row merge engine to support duplicate the log stream. So that, user can construct a deduplicated table.

### Solution

_No response_

### Anything else?

_No response_

Aitozi:
### Purpose

Linked issue: close #1473

Add supports of first_row engine. It should work with inputs with only Insert/Update_after and with no sequence id.
By this merge engine, user can build a deduplicated table on the certain primary key and produce an insert only changelog stream, similarly to to the first_row deduplicate in flink stream. 

The point of this is to help built a reusable deduplicated paimon table for serval downstream consumer.

### Test

- FirstRowITCase
- LooupLevelsTest
- LookupChangelogMergeFunctionWrapperTest.java

### Documentation

I'd like to open another PR to doc this merge engine

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/mergetree/LookupLevels.java
+++ b/paimon-core/src/main/java/org/apache/paimon/mergetree/LookupLevels.java
@@ -34,6 +34,7 @@ import org.apache.paimon.types.RowKind;
 import org.apache.paimon.types.RowType;
 import org.apache.paimon.utils.FileIOUtils;
 import org.apache.paimon.utils.IOFunction;
+import org.apache.paimon.utils.Preconditions;
 
 import org.apache.paimon.shade.guava30.com.google.common.cache.Cache;
 import org.apache.paimon.shade.guava30.com.google.common.cache.CacheBuilder;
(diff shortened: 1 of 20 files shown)


paimon-core/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -1158,7 +1158,9 @@ public enum MergeEngine implements DescribedEnum {
 
         PARTIAL_UPDATE("partial-update", "Partial update non-null fields."),
 
-        AGGREGATE("aggregation", "Aggregate fields with same primary key.");
+        AGGREGATE("aggregation", "Aggregate fields with same primary key."),
+
+        FIRST_ROW("first-row", "De-duplicate and keep the first row.");
```
JingsongLi, line 1163: Also document this in primary key table page.

paimon-core/src/main/java/org/apache/paimon/mergetree/LookupLevels.java
```diff
@@ -102,26 +103,59 @@ public void notifyDropFile(String file) {
...
-            SortedRun level = levels.runOfLevel(i);
-            kv = lookup(key, level);
-            if (kv != null) {
-                break;
-            }
-        }
 
-        return kv;
+        switch (context.order) {
+            case High2Lower:
```
JingsongLi, line 110: Can you explain why here using `High2Lower` and `Low2Higher`? My understanding is that we can just check key contains.

paimon-core/src/main/java/org/apache/paimon/mergetree/compact/FirstRowMergeFunction.java
```diff
@@ -0,0 +1,88 @@
...
+    public void reset() {
+        this.first = null;
+    }
+
+    @Override
+    public void add(KeyValue kv) {
+        RowKind rowKind = kv.valueKind();
+        Preconditions.checkArgument(
+                rowKind.isAdd(),
+                String.format("First row merge engine don't accept %s message", rowKind));
```
JingsongLi, line 56: Do not use `String.format`, you should use:
```
checkArgument(
            boolean condition,
            @Nullable String errorMessageTemplate,
            @Nullable Object... errorMessageArgs
```
This will not invoke `String.format` in normal code path.

paimon-core/src/main/java/org/apache/paimon/mergetree/LookupLevels.java
Aitozi: I think you are right, this solution is not necessary, will refactor this part.

paimon-core/src/test/java/org/apache/paimon/mergetree/compact/SortMergeReaderTestBase.java
```diff
@@ -139,4 +144,25 @@ public void testCancelingRecords(SortEngine sortEngine) throws IOException {
             runTest(parseData("1, 2, +, 100", "1, 1, +, -100"), sortEngine);
         }
     }
+
+    /** Test for {@link SortMergeReader} with {@link FirstRowMergeFunction}. */
+    public static class WithFirstRowMergeFunctionTest extends SortMergeReaderTestBase {
```
JingsongLi, line 149: This is very good!

paimon-core/src/main/java/org/apache/paimon/mergetree/compact/LookupMergeTreeCompactRewriter.java
```diff
@@ -89,7 +89,7 @@ protected boolean upgradeChangelog(int outputLevel, DataFileMeta file) {
     protected MergeFunctionWrapper<ChangelogResult> createMergeWrapper(int outputLevel) {
         return new LookupChangelogMergeFunctionWrapper(
                 mfFactory,
-                key -> {
+                (key) -> {
```
JingsongLi, line 92: revert this?

paimon-core/src/main/java/org/apache/paimon/mergetree/compact/FullChangelogMergeFunctionWrapper.java
```diff
@@ -110,8 +112,9 @@ public ChangelogResult getResult() {
             } else {
                 if (merged == null || !isAdd(merged)) {
                     reusedResult.addChangelog(replace(reusedBefore, RowKind.DELETE, topLevelKv));
-                } else if (!changelogRowDeduplicate
-                        || !valueEqualiser.equals(topLevelKv.value(), merged.value())) {
+                } else if (!isFirstRow
```
JingsongLi, line 115: Do we have more intuitive modifications for `FullChangelogMergeFunctionWrapper` and `LookupChangelogMergeFunctionWrapper`?

If it is a first row, we should just return old value. And produce no changelog.
Can we just have another wrapper? Maybe...

paimon-core/src/main/java/org/apache/paimon/mergetree/compact/LookupMergeTreeCompactRewriter.java
Aitozi: done

paimon-core/src/main/java/org/apache/paimon/mergetree/compact/FullChangelogMergeFunctionWrapper.java
Aitozi: I have not come up with a good wrapper to handle this, but I improve the logic in `FullChangelogMergeFunctionWrapper` with a short-cut for first row

Aitozi:
please take a look again.

JingsongLi (approved):
+1

PR #1562, opened 2023-07-13
[flink] fix the issue that the historical snapshot may be lost when Enumerator performs snapshotState

### LINKED ISSUE #1561: [Bug][flink] historical snapshot may be lost when Enumerator performs snapshotState.
### Search before asking

### Paimon version

0.5-SNAPSHOT

### Compute Engine

Flink 1.17.1

### Minimal reproduce step

This is a corner case. We use an asynchronous thread (`workerExecutor`) to perform the scan operation, and add the result of the scan to the `SplitAssigner` in (`coordinatorExecutor`). There is no guarantee that it will be executed in the one by one order of `workerExecutor(scan) -> coordinatorExecutor(processDiscoveredSplits) -> workerExecutor(scan)....` 

When the discoveryInterval is small or the GC is heavy, the execution order of `scan -> scan -> processDiscoveredSplits -> processDiscoveredSplits` may appear.The `SnapshotId` of the next scan will be updated in `processDiscoveredSplits`. If it is executed in the order of `scan -> scan -> processDiscoveredSplits -> checkpoint -> failover`, the intermediate snapshot will be lost.

The Checkpoint made by Enumerator should not lose the historical snapshot.

### Anything else?

_No response_

liming30:
### Purpose

[flink] fix the issue that the historical snapshot may be lost when Enumerator performs snapshotState

Linked issue: close #1561 

### Tests

Add `org.apache.paimon.flink.source.ContinuousFileSplitEnumeratorTest#testEnumeratorWithCheckpoint` to verify the correctness of checkpoint.

### API and Format

No.

### Documentation

No.

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/ContinuousFileSplitEnumerator.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/ContinuousFileSplitEnumerator.java
@@ -106,7 +106,8 @@ public class ContinuousFileSplitEnumerator
 
     @Override
     public void start() {
-        context.callAsync(scan::plan, this::processDiscoveredSplits, 0, discoveryInterval);
+        context.callAsync(
+                this::scanNextSnapshot, this::processDiscoveredSplits, 0, discoveryInterval);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


liming30:
I'm not sure if this #1507 is trying to fix this, but it looks like `handleSplitRequest` and `processDiscoveredSplits` both run in a `coordinatorExecutor` and there shouldn't be a problem with thread conflicts.

JingsongLi:
@liming30 Really Thanks! This is a new bug!

PR #1567, opened 2023-07-14
[flink][mysql-cdc] MySQL CDC should ignore new tables that don't have primary keys or should be excluded

yuzelin:
### Purpose

Linked issue: close #xxx


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/cdc/mysql/MySqlDebeziumJsonEventParser.java
```diff
@@ -382,4 +431,28 @@ private Map<String, String> keyCaseInsensitive(Map<String, String> origin) {
...
     }
+
+    private boolean shouldExcludeCurrentTable() {
+        if (excludedTables.contains(currentTable)) {
+            return true;
+        }
+
+        boolean shouldInclude = true;
+        if (includingPattern != null) {
+            shouldInclude = includingPattern.matcher(currentTable).matches();
```
JingsongLi, line 442: NOTE this is executed per record.
We should have an `includeTables` to improve performance

JingsongLi (approved):
+1

PR #1569, opened 2023-07-15
[Core]migrate CoreOptions to common module

wg1026688210:
Currently, CoreOption is placed in the paimon-core module, which causes that the paimon-format module is unable to access it. I want to  migrate CoreOption to the paimon-common module so that format can directly read some configurations from CoreOption for reducing redundant code and improve code maintainability  , which  prepare for enhancing the format config in the next pr . 
related #1375


JingsongLi (approved):
+1

PR #1588, opened 2023-07-18
[core] Automatic Creation for tags

JingsongLi:
### Purpose

Paimon supports automatic creation of tags in writing job.

**Step 1: Choose Creation Mode**

You can set `'tag.automatic-creation'` to `process-time` or `watermark`:
- `process-time`: Create TAG based on the time of the machine.
- `watermark`: Create TAG based on the watermark of the Sink input.

If you choose Watermark, you may need to specify the time zone of watermark, if watermark is not in the
UTC time zone, please configure `'sink.watermark-time-zone'`.

**Step 2: Choose Creation Period**

What frequency is used to generate tags. You can choose `'daily'`, `'hourly'` and `'two-hours'` for `'tag.creation-period'`.

If you need to wait for late data, you can configure a delay time: `'tag.creation-delay'`.

**Step 3: Automatic deletion of tags**

You can configure `'tag.num-retained-max'` to delete tags automatically.

### DIFF
--- a/docs/content/maintenance/manage-tags.md
+++ b/docs/content/maintenance/manage-tags.md
@@ -37,7 +37,7 @@ snapshot. A typical usage is creating tags daily, then you can maintain the hist
 
 Paimon supports automatic creation of tags in writing job.
 
-**Option 1: Choose Creation Mode**
+**Step 1: Choose Creation Mode**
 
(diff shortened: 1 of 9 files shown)


yuzelin (approved):
+1

PR #1596, opened 2023-07-18
[Format] dictionary option

### LINKED ISSUE #1375: [Feature] Optimize memory for big column

wg1026688210:
close #1375 
add two  format options to enable dictionary for ORC/PARQUET

1.  format.dictionary-enable
2.  fields.{fieldName}.dictionary-enable
### test
paimon-format/src/test/java/org/apache/paimon/format/AbstractDictionaryReaderWriterTest.java
paimon-format/src/test/java/org/apache/paimon/format/orc/OrcDictionaryReaderWriterTest.java
paimon-format/src/test/java/org/apache/paimon/format/parquet/ParquetDictionaryReaderWriterTest.java

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
@@ -1239,7 +1239,7 @@ public class CoreOptions implements Serializable {
         // set whole table dictionary option
         boolean globalDictionaryEnable = options.get(CoreOptions.FORMAT_FIELDS_DICTIONARY);
 
-        Map<String, Boolean> fieldsDictionaryEnabled = new HashMap<>();
+        Map<String, Boolean> fieldsDictionaryOption = new HashMap<>();
 
         // set field's
         for (String key : options.keySet()) {
@@ -1248,11 +1248,11 @@ public class CoreOptions implements Serializable {
             if (key.startsWith(prefix) && key.endsWith(suffix)) {
                 boolean fieldDictionaryOpt = options.getBoolean(key);
                 String fieldName = key.substring(prefix.length(), key.length() - suffix.length());
-                fieldsDictionaryEnabled.put(fieldName, fieldDictionaryOpt);
+                fieldsDictionaryOption.put(fieldName, fieldDictionaryOpt);
             }
         }
 
-        return new DictionaryOptions(globalDictionaryEnable, fieldsDictionaryEnabled);
+        return new DictionaryOptions(globalDictionaryEnable, fieldsDictionaryOption);
     }
 
     public Map<String, String> commitCallbacks() {
--- a/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
@@ -58,6 +58,10 @@ public class CoreOptions implements Serializable {
 
     public static final String FIELDS_PREFIX = "fields";
 
+    public static final String FORMAT_PREFIX = "format";
+
+    public static final String DICTIONARY_PREFIX = "fields-dictionary";
+
     public static final ConfigOption<Integer> BUCKET =
             key("bucket")
                     .intType()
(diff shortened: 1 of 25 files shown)


paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -770,6 +774,14 @@ public class CoreOptions implements Serializable {
                             "Parameter string for the constructor of class #. "
                                     + "Callback class should parse the parameter by itself.");
 
+    public static final ConfigOption<Boolean> FORMAT_FIELDS_DICTIONARY =
+            key("format.fields-dictionary.enable")
+                    .booleanType()
+                    .defaultValue(true)
+                    .withDescription(
+                            "Enable or disable the dictionary for Parquet/Orc,"
+                                    + "'format.fields-dictionary.{fieldName}.enable' can be set to enable or disable the dictionary for a specific column.");
```
JingsongLi, line 783: `fields.#.dictionary-enable`?

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -770,6 +774,14 @@ public class CoreOptions implements Serializable {
                             "Parameter string for the constructor of class #. "
                                     + "Callback class should parse the parameter by itself.");
 
+    public static final ConfigOption<Boolean> FORMAT_FIELDS_DICTIONARY =
+            key("format.fields-dictionary.enable")
```
JingsongLi, line 778: format.dictionary-enable ?

paimon-common/src/main/java/org/apache/paimon/format/FileFormat.java
```diff
@@ -120,6 +120,14 @@ public static FileFormat getFileFormat(Options options, String formatIdentifier)
         int readBatchSize = options.get(CoreOptions.READ_BATCH_SIZE);
         return FileFormat.fromIdentifier(
                 formatIdentifier,
-                new FormatContext(options.removePrefix(formatIdentifier + "."), readBatchSize));
+                new FormatContext(
```
JingsongLi, line 123: Add `dictionaryDisabledFields`.

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -1194,6 +1206,43 @@ public Map<String, String> getFieldDefaultValues() {
         return defaultValues;
     }
 
+    public List<String> getDisableDictionaryFields(List<String> fieldNames) {
```
JingsongLi, line 1209: getDictionaryDisabledFields

JingsongLi:
Thanks @wg1026688210 for the contribution.

I see that PR has become increasingly complex, and I start to think about whether we really need to support this change.
We can have a unified configuration, but this should only be the cost of translation, not a significant change.

Actually, currently ORC already supports column configuration, while Parquet supports global configuration.
Do we need to support the configuration of parquet columns? I feel more like Parquet's job than ours.

So back here, can we simply document it in the orc/parquet configuration way instead of introducing thousands of lines of translation work.

wg1026688210:
> Thanks @wg1026688210 for the contribution.
> 
> I see that PR has become increasingly complex, and I start to think about whether we really need to support this change. We can have a unified configuration, but this should only be the cost of translation, not a significant change.
> 
> Actually, currently ORC already supports column configuration, while Parquet supports global configuration. Do we need to support the configuration of parquet columns? I feel more like Parquet's job than ours.
> 
> So back here, can we simply document it in the orc/parquet configuration way instead of introducing thousands of lines of translation work.

Okay, I will create another PR to update the documentation related to the format dictionary.

wg1026688210:
This PR will  be closed, and if there is a need for it, I will open it again.

PR #1599, opened 2023-07-19
[format] ParquetDataColumnReader supports reading timestamp with precision <= 6

yuzelin:
### Purpose

Currently, the default implementation of `ParquetDataColumnReader` doesn't support reading timestamp.
This PR make it possible to read timestamp if precision is less than or equal to 6.

### API and Format
Parquet format


JingsongLi:
Can you add a parquet sub class for `FormatReadWriteTest`?

JingsongLi (approved):
+1

PR #1607, opened 2023-07-20
[hive] Refactor schema check between Paimon schema and Hive DDL

tsreaper:
### Purpose

When testing Paimon Hive connector with some Hive compatible system we meet the following problems:
1. Some system does not support `partition_columns` in properties. Instead they have a `TABLE_TOTAL_PARTITIONS` which records the number of partitions in that table. We should check both properties to determine if the table is a partitioned table.
2. Hive client and Hive workers may have different network environments. It is possible that Hive client cannot access the schema files. In this case we can create Hive table based on the Hive DDL given by user.

This PR solves the two problems above.

### Tests

### API and Format

No.

### Documentation

No.


JingsongLi (approved):
+1

PR #1609, opened 2023-07-20
[Spark] supports saving a DataFrame to a paimon location

YannByron:
### Purpose

Let paimon extend `CreatableRelationProvider` that allow to saving DataFrame to a paimon location.
Additional, this is required by supporting schema merging when write.


JingsongLi (approved):
+1 Thanks @YannByron

PR #1614, opened 2023-07-20
[core] Fix missing comment after updating column type.

zhuangchong:
### Purpose

Fix missing comment after updating column type.


JingsongLi (approved):
+1

PR #1618, opened 2023-07-20
[core] Add test for histogram metric

### LINKED ISSUE #1617: [Feature] Add test for histogram metric
### Search before asking

### Motivation

Add test for paimon histogram metric

### Solution

_No response_

### Anything else?

_No response_

schnappi17:
### Purpose

Linked issue: close #1617


paimon-core/src/test/java/org/apache/paimon/metrics/AbstractHistogramTest.java
```diff
@@ -0,0 +1,60 @@
...
+package org.apache.paimon.metrics;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.offset;
+
+/**
+ * Abstract base class for testing {@link Histogram} and {@link HistogramStatistics}
+ * implementations.
+ */
+public class AbstractHistogramTest {
```
tsreaper, line 28: Why do we need an abstract test class? There is only one subclass of `Histogram`.

PR #1620, opened 2023-07-20
[core] Introduce metric group

### LINKED ISSUE #1619: [Feature] Introduce metric group
### Search before asking

### Motivation

Introduce metric group, which is used for metrics registry. 

### Solution

subtask of #742 

### Anything else?

_No response_

schnappi17:
### Purpose

Linked issue: close #1619

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/metrics/AbstractMetricGroup.java
+++ b/paimon-core/src/main/java/org/apache/paimon/metrics/AbstractMetricGroup.java
@@ -1,0 +1,187 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.paimon.metrics;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nullable;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Contains key functionality for adding metrics and carries metrics.
+ *
+ * <p>A MetricGroup can be {@link #close() closed}. Upon closing, the group de-register all metrics.
+ *
+ * <p>The {@link #close()} method and {@link #addMetric(String, Metric)} method won't be invoked by
+ * multiple threads at the same time, {@link #addMetric(String, Metric)} and {@link #getMetrics()}
+ * have multi-threads problems, so they should be synchronized by the object lock.
+ */
+public abstract class AbstractMetricGroup implements MetricGroup {
+    protected static final Logger LOG = LoggerFactory.getLogger(MetricGroup.class);
+
+    // ------------------------------------------------------------------------
+
+    /** The map containing all tags and their associated values, lazily computed. */
+    protected Map<String, String> tags;
+
+    /** Flag indicating whether this group has been closed. */
+    private boolean closed = false;
+
+    private final Map<String, Metric> metrics = new HashMap<>();
+
+    private final String table;
+
+    // ------------------------------------------------------------------------
+
+    public AbstractMetricGroup(String table, @Nullable Map<String, String> tags) {
+        this.table = table;
+        this.tags = tags;
+        Metrics.getInstance().addGroup(this);
+    }
+
+    @Override
+    public Map<String, String> getAllTags() {
+        return tags;
+    }
+
+    /**
+     * Returns the fully qualified metric name using the configured delimiter for the reporter with
+     * the given index, for example {@code "myTable.bucket-1.metricName"}.
+     *
+     * @param metricName metric name
+     * @param delimiter delimiter to use
+     * @return fully qualified metric name
+     */
+    public String getMetricIdentifier(String metricName, String delimiter) {
+        return String.join(delimiter, table, metricName);
+    }
+
+    /**
+     * Creates and registers a new {@link org.apache.paimon.metrics.Counter}.
+     *
+     * @param name name of the counter
+     * @return the created counter
+     */
+    public Counter counter(String name) {
+        return counter(name, new SimpleCounter());
+    }
+
+    /**
+     * Registers a {@link org.apache.paimon.metrics.Counter}.
+     *
+     * @param name name of the counter
+     * @param counter counter to register
+     * @param <C> counter type
+     * @return the given counter
+     */
+    public <C extends Counter> C counter(String name, C counter) {
+        addMetric(name, counter);
+        return counter;
+    }
+
+    /**
+     * Registers a new {@link org.apache.paimon.metrics.Gauge}.
+     *
+     * @param name name of the gauge
+     * @param gauge gauge to register
+     * @param <T> return type of the gauge
+     * @return the given gauge
+     */
+    public <T, G extends Gauge<T>> G gauge(String name, G gauge) {
+        addMetric(name, gauge);
+        return gauge;
+    }
+
+    /**
+     * Registers a new {@link Histogram} with Paimon.
+     *
+     * @param name name of the histogram
+     * @param histogram histogram to register
+     * @param <H> histogram type
+     * @return the registered histogram
+     */
+    public <H extends Histogram> H histogram(String name, H histogram) {
+        addMetric(name, histogram);
+        return histogram;
+    }
+
+    /**
+     * Adds the given metric to the group and registers it at the registry, if the group is not yet
+     * closed, and if no metric with the same name has been registered before.
+     *
+     * @param metricName the name to register the metric under
+     * @param metric the metric to register
+     */
+    protected void addMetric(String metricName, Metric metric) {
+        if (metric == null) {
+            LOG.warn(
+                    "Ignoring attempted registration of a metric due to being null for name {}.",
+                    metricName);
+            return;
+        }
+        // add the metric only if the group is still open
+        synchronized (this) {
+            if (!isClosed()) {
+                switch (metric.getMetricType()) {
+                    case COUNTER:
+                    case GAUGE:
+                    case HISTOGRAM:
+                        metrics.put(metricName, metric);
+                        break;
+                    default:
+                        LOG.warn(
+                                "Cannot add unknown metric type {}. This indicates that the paimon "
+                                        + "does not support this metric type.",
+                                metric.getClass().getName());
+                }
+            }
+        }
+    }
+
+    @Override
+    public Map<String, Metric> getMetrics() {
+        Map<String, Metric> metricsMap = new HashMap<>();
+        synchronized (this) {
+            metrics.forEach((k, v) -> metricsMap.put(k, v));
+        }
+        return metricsMap;
+    }
+
+    /**
+     * Returns the name for this group, meaning what kind of entity it represents, for example
+     * "bucket".
+     *
+     * @return logical name for this group
+     */
+    public abstract String getGroupName();
+
+    public void close() {
+        if (!closed) {
+            closed = true;
+            metrics.clear();
+        }
+    }
+
+    public final boolean isClosed() {
+        return closed;
+    }
+}
--- a/paimon-core/src/main/java/org/apache/paimon/metrics/MetricGroup.java
+++ b/paimon-core/src/main/java/org/apache/paimon/metrics/MetricGroup.java
@@ -1,0 +1,102 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.paimon.metrics;
+
+import org.apache.paimon.annotation.Public;
+
+import java.util.Map;
+
+/**
+ * A MetricGroup is a named container for {@link Metric Metrics} and further metric subgroups.
+ *
+ * <p>Instances of this class can be used to register new metrics with Paimon.
+ */
+@Public
+public interface MetricGroup {
+
+    /**
+     * Creates and registers a new {@link org.apache.paimon.metrics.Counter} with Paimon.
+     *
+     * @param name name of the counter
+     * @return the created counter
+     */
+    Counter counter(String name);
+
+    /**
+     * Registers a {@link org.apache.paimon.metrics.Counter} with Paimon.
+     *
+     * @param name name of the counter
+     * @param counter counter to register
+     * @param <C> counter type
+     * @return the given counter
+     */
+    <C extends Counter> C counter(String name, C counter);
+
+    /**
+     * Registers a new {@link org.apache.paimon.metrics.Gauge} with Paimon.
+     *
+     * @param name name of the gauge
+     * @param gauge gauge to register
+     * @param <T> return type of the gauge
+     * @return the given gauge
+     */
+    <T, G extends Gauge<T>> G gauge(String name, G gauge);
+
+    /**
+     * Registers a new {@link Histogram} with Paimon.
+     *
+     * @param name name of the histogram
+     * @param histogram histogram to register
+     * @param <H> histogram type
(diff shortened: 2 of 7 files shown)


paimon-core/src/main/java/org/apache/paimon/metrics/AbstractMetricGroup.java
```diff
@@ -0,0 +1,187 @@
...
+            }
+        }
+    }
+
+    @Override
+    public Map<String, Metric> getMetrics() {
+        Map<String, Metric> metricsMap = new HashMap<>();
+        synchronized (this) {
+            metrics.forEach((k, v) -> metricsMap.put(k, v));
+        }
```
tsreaper, line 165: Use `ConcurrentHashMap`. No need for `synchronized`.

paimon-core/src/main/java/org/apache/paimon/metrics/groups/GenericMetricGroup.java
```diff
@@ -0,0 +1,36 @@
...
+
+    public static final String GROUP_NAME = "table";
+
+    public GenericMetricGroup(String table) {
+        super(table, null);
+    }
+
+    @Override
+    public String getGroupName() {
+        return GROUP_NAME;
```
tsreaper, line 34: Same as above.

paimon-core/src/main/java/org/apache/paimon/metrics/AbstractMetricGroup.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    /**
+     * Creates and registers a new {@link org.apache.paimon.metrics.Counter}.
+     *
+     * @param name name of the counter
+     * @return the created counter
+     */
+    public Counter counter(String name) {
+        return counter(name, new SimpleCounter());
```
tsreaper, line 85: From a user's perspective, what I'm expecting is that, if I call `counter(name)` with the same name twice, I should get the same `Counter` object. Your implementation will throw away the counter created earlier.

paimon-core/src/test/java/org/apache/paimon/metrics/MetricGroupTest.java
```diff
@@ -0,0 +1,59 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.paimon.metrics;
+
+import org.apache.paimon.metrics.groups.GenericMetricGroup;
+
+import org.junit.Test;
```
tsreaper, line 23: Use junit5 (`org.junit.jupiter.api.Test`).

paimon-core/src/test/java/org/apache/paimon/metrics/groups/BucketMetricGroupTest.java
```diff
@@ -0,0 +1,40 @@
...
+public class BucketMetricGroupTest {
+    // ------------------------------------------------------------------------
+    //  scope name tests
+    // ------------------------------------------------------------------------
+
+    @Test
+    public void testGenerateScopeDefault() {
+        BucketMetricGroup group = BucketMetricGroup.createBucketMetricGroup("myTable", 1, "dt=1");
+
+        assertEquals(2, group.getAllTags().size());
```
tsreaper, line 35: Use `assertThat(group.getAllTags().size()).isEqualTo(2)` in junit5.

paimon-core/src/main/java/org/apache/paimon/metrics/AbstractMetricGroup.java
```diff
@@ -0,0 +1,187 @@
...
+
+    private final Map<String, Metric> metrics = new HashMap<>();
+
+    private final String table;
+
+    // ------------------------------------------------------------------------
+
+    public AbstractMetricGroup(String table, @Nullable Map<String, String> tags) {
+        this.table = table;
+        this.tags = tags;
```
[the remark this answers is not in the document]
schnappi17: In my option, `table` should exists in the metric name for users must concern about the metrics of one specific table, which means that people would rather filter a table first, and then query the metrics they need, but not query the metrics first and then specify a tag of `table`. How do you think?

paimon-core/src/main/java/org/apache/paimon/metrics/AbstractMetricGroup.java
schnappi17: I think `ConcurrentHashMap` won't guarantee the sequence of iterator and `put` operation right? Or there is no need to guarantee the sequence of iterator and the `put` operation you think?

paimon-core/src/main/java/org/apache/paimon/metrics/MetricGroup.java
```diff
@@ -0,0 +1,102 @@
...
+     * Registers a new {@link Histogram} with Paimon.
+     *
+     * @param name name of the histogram
+     * @param histogram histogram to register
+     * @param <H> histogram type
+     * @return the registered histogram
+     */
+    default <H extends Histogram> H histogram(int name, H histogram) {
+        return histogram(String.valueOf(name), histogram);
+    }
```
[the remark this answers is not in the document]
schnappi17: I missed out this, thank you for the detailed reviewing.

paimon-core/src/main/java/org/apache/paimon/metrics/Metrics.java
```diff
@@ -0,0 +1,52 @@
...
+     * instances.
+     */
+    public synchronized void addGroup(AbstractMetricGroup group) {
+        metricGroups.add(group);
+    }
+
+    /** Get metric groups. */
+    public synchronized List<MetricGroup> getMetricGroups() {
+        return metricGroups;
+    }
```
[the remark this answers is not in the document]
schnappi17: Yes, get it. Thanks!

paimon-core/src/test/java/org/apache/paimon/metrics/MetricGroupTest.java
```diff
@@ -0,0 +1,59 @@
...
+        // these will fail is the registration is propagated
+        group.counter("testcounter");
+        group.gauge(
+                "testgauge",
+                new Gauge<Object>() {
+                    @Override
+                    public Object getValue() {
+                        return null;
+                    }
+                });
```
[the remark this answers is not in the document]
schnappi17: It's test for registering metrics on a closed group, and the registering is failed. I added an assertion after the registering.

paimon-core/src/main/java/org/apache/paimon/metrics/AbstractMetricGroup.java
schnappi17: Make sense, add the handling for the metric name collision in `addMetric` method, which is a common metrics name collision handling for all the metric type. We keep the old metric and discard the new one with a warning message.

paimon-core/src/test/java/org/apache/paimon/metrics/MetricGroupTest.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    @Test
+    public void tolerateMetricNameCollisions() {
+        final String name = "abctestname";
+        GenericMetricGroup group = new GenericMetricGroup("testgroup");
+
+        assertNotNull(group.counter(name));
+        assertNotNull(group.counter(name));
+    }
```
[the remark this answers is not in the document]
schnappi17: Yes you're right, here checked the registered counters with metric name collision are the same object.

paimon-core/src/main/java/org/apache/paimon/metrics/AbstractMetricGroup.java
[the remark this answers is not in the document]
tsreaper: I disagree with this idea. If you take a look at the metric names in Flink, you will see that job id, operator id, etc., etc., are all tags. Many visualization software, for example Grafana, supports plotting the same metric with different tags at the same time. This is convenient for the users to quickly spot what table is behaving unexpectedly. If the metric names are different it would be difficult to plot all of them in the same graph.

paimon-core/src/main/java/org/apache/paimon/metrics/AbstractMetricGroup.java
tsreaper: It doesn't matter if the metrics reporter reports a metric group slightly earlier or later, so there is no need to maintain strict consistency here.

paimon-core/src/main/java/org/apache/paimon/metrics/AbstractMetricGroup.java
```diff
@@ -0,0 +1,201 @@
...
+     * Registers a {@link org.apache.paimon.metrics.Counter}.
+     *
+     * @param name name of the counter
+     * @param counter counter to register
+     * @param <C> counter type
+     * @return the given counter
+     */
+    public <C extends Counter> C counter(String name, C counter) {
+        Metric metric = addMetric(name, counter);
+        return (C) metric;
```
tsreaper, line 98:
```suggestion
        return (C) addMetric(name, counter);
```

paimon-core/src/main/java/org/apache/paimon/metrics/AbstractMetricGroup.java
```diff
@@ -0,0 +1,201 @@
...
+     * Registers a new {@link org.apache.paimon.metrics.Gauge}.
+     *
+     * @param name name of the gauge
+     * @param gauge gauge to register
+     * @param <T> return type of the gauge
+     * @return the given gauge
+     */
+    public <T, G extends Gauge<T>> G gauge(String name, G gauge) {
+        Metric metric = addMetric(name, gauge);
+        return (G) metric;
```
tsreaper, line 111:
```suggestion
        return (G) addMetric(name, gauge);
```

paimon-core/src/main/java/org/apache/paimon/metrics/AbstractMetricGroup.java
```diff
@@ -0,0 +1,201 @@
...
+     * Registers a new {@link Histogram} with Paimon.
+     *
+     * @param name name of the histogram
+     * @param histogram histogram to register
+     * @param <H> histogram type
+     * @return the registered histogram
+     */
+    public <H extends Histogram> H histogram(String name, H histogram) {
+        Metric metric = addMetric(name, histogram);
+        return (H) metric;
```
tsreaper, line 124:
```suggestion
        return (H) addMetric(name, histogram);
```

paimon-core/src/main/java/org/apache/paimon/metrics/AbstractMetricGroup.java
```diff
@@ -0,0 +1,201 @@
...
+        if (!isClosed()) {
+            switch (metric.getMetricType()) {
+                case COUNTER:
+                case GAUGE:
+                case HISTOGRAM:
+                    // immediately put without a 'contains' check to optimize the common case
+                    // (no
+                    // collision)
+                    // collisions are resolved later
+                    Metric prior = metrics.put(metricName, metric);
```
tsreaper, line 151:
```suggestion
                    // immediately put without a 'contains' check to optimize the common case
                    // (no collision), collisions are resolved later
                    Metric prior = metrics.put(metricName, metric);
```

paimon-core/src/main/java/org/apache/paimon/metrics/AbstractMetricGroup.java
```diff
@@ -0,0 +1,201 @@
...
+
+                    // check for collisions with other metric names
+                    if (prior != null) {
+                        // we had a collision. put back the original value
+                        metrics.put(metricName, prior);
+
+                        // we warn here, rather than failing, because metrics are tools that
+                        // should not
+                        // fail the
+                        // program when used incorrectly
```
tsreaper, line 161:
```suggestion
                        // we warn here, rather than failing, because metrics are tools that
                        // should not fail the program when used incorrectly
```

paimon-core/src/main/java/org/apache/paimon/metrics/Metrics.java
```diff
@@ -0,0 +1,51 @@
...
+     * instances.
+     */
+    public synchronized void addGroup(AbstractMetricGroup group) {
+        metricGroups.add(group);
+    }
+
+    /** Get metric groups. */
+    public synchronized ConcurrentLinkedQueue<MetricGroup> getMetricGroups() {
+        return metricGroups;
+    }
```
tsreaper, line 50: Why `synchronized`? You already used `ConcurrentLinkedQueue`.

paimon-core/src/main/java/org/apache/paimon/metrics/AbstractMetricGroup.java
```diff
@@ -0,0 +1,201 @@
...
+        return Collections.unmodifiableMap(metrics);
+    }
+
+    /**
+     * Returns the name for this group, meaning what kind of entity it represents, for example
+     * "bucket".
+     *
+     * @return logical name for this group
+     */
+    public abstract String getGroupName();
```
tsreaper, line 189: For example "bucket"? Bucket is just a tag for the metrics (at least for now). Shouldn't it be "compact" or "commit"?

paimon-core/src/main/java/org/apache/paimon/metrics/MetricGroup.java
```diff
@@ -0,0 +1,90 @@
...
+    String getMetricIdentifier(String metricName, String delimiter);
+
+    /** Returns a map of all variables and their associated value. */
+    Map<String, String> getAllTags();
+
+    /**
+     * Returns the name for this group, meaning what kind of entity it represents, for example
+     * "bucket".
+     */
+    String getGroupName();
```
tsreaper, line 86: Same as above.

paimon-core/src/test/java/org/apache/paimon/metrics/MetricGroupTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Test
+    public void tolerateMetricNameCollisions() {
+        final String name = "abctestname";
+        GenericMetricGroup group = GenericMetricGroup.createGenericMetricGroup("myTable", "commit");
+
+        Counter counter1 = group.counter(name);
+
+        // return the old one with the metric name collision
+        assertThat(group.counter(name)).isEqualTo(counter1);
```
tsreaper, line 60: Use `isSameAs` to check for the exact same object.

```suggestion
        // return the old one with the metric name collision
        assertThat(group.counter(name)).isSameAs(counter1);
```

paimon-core/src/test/java/org/apache/paimon/metrics/MetricGroupTest.java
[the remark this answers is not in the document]
tsreaper: I'm wondering why we need this test. Under what circumstances will we add a metric after the metric group is closed?

paimon-core/src/main/java/org/apache/paimon/metrics/AbstractMetricGroup.java
schnappi17: Thanks for your detailed reviewing, I missed this desc after re-designed the sense of group name.

paimon-core/src/test/java/org/apache/paimon/metrics/MetricGroupTest.java
schnappi17: Thanks for reminding, learn much!

paimon-core/src/main/java/org/apache/paimon/metrics/Metrics.java
schnappi17: Yes, the `synchronized` here could be removed by using `ConcurrentLinkedQueue`, sorry for missing this after updating

paimon-core/src/main/java/org/apache/paimon/metrics/AbstractMetricGroup.java
schnappi17: Thanks for helping correct this.

paimon-core/src/main/java/org/apache/paimon/metrics/AbstractMetricGroup.java
schnappi17: Yes, this is a redundant coding, updated it.

paimon-core/src/main/java/org/apache/paimon/metrics/AbstractMetricGroup.java
```diff
@@ -0,0 +1,194 @@
...
+    protected Map<String, String> tags;
+
+    /** Flag indicating whether this group has been closed. */
+    private boolean closed = false;
+
+    private final ConcurrentHashMap<String, Metric> metrics = new ConcurrentHashMap<>();
+
+    // ------------------------------------------------------------------------
+
+    public AbstractMetricGroup(@Nullable Map<String, String> tags) {
```
tsreaper, line 55: Why is it `@Nullable`? What's the difference between a `null` tag map and an empty tag map?

paimon-core/src/test/java/org/apache/paimon/metrics/MetricGroupTest.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.HashMap;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/** Tests for the {@link MetricGroup}. */
+public class MetricGroupTest {
+    @Test
+    public void groupRegisterMetrics() {
```
tsreaper, line 34: Test method names should begin with `test`. For example `testRegisterMetricsToGroup`.

paimon-core/src/test/java/org/apache/paimon/metrics/MetricGroupTest.java
```diff
@@ -0,0 +1,71 @@
...
+                                put("table", "myTable");
+                            }
+                        });
+        assertThat(group.getMetrics().size()).isEqualTo(2);
+        group.close();
+        assertTrue(group.isClosed());
+    }
+
+    @Test
+    public void tolerateMetricNameCollisions() {
```
tsreaper, line 62: Same as above.

PR #1622, opened 2023-07-21
[flink][cdc] Setting memory pool for multiple tables sink

yuzelin:
### Purpose

Limit memory pools in `CdcRecordStoreMultiWriteOperator` to avoid possible OOM.


JingsongLi (approved):
+1

PR #1623, opened 2023-07-21
[flink] flink web ui add write-only tag to writer(which is write only)

leaves12138:
Writer       ---->>>          Writer(write-only)    
(if the writer is write-only, table config with write-only=true)


JingsongLi (approved):
+1

PR #1624, opened 2023-07-21
[improve] support zorder rewrite the unaware-bucket table

leaves12138:
This class are copied from iceberg:
paimon-core/src/main/java/org/apache/paimon/utils/ByteBuffers.java
paimon-core/src/main/java/org/apache/paimon/utils/ZOrderByteUtils.java
paimon-core/src/test/java/org/apache/paimon/utils/TestZOrderByteUtil.java

This bytes(every column devoted to z-index) is fixed to 8 bytes, so if the column is short or byte etc, it will fill the upper byte with zero. If the column is string or something like byte[] more then 8, it will truncate it to 8 byte.

The action will invoke like this:

 Table(sql select) ----------------> DataStream[Row] ----------[with zorder]---------> DataStream[Row] -----------------> Table(with z-order) -----------------> Table(sorted) -----[overwrite target]-----> Table(overwrited)

usage example:
<flink run paimon jar>
zorder-rewrite
--warehouse
/Users/yejunhao/paimontest/GlobalBatchReadWrite
--database
my_db
--table
Orders1
--sql-select
"SELECT * FROM my_db.Orders1 WHERE f0 = 0"
--zorder-by
f0,f1,f2,f3,f4,f7,f8,f9,f10,f11,f12,f13,f14,f15


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/ZorderRewriteAction.java
```diff
@@ -0,0 +1,62 @@
...
+ * Rewrite the target table by z-order with specified columns.
+ *
+ * <p>The effect is similar to sql: "INSERT OVERWRITE target_table SELECT * FROM SOURCE ZORDER BY
+ * col1,col2,... "
+ *
+ * <p>Example usage: zorder-rewrite --warehouse /tmp/paimon/warehouse --database my_db --table
+ * Orders1 --sql-select "SELECT * FROM my_db.Orders1 WHERE f0 < 10" --zorder-by
+ * f0,f1,f2,f3,f4,f7,f8,f9,f10,f11,f12,f13,f14,f15
+ */
+public class ZorderRewriteAction extends FlinkActionEnvironmentBase {
```
JingsongLi, line 38: See https://iceberg.apache.org/docs/latest/spark-procedures/#rewrite_data_files
A `sort_order` to declare z-order.

Maybe we can have a sort action. and with a sort-order to declare z-order too.

JingsongLi (commented):
Copy codes from other project, you can refer to https://github.com/apache/incubator-paimon/pull/1403

1. LICENSE
2. File header

JingsongLi (commented):
Please document this feature.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/ActionFactory.java
```diff
@@ -160,4 +160,20 @@ static void parseKeyValueString(Map<String, String> map, String kvString) {
         }
         map.put(kv[0].trim(), kv[1].trim());
     }
+
+    default String getSqlOrderBy(MultipleParameterTool params) {
```
JingsongLi, line 164: Why in base class?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/ActionFactory.java
```diff
@@ -160,4 +160,20 @@ static void parseKeyValueString(Map<String, String> map, String kvString) {
...
+        String sqlOrderBy = params.get("sql-order-by");
+        if (sqlOrderBy == null) {
+            throw new IllegalArgumentException("Please specify \"sql-order-by\".");
+        }
+        return sqlOrderBy;
+    }
+
+    default String getSqlSelect(MultipleParameterTool params) {
```
JingsongLi, line 172: Why in base class?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/FlinkActionEnvironmentBase.java
```diff
@@ -0,0 +1,86 @@
...
+import org.apache.flink.table.api.EnvironmentSettings;
+import org.apache.flink.table.api.bridge.java.StreamTableEnvironment;
+import org.apache.flink.table.types.logical.LogicalType;
+
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+/** Flink Action Base with batch environment and stream environment. */
+public abstract class FlinkActionEnvironmentBase extends ActionBase {
```
JingsongLi, line 36: You can just put this into `ActionBase`

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/OrderRewriteActionFactory.java
```diff
@@ -0,0 +1,86 @@
...
+                "Action \"order-rewrite\" is similar to sql \"INSERT OVERWRITE target_table SELECT * FROM SOURCE ZORDER BY col1,col2,... \".");
+        System.out.println();
+
+        System.out.println("Syntax:");
+        System.out.println(
+                "  order-rewrite --warehouse <warehouse-path>\n"
+                        + "             --database <database-name>\n"
+                        + "             --warehouse <warehouse-path>\n"
+                        + "             --table <target-table-name>\n"
+                        + "             --sql-select \"sql\"\n"
```
JingsongLi, line 74: Should we have sql-select and sql-order-by? Can you list some reference system?

leaves12138: Learned from iceberg: https://iceberg.apache.org/docs/latest/spark-procedures/#rewrite_data_files

Iceberg rewrite action has 
sort_order: For Zorder use a comma separated list of columns within zorder(). (Supported in Spark 3.2 and Above) Example: zorder(c1,c2,c3). Else, Comma separated sort orders in the format (ColumnName SortDirection NullOrder). Where SortDirection can be ASC or DESC. NullOrder can be NULLS FIRST or NULLS LAST. Defaults to the table’s sort order
where:predicate as a string used for filtering the files. Note that all files that may contain data matching the filter will be selected for rewriting

eg:
CALL catalog_name.system.rewrite_data_files(table => 'db.sample', where => 'id = 3 and name = "foo", strategy => 'sort', sort_order => 'zorder(c1,c2)')

But I thing use write sql-select is more flexible, which mean they could rewrite source table to target table by zorder without changing origin table. So i change "where" to "sql-select"

YannByron:
I notice that most of codes is in Flink. Can extract more codes to the common/core module so that other engines (like spark) can use them?

leaves12138:
This pull request re-open as https://github.com/apache/incubator-paimon/pull/1846

PR #1625, opened 2023-07-22
[flink] upgrade to flink cdc 2.4

zhangjun0x01:
### Purpose

flink cdc 2.4 has a bug , for new added table in the runtime ,we can get the table ,but we can not get the data of the new table, i check it , it is caused by `scanNewlyAddedTables ` param, so I remove it , then all test case passed, I think it meets all our current scenario

### DIFF
--- a/paimon-flink/paimon-flink-common/pom.xml
+++ b/paimon-flink/paimon-flink-common/pom.xml
@@ -35,7 +35,7 @@ under the License.
 
     <properties>
         <flink.version>1.17.1</flink.version>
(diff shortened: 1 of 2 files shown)


zhangjun0x01:
image

yuzelin:
I think it's better to keep our flink-cdc version at 2.3.0. I have reported this bug to flink-cdc repo.

zhuangchong:
I will close this PR, if you have any questions you can open this RP again.

PR #1626, opened 2023-07-22
[flink] unify get catalog method for flink action

zhangjun0x01:
### Purpose

 unify get catalog method for all test case  in flink action .


JingsongLi (approved):
+1

PR #1627, opened 2023-07-23
[filnk] make kafka sync action to public

zhangjun0x01:
### Purpose

when use flink api to start a flink kafka sync action job  ,  it will throw an `Cannot be accessed from outside package` exception , because it is not public.


JingsongLi (commented):
Also modify DropPartitionAction and MergeIntoAction.

zhangjun0x01:
> Also modify DropPartitionAction and MergeIntoAction.

done

JingsongLi (approved):
+1

PR #1629, opened 2023-07-24
[checkstyle] Forbid directly using guava in checkstyle

tsreaper:
### Purpose

Paimon provides paimon-shaded-guava to avoid class conflicts. We should enforce the developers to use this shaded guava by adding a rule in checkstyle.

### Tests

No tests for this change.

### API and Format

No.

### Documentation

No.


JingsongLi (changes requested):
Already have this in checkstyle file.

PR #1630, opened 2023-07-24
[flink][cdc] Remove Druid dependency

yuzelin:
### Purpose

Using ddl to parse new table cannot always work (for example, `create table like`). So stop using Druid to parse sql.


JingsongLi (approved):
+1

PR #1636, opened 2023-07-24
[flink] FileStoreLookupFunction should not throw OutOfRangeException

JingsongLi:
### Purpose

When the main table occasionally has no data, while the dimension table always has data, there is no data to drive the loading of the dimension table, which may result in the snapshot of the dimension table being expired when waiting for a long time to load.

We should avoid making mistakes in this situation.


yuzelin (approved):
+1

PR #1637, opened 2023-07-24
[hive] PaimonInputFormat#getSplits should deal with multiple partitions in INPUT_DIR

tsreaper:
### Purpose

For a partitioned table, the location of its input split must point to a partition, not the root directory of that table. This PR fixes the issue.

### Tests

* `HiveE2eTest#testMetastorePartitionedTable`.

### API and Format

No.

### Documentation

No.


JingsongLi (approved):
+1

PR #1644, opened 2023-07-25
[core] Optimize First row merge engine with lookup

JingsongLi:
### Purpose

Introduce `ContainsLevels` to optimize first row.

And first row only works with lookup changelog-producer.

Benifit:

1. One is to generate a changelog through contains, which should be highly efficient. It only reads keys and does not read the entire row of data.
2. The second one is that there is data at the upper level, and the output of data at the lower level is directly null. After merging, it is directly deleted


Aitozi (commented):
Very nice improvement, Only left some minor comments.

PR #1647, opened 2023-07-25
[improve] add system table to show all table configs in `sys.all_table_options`

leaves12138:
add system table system$all_table_options to enable config search

sql:
```
SELECT * FROM sys.all_table_options
```

add built-in system "sys"
sql:
```
USE sys;
SHOW TABLES;
```
will result in:
all_table_options

We will add more system table in the future

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
@@ -40,7 +40,7 @@ public abstract class AbstractCatalog implements Catalog {
 
     public static final String DB_SUFFIX = ".db";
     protected static final String TABLE_DEFAULT_OPTION_PREFIX = "table-default.";
-    protected static final List<String> globalTables =
+    protected static final List<String> GLOBAL_TABLES =
             Arrays.asList(AllTableOptionsTable.ALL_TABLE_OPTIONS);
 
     protected final FileIO fileIO;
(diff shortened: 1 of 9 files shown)


JingsongLi (commented):
Do we need a system database?

JingsongLi:
Do we have a reference system? For example, other databases

leaves12138:
Add built-in database: sys

We can search all system table by sql:
use sys;
show tables;

But don't show sys in `list databases` to protect this database;

We can search all table_options by :
```
SELECT * FROM sys.all_table_options
```

JingsongLi:
Why naming to sys? Do you have some reference system?

leaves12138:
> Why naming to sys? Do you have some reference system?

Yes, I have referred some database such as mysql, oracle, the result is :

mysql:    have a system database named: sys  https://dev.mysql.com/doc/refman/8.0/en/sys-schema-usage.html
oracle:   have a system database named: sys   https://docs.oracle.com/database/timesten-18.1/TTSYS/systemtables.htm#TTSYS346
sql server: https://learn.microsoft.com/en-us/sql/relational-databases/system-catalog-views/sys-objects-transact-sql?view=sql-server-ver16
postgresql: no system database, system table are head of "pg_": http://www.postgres.cn/docs/9.4/catalogs.html
doris: no system database, one system table: rowsets https://doris.apache.org/docs/dev/admin-manual/system-table/rowsets/
hologres：system database just called hologres: https://www.alibabacloud.com/help/zh/hologres/user-guide/system-tables

hudi pandas iceberg: no system database

paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
```diff
@@ -61,9 +66,128 @@ protected AbstractCatalog(FileIO fileIO, Map<String, String> options) {
                                         options.get(key)));
     }
 
+    @Override
+    public List<String> listDatabases() {
+        // we hide system database
+        return listDatabasesImpl();
```
JingsongLi, line 72: We don't need to introduce so many impl methods...

JingsongLi (approved):
+1

PR #1649, opened 2023-07-25
[flink][cdc] Multiple tale write should share one write buffer

yuzelin:
### Purpose

When synchronizing multiple tables, all table writes should share one write buffer so that writers can preempt memory from those of other tables.

### Tests

`MySqlSyncDatabaseActionITCase#testSyncManyTableWithLimitedMemory`

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/memory/MemoryPoolFactory.java
+++ b/paimon-core/src/main/java/org/apache/paimon/memory/MemoryPoolFactory.java
@@ -18,6 +18,8 @@
 
 package org.apache.paimon.memory;
 
+import org.apache.paimon.annotation.VisibleForTesting;
+
 import org.apache.paimon.shade.guava30.com.google.common.collect.Iterators;
 
 import java.util.List;
@@ -53,6 +55,11 @@ public class MemoryPoolFactory {
     public void notifyNewOwner(MemoryOwner owner) {
         checkNotNull(owners);
         owner.setMemoryPool(createSubPool(owner));
+    }
+
+    @VisibleForTesting
+    public Iterable<MemoryOwner> memoryOwners() {
+        return owners;
     }
 
     MemorySegmentPool createSubPool(MemoryOwner owner) {
--- a/paimon-core/src/main/java/org/apache/paimon/memory/MemoryPoolFactory.java
+++ b/paimon-core/src/main/java/org/apache/paimon/memory/MemoryPoolFactory.java
@@ -18,7 +18,11 @@
 
 package org.apache.paimon.memory;
 
+import org.apache.paimon.shade.guava30.com.google.common.collect.Iterators;
+
 import java.util.List;
+
+import static org.apache.paimon.utils.Preconditions.checkNotNull;
 
 /**
  * A factory which creates {@link MemorySegmentPool} from {@link MemoryOwner}. The returned memory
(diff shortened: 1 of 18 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/TableWriteOperator.java
```diff
@@ -91,7 +93,18 @@ void initStateAndWriter(
...
+        write =
+                storeSinkWriteProvider.provide(
+                        table,
+                        commitUser,
+                        state,
+                        ioManager,
+                        new MemoryPoolFactory(
+                                memoryPool != null
+                                        ? memoryPool
+                                        : new HeapMemorySegmentPool(
```
JingsongLi, line 105: default null？

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/StoreSinkWriteImpl.java
```diff
@@ -80,12 +80,13 @@ private TableWriteImpl<?> newTableWrite(FileStoreTable table) {
...
-                .withMemoryPool(
-                        memoryPool != null
-                                ? memoryPool
-                                : new HeapMemorySegmentPool(
-                                        table.coreOptions().writeBufferSize(),
-                                        table.coreOptions().pageSize()))
+                .withMemoryPoolFactory(
+                        memoryPoolFactory != null
+                                ? memoryPoolFactory
+                                : new MemoryPoolFactory(
```
JingsongLi, line 86: maybe here default null?

paimon-core/src/main/java/org/apache/paimon/memory/MemoryPoolFactory.java
```diff
@@ -28,15 +32,26 @@ public class MemoryPoolFactory {
...
+
+    public MemoryPoolFactory(MemorySegmentPool innerPool) {
         this.innerPool = innerPool;
         this.totalPages = innerPool.freePages();
-        this.owners = owners;
+    }
+
+    public MemoryPoolFactory addOwners(Iterable<MemoryOwner> newOwners) {
+        if (this.owners == null) {
+            this.owners = newOwners;
```
JingsongLi, line 45: add test here.

here should store `List<Iterable<MemoryOwner>>`.
If there is a new write generated later, it will not enter this factory

paimon-core/src/main/java/org/apache/paimon/table/sink/TableWrite.java
```diff
@@ -36,8 +36,8 @@ public interface TableWrite extends AutoCloseable {
     /** With {@link IOManager}, this is needed if 'write-buffer-spillable' is set to true. */
     TableWrite withIOManager(IOManager ioManager);
 
-    /** With {@link MemorySegmentPool} for the current table write. */
-    TableWrite withMemoryPool(MemorySegmentPool memoryPool);
+    /** With {@link MemoryPoolFactory} for the current table write. */
+    TableWrite withMemoryPoolFactory(MemoryPoolFactory memoryPoolFactory);
```
JingsongLi, line 40: Can we not modify this public method?

paimon-core/src/main/java/org/apache/paimon/table/sink/TableWrite.java
```diff
@@ -36,8 +37,19 @@ public interface TableWrite extends AutoCloseable {
...
+     */
+    default TableWrite withMemoryPool(MemorySegmentPool memoryPool) {
+        return withMemoryPoolFactory(new MemoryPoolFactory(memoryPool));
+    }
+
+    /**
+     * With {@link MemoryPoolFactory} for the current table write. Cannot be called with {@link
+     * #withMemoryPool} together.
+     */
+    TableWrite withMemoryPoolFactory(MemoryPoolFactory memoryPoolFactory);
```
JingsongLi, line 52: Add this in `TableWriteImpl`. We can avoid making changes to this public classe

JingsongLi (approved):
+1

PR #1651, opened 2023-07-25
[test] Fix unstable test: UnawareBucketAppendOnlyTableITCase.testCompactionInStreamingMode

leaves12138:
This test is unstable, guess the rows generate too fast, and job start too slow to generate 4 file of 4 rows. So make more rows by datagen.


JingsongLi (approved):
+1

PR #1657, opened 2023-07-25
[bug] [bugfix] fix spark generic catalog could not load existing paimon tables

leaves12138:
fix spark generic catalog could not load existing paimon tables

spark_catalog will throw NoSuchDatabaseException in loadTable(ident)


JingsongLi (approved):
+1

PR #1660, opened 2023-07-25
[flink] fix rename table error in sync action

zhangjun0x01:
### Purpose

Linked issue: close https://github.com/apache/incubator-paimon/issues/1659

I fixed it ,the result is : 
During the data synchronization,  we rename a table,  the old table will stop data synchronization , the new table will not be synchronized either.


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/cdc/CdcMultiplexRecordChannelComputer.java
```diff
@@ -51,9 +57,13 @@ public void setup(int numChannels) {
 
     @Override
     public int channel(CdcMultiplexRecord multiplexRecord) {
+        int recordChannel =
+                computeChannelComputer(multiplexRecord) != null
```
JingsongLi, line 61: reuse `computeChannelComputer(multiplexRecord)` to a local field

zhangjun0x01: done

JingsongLi (approved):
+1

PR #1663, opened 2023-07-26
[flink] parse mysql-conf error when it contain `=`

zhangjun0x01:
### Purpose

when the mysql conf contain `=` (for example `password=VdwrHwd3=9bxN@#fySW`) , it can not be parsed correctly.

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/ActionFactory.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/ActionFactory.java
@@ -151,7 +151,7 @@ public interface ActionFactory extends Factory {
     }
 
     static void parseKeyValueString(Map<String, String> map, String kvString) {
-        String[] kv = kvString.split("=");
+        String[] kv = kvString.split("=", 2);
         if (kv.length != 2) {
             throw new IllegalArgumentException(
                     String.format(


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/ActionFactory.java
```diff
@@ -151,8 +151,8 @@ static Map<String, String> parseCommaSeparatedKeyValues(String keyValues) {
     }
 
     static void parseKeyValueString(Map<String, String> map, String kvString) {
-        String[] kv = kvString.split("=");
-        if (kv.length != 2) {
+        String[] kv = kvString.split("=", 2);
+        if (kv.length < 2) {
```
JingsongLi, line 155: Why != to < ?

zhangjun0x01: because we add the `limit` for `split` method, the max length of `kv` array is 2,  the exception is that the expression does not have `=`,  which means that `kv.length < 2`

JingsongLi: `kv.length != 2` Where is the difference?

zhangjun0x01: `!= 2` is also correct. What I started thinking was that after adding a limit 2 , there would be no situation `>2`, so I changed it to `<2`

I rollback the code.

JingsongLi (approved):
+1

PR #1669, opened 2023-07-27
[hotfix] Clean up the redundant code of the method of obtaining timestamp type precision.

zhuangchong:
### Purpose

Clean up the redundant code of the method of obtaining timestamp type precision.


JingsongLi (approved):
+1

PR #1674, opened 2023-07-27
[flink][mysql-cdc] Support Synchronizing multiple shards into one database

### LINKED ISSUE #1161: [Feature] Support Synchronizing from multiple database in CDC sync-database action
### Search before asking

### Motivation

Currently, `mysql-sync-database` only support sync single database. But a common scenario is a table has been split into multiple database, for example, the `db1` and `db2` both have a table called `tbl`.

We want support that the two tables can be synchronized into one table, that means we should support synchronizing from multiple database.

### Solution

The `mysql-sync-table` action can set database name with a regular expression to support synchronizing from multiple database. The solution can be applied to sync database action.

### Anything else?

_No response_

yuzelin:
### Purpose

Linked issue: close #1161

### DIFF
--- a/docs/content/how-to/cdc-ingestion.md
+++ b/docs/content/how-to/cdc-ingestion.md
@@ -196,6 +196,37 @@ The command to recover from previous snapshot and add new tables to synchronize 
     --including-tables 'product|user|address|order|custom'
 ```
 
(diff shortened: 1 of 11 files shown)


JingsongLi (approved):
+1

PR #1677, opened 2023-07-27
[core] fix typo in SchemaManager.java

eltociear:
### Purpose
loosing -> losing


JingsongLi (changes requested):
Can you fix related tests?

JingsongLi:
It looks like `loosing` is OK sometimes..

PR #1678, opened 2023-07-27
[spark] supports schema merging when write

YannByron:
### Purpose

Linked issue: close https://github.com/apache/incubator-paimon/issues/1566


paimon-common/src/main/java/org/apache/paimon/types/ILength.java
```diff
@@ -0,0 +1,25 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.paimon.types;
+
+/** An interface with a getLength method. */
+public interface ILength {
```
JingsongLi, line 22: Do you think https://github.com/apache/incubator-paimon/pull/1685 works for you?

paimon-core/src/main/java/org/apache/paimon/schema/SchemaMergingUtils.java
```diff
@@ -0,0 +1,196 @@
...
+import org.apache.paimon.types.RowType;
+
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+/** The util class for merging the schemas. */
+public class SchemaMergingUtils {
```
JingsongLi, line 40: Maybe we can merge this class to `SchemaManager`?
And introduce a method to `mergeSchema(RowType rowType)` to `FileStore`?

paimon-core/src/main/java/org/apache/paimon/schema/SchemaMergingUtils.java
```diff
@@ -0,0 +1,196 @@
...
+     */
+    static DataType merge(DataType base0, DataType update0, AtomicInteger highestFieldId) {
+        // Here we try t0 merge the base0 and update0 without regard to the nullability,
+        // and set the base0's nullability to the return's.
+        DataType base = base0.copy(true);
+        DataType update = update0.copy(true);
+
+        if (base.equals(update)) {
+            return base0;
+        } else if (base instanceof RowType && update instanceof RowType) {
```
JingsongLi, line 85: Maybe here can use `DataTypeDefaultVisitor`?

paimon-core/src/main/java/org/apache/paimon/schema/SchemaMergingUtils.java
```diff
@@ -0,0 +1,196 @@
...
+        } else if (base instanceof DecimalType && update instanceof DecimalType) {
+            if (base.equals(update)) {
+                return base0;
+            } else {
+                throw new UnsupportedOperationException(
+                        String.format(
+                                "Failed to merge decimal types with different precision or scale: %s and %s",
+                                base, update));
+            }
+        } else if (DataTypeCasts.supportsImplicitCast(base, update)) {
```
JingsongLi, line 159: Currently, we have using `supportsExplicitCast` to manually casting (From DDL alter table).
Do you think we should use `supportsExplicitCast`? Or provide an option to let user to choose?

JingsongLi (commented):
Thanks @YannByron , Nice working!

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -818,6 +818,13 @@ public class CoreOptions implements Serializable {
                                     + " the value should be the user configured local time zone. The option value is either a full name"
                                     + " such as 'America/Los_Angeles', or a custom timezone id such as 'GMT-08:00'.");
 
+    public static final ConfigOption<Boolean> MERGE_SCHEMA =
```
JingsongLi, line 821: Maybe we can introduce a `SparkConnectorOptions`? This is only for spark.

YannByron: I think so.

paimon-common/src/main/java/org/apache/paimon/types/ILength.java
YannByron: sure. very nice. I can't find it before.

paimon-core/src/main/java/org/apache/paimon/schema/SchemaMergingUtils.java
YannByron: >> Maybe we can merge this class to SchemaManager?
Even though merge this to `SchemaManager`, all the methods will be `static`. i think there  is no different.

>> And introduce a method to mergeSchema(RowType rowType) to FileStore?
yes, better than the current `commitSchema`.

paimon-core/src/main/java/org/apache/paimon/schema/SchemaMergingUtils.java
YannByron: that needs an option if we want to support Explicit-Casting.

paimon-core/src/main/java/org/apache/paimon/schema/SchemaMergingUtils.java
YannByron: I think `DataTypeVisitor` is more used to extract some attributes, here I wanna know if a datatype is exactly what i want, and do some process if true. So maybe `instanceof` is clear enough.

JingsongLi: Sounds reasonable to me.

paimon-common/src/main/java/org/apache/paimon/types/DataField.java
```diff
@@ -76,6 +77,11 @@ public DataType type() {
         return type;
     }
 
+    @VisibleForTesting
```
JingsongLi, line 80: A newId method looks good to me.
We can just remove `VisibleForTesting`. (Public class should not have public VisibleForTesting method)

paimon-core/src/main/java/org/apache/paimon/AbstractFileStore.java
```diff
@@ -142,6 +142,11 @@ public CoreOptions options() {
         return options;
     }
 
+    @Override
+    public boolean mergeSchema(RowType rowType, Boolean allowExplicitCast) {
```
JingsongLi, line 146: boolean instead of Boolean?

JingsongLi (commented):
Thanks for the update, just left some minor comments.

paimon-core/src/main/java/org/apache/paimon/AbstractFileStore.java
JingsongLi: All related call links use `boolean`. (Looks like you're confusing Java and scala?)

YannByron: yep, my fault.

JingsongLi (approved):
+1 Thanks @YannByron !

PR #1680, opened 2023-07-28
[fix] Fix problem: after compaction, all the result files are regarded as small files

leaves12138:
The RollingFileWriter judge if need rolling by:
file(which is written).getPos + bytes(in memory) * judgedCompressionRate
The first part is the real size, but the second part is not accurate.
When use zstd compression, the output file size is smaller than target file size, and the output file will be regarded as small files, which cause all the files in compaction all the way.

We filter middle size file to avoid this, the file which is larger than targetFileSize * 0.7 will not participate in compaction anymore.


JingsongLi (approved):
+1

PR #1686, opened 2023-07-31
[core] Fix the value of partitions table primary Keys.

zhuangchong:
### Purpose

Fix the value of partitions table primary Keys.


JingsongLi (approved):
+1

PR #1688, opened 2023-07-31
[test] fix unstable hive catalog test

zhuangchong:
### Purpose

fix unstable hive catalog test.

See https://github.com/apache/incubator-paimon/blob/c617beb74eb94b627de05809333dbceaf1b81ad5/paimon-hive/paimon-hive-common/src/main/java/org/apache/paimon/hive/HiveTypeUtils.java#L216-L218


JingsongLi (approved):
+1, I just merged this for not blocking master. We can improve all exception assertion in future.

zhuangchong:
@JingsongLi @FangYongs Thanks, I will improve all exception assertion.

PR #1696, opened 2023-08-01
[Spark][Java api]Support truncate table

waywtdcc:
### Purpose

fix issue: https://github.com/apache/incubator-paimon/issues/1697

Support spark truncate table.The previous execution of the truncate statement did not delete the data, and the query can still find the result


waywtdcc:
#https://github.com/apache/incubator-paimon/issues/1697

YannByron:
hi @waywtdcc,  I think `Truncate` has been supported after https://github.com/apache/incubator-paimon/pull/1459. Can only put your UT in master branch to verify. Look forward to your feedback, thanks.

paimon-core/src/main/java/org/apache/paimon/table/AbstractFileStoreTable.java
```diff
@@ -82,6 +86,7 @@ public abstract class AbstractFileStoreTable implements FileStoreTable {
     protected final TableSchema tableSchema;
     protected final Lock.Factory lockFactory;
     @Nullable protected final MetastoreClient.Factory metastoreClientFactory;
+    private Set<Object> metaPaths = new HashSet<>(Arrays.asList("manifest", "schema"));
```
Zouxxyy, line 89: I think we need to think about what truncate needs to keep, users should expect to get a brand new table when using truncate

waywtdcc: I think the truncate table should be a brand new table that does not contain historical data. For example, hudi directly deletes data files, which has much higher performance than [#1459](https://github.com/apache/incubator-paimon/pull/1459)  per data traversal

waywtdcc:
@Zouxxyy hello,can you help me review this pr?

JingsongLi:
I think @YannByron is right, we just not add test.

@waywtdcc I will merge your test.

JingsongLi:
Closed this via 578c4085ad8fb2a012c3f9cb092fe711f4a1e1b5

waywtdcc:
@JingsongLi 
image
Currently, truncate traverses each piece of data and then deletes it. I think this is too inefficient. Can you add a parameter such as truncate.delete.all-history to indicate whether to delete files or traverse each piece of data to delete?

PR #1701, opened 2023-08-01
[spark]Fix spark write commit.force-compact parameter does not take effect

waywtdcc:
### Purpose
[spark]Fix spark write commit.force-compact parameter does not take effect


JingsongLi:
Hi @waywtdcc .
We don't force full-compaction for every Spark write job.

JingsongLi:
You can re-open this if you have question.

waywtdcc:
> Hi @waywtdcc . We don't force full-compaction for every Spark write job.

What is the significance of setting commit.force-compact in spark?

zhongtiancai:
> What is the significance of setting commit.force-compact in spark?
I have the same doubts  in flink, This options appears to only wait for the last compaction to complete. @JingsongLi

PR #1702, opened 2023-08-01
[flink]do not throw exception when cdc table change is null

zhangjun0x01:
### Purpose

Linked issue: close (https://github.com/apache/incubator-paimon/issues/1700)


JingsongLi (approved):
+1

PR #1704, opened 2023-08-01
[bug]  fix bug: Stream read snapshot slow while reading from history, waiting for time interval to trigger stream scan

leaves12138:
Stream read too slow, tasks will wait time interval to trigger snapshot scan. 
Every request will trigger a snapshot scan immediately (on if the last scan hasn't reach the latest one).


JingsongLi (approved):
+1

PR #1705, opened 2023-08-01
[test]  Improve all exception assertion

zhuangchong:
### Purpose

The related pr: #1688


JingsongLi (approved):
Really thanks @zhuangchong , looks good to me!

PR #1710, opened 2023-08-02
[filnk] replace the string class name with constant

zhangjun0x01:
### Purpose

I think we should better use constant instead of string


JingsongLi (approved):
+1

PR #1717, opened 2023-08-02
[doc] Use correct parameters for Presto & Trino

lvyanquan:
### Purpose
file.format is defined here
https://github.com/apache/incubator-paimon/blob/41997b2170dcfa32aa867bbc4573989911f05f61/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java#L91
bucket-key
https://github.com/apache/incubator-paimon/blob/41997b2170dcfa32aa867bbc4573989911f05f61/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java#L67
changelog-producer
https://github.com/apache/incubator-paimon/blob/41997b2170dcfa32aa867bbc4573989911f05f61/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java#L373
primary_key & partitioned_by shouldn't be changed 
https://github.com/apache/incubator-paimon-trino/blob/75360d8e0b11cf5026f5219d0c9af02afee2b984/paimon-trino-common/src/main/java/org/apache/paimon/trino/TrinoTableOptions.java#L36

Linked issue: close #xxx

### DIFF
--- a/docs/content/engines/presto.md
+++ b/docs/content/engines/presto.md
@@ -124,12 +124,12 @@ CREATE TABLE paimon.test_db.orders (
     order_date date
 )
 WITH (
-    file.format = 'ORC',
+    file_format = 'ORC',
     primary_key = ARRAY['order_key','order_date'],
-    partitioned_by = ARRAY['orderdate'],
+    partitioned_by = ARRAY['order_date'],
     bucket = '2',
-    bucket-key = 'order_key',
-    changelog-producer = 'input'
+    bucket_key = 'order_key',
+    changelog_producer = 'input'
 )
 ```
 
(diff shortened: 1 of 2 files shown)


docs/content/engines/presto.md
```diff
@@ -124,12 +124,12 @@ CREATE TABLE paimon.test_db.orders (
     order_date date
 )
 WITH (
-    file_format = 'ORC',
+    file.format = 'ORC',
```
JingsongLi, line 127: Have you tested this?

leaves12138: You can run paimon-build:
mvn spotless:apply && mvn clean install -DskipTests && mvn package -Pgenerate-docs -pl paimon-docs -nsu -DskipTests && cd paimon-core && mvn test && cd ../paimon-flink && mvn test && cd ../paimon-docs && mvn test && cd ..
in paimon base directory to test this

JingsongLi: I mean, Trino has made special treatment because "." or "-" is not allowed in trino.

lvyanquan: Thanks for pointing this and suggestion, after testing and find out that those parameters don't need change.

截屏2023-08-03 下午3 23 04

JingsongLi (approved):
+1

PR #1719, opened 2023-08-03
[core] Supports cross partition update

JingsongLi:
### Purpose

This PR aims to address the need for cross partition update capability when the primary key does not contain partitions.

### Tests

UT & IT

### Documentation

Done


leaves12138:
We can add some doc about cross partition update do not support Partial Update in Exist-Engine DELETE

leaves12138:
+1

PR #1722, opened 2023-08-03
[core] Partial update supports retraction with sequence group

JingsongLi:
### Purpose

By default, Partial update can not accept delete records, you can choose one of the following solutions:
1. Configure 'partial-update.ignore-delete' to ignore delete records.
2. Configure 'sequence-group' to retract partial columns.

Sequence group delete corresponding fields.


zhuangchong (approved):
+1

PR #1725, opened 2023-08-03
[paimon-hive-catalog] dynamically update the fields in the partition metadata when add Partition

hekaifei:
[paimon-hive-catalog] dynamically update the fields in the partition metadata when add Partition

Purpose

When the hive engine queries the partition table, it cannot find the latest fields in the partition.
The latest fields should be used when adding partitions, The fields should also be checked when the partition exists.

### DIFF
--- a/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveMetastoreClient.java
+++ b/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveMetastoreClient.java
@@ -26,6 +26,7 @@ import org.apache.paimon.schema.TableSchema;
 import org.apache.paimon.utils.PartitionPathUtils;
 import org.apache.paimon.utils.RowDataPartitionComputer;
 
+import org.apache.commons.collections.CollectionUtils;
 import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.hadoop.hive.metastore.IMetaStoreClient;
 import org.apache.hadoop.hive.metastore.api.NoSuchObjectException;
@@ -43,7 +44,6 @@ public class HiveMetastoreClient implements MetastoreClient {
     private final RowDataPartitionComputer partitionComputer;
 
     private final IMetaStoreClient client;
-    private final StorageDescriptor sd;
 
     private HiveMetastoreClient(Identifier identifier, TableSchema schema, IMetaStoreClient client)
             throws Exception {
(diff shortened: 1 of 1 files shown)


YannByron:
> When the hive engine queries the partition table, it cannot find the latest fields in the partition.

@hekaifei hi, can you explain why the latest fields can't be used in this case. I thought the latest schema should be synced to HMS, and If I remember correctly, Hive uses Table-level fields when fields is empty in `Partition`.

hekaifei:
> @hekaifei hi, can you explain why the latest fields can't be used in this case. I thought the latest schema should be synced to HMS, and If I remember correctly, Hive uses Table-level fields when fields is empty in `Partition`.

@YannByron  The original code had two problems

1. There is no problem with table schema sync to HMS， but when adding a new partition, the table schema is not obtained in real time, only initialized once

   https://github.com/apache/incubator-paimon/blob/master/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveMetastoreClient.java

   ```sql
   private final StorageDescriptor sd;
   
   private HiveMetastoreClient(Identifier identifier, TableSchema schema, IMetaStoreClient client)
           throws Exception {
       this.identifier = identifier;
       this.partitionComputer =
               new RowDataPartitionComputer(
                       new CoreOptions(schema.options()).partitionDefaultName(),
                       schema.logicalPartitionType(),
                       schema.partitionKeys().toArray(new String[0]));
   
       this.client = client;
       this.sd = client.getTable(identifier.getDatabaseName(), identifier.getObjectName()).getSd();
   }
   ```

2. An existing partition will not resolve new fields because of the partition metadata，We should update the partition metadata according to the latest data。

   one example :

   ```sql
   create table test_partition(
   c1 string
   )partitioned by (part string);
   
   insert into test_partition partition(part='part1') values('col1');
   
   alter table test_partition add columns(c2 string);
   
   insert into test_partition partition(part='part1') values('col1','col2');
   
   -- query result 
   hive> select * from test_partition where part='part1';
   OK
   col1    NULL    part1
   col1    NULL    part1
   ```

   If update partition metadata , We can query the new field in existing partitions

   like doing this：

   ```sql
   alter table test_partition partition(part='part1') add columns(c2 string);
    
   hive> select * from test_partition where part='part1';
   OK
   col1    NULL    part1
   col1    col2    part1
   ```

JingsongLi:
@hekaifei Thanks for your contribution.

I saw the solution from @YannByron .

We may need some testing to see whether "`Hive uses Table-level fields when fields is empty in Partition`" can work.

PR #1726, opened 2023-08-03
[hive] Support use partitioned by to create hive partition table.

zhuangchong:
### Purpose

Support use `partitioned by` to create hive partition table.


JingsongLi (approved):
+1

PR #1727, opened 2023-08-03
[flink][mysql-cdc] All FileStoreWrites generated in database synchronization job should use the same compaction executor service

yuzelin:
### Tests

`CdcRecordStoreMultiWriteOperatorTest#testUsingTheSameCompactExecutor`

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreWrite.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreWrite.java
@@ -70,6 +70,7 @@ public abstract class AbstractFileStoreWrite<T>
     protected final Map<BinaryRow, Map<Integer, WriterContainer<T>>> writers;
 
     private ExecutorService lazyCompactExecutor;
+    private boolean closeCompactExecutorWhenLeaving = true;
     private boolean ignorePreviousFiles = false;
     protected boolean isStreamingMode = false;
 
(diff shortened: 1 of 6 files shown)


paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreWrite.java
```diff
@@ -102,6 +102,10 @@ public void withIgnorePreviousFiles(boolean ignorePreviousFiles) {
         this.ignorePreviousFiles = ignorePreviousFiles;
     }
 
+    public void withCompactExecutor(ExecutorService compactExecutor) {
+        this.lazyCompactExecutor = compactExecutor;
```
JingsongLi, line 106: should not close it when close write.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/cdc/CdcRecordStoreMultiWriteOperator.java
```diff
@@ -96,6 +101,10 @@ public void initializeState(StateInitializationContext context) throws Exception
         state = new StoreSinkWriteState(context, (tableName, partition, bucket) -> true);
         tables = new HashMap<>();
         writes = new HashMap<>();
+        compactExecutor =
```
JingsongLi, line 104: close this, in close.

JingsongLi (approved):
+1

PR #1734, opened 2023-08-04
[flink] add option to support disable creating table in default database

legendtkl:
### Purpose

Linked issue: close https://github.com/apache/incubator-paimon/issues/1729

### Tests

FlinkCatalogTest

### DIFF
--- a/docs/layouts/shortcodes/generated/catalog_configuration.html
+++ b/docs/layouts/shortcodes/generated/catalog_configuration.html
@@ -27,6 +27,12 @@ under the License.
     </thead>
     <tbody>
         <tr>
+            <td><h5>disable-create-table-in-default-db</h5></td>
+            <td style="word-wrap: break-word;">false</td>
+            <td>Boolean</td>
+            <td>If true, creating table in default database is not allowed. Default is false.</td>
+        </tr>
+        <tr>
             <td><h5>fs.allow-hadoop-fallback</h5></td>
             <td style="word-wrap: break-word;">true</td>
             <td>Boolean</td>
--- a/docs/layouts/shortcodes/generated/flink_catalog_configuration.html
+++ b/docs/layouts/shortcodes/generated/flink_catalog_configuration.html
@@ -33,6 +33,12 @@ under the License.
             <td></td>
         </tr>
         <tr>
+            <td><h5>disable-create-table-in-default-db</h5></td>
+            <td style="word-wrap: break-word;">false</td>
+            <td>Boolean</td>
+            <td>If true, creating table in default database is not allowed. Default is false.</td>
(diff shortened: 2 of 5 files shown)


zhuangchong:
I understand that it is supported in the paimon catalog, not just in the Flink engine. cc @FangYongs

FangYongs:
@zhuangchong Agree, and @legendtkl could you move this to paimon catalog instead of flink catalog?

legendtkl:
> @zhuangchong Agree, and @legendtkl could you move this to paimon catalog instead of flink catalog?

Got it. @FangYongs @zhuangchong

legendtkl:
Hi, @FangYongs , @zhuangchong , I have addressed the review comment, would you help review it?

Thanks.

paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
```diff
@@ -205,4 +209,11 @@ public static Path databasePath(String warehouse, String database) {
     protected boolean isSystemDatabase(String database) {
         return SYSTEM_DATABASE_NAME.equals(database);
     }
+
+    protected boolean allowCreateTableInDefaultDB(Identifier identifier) {
+        return this.options
+                .getOptional(DISABLE_CREATE_TABLE_IN_DEFAULT_DB)
+                .map(val -> !(val & DEFAULT_DATABASE.equals(identifier.getDatabaseName())))
```
FangYongs, line 216: Use `&&`？

FangYongs: In addition, users can configure default database name with option `default-database` for flink

FangYongs:
Thanks @legendtkl , left some comments

legendtkl:
Hi, @FangYongs and @zhuangchong , let's have a discussion about the implementation.

1. Now flink engine support to specify the default database name with option `default-database` for flink, this will be stored in `org.apache.flink.table.catalog.AbstractCatalog`
2. But for paimon catalog, we can't access the default database specified by flink. If we check it in Paimon catalog, the info is missed.
3. So I propose to support this issue in FlinkCatalog now only. For Paimon default database support we can open a new ticket for it.

What do you think?

FangYongs:
Thanks @legendtkl , it sounds good to me to support flink catalog in this PR

FangYongs:
@legendtkl There're some conflicts

legendtkl:
@FangYongs resolved, PTAL.

Thanks.

FangYongs:
Thanks @legendtkl , +1

FangYongs (approved):

PR #1738, opened 2023-08-04
[fs] Introduce requiredOptions to FileIOLoader

JingsongLi:
### Purpose

At present, when we put Paimon-s3.jar into the production environment, we must configure s3 related configurations, and fallback FileIO will not be useful.

We can actually make them coexist and introduce the requiredOptions mechanism to use fallback FileIO.

It is not recommended to use `FileIOUtils.checkAccess` here. Accessing real S3FileIO may cause decompression of temporary files and may also cause blockages.


leaves12138:
+1

PR #1739, opened 2023-08-05
[Flink] Support numBytesSend metrics for paimon sink

### LINKED ISSUE #1737: [Feature] Support numBytesSend metric for paimon sink
### Search before asking

### Motivation

As flip-33 mentioned, we need to implement sink metrics, the numBytesSend metrics is the first important metric we support.

### Solution

_No response_

### Anything else?

_No response_

schnappi17:
### Purpose

Linked issue: close #1737 

Update counter when commit with committables, the value is the size of data files. The counter metric is already registered by flink.

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/StoreCommitter.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/StoreCommitter.java
@@ -95,7 +95,12 @@ public class StoreCommitter implements Committer<Committable, ManifestCommittabl
         for (ManifestCommittable committable : committables) {
             List<CommitMessage> commitMessages = committable.fileCommittables();
             for (CommitMessage commitMessage : commitMessages) {
-                long dataFileSizeInc = ((CommitMessageImpl) commitMessage).newFilesIncrement().newFiles().stream().mapToLong(f -> f.fileSize()).reduce(Long::sum).getAsLong();
+                long dataFileSizeInc =
+                        ((CommitMessageImpl) commitMessage)
+                                .newFilesIncrement().newFiles().stream()
+                                        .mapToLong(f -> f.fileSize())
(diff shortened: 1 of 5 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/StoreCommitter.java
```diff
@@ -83,4 +88,22 @@ public Map<Long, List<Committable>> groupByCheckpoint(Collection<Committable> co
...
+
+    @VisibleForTesting
+    static long calcDataBytesSend(List<ManifestCommittable> committables) {
+        long bytesSend = 0;
+        for (ManifestCommittable committable : committables) {
+            List<CommitMessage> commitMessages = committable.fileCommittables();
+            for (CommitMessage commitMessage : commitMessages) {
+                long dataFileSizeInc =
+                        ((CommitMessageImpl) commitMessage)
+                                .newFilesIncrement().newFiles().stream()
```
tsreaper, line 100: Compact files and changelogs should also be calculated.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/StoreCommitter.java
```diff
@@ -60,8 +64,9 @@ public ManifestCommittable combine(
     }
 
     @Override
-    public void commit(List<ManifestCommittable> committables)
+    public void commit(List<ManifestCommittable> committables, OperatorIOMetricGroup metricGroup)
             throws IOException, InterruptedException {
+        metricGroup.getNumBytesOutCounter().inc(calcDataBytesSend(committables));
```
tsreaper, line 69: Count this only after the commit is successful. Otherwise if the commit is retried through fail over, this will be counted multiple times.

tsreaper (approved):

PR #1740, opened 2023-08-05
[Improve][Doc][Flink]Improve some small content

MonsterChenzhuo:
### Purpose
1. Ensure uniformity of semicolon termination for SQL statements within the document.

2. Refactor the logic inside `getSchemaChanges` into a separate method to eliminate redundancy.

3. some method invocations may result in 'NullPointerException'. It would be more appropriate to perform null checks beforehand instead of waiting for the invocation to trigger it.

4. Be cautious about unboxing 'nextSnapshotId', as it may lead to 'java.lang.NullPointerException'.

5. Note that '@Nonnull' annotation cannot be used on members of primitive types.

6. Improve the performance of `list.containsAll(collection)` by wrapping 'fieldNames' in a `HashSet` constructor.

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/align/AlignedContinuousFileSplitEnumerator.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/align/AlignedContinuousFileSplitEnumerator.java
@@ -220,8 +220,9 @@ public class AlignedContinuousFileSplitEnumerator extends ContinuousFileSplitEnu
         PlanWithNextSnapshotId nextPlan = pendingPlans.poll();
         if (nextPlan != null) {
             nextSnapshotId = nextPlan.nextSnapshotId();
+            Objects.requireNonNull(nextSnapshotId);
             TableScan.Plan plan = nextPlan.plan();
-            if (plan.splits().isEmpty() && nextSnapshotId != null) {
+            if (plan.splits().isEmpty()) {
                 addSplits(
                         Collections.singletonList(
                                 new FileStoreSourceSplit(
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/align/AlignedContinuousFileSplitEnumerator.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/align/AlignedContinuousFileSplitEnumerator.java
@@ -45,6 +45,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.TreeMap;
 import java.util.concurrent.ArrayBlockingQueue;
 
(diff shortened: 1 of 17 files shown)


MonsterChenzhuo:
@JingsongLi PTAL,thanks.

JingsongLi:
@MonsterChenzhuo Thanks for your contribution! I will take a look this afternoon.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/align/AlignedContinuousFileSplitEnumerator.java
```diff
@@ -220,7 +221,7 @@ protected void processDiscoveredSplits(PlanWithNextSnapshotId ignore, Throwable
         if (nextPlan != null) {
             nextSnapshotId = nextPlan.nextSnapshotId();
             TableScan.Plan plan = nextPlan.plan();
-            if (plan.splits().isEmpty()) {
+            if (plan.splits().isEmpty() && nextSnapshotId != null) {
```
JingsongLi, line 224: `checkArguement(nextSnapshotId != null)`?

MonsterChenzhuo: yeah
图片

JingsongLi: I mean we should just `checkArguement` instead of falling into the next branch.

MonsterChenzhuo: 图片
I noticed that the constructor defines nextSnapshotId as nullable. I'm not sure whether the else logic, specifically addSplits(splitGenerator.createSplits(plan));, will be triggered when nextSnapshotId is null. Since I'm not very familiar with the code logic, I am uncertain about the best approach.

1：
if (nextPlan != null) {
            nextSnapshotId = nextPlan.nextSnapshotId();
            Objects.requireNonNull(nextSnapshotId);
            TableScan.Plan plan = nextPlan.plan();
            if (plan.splits().isEmpty()) {
                addSplits(
                        Collections.singletonList(
                                new FileStoreSourceSplit(
                                        PLACEHOLDER_SPLIT,
                                        new PlaceholderSplit(nextSnapshotId - 1))));
            } else {
                addSplits(splitGenerator.createSplits(plan));
            }
        }
2：
  if (nextPlan != null) {
      nextSnapshotId = nextPlan.nextSnapshotId();
      TableScan.Plan plan = nextPlan.plan();
      if (plan.splits().isEmpty()) {
          Objects.requireNonNull(nextSnapshotId);
          addSplits(
                  Collections.singletonList(
                          new FileStoreSourceSplit(
                                  PLACEHOLDER_SPLIT,
                                  new PlaceholderSplit(nextSnapshotId - 1))));
      } else {
          addSplits(splitGenerator.createSplits(plan));
      }
  }

MonsterChenzhuo: done.

JingsongLi (approved):
+1 Thanks @MonsterChenzhuo

PR #1749, opened 2023-08-07
[flink] Support source metrics of pending records

### LINKED ISSUE #1748: [Feature] Support records lag metric for paimon source
### Search before asking

### Motivation

Support records lag metric for paimon source in streaming mode.

### Solution

1. Each reader report the split's row count when fetching a new split to a map with key of subtask(reader).
2. Split enumerator acc the discovered splits' row count when scanning the next snapshot.
3. The gauge registered calc with `discoveredRecords` - sum of all readers fetched splits' row count.

### Anything else?

_No response_

schnappi17:
### Purpose

Linked issue: close #1748


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/ContinuousFileStoreSource.java
```diff
@@ -40,6 +48,7 @@ public class ContinuousFileStoreSource extends FlinkSource {
 
     protected final Map<String, String> options;
     protected final BucketMode bucketMode;
+    protected ConcurrentHashMap<Integer, Long> consumedRecords = new ConcurrentHashMap<>();
```
tsreaper, line 51: `SourceReader` runs in task manager, while `SplitEnumerator` runs in job manager. They can't communicate only through memory.

tsreaper (changes requested):

PR #1753, opened 2023-08-07
[core] flink file clean action

### LINKED ISSUE #1571: [Feature] Introduce file clean action
### Search before asking

### Motivation

Currently, the `FileStoreExpireImpl` delete file quietly, which means if there occurs error, the process won't throw exceptions. It's possible that some files won't be deleted when snapshot expiring, so I propose to introduce an action to clean files unused by current snapshots. 

### Solution

Scan all files and filter out that still used by at least one snapshot, then delete those not used by any snapshot.

### Anything else?

_No response_

john8628:
### Purpose

Linked issue: close #1571

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/FileCleanAction.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/FileCleanAction.java
@@ -1,0 +1,205 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.paimon.flink.action;
+
+import org.apache.paimon.fs.FileIO;
+import org.apache.paimon.fs.FileStatus;
+import org.apache.paimon.fs.Path;
+import org.apache.paimon.table.AbstractFileStoreTable;
+import org.apache.paimon.table.FileStoreTable;
+
+import org.apache.commons.compress.utils.Lists;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.types.Row;
+import org.apache.flink.util.CloseableIterator;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Predicate;
+
+/** File Clean table orphan file action for Flink. */
+public class FileCleanAction extends TableActionBase {
+
+    private static final Logger LOG = LoggerFactory.getLogger(FileCleanAction.class);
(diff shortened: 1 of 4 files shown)


zhangjun0x01:
hi ,@john8628 , could you add the test case and document ?

zhangjun0x01:
I think we should not only check the data files, but also check metadata

john8628:
> hi ,@john8628 , could you add the test case and document ?
Thank you for your review. I will incorporate these shortly, once I am certain about the legality of my implementation.

> I think we should not only check the data files, but also check metadata
yean metadata should not deleted;  I will implement checks to ensure they are not deleted and remain intact.

zhangjun0x01:
In addition, it is also very important to add an    `olderThan` ,  because the process of writing data is to first write data files. and then write snapshot. When we have written the data but have not yet submitted the snapshot, if the clean action is running at this time, it will delete the newly written data.

john8628:
> In addition, it is also very important to add an `olderThan` , because the process of writing data is to first write data files. and then write snapshot. When we have written the data but have not yet submitted the snapshot, if the clean action is running at this time, it will delete the newly written data.

Thanks for your raising this important consideration;  You're absolutely right, including an "olderThan" parameter is essential in order to prevent the accidental deletion of newly written data when the clean action is executed during the period between writing data files and submitting snapshot;

yuzelin:
Hi, please run `mvn spotless::apply` in your dev environment to fix Checkstyle violation.

john8628:
> Hi, please run `mvn spotless::apply` in your dev environment to fix Checkstyle violation.

OK , thanks , i will do it ;

john8628:
@yuzelin @zhangjun0x01 @JingsongLi PTAL

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/FileCleanAction.java
```diff
@@ -0,0 +1,205 @@
...
+                "Scan all files and filter out that still used by at least one snapshot, then delete those not used by any snapshot..");
+        // 1: get all files  of the table path
+        List<String> allFileList = buildAllFileList(basePath.getPath());
+        // 2: get all the used files of the table;exclude the manifest , schema  and snapshot files;
+        List<String> validFileNameList = buildValidFileNameList();
+        // 3: get the diff of  the files
+        allFileList.remove(validFileNameList);
+        LOG.info("orphan files:{}", allFileList.size());
+        // 4:delete the file
+        allFileList.stream()
```
wg1026688210, line 84: How about adding a thread num opt.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/FileCleanAction.java
```diff
@@ -0,0 +1,205 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Predicate;
+
+/** File Clean table orphan file action for Flink. */
+public class FileCleanAction extends TableActionBase {
+
+    private static final Logger LOG = LoggerFactory.getLogger(FileCleanAction.class);
+    private static final int MAX_DRIVER_LISTING_DEPTH = 3;
```
wg1026688210, line 46: Do we need to infer based on the number of partition fields and bucket directories?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/FileCleanActionFactory.java
```diff
@@ -0,0 +1,64 @@
...
+        System.out.println();
+
+        System.out.println("Syntax:");
+        System.out.println(
+                "  file-clean --warehouse <warehouse-path> --database <database-name> "
+                        + "--table <table-name> ");
+
+        System.out.println("Examples:");
+        System.out.println(
+                "  file-clean --path hdfs:///path/to/warehouse/test_db.db/test_table --where id > (SELECT count(*) FROM employee)");
```
wg1026688210, line 60: what 's the usage of `where` opt

wg1026688210 (commented):
Left some minor comments

john8628:
> Left some minor comments

thanks for your reply, good  idea；

john8628:
sorry , i can't continue to work for this PR ,

zhangjun0x01:
> sorry , i can't continue to work for this PR ,

I would like to do it .

PR #1758, opened 2023-08-08
[flink]Improve get kafka schema

MonsterChenzhuo:
### Purpose

The current configuration for retry attempts and retry intervals seems inappropriate: the number of retry attempts is set at 100, and the retry interval is set at 100 milliseconds. It is suggested to adjust these values to a more reasonable setting, such as 5 retry attempts and a retry interval of 1000 milliseconds. Additionally, implementing an exponential backoff strategy would be beneficial, which progressively increases the retry waiting time with each attempt.

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/cdc/kafka/KafkaSchema.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/cdc/kafka/KafkaSchema.java
@@ -41,7 +41,8 @@ import java.util.Properties;
 /** Utility class to load canal kafka schema. */
 public class KafkaSchema {
 
-    private static final int MAX_RETRY = 100;
+    private static final int MAX_RETRY = 5;
+    private static final int POLL_TIMEOUT_MILLIS = 100;
 
     private final String databaseName;
     private final String tableName;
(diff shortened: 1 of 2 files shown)


MonsterChenzhuo:
@JingsongLi  PTAL,thanks.

JingsongLi:
> It is not ideal that the KafkaSchema#getKafkaSchema() method could potentially execute the new CanalRecordParser operation without checking if records are available first. It is recommended to include a check for the presence of records before performing the new CanalRecordParser operation.

I cannot get it. What do you mean?

MonsterChenzhuo:
> > It is not ideal that the KafkaSchema#getKafkaSchema() method could potentially execute the new CanalRecordParser operation without checking if records are available first. It is recommended to include a check for the presence of records before performing the new CanalRecordParser operation.
> 
> I cannot get it. What do you mean?
If records is null, the process will proceed to create a new CanalRecordParser. What I mean is, before creating a new CanalRecordParser, there should be a check to ensure that records is not empty.
图片

MonsterChenzhuo:
> > It is not ideal that the KafkaSchema#getKafkaSchema() method could potentially execute the new CanalRecordParser operation without checking if records are available first. It is recommended to include a check for the presence of records before performing the new CanalRecordParser operation.
> 
> I cannot get it. What do you mean?

This passage was written incorrectly, and I have already deleted it. Please disregard it.

JingsongLi:
@MonsterChenzhuo So you can revert changes, and just modify sleep time?

MonsterChenzhuo:
> @MonsterChenzhuo So you can revert changes, and just modify sleep time?

Yeah

JingsongLi (approved):
+1

PR #1762, opened 2023-08-08
[core] The config of changelog format and file format are separated

chestnutqiang:
### Purpose

Linked issue: none

### Tests

org.apache.paimon.flink.ChangelogWithKeyFileStoreTableITCase#testInputChangelogProducerBatchRandom
org.apache.paimon.flink.ChangelogWithKeyFileStoreTableITCase#testInputChangelogProducerStreamingRandom
org.apache.paimon.flink.ChangelogWithKeyFileStoreTableITCase#testStandAloneInputJobRandom

### API and Format

none

### Documentation

auto generate

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
@@ -139,6 +139,20 @@ public class CoreOptions implements Serializable {
                             "Default file compression format, can be overridden by "
                                     + FILE_COMPRESSION_PER_LEVEL.key());
 
+    public static final ConfigOption<FileFormatType> CHANGELOG_FORMAT =
+            key("changelog.format")
+                    .enumType(FileFormatType.class)
+                    .defaultValue(FileFormatType.ORC)
+                    .withDescription(
+                            "Specify the message format of changelog files, currently orc, parquet and avro are supported.");
+
+    public static final ConfigOption<String> CHANGELOG_COMPRESSION =
+            key("changelog.compression")
+                    .stringType()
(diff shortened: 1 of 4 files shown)


paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -139,6 +138,40 @@ public class CoreOptions implements Serializable {
...
+                    .withDescription(
+                            "Specify the message format of changelog files, currently orc, parquet and avro are supported.");
+
+    public static final ConfigOption<String> CHANGELOG_COMPRESSION =
+            key("changelog.compression")
+                    .stringType()
+                    .noDefaultValue()
+                    .withDescription("Default changelog compression format.");
+
+    public static final ConfigOption<Map<String, String>> CHANGELOG_COMPRESSION_PER_LEVEL =
```
JingsongLi, line 154: I don't want to introduce per level for changelog.
Can we just introduce unified format and compression?

chestnutqiang: It depends on whether any compaction happens on the changelog and how small a file there is if you don't compate. I mean, changelog doesn't need to sort even when it does small file merges because it's in chronological order.

JingsongLi: Yes, so there should be no per level

JingsongLi:
Thanks @chestnutqiang for the contribution.

But the modification is too big. I still not see the strong requirement to separate the config of changelog format and file format.

PR #1770, opened 2023-08-09
[flink] Flink Lookup supports async mode

JingsongLi:
### Purpose

For Flink Lookup Join:

### Retry Lookup

If the records of `Orders` (main table) join missing because the corresponding data of `customers` (lookup table) is not ready.
You can consider using Flink's [Delayed Retry Strategy For Lookup](https://nightlies.apache.org/flink/flink-docs-master/docs/dev/table/sql/queries/hints/#3-enable-delayed-retry-strategy-for-lookup).
Only for Flink 1.16+.

```sql
-- enrich each order with customer information
SELECT /*+ LOOKUP('table'='c', 'retry-predicate'='lookup_miss', 'retry-strategy'='fixed_delay', 'fixed-delay'='1s', 'max-attempts'='600') */
o.order_id, o.total, c.country, c.zip
FROM Orders AS o
JOIN customers
FOR SYSTEM_TIME AS OF o.proc_time AS c
ON o.customer_id = c.id;
```

### Async Retry Lookup

The problem with synchronous retry is that one record will block subsequent records, causing the entire job to be blocked.
You can consider using async to avoid blocking.

```sql
-- enrich each order with customer information
SELECT /*+ LOOKUP('table'='c', 'retry-predicate'='lookup_miss', 'async'='true', 'output-mode'='allow_unordered', 'retry-strategy'='fixed_delay', 'fixed-delay'='1s', 'max-attempts'='600') */
o.order_id, o.total, c.country, c.zip
FROM Orders AS o
JOIN customers
FOR SYSTEM_TIME AS OF o.proc_time AS c
ON o.customer_id = c.id;
```


docs/content/how-to/lookup-joins.md
```diff
@@ -78,14 +82,42 @@ FOR SYSTEM_TIME AS OF o.proc_time AS c
...
+FROM Orders AS o
+JOIN customers
+FOR SYSTEM_TIME AS OF o.proc_time AS c
+ON o.customer_id = c.id;
+```
+
+### Async Retry Lookup
+
+The problem with synchronous retry is that one record will block subsequent records, causing the entire job to be blocked.
+You can consider using async to avoid blocking.
```
Myasuka, line 104: I think the output mode of `allow_unordered` should be also mentioned in the docs. Unordered output might produce unexpected results, however, ordered output could impact the performance significantly. We cannot say async retry could resolve the poor performance without unordered output.

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/LookupJoinITCase.java
```diff
@@ -517,6 +517,29 @@ public void testRetryLookup() throws Exception {
         iterator.close();
     }
 
+    @Test
+    public void testAsyncRetryLookup() throws Exception {
```
Myasuka, line 521: It seems this unit test does not have difference with existed `testRetryLookup`, do you think we should test the `async` behavior with existing sync lookup tests?

Myasuka (commented):
Thanks for creating this PR to cover my previous one.
I have left some comments, please take a look.

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/LookupJoinITCase.java
JingsongLi: It is better to separate them. Every test should launch a streaming job. It is good to just launch one job in one test.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkConnectorOptions.java
```diff
@@ -224,6 +224,18 @@ public class FlinkConnectorOptions {
...
+    public static final ConfigOption<Boolean> LOOKUP_ASYNC =
+            ConfigOptions.key("lookup.async")
+                    .booleanType()
+                    .defaultValue(false)
+                    .withDescription("Whether to enable async lookup join.");
+
+    public static final ConfigOption<Integer> LOOKUP_ASYNC_THREAD_NUMBER =
+            ConfigOptions.key("lookup.async-thread-number")
+                    .intType()
+                    .defaultValue(16)
```
Myasuka, line 236: I remember that you ever mentioned that the number of async threads cannot be larger than 1 as the rocksDB cache is not thread-safe.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkConnectorOptions.java
```diff
@@ -224,6 +224,18 @@ public class FlinkConnectorOptions {
                     .withDescription(
                             "If the new snapshot has not been generated when the checkpoint starts to trigger, the enumerator will block the checkpoint and wait for the new snapshot. Set the maximum waiting time to avoid infinite waiting, if timeout, the checkpoint will fail. Note that it should be set smaller than the checkpoint timeout.");
 
+    public static final ConfigOption<Boolean> LOOKUP_ASYNC =
+            ConfigOptions.key("lookup.async")
```
Myasuka, line 228: Just as I mentioned in previous PR, it's a bit strange to mix the hint of `Lookup('async' = 'true')` with the `lookup.async` option.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkConnectorOptions.java
JingsongLi: So there is a lock to lookup.

This parameter is due to the implementation of Flink. Currently, in the case of join missing, the Flink implementation will sleep and wait, which will cause the thread to be blocked. So we need more threads to ensure allow_unordered availability.

See `RetryableAsyncLookupFunctionDelegator.lookupWithRetry` in Flink.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkConnectorOptions.java
JingsongLi: Yes, but it is a necessity.

Flink will choose asynchrony by default, but we do not want to provide asynchrony by default because asynchrony is meaningless.

So I introduce this option to provide sync mode by default.

Myasuka: I see, since Flink's `preferAsync` [returns true except async option in hint is false](https://github.com/apache/flink/blob/master/flink-table/flink-table-planner/src/main/java/org/apache/flink/table/planner/plan/utils/LookupJoinUtil.java#L482-L485), how about we could simplify the example to not include the hint of `Lookup('async' = 'true')`?

JingsongLi: Good point! You are right!

xwmr-max:
Hi, @JingsongLi ,currently in paimon, flink look up join is leveraging rocksdb's backend storage capabilities. This is a conversion from the original memory to back-end storage, and the timeliness is not inferior. However, I have a question: paimon has the storage capability of LSM, why not use the ordering capability of LSM to implement lookup?

JingsongLi:
> Hi, @JingsongLi ,currently in paimon, flink look up join is leveraging rocksdb's backend storage capabilities. This is a conversion from the original memory to back-end storage, and the timeliness is not inferior. However, I have a question: paimon has the storage capability of LSM, why not use the ordering capability of LSM to implement lookup?

Yes, we can change to lookup LSM directly if the lookup key is the primary key of the table. This is a TODO.

Myasuka (approved):
LGTM.

PR #1771, opened 2023-08-09
[core] improve_universal_compaction

zhangjun0x01:
### Purpose

Linked issue: close https://github.com/apache/incubator-paimon/issues/1769


JingsongLi:
I think this is worthing to do, but related to core mechanism, we should think carefully.

zhangjun0x01:
> I think this is worthing to do, but related to core mechanism, we should think carefully.

yes， did you has some suggestion ?  thanks

PR #1784, opened 2023-08-10
[core] Validate changelog producer on table without primary keys

### LINKED ISSUE #1782: [Feature] Validate write-mode: auto and changelog producer not NONE conflict if no primary key
### Search before asking

### Motivation

Now we have table schema validation on WriteMode.APPEND_ONLY  and changelog-produer not NONE. This can throw runtime exception when users set the writeMode to APPEND_ONLY explicitly. 

But when there is a table without primary key and user set the changelog-producer to something else but not NONE, this validation is passed at first because the writeMode is AUTO  when validating. Then paimon will help users to set the writeMode to APPEND_ONLY  because there is no primary key in table.

Thus we have a table without primary key, with writeMode APPEND_ONLY and changelog-producer != NONE. This will cause the reader waiting for changelog files but the writer doesn't have any changelog to write. 

### Solution

_No response_

### Anything else?

_No response_

schnappi17:
### Purpose

Linked issue: close #1782

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
+++ b/paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
@@ -97,7 +97,7 @@ public class SchemaValidation {
                 && changelogProducer != ChangelogProducer.NONE) {
             throw new UnsupportedOperationException(
                     String.format(
-                            "Can not set %s on table without primary keys.",
+                            "Can not set %s on table without primary keys, please define primary keys.",
                             CHANGELOG_PRODUCER.key()));
         }
 
--- a/paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
+++ b/paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
@@ -92,6 +92,15 @@ public class SchemaValidation {
                             WRITE_MODE.key(), APPEND_ONLY, CHANGELOG_PRODUCER.key()));
         }
 
+        if (options.writeMode() == WriteMode.AUTO
+                && schema.primaryKeys().isEmpty()
+                && changelogProducer != ChangelogProducer.NONE) {
+            throw new UnsupportedOperationException(
+                    String.format(
+                            "Can not set %s on table without primary keys.",
+                            CHANGELOG_PRODUCER.key()));
+        }
+
         checkArgument(
                 options.snapshotNumRetainMin() > 0,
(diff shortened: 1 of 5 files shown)


schnappi17:
@JingsongLi Could you help to review this pr?

paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
```diff
@@ -92,6 +92,15 @@ public static void validateTableSchema(TableSchema schema) {
                             WRITE_MODE.key(), APPEND_ONLY, CHANGELOG_PRODUCER.key()));
         }
 
+        if (options.writeMode() == WriteMode.AUTO
+                && schema.primaryKeys().isEmpty()
```
SteNicholas, line 96:
```suggestion
                && CollectionUtil.isNullOrEmpty(schema.primaryKeys())
```

paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
```diff
@@ -92,6 +92,15 @@ public static void validateTableSchema(TableSchema schema) {
                             WRITE_MODE.key(), APPEND_ONLY, CHANGELOG_PRODUCER.key()));
         }
 
+        if (options.writeMode() == WriteMode.AUTO
```
SteNicholas, line 95: Should we check the write mode?

paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
```diff
@@ -92,6 +92,15 @@ public static void validateTableSchema(TableSchema schema) {
...
+                && CollectionUtil.isNullOrEmpty(schema.primaryKeys())
+                && changelogProducer != ChangelogProducer.NONE) {
+            throw new UnsupportedOperationException(
+                    String.format(
+                            "Can not set %s on table without primary keys.",
```
SteNicholas, line 100: We could improve the exception message to help users to solve, like `when set write mode to none, the table must have primary key` etc. Meanwhile, adds this information into document to reduce the situation that the users have this validation failure.

SteNicholas (approved):
LGTM.

schnappi17:
Error:  Failures: 
Error:    MySqlSyncDatabaseTableListITCase.testActionRunResult:133

yuzelin:
I am trying to fix that error, I'll ping you after I fix it.

PR #1796, opened 2023-08-11
[flink] Fix to  use : instead of -> for operator name

hehuiyuan:
### Purpose

Ambiguous operator name, we can make it more clearer.
image

 use : instead of -> for operator name


JingsongLi (approved):
+1 I second you!

PR #1798, opened 2023-08-11
[hive] extern table remove file by mistake when create table failed

wg1026688210:
### Purpose
When attempting to create a  hive extern table  for an existing Paimon, if the table creation fails, the Paimon table will be deleted, which is not expected.
The pr is for reproduce the problem,and then fix it

### Tests
CreateTableITCase#createExternTableFailing


paimon-hive/paimon-hive-connector-common/src/main/java/org/apache/paimon/hive/PaimonMetaHook.java
```diff
@@ -62,6 +63,9 @@ public class PaimonMetaHook implements HiveMetaHook {
     private static final String COMMENT = "comment";
     private final Configuration conf;
 
+    // paimon table existed before create hive table
+    private Set<Identifier> existingPaimonTable = new HashSet<>();
```
JingsongLi, line 67: final

JingsongLi (approved):
+1

PR #1800, opened 2023-08-11
[flink][mysql-cdc] MySQL CDC supports convert TIME type to STRING type when using Hive catalog

yuzelin:
### Purpose

This is because Hive doesn't support TIME.

### Tests
`MySqlSyncTableActionITCase#testSpecialCastRules`

### Documentation

Refactor cdc doc


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/cdc/mysql/MySqlSyncDatabaseAction.java
```diff
@@ -172,13 +174,18 @@ public void build(StreamExecutionEnvironment env) throws Exception {
                         + "use mysql-sync-table instead.");
         boolean caseSensitive = catalog.caseSensitive();
 
+        SpecialCastRules specialCastRules =
+                new SpecialCastRules(
+                        mySqlConfig.get(MYSQL_CONVERTER_TINYINT1_BOOL),
+                        catalog instanceof HiveCatalog);
```
JingsongLi, line 180: Can Catalog returns a `List<DataTypeRoot> unsupportedTypes()`?

JingsongLi:
Or can we just support this conversion in `HiveCatalog`?

PR #1801, opened 2023-08-11
[doc] refine the paimon-doc readme doc

legendtkl:
### Purpose

refine the paimon-doc readme, and give the more accurate expression for doc generating.


JingsongLi (approved):
+1

PR #1810, opened 2023-08-14
[core] Fix that tag auto creation cannot work after restarting job

yuzelin:
### Purpose

After restarting job, the `TagAutoCreation` may be initialized to expired snapshot. 

### Tests

`TagAutoCreationTest#testTag`

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/tag/TagAutoCreation.java
+++ b/paimon-core/src/main/java/org/apache/paimon/tag/TagAutoCreation.java
@@ -107,7 +107,8 @@ public class TagAutoCreation {
                     firstNonNull(snapshotManager.earliestSnapshotId(), FIRST_SNAPSHOT_ID);
         } else {
             Snapshot lastTag = tags.lastKey();
-            this.nextSnapshot = lastTag.id() + 1;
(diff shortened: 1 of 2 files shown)


paimon-core/src/main/java/org/apache/paimon/tag/TagAutoCreation.java
```diff
@@ -115,6 +115,8 @@ private TagAutoCreation(
     }
 
     public void run() {
+        // avoid snapshot has been expired
+        nextSnapshot = Math.max(snapshotManager.earliestSnapshotId(), nextSnapshot);
```
JingsongLi, line 119:
```
    public void run() {
        while (true) {
            if (snapshotManager.snapshotExists(nextSnapshot)) {
                tryToTag(snapshotManager.snapshot(nextSnapshot));
                nextSnapshot++;
            } else {
                Long earliest = snapshotManager.earliestSnapshotId();
                if (earliest != null && earliest > nextSnapshot) {
                    nextSnapshot = earliest;
                } else {
                    break;
                }
            }
        }
    }
```

JingsongLi (approved):
+1

PR #1814, opened 2023-08-14
[core]CI execution optimization

MonsterChenzhuo:
### Purpose

-T 1C to speed up build efficiency

图片

after change：

图片

### DIFF
--- a/.github/workflows/check-licensing.yml
+++ b/.github/workflows/check-licensing.yml
@@ -37,7 +37,6 @@ jobs:
         with:
           java-version: 8
           distribution: 'adopt'
-
       - name: Build
         run: |
           set -o pipefail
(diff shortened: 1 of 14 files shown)


JingsongLi:
Let's see if optimization is effective.

zhuangchong:
No need to use `mvnw`.

the related pr: #1178

MonsterChenzhuo:
> No need to use `mvnw`.
> 
> the related pr: #1178

Thanks for the reminder, I have removed mvnw

MonsterChenzhuo:
@JingsongLi PTAL

MonsterChenzhuo:
> Let's see if optimization is effective.

@JingsongLi PTAL,thanks.

A few minutes of optimization.Is this PR still worth merging?
图片

after change：

图片

JingsongLi (approved):
+1

This PR is not complicate.
And it looks like really improve a bit.
So I think it is worth to be merged.

PR #1817, opened 2023-08-15
[doc] Added a link to the configurations documentation in the `Options Table` section of the system table documentation.

zhuangchong:
### Purpose

Added a link to the configurations documentation in the `Options Table` section of the system table documentation.

before change:
image

after change:
iShot_2023-08-15_13 06 33


JingsongLi (approved):
+1

PR #1823, opened 2023-08-15
[core] supports asynchronous deletion of expired snapshot files.

### LINKED ISSUE #1790: [Feature] Move the behavior of deleting files in SnapshotDeletion to the asynchronous thread pool
### Search before asking

### Motivation

We are trying to run a high-traffic Flink job written to paimon. Paimon is set to 400 buckets, and the checkpoint interval is set to 3 minutes. The `Committer` operator will check the expired snapshot and delete invalid data files when the checkpoint is successful. 

Because a single thread is used to delete a large number of files, it takes a long time, causing backpressure on the `Writer` operator, and the next checkpoint will fail.

```
"Global Committer -> xxx -> Sink: end (1/1)#1" Id=573 TIMED_WAITING on org.apache.hadoop.ipc.Client$Call@4bd116c4
	at java.base@11.0.19/java.lang.Object.wait(Native Method)
	-  waiting on org.apache.hadoop.ipc.Client$Call@4bd116c4
	at app//org.apache.hadoop.ipc.Client.call(Client.java:1661)
	at app//org.apache.hadoop.ipc.Client.call(Client.java:1583)
	at app//org.apache.hadoop.ipc.ProtobufRpcEngine$Invoker.invoke(ProtobufRpcEngine.java:352)
	at app//com.sun.proxy.$Proxy28.delete(Unknown Source)
	at app//org.apache.hadoop.hdfs.protocolPB.ClientNamenodeProtocolTranslatorPB.delete(ClientNamenodeProtocolTranslatorPB.java:631)
	at jdk.internal.reflect.GeneratedMethodAccessor115.invoke(Unknown Source)
	at java.base@11.0.19/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base@11.0.19/java.lang.reflect.Method.invoke(Method.java:566)
	at app//org.apache.hadoop.io.retry.RetryInvocationHandler.invokeMethod(RetryInvocationHandler.java:328)
	at app//org.apache.hadoop.io.retry.RetryInvocationHandler.invoke(RetryInvocationHandler.java:121)
	at app//com.sun.proxy.$Proxy29.delete(Unknown Source)
	at app//org.apache.hadoop.hdfs.DFSClient.delete(DFSClient.java:2522)
	at app//org.apache.hadoop.hdfs.DistributedFileSystem$20.doCall(DistributedFileSystem.java:877)
	at app//org.apache.hadoop.hdfs.DistributedFileSystem$20.doCall(DistributedFileSystem.java:873)
	at app//org.apache.hadoop.fs.FileSystemLinkResolver.resolve(FileSystemLinkResolver.java:81)
	at app//org.apache.hadoop.hdfs.DistributedFileSystem.delete(DistributedFileSystem.java:884)
	at org.apache.paimon.fs.hadoop.HadoopFileIO.delete(HadoopFileIO.java:103)
	at org.apache.paimon.fs.FileIO.deleteQuietly(FileIO.java:144)
	at org.apache.paimon.operation.FileStoreExpireImpl.lambda$expireMergeTreeFiles$3(FileStoreExpireImpl.java:290)
	at org.apache.paimon.operation.FileStoreExpireImpl$$Lambda$2738/0x0000000841215440.accept(Unknown Source)
	at java.base@11.0.19/java.util.HashMap.forEach(HashMap.java:1337)
	at org.apache.paimon.operation.FileStoreExpireImpl.expireMergeTreeFiles(FileStoreExpireImpl.java:287)
	at org.apache.paimon.operation.FileStoreExpireImpl.expireMergeTreeFiles(FileStoreExpireImpl.java:256)
	at org.apache.paimon.operation.FileStoreExpireImpl.expireUntil(FileStoreExpireImpl.java:185)
	at org.apache.paimon.operation.FileStoreExpireImpl.expire(FileStoreExpireImpl.java:139)
	at org.apache.paimon.table.sink.TableCommitImpl.expire(TableCommitImpl.java:141)
	at org.apache.paimon.table.sink.TableCommitImpl.commitMultiple(TableCommitImpl.java:118)
	at org.apache.paimon.flink.sink.StoreCommitter.commit(StoreCommitter.java:76)
	at org.apache.paimon.flink.sink.CommitterOperator.commitUpToCheckpoint(CommitterOperator.java:160)
	at org.apache.paimon.flink.sink.CommitterOperator.notifyCheckpointComplete(CommitterOperator.java:154)
```

### Solution

Since the deletion of these data files does not affect reading, we can use an asynchronous thread pool to delete files and provide an Action to clean up zombie files.

### Anything else?

_No response_

liming30:
### Purpose

Linked issue: close #1790 

[core] supports asynchronous deletion of expired snapshot files. 

I copied 2 files from `flink` repository to help implement the test code: `ManuallyTriggeredScheduledExecutorService`、`ScheduledTask`.

Added 3 configuration options: 

- `snapshot.expire.execution-mode`: decide whether to expire asynchronously. If snapshots are generated faster than deleted, this may cause a backlog of snapshots.
- `snapshot.expire.io.threadpool.size`: threadpool size for deleting files concurrently.
- `snapshot.expire.limit`: The maximum number of snapshots allowed to expire at a time.

### Tests

Added the following unit tests to verify correctness.

- `org.apache.paimon.table.FileStoreTableTestBase#testAsyncExpireExecutionMode`
- `org.apache.paimon.table.FileStoreTableTestBase#testExpireWithLimit`

### API and Format

No
### Documentation

No


paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -194,6 +194,19 @@ public class CoreOptions implements Serializable {
                     .defaultValue(Duration.ofHours(1))
                     .withDescription("The maximum time of completed snapshots to retain.");
 
+    public static final ConfigOption<SnapshotExpireDeleteMode> SNAPSHOT_EXPIRE_MODE =
```
JingsongLi, line 197: Maybe we don't need this.

The concurrent deletion mode can be the default mode, which appears harmless.

liming30: OK, I've removed this config option.

paimon-core/src/main/java/org/apache/paimon/operation/FileStoreExpireImpl.java
```diff
@@ -179,31 +191,57 @@ public void expireUntil(long earliestId, long endExclusiveId) {
...
-            }
-
-            Snapshot snapshot = snapshotManager.snapshot(id);
-            snapshotDeletion.cleanUnusedManifests(snapshot, skippingSet);
-
-            // delete snapshot last
-            snapshotManager.fileIO().deleteQuietly(snapshotManager.snapshotPath(id));
-        }
+        final long startId = beginInclusiveId, endId = endExclusiveId;
+        cleanEmptyDirFuture.whenCompleteAsync(
```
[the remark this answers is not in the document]
liming30: I modified it to wait for the previous `future` to complete before continuing in the next `expire`.  Please help review it when you have time.

siyangzeng:
@liming30 I am very glad that you are also dealing with this problem. Can you consider limiting the amount of cleaning on this basis to avoid too much cleaning at one time and make the whole process smoother?
Consider incorporating this strategy into this  https://github.com/apache/incubator-paimon/pull/1830

liming30:
> @liming30 I am very glad that you are also dealing with this problem. Can you consider limiting the amount of cleaning on this basis to avoid too much cleaning at one time and make the whole process smoother? Consider incorporating this strategy into this #1830

@siyangzeng Thanks for your suggestion. This can also help us control the time-consuming of expire, I will add this feature in the next commit.

paimon-core/src/main/java/org/apache/paimon/AbstractFileStore.java
```diff
@@ -230,4 +234,10 @@ public TagAutoCreation newTagCreationManager() {
         return TagAutoCreation.create(
                 options, snapshotManager(), newTagManager(), newTagDeletion());
     }
+
+    protected ExecutorService createIoExecutor() {
+        return Executors.newFixedThreadPool(
+                options.snapshotExpireIoThreadPoolSize(),
+                new ExecutorThreadFactory("expire-thread"));
```
JingsongLi, line 241: `Thread.currentThread().getName() + "-expire-thread"`

paimon-core/src/main/java/org/apache/paimon/utils/DirectExecutorService.java
```diff
@@ -0,0 +1,268 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+/**
+ * The direct executor service directly executes the runnables and the callables in the calling
+ * thread.
+ *
+ * <p>Copy from Flink.
+ */
+public class DirectExecutorService implements ExecutorService {
```
JingsongLi, line 42: Just use `MoreExecutors.directExecutor()`?

paimon-core/src/main/java/org/apache/paimon/operation/FileDeletionBase.java
```diff
@@ -64,6 +67,7 @@ public abstract class FileDeletionBase {
     protected final ManifestList manifestList;
     protected final IndexFileHandler indexFileHandler;
     protected final Map<BinaryRow, Set<Integer>> deletionBuckets;
+    protected final DirectExecutorService directIoExecutor;
```
JingsongLi, line 70: If you use `MoreExecutors.directExecutor()`, it don't need to be a class field.

JingsongLi:
Overall, I have a question. Will the current deletions be strictly in order? For example, deleting the datafiles first and then the manifestfiles will ensure the reentrancy of the expiration cleanup and avoid leaving zombie files due to failover.

JingsongLi:
Instead of the fragmented generation of async futures like now, can we just introducing some concurrent deletion capabilities, such as introducing an interface called `void deleteFiles(files, executor)`, this method is sync, but will execute deletion concurrently.

liming30:
> Overall, I have a question. Will the current deletions be strictly in order? For example, deleting the datafiles first and then the manifestfiles will ensure the reentrancy of the expiration cleanup and avoid leaving zombie files due to failover.

@JingsongLi It can only guarantee the order of deleting `dataFiles` first, and then deleting `manifests`. But there is no guarantee that zombie files will not be generated, because we always `commitEarliestHint` first to ensure that the expired snapshot is invisible, and we do not wait for the deletion of `dataFiles`/`manifests` to end. In case of failover, these files may become zombie files.

> 2. Maybe: make the entire expire asynchronous and completely decouple from the commit, without blocking the normal process of the commit.

Do you mean we can introduce an `expire` operator?

JingsongLi:
Hi @liming30 
> because we always commitEarliestHint first to ensure that the expired snapshot is invisible

This is not true. Please check again.

> Do you mean we can introduce an expire operator?

No, we just need a async thread in `TableCommitImpl`, can close this thread (single thread executor?) in close.

liming30:
@JingsongLi Thanks for your suggestion, I refactored the implementation of the code: use a single thread executor to expire asynchronously in `TableCommitImpl`, use a threadpool to delete files concurrently in `FileDeletionBase` and wait for the result of the deletion synchronously.  

I kept the `snapshot.expire.execution-mode` configuration option and set it to `sync` by default for two reasons:
1. After adding the ability to delete files concurrently, if the production speed is still lower than the deletion speed, using `sync` as the default option can help users perceive this and make adjustments.
2. A large number of tests assume that `expire` will be completed at the end of the `commit` method. If the default setting is `async`, it may increase the cost of writing test code.

paimon-core/src/test/java/org/apache/paimon/utils/ManuallyTriggeredScheduledExecutorService.java
```diff
@@ -0,0 +1,333 @@
...
+/**
+ * Simple {@link ScheduledExecutorService} implementation for testing purposes. It spawns no
+ * threads, but lets you trigger the execution of tasks manually.
+ *
+ * <p>This class is helpful when implementing tests tasks synchronous and control when they run,
+ * which would otherwise asynchronous and require complex triggers and latches to test.
+ *
+ * <p>Copy from Flink.
+ */
+public class ManuallyTriggeredScheduledExecutorService implements ScheduledExecutorService {
```
JingsongLi, line 47: Do we really need this? This is only for testing.
I think maybe we can just get rid of these two classes.

paimon-core/src/main/java/org/apache/paimon/AbstractFileStore.java
```diff
@@ -60,6 +63,8 @@ public abstract class AbstractFileStore<T> implements FileStore<T> {
 
     @Nullable private final SegmentsCache<String> writeManifestCache;
 
+    private ExecutorService lazyIoExecutor;
```
JingsongLi, line 66: This `IoExecutor` is created in here, but passed to new classes. And closed in new classes.
Its lifetime is weird...

Can we just use `FileUtils.COMMON_IO_FORK_JOIN_POOL`?

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -194,6 +194,26 @@ public class CoreOptions implements Serializable {
                     .defaultValue(Duration.ofHours(1))
                     .withDescription("The maximum time of completed snapshots to retain.");
 
+    public static final ConfigOption<ExpireExecutionMode> SNAPSHOT_EXPIRE_EXECUTION_MODE =
+            key("snapshot.expire.execution-mode")
```
JingsongLi, line 198: I think about this again. Maybe it is not good to provide async mode now. Maybe there will be some thread conflicts.

The performance maybe good enough now, so we don't need to provide async mode.

JingsongLi (commented):
Thanks @liming30 for the contribution, left comments.

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
liming30: For a table with a particularly large number of files, simply increasing the number of threads may not be able to delete expired files in time. In our production scenario, a table is divided into partitions by day, and 10,000+ files are generated each commit, and each partition has about 5 million files. When the partition expires, it is difficult to delete all within the checkpoint interval.

paimon-core/src/main/java/org/apache/paimon/AbstractFileStore.java
liming30: Resolved.

liming30:
Sorry for taking so long to reply. I have addressed the relevant comments. Please review it when you have time.

JingsongLi (commented):
+1

JingsongLi:
You can document this in a separate PR @liming30

PR #1825, opened 2023-08-15
[improve] extract field in value that already contains in key

leaves12138:
The key-value storage (in primary-key table) are redundant.
Now, the format of a row is:   _KEY_a, _KEY_b, _FIELD_SEQUENCE, _ROW_KIND, a, b, c, d, e, f, g
The column "a" and "b" have double restored. A optimization is:
<pre>
convert    _KEY_a, _KEY_b, _FIELD_SEQUENCE, _ROW_KIND, a, b, c, d, e, f, g

to         _KEY_a, _KEY_b, _FIELD_SEQUENCE, _ROW_KIND, c, d, e, f, g
</pre>

While reading, we find the `a`, `b` from key and fill them to value.

This PR is not complete, unit tests and compatibility test will need to be done.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/io/KeyValueFileWriterFactory.java
+++ b/paimon-core/src/main/java/org/apache/paimon/io/KeyValueFileWriterFactory.java
@@ -165,7 +165,7 @@ public class KeyValueFileWriterFactory {
         public KeyValueFileWriterFactory build(
                 BinaryRow partition, int bucket, CoreOptions options) {
             RowType fileRowType = KeyValue.schema(keyType, valueType);
-            RowType storageRowType = RowTypeUtils.toStorageRowType(fileRowType);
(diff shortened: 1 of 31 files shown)


paimon-core/src/main/java/org/apache/paimon/io/KeyValueFileWriterFactory.java
```diff
@@ -164,11 +165,14 @@ private Builder(
         public KeyValueFileWriterFactory build(
                 BinaryRow partition, int bucket, CoreOptions options) {
             RowType fileRowType = KeyValue.schema(keyType, valueType);
+            RowType storageRowType = RowTypeUtils.toStorageRowType(fileRowType);
```
JingsongLi, line 168: Here we need an option, by default, we can use old style row type. Because we need to be as compatible with old readers as possible.

Then wait for 1-2 versions, and switch to the default mode to thin mode.

PR #1828, opened 2023-08-16
[core] AddPartitionCommitCallback should only be created for partitioned table

tsreaper:
### Purpose

Currently when creating `AddPartitionCommitCallback`, we forget to check whether the table is a partitioned table. This PR fixes the issue.

### Tests

* `HiveCatalogITCaseBase#testAddPartitionsToMetastoreForUnpartitionedTable`.

### API and Format

No.

### Documentation

No.


JingsongLi (approved):
+1

PR #1830, opened 2023-08-16
[core] Add expiration cleanup limit

siyangzeng:
### Purpose

【common】add parameters for cleaning limits
【core】limit bulk deletions when snapshots expire

after the consumer group expires, a large number of snapshots will be deleted. In severe cases, it may cause continuous checkpoint timeout and affect the running of tasks. Therefore, we need to limit the number of deleted snapshots to make the whole process smoother.

### Tests

No

### API and Format

No

### Documentation

No


JingsongLi:
Related PR: https://github.com/apache/incubator-paimon/pull/1823

PR #1833, opened 2023-08-17
[hive] After Flink changed the table name when using hive metastore, the modified table could not be found.

### LINKED ISSUE #1831: [Bug] [hive] After modifying the table name when flink uses hive metastore, the modified table cannot be found
### Search before asking

### Paimon version

0.5

### Compute Engine

flink

### Minimal reproduce step

After modifying the table name when flink uses hive metastore, the modified table cannot be found.

5A6300F8-CEF7-4e36-8879-4C3B6E6BF5D3

The modified table can be queried

### Anything else?

_No response_

zhuangchong:
…

### Purpose

Linked issue: close #1831

### Tests

HiveCatalogITCaseBase#testRenameTable

### DIFF
--- a/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
+++ b/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
@@ -358,31 +358,32 @@ public class HiveCatalog extends AbstractCatalog {
 
         try {
             checkIdentifierUpperCase(toTable);
-
-            // Rename the file system's table directory. Maintain consistency between tables in the
-            // file system and tables in the Hive Metastore.
-            Path fromPath = getDataTableLocation(fromTable);
-            Path toPath = getDataTableLocation(toTable);
-            try {
-                fileIO.rename(fromPath, toPath);
-            } catch (IOException e) {
-                throw new RuntimeException(
-                        "Failed to rename changes of table "
-                                + toTable.getFullName()
-                                + " to underlying files.",
-                        e);
-            }
-
             String fromDB = fromTable.getDatabaseName();
             String fromTableName = fromTable.getObjectName();
             Table table = client.getTable(fromDB, fromTableName);
-
-            // update location
-            locationHelper.specifyTableLocation(table, toPath.toString());
-
             table.setDbName(toTable.getDatabaseName());
             table.setTableName(toTable.getObjectName());
             client.alter_table(fromDB, fromTableName, table);
+
+            Path fromPath = getDataTableLocation(fromTable);
+            if (new SchemaManager(fileIO, fromPath).listAllIds().size() > 0) {
+                // Rename the file system's table directory. Maintain consistency between tables in
+                // the
+                // file system and tables in the Hive Metastore.
+                Path toPath = getDataTableLocation(toTable);
+                try {
+                    fileIO.rename(fromPath, toPath);
+                } catch (IOException e) {
+                    throw new RuntimeException(
+                            "Failed to rename changes of table "
+                                    + toTable.getFullName()
+                                    + " to underlying files.",
+                            e);
+                }
+
+                // update location
+                locationHelper.specifyTableLocation(table, toPath.toString());
+            }
         } catch (TException e) {
             throw new RuntimeException("Failed to rename table " + fromTable.getFullName(), e);
         }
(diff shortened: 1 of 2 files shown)


paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
```diff
@@ -358,9 +358,28 @@ public void renameTable(Identifier fromTable, Identifier toTable, boolean ignore
 
         try {
             checkIdentifierUpperCase(toTable);
+
+            // Rename the file system's table directory. Maintain consistency between tables in the
+            // file system and tables in the Hive Metastore.
+            Path fromPath = getDataTableLocation(fromTable);
+            Path toPath = getDataTableLocation(toTable);
+            try {
+                fileIO.rename(fromPath, toPath);
```
JingsongLi, line 367: Maybe check `fromPath` existence first? Maybe hive will do this renaming before.

JingsongLi: https://github.com/apache/incubator-paimon/pull/471
See original PR.

zhuangchong: The version of hive used by the user is 3.1.3. After the rename operation, the location of the new table has not changed, and it points to the original path.

465A537E-183E-4543-8F47-061B0E581173

B79371E6-AAB0-4fbb-8AD6-555E8818C8C3

JingsongLi: You can see discussion in https://github.com/apache/incubator-paimon/pull/471 .
Some Hive version maybe rename the directory to new. So we can add this check first. 
Or can you confirm all Hive versions do not rename the directory?

JingsongLi: We are fixing this bug, so I know there is the problem.

zhangjun0x01:
I test again in hive3.1.3, it is correct. that is very strange

```
Flink SQL> CREATE CATALOG paimon_hive_catalog WITH (
>     'type' = 'paimon',
>     'metastore' = 'hive',
>     'uri' = 'thrift://localhost:9083',
>     'warehouse' = 'hdfs://localhost/user/hive/warehouse'
> );
> 

Flink SQL> use catalog paimon_hive_catalog;


Flink SQL> create database db6;

Flink SQL> use db6;

Flink SQL> create table t1(id int);

Flink SQL> show tables;
+------------+
| table name |
+------------+
|         t1 |
+------------+
1 row in set

Flink SQL> alter table t1 rename to t2;

Flink SQL> show tables;
+------------+
| table name |
+------------+
|         t2 |
+------------+
1 row in set

Flink SQL> show create table t2;
+------------------------------------------------------------------------------------------------------------------------------------+
|                                                                                                                             result |
+------------------------------------------------------------------------------------------------------------------------------------+
| CREATE TABLE `paimon_hive_catalog`.`db6`.`t2` (
  `id` INT
) WITH (
  'path' = 'hdfs://localhost/user/hive/warehouse/db6.db/t2'
)
 |
+------------------------------------------------------------------------------------------------------------------------------------+
1 row in set

```

zhangjun0x01:
We need to confirm if the user is using object storage. I remember this situation occurred once when I was using s3.

In addition, I think we may need more user environment information to troubleshoot the issue

zhangjun0x01:
I test in s3  with hive catalog  , the table could no be found after rename table 

```

Flink SQL> show create table t1;
+----------------------------------------------------------------------------------------------------------------------------------------------------------------+
|                                                                                                                                                         result |
+----------------------------------------------------------------------------------------------------------------------------------------------------------------+
| CREATE TABLE `paimon_hive_catalog_dev`.`paimon_zhangjun`.`t1` (
  `id` INT
) COMMENT 'aa'
WITH (
  'path' = 's3://xxxxx/paimon/paimon_zhangjun.db/t1'
)
 |
+----------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set

Flink SQL> show tables;
+------------+
| table name |
+------------+
|         t1 |
+------------+
1 row in set

Flink SQL> alter table t1 rename to t2;

Flink SQL> show tables;
Empty set

Flink SQL> 


```

zhuangchong:
Thanks @zhangjun0x01 @JingsongLi 

I suggest that at the code layer, first execute the hive client alert_table method to rename, then check toPath, if it does not exist, call fileIO rename, and update hive location. What do you think? @JingsongLi @zhangjun0x01

zhangjun0x01:
> Thanks @zhangjun0x01 @JingsongLi
> 
> I suggest that at the code layer, first execute the hive client alert_table method to rename, then check fromPath, if it exists, call fileIO rename, and update hive location. What do you think? @JingsongLi @zhangjun0x01

I think it is ok, add a check to ensure that the location is correctly renamed

JingsongLi (approved):
+1 Thanks @zhuangchong

PR #1835, opened 2023-08-17
[build] Try to minimize paimon-common jar

JingsongLi:
### Purpose

Now, paimon-common.jar is too big, about 26 MB.
We should minimize it, actually, the root reason is `fastutil`, it is very big.

I am trying to add minimizeJar to shade plugin. After, the jar is 3 MB.


yuzelin (approved):
+1

PR #1836, opened 2023-08-17
[Spark] Spark supports PARTITION MANAGEMENT

choucmei:
### Purpose

Linked feature: close #[1584](https://github.com/apache/incubator-paimon/issues/1584)

### Tests

org.apache.paimon.spark.SparkReadITCase#testShowPartition
org.apache.paimon.spark.SparkReadITCase#testDropPartition

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/SparkTable.java
+++ b/paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/SparkTable.java
@@ -53,7 +53,8 @@ public class SparkTable
         implements org.apache.spark.sql.connector.catalog.Table,
                 SupportsRead,
                 SupportsWrite,
-                SupportsDelete {
+                SupportsDelete,
+                PaimonPartitionManagement {
 
     private final Table table;
 
--- a/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonPartitionManagement.scala
+++ b/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonPartitionManagement.scala
@@ -1,0 +1,129 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.paimon.spark
+
+import org.apache.paimon.data.BinaryRow
+import org.apache.paimon.operation.FileStoreCommit
+import org.apache.paimon.table.{AbstractFileStoreTable, Table}
+import org.apache.paimon.table.sink.BatchWriteBuilder
+import org.apache.paimon.table.source.TableScan
+import org.apache.paimon.types.RowType
+import org.apache.paimon.utils.{FileStorePathFactory, RowDataPartitionComputer}
+
+import org.apache.spark.sql.Row
+import org.apache.spark.sql.catalyst.{CatalystTypeConverters, InternalRow}
+import org.apache.spark.sql.connector.catalog.SupportsPartitionManagement
+import org.apache.spark.sql.types.StructType
+
+import java.util
+import java.util.{Collections, UUID}
+
+import scala.collection.JavaConverters._
+
+trait PaimonPartitionManagement extends SupportsPartitionManagement {
+  self: SparkTable =>
+
+  def partitionKeys() = getTable.partitionKeys
+
+  def tableRowType() = new RowType(
+    getTable.rowType.getFields.asScala
+      .filter(dataFiled => partitionKeys.contains(dataFiled.name()))
+      .asJava)
+
+  override def partitionSchema(): StructType = {
+    SparkTypeUtils.fromPaimonRowType(tableRowType)
+  }
+
+  override def dropPartition(internalRow: InternalRow): Boolean = {
+    // convert internalRow to row
+    val row: Row = CatalystTypeConverters
+      .createToScalaConverter(partitionSchema())
+      .apply(internalRow)
+      .asInstanceOf[Row]
+    val rowDataPartitionComputer = new RowDataPartitionComputer(
+      FileStorePathFactory.PARTITION_DEFAULT_NAME.defaultValue,
+      tableRowType,
+      partitionKeys.asScala.toArray)
+    val partitionMap = rowDataPartitionComputer.generatePartValues(new SparkRow(tableRowType, row))
+    getTable match {
+      case table: AbstractFileStoreTable =>
+        val commit: FileStoreCommit = table.store.newCommit(UUID.randomUUID.toString)
+        commit.dropPartitions(
+          Collections.singletonList(partitionMap),
+          BatchWriteBuilder.COMMIT_IDENTIFIER)
+      case _ =>
+        throw new UnsupportedOperationException(
+          "Only AbstractFileStoreTable supports drop partitions.")
+    }
+    true
+  }
+
+  override def replacePartitionMetadata(
+      ident: InternalRow,
+      properties: util.Map[String, String]): Unit = {
+    throw new UnsupportedOperationException("Replace partition is not supported")
+  }
+
+  override def loadPartitionMetadata(ident: InternalRow): util.Map[String, String] = {
+    throw new UnsupportedOperationException("Load partition is not supported")
+  }
+
+  override def listPartitionIdentifiers(
+      partitionCols: Array[String],
+      internalRow: InternalRow): Array[InternalRow] = {
+    assert(
+      partitionCols.length == internalRow.numFields,
+      s"Number of partition names (${partitionCols.length}) must be equal to " +
+        s"the number of partition values (${internalRow.numFields})."
+    )
+    val schema: StructType = partitionSchema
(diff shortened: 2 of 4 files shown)


paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/SparkTable.java
```diff
@@ -137,4 +153,97 @@ public Map<String, String> properties() {
...
+    public void createPartition(InternalRow internalRow, Map<String, String> map)
+            throws PartitionsAlreadyExistException, UnsupportedOperationException {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public boolean dropPartition(InternalRow internalRow) {
+        StructType structType = partitionSchema();
+        StructField[] fields = structType.fields();
+        HashMap<String, String> partitions = new HashMap<>();
```
JingsongLi, line 178: Can you use `RowDataPartitionComputer` here?

And we can convert internalRow to paimon `SparkRow`.

choucmei: Do you mean it like this?
```
        List<String> partitionKeys = table.partitionKeys();
        RowType rowType =
                new RowType(
                        table.rowType().getFields().stream()
                                .filter(dataField -> partitionKeys.contains(dataField.name()))
                                .collect(Collectors.toList()));
        StructType structType = SparkTypeUtils.fromPaimonRowType(rowType);
        String[] partitionCols = table.partitionKeys().toArray(new String[partitionKeys.size()]);
        RowDataPartitionComputer rowDataPartitionComputer = new RowDataPartitionComputer(FileStorePathFactory.PARTITION_DEFAULT_NAME.defaultValue(), rowType, partitionCols);
        LinkedHashMap<String, String> partitions = rowDataPartitionComputer.generatePartValues(new SparkRow(rowType, Row.fromSeq(internalRow.toSeq(structType))));
```
`internalRow` need convert to `Row` ,use `SparkRow` to wrapper this

YannByron:
@choucmei  Can we separate the `SupportsPartitionManagement` part currently in `SparkTable` into an individual class, like `PaimonPartitionManagement`? And another small ask, can you use scala to implement this ?

choucmei:
> @choucmei Can we separate the `SupportsPartitionManagement` part currently in `SparkTable` into an individual class, like `PaimonPartitionManagement`? And another small ask, can you use scala to implement this ?

`PaimonPartitionManagement` implments `SupportsPartitionManagement`,then `SparkTable` extend `PaimonPartitionManagement`   . Right?

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonPartitionManagement.scala
```diff
@@ -0,0 +1,125 @@
...
+import scala.collection.JavaConverters._
+
+abstract class PaimonPartitionManagement(table: Table) extends SupportsPartitionManagement {
+
+  override def partitionSchema(): StructType = {
+    val partitionKeys = table.partitionKeys
+    val rowType = new RowType(
+      table.rowType.getFields.asScala
+        .filter(dataFiled => partitionKeys.contains(dataFiled.name()))
+        .asJava)
```
YannByron, line 45: Better to extract  a private method here that return the partition RowType, and this method can reuse in `partitionSchema` and  `dropPartition` .

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonPartitionManagement.scala
```diff
@@ -0,0 +1,125 @@
...
+    SparkTypeUtils.fromPaimonRowType(rowType)
+  }
+
+  override def dropPartition(internalRow: InternalRow): Boolean = {
```
YannByron, line 49: we need to check whether this table (eg: SystemTable) can support `dropPartition`, maybe like this: 
table match {
    case t: FileStoreTable => 
        // ....
    case _ => throw ...
}

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonPartitionManagement.scala
```diff
@@ -0,0 +1,125 @@
...
+    val row: Row = CatalystTypeConverters
+      .createToScalaConverter(partitionSchema())
+      .apply(internalRow)
+      .asInstanceOf[Row]
+    val rowDataPartitionComputer = new RowDataPartitionComputer(
+      FileStorePathFactory.PARTITION_DEFAULT_NAME.defaultValue,
+      rowType,
+      partitionKeys.asScala.toArray)
+    val partitionMap = rowDataPartitionComputer.generatePartValues(new SparkRow(rowType, row))
+    val identifier: Long = BatchWriteBuilder.COMMIT_IDENTIFIER
```
YannByron, line 65: Internal variables do not need to declare its type identifier.

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonPartitionManagement.scala
```diff
@@ -0,0 +1,125 @@
...
+    val commit: FileStoreCommit =
+      (table.asInstanceOf[AbstractFileStoreTable]).store.newCommit(UUID.randomUUID.toString)
```
YannByron, line 67: `FileStoreCommit newCommit(String commitUser);`

here why use UUID to fake the commit user?

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonPartitionManagement.scala
```diff
@@ -0,0 +1,125 @@
...
+    )
+    val schema: StructType = partitionSchema
+    assert(
+      partitionCols.forall(fieldName => schema.fieldNames.contains(fieldName)),
+      s"Some partition names ${partitionCols.mkString("[", ", ", "]")} don't belong to " +
+        s"the partition schema '${schema.sql}'."
+    )
+
+    val tableScan: TableScan = table.newReadBuilder.newScan
+    val binaryRows: List[BinaryRow] = tableScan.listPartitions.asScala.toList
```
YannByron, line 98: the current implement is ok. but maybe an optimization can be considered, that is we can convert `partitionCols` and `internalRow` to the Paimon `Predicate`, and `newReadBuilder.withFilter(...).newScan` to get the result directly. That's just a suggestion, you can stay this way.

paimon-spark/paimon-spark-common/src/test/java/org/apache/paimon/spark/SparkReadITCase.java
```diff
@@ -704,4 +704,101 @@ public void testCreateNestedField() {
                                 "a INT",
                                 "b STRUCT<b1: STRUCT<b11: INT, b12: INT>, b2: BIGINT>"));
     }
+
+    @Test
+    public void testShowPartition() {
```
YannByron, line 709: sorry that forget to remind you in the first review, that we’d better to move this to `test/scala/org/apache/spark/sql/PaimonMangermentTest.scala`.

YannByron: after this, we can use `T` to replace `partition_table_01` or `drop_partition_table_01` table name, that will be dropped automatically when this UT finishes.

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonPartitionManagement.scala
choucmei: because i saw `((AbstractFileStoreTable) table).store().newCommit(UUID.randomUUID().toString());` in other modules.

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonPartitionManagement.scala
choucmei:
```
      val predicate = PredicateBuilder.partition(partitionMap, tableRowType)
      getTable.newReadBuilder.withFilter(predicate).newScan()
      val binaryRows = tableScan.listPartitions.asScala.toList
```
I tried using partition filters before, but tableScan. listPartitions is empty.

docs/content/maintenance/write-performance.md
```diff
@@ -28,7 +28,9 @@ under the License.
 
 Paimon's write performance is closely related to checkpoint, so if you need greater write throughput:
 
-1. Increase the checkpoint interval, or just use batch mode.
+1. Flink Configuration (`'flink-conf.yaml'` or `SET` in SQL): Increase the checkpoint interval
+   (`'execution.checkpointing.interval'`), increase max concurrent checkpoints to 3
+   (`'execution.checkpointing.max-concurrent-checkpoints'`), or just use batch mode.
```
YannByron, line 33: ??

paimon-spark/paimon-spark-common/src/test/scala/org/apache/paimon/spark/sql/PaimonPartitionManagementTest.scala
```diff
@@ -0,0 +1,200 @@
...
+package org.apache.paimon.spark.sql
+
+import org.apache.paimon.WriteMode.CHANGE_LOG
+import org.apache.paimon.spark.PaimonSparkTestBase
+
+import org.apache.spark.sql.{AnalysisException, Row}
+
+class PaimonPartitionManagementTest extends PaimonSparkTestBase {
+
+  writeModes.foreach {
```
YannByron, line 27: I think we only need two tests: one for non-partitioned table, one for partitioned table. Each of test should convert the `drop partition` and `show partitions` cases. maybe it'll more clear.

YannByron:
@choucmei if you don't mind, can you apply this patch.
[0001-followup-optimize-PaimonPartitionManagementTest.patch](https://github.com/apache/incubator-paimon/files/12415083/0001-followup-optimize-PaimonPartitionManagementTest.patch)

choucmei:
> 
Thanks for the advice.

YannByron (approved):

JingsongLi (approved):
+1

PR #1837, opened 2023-08-17
[hive]hive create table should force primary key not null

Alibaba-HZY:
hive create table should force primary key not null

### Tests

HiveCatalogITCaseBase

### API and Format

no

### Documentation

no

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/schema/SchemaManager.java
+++ b/paimon-core/src/main/java/org/apache/paimon/schema/SchemaManager.java
@@ -151,7 +151,8 @@ public class SchemaManager implements Serializable {
                     fields.stream()
                             .map(
                                     field -> {
-                                        if (finalPrimaryKeys.contains(field.name())) {
+                                        if (finalPrimaryKeys.contains(field.name())
+                                                && field.type().isNullable()) {
                                             return new DataField(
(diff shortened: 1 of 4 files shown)


paimon-core/src/main/java/org/apache/paimon/schema/SchemaManager.java
```diff
@@ -146,7 +146,23 @@ public TableSchema createTable(Schema schema) throws Exception {
                 }
                 options.remove(CoreOptions.PRIMARY_KEY.key());
             }
-
+            List<String> finalPrimaryKeys = primaryKeys;
+            fields =
```
JingsongLi, line 150: Actually, I don't get why.

`Schema.normalizeFields` will convert pk fields to not-null.

Alibaba-HZY: when hive create table primary key  is created by 
image
is not from Schema.normalizeFields

JingsongLi: I see, maybe we should move the logical from `SchemaManager` to `Schema`.

JingsongLi (approved):
+1

PR #1840, opened 2023-08-17
[core] Remove append-only.assert-disorder

JingsongLi:
### Purpose

There could overlap in append files, if the users have multiple write jobs.

So we can not check by default.


yuzelin (approved):
+1

PR #1841, opened 2023-08-17
[flink] only sync user table in the sync action

zhangjun0x01:
### Purpose

add table type to avoid sync `VIEW`,`SYSTEM DATABASE` like `mysql`


zhuangchong:

PR #1842, opened 2023-08-17
[spark] Supports Spark Streaming Source

YannByron:
### Purpose

Linked issue: close #xxx

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/sources/PaimonMicroBatchStream.scala
+++ b/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/sources/PaimonMicroBatchStream.scala
@@ -1,0 +1,81 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.paimon.spark.sources
+
+import org.apache.paimon.spark.{SparkInputPartition, SparkReaderFactory}
+import org.apache.paimon.table.FileStoreTable
+import org.apache.paimon.table.source.ReadBuilder
+
+import org.apache.spark.internal.Logging
+import org.apache.spark.sql.connector.read.{InputPartition, PartitionReaderFactory}
+import org.apache.spark.sql.connector.read.streaming.{MicroBatchStream, Offset}
+
+class PaimonMicroBatchStream(
+    originTable: FileStoreTable,
+    readBuilder: ReadBuilder,
+    checkpointLocation: String)
+  extends MicroBatchStream
+  with StreamHelper
+  with Logging {
+
+  private var committedOffset: Option[PaimonSourceOffset] = None
+
+  lazy val initOffset: PaimonSourceOffset = PaimonSourceOffset(getStartingContext)
+
+  override def latestOffset(): Offset = {
+    getLatestOffset
+  }
+
+  override def planInputPartitions(start: Offset, end: Offset): Array[InputPartition] = {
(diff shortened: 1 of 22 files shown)


YannByron:
A following pr will to support the admission control to limit he rate of data ingested.

JingsongLi:
Hi @YannByron , thanks for the contribution.

I don't get why you need to introduce a new `StreamingScanner`. Can we just use `Table.newReadBuilder.newStreamScan`?
What is the requirement in Spark streaming?

JingsongLi:
> I try to refactor `StreamTableScan` for spark streaming read: #1905

Merged.

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/sources/PaimonMicroBatchStream.scala
```diff
@@ -0,0 +1,182 @@
...
+
+    // choose the bigger one between the earliest snapshotId and the provided snapshotId if exists as the init snapshotId
+    val earliestSnapshotId = Try(table.snapshotManager().earliestSnapshotId())
+      .map(_.longValue())
+      .getOrElse(Long.MinValue)
+
+    var scanMode = FROM_SNAPSHOT
+    var startSnapshotId = currentSnapshot.map(_.id()).getOrElse(1L)
+    var withStartSnapshot = false
+    options.get(SCAN_MODE) match {
```
JingsongLi, line 113: Can we ignore `SCAN_MODE` here?
We may support more scan modes in the future, but if we decouple here, it will be difficult to maintain.

When initializing for the first time, can we have a `null` offset so that we don't have to forcibly generate a startOffset.

YannByron: Sorry that I can't understand why we should have a null offset instead of an initialized offset.
Spark will call  `initialOffset` to get the starting offset which is parsed by `SCAN_MODE` and other related options.

YannByron: And i don't think we can support other scan modes without changing any codes.

JingsongLi: Logically speaking, Spark does not need to perceive specific scan modes, for example, it only generates offsets at the beginning: `(startOffset: Nil, endOffset: $last_snapshot)`.
Then, during scan, if startOffset is Nil, directly call StreamTableScan.scan.
At this point, obtain the snapshotId to which the plan belongs, set offset to: snapshotId + splitNumber + isInitial: true.

This process does not require Spark to perceive scan modes.

YannByron: Understood. I still think we need to initialize `initialOffset` by parsing `ScanMode` and other related configs. Cause:
1. `initialOffset` will return an accurate offset, not just `null`, due to the definition of the method.
2. As `startOffset` in the first batch, `initialOffset` will also be used in many calls, like `Offset latestOffset(Offset startOffset, ReadLimit limit)` after supporting read limit.
3. Cause more complicated logic, just to avoid perceive scan mode. maybe some unexpected errors will occur.

JingsongLi: I still adhere to my viewpoint for the following reasons:

I think it just depends on our definition, and the initialization state is also an initialOffset. I don't necessarily mean to use null to represent it, it depends on the abstract space provided by the Spark interface.

When using StreamTableScan, we don't have any requirements to couple scan modes in Spark code, which is strange and will make it increasingly troublesome later on.
When we added a scan mode (which is common and has occurred multiple times in the past), we found that we need to modify Spark's code, while Flink's code does not need to be modified, which is strange.

YannByron: Your point is avoiding to perceive scan mode just for spark interface, while my point is that `initialOffset` should be correct and meaningful. I have an idea to meet these: let `StreamTableScan` return this starting snapshot id and whether it'll scan the full snapshot.

JingsongLi: We can define initialOffset is something like `-unbounded`, because Paimon store all data for snapshot scan. Is this possible?

YannByron: related to https://github.com/apache/incubator-paimon/pull/1955.

JingsongLi (approved):
+1

PR #1844, opened 2023-08-18
[flink][mysql-cdc] Map MySQL BINARY to Paimon VARBINARY and fix parse of MySQL BIT(N) value

yuzelin:
### Purpose

1. Binary value is passed as bytes, so we should map the MySQL Binary to byte type. I choose VARBINARY because it can retain the length information.

2. Current parse of BIT(N) is incorrect. The value is of little endian form, and shouldn't be stored by Java String.
Fix:
1. convert bytes to normal form;
2. encode the value to String using Base64 

### Tests

`MySqlCdcDataTypeModeITCase#testReadAllTypes`
`MySqlSyncTableActionITCase#testAllTypes`

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/cdc/CdcRecord.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/cdc/CdcRecord.java
@@ -25,6 +25,7 @@ import java.io.Serializable;
 import java.util.Collections;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 
 /** A data change message from the CDC source. */
 @Experimental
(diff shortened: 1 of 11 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/cdc/CdcRecord.java
```diff
@@ -35,10 +36,16 @@ public class CdcRecord implements Serializable {
     private RowKind kind;
 
     private final Map<String, String> fields;
+    private final Set<String> encodedFields;
```
JingsongLi, line 39: -1 for here, we should not modify common data structure.

JingsongLi (approved):
+1

PR #1845, opened 2023-08-18
[flink] Add local merge operator for Flink

tsreaper:
### Purpose

Local merge aims to resolve data skew on primary keys. If your job suffers from primary key data skew (for example, you want to count the number of views for each pages in a website, and some particular pages are very popular among the users), you can use local merge so that input records will be buffered and merged before they're shuffled by bucket and written into sink. This is particularly useful when the same primary key is updated frequently between snapshots.

This PR implements local merge operator for Flink.

### Tests

* `ChangelogWithKeyFileStoreTableITCase` now randomly runs with local merge.

### API and Format

No.

### Documentation


JingsongLi (approved):
+1

PR #1851, opened 2023-08-20
Doc/add dictionary option

### LINKED ISSUE #1375: [Feature] Optimize memory for big column

wg1026688210:
### Documentation
add the usage of dictionary option in writer-perform page to guide user when and how to use dictionary option in what scenarios

close #1375


JingsongLi (approved):
+1

PR #1857, opened 2023-08-21
[flink] Introduce FlinkGenericCatalog

JingsongLi:
### Purpose

Using FlinkGenericCatalog, you need to use Hive metastore. Then, you can use all the tables from Paimon, Hive, and
Flink Generic Tables (Kafka and other tables)!

In this mode, you should use 'connector' option for creating tables.

```sql
CREATE CATALOG my_catalog WITH (
    'type'='paimon-generic',
    'hive-conf-dir'='...',
    'hadoop-conf-dir'='...'
);

USE CATALOG my_catalog;

-- create a word count table
CREATE TABLE word_count (
    word STRING PRIMARY KEY NOT ENFORCED,
    cnt BIGINT
) WITH (
    'connector'='paimon'
);
```

### DIFF
--- a/docs/content/engines/flink.md
+++ b/docs/content/engines/flink.md
@@ -125,6 +125,10 @@ You can now start Flink SQL client to execute SQL scripts.
 
 **Step 5: Create a Catalog and a Table**
 
+{{< tabs "Create Flink Catalog" >}}
+
+{{< tab "Catalog" >}}
+
 ```sql
 -- if you're trying out Paimon in a distributed environment,
 -- the warehouse path should be set to a shared file system, such as HDFS or OSS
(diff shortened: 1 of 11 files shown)


hdygxsj:
Do we have to use hive metastore? Why not consider persisting to paimon's own file system

yuzelin (approved):
+1

SteNicholas (approved):
LGTM.

zhuangchong (approved):
+1

JingsongLi:
> Do we have to use hive metastore? Why not consider persisting to paimon's own file system

Hi @hdygxsj , I replied you in https://github.com/apache/incubator-paimon/issues/1804 , please take a look~

PR #1868, opened 2023-08-22
[hotfix] Fix that MySQL sync-database action parses table prefix incorrectly

yuzelin:


JingsongLi (approved):
I merged this first, but this fix needs testing.

PR #1869, opened 2023-08-22
[hotfix] Fix savepoint not auto create tag

wxplovecc:
### Purpose

Savepoint not trigger `notifyCheckpointComplete` method

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/AutoTagForSavepointCommitterOperator.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/AutoTagForSavepointCommitterOperator.java
@@ -127,8 +127,10 @@ public class AutoTagForSavepointCommitterOperator<CommitT, GlobalCommitT>
     @Override
     public void notifyCheckpointComplete(long checkpointId) throws Exception {
         commitOperator.notifyCheckpointComplete(checkpointId);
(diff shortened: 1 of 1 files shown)


wxplovecc:
cc @liming30

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/AutoTagForSavepointCommitterOperator.java
```diff
@@ -127,8 +127,10 @@ public OperatorSnapshotFutures snapshotState(
     @Override
     public void notifyCheckpointComplete(long checkpointId) throws Exception {
         commitOperator.notifyCheckpointComplete(checkpointId);
-        if (identifiersForTags.remove(checkpointId)) {
-            createTagForIdentifiers(Collections.singletonList(checkpointId));
+        // savepoint do not trigger notifyCheckpointComplete method
+        long savepointId = checkpointId - 1;
```
liming30, line 131: We may need to make tags for all previous successful savepoints that are less than `checkpointId`, instead of just `checkpointId - 1 `(considering the case where multiple savepoints are triggered consecutively).  
This feature may work fine under `stop-with-savepoint`.

liming30:
@wxplovecc Thanks for your contribution, left one comment.

wxplovecc:
> 

Get it close this

PR #1872, opened 2023-08-23
[Refactor] Add Kafka cdc computed column test and reduce unnecessary e2e tests

yuzelin:
### Purpose

1. Kafka computed column test is missing.
2. tinyint1 conversion test shouldn't be in test base, because it will increase many unnecessary tests.


JingsongLi (approved):
+1

PR #1877, opened 2023-08-23
[flink] add table comment for create table in sync action

zhangjun0x01:
### Purpose

in flink sync action , sync the table comment when create new table.

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/cdc/mysql/MySqlActionUtils.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/cdc/mysql/MySqlActionUtils.java
@@ -27,6 +27,7 @@ import org.apache.paimon.flink.action.cdc.mysql.schema.MySqlSchemasInfo;
 import org.apache.paimon.flink.action.cdc.mysql.schema.MySqlTableInfo;
 import org.apache.paimon.schema.Schema;
 import org.apache.paimon.types.DataType;
+import org.apache.paimon.utils.StringUtils;
 
 import com.ververica.cdc.connectors.mysql.source.MySqlSource;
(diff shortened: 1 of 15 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/cdc/mysql/MySqlActionUtils.java
```diff
@@ -79,13 +79,13 @@ static Connection getConnection(Configuration mySqlConfig, boolean tinyint1NotBo
             throws Exception {
         String url =
                 String.format(
-                        "jdbc:mysql://%s:%d%s",
+                        "jdbc:mysql://%s:%d?useInformationSchema=true%s",
```
JingsongLi, line 82: Should we use an option to open this feature?
I don't know how much impact it has in performance.

zhangjun0x01: the param in default `true` in  mysql 8.0.3 ([doc](https://dev.mysql.com/doc/connector-j/8.1/en/connector-j-connp-props-metadata.html)) ,but it is default `false` in mysql    5.7,  if we add an option and when we set it to `false` ,do    we should set it to `false` explicitly in mysql8 ?

JingsongLi: Default auto?

JingsongLi: Or default we can just disable table comments.

zhuangchong:

PR #1881, opened 2023-08-24
[fix] clear sort buffer after zorder sort complete

leaves12138:
clear buffer after sort


JingsongLi:
Please add a test for `SortOperator`.

JingsongLi (approved):
+1

PR #1905, opened 2023-08-28
[core] Refactor StreamTableScan interface to spark streaming read

JingsongLi:
### Purpose

At present, the interface exposed by StreamTableScan is only suitable for Flink, and Spark does not save the returned splits. It needs to read the current snapshot again during failover, but currently only returns nextSnapshotId.

This PR includes the current snapshotId in the returned plan and returns ScanMode to resolve the full stage restore.


yuzelin (approved):
+1

PR #1910, opened 2023-08-29
[core] Support incremental between scan mode

wxplovecc:
### Purpose

Support incremental between scan mode , default is `DELTA`, `CHANGELOG` mode used in some batch write scene
with `changelog-producer.row-deduplicate` on , mult write the same table users can scan only changelog for the incremental 
changelog

### Tests

IncrementalStartingScannerTest.testScan


paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -1589,6 +1601,50 @@ public static StreamingCompactionType fromValue(String value) {
         }
     }
 
+    /** Specifies this scan type for incremental scan . */
+    public enum IncrementalBetweenScanMode implements DescribedEnum {
+        DELATA("delta", "Scan newly changed files of a snapshot."),
```
JingsongLi, line 1606: DELTA instead of DELATA

paimon-core/src/main/java/org/apache/paimon/table/source/snapshot/IncrementalStartingScanner.java
```diff
@@ -111,4 +125,13 @@ private List<DataSplit> readDeltaSplits(SnapshotReader reader, Snapshot s) {
         }
         return (List) reader.withSnapshot(s).withMode(ScanMode.DELTA).read().splits();
     }
+
+    @SuppressWarnings({"unchecked", "rawtypes"})
+    private List<DataSplit> readChangeLogSplits(SnapshotReader reader, Snapshot s) {
+        if (s.commitKind() != CommitKind.COMPACT) {
```
JingsongLi, line 131: Maybe we should not ignore APPEND for changelog.

JingsongLi (changes requested):

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
wxplovecc: Done

paimon-core/src/main/java/org/apache/paimon/table/source/snapshot/IncrementalStartingScanner.java
wxplovecc: Yes you are right, APPEND is also need

JingsongLi (approved):
+1

PR #1913, opened 2023-08-29
[flink] fix cannot deserialize value

MonsterChenzhuo:
### Purpose

fix aused by: java.lang.IllegalArgumentException: Cannot deserialize value of type `java.lang.String` from Object value (token `JsonToken.START_OBJECT`)
 at [Source: UNKNOWN; byte offset: #UNKNOWN]


JingsongLi (approved):
+1

PR #1914, opened 2023-08-30
[flink] Fix global dynamic bucket mode

JingsongLi:
### Purpose

1. Introduce ExternalBuffer, records should wait bootstrap.
2. Remove CDC record supports for global dynamic bucket mode. (ExternalBuffer only supports Row)
3. Fix global dynamic bucket mode.


leaves12138:
+1

PR #1919, opened 2023-08-30
[flink] fix exception related to field size after adding a new table

MonsterChenzhuo:
### Purpose

Fix the issue of throwing an exception related to field size after adding a new table.

### Tests

MongoDBSyncDatabaseActionITCase#testDynamicTableCreationInMongoDB()


JingsongLi (changes requested):
Can you add a case to cover this fixing?

MonsterChenzhuo:
> Can you add a case to cover this fixing?

没有问题，整整在

JingsongLi:
@MonsterChenzhuo See `TestCaseInsensitiveCatalogFactory`.

MonsterChenzhuo:
> @MonsterChenzhuo See `TestCaseInsensitiveCatalogFactory`.

@JingsongLi  PTAL,thanks.

JingsongLi (approved):
+1

PR #1920, opened 2023-08-30
[Hive] Set the primary key column to be non-null. Automatically when create table by hive

wg1026688210:
### Purpose
Currently, the primary key of the table created by Hive is nullable, which causes the write tasks to fail. This PR is to set primary key as non-null when create table by hive  to work around this problem.

### Tests
paimon-hive/paimon-hive-connector-common/src/test/java/org/apache/paimon/hive/CreateTableITCase.java#testPrimaryKeyFieldNotNull


wg1026688210:
the latest version of master has sovled it ,so I close  it.

PR #1921, opened 2023-08-30
[flink] Remove Redundant Test Methods

MonsterChenzhuo:
### Purpose

Remove Redundant Test Methods

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/Action.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/Action.java
@@ -18,9 +18,13 @@
 
 package org.apache.paimon.flink.action;
 
+import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
+
 /** Abstract class for Flink actions. */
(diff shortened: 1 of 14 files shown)


MonsterChenzhuo:
@yuzelin PTAL,thanks.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/Action.java
```diff
@@ -18,9 +18,13 @@
...
 
+import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
+
 /** Abstract class for Flink actions. */
 public interface Action {
 
     /** The execution method of the action. */
     void run() throws Exception;
+
+    default void build(StreamExecutionEnvironment env) throws Exception {}
```
yuzelin, line 29: Actions that override this method should be annotated.

MonsterChenzhuo: done.

yuzelin (approved):
+1

PR #1924, opened 2023-08-30
[flink] Mongodb CDC Ingestion  Support computed column

MonsterChenzhuo:
### Purpose

mongodb CDC Ingestion support computed column

### Tests

MongoDBSyncTableActionITCase#testComputedColumn()


JingsongLi (approved):
+1

PR #1925, opened 2023-08-31
[test] fix test problem: libc++abi: pure virtual function called

leaves12138:
Test always fails because some error called: pure virtual function called.

Track it, found the root cause is: "Java: Finalizer" Thread collapse while gc the rocksdb object.

crash stack:
image

rocksdb related issue:
https://github.com/facebook/rocksdb/issues/649


yuzelin (approved):
Hope it is useful.

MonsterChenzhuo:
hi @leaves12138  Could you please explain how this crash stack was obtained?

leaves12138:
> hi @leaves12138 Could you please explain how this crash stack was obtained?

system log on mac

PR #1927, opened 2023-08-31
[core] Fix method references may be null

MonsterChenzhuo:
### Purpose

`this::toRow` the method reference parameter may be null

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/system/CatalogOptionsTable.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/system/CatalogOptionsTable.java
@@ -38,7 +38,6 @@ import org.apache.paimon.utils.ProjectedRow;
 
 import org.apache.paimon.shade.guava30.com.google.common.collect.Iterators;
 
-import java.io.IOException;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Iterator;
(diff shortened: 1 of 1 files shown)


MonsterChenzhuo:
@yuzelin PTAL,thanks.

paimon-core/src/main/java/org/apache/paimon/table/system/CatalogOptionsTable.java
```diff
@@ -167,14 +166,14 @@ public TableRead withIOManager(IOManager ioManager) {
...
-        public RecordReader<InternalRow> createReader(Split split) throws IOException {
+        public RecordReader<InternalRow> createReader(Split split) {
             if (!(split instanceof CatalogOptionsTable.CatalogOptionsSplit)) {
                 throw new IllegalArgumentException("Unsupported split: " + split.getClass());
             }
             Iterator<InternalRow> rows =
                     Iterators.transform(
                             ((CatalogOptionsSplit) split).catalogOptions.entrySet().iterator(),
-                            this::toRow);
+                            entry -> (entry == null) ? null : this.toRow(entry));
```
yuzelin, line 176: I think this is not much necessary.

MonsterChenzhuo: ok

PR #1929, opened 2023-08-31
[test] add wait time for runnable in thread pool

leaves12138:
Add wait time for runnable in thread pool, the last runnable may not be stopped by shutdown.

### DIFF
--- a/paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/sink/AppendOnlyTableCompactionWorkerOperatorTest.java
+++ b/paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/sink/AppendOnlyTableCompactionWorkerOperatorTest.java
@@ -41,7 +41,6 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
 import java.util.stream.Collectors;
 
 /** Tests for {@link AppendOnlyTableCompactionWorkerOperator}. */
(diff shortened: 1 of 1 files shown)


JingsongLi:
Thanks @leaves12138 for your contribution.

But can you explain what is the problem?

> the last runnable may not be stopped by shutdown.

So what?

leaves12138:
> Thanks @leaves12138 for your contribution.
> 
> But can you explain what is the problem?
> 
> > the last runnable may not be stopped by shutdown.
> 
> So what?

When executorPool called shutdown, maybe there is a runnable thread in thread pool, and have not get some place to handle InterruptedException. Then it will still hold the open output stream. The method `TableTestBase.after` will check if all the output streams are closed. It will throw an assert exception.

JingsongLi:
> > Thanks @leaves12138 for your contribution.
> > But can you explain what is the problem?
> > > the last runnable may not be stopped by shutdown.
> > 
> > 
> > So what?
> 
> When executorPool called shutdown, maybe there is a runnable thread in thread pool, and have not get some place to handle InterruptedException. Then it will still hold the open output stream. The method `TableTestBase.after` will check if all the output streams are closed. It will throw an assert exception.

I still cannot get it. Are there some cases fail?

leaves12138:
> > > Thanks @leaves12138 for your contribution.
> > > But can you explain what is the problem?
> > > > the last runnable may not be stopped by shutdown.
> > > 
> > > 
> > > So what?
> > 
> > 
> > When executorPool called shutdown, maybe there is a runnable thread in thread pool, and have not get some place to handle InterruptedException. Then it will still hold the open output stream. The method `TableTestBase.after` will check if all the output streams are closed. It will throw an assert exception.
> 
> I still cannot get it. Are there some cases fail?

occasionally, 
image
This may throws en error, because there is still one thread running in thread pool occupied an output stream or input stream

JingsongLi (approved):
+1

PR #1940, opened 2023-09-04
[flink] Decouple tag Auto Creation logic

MonsterChenzhuo:
### Purpose

 Decouple tag Auto Creation logic

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/tag/TagAutoCreation.java
+++ b/paimon-core/src/main/java/org/apache/paimon/tag/TagAutoCreation.java
@@ -20,57 +20,34 @@ package org.apache.paimon.tag;
 
 import org.apache.paimon.CoreOptions;
 import org.apache.paimon.Snapshot;
-import org.apache.paimon.data.Timestamp;
 import org.apache.paimon.operation.TagDeletion;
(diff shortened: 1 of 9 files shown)


MonsterChenzhuo:
@yuzelin PTAL，thanks.

JingsongLi:
I don't fully understand why we need to decouple tag Auto Creation logic.

JingsongLi:
https://docs.oracle.com/javase/tutorial/java/javaOO/nested.html

Compelling reasons for using nested classes include the following:
- It is a way of logically grouping classes that are only used in one place: If a class is useful to only one other class, then it is logical to embed it in that class and keep the two together. Nesting such "helper classes" makes their package more streamlined.
- It increases encapsulation: Consider two top-level classes, A and B, where B needs access to members of A that would otherwise be declared private. By hiding class B within class A, A's members can be declared private and B can access them. In addition, B itself can be hidden from the outside world.
- It can lead to more readable and maintainable code: Nesting small classes within top-level classes places the code closer to where it is used.

So from my perspective, unless:
- nested classes become very complex
- nested classes need to be used by multiple classes

In this cases, nested class should be an external separate class.

But now, I don't think it is too complicated or used by multiple classes.

PR #1949, opened 2023-09-06
[flink] Auto create new tag for flink checkpoint

### LINKED ISSUE #1942: [Feature] auto create new tag for flink checkpoint
### Search before asking

### Motivation

企业微信截图_724b619c-2161-4831-9f96-b75fcfe3bfbb
Paimon's auto-tag didn't get created yesterday. Could it be that it won't trigger if there are no data changes? Is it triggered on write?
企业微信截图_3eecf419-c8e8-4a96-8bd3-2b121c467111
As expected, it's triggered on write. I manually inserted a few rows of data, and the tags were generated.

This can't be used in a batch processing scenario because batch processing requires an accurate trigger signal. When no data is written into the stream table, the auto-generation of tags is interrupted. Therefore, we need a mechanism for generating tags based on checkpoints. This mechanism should take into account settings for parameters like 'tag.automatic-creation', 'tag.creation-period', 'tag.creation-delay', and 'tag.num-retained-max'.

- [ ] Decouple tag Auto Creation logic #1940
- [ ] auto create new tag for flink checkpoint #1949 

### Solution

_No response_

### Anything else?

_No response_

MonsterChenzhuo:
### Purpose

close #1942

图片
The generation of tags is driven by checkpoints. Although no new data is written to trigger the creation of a new snapshot, tags can still be generated and point to the current latest snapshot.


docs/content/maintenance/manage-tags.md
```diff
@@ -257,6 +257,11 @@ Paimon's snapshot is similar to flink's checkpoint, and both will automatically
 allows snapshots to be retained for a long time. Therefore, we can combine the two features of paimon's tag and flink's 
 savepoint to achieve incremental recovery of job from the specified savepoint.
 
+At the same time to open the change parameter will also promote the automatic creation of the tag, now although open the "tag.automatic-creation" parameter but,
```
JingsongLi, line 260: remove this sentence.. there is no `sink.savepoint.auto-tag` here.

Document after `You can set `sink.savepoint.auto-tag` to `true` to enable the feature of automatically creating tags for savepoint.`

JingsongLi, line 260: Just simplify documentation. For example: `This option will also force creating a Tag for the tag.automatic-creation=process-time mode at an appropriate time even when no data is written.`

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/AutoTagForSavepointCommitterOperator.java
```diff
@@ -58,31 +60,27 @@ public class AutoTagForSavepointCommitterOperator<CommitT, GlobalCommitT>
         implements OneInputStreamOperator<CommitT, CommitT>,
                 SetupableStreamOperator,
                 BoundedOneInput {
-    public static final String SAVEPOINT_TAG_PREFIX = "savepoint-";
-
+    @VisibleForTesting public static final String SAVEPOINT_TAG_PREFIX = "savepoint-";
     private static final long serialVersionUID = 1L;
-
```
JingsongLi, line 64: Don't remove this empty line, this is just personal code style

paimon-core/src/main/java/org/apache/paimon/utils/TagManager.java
```diff
@@ -61,7 +61,9 @@ public Path tagPath(String tagName) {
     /** Create a tag from given snapshot and save it in the storage. */
     public void createTag(Snapshot snapshot, String tagName) {
         checkArgument(!StringUtils.isBlank(tagName), "Tag name '%s' is blank.", tagName);
-        checkArgument(!tagExists(tagName), "Tag name '%s' already exists.", tagName);
+        if (tagExists(tagName)) {
```
JingsongLi, line 64: Why add this?

paimon-core/src/main/java/org/apache/paimon/tag/TagAutoCreation.java
```diff
@@ -115,10 +114,20 @@ private TagAutoCreation(
         }
     }
 
+    public void refreshTags() {
```
JingsongLi, line 117: We should not add new method for `TagAutoCreation`.
We should just invoke `run`.

We can create a method:
```
private void tryToTag(Snapshot snapshot, LocalDateTime time) {
        if (nextTag == null
                || isAfterOrEqual(time.minus(delay), periodHandler.nextTagTime(nextTag))) {
            LocalDateTime thisTag = periodHandler.normalizeToTagTime(time);
            String tagName = periodHandler.timeToTag(thisTag);
            tagManager.createTag(snapshot, tagName);
            nextTag = periodHandler.nextTagTime(thisTag);

            if (numRetainedMax != null) {
                SortedMap<Snapshot, String> tags = tagManager.tags();
                if (tags.size() > numRetainedMax) {
                    int toDelete = tags.size() - numRetainedMax;
                    int i = 0;
                    for (String tag : tags.values()) {
                        tagManager.deleteTag(tag, tagDeletion, snapshotManager);
                        i++;
                        if (i == toDelete) {
                            break;
                        }
                    }
                }
            }
        }
    }
```

And in `run`:
```
    public void run() {
        while (true) {
            if (snapshotManager.snapshotExists(nextSnapshot)) {
                tryToTag(snapshotManager.snapshot(nextSnapshot));
                nextSnapshot++;
            } else {
                // avoid snapshot has been expired
                Long earliest = snapshotManager.earliestSnapshotId();
                if (earliest != null && earliest > nextSnapshot) {
                    nextSnapshot = earliest;
                } else {
                    break;
                }
            }
        }

        if (timeExtractor instanceof ProcessTimeExtractor) {
            tryToTag(snapshotManager.snapshot(nextSnapshot - 1), LocalDateTime.now());
        }
    }
```

Something like above codes.

JingsongLi (changes requested):

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/AutoTagForSavepointCommitterOperator.java
```diff
@@ -94,7 +92,7 @@ public void initializeState(StreamTaskStateInitializer streamTaskStateManager)
         } finally {
             snapshotManager = snapshotManagerFactory.get();
             tagManager = tagManagerFactory.get();
-
+            tagAutoCreation = tagAutoCreationFactory.get();
```
JingsongLi, line 95: Maybe you can try to extract `tagAutoCreation` from `CommitterOperator`. It is better to maintain only one object for in the committer.

docs/content/maintenance/manage-tags.md
MonsterChenzhuo: done.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/AutoTagForSavepointCommitterOperator.java
MonsterChenzhuo: done.

paimon-core/src/main/java/org/apache/paimon/utils/TagManager.java
MonsterChenzhuo: done.

paimon-core/src/main/java/org/apache/paimon/tag/TagAutoCreation.java

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/AutoTagForSavepointCommitterOperator.java
MonsterChenzhuo: done.

MonsterChenzhuo:
@JingsongLi  PATL,thanks.

MonsterChenzhuo:
Colleagues reported that some problems were found during testing and may need to be modified tomorrow.

JingsongLi:
I think we can use a generic way to implement this.
For example, if the extractor is `ProcessTimeExtractor` in `TagAutoCreation`, we can launch a async thread to check tag creation.

JingsongLi:
When I developed https://github.com/apache/incubator-paimon/pull/2134
I found that snapshots are our fundamental solution, and another way to solve this problem is to force the creation of snapshots for tables (in this configuration case, or add a new configuration option), which is a relatively simple thing.

JingsongLi:
#2188 should fixes this.

PR #1960, opened 2023-09-07
[doc] Fix Hive data type error in hive type conversion document.

zhuangchong:
### Purpose

Fix Hive data type error in hive type conversion document.

See https://github.com/apache/hive/blob/a4f895ac2eb772ff211c6f98418c7c12387a3a13/serde/src/java/org/apache/hadoop/hive/serde2/typeinfo/TypeInfoFactory.java#L46-L61


JingsongLi (approved):
+1

PR #1962, opened 2023-09-07
[flink] CDC Ingestion options change did not trigger schema update

### LINKED ISSUE #1951: [Bug] options change did not trigger schema update
### Search before asking

### Paimon version

0.5

### Compute Engine

flink

### Minimal reproduce step

Synchronize the MySQL table to Paimon, stop the Flink program, modify the table-conf configuration to change 'tag.creation-period' from 'daily' to 'hourly', but the changes do not take effect.

Changing the parameter does not take effect, which is not as expected.

### Anything else?

_No response_

MonsterChenzhuo:
### Purpose

close #1951

The current PR has implemented the following: After canceling the single-table synchronization Flink job and modifying the settings, the changed or deleted table parameters take effect upon restart. For full database synchronization, in scenarios where new tables are added upon restart, the newly added tables can execute according to the new table configuration items.

### Tests

KafkaSchemaITCase#testOptionsChange()
MongoDBSyncTableActionITCase#testOptionsChange()
MySqlSyncTableActionITCase#testOptionsChange()
MySqlSyncDatabaseActionITCase#testNewlyAddedTablesOptionsChange()
MongoDBSyncDatabaseActionITCase#testNewlyAddedTablesOptionsChange()
KafkaSchemaITCase#testNewlyAddedTablesOptionsChange()


paimon-core/src/main/java/org/apache/paimon/tag/TagAutoCreation.java
```diff
@@ -109,8 +109,13 @@ private TagAutoCreation(
         } else {
             Snapshot lastTag = tags.lastKey();
             this.nextSnapshot = lastTag.id() + 1;
-
-            LocalDateTime time = periodHandler.tagToTime(tags.get(lastTag));
+            LocalDateTime time;
+            String lastTagValue = tags.get(lastTag);
+            if (periodHandler instanceof DailyTagPeriodHandler) {
```
JingsongLi, line 114: Is here a test for this?

JingsongLi:
Hi @MonsterChenzhuo , I think the table config should be dynamic options, it should not be updated to the schema.

paimon-core/src/main/java/org/apache/paimon/tag/TagAutoCreation.java

MonsterChenzhuo:
> Hi @MonsterChenzhuo , I think the table config should be dynamic options, it should not be updated to the schema.

done.

MonsterChenzhuo:
@JingsongLi  PTAL,thanks.

JingsongLi (approved):
+1 Thanks @MonsterChenzhuo

PR #1967, opened 2023-09-08
[flink] Fixed to a more reasonable timeout

MonsterChenzhuo:
### Purpose

fixed to a more reasonable timeout.


JingsongLi:
Sometimes we may need a bigger timeout... 60 s is too small..

PR #1969, opened 2023-09-08
[core] Cache record in append-only write using `ExternalBuffer`, avoid out-of-memory error while writing lots of partitions

leaves12138:
### Problem
  While writing to append-only table, we have no cache in append-only writer. If the records between two checkpoint interval too many, we will get a out-of-memory exception. In conclusion, it is not a good idea to cache all the records between two checkpoint in memory. So here, introduce ExternalBuffer to cache the record, and flush it while memory pool is exhausted.

### Influence 
1、 The write speed will be slowed down.  We have to serialize all the record to segment, and deserialize it while checkpoint comming.
2、The  checkpoint complete will be delayed.  If we spill record on disk, we have to read them and write it to a file.

### Discuss
Now `ExternalBuffer` default enable spill and have no parameter to disable that. Whether we need an option to disable spill function, and to be compatible with the table-config "write-buffer-spillable".  Or just enable spill on `ExternalBuffer` by default.

Please give me some advice about the discuss ...

### Tests
Added tests for `SpillableBuffer` and `AppendOnlyWrite`

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
@@ -277,8 +277,8 @@ public class CoreOptions implements Serializable {
                     .withDescription(
                             "Whether the write buffer can be spillable. Enabled by default when using object storage.");
 
-    public static final ConfigOption<Boolean> USE_WRITE_BUFFER =
-            key("use-write-buffer")
+    public static final ConfigOption<Boolean> WRITE_BUFFER_FOR_APPEND =
+            key("write-buffer-for-append")
                     .booleanType()
                     .defaultValue(false)
                     .withDescription(
@@ -996,7 +996,7 @@ public class CoreOptions implements Serializable {
     }
 
     public boolean useWriteBuffer() {
-        return options.get(USE_WRITE_BUFFER);
+        return options.get(WRITE_BUFFER_FOR_APPEND);
     }
 
     public long sortSpillBufferSize() {
--- a/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
@@ -277,6 +277,13 @@ public class CoreOptions implements Serializable {
                     .withDescription(
                             "Whether the write buffer can be spillable. Enabled by default when using object storage.");
 
+    public static final ConfigOption<Boolean> USE_WRITE_BUFFER =
+            key("use-write-buffer")
+                    .booleanType()
+                    .defaultValue(false)
+                    .withDescription(
+                            "This option only works for append-only table. Whether the write use write buffer to avoid out-of-memory error.");
+
     public static final ConfigOption<MemorySize> WRITE_MANIFEST_CACHE =
             key("write-manifest-cache")
                     .memoryType()
@@ -986,6 +993,10 @@ public class CoreOptions implements Serializable {
     public boolean writeBufferSpillable(boolean usingObjectStore, boolean isStreaming) {
         // if not streaming mode, we turn spillable on by default.
         return options.getOptional(WRITE_BUFFER_SPILLABLE).orElse(usingObjectStore || !isStreaming);
+    }
+
+    public boolean useWriteBuffer() {
+        return options.get(USE_WRITE_BUFFER);
     }
 
     public long sortSpillBufferSize() {
--- a/paimon-core/src/main/java/org/apache/paimon/append/AppendOnlyWriter.java
+++ b/paimon-core/src/main/java/org/apache/paimon/append/AppendOnlyWriter.java
@@ -24,7 +24,7 @@ import org.apache.paimon.compact.CompactManager;
 import org.apache.paimon.data.InternalRow;
 import org.apache.paimon.data.serializer.InternalRowSerializer;
 import org.apache.paimon.disk.IOManager;
-import org.apache.paimon.disk.SpillableBuffer;
+import org.apache.paimon.disk.InternalRowBuffer;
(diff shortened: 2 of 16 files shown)


leaves12138:
Now I add the spillable config to SpillableBuffer (which is renamed from ExternalBuffer). And if spillable is false, append-only writer will flush the files to disk, which may cause small too many small files.

leaves12138:
Please review my code and give me some options.

wxplovecc:
Append only table write to files direct in my understanding, the OOM maybe caused by too many writers?

JingsongLi:
I think we should keep directly writing for append table.

leaves12138:
> Append only table write to files direct in my understanding, the OOM maybe caused by too many writers?

Records may cache in OrcWriter/ParquetWriter, and too many writers cache too many records cause the OOM

leaves12138:
> I think we should keep directly writing for append table by default. For example, 256MB in memory buffer may can not write 128MB orc file, but previous behavior can.

reasonable

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -277,6 +277,13 @@ public class CoreOptions implements Serializable {
                     .withDescription(
                             "Whether the write buffer can be spillable. Enabled by default when using object storage.");
 
+    public static final ConfigOption<Boolean> USE_WRITE_BUFFER =
+            key("use-write-buffer")
```
JingsongLi, line 281: `write-buffer-for-append`?

JingsongLi, line 281: Also document this in append-only documentation page.

paimon-core/src/main/java/org/apache/paimon/disk/SpillableBuffer.java
```diff
@@ -53,14 +53,18 @@ public class ExternalBuffer {
...
+    private final boolean spillable;
 
     private final List<ChannelWithMeta> spilledChannelIDs;
     private int numRows;
 
     private boolean addCompleted;
 
-    public ExternalBuffer(
-            IOManager ioManager, MemorySegmentPool pool, AbstractRowDataSerializer<?> serializer) {
+    public SpillableBuffer(
```
JingsongLi, line 63: Don't change this name... The renaming is not so solid.

paimon-core/src/main/java/org/apache/paimon/disk/SpillableBuffer.java
```diff
@@ -53,14 +53,18 @@ public class ExternalBuffer {
...
+            IOManager ioManager,
+            MemorySegmentPool pool,
+            AbstractRowDataSerializer<?> serializer,
+            boolean spillable) {
```
JingsongLi, line 67: You should first introduce a new interface, and ExternalBuffer and InMemoryBuffer implements it.

paimon-core/src/main/java/org/apache/paimon/sort/BinaryInMemorySortBuffer.java
```diff
@@ -182,6 +182,7 @@ public boolean write(InternalRow record) throws IOException {
         long currOffset = currentDataBufferOffset + skip;
 
         writeIndexAndNormalizedKey(record, currOffset);
+        sortIndexBytes += indexEntrySize;
```
JingsongLi, line 185: This can be a separate PR.

paimon-core/src/main/java/org/apache/paimon/append/AppendOnlyWriter.java
```diff
@@ -63,12 +69,17 @@ public class AppendOnlyWriter implements RecordWriter<InternalRow> {
     private final List<DataFileMeta> compactAfter;
     private final LongCounter seqNumCounter;
     private final String fileCompression;
+    private final boolean useWriteBuffer;
+    private final boolean spillable;
     private final FieldStatsCollector.Factory[] statsCollectors;
+    private final IOManager ioManager;
 
+    private SpillableBuffer writeBuffer;
```
JingsongLi, line 77: You can create a small inner `interface`. And with two small inner class implement it.

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
leaves12138: done

paimon-core/src/main/java/org/apache/paimon/disk/SpillableBuffer.java
leaves12138: done

paimon-core/src/main/java/org/apache/paimon/sort/BinaryInMemorySortBuffer.java
leaves12138: reverted

paimon-core/src/main/java/org/apache/paimon/append/AppendOnlyWriter.java
leaves12138: done

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
leaves12138: done

paimon-core/src/main/java/org/apache/paimon/disk/SpillableBuffer.java
leaves12138: reverted

JingsongLi (approved):
+1

JingsongLi (approved):
+1

PR #1975, opened 2023-09-08
[core] Introduce index-ttl and bootstrap-min-partition for cross-partition-upsert

JingsongLi:
### Purpose

Cross-partition-upsert Performance: For tables with a large amount of data, there will be a significant loss in performance. Moreover,
initialization takes a long time.

If your upsert does not rely on too old data, you can consider configuring index TTL and bootstrap-min-partition to
reduce Index and initialization time:
- `'cross-partition-upsert.index-ttl'`: The TTL in rocksdb index, this can avoid maintaining too many indexes and lead
  to worse and worse performance.
- `'cross-partition-upsert.bootstrap-min-partition'`: The min partition bootstrap of rocksdb index, bootstrap will only
  read the partitions above it, and the smaller partitions will not be read into the index. This can reduce job startup
  time and excessive initialization of index.

But please note that this may also cause data duplication.

### DIFF
--- a/docs/content/concepts/primary-key-table.md
+++ b/docs/content/concepts/primary-key-table.md
@@ -64,13 +64,13 @@ Performance:
    entries in a partition takes up **1 GB** more memory, partitions that are no longer active do not take up memory.
 2. For tables with low update rates, this mode is recommended to significantly improve performance.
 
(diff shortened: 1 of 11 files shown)


yuzelin (approved):
+1

PR #1982, opened 2023-09-11
[core] Introduce PartitionPredicate to improve performance for multiple partitions

JingsongLi:
### Purpose

When considering tens of thousands of partitions, in Committer node:
- `FileStoreCommitImpl.readAllEntriesFromChangedPartitions` will read all partitions, the predicate will be tens of thousands of `OR`, this will be very low performance.
- We can optimize this by using specific `PartitionPredicate`.

### Tests

- `PartitionPredicateTest`.


yuzelin (approved):
+1

PR #1986, opened 2023-09-12
[flink] Introduce fine-grained options to increase Committer heap memory only

JingsongLi:
### Purpose

Committer node may use a large memory if the amount of data written to the table is particularly large, OOM may occur
if the memory is too small. In this case, you need to increase the Committer heap memory, but you may not want to
increase the memory of Flink's TaskManager uniformly, which may lead to a waste of memory.

You can use fine-grained-resource-management of Flink to increase committer heap memory only:
1. Configure Flink Configuration `cluster.fine-grained-resource-management.enabled: true`. (This is default after Flink 1.18)
2. Configure Paimon Table Options: `sink.committer-cpu` (for example 1.0) and `sink.committer-memory` (for example 300 MB, depends on your `TaskManager`).


yuzelin (approved):
+1

PR #1993, opened 2023-09-12
[core] Clean StreamTableScan public interface

JingsongLi:
### Purpose

After Spark streaming source implementation. We should clean StreamTableScan public interface.


yuzelin (approved):
+1

PR #1996, opened 2023-09-13
[common] Load plugin classes directly from jar without decompression

tsreaper:
### Purpose

Currently to load classes from plugins, `PluginLoader` have to decompress a jar file out of Paimon distribution jar and load that jar file. It is hard for a long existing client (for example a SQL gateway service) to clean up these jar files.

This PR fixes the issue by loading plugin classes directly from distribution jar without decompression.

### Tests

### API and Format

No.

### Documentation

No.

### DIFF
--- a/paimon-codegen-loader/pom.xml
+++ b/paimon-codegen-loader/pom.xml
@@ -65,10 +65,10 @@ under the License.
                 <artifactId>maven-dependency-plugin</artifactId>
                 <executions>
                     <execution>
-                        <id>copy-table-planner-jars</id>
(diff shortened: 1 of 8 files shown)


JingsongLi:
This solution is not work in JDK 11 and JDK 17.

tsreaper:
> This solution can not work in JDK 11 and JDK 17.

Changed the solution a bit and now it works on both JDK 11 and 17.

JingsongLi (approved):
Looks good to me!

PR #1997, opened 2023-09-13
[flink] The table's primary key use a fixed constraint name.

### LINKED ISSUE #1970: [Feature] The table's primary key should use a fixed constraint name.
### Search before asking

### Motivation

Now, the table's primary key uses UUID as the constraint name, and there are no issues with regular usage. However, starting from Flink 1.18, it allows setting the state TTL for each operator (https://nightlies.apache.org/flink/flink-docs-master/docs/dev/table/concepts/overview/#configure-operator-level-state-ttl). This leads to the following problem:

1. When exporting a JSON plan using `COMPILE PLAN`, it calls `FlinkCatalog#toCatalogTable` to retrieve the table's constraint name, which will be a UUID as expected.
2. After modifying the operator's TTL.
3. Then, when executing the modified JSON plan using `EXECUTE PLAN`, it once again calls `FlinkCatalog#toCatalogTable` to fetch the table's constraint name. Naturally, this results in a new UUID, which is compared to the UUID in the JSON plan(comparing the schema before and after). When a mismatch is detected, it results in an error and the execution fails.

**first:**
// add primary keys
        if (table.primaryKeys().size() > 0) {
            &emsp;builder.primaryKey(table.primaryKeys().toArray(new String[0]));
        }
**and then:**
public Builder primaryKey(String... columns) {
            &emsp;return primaryKey(UUID.randomUUID().toString(), columns);
        }

### Solution

Generate a fixed constraint name for the primary key following the strategy used in Flink's Schema. https://github.com/apache/flink/blob/c4a76e2f006d834771092a8fac96bfc63361e20b/flink-table/flink-table-common/src/main/java/org/apache/flink/table/api/Schema.java#L553

### Anything else?

_No response_

hzjhjjyy:
### Purpose

Linked issue: close #1970

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkCatalog.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkCatalog.java
@@ -604,7 +604,9 @@ public class FlinkCatalog extends AbstractCatalog {
 
         // add primary keys
         if (table.primaryKeys().size() > 0) {
-            builder.primaryKey(table.primaryKeys().toArray(new String[0]));
+            builder.primaryKey(
+                    table.primaryKeys().stream().collect(Collectors.joining("_", "PK_", "")),
+                    table.primaryKeys().toArray(new String[0]));
         }
 
         TableSchema schema = builder.build();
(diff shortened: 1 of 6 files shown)


JingsongLi (approved):
+1

PR #2004, opened 2023-09-13
[flink] Use InternalRow instead of RowData in sink sub topo

JingsongLi:
### Purpose

This can avoid many conversion cost between InternalRow and RowData.


yuzelin (approved):
+1

PR #2012, opened 2023-09-14
[cdc] mysql cdc does not obtain whether the original field data type is allowed to be null

### LINKED ISSUE #2011: [Bug] mysql cdc does not obtain whether the original field data type is allowed to be null
### Search before asking

### Paimon version

0.6

### Compute Engine

mysql cdc

### Minimal reproduce step

When mysql synchronizes the entire database, a field in a table is not allowed to be empty. After synchronizing to paimon, this field in the paimon table schema is allowed to be null.

mysql table schema is consistent with paimon table schema.

### Anything else?

_No response_

zhuangchong:
### Purpose

Linked issue: close #2011


yuzelin (approved):
+1

zhuangchong:
Thanks. @yuzelin

PR #2013, opened 2023-09-14
[flink] Support compact procedure

yuzelin:
### Tests

`CompactActionITCase`
`SortCompactActionITCase`

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
@@ -27,6 +27,7 @@ import org.apache.paimon.table.Table;
 import java.io.Serializable;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 
 /**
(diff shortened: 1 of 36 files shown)


paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
```diff
@@ -212,6 +213,12 @@ default boolean caseSensitive() {
         return true;
     }
 
+    /** Return the warehouse path. */
+    String warehouse();
```
JingsongLi, line 217: Cast to `AbstractCatalog` to get warehouse and options.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/procedure/CompactProcedure.java
```diff
@@ -0,0 +1,133 @@
...
+            }
+            action.withPartitions(partitions);
+        }
+
+        StreamExecutionEnvironment env = procedureContext.getExecutionEnvironment();
+        action.build(env);
+
+        ReadableConfig conf = StreamExecutionEnvironmentUtils.getConfiguration(env);
+        String name = conf.getOptional(PipelineOptions.NAME).orElse(jobName);
+        if (conf.get(ExecutionOptions.RUNTIME_MODE) == RuntimeExecutionMode.STREAMING) {
```
JingsongLi, line 125: We should use `TABLE_DML_SYNC` in Flink.

JingsongLi (approved):
+1

PR #2018, opened 2023-09-14
Support set execute mode for compact action

hehuiyuan:
### Purpose

Support set execute mode for compact action.

We can set execute mode by main args  ,  this  default value of  `execution.runtime-mode`  is streaming.

For example , 
When we submit a  compcat flink job by  rest service , we can set main args :
compact --path hdfs://ns1007/user/mart_sch/tmp.db/paimon_test_table26 --catalog-conf metastore=hive --execution.runtime-mode batch

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/CompactAction.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/CompactAction.java
@@ -22,10 +22,13 @@ import org.apache.paimon.CoreOptions;
 import org.apache.paimon.flink.compact.UnawareBucketCompactionTopoBuilder;
 import org.apache.paimon.flink.sink.CompactorSinkBuilder;
 import org.apache.paimon.flink.source.CompactorSourceBuilder;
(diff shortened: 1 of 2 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/CompactAction.java
```diff
@@ -41,6 +38,7 @@
 public class CompactAction extends TableActionBase {
 
     private List<Map<String, String>> partitions;
+    private Boolean isStreaming;
```
SteNicholas, line 41: Use `RuntimeExecutionMode` instead of `isStreaming`.

JingsongLi:
You can use `-D execution.runtime-mode=batch`.

hehuiyuan:
> You can use `-D execution.runtime-mode=batch`.

flink run shell is ok .  But  it is ok if it is a common flink jar job?

hehuiyuan:
Hi @JingsongLi ,  we still have some requirements that need to be set through parameter for streaming or  batch execution mode.

PR #2026, opened 2023-09-15
[core] Partial-update streaming read supports input changelog-producer

JingsongLi:
### Purpose

We can support input changelog-producer for Partial-update streaming reading, but only returns input records.
This is also a requirement for scenarios where the downstream only wants to read incremental data.


yuzelin (approved):
+1

PR #2027, opened 2023-09-15
[flink] Translate actions to procedures

yuzelin:
### Purpose

DeleteAction is ignored because we can use flink `DELETE` statement.


JingsongLi (approved):
+1

PR #2030, opened 2023-09-15
[flink] Fix kafka CDC Ingestion multi-topic exception

MonsterChenzhuo:
### Purpose

fix kafka CDC Ingestion multi-topic exception


JingsongLi (approved):
+1

PR #2033, opened 2023-09-16
[core] Fix outputlevel is still 0 in extreme scenarios

lipenglin:
### Purpose
Fix outputlevel is still 0 in extreme scenarios.

This may not be a common case, but it is indeed possible.

For a newly created paimon pk table, with default parameters for num-levels and
others, without adjusting the max-size-amp, size-ratio, and other
configurations, assuming all configurations are using default values.

For a specific bucket:
file1 (1kb) is written,
file2 (1kb) is written,
file3 (1kb) is written,
file4 (1mb) is written,
file5 (10mb) is written

The srS of level0 will be sorted according to
https://github.com/apache/incubator-paimon/blob/master/paimon-core/src/main/java/org/apache/paimon/mergetree/Levels.java#L58 
In this situation, these five level0 files may not go through the pickForSizeAmp
function, outputlevel is 0. This may cause a larger file to appear in level0. I
think setting the outputlevel to 5 would be more ideal.

Linked issue: close #xxx

### Tests
Add UT org.apache.paimon.mergetree.compact.UniversalCompactionTest#testExtremeCaseNoOutputLevel0

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/UniversalCompaction.java
+++ b/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/UniversalCompaction.java
@@ -171,6 +171,10 @@ public class UniversalCompaction implements CompactStrategy {
             }
         }
 
+        if (outputLevel == 0) {
+            outputLevel = maxLevel;
+        }
+
         return CompactUnit.fromLevelRuns(outputLevel, runs.subList(0, runCount));
     }
 }
--- a/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/UniversalCompaction.java
+++ b/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/UniversalCompaction.java
@@ -171,7 +171,7 @@ public class UniversalCompaction implements CompactStrategy {
             }
         }
 
-        if (outputLevel == 0) {
+        if (runCount == runs.size()) {
             outputLevel = maxLevel;
         }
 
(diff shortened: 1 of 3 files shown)


JingsongLi:
See `UniversalCompaction.createUnit`, we force picking to output level 1...

lipenglin:
> See `UniversalCompaction.createUnit`, we force picking to output level 1...

But actually, this result is 0.

`        UniversalCompaction compaction = new UniversalCompaction(200, 1, 5);

        Optional<CompactUnit> pick =
                compaction.pick(
                        6,
                        Arrays.asList(
                                level(0, 1),
                                level(0, 1),
                                level(0, 1),
                                level(0, 1024),
                                level(0, 1024 * 1024)));

        assertThat(pick.isPresent()).isTrue();
        long[] results = pick.get().files().stream().mapToLong(DataFileMeta::fileSize).toArray();
        assertThat(results).isEqualTo(new long[] {1, 1, 1, 1024, 1024 * 1024});
        // pick.get().outputLevel() -> the result is 0
`

paimon-core/src/main/java/org/apache/paimon/mergetree/compact/UniversalCompaction.java
```diff
@@ -171,6 +171,10 @@ static CompactUnit createUnit(List<LevelSortedRun> runs, int maxLevel, int runCo
             }
         }
 
+        if (outputLevel == 0) {
```
JingsongLi, line 174:
```
if (runCount == runs.size()) {
    outputLevel = maxLevel;
}
```

JingsongLi:
@lipenglin You are right, see above comment.

lipenglin:
> @lipenglin You are right, see above comment.

done.

PTAL

JingsongLi:
Please fix tests.

lipenglin:
> Please fix tests.

done

JingsongLi (approved):
+1

PR #2036, opened 2023-09-18
[cdc] Refactor mysql schema build in mysql cdc action

zhuangchong:
### Purpose

1.Use public `Schema` to replace `MysqlSchema`
2.Add mysql table comments to paimon table schema


yuzelin (approved):
+1

PR #2042, opened 2023-09-19
[flink] optimize code style and remove useless code

leaves12138:
Just remove useless code, the sink builder will use the sorter parallelism, so we don't need to set parallelism.

No other influence.

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/SortCompactAction.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/SortCompactAction.java
@@ -116,16 +116,11 @@ public class SortCompactAction extends CompactAction {
         DataStream<RowData> source = sourceBuilder.withEnv(env).withContinuousMode(false).build();
         TableSorter sorter =
                 TableSorter.getSorter(env, source, fileStoreTable, sortStrategy, orderColumns);
-        DataStream<RowData> sorted = sorter.sort();
 
-        FlinkSinkBuilder flinkSinkBuilder = new FlinkSinkBuilder(fileStoreTable);
-        flinkSinkBuilder.withInput(sorted).withOverwritePartition(new HashMap<>());
-        String sinkParallelism = tableConfig.get(FlinkConnectorOptions.SINK_PARALLELISM.key());
-        if (sinkParallelism != null) {
-            flinkSinkBuilder.withParallelism(Integer.parseInt(sinkParallelism));
-        }
-
-        flinkSinkBuilder.build();
+        new FlinkSinkBuilder(fileStoreTable)
+                .withInput(sorter.sort())
+                .withOverwritePartition(new HashMap<>())
+                .build();
     }
 
     public SortCompactAction withOrderStrategy(String sortStrategy) {


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/SortCompactAction.java
```diff
@@ -116,16 +116,11 @@ public void build(StreamExecutionEnvironment env) {
...
-        flinkSinkBuilder.withInput(sorted).withOverwritePartition(new HashMap<>());
-        String sinkParallelism = tableConfig.get(FlinkConnectorOptions.SINK_PARALLELISM.key());
-        if (sinkParallelism != null) {
-            flinkSinkBuilder.withParallelism(Integer.parseInt(sinkParallelism));
-        }
-
-        flinkSinkBuilder.build();
+        new FlinkSinkBuilder(fileStoreTable)
+                .withInput(sorter.sort())
+                .withOverwritePartition(new HashMap<>())
```
SteNicholas, line 122: Could the `overwritePartition` of `withOverwritePartition` be nullable?

leaves12138: sure

SteNicholas: When `overwritePartition` is nullable, does this need to invoke `withOverwritePartition`?

leaves12138: Yes, it need to tag it as overwrite. If we don't call withOverwritePartition(new HastMap()), it will not be a compact and overwrite action, the data will be doubled

leaves12138: See `FlinkWriteSink` constructor.

leaves12138:
add comment for overwritePartition, @JingsongLi, do we need to add an extra parameter to tag whether to overwrite

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/SortCompactAction.java
```diff
@@ -116,16 +116,12 @@ public void build(StreamExecutionEnvironment env) {
...
-        flinkSinkBuilder.withInput(sorted).withOverwritePartition(new HashMap<>());
-        String sinkParallelism = tableConfig.get(FlinkConnectorOptions.SINK_PARALLELISM.key());
-        if (sinkParallelism != null) {
-            flinkSinkBuilder.withParallelism(Integer.parseInt(sinkParallelism));
-        }
-
-        flinkSinkBuilder.build();
+        new FlinkSinkBuilder(fileStoreTable)
+                .withInput(sorter.sort())
+                // we pass an object here to tag it an overwrite action.
```
SteNicholas, line 122:
```suggestion
                // This should use empty map to tag it on overwrite action, otherwise there is no overwrite action.
```

SteNicholas: @JingsongLi, this invoker makes confusing to me, could we optimize the logic of `commitMultiple`? IMO, null or empty map should be the same behavior in which there is no overwrite action.

SteNicholas (commented):
@leaves12138, thanks for updates.

SteNicholas (approved):
LGTM. Thanks for @leaves12138 optimization. cc @JingsongLi.

JingsongLi (approved):
+1

PR #2044, opened 2023-09-20
[core] Fix that time travel to tag with expired snapshot throws FileNotFoundException

yuzelin:
### Purpose

`StaticFromTagStartingScanner` should read a tag instead of a snapshot id.


JingsongLi (approved):
+1

PR #2047, opened 2023-09-21
[flink][cdc] Fix that setting both table affix and including pattern causes table cannot be synchronized in kafka cdc

yuzelin:
### Purpose

Table name should be converted in `RichCdcMultiplexRecordEventParser` instead of `RecordParser`


zhuangchong (approved):
LGTM.

PR #2048, opened 2023-09-21
[hive]When creating a hive external table pointing to the existing paimon table, the field name is capitalized, and hive show table does not display the field.

zhuangchong:
…

### Purpose

When creating a hive external table pointing to the existing paimon table, the field name is capitalized, and hive show table does not display the field.


JingsongLi (approved):
+1

PR #2058, opened 2023-09-22
[flink] create tags for the previous savepoint after a successful checkpoint.

### LINKED ISSUE #2057: [Bug] Flink does not automatically create tags after making savepoint.
### Search before asking

### Paimon version

0.6-SNAPSHOT

### Compute Engine

flink-1.17.1

### Minimal reproduce step

1. enable `sink.savepoint.auto-tag`
2. trigger savepoint for flink job.
3. check whether the tag is created.

A tag should be created for the snapshot corresponding to the savepoint.

### Anything else?

The reason is that currently flink will not send a complete message to the task after the savepoint is successful (unless use `stop-with-savepoint`).

liming30:
### Purpose

Linked issue: close #2057 

Currently flink will not send a complete message to the task after the savepoint is successful (unless use `stop-with-savepoint`).  Therefore, we need to create tags for the previous savepoint after the next checkpoint is successful.

### Tests

- `org.apache.paimon.flink.sink.AutoTagForSavepointCommitterOperatorTest#testAbortSavepointAndCleanTag` to verify that the tag can be cleaned after the savepoint abort.
- `org.apache.paimon.flink.sink.AutoTagForSavepointCommitterOperatorTest#testAutoTagForSavepoint` to verify that the tag can be automatically created for the savepoint after the checkpoint is successful.

### API and Format

No
### Documentation

No

### DIFF
--- a/docs/content/maintenance/manage-tags.md
+++ b/docs/content/maintenance/manage-tags.md
@@ -258,6 +258,15 @@ Paimon's snapshot is similar to flink's checkpoint, and both will automatically 
 allows snapshots to be retained for a long time. Therefore, we can combine the two features of paimon's tag and flink's 
 savepoint to achieve incremental recovery of job from the specified savepoint.
 
+{{< hint warning >}}
+Starting from Flink 1.15 intermediate savepoints (savepoints other than created with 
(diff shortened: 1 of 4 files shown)


JingsongLi (approved):
+1

PR #2062, opened 2023-09-23
[core] Support remove metric groups from metric instance

### LINKED ISSUE #2061: [Feature] Support delete metric group
### Search before asking

### Motivation

Now we support add metric groups to Paimon metrics instance, which will be exist in the memory always, in the long running system like flink session cluster, this will cause OOM eventually.
We need to support deleting metric groups.

### Solution

_No response_

### Anything else?

_No response_

schnappi17:
### Purpose

Linked issue: close #2061

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/metrics/AbstractMetricGroup.java
+++ b/paimon-core/src/main/java/org/apache/paimon/metrics/AbstractMetricGroup.java
@@ -183,6 +183,7 @@ public abstract class AbstractMetricGroup implements MetricGroup {
         if (!closed) {
             closed = true;
             metrics.clear();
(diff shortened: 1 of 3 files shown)


paimon-core/src/main/java/org/apache/paimon/metrics/AbstractMetricGroup.java
```diff
@@ -183,6 +183,7 @@ public void close() {
         if (!closed) {
             closed = true;
             metrics.clear();
+            Metrics.getInstance().removeGroup(this);
```
FangYongs, line 186: Add a test unit for this

FangYongs:
Thanks @schnappi17 for contribution, LGTM +1

PR #2076, opened 2023-09-26
[core] reduce unnecessary traversals to optimize the assign logic of PartitionIndex.

### LINKED ISSUE #2075: [Feature] Optimize the assign performance of PartitionIndex.
### Search before asking

### Motivation

Currently, `PartitionIndex` will retain all bucket information. When a new key arrives, it will traverse and check the bucket capacity to find the first bucket that is not full. There are many invalid traversals in this process, such as 1. checking buckets that do not belong to the assigner, 2. checking buckets that are already full. I think we can skip these checks.

### Solution

_No response_

### Anything else?

_No response_

liming30:
### Purpose

Linked issue: close #2075 

 [core] reduce unnecessary traversals to optimize the assign logic of PartitionIndex.

I made a simple benchmark and continuously used random keys to call the `assign` of `PartitionIndex`. The results showed that removing unnecessary traversals can improve performance by 50% to 200%.

|   | benchmark 1s |  benchmark 5s |  benchmark 10s |  benchmark 30s |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| before | 4247.443 ops/ms  | 2342.67 ops/ms  | 1766.141 ops/ms  | 1287.669 ops/ms  |
| after | 6324.565 ops/ms  | 5738.319 ops/ms  | 5200.766 ops/ms  | 4563.41 ops/ms  |

### Tests

The existing `org.apache.paimon.index.HashBucketAssignerTest` can verify this change.

### API and Format

No

### Documentation

No

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/index/PartitionIndex.java
+++ b/paimon-core/src/main/java/org/apache/paimon/index/PartitionIndex.java
@@ -27,8 +27,11 @@ import java.io.EOFException;
 import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.function.IntPredicate;
 
 import static org.apache.paimon.index.HashIndexFile.HASH_INDEX;
@@ -38,7 +41,9 @@ public class PartitionIndex {
 
     public final Int2ShortHashMap hash2Bucket;
 
-    public final Map<Integer, Long> bucketInformation;
+    public final Map<Integer, Long> nonFullBucketInformation;
+
+    public final Set<Integer> totalBucket;
 
     private final long targetBucketRowNumber;
 
@@ -51,7 +56,8 @@ public class PartitionIndex {
             Map<Integer, Long> bucketInformation,
             long targetBucketRowNumber) {
         this.hash2Bucket = hash2Bucket;
-        this.bucketInformation = bucketInformation;
+        this.nonFullBucketInformation = bucketInformation;
+        this.totalBucket = new HashSet<>(bucketInformation.keySet());
         this.targetBucketRowNumber = targetBucketRowNumber;
         this.lastAccessedCommitIdentifier = Long.MIN_VALUE;
         this.accessed = true;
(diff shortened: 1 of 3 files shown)


paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/commands/WriteIntoPaimonTable.scala
```diff
@@ -245,7 +245,8 @@ object WriteIntoPaimonTable {
               indexFileHandler,
               partition,
               targetBucketRowNumber,
-              (_) => true))
+              (_) => true,
+              buckFilter))
```
JingsongLi, line 249: Due to Spark AQE, it should be dangerous to do this filter.

JingsongLi:
Thanks @liming30 , benchmark looks very good~

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/commands/WriteIntoPaimonTable.scala
liming30: `buckFilter` should only be related to `numSparkPartitions` and `sparkPartitionId`, which is determined at runtime. Even if `loadIndex` loads all buckets, we still need to use `buckFilter` to filter when `assign`.  Can you help provide a case where there will be problems if only the relevant buckets are loaded in advance?

paimon-core/src/main/java/org/apache/paimon/index/PartitionIndex.java
```diff
@@ -66,41 +72,45 @@ public int assign(int hash, IntPredicate bucketFilter) {
...
                         maxBucket, targetBucketRowNumber));
     }
 
     public static PartitionIndex loadIndex(
             IndexFileHandler indexFileHandler,
             BinaryRow partition,
             long targetBucketRowNumber,
-            IntPredicate loadFilter) {
+            IntPredicate loadFilter,
+            IntPredicate bucketFilter) {
```
JingsongLi, line 113: Have you consider rescale case?

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/commands/WriteIntoPaimonTable.scala
JingsongLi: My mistake. You are right.

JingsongLi (approved):
+1

PR #2078, opened 2023-09-27
[cdc] Optimize Duplicate Field Detection in MySQL CDC Schema Building.

zhuangchong:
### Purpose

Optimize Duplicate Field Detection in MySQL CDC Schema Building.

### DIFF
--- a/paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/action/cdc/CdcActionCommonUtils.java
+++ b/paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/action/cdc/CdcActionCommonUtils.java
@@ -140,8 +140,9 @@ public class CdcActionCommonUtils {
         if (caseSensitive) {
             return column;
         }
-        checkArgument(existedFields.add(column.toLowerCase()), columnDuplicateErrMsg.apply(column));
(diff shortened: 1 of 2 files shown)


paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/action/cdc/CdcActionCommonUtils.java
```diff
@@ -140,8 +140,9 @@ public static String columnCaseConvertAndDuplicateCheck(
         if (caseSensitive) {
             return column;
         }
-        checkArgument(existedFields.add(column.toLowerCase()), columnDuplicateErrMsg.apply(column));
-        return column.toLowerCase();
```
yuzelin, line 144: Here I think we can keep the filed name case in error message.

zhuangchong: is reserved, the variable used in the error message is `column`, and the returned result variable is `columnLowerCase`

yuzelin (approved):
+1

zhuangchong:
Thanks. @yuzelin

PR #2081, opened 2023-09-28
[cdc] CDC schema evolution supports all type conversions already supported by paimon

### LINKED ISSUE #2080: [Feature] CDC schema evolution supports all type conversions already supported by paimon
### Search before asking

### Motivation

CDC schema evolution supports all type conversions already supported by paimon

paimon supported type conversions:
https://paimon.apache.org/docs/master/how-to/altering-tables/#changing-column-type

### Solution

_No response_

### Anything else?

_No response_

zhuangchong:
### Purpose

Linked issue: close #2080

### DIFF
--- a/docs/content/how-to/cdc-ingestion.md
+++ b/docs/content/how-to/cdc-ingestion.md
@@ -667,14 +667,7 @@ behaviors of `RENAME TABLE` and `DROP COLUMN` will be ignored, `RENAME COLUMN` w
 
 * Adding columns.
 
-* Altering column types. More specifically,
-
-  * altering from a string type (char, varchar, text) to another string type with longer length,
-  * altering from a binary type (binary, varbinary, blob) to another binary type with longer length,
-  * altering from an integer type (tinyint, smallint, int, bigint) to another integer type with wider range,
-  * altering from a floating-point type (float, double) to another floating-point type with wider range,
-
-  are supported.
+* Altering column types. See [ChangingColumnType]({{< ref "how-to/altering-tables#changing-column-type" >}}) for supported type conversions.
 
 ## Computed Functions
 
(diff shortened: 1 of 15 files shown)


JingsongLi (changes requested):
All type conversions may result blocking, if there is casting failure.

zhuangchong:
> All type conversions may result blocking, if there is casting failure.

Thanks. @JingsongLi 

I found that if the data conversion fails in the cdc sink operator, it will perform cyclic conversion processing, which will eventually lead to blocking. 

What do you think if this abnormal data is printed and filtered when the conversion fails?

JingsongLi:
> > All type conversions may result blocking, if there is casting failure.
> 
> Thanks. @JingsongLi
> 
> I found that if the data conversion fails in the cdc sink operator, it will perform cyclic conversion processing, which will eventually lead to blocking.
> 
> What do you think if this abnormal data is printed and filtered when the conversion fails?

We should not just throw abnormal data because of distributed processing.

So for our CDC, we don't support all type conversions.

zhuangchong:
> So for our CDC, we don't support all type conversions.

Explicit conversion may cause data conversion to fail, CDC can only support implicit conversion. This allows CDC schema changes to be supported more richly than before.

https://github.com/apache/incubator-paimon/blob/3abccb832a8bd9390c24a9d8ab61fa570750d84b/paimon-common/src/main/java/org/apache/paimon/types/DataTypeCasts.java#L180-L202

JingsongLi:
Not just implicit conversion.
Consider varchar(10) -> varchar(2) -> varchar(10). This is an implicit conversion, but is NOT OK for our CDC. The data may be lost because of varchar(2).

zhuangchong:
@JingsongLi Thanks. 

I didn't consider it comprehensively, so I will close this PR.

PR #2087, opened 2023-10-01
[core] Introduce bucket level value filter (filter push down) in primary key table

leaves12138:
### Purpose
Introduce bucket level value filter in primary key.

Current:  primary key does not support value filter.

Improve: if the bucket range min/max doesn't contains the filter target, we don't need to search the bucket any more.

### Tests

Add test.

### API and Format

The may speed up select in some conditions. No other influence to user.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreScan.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreScan.java
@@ -45,6 +45,7 @@ import javax.annotation.Nullable;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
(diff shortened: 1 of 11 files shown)


paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreScan.java
```diff
@@ -275,6 +276,20 @@ private Pair<Snapshot, List<ManifestEntry>> doPlan(
...
             }
         }
+
+        // We group files by bucket here, and filter them by the whole bucket filter.
+        // Why do this: because in primary key table, we can't just filter the value
+        // by the stat in files (see `ChangelogWithKeyFileStoreTable.nonPartitionFilterConsumer`),
+        // but we can do this by filter the whole bucket files
+        files =
+                files.stream()
+                        .collect(Collectors.groupingBy(ManifestEntry::bucket))
```
JingsongLi, line 286: Partition -> bucket

paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreScan.java
```diff
@@ -339,6 +354,9 @@ private boolean filterByLevel(ManifestEntry entry) {
     /** Note: Keep this thread-safe. */
     protected abstract boolean filterByStats(ManifestEntry entry);
 
+    /** Note: Keep this thread-safe. */
+    protected abstract boolean filterByStats(List<ManifestEntry> entries);
```
JingsongLi, line 358: `filterWholeBucketByStats`

JingsongLi:
Thanks @leaves12138 , left comments above.

paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreScan.java
leaves12138: done

paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreScan.java
leaves12138: done, added test for this

JingsongLi (approved):
+1

PR #2092, opened 2023-10-07
[core] Retry read consumer fromPath

wangyp501:
### Purpose
Retry read consumer fromPath , Avoid no content and Job restart

Exception：
0e94f5f70b46e6c0e6a6140ee6632549

### Tests
ConsumerManagerTest.java
testReadConsumer()

### Documentation
READ_CONSUMER_RETRY_NUM = 3;
READ_CONSUMER_RETRY_INTERVAL = 100 MILLISECONDS;

### DIFF
--- a/docs/content/concepts/append-only-table.md
+++ b/docs/content/concepts/append-only-table.md
@@ -241,12 +241,14 @@ You can trigger action by shell script:
     --warehouse hdfs:///path/to/warehouse \
     --database test_db \
     --table <tableName> \
+    --table-conf <key>=<value> \
     --order-strategy <orderType> \
     --order-by <col1,col2,...>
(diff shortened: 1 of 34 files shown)


wangyp501:
@JingsongLi PTAL,thanks.

JingsongLi (approved):
+1

wangyp501:
您好，我是王云朋，您的邮件我已经收到。

dragonkid:
I think this is still not the final resolution. It seems the exception will still be thrown on retries over the limit.
With the number of consumers on the same table growing, the probability of reading empty from `fromPath` is larger.

So, maybe the job will still fail randomly under this resolution?

PR #2097, opened 2023-10-08
[flink] simplify the usage of flink compact procedure

leaves12138:
### Purpose

Simplify the usage of compact procedure in flink action.

Just change the way to pass the parameters, no other influence to user.

### Documentation

Wait for @yuzelin repair merge-into and add doc for procedure together


JingsongLi (approved):
+1

PR #2099, opened 2023-10-08
[flink] Introduce Pulsar synchronizing table action

yuzelin:
### Purpose

Abstract message queue action base from the kafka action.

### Documentation

TODO


JingsongLi (changes requested):
Thanks @yuzelin for the contribution.

Please do not change package name. For example, the user of https://github.com/apache/incubator-paimon/pull/2098 uses KafkaAction to integrate to their internal platform. If there is no strong reason, we should avoid moving classes.

yuzelin:
> Thanks @yuzelin for the contribution.
> 
> Please do not change package name. For example, the user of #2098 uses KafkaAction to integrate to their internal platform. If there is no strong reason, we should avoid moving classes.

Reverted the KafkaAction movement. The tests are still moved to new place.

JingsongLi (approved):
+1

PR #2100, opened 2023-10-09
[flink] Introduce zorder/order sort compact for dynamic bucket table

leaves12138:
### Purpose

Speed up query performance with dynamic table.

We rewrite the dynamic table after z-order sorting the records, with the bucket filter function, we can speed up query. (The whole bucket may be filtered)

### Tests

End to end test in my pc.

SQL to generate data:
```
DROP TABLE IF EXISTS Orders1;
CREATE TABLE IF NOT EXISTS Orders1 (
    pk        INT,
    f0        INT,
    f1        INT,
    v         STRING,
    PRIMARY KEY(pk) NOT ENFORCED
)
WITH (
'bucket' = '-1',
'dynamic-bucket.target-row-num' = '10000'
);


CREATE TEMPORARY TABLE Orders_in (
    pk        INT,
    f0        INT,
    f1        INT,
    v         STRING
) WITH (
    'connector' = 'datagen',
    'fields.f0.min'   = '0',
    'fields.f0.max'   = '10000',
    'fields.f1.min'   = '0',
    'fields.f1.max'   = '10000',
    'number-of-rows' = '1000000'
);

SET 'parallelism.default' = '2';
SET 'execution.checkpointing.interval' = '20 s';

SET 'execution.runtime-mode' = 'streaming';
SET 'sql-client.execution.result-mode' = 'TABLEAU';

INSERT INTO Orders1 SELECT * FROM Orders_in;
```

Query sql:
the query sql `SELECT * FROM Orders1 where f1 < 200 and f1 > 100;`

Before sort:
take 2.5 s to get result, no bucket was filtered by value push down
image

After sort:
take 0.5s to get result, only 17% data was passed, 83% data was filtered by value push down.
image

### API and Format

Usage example:
flink run ./paimon-flink-action-0.6-SNAPSHOT.jar compact \
--warehouse \
/Users/paimontest/DynamicBucket \
--database \
my_db \
--table \
Orders1 \
--order-strategy zorder \
--order-by f0,f1

No other influence to user.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/index/SimpleBucketAssigner.java
+++ b/paimon-core/src/main/java/org/apache/paimon/index/SimpleBucketAssigner.java
@@ -1,0 +1,85 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.paimon.index;
+
+import org.apache.paimon.data.BinaryRow;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/** Only for batch sort compact bucket assign. */
+public class SimpleBucketAssigner implements BucketAssigner {
+
+    private final int numAssigners;
+    private final int assignId;
+    private final long targetBucketRowNumber;
+
+    private final Map<BinaryRow, SimplePartitionIndex> partitionIndex;
+
+    public SimpleBucketAssigner(int numAssigners, int assignId, long targetBucketRowNumber) {
+        this.numAssigners = numAssigners;
+        this.assignId = assignId;
+        this.targetBucketRowNumber = targetBucketRowNumber;
+        this.partitionIndex = new HashMap<>();
+    }
+
+    @Override
+    public int assign(BinaryRow partition, int hash) {
+        SimplePartitionIndex index =
+                this.partitionIndex.computeIfAbsent(partition, p -> new SimplePartitionIndex());
+        return index.assign();
+    }
+
+    @Override
+    public void prepareCommit(long commitIdentifier) {
+        // do nothing
+    }
+
+    /** Simple partition bucket assigner. */
+    private class SimplePartitionIndex {
+
+        private final Map<Integer, Long> bucketInformation;
+        private int currentBucket;
+
+        private SimplePartitionIndex() {
+            bucketInformation = new HashMap<>();
+            loadNewBucket();
+        }
+
+        public int assign() {
+            Long num = bucketInformation.computeIfAbsent(currentBucket, i -> 0L);
+            if (num >= targetBucketRowNumber) {
+                loadNewBucket();
+            }
+            bucketInformation.compute(currentBucket, (i, l) -> l == null ? 1L : l + 1);
+            return currentBucket;
+        }
+
+        private void loadNewBucket() {
+            for (int i = 0; i < Short.MAX_VALUE; i++) {
+                if (i % numAssigners == assignId && !bucketInformation.containsKey(i)) {
+                    currentBucket = i;
+                    return;
+                }
+            }
+            throw new RuntimeException(
+                    "Can't find a suitable bucket to assign, all the bucket are assigned?");
+        }
+    }
+}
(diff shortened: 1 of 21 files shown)


paimon-core/src/main/java/org/apache/paimon/index/SimpleBucketAssigner.java
```diff
@@ -0,0 +1,85 @@
...
+
+package org.apache.paimon.index;
+
+import org.apache.paimon.data.BinaryRow;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/** Only for batch sort compact bucket assign. */
+public class SimpleBucketAssigner implements BucketAssigner {
```
JingsongLi, line 27: For hash collision, you can not just use a new simple bucket assigner.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkSinkBuilder.java
```diff
@@ -44,6 +44,7 @@ public class FlinkSinkBuilder {
     @Nullable private Map<String, String> overwritePartition;
     @Nullable private LogSinkFunction logSinkFunction;
     @Nullable private Integer parallelism;
+    private boolean compactSink = false;
```
JingsongLi, line 47: You should just use overwrite to judge.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/DynamicBucketCompactSink.java
```diff
@@ -0,0 +1,60 @@
...
+
+        // bucket-assigner
+        BucketAssignerOperator<InternalRow> assignerOperator =
+                new BucketAssignerOperator<>(initialCommitUser, table, extractorFunction(), true);
+        TupleTypeInfo<Tuple2<InternalRow, Integer>> rowWithBucketType =
+                new TupleTypeInfo<>(input.getType(), BasicTypeInfo.INT_TYPE_INFO);
+        DataStream<Tuple2<InternalRow, Integer>> bucketAssigned =
+                input.transform("dynamic-bucket-assigner", rowWithBucketType, assignerOperator)
+                        .setParallelism(input.getParallelism());
+        return sinkFrom(bucketAssigned, initialCommitUser);
```
JingsongLi, line 58: Add test for overwritten, it seems that there is bug in AbstractFileStoreWrite for index maintaining.

paimon-core/src/main/java/org/apache/paimon/index/SimpleHashBucketAssigner.java
```diff
@@ -0,0 +1,93 @@
...
+
+package org.apache.paimon.index;
+
+import org.apache.paimon.data.BinaryRow;
+import org.apache.paimon.utils.Int2ShortHashMap;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/** Only for batch sort compact bucket assign. */
```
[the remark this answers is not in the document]
leaves12138: done

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkSinkBuilder.java
leaves12138: done

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/DynamicBucketCompactSink.java
leaves12138: fixed this bug and added two tests for this

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/BucketAssignerOperator.java
```diff
@@ -32,7 +34,7 @@
 import org.apache.flink.streaming.runtime.streamrecord.StreamRecord;
 
 /** Assign bucket for the input record, output record with bucket. */
-public class HashBucketAssignerOperator<T> extends AbstractStreamOperator<Tuple2<T, Integer>>
+public class BucketAssignerOperator<T> extends AbstractStreamOperator<Tuple2<T, Integer>>
```
JingsongLi, line 37: revert this name

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkSinkBuilder.java
```diff
@@ -108,9 +108,13 @@ public DataStreamSink<?> build() {
     private DataStreamSink<?> buildDynamicBucketSink(
             DataStream<InternalRow> input, boolean globalIndex) {
         checkArgument(logSinkFunction == null, "Dynamic bucket mode can not work with log system.");
-        return globalIndex
-                ? new GlobalDynamicBucketSink(table, overwritePartition).build(input, parallelism)
-                : new RowDynamicBucketSink(table, overwritePartition).build(input, parallelism);
+        return overwritePartition != null
```
JingsongLi, line 111: Here should only work for `compact & !globalIndex`

leaves12138: Added a todo here for `Global Index Sort Compact`, because Global Index table sort compact will need extra tests and   
and end-to-end tests to guarantee it works.

leaves12138: done this judge

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/BucketAssignerOperator.java
leaves12138: done

JingsongLi (approved):
+1

PR #2104, opened 2023-10-10
[flink] Support to set the table option by dynamic global job configu…

### LINKED ISSUE #2037: [Feature] Support to set the table option by dynamic global job configuration
### Search before asking

### Motivation

When using Flink to query the Paimon table, supporting overwriting the catalog table option by `SET` syntax will be more convenient. The dynamic options can only take effect in the current session.

### Solution

_No response_

### Anything else?

_No response_

Aitozi:
…ration

### Purpose

Linked issue: close #2037

### DIFF
--- a/docs/content/engines/flink.md
+++ b/docs/content/engines/flink.md
@@ -301,3 +301,19 @@ Users can set memory weight in SQL for Flink Managed Memory, then Flink sink ope
 INSERT INTO paimon_table /*+ OPTIONS('sink.use-managed-memory-allocator'='true', 'sink.managed.writer-buffer-memory'='256M') */
 SELECT * FROM ....;
 ```
+## Setting dynamic options
+
+When interacting with the Paimon table, table options can be tuned without changing the options in the catalog. Paimon will extract job-level dynamic options and take effect in the current session.
+The dynamic option's key format is `paimon.${catalogName}.${dbName}.${tableName}.${config_key}`. The catalogName/dbName/tableName can be `*`, which means matching all the specific parts. 
+
+For example:
+
+```sql
+-- set scan.timestamp-millis=1697018249000 for the table mycatalog.default.T
+SET 'paimon.mycatalog.default.T.scan.timestamp-millis' = '1697018249000';
+SELECT * FROM T;
+
+-- set scan.timestamp-millis=1697018249000 for the table default.T in any catalog
+SET 'paimon.*.default.T.scan.timestamp-millis' = '1697018249000';
+SELECT * FROM T;
+```
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/AbstractFlinkTableFactory.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/AbstractFlinkTableFactory.java
@@ -315,7 +315,7 @@ public abstract class AbstractFlinkTableFactory
 
         String template =
                 String.format(
-                        "(%s|\\*)\\.(%s|\\*)\\.(%s|\\*)\\.(%s|\\*)\\.(.+)",
+                        "(%s)\\.(%s|\\*)\\.(%s|\\*)\\.(%s|\\*)\\.(.+)",
                         FlinkConnectorOptions.TABLE_DYNAMIC_OPTION_PREFIX,
                         context.getObjectIdentifier().getCatalogName(),
                         context.getObjectIdentifier().getDatabaseName(),
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/AbstractFlinkTableFactory.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/AbstractFlinkTableFactory.java
@@ -36,6 +36,7 @@ import org.apache.paimon.lineage.TableLineageEntity;
 import org.apache.paimon.lineage.TableLineageEntityImpl;
 import org.apache.paimon.options.Options;
 import org.apache.paimon.schema.Schema;
+import org.apache.paimon.schema.SchemaManager;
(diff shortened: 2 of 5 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/AbstractFlinkTableFactory.java
```diff
@@ -263,4 +286,57 @@ static boolean schemaEquals(RowType rowType1, RowType rowType2) {
...
+     * The dynamic option's format is:
+     *
+     * <p>{@link
+     * FlinkConnectorOptions#TABLE_DYNAMIC_OPTION_PREFIX}.${catalog}.${database}.${tableName}.key =
+     * value. These job level configs will be extracted and injected into the target table option.
+     *
+     * @param context The table factory context.
+     * @return The dynamic options of this target table.
+     */
+    static Map<String, String> getDynamicTableConfigOptions(DynamicTableFactory.Context context) {
```
JingsongLi, line 300: Can you also document this in flink page?

Aitozi: Updated

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/AbstractFlinkTableFactory.java
```diff
@@ -263,4 +286,57 @@ static boolean schemaEquals(RowType rowType1, RowType rowType2) {
...
+                        FlinkConnectorOptions.TABLE_DYNAMIC_OPTION_PREFIX,
+                        context.getObjectIdentifier().getCatalogName(),
+                        context.getObjectIdentifier().getDatabaseName(),
+                        context.getObjectIdentifier().getObjectName());
+        Pattern pattern = Pattern.compile(template);
+
+        conf.keySet()
+                .forEach(
+                        (key) -> {
+                            Matcher matcher = pattern.matcher(key);
```
JingsongLi, line 328: `pattern.matcher` is very slow, can we just check starts with `paimon.` first?

docs/content/engines/flink.md
```diff
@@ -301,3 +301,19 @@ Users can set memory weight in SQL for Flink Managed Memory, then Flink sink ope
...
+
+For example:
+
+```sql
+-- set scan.timestamp-millis=1697018249000 for the table mycatalog.default.T
+SET 'paimon.mycatalog.default.T.scan.timestamp-millis' = '1697018249000';
+SELECT * FROM T;
+
+-- set scan.timestamp-millis=1697018249000 for the table default.T in any catalog
+SET 'paimon.*.default.T.scan.timestamp-millis' = '1697018249000';
```
JingsongLi, line 317: Can we just support paimon.{key} for all tables?

Aitozi: for all table may cause unexpected behavior, if we use two table but only want to set option to one of them

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/AbstractFlinkTableFactory.java
Aitozi: It's only happened in client planner phase and the pattern is not complex

JingsongLi: You can do a little benchmark for these two methods, and I think there may be hundreds options need to be checked.

docs/content/engines/flink.md
JingsongLi: Yes, there is unexpected behavior, but there is also user case to define options for all tables.

JingsongLi: Maybe not `scan.timestamp-millis`, but some options about execution.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/AbstractFlinkTableFactory.java
Aitozi: ok, will do a test

docs/content/engines/flink.md
Aitozi: > Yes, there is unexpected behavior, but there is also user case to define options for all tables.

if user want to define for all tables they can set `paimon.*.*.*.config_key=value`. This means all tables will take this config_key. if user want to set options under a database, they can set `paimon.*.db.*.config_key=value`. They just have to keep in mind that these three part controls the catalog, database, and table scope.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/AbstractFlinkTableFactory.java
Aitozi: OpenJDK 64-Bit Server VM 1.8.0_292-b10 on Mac OS X 10.16
Apple M1 Pro
test-pattern:                                                                                        Best/Avg Time(ms)    Row Rate(M/s)      Per Row(ns)   Relative
OPERATORTEST_test-pattern_pattern                                                                       12530 / 12779              0.0          41767.0       1.0X
OPERATORTEST_test-pattern_startswith                                                                       500 /  538              0.6           1666.2      25.1X

Per row will run for 100 key. so the max cost for 100 key match is 41767 ns. The startsWith is much faster, but I think both of the cost is acceptable, and the startWith check is not flexible when we only want to set  for the specific table

Aitozi:
Updated, Please take a look again @JingsongLi

JingsongLi (approved):
+1

PR #2105, opened 2023-10-10
[core] FileDeletionBase should cancel reading manifest files when any file is failed to read

yuzelin:
### Purpose

For example, if a data file is upgraded, it has an ADD entry and a DELETE entry. If it only reads the DELETE entry successfull, the upgraded data file will be deleted.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/FileDeletionBase.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/FileDeletionBase.java
@@ -41,6 +41,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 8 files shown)


paimon-core/src/main/java/org/apache/paimon/operation/FileDeletionBase.java
```diff
@@ -227,32 +222,17 @@ protected Iterable<ManifestEntry> tryReadDataManifestEntries(Snapshot snapshot)
...
-                                        } catch (Exception e) {
-                                            LOG.warn("Failed to read manifest file " + file, e);
-                                            return Collections.emptyList();
-                                        }
-                                    }
-                                });
+    /** Cancel reading when any file is failed to read. */
+    private List<ManifestEntry> readManifestEntries(List<ManifestFileMeta> manifestFileMetas) {
+        try {
+            return manifestFileMetas.stream()
```
JingsongLi, line 228: Here you break the lazy load for entries.

You can just cancel in the outer.

yuzelin: Refactored. 
1. for data skipping, if the read failed, we should skip the cleaning because skipping set is not complete, we may delete used data file;
2. if we should delete the data file in any case (like `deleteAddedDataFiles`),  we can just ignore unavailable data file.

PR #2106, opened 2023-10-10
[flink] Reset consumer action supports delete a consumer with a given consumer ID

zhuangchong:
…r ID.

### Purpose

reset consumer action supports delete a consumer with a given consumer ID

### Tests

ConsumerActionITCase#testResetConsumer


JingsongLi (approved):
+1

PR #2118, opened 2023-10-11
[core] Introduce file-reader-async-threshold to speed up merging

JingsongLi:
### Purpose

Currently, when there are too many files in a bucket, reading too many files can lead to poor overall performance, slow merging, and high query latency.

This PR aims to use asynchronous threads to read files and simultaneously read multiple files to accelerate the speed of merging.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/utils/AsyncRecordReader.java
+++ b/paimon-core/src/main/java/org/apache/paimon/utils/AsyncRecordReader.java
@@ -1,0 +1,76 @@
+package org.apache.paimon.utils;
+
+import org.apache.paimon.reader.RecordReader;
+import org.apache.paimon.reader.RecordReader.RecordIterator;
+
+import javax.annotation.Nullable;
+
+import java.io.IOException;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.LinkedBlockingQueue;
+
+/** A {@link RecordReader} to use ASYNC_EXECUTOR to read records async. */
+public class AsyncRecordReader<T> implements RecordReader<T> {
+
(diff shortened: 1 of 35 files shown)


leaves12138:
+1

paimon-core/src/main/java/org/apache/paimon/utils/AsyncRecordReader.java
```diff
@@ -0,0 +1,98 @@
...
+    private final Future<Void> future;
+
+    private boolean isEnd = false;
+
+    public AsyncRecordReader(IOExceptionSupplier<RecordReader<T>> supplier) {
+        this.queue = new LinkedBlockingQueue<>();
+        this.future =
+                ASYNC_EXECUTOR.submit(
+                        () -> {
+                            try (RecordReader<T> reader = supplier.get()) {
```
leaves12138, line 48: Maybe we can throw the exception, and announce the worker thread to stop to avoid worker thread blocked at queue.get() forever.

paimon-core/src/main/java/org/apache/paimon/utils/AsyncRecordReader.java
```diff
@@ -0,0 +1,98 @@
...
+
+    @Nullable
+    @Override
+    public RecordIterator<T> readBatch() throws IOException {
+        if (isEnd) {
+            return null;
+        }
+
+        try {
+            Element element = queue.take();
```
leaves12138, line 70: We may block here forever if something bad happens in the read thread

leaves12138:
java.lang.IllegalStateException: Trying to access closed classloader. Please check if you store classloaders directly or indirectly in static fields. If the stacktrace suggests that the leak occurs in a third party library and cannot be fixed immediately, you can disable this check with the configuration 'classloader.check-leaked-classloader'.

If reuse cached thread will throw this exception


image

leaves12138:
+1

PR #2124, opened 2023-10-12
[core] Move common code from FileSystemCatalog and HiveCatalog into AbstractCatalog

tsreaper:
### Purpose

Currently, `FileSystemCatalog` and `HiveCatalog` contains a lot of common code. For example, both catalog needs to check if the table to be deleted is a system table, and if that table exists.

This PR extracts common code into `AbstractCatalog`, which makes introducing new catalog easier.

### Tests

### API and Format

No.

### Documentation

No.


JingsongLi (approved):
+1

Issue #1797, opened 2023-08-11
[Feature] Register commit metrics to flink metric groups


schnappi17:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/incubator-paimon/issues) and found nothing similar.

### Motivation

Paimon metrics should be exposed outside, considering the flink metric system is mature in practice, we plan to register paimon metrics to flink metric groups.

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

PR #2131, opened 2023-10-16
[WIP] add writer metric

wg1026688210:
1.add writer metric logic
2.Adjust the metric repository from the process level to the table level to prevent memory leaks caused by unnecessary strong references.
#1797


wg1026688210:
@schnappi17 @JingsongLi @tsreaper PTAL

PR #2133, opened 2023-10-16
[flink] Correct the partitioning for multi-table CDC between write to commit

yuzelin:
### Purpose

Currently， the parallelism is set to 1 but actually the cdc records can be shuffled by table name. This PR fix it.


paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/sink/cdc/FlinkCdcMultiTableSink.java
```diff
@@ -116,15 +117,24 @@ public DataStreamSink<?> sinkFrom(
...
+                partitioned
+                        .transform(
+                                GLOBAL_COMMITTER_NAME,
+                                typeInfo,
+                                new CommitterOperator<>(
+                                        true,
+                                        commitUser,
+                                        createCommitterFactory(),
+                                        createCommittableStateManager()))
+                        .setParallelism(input.getParallelism());
```
JingsongLi, line 137: add a test for this.

JingsongLi (approved):
+1

PR #2134, opened 2023-10-16
[hive] Introduce metastore.tag-to-partition for Hive metastore

### LINKED ISSUE #2125: [Feature] Introduce metastore.tag-to-partition-field for Hive metastore
### Search before asking

### Motivation

When using primary key tables, an unpartitioned approach is often used to maintain updates, in order to mirror and synchronize tables from upstream database tables. This allows users to query the latest data.

However, the tradition of Hive data warehouses is not like this. Offline data warehouses require an immutable view every day to ensure the idempotence of calculations. So we created a Tag mechanism to output these views.

However, the traditional use of Hive data warehouses is more accustomed to using partitions to specify the query's Tag, and is more accustomed to using Hive computing engines.

So, we are considering mapping a non partitioned primary key table to the partition table in Hive metastore, and mapping the partition field to the name of the Tag to be fully compatible with Hive.

### Solution

Subtasks:
1. Introduce `metastore.tag-to-partition-field`, option type string, the field name for partition field, for Hive metastore, it will create a partition field to represent tag.
2. Flink & Spark and other engines can not see the partition field, because they use the schema in File System.
3. Before Hive engine query it, we should create Partition to metastore when we create Tag for the table.
    a. This requires Tag callback mechanism.
4. After partitions created, In `PaimonInputFormat.getSplits`, the locations will contains all partitions, we should convert these locations into Tags. And generate splits for tags.
5. In `PaimonInputFormat.getRecordReader`, we should generate partition field for the rows, so we should put tag information into `PaimonInputSplit` too.
6. This table can not be written from Hive compute engine.

### Anything else?

_No response_

JingsongLi:
### Purpose

Linked issue: close #2125

The Tag will maintain the manifests and data files of the snapshot.
A typical usage is creating tags daily, then you can maintain the historical data of each day for batch reading.

When using primary key tables, a non-partitioned approach is often used to maintain updates, in order to mirror and
synchronize tables from upstream database tables. This allows users to query the latest data. The tradition of Hive
data warehouses is not like this. Offline data warehouses require an immutable view every day to ensure the idempotence
of calculations. So we created a Tag mechanism to output these views.

However, the traditional use of Hive data warehouses is more accustomed to using partitions to specify the query's Tag,
and is more accustomed to using Hive computing engines.

So, we introduce `'metastore.tag-to-partition'` to mapping a non-partitioned primary key table to the partition table
in Hive metastore, and mapping the partition field to the name of the Tag to be fully compatible with Hive.

## Example

**Step 1: Create table and tag in Flink SQL**

```sql
CREATE CATALOG my_hive WITH (
    'type' = 'paimon',
    'metastore' = 'hive',
    'uri' = 'thrift://<hive-metastore-host-name>:<port>',
    -- 'hive-conf-dir' = '...', this is recommended in the kerberos environment
    -- 'hadoop-conf-dir' = '...', this is recommended in the kerberos environment
    'warehouse' = 'hdfs:///path/to/warehouse'
);

USE CATALOG my_hive;

CREATE TABLE mydb.T (
    pk INT,
    col1 STRING,
    col2 STRING
) WITH (
    'bucket' = '-1',
    'metastore.tag-to-partition' = 'dt'
);

INSERT INTO t VALUES (1, '10', '100'), (2, '20', '200');

-- create tag '2023-10-16' for snapshot 1
CALL my_hive.system.create_tag('mydb.T', '2023-10-16', 1);
```

**Step 2: Query table in Hive with Partition Pruning**

```sql
SHOW PARTITIONS T;
/*
OK
dt=2023-10-16
*/

SELECT * FROM T WHERE dt='2023-10-16';
/*
OK
1 10 100 2023-10-16
2 20 200 2023-10-16
*/
```

### DIFF
--- a/docs/content/migration/_index.md
+++ b/docs/content/migration/_index.md
@@ -1,10 +1,9 @@
 ---
-title: Project
-icon: <i class="fa fa-sitemap title maindish" aria-hidden="true"></i>
+title: Migration
+icon: <i class="fa fa-briefcase title maindish" aria-hidden="true"></i>
 bold: true
 bookCollapseSection: true
-sectionBreak: true
-weight: 9
+weight: 8
 ---
 <!--
 Licensed to the Apache Software Foundation (ASF) under one
--- a/docs/content/migration/upsert-to-partitioned.md
+++ b/docs/content/migration/upsert-to-partitioned.md
@@ -1,0 +1,99 @@
+---
+title: "Upsert To Partitioned"
+weight: 1
+type: docs
+aliases:
+- /migration/upsert-to-partitioned.html
+---
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+# Upsert To Partitioned
+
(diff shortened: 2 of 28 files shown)


JingsongLi:
Current question: How to query the data for the day? The tag creation has not been completed yet.

leaves12138:
+1

PR #2135, opened 2023-10-16
[flink][cdc] Add operator and transformation name for cdc pipeline

yuzelin:
### Purpose

A tiny refactor for better experience on Flink UI.


paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/sink/cdc/CdcDynamicTableParsingProcessFunction.java
```diff
@@ -100,7 +100,10 @@ public void processElement(T raw, Context context, Collector<Void> collector) th
                             try {
                                 catalog.createTable(identifier, schema, true);
                             } catch (Exception e) {
-                                LOG.error("create newly added paimon table error.", e);
+                                LOG.error(
+                                        "Cannot create newly added Paimon table "
+                                                + identifier.getFullName(),
+                                        e);
```
zhuangchong, line 106: LOG.error("Cannot create newly added Paimon table: {}.", identifier.getFullName(), e);  ?

yuzelin: Seems this method will consider the exception as the second string argument? 
```
@param format the format string
@param arg1   the first argument
@param arg2   the second argument
public void error(String format, Object arg1, Object arg2);
```

zhuangchong: slf4j will internally judge the last parameter.

image

https://github.com/qos-ch/slf4j/blob/5ff6f2c385c36ea4f8b85cacd69f7ca891c37818/slf4j-api/src/main/java/org/slf4j/helpers/AbstractLogger.java#L389-L395

yuzelin: That makes sense.

zhuangchong (approved):
+1

JingsongLi:
+1

PR #2140, opened 2023-10-17
[flink] Place all the built-in procedures into sys database

yuzelin:
### Purpose

We follow the behavior of [iceberg procedures](https://iceberg.apache.org/docs/1.3.1/spark-procedures/)


JingsongLi (approved):
+1

PR #2143, opened 2023-10-17
[ci] Separate Build and Test to reduce log output for test

JingsongLi:
### Purpose

Now, test outputs too more logs, it is difficult to find the root exception.


leaves12138:
+1

PR #2145, opened 2023-10-17
[flink] Refactor MergeIntoProcedure to make the api easier to use

yuzelin:
### Purpose

Without named argument, it is not easy to provide a simple merge_into procedure to users.
So we should simplify the API.

This PR only provide matched-upsert, matched-delete and not-matched-insert to users.


JingsongLi (approved):
+1

PR #2147, opened 2023-10-18
[core] Remove Changelog Value Count mode

JingsongLi:
### Purpose

```
CREATE TABLE T (
  i INT,
  j INT
) WITH (
  'write-mode' = 'changelog'
);
```

The table can accept +I -U +U -D all changelog inputs. Paimon will automatically maintain the LSM structure, updating, deleting, and counting according to the entire row.

But this mode has very poor performance, so it should be compared across the entire row.

We can delete this mode because no one use it.


yuzelin (approved):
+1

PR #2149, opened 2023-10-18
[flink] Remove cross-partition-upsert.bootstrap-min-partition

JingsongLi:
### Purpose

- `cross-partition-upsert.bootstrap-min-partition` is hard to be configured, actually we need a TTL instead of startup partition, a long running job should not rely on a static startup partition.
- now, use `cross-partition-upsert.index-ttl` to filter scan splits, this can just replace old `bootstrap-min-partition` feature.


yuzelin (approved):
+1

PR #1638, opened 2023-07-24
[core] Support commit metrics

schnappi17:
### Purpose

Linked issue: close #1468

PR #2150, opened 2023-10-18
[core] close commits in UTs

schnappi17:
### Purpose

Linked issue: subtask of #1638 

There are non-standard table commit creation in paimon UTs, we need to close the table commits after use in case of commit resources leak.


FangYongs:
Thanks @schnappi17 for contribution, LGTM

PR #2151, opened 2023-10-18
[flink] support partition push down

### LINKED ISSUE #2127: [Feature] Add support of partition push down for flink source
### Search before asking

### Motivation

Currently, Paimon Flink source do not implement `SupportsPartitionPushdown`, and rely on filter push down to prune the unrelated partition. However, this brings two optimization bad case:

- Currently, the paimon's `SupportsFilterPushDown` do not consume the filter expression after applyFilters.

```java
@Override
public Result applyFilters(List<ResolvedExpression> filters) {
    source.pushFilters(filters);
    return Result.of(filters, filters);
}
```
So in the case below, the limit can not be pushed into scan.

```sql
SELECT * FROM T where a = 1 limit 3
```

image

- Using filter push down to prune partition is not as effective as partition push down. For example, the filter condition '%pattern%' can not generate a valid partition predicate. And, in partition push down, some complex condition on partition field even function call condition is workable.

So, I proposal to support partition push down for paimon flink source, for more effective partition pruning.

### Solution

_No response_

### Anything else?

_No response_

Aitozi:
### Purpose

Linked issue: close #2127

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/DefaultValueAssigner.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/DefaultValueAssigner.java
@@ -133,17 +133,17 @@ public class DefaultValueAssigner {
                             return Optional.of(predicate);
                         };
 
-                ArrayList<Predicate> filterWithouDefaultValueField = new ArrayList<>();
+                ArrayList<Predicate> filterWithoutDefaultValueField = new ArrayList<>();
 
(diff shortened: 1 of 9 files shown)


Aitozi:
CC @JingsongLi

JingsongLi:
I still think we can have a better filter pushdown instead of introducing partition pushdown. Because in Paimon, list partitions is a heavy work.

Aitozi:
OK, I'll try to refactor this PR

PR #2152, opened 2023-10-19
[core] Introduce cross-partition-upsert.bootstrap-parallelism

JingsongLi:
### Purpose

Introduce `cross-partition-upsert.bootstrap-parallelism` for accelerate bootstrap.

When the job failover, it is necessary to read all indexes, which is a slow process, with hundreds of millions of data in several parallelism jobs that may take half an hour.

This PR uses the most violent way (parallel execution) to optimize bootstrap.


leaves12138:
+1

PR #2158, opened 2023-10-21
Add validation that append table of unware bucket mode do not support…

wg1026688210:
### Purpose

Currently , append table of unware bucket mode do not support compaction 'full-compaction.delta-commits'.add a validation when creating table

### Tests
AppendOnlyTableITCase#testCreateUnawareBucketTableWithFullCompaction


JingsongLi (approved):
+1

PR #2165, opened 2023-10-23
[hotfix] Flink procedures should pass the execution environment to action code

yuzelin:


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/procedure/CompactDatabaseProcedure.java
```diff
@@ -107,6 +107,7 @@ public String[] call(
                         .includingTables(nullable(includingTables))
                         .excludingTables(nullable(excludingTables))
                         .withDatabaseCompactMode(nullable(mode));
+        action.withStreamExecutionEnvironment(procedureContext.getExecutionEnvironment());
```
JingsongLi, line 110: Set env to `ActionBase` in `ProcedureBase` instead of every procedures.

JingsongLi (approved):
+1

PR #2167, opened 2023-10-23
[flink] add `drop partition`  for flink sql

zhangjun0x01:
### Purpose

add `drop partition`  for flink sql 

### Tests

org.apache.paimon.flink.CatalogTableITCase#testDropPartition

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
@@ -23,23 +23,28 @@ import org.apache.paimon.factories.FactoryUtil;
 import org.apache.paimon.fs.FileIO;
 import org.apache.paimon.fs.Path;
 import org.apache.paimon.lineage.LineageMetaFactory;
+import org.apache.paimon.operation.FileStoreCommit;
 import org.apache.paimon.operation.Lock;
 import org.apache.paimon.options.Options;
 import org.apache.paimon.schema.Schema;
 import org.apache.paimon.schema.SchemaChange;
 import org.apache.paimon.schema.TableSchema;
+import org.apache.paimon.table.AbstractFileStoreTable;
 import org.apache.paimon.table.CatalogEnvironment;
 import org.apache.paimon.table.FileStoreTable;
 import org.apache.paimon.table.FileStoreTableFactory;
 import org.apache.paimon.table.Table;
+import org.apache.paimon.table.sink.BatchWriteBuilder;
 import org.apache.paimon.table.system.SystemTableLoader;
 import org.apache.paimon.utils.StringUtils;
 
 import javax.annotation.Nullable;
 
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.UUID;
 import java.util.stream.Collectors;
 
 import static org.apache.paimon.options.CatalogOptions.LINEAGE_META;
@@ -105,6 +110,16 @@ public abstract class AbstractCatalog implements Catalog {
         }
 
         createDatabaseImpl(name);
+    }
+
+    @Override
+    public void dropPartition(Identifier identifier, Map<String, String> partitionSpec)
+            throws TableNotExistException {
+        Table table = getTable(identifier);
+        AbstractFileStoreTable fileStoreTable = (AbstractFileStoreTable) table;
+        FileStoreCommit commit = fileStoreTable.store().newCommit(UUID.randomUUID().toString());
+        commit.dropPartitions(
+                Collections.singletonList(partitionSpec), BatchWriteBuilder.COMMIT_IDENTIFIER);
     }
 
     protected abstract void createDatabaseImpl(String name);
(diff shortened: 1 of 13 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkCatalog.java
```diff
@@ -842,8 +851,23 @@ public final void createPartition(
     @Override
     public final void dropPartition(
             ObjectPath tablePath, CatalogPartitionSpec partitionSpec, boolean ignoreIfNotExists)
-            throws CatalogException {
-        throw new UnsupportedOperationException();
+            throws PartitionNotExistException, CatalogException {
```
zhuangchong, line 854: I understand that paimon catalog should be used to drop partition. For example, when using paimon hive catalog, hive metastore needs to be updated at the same time.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkCatalog.java
```diff
@@ -842,8 +851,23 @@ public final void createPartition(
...
+            Identifier identifier = toIdentifier(tablePath);
+            Table table = catalog.getTable(identifier);
+            AbstractFileStoreTable fileStoreTable = (AbstractFileStoreTable) table;
+            FileStoreCommit commit = fileStoreTable.store().newCommit(UUID.randomUUID().toString());
+            List<Map<String, String>> partitions =
+                    Collections.singletonList(partitionSpec.getPartitionSpec());
+            commit.dropPartitions(partitions, BatchWriteBuilder.COMMIT_IDENTIFIER);
+        } catch (Catalog.TableNotExistException e) {
+            throw new CatalogException(e);
+        }
```
zhuangchong, line 870: I understand that paimon catalog should be used to drop partition. For example, when using paimon hive catalog, hive metastore needs to be updated at the same time.

zhangjun0x01: when we create a paimon partitioned table with hive catalog , [we do not save the partitions info in the hive catalog](https://github.com/apache/incubator-paimon/blob/fa68a1e91ee77aeef0c0b99160cc093e56d2405a/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java#L455) , so when we drop table partition, we only need to modify the paimon metadata.

zhuangchong: It's handled here.


https://github.com/apache/incubator-paimon/blob/fa68a1e91ee77aeef0c0b99160cc093e56d2405a/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java#L479-L496

zhangjun0x01: yes， you are right, I ignored this parameter `METASTORE_PARTITIONED_TABLE`, I update it

paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
```diff
@@ -113,6 +118,16 @@ public void createDatabase(String name, boolean ignoreIfExists)
         createDatabaseImpl(name);
     }
 
+    @Override
+    public void dropPartition(Identifier identifier, Map<String, String> partitionSpec)
+            throws TableNotExistException, PartitionNotExistException {
+        Table table = getTable(identifier);
+        AbstractFileStoreTable fileStoreTable = (AbstractFileStoreTable) table;
+        FileStoreCommit commit = fileStoreTable.store().newCommit(UUID.randomUUID().toString());
+        commit.dropPartitions(
```
JingsongLi, line 127: It looks that we should delete partitions of metastore in `FileStoreCommit`.

JingsongLi:
I think maybe we should not delete partition for metastore, this can be finished by users.
Because we have time travel, the partitions are still visible.

zhangjun0x01:
> I think maybe we should not delete partition for metastore, this can be finished by users. Because we have time travel, the partitions are still visible.

I removed the code follow your suggestion.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkCatalog.java
```diff
@@ -826,7 +826,12 @@ public final CatalogPartition getPartition(
     @Override
     public final boolean partitionExists(ObjectPath tablePath, CatalogPartitionSpec partitionSpec)
             throws CatalogException {
-        return false;
+        try {
+            List<CatalogPartitionSpec> partitionSpecs = getPartitionSpecs(tablePath, partitionSpec);
```
JingsongLi, line 830: Here we need an effective method to check partition. For example, we can provide a `SnapshotReaderImpl.withPartitionFilter` to optimize scanning.

zhangjun0x01: I updated it

yuzelin:
Resolve conflict.

JingsongLi (approved):
+1

PR #2169, opened 2023-10-23
[core] Refactor metrics to remove Metrics singleton

tsreaper:
### Purpose

Currently, Paimon's metrics system records all metrics in the `Metrics` singleton. With this singleton, we have to carefully manage the creation and deletion of metrics, or it will result in resource leaks.

This PR refactors the metrics system to remove this Metrics singleton. The life cycle of Paimon's metrics will be entirely managed by the metrics system of computation engines like Flink and Spark. Paimon only needs to concern about how to calculate these metrics.

### Tests

Existing tests should cover the changes.

### API and Format

No.

### Documentation

No.


paimon-core/src/main/java/org/apache/paimon/table/sink/TableCommitImpl.java
```diff
@@ -129,6 +131,12 @@ public TableCommitImpl ignoreEmptyCommit(boolean ignoreEmptyCommit) {
         return this;
     }
 
+    @Override
+    public InnerTableCommit withMetricRegistry(AbstractMetricRegistry registry) {
+        commit.withMetrics(new CommitMetrics(registry, tableName));
+        return this;
+    }
+
```
zhuangchong, line 139: I didn't find where to call this method `withMetricRegistry`

tsreaper: This method is called from Flink / Spark operator. We'll introduce a `FlinkMetricRegistry` in the next PR.

paimon-core/src/main/java/org/apache/paimon/metrics/AbstractMetricRegistry.java
```diff
@@ -0,0 +1,50 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.paimon.metrics;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+/** Factory to create {@link MetricGroup}s. */
+public abstract class AbstractMetricRegistry {
```
JingsongLi, line 25: Just name this `MetricRegistry`?

JingsongLi (commented):
CC @schnappi17

paimon-core/src/main/java/org/apache/paimon/metrics/MetricGroup.java
```diff
@@ -56,28 +46,19 @@ public interface MetricGroup {
      * @param <T> return type of the gauge
      * @return the given gauge
      */
-    <T, G extends Gauge<T>> G gauge(String name, G gauge);
+    <T> Gauge<T> gauge(String name, Gauge<T> gauge);
```
schnappi17, line 49: So do we need a wrapper to wrap the PAIMON basic metric types to FLINK basic metrics types? Because finally we need to register FLINK(or SPARK) Metric to FLINK metric groups.

paimon-core/src/main/java/org/apache/paimon/metrics/MetricGroupImpl.java
```diff
@@ -0,0 +1,122 @@
...
+/** Default implementation of {@link MetricGroup}. */
+public class MetricGroupImpl implements MetricGroup {
+
+    private static final Logger LOG = LoggerFactory.getLogger(MetricGroupImpl.class);
+
+    private final String groupName;
+    private final Map<String, String> variables;
+    private final Map<String, Metric> metrics;
+
+    public MetricGroupImpl(String groupName) {
```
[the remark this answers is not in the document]
tsreaper: Yes. This is the default implementation of `MetricGroup`. Java API users may want to pass this into `InnerTableCommit` to fetch metrics. The usage of this class will be given in the documents.

paimon-core/src/main/java/org/apache/paimon/metrics/MetricGroup.java
tsreaper: Yes. Flink support will come in the next PR.

JingsongLi (approved):
+1

PR #2173, opened 2023-10-24
[hive] Supports modifying hive table parameters when using hive catalog

zhuangchong:
### Purpose

Supports modifying hive table parameters when using hive catalog

### DIFF
--- a/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
+++ b/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
@@ -293,17 +293,7 @@ public class HiveCatalog extends AbstractCatalog {
                     e);
         }
 
-        Table table =
-                newHmsTable(
-                        identifier,
-                        tableSchema.options().entrySet().stream()
-                                .filter(entry -> entry.getKey().startsWith(HIVE_PREFIX))
-                                .collect(
-                                        Collectors.toMap(
-                                                entry ->
-                                                        entry.getKey()
-                                                                .substring(HIVE_PREFIX_LENGTH),
-                                                Map.Entry::getValue)));
+        Table table = newHmsTable(identifier);
         try {
             updateHmsTable(table, identifier, tableSchema);
             client.createTable(table);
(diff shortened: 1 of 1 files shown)


JingsongLi:
I will fix the unstable test in https://github.com/apache/incubator-paimon/pull/2174

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
```diff
@@ -502,6 +485,24 @@ private void updateHmsTable(Table table, Identifier identifier, TableSchema sche
...
+        // set table parameters
+        Map<String, String> parameters =
+                schema.options().entrySet().stream()
+                        .filter(entry -> entry.getKey().startsWith(HIVE_PREFIX))
+                        .collect(
+                                Collectors.toMap(
+                                        entry -> entry.getKey().substring(HIVE_PREFIX_LENGTH),
+                                        Map.Entry::getValue,
+                                        (v1, v2) -> v2));
+        table.setParameters(parameters);
```
JingsongLi, line 497: How about previous parameters in the table?

JingsongLi:
Can you explain what is this used to?

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
zhuangchong: Already modified, keep the previous parameters when modifying the table.

zhuangchong:
> Can you explain what is this used to?

Currently, hive catalog supports adding hive table parameters when creating a table, but does not support adding or modifying hive table parameters when modifying the table.

JingsongLi:
> > Can you explain what is this used to?
> 
> Currently, hive catalog supports adding hive table parameters when creating a table, but does not support adding or modifying hive table parameters when modifying the table.

Is this come from some user case?

zhuangchong:
> > > Can you explain what is this used to?
> > 
> > 
> > Currently, hive catalog supports adding hive table parameters when creating a table, but does not support adding or modifying hive table parameters when modifying the table.
> 
> Is this come from some user case?

nothing now,  I understand that it is necessary to modify the table.

For example: 
1.modify original  table parameters: table leader, calculation period, usage status, subject field, etc.
1.add new table parameters: table leader, calculation period, usage status, subject field, etc.

create table case: https://github.com/apache/incubator-paimon/issues/1983

PR #2180, opened 2023-10-26
[core] Add validation to forbid setting stream-read-overwrite with 'full-compaction' or 'lookup' changelog producer

yuzelin:
### Purpose

This is because `full-compaction` and `lookup` changelog producers will output the changes in later compact snapshot. If we read the changes of the overwrite snapshot, the output is duplicated.


paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
```diff
@@ -87,6 +88,16 @@ public static void validateTableSchema(TableSchema schema) {
                             "Can not set %s on table without primary keys, please define primary keys.",
                             CHANGELOG_PRODUCER.key()));
         }
+        if (options.streamingReadOverwrite()
+                && (changelogProducer == ChangelogProducer.FULL_COMPACTION
+                        || changelogProducer == ChangelogProducer.LOOKUP)) {
+            throw new UnsupportedOperationException(
+                    String.format(
+                            "Cannot set %s to true when changelog producer is %s or %s.",
```
JingsongLi, line 96: Add why, and add a test in `CatalogTableITCase`.

JingsongLi (approved):
+1

PR #2184, opened 2023-10-27
[core] Bulk load records for GlobalIndexAssigner when index is empty

JingsongLi:
### Purpose

When there is no index, and it is a batch job, we don't need to use rocksdb, we can just use sort to bulk load records.


Zouxxyy:
+1

PR #2186, opened 2023-10-27
[core] Move codes of cross partition to paimon-core

JingsongLi:
### Purpose

These codes should be reused by paimon-spark.


Zouxxyy:
+1

PR #2187, opened 2023-10-27
[core]Show the latest modification time of a specified partition.

wg1026688210:
Show the latest modification time of a specified partition.

for example:
A table has 3 partitions: dt, hour, and eventid, but the modify time of only one specified parition dt should be show

2023-10-30/22/click, modifyTime:100
2023-10-30/23/show,modifyTime:101    
2023-10-31/00/show,modifyTime:102

tranfer to

2023-10-30,modifyTime:101 
2023-10-31,modifyTime:102

### Tests
FileStoreTableTestBase#testMultiParitionModifyTime


JingsongLi:
You can use `SELECT MAX(update_time) FROM T$files WHERE partition='xxxx'` to compute update time.

PR #2188, opened 2023-10-27
[core] Force creating snapshot for tag with auto process time creation

### LINKED ISSUE #1942: [Feature] auto create new tag for flink checkpoint

JingsongLi:
### Purpose

Fixes #1942

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/manifest/WrappedManifestCommittable.java
+++ b/paimon-core/src/main/java/org/apache/paimon/manifest/WrappedManifestCommittable.java
@@ -28,13 +28,42 @@ import java.util.TreeMap;
 /** Manifest commit message. */
 public class WrappedManifestCommittable {
 
-    private Map<Identifier, ManifestCommittable> manifestCommittables;
+    private final long checkpointId;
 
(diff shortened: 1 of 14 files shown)


yuzelin (approved):
+1

PR #2204, opened 2023-10-30
[core] Add validation for sequence group during create table

### LINKED ISSUE #2203: [Feature] Add validation for sequence group during create table
### Search before asking

### Motivation

Currently, the sequence group options is validated during the runtime. Some checks should be moved to the SchemaValidation phase.

### Solution

_No response_

### Anything else?

_No response_

Aitozi:
### Purpose

Linked issue: close #2203

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
+++ b/paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
@@ -37,13 +37,16 @@ import org.apache.paimon.types.VarCharType;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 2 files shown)


paimon-core/src/main/java/org/apache/paimon/mergetree/compact/PartialUpdateMergeFunction.java
```diff
@@ -205,27 +205,8 @@ private Factory(Options options, RowType rowType) {
                     SequenceGenerator sequenceGen =
                             new SequenceGenerator(sequenceFieldName, rowType);
                     Arrays.stream(v.split(","))
-                            .map(
```
JingsongLi, line 208: We don't need to delete here, we can keep it as it is.

JingsongLi (commented):
Thanks @Aitozi left one comment

paimon-core/src/main/java/org/apache/paimon/mergetree/compact/PartialUpdateMergeFunction.java
Aitozi: Make sense, updated.

JingsongLi (approved):
+1

PR #2207, opened 2023-10-30
[doc] Fixed an error with the Hadoop configuration in presto and trino

huyuanfeng2018:
Fixed an error with the Hadoop configuration in presto and trino

### Purpose

Linked issue: close #xxx


JingsongLi (approved):
Thanks @huyuanfeng2018

PR #2212, opened 2023-10-31
[spark] Support cross partition write

Zouxxyy:
### Purpose

spark support `Cross Partitions Upsert Dynamic Bucket Mode`

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/crosspartition/KeyPartOrRow.java
+++ b/paimon-core/src/main/java/org/apache/paimon/crosspartition/KeyPartOrRow.java
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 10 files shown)


waywtdcc:
How is this PR going?

Zouxxyy:
@waywtdcc cross partition write needs to be refactored, will rebase after refactoring

waywtdcc:
Now, How is this PR going?

Zouxxyy:
@waywtdcc sorry for delay,  we choose to  closed it, The current implementation is friendly to streaming scenarios, but the batch cost is too high. We need to find a more efficient implementation in batch scenarios.

waywtdcc:
Doesn't flink batch also support it? The logic of the two should be the same, right?

PR #2218, opened 2023-10-31
[flink] exclude dependency of protobuf

leaves12138:
### Purpose

To run test in idea, we need to exclude the low version of protobuf. Otherwise, idea will will the low version protobuf (it can't recognize shade)


JingsongLi (approved):
+1

PR #2221, opened 2023-10-31
[flink] support the feature of ContinuousFileStoreSource using consumer-id.

### LINKED ISSUE #2157: [Feature] Support consumer-id for ContinuousFileStoreSource.
### Search before asking

### Motivation

After setting the `consumer-id`, the `MonitorFunction` that implements `RichSourceFunction` will be used as the `source`. However, this `RichSourceFunction` does not support new features introduced after flink [Flip-27](https://cwiki.apache.org/confluence/display/FLINK/FLIP-27%3A+Refactor+Source+Interface), such as watermark alignment. 

At the same time, in the existing implementation, if the user wants to consume in `xxx_FULL` mode, he needs to wait for the full phase consumption to be completed before making a checkpoint. During this period, if the job failover, it will cause full consumption again.

### Solution

`ContinuousFileStoreSource` is implemented using the Flip-27 interface, which can be extended to support `consumer-id`.

1. The splits will be allocated strictly in the order of snapshots. The `reader` will request the next splits after consuming the allocated splits. This means that the snapshots smaller than split in the reader instance have been consumed;
2. `Enumerator` collects the snapshot id of the split currently assigned to each `reader`, and takes min (the snapshot-id currently being processed by all readers) as the `nextSnapshotId` of the `consumer`;
3. When the checkpoint completed, notify `StreamTableScan` to update the `consumer`.

### Anything else?

_No response_

liming30:
### Purpose

Linked issue: close #2157 

[flink] support the feature of ContinuousFileStoreSource using consumer-id.

`Enumerator` will assign split to the `reader` in the order of snapshots, and the `reader` will also consume in the order of snapshots. Therefore, the consumption progress of a single reader is:

1. If the reader is in the `enumerator's` waiting list, then the consumption progress of the `reader` is equal to the snapshot assigned by `splitAssigner` next time. If there is no split in `splitAssigner`, it is equal to the snapshot of the next scan by enumerator.
2. If the reader is not in the `enumerator's` waiting list, the `reader's` consumption progress is equal to the greater of the snapshot consumed by the consumer and the snapshot recently assigned to it by `splitAssigner`.

The minimum value of the consumption progress of all readers is the current position of the `consumer`.

### Tests

1. `org.apache.paimon.flink.source.ContinuousFileSplitEnumeratorTest#testEnumeratorWithConsumer`
2. `org.apache.paimon.flink.source.FileStoreSourceReaderTest#testReaderOnSplitFinished`

### API and Format

No
### Documentation

No

### DIFF
--- a/docs/content/how-to/querying-tables.md
+++ b/docs/content/how-to/querying-tables.md
@@ -332,6 +332,17 @@ NOTE: The consumer will prevent expiration of the snapshot. You can specify 'con
 lifetime of consumers.
 {{< /hint >}}
 
+You can set `consumer.use-legacy-mode` to `false` to use the consumer function implemented based on 
+[flip-27](https://cwiki.apache.org/confluence/display/FLINK/FLIP-27%3A+Refactor+Source+Interface), which can provide 
+more capabilities, such as watermark alignment.
+
+{{< hint warning >}}
+1. When there is no watermark definition, the consumer in non-legacy mode cannot provide the ability to pass the 
+watermark in the snapshot to the downstream. 
+2. Since the implementation of legacy mode and non-legacy mode are completely different, the state of flink is 
+incompatible and cannot be restored from the state when switching modes.
+{{< /hint >}}
+
 You can reset a consumer with a given consumer ID and next snapshot ID and delete a consumer with a given consumer ID.
 
 {{< hint info >}}
--- a/docs/content/how-to/querying-tables.md
+++ b/docs/content/how-to/querying-tables.md
@@ -332,15 +332,16 @@ NOTE: The consumer will prevent expiration of the snapshot. You can specify 'con
 lifetime of consumers.
 {{< /hint >}}
 
-You can set `consumer.use-legacy-mode` to `false` to use the consumer function implemented based on 
-[flip-27](https://cwiki.apache.org/confluence/display/FLINK/FLIP-27%3A+Refactor+Source+Interface), which can provide 
-more capabilities, such as watermark alignment.
+By default, the consumer uses `exactly-once` mode to record consumption progress, which strictly ensures that what is 
+recorded in the consumer is the snapshot-id + 1 that all readers have exactly consumed. You can set `consumer.mode` to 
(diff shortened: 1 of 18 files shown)


JingsongLi (changes requested):
Thanks @liming30 
I have two comments:
1. Option can be `consumer.mode`: [EXACTLY_ONCE, AT_LEAST_ONCE]
2. Extract a util class from `ContinuousFileSplitEnumerator`.

liming30:
> Thanks @liming30 I have two comments:
> 
> 1. Option can be `consumer.mode`: [EXACTLY_ONCE, AT_LEAST_ONCE]
> 2. Extract a util class from `ContinuousFileSplitEnumerator`.

@JingsongLi I have addressed this comment. Please help review it if you have time.

scottxing:
Paimon version 0.7, when a large amount of data passes through paimon cdc, about 100 million records are dropped to the paimon ods table. The table attribute sets changelog as input. Then, at this time, I write a flink sql job (using the consumer-id setting), and read This table is inserted into another paimon dwd table (the changelog attribute is lookup). After starting this job, the checkpoint has been stuck at 0% and cannot be completed, so the snapshot cannot be committed. As a result, my other flink sql job cannot check the paimon dwd table. to data. This leads to the phenomenon that a large amount of data from one paimon table must be completely written to another paimon table before it can then be transferred from this paimon table to the next. Data cannot flow smoothly from job to job like a stream. How to solve it?

scottxing:
@JingsongLi @liming30  Paimon version 0.7, when a large amount of data passes through paimon cdc, about 100 million records are dropped to the paimon ods table. The table attribute sets changelog as input. Then, at this time, I write a flink sql job (using the consumer-id setting), and read This table is inserted into another paimon dwd table (the changelog attribute is lookup). After starting this job, the checkpoint has been stuck at 0% and cannot be completed, so the snapshot cannot be committed. As a result, my other flink sql job cannot check the paimon dwd table. to data. This leads to the phenomenon that a large amount of data from one paimon table must be completely written to another paimon table before it can then be transferred from this paimon table to the next. Data cannot flow smoothly from job to job like a stream. How to solve it?

PR #2234, opened 2023-11-01
[cdc] Fix NPE in obtaining metadata column op_ts value in mysql cdc action.

### LINKED ISSUE #2182: When using the MySQL CDC synchronization tool, I found that the --metadata-column parameter, specifically op_ts, cannot be properly converted from a timestamp to TIMESTAMP_LTZ(3).
### Search before asking

### Paimon version

0.6-SNAPSHOT

### Compute Engine

flink: 1.17.1
flink cdc: 2.3

### Minimal reproduce step

> 1：use paimon-flink-action-0.6-SNAPSHOT.jar

> 2： use param --metadata-column op_ts,database_name,table_name

> 3：I've noticed that there's no tm_ms field in the org.apache.paimon.flink.action.cdc.mysql.format.DebeziumEvent file under the public class Source. I added this field, thinking I had solved the issue. 

> 4：However, during the usage, I found that the op_ts field is using the org.apache.paimon.utils.TypeUtils utility for conversion, and I traced it to the castFromStringInternal method. It appears that TIMESTAMP_WITHOUT_TIME_ZONE cannot properly convert the timestamp.

**I would like the op_ts field to be written as either a timestamp or a bigint type.**

### Anything else?

**1：Why isn't the data type of the op_ts field bigint?**

zhuangchong:
### Purpose

Linked issue: close #2182 

Fix NPE in obtaining metadata column `op_ts` value in mysql cdc action


paimon-flink/paimon-flink-cdc/src/test/java/org/apache/paimon/flink/action/cdc/mysql/MySqlSyncTableActionITCase.java
```diff
@@ -922,13 +922,14 @@ public void testMetadataColumns() throws Exception {
...
+                            DataTypes.TIMESTAMP_WITH_LOCAL_TIME_ZONE(3).notNull()
                         },
-                        new String[] {"pk", "_date", "table_name", "database_name"});
+                        new String[] {"pk", "_date", "table_name", "database_name", "op_ts"});
         waitForResult(
                 Arrays.asList(
-                        "+I[1, 2023-07-30, test_metadata_columns, metadata]",
-                        "+I[2, 2023-07-30, test_metadata_columns, metadata]"),
+                        "+I[1, 2023-07-30, test_metadata_columns, metadata, 1970-01-01T00:00]",
+                        "+I[2, 2023-07-30, test_metadata_columns, metadata, 1970-01-01T00:00]"),
```
yuzelin, line 932: `op_ts` is always '0' when reading from snapshot so the test result is `1970-01-01T00:00`. Am I right? If so, please add comment to explain it (and you can add this [link](https://ververica.github.io/flink-cdc-connectors/master/content/connectors/mysql-cdc.html#available-metadata))

zhuangchong: This is it, comments have been added.

yuzelin (approved):
+1

PR #2239, opened 2023-11-01
[core] [bug] fix InMemoryBuffer preempt loop

leaves12138:
`InMemoryBuffer` put method may cause loop preempt memory, and finally it will preempt itself.

And flush should never ask for memory.

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/data/SimpleCollectingOutputView.java
+++ b/paimon-common/src/main/java/org/apache/paimon/data/SimpleCollectingOutputView.java
@@ -52,7 +52,6 @@ public class SimpleCollectingOutputView extends AbstractPagedOutputView {
         this.fullSegments = fullSegmentTarget;
         this.memorySource = memSource;
         this.fullSegments.add(getCurrentSegment());
-        segmentNum = 0;
     }
 
     public void reset() {
--- a/paimon-common/src/main/java/org/apache/paimon/data/SimpleCollectingOutputView.java
+++ b/paimon-common/src/main/java/org/apache/paimon/data/SimpleCollectingOutputView.java
@@ -52,6 +52,7 @@ public class SimpleCollectingOutputView extends AbstractPagedOutputView {
         this.fullSegments = fullSegmentTarget;
         this.memorySource = memSource;
         this.fullSegments.add(getCurrentSegment());
+        segmentNum = 0;
     }
 
     public void reset() {
--- a/paimon-core/src/main/java/org/apache/paimon/disk/InMemoryBuffer.java
+++ b/paimon-core/src/main/java/org/apache/paimon/disk/InMemoryBuffer.java
@@ -23,6 +23,7 @@ import org.apache.paimon.data.InternalRow;
 import org.apache.paimon.data.RandomAccessInputView;
 import org.apache.paimon.data.SimpleCollectingOutputView;
 import org.apache.paimon.data.serializer.AbstractRowDataSerializer;
+import org.apache.paimon.data.serializer.InternalRowSerializer;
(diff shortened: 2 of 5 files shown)


paimon-core/src/main/java/org/apache/paimon/disk/InMemoryBuffer.java
```diff
@@ -110,7 +115,9 @@ public long memoryOccupancy() {
 
     @Override
     public InMemoryBufferIterator newIterator() {
-        tryInitialize();
+        if (!isInitialized) {
+            return EMPTY_ITERATOR;
```
JingsongLi, line 119: Add comment: to avoid request memory.

paimon-core/src/main/java/org/apache/paimon/disk/InMemoryBuffer.java
```diff
@@ -126,6 +133,9 @@ long getCurrentDataBufferOffset() {
     }
 
     int getNumRecordBuffers() {
+        if (currentDataBufferOffset < 0) {
```
JingsongLi, line 136: if (!isInitialized) {
   return 0;
}

paimon-core/src/main/java/org/apache/paimon/disk/InMemoryBuffer.java
```diff
@@ -190,4 +200,33 @@ public BinaryRow next() throws IOException {
...
+        }
+
+        @Override
+        public boolean advanceNext() {
+            return false;
+        }
+
+        @Override
+        public BinaryRow getRow() {
+            return null;
```
JingsongLi, line 219: throw new UnsupportedException

paimon-core/src/main/java/org/apache/paimon/disk/InMemoryBuffer.java
```diff
@@ -68,7 +72,8 @@ private void tryInitialize() {
     @Override
     public void reset() {
         if (this.isInitialized) {
-            this.currentDataBufferOffset = 0;
+            // we need to set -1 here to avoid been preempt
+            this.currentDataBufferOffset = -1;
```
JingsongLi, line 76: I don't know why to set to -1.

I think we should modify `MemoryPoolFactory.preemptMemory`, if (memoryOccupancy > 0) assign to max.

JingsongLi (changes requested):

paimon-common/src/main/java/org/apache/paimon/data/SimpleCollectingOutputView.java
```diff
@@ -52,6 +52,7 @@ public SimpleCollectingOutputView(
         this.fullSegments = fullSegmentTarget;
         this.memorySource = memSource;
         this.fullSegments.add(getCurrentSegment());
+        segmentNum = 0;
```
JingsongLi, line 55: revert this.

paimon-core/src/test/java/org/apache/paimon/disk/InMemoryBufferTest.java
```diff
@@ -143,4 +147,70 @@ public void testEmpty() throws Exception {
         RowBufferIterator iterator = buffer.newIterator();
         assertThat(iterator.advanceNext()).isFalse();
     }
+
+    @Test
+    public void testMemoryPoolWorksWellWithInMemoryBuffer() throws Exception {
```
JingsongLi, line 152: There are lots of code style issues, you can check your ide support.

leaves12138: done

JingsongLi (approved):
+1

PR #2244, opened 2023-11-02
[core] Add support of partial update with aggregate function

### LINKED ISSUE #2242: [Feature] Support to specify the aggregate fields for the partial-update merge engine
### Search before asking

### Motivation

Currently, the aggregation and partial-upadate are individual merge engine. And the partial-update will use `last_value_not_null`  
 aggregation semantic for the field not in sequence group, and `last_value` for the field in the sequence group.

We can also support user set the aggregation function for the fields in partial-update merge engine. 
This will benefit the case:  
When merging multiple rows for the primary key into a wide table, user can set the aggregate strategy for the different field eg: sum/first_value/max ... 

### Solution

Introduce fieldAggregators for the `PartialUpdateMergeFunction` and unify the current merge solution

### Anything else?

_No response_

### LINKED ISSUE #2248: [Feature] Add first_value aggregate function
### Search before asking

### Motivation

To support keep the first value of the column

### Solution

_No response_

### Anything else?

_No response_

Aitozi:
### Purpose

Linked issue: close #2242
Linked issue: close #2248


paimon-core/src/main/java/org/apache/paimon/mergetree/compact/PartialUpdateMergeFunction.java
```diff
@@ -106,36 +116,23 @@ public void add(KeyValue kv) {
...
-                    row.setField(i, field);
-                }
+                row.setField(i, aggregator.agg(accumulator, field));
             } else {
                 Long currentSeq = sequenceGen.generateNullable(kv.value());
                 if (currentSeq != null) {
                     Long previousSeq = sequenceGen.generateNullable(row);
                     if (previousSeq == null || currentSeq >= previousSeq) {
-                        row.setField(i, field);
+                        row.setField(i, aggregator.agg(accumulator, field));
```
JingsongLi, line 135: This is strange, we may introduce `FieldAggregator.aggForOldSequence(...)`, for example, sum should be also agg if it is a late record?

There should be tests..

paimon-core/src/main/java/org/apache/paimon/mergetree/compact/PartialUpdateMergeFunction.java
```diff
@@ -155,7 +152,17 @@ private void retractWithSequenceGroup(KeyValue kv) {
                             row.setField(i, getters[i].getFieldOrNull(kv.value()));
                         } else {
                             // retract normal field
-                            row.setField(i, null);
+                            if (kv.valueKind() == RowKind.UPDATE_BEFORE) {
```
JingsongLi, line 155: We should retract for delete too.

JingsongLi (changes requested):

paimon-core/src/main/java/org/apache/paimon/mergetree/compact/PartialUpdateMergeFunction.java
Aitozi: You are right, will fix this

paimon-core/src/main/java/org/apache/paimon/mergetree/compact/PartialUpdateMergeFunction.java
Aitozi: I think only agg#retract is not enough, I think we should have 3 strategy for the retract record (DELETE/UPDATE_BEFORE).

And each aggregate function will be assigned with a retract strategy.

- `ignore_retract`  if user enabled `fields.${filed}.ignore-retract`
- `set_null`  for the last_value / last_not_null_value  user can use this to handle the retracted value to set null. This should be default for the field which is not declared as an agg field in partial-update for compatibility
- default call `agg#retract`

What do you think @JingsongLi

JingsongLi: Yes, correct

paimon-core/src/main/java/org/apache/paimon/mergetree/compact/PartialUpdateMergeFunction.java
macdoor: how about first_value / first_not_null_value?

Aitozi: This will be done in the `first_value` agg function. For the `first_value`,  when an older sequence record in , it should updated to the new value.

Aitozi:
This PR also add support of the first_value aggregation function. And will update the  related doc in a following PR. Please take a look  again. @JingsongLi

JingsongLi:
Thanks @Aitozi , I will take a review next week.

paimon-core/src/main/java/org/apache/paimon/mergetree/compact/PartialUpdateMergeFunction.java
```diff
@@ -300,5 +320,84 @@ public AdjustedProjection adjustProjection(@Nullable int[][] projection) {
...
+            List<DataType> fieldTypes = rowType.getFieldTypes();
+            Map<Integer, FieldAggregator> fieldAggregators = new HashMap<>();
+            for (int i = 0; i < fieldNames.size(); i++) {
+                String fieldName = fieldNames.get(i);
+                DataType fieldType = fieldTypes.get(i);
+                // aggregate by primary keys, so they do not aggregate
+                boolean isPrimaryKey = primaryKeys.contains(fieldName);
+                String strAggFunc =
+                        options.get(
+                                key(FIELDS_PREFIX + "." + fieldName + "." + AGG_FUNCTION)
```
JingsongLi, line 341: Maybe extract methods here?

JingsongLi (changes requested):
Can you keep this PR simple? Just support agg for partial-update.

Multiplex things here make me confused.

paimon-core/src/main/java/org/apache/paimon/mergetree/compact/PartialUpdateMergeFunction.java
Aitozi: Updated

Aitozi:
> Can you keep this PR simple? Just support agg for partial-update.
> 
> Multiplex things here make me confused.

I've reset the commit to only keep the commits of https://github.com/apache/incubator-paimon/issues/2242

macdoor:
please keep first_value

macdoor:
got it. first_value pr will be submitted seperately

JingsongLi:
I took a quick look at it, I think the logic is correct now.

But I still have some concerns about the codes abstraction, now everything is rely on Agg function. What I see both partial-update and agg become more complicated.

Aitozi:
> now everything is rely on Agg function

Thanks for your comments, I think the previous behavior of the partial-update  have two behavior:

- in sequence group.  it means a lastValue agg
- not in sequence group. it means a last_not_null agg

So a unified solution with fieldAggregators looks more consolidated. 

But, it will brings some more function call than before, do you care about this ?

JingsongLi:
I think we can have some concept first:
1. User only need to use both sequenceGroup and agg.
2. Sequence field should not have agg func.
3. In `updateWithSequenceGroup`:
```
if (previousSeq == null || currentSeq >= previousSeq) {
     row.setField(i, aggregator == null ? field : aggregator.agg(getField(i), field));
} else if (aggregator != null) {
     row.setField(i, aggregator.agg(field, getField(i)));
}
```
Something like this.
4. For `retractWithSequenceGroup`, maybe we should do something similiar.

JingsongLi (approved):
+1

PR #2247, opened 2023-11-02
[flink] flink can not streaming read deleted data

zhangjun0x01:
### Purpose

Linked issue: close (https://github.com/apache/incubator-paimon/issues/2246)


zhangjun0x01:
When delete predicate is partition keys, there are still issues, but I think it's better  to handle the partition issues in another PR uniformly , including 
`delete from table where partitionKey = '111' `
`alter table drop partition( partitionKey = '111') `
`drop partition action`

PR #2254, opened 2023-11-03
[core] Sort the results of the `show tables` command

zhangjun0x01:
### Purpose

When there are hundreds of tables in the database,  and the results of the `show tables` command are not sorted, we need to search one by one to look up if a table exists.

If the results are sorted, we can quickly locate it based on the dictionary order.


yuzelin (approved):
+1

JingsongLi (approved):
+1

PR #2255, opened 2023-11-03
[flink][cdc] Register MySQL JDBC driver when constructing cdc action

yuzelin:
### Purpose

Driver should be registered before get JDBC connection when using streampark to submit jobs.


zhuangchong (approved):
LGTM.

PR #2262, opened 2023-11-05
[core] Fix auto process time create snapshot for tag

MonsterChenzhuo:
### Purpose

https://github.com/apache/incubator-paimon/pull/2188/files
I conducted relevant tests locally using MySQL, and here are the screenshots:
企业微信截图_9d6c3269-8582-46cf-930a-09bc4d0f7e96
There is an issue that arises, for instance, when I set the tag creation to once a day and checkpoints to occur every 2 minutes. This results in a large number of snapshot and manifest files being generated. I believe there is room for optimization. Each time a snapshot is forcefully created, it should be checked whether it coincides with the next scheduled tag creation time. If it does, proceed with the snapshot creation; otherwise, skip it.

The following image is a screenshot taken after conducting tests with the logical corrections applied:
图片

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/sink/TableCommitImpl.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/sink/TableCommitImpl.java
@@ -120,7 +120,9 @@ public class TableCommitImpl implements InnerTableCommit {
     }
 
     public boolean forceCreatingSnapshot() {
-        return tagAutoCreation != null && tagAutoCreation.forceCreatingSnapshot();
+        return tagAutoCreation != null
+                && tagAutoCreation.canProceedWithNextTag()
+                && tagAutoCreation.forceCreatingSnapshot();
     }
 
     @Override
--- a/paimon-core/src/main/java/org/apache/paimon/table/sink/TableCommitImpl.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/sink/TableCommitImpl.java
@@ -120,9 +120,7 @@ public class TableCommitImpl implements InnerTableCommit {
     }
 
     public boolean forceCreatingSnapshot() {
-        return tagAutoCreation != null
-                && tagAutoCreation.canProceedWithNextTag()
-                && tagAutoCreation.forceCreatingSnapshot();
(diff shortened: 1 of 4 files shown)


leaves12138 (changes requested):

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/CommitterOperator.java
```diff
@@ -109,6 +119,18 @@ public void initializeState(StateInitializationContext context) throws Exception
         committer = committerFactory.create(commitUser, getMetricGroup());
 
         committableStateManager.initializeState(context, committer);
+
+        Class<?> storeCommitterClass = StoreCommitter.class;
+        Class<?> tableCommitImplClass = TableCommitImpl.class;
+        Field commitField = storeCommitterClass.getDeclaredField(COMMIT_FIELD_NAME);
```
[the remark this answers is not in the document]
MonsterChenzhuo: done.

paimon-core/src/main/java/org/apache/paimon/table/sink/TableCommitImpl.java
```diff
@@ -120,7 +120,9 @@ public TableCommitImpl(
     }
 
     public boolean forceCreatingSnapshot() {
-        return tagAutoCreation != null && tagAutoCreation.forceCreatingSnapshot();
+        return tagAutoCreation != null
+                && tagAutoCreation.canProceedWithNextTag()
```
JingsongLi, line 124: `forceCreatingSnapshot` first and then `canProceedWithNextTag`?

JingsongLi:
BTW, this is good idea! Thanks @MonsterChenzhuo

paimon-core/src/main/java/org/apache/paimon/table/sink/TableCommitImpl.java
MonsterChenzhuo: done.

MonsterChenzhuo:
> BTW, this is a good idea! Thanks @MonsterChenzhuo

😊 PTAL,thanks.

JingsongLi (approved):
+1

PR #2268, opened 2023-11-06
[spark] Support UPDATE command

Zouxxyy:
### Purpose

spark support update, usage restrictions:

- is primary key table
- MergeEngine need to be deduplicate or partial-update
- supports update ordinary field and StructType's nested field

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/scala/org/apache/spark/sql/catalyst/optimizer/RewriteRowLeverCommands.scala
+++ b/paimon-spark/paimon-spark-common/src/main/scala/org/apache/spark/sql/catalyst/optimizer/RewriteRowLeverCommands.scala
@@ -17,46 +17,67 @@
  */
 package org.apache.spark.sql.catalyst.optimizer
 
-import org.apache.paimon.CoreOptions.{MERGE_ENGINE, MergeEngine}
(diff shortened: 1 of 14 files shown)


paimon-spark/paimon-spark-common/src/main/scala/org/apache/spark/sql/catalyst/optimizer/RewriteRowLeverCommands.scala
```diff
@@ -17,46 +17,67 @@
...
+import org.apache.spark.sql.catalyst.expressions.{Attribute, Expression, PredicateHelper, SubqueryExpression}
+import org.apache.spark.sql.catalyst.plans.logical._
 import org.apache.spark.sql.catalyst.rules.Rule
 import org.apache.spark.sql.connector.catalog.SupportsDelete
 import org.apache.spark.sql.execution.datasources.DataSourceStrategy
 import org.apache.spark.sql.execution.datasources.v2.DataSourceV2Relation
 
+import java.util
+
 object RewriteRowLeverCommands extends Rule[LogicalPlan] with PredicateHelper {
```
YannByron, line 35: Lever? or Level?

JingsongLi (approved):
Looks good to me!

PR #2269, opened 2023-11-06
[WIP]Strategically Engineering Wide Tables with Foreign Key-Based Joins

### LINKED ISSUE #2270: [Feature] Strategically Engineering Wide Tables with Foreign Key-Based Joins
### Search before asking

### Motivation

When performing a join operation with Flink's dual stream join, there is an issue with excessive state storage. Flink Lookup Join only concerns itself with changes in the primary stream; changes in the dimension table cannot update data that has already been joined. With Partial Update association, there is a problem where the joined tables need to have a common primary key to ensure matching between two or more data sources. All the aforementioned solutions for generating wide tables through multi-table associations have their limitations.

I would like to implement an entirely new method to overcome these shortcomings: a dynamic dimension table-driven Lookup Join, which I also refer to as 'ForeignKey Widening'.

### Solution

_No response_

### Anything else?

_No response_

MonsterChenzhuo:
### Purpose

Linked issue: close #2270


JingsongLi:
Close this now to wait for more design.

PR #2275, opened 2023-11-07
[core]Fix modify tag period exception

MonsterChenzhuo:
### Purpose

Fix modify tag period exception.
企业微信截图_067933cf-b106-4862-925a-0cc16eae5f26

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/tag/TagAutoCreation.java
+++ b/paimon-core/src/main/java/org/apache/paimon/tag/TagAutoCreation.java
@@ -74,13 +74,8 @@ public class TagAutoCreation {
         } else {
             Snapshot lastTag = tags.lastKey();
             this.nextSnapshot = lastTag.id() + 1;
-            LocalDateTime time;
-            String lastTagValue = tags.get(lastTag);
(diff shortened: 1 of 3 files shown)


paimon-core/src/main/java/org/apache/paimon/tag/TagAutoCreation.java
```diff
@@ -74,8 +74,13 @@ private TagAutoCreation(
...
             this.nextSnapshot = lastTag.id() + 1;
-
-            LocalDateTime time = periodHandler.tagToTime(tags.get(lastTag));
+            LocalDateTime time;
+            String lastTagValue = tags.get(lastTag);
+            if (periodHandler instanceof TagPeriodHandler.DailyTagPeriodHandler) {
+                time = periodHandler.tagToTime(lastTagValue.split(" ")[0]);
+            } else {
+                time = periodHandler.tagToTime(lastTagValue);
+            }
```
yuzelin, line 83: You can just place the `split` logic into `DailyTagPeriodHandler#tagToTime`.

yuzelin (approved):
+1

PR #2276, opened 2023-11-07
 [flink] fix the problem of incorrect timeout logic in testAsyncCompactionWorks.

liming30:
### Purpose

 [flink] fix the problem of incorrect timeout logic in testAsyncCompactionWorks.

### Tests

No
### API and Format

No
### Documentation

No


JingsongLi (approved):
+1

PR #2285, opened 2023-11-08
[bug] close write anyway to avoid connection leak

leaves12138:
### Purpose

Close  rewriter anyway no mater what to avoid connection leak.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/AppendOnlyFileStoreWrite.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/AppendOnlyFileStoreWrite.java
@@ -147,7 +147,7 @@ public class AppendOnlyFileStoreWrite extends MemoryFileStoreWrite<InternalRow> 
             if (toCompact.isEmpty()) {
                 return Collections.emptyList();
             }
-            RowDataRollingFileWriter rewriter =
+            try (RowDataRollingFileWriter rewriter =
                     new RowDataRollingFileWriter(
                             fileIO,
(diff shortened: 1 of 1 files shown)


liming30 (approved):
+1

paimon-core/src/main/java/org/apache/paimon/operation/AppendOnlyFileStoreWrite.java
```diff
@@ -147,7 +147,7 @@ public AppendOnlyCompactManager.CompactRewriter compactRewriter(
             if (toCompact.isEmpty()) {
                 return Collections.emptyList();
             }
-            RowDataRollingFileWriter rewriter =
+            try (RowDataRollingFileWriter rewriter =
```
JingsongLi, line 150: Can you use try finally to avoid closing twice?

JingsongLi (changes requested):
See my comment.

paimon-core/src/main/java/org/apache/paimon/operation/AppendOnlyFileStoreWrite.java
leaves12138: Sry forget to remove the close inside.

liming30: But `rewriter.result()` requires that the `writer` has been closed.

liming30 (approved):
+1

PR #2290, opened 2023-11-08
[oss] Use jindo sdk instead of pure oss sdk

tsreaper:
### Purpose

Currently the old `aliyun-sdk-oss` is not maintained any more and has slow performance. Jindo sdk is the new sdk for aliyun oss and is under active development.

This PR replace pure oss sdk with jindo sdk.

### Tests

This PR is tested by hand.

### API and Format

No.

### Documentation

No.


JingsongLi (approved):
+1

PR #2293, opened 2023-11-09
[hotfix] Exclude conflict jar for paimon-flink tests

yuzelin:
### Purpose

This is because shade doesn't works in IDEA.


leaves12138:
+1

Issue #1804, opened 2023-08-12
[Feature] Flink catalog as persistent storage for Flink metadata


hdygxsj:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/incubator-paimon/issues) and found nothing similar.

### Motivation

Why isn't the flink catalog provided by paimon available as persistent storage for pure Flink metadata, like hive's catalog？
<img width="1062" alt="截屏2023-08-13 02 18 33" src="https://github.com/apache/incubator-paimon/assets/35210666/e6faaf3d-0bab-4d42-bdd8-ebb50e043ec7">
<img width="895" alt="截屏2023-08-13 02 19 52" src="https://github.com/apache/incubator-paimon/assets/35210666/48c808c5-eb49-4df4-b4d7-1417de44f3f9">

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

JingsongLi:
> @hdygxsj Do you means to save catalog information to a storage? We would like to do this in #1481

No, I think this issue means something like `SparkGenericCatalog`. We need a `FlinkGenericCatalog` too. I am +1 for this.

hdygxsj:
> > @hdygxsj Do you means to save catalog information to a storage? We would like to do this in #1481
> 
> No, I think this issue means something like `SparkGenericCatalog`. We need a `FlinkGenericCatalog` too. I am +1 for this.

Can you give me a chance to finish it？I'm willing to submit a PR!

FangYongs:
Got it, I'll take it 👍

FangYongs:
> > > @hdygxsj Do you means to save catalog information to a storage? We would like to do this in #1481
> > 
> > 
> > No, I think this issue means something like `SparkGenericCatalog`. We need a `FlinkGenericCatalog` too. I am +1 for this.
> 
> Can you give me a chance to finish it？I'm willing to submit a PR!

Assigned to you 👍

pegasas:
great idea & work!
This will push Flink further into cloud-native steps!
Also a good way of spreading palmon!

Following this issue.

hdygxsj:
> great idea & work! This will push Flink further into cloud-native steps! Also a good way of spreading palmon!
> 
> Following this issue.
> 
[...]

PR #2298, opened 2023-11-09
[flink] New external catalog for storing paimon and non-paimon tables

hdygxsj:
### Purpose

close issue #1804
Supports the use of paimon external catalog to persist non-paimon tables so that users do not need to create their other tables after they have created them.

### Tests

Added unit tests for FlinkExternalCatalog

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkExternalCatalog.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkExternalCatalog.java
@@ -190,7 +190,9 @@ public class FlinkExternalCatalog extends AbstractCatalog {
     public CatalogBaseTable getTable(ObjectPath tablePath)
             throws TableNotExistException, CatalogException {
         try {
-            return paimon.getTable(tablePath);
+            CatalogTable table = paimon.getTable(tablePath);
+            table.getOptions().put("connector","paimon");
+            return table;
         } catch (TableNotExistException | CatalogException e) {
             Path tableSchemaPath = new Path(externalTableSchemaDir(tablePath) + "schema");
             try {
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkExternalCatalog.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkExternalCatalog.java
@@ -191,7 +191,6 @@ public class FlinkExternalCatalog extends AbstractCatalog {
             throws TableNotExistException, CatalogException {
         try {
             CatalogTable table = paimon.getTable(tablePath);
-            Map<String, String> options = table.getOptions();
             if (!(table instanceof SystemCatalogTable)) {
                 table.getOptions().put("connector", "paimon");
             }
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkExternalCatalog.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkExternalCatalog.java
@@ -18,6 +18,7 @@
 
 package org.apache.paimon.flink;
 
+import org.apache.flink.table.catalog.ResolvedCatalogTable;
 import org.apache.paimon.fs.FileIO;
 import org.apache.paimon.fs.FileStatus;
 import org.apache.paimon.fs.Path;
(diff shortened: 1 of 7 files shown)


zhuangchong (changes requested):
`FlinkGenericCatalog` is currently supported. What is the difference between this PR and FlinkGenericCatalog?

hdygxsj:
> `FlinkGenericCatalog` is currently supported. What is the difference between this PR and FlinkGenericCatalog?

 It can enable Paimon's FileSystem to save the connection information of all Flink connectors without Hive metastore

JingsongLi:
Thanks @hdygxsj , Can you add documentation too?

hdygxsj:
> Thanks @hdygxsj , Can you add documentation too?

Already done

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkExternalCatalog.java
```diff
@@ -0,0 +1,488 @@
...
+    }
+
+    @Override
+    public void close() throws CatalogException {
+        paimon.close();
+    }
+
+    @Override
+    public List<String> listDatabases() throws CatalogException {
+        return paimon.listDatabases();
```
JingsongLi, line 132: Do we need to list databases for external tables?

hdygxsj: I think saving different external tables to different databases would give the user a better experience. For example
d6e7c95bf3ca4c629c2c634a9099135

JingsongLi: I cannot get your point.
I mean list databases for external tables. There is only paimon databases.

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/FlinkExternalCatalogTest.java
```diff
@@ -0,0 +1,243 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+/** Test for {@link FlinkExternalCatalog}. */
+public class FlinkExternalCatalogTest {
```
JingsongLi, line 52: You should add a test for `CatalogTestBase`, and mix paimon and external tables, to show the behavior of this catalog.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkExternalCatalog.java
hdygxsj: I made the external table use the same databases as the paimon table.

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/FlinkExternalCatalogTest.java
hdygxsj: Do you mean that I need to add mixed paimon and external table test cases to CatalogTestBase? The current External Table information is not stored using org.apache.paimon.catalog.Catalog , I persist it to a separate file path in FLINK_EXTERNAL_METADATA/TABLE by FileIO. It is also intended to persist the user's UDF in FLINK_EXTERNAL_METADATA/FUNCTION later.So  I cannot mix paimon and external tables in the org.apache.paimon.catalog.Catalog. FlinkExternalCatalogTest already has test cases that use a mix of paimon and external tables. Do you have any better suggestions?

-----
我的英语不太好，我的意思如下：
你的意思是我需要在CatalogTestBase 中增加使用混合使用paimon表和外部表的测试用例吗？我现在持久化外部表没有使用org.apache.paimon.catalog.Catalog，而是把它的元数据信息序列化后通过FileIO保存到了FLINK_EXTERNAL_METADATA/TABLE路径下。并且以后打算把UDF保存到FLINK_EXTERNAL_METADATA/FUNCTION。
所以我没法在org.apache.paimon.catalog.Catalog中混合使用paimon表和外部表。FlinkExternalCatalogTest 中已经有了混合使用paimon和tables的用例。您有什么更好的建议吗。

image

hdygxsj: Added integration tests that mix Paimon tables with non-paimon

JingsongLi:
Thanks! I am busy now, I need some time to take a look again.

macdoor:
Hive is outdate. This feature allows paimon to no longer rely on hive in scenarios where paimon and kafka are used together.

hdygxsj:
Anything need  to fix? It seems possible to execute these test cases successfully in my local environment
image
image
image

PR #2299, opened 2023-11-10
[flink] Correct pom of paimon-flink-1.17

yuzelin:
### Purpose

paimon-flink-common use flink-1.18, so paimon-flink-1.17 should exculdes denpendencies of paimon-flink-common


leaves12138 (approved):
+1

PR #2302, opened 2023-11-13
[flink] Add config "table-conf sink.parallelism" for ordinary compact

leaves12138:
### Purpose

Enable table-conf sink.parallelism for ordinary dedicated compact job. We can use --table-conf sink.parallelism=xxx to modify the parallelism of dedicated compaction job.

### API and Format

Example: 
```sql
flink run ./paimon-flink-action-0.6-SNAPSHOT.jar compact \
--warehouse \
/Users/paimontest/GlobalBatchReadWrite \
--database \
my_db \
--table \
Orders1 \
--table-conf sink.parallelism=4
```

### Documentation

Added documentation for dedicated compaction job.


JingsongLi (approved):
+1

PR #2303, opened 2023-11-13
[e2e] Fix unstable e2e tests with Kafka service

yuzelin:
### Purpose

Try to fix the frequently occurred "container is not running" error.


JingsongLi (approved):
+1

PR #2297, opened 2023-11-09
[Bug] Fix alter table column do not work for the non-physical column

Aitozi:
### Purpose

Linked issue: close #2289

PR #2307, opened 2023-11-13
[flink] [fix] Flink add computed column should be nullable

leaves12138:
### Purpose

sql "ALTER TABLE T add (__proc_time AS PROCTIME())" should be work well with paimon. And the clumn __proc_time which is ComputedColumn should be nullable.


JingsongLi:
You should add test to use computed column and watermark for reading.

leaves12138:
> You should add test to use computed column and watermark for reading, a real ITCase.

done

JingsongLi:
@leaves12138 Maybe you can just take a review to #2297, is that what you want?

leaves12138:
Already resolved in https://github.com/apache/incubator-paimon/pull/2297#pullrequestreview-1728717978

PR #2311, opened 2023-11-14
[core] Remove RocksDB log relocation to flink log

JingsongLi:
### Purpose

Now, RocksDB log file will output to flink/log, for session cluster, this will outputs more and more log files...
We should keep RocksDB default configuration, log file should be along to data files.


SteNicholas (approved):
LGTM. Merging.

PR #2312, opened 2023-11-14
[flink][kafka-cdc] Refactor Kafka CDC RecordParser

yuzelin:
### Purpose

Currently the RecordParser has strong validation of the Json format. But in some cases, the desired data is not present (data type, table name, etc). We should make sure we can try best to synchronize data. This PR has done:

1. Remove the strong validation before handle data. Now the RecordParser will check specific field when really need it.
2. If the json is invalid, log it before throwing exception (for trouble shooting).
3. Make database, table and primary keys optional.
4. Handle case sensitivity at last.

### DIFF
--- a/docs/content/cdc-ingestion/kafka-cdc.md
+++ b/docs/content/cdc-ingestion/kafka-cdc.md
@@ -63,7 +63,12 @@ If a message in a Kafka topic is a change event captured from another database u
 </table>
 
 {{< hint info >}}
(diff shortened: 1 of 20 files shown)


JingsongLi (changes requested):
Can you add tests for what are you want to support?

JingsongLi (changes requested):
Can you add tests for what are you want to support?

JingsongLi (approved):
+1

PR #2313, opened 2023-11-14
[core] Filesystem catalog supports setting case-sensitive option.

zhuangchong:
### Purpose

 Filesystem catalog supports setting case-sensitive option.


paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/action/cdc/kafka/KafkaSyncDatabaseAction.java
```diff
@@ -183,21 +180,9 @@ public void build() throws Exception {
...
-                tablePrefix.equals(tablePrefix.toLowerCase()),
-                String.format(
-                        "Table prefix [%s] cannot contain upper case in case-insensitive catalog.",
-                        tablePrefix));
-        checkArgument(
-                tableSuffix.equals(tableSuffix.toLowerCase()),
-                String.format(
-                        "Table suffix [%s] cannot contain upper case in case-insensitive catalog.",
-                        tableSuffix));
+        catalog.validateCaseInsensitive("Database", database);
```
JingsongLi, line 183: This should be just a util method... Not a public method.

JingsongLi:
Thanks @zhuangchong , left comment about public method in `Catalog`.

zhuangchong:
> Thanks @zhuangchong , left comment about public method in `Catalog`.

Thanks @JingsongLi ,  done.

JingsongLi (approved):
+1

PR #2220, opened 2023-10-31
[cdc] refactor mysql cdc action uniformly uses flatmap to process records

zhuangchong:
### Purpose

Linked issue: close #xxx
refactor mysql cdc action uniformly uses flatmap to process records.

make CDC processing procedures consistent.

PR #2314, opened 2023-11-14
[cdc] Remove JsonPrimaryKeyDeserializationSchema and refactor NewTableSchemaBuilder

yuzelin:
### Purpose

JsonPrimaryKeyDeserializationSchema is unnecessary because #2312 has remove check for pk names.
#2220 removed `MySqlTableSchemaBuilder` so the `NewTableSchemaBuilder ` interface can be refactored to a class.


JingsongLi (approved):
+1

PR #2315, opened 2023-11-14
[Hive] Using hive table location when catalog is HiveCatalog

humengyu2012:
### Purpose

We want to query Paimon tables with Presto, but our Paimon tables are scattered across different warehouse directories(such as `/user1/warehouse`, `/user2/warehouse`).

Therefore, we use Hive catalog to discover the addresses of different Paimon warehouses. However, in the current implementation of the Hive catalog, it does not use the Hive table's location as the location for Paimon tables.

In this pull request, we have re-implemented the `org.apache.paimon.hive.HiveCatalog#getDataTableLocation` method, which achieves the desired effect.

### DIFF
--- a/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
+++ b/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
@@ -167,15 +167,22 @@ public class HiveCatalog extends AbstractCatalog {
             String databaseName = identifier.getDatabaseName();
             String tableName = identifier.getObjectName();
             if (client.tableExists(databaseName, tableName)) {
-                return new Path(
-                        locationHelper.getTableLocation(client.getTable(databaseName, tableName)));
+                String location =
+                        locationHelper.getTableLocation(client.getTable(databaseName, tableName));
+                if (location != null) {
+                    return new Path(location);
+                }
             } else {
                 // If the table does not exist,
                 // we should use the database path to generate the table path.
-                return new Path(
-                        locationHelper.getDatabaseLocation(client.getDatabase(databaseName)),
-                        tableName);
+                String dbLocation =
+                        locationHelper.getDatabaseLocation(client.getDatabase(databaseName));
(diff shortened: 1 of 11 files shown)


paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
```diff
@@ -161,6 +161,18 @@ public Optional<MetastoreClient.Factory> metastoreClientFactory(Identifier ident
         }
     }
 
+    @Override
+    public Path getDataTableLocation(Identifier identifier) {
+        try {
+            return new Path(
```
JingsongLi, line 167: Introduce a method to `LocationHelper`: `getTableLocation`.

JingsongLi (commented):
Left one comment.

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
```diff
@@ -161,6 +161,18 @@ public Optional<MetastoreClient.Factory> metastoreClientFactory(Identifier ident
...
+                    client.getTable(identifier.getDatabaseName(), identifier.getObjectName())
+                            .getSd()
+                            .getLocation());
+        } catch (TException e) {
```
JingsongLi, line 171:
```
catch (NoSuchObjectException e) {
     return super.getDataTableLocation(identifier);
}
```

humengyu2012: Regarding HiveCatalog, I have a question: should we fallback to the parent implemets, or should we adhere to the Hive specification and create tables under the database path? @JingsongLi

JingsongLi: OK let's just throw exception.

JingsongLi (approved):
+1

PR #2316, opened 2023-11-14
[hotfix] Fix FlinkActionsE2eTest testDropPartition timeout

zhuangchong:
### Purpose

Fix https://github.com/apache/incubator-paimon/actions/runs/6862065679/job/18659032196#step:5:183


JingsongLi (approved):
+1

PR #2318, opened 2023-11-15
[core] Close CompactMetrics after closing its corresponding writer

tsreaper:
### Purpose

Currently we don't have a `close` method in `MetricGroup`. When using bucket-wise metric groups (for example `CompactMetrics`), because more and more buckets will be written as the job runs, we need to clean up the resources (for example, their corresponding Flink metric group) in time.

This PR closes `CompactMetrics` after its corresponding writer is closed.

### Tests

As `CompactMetrics` is not bridged into Flink's metrics system at this time, we can't come up with unit tests. This should be tested in the PR when bridging `CompactMetrics` with Flink's metrics system.

### API and Format

No.

### Documentation

No.


wg1026688210:
+1

JingsongLi (approved):
+1

PR #2319, opened 2023-11-15
[core] Introduce RawFile public API to indicate files which can be read without merging

tsreaper:
### Purpose

Currently many C++ big data systems (for example StarRocks) have supported Paimon. However as Paimon's data files use the LSM tree structure, these data files must be merged before reading, and these C++ systems have to use Paimon's Java reader. This affects the performance a lot.

However, under certain circumstances (for example, when a table is fully compacted) the data files can be read without merging. This PR introduces a new public API called `RawFile` to indicate such files. API users can try to extract `RawFile` from `DataSplit` and directly read the files with their own very fast file reader.

### Tests

* `DataSplitTest`.

### API and Format

Yes. It introduces a new public API.

### Documentation

No. This change should only be used by other big data system developers, not users.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/source/RawFile.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/source/RawFile.java
@@ -34,15 +34,10 @@ public class RawFile {
     private final String path;
     private final long offset;
     private final long length;
-    private final CoreOptions.FileFormatType format;
+    private final String format;
     private final long schemaId;
 
-    public RawFile(
-            String path,
-            long offset,
-            long length,
-            CoreOptions.FileFormatType format,
-            long schemaId) {
+    public RawFile(String path, long offset, long length, String format, long schemaId) {
         this.path = path;
         this.offset = offset;
         this.length = length;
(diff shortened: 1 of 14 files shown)


paimon-core/src/main/java/org/apache/paimon/table/source/RawTableFile.java
```diff
@@ -0,0 +1,72 @@
...
+ * @since 0.6.0
+ */
+@Public
+public class RawTableFile {
+
+    private final String path;
+    private final long offset;
+    private final long length;
+
+    public RawTableFile(String path, long offset, long length) {
```
JingsongLi, line 37: add `String format`.

JingsongLi, line 37: Add schema ID.

JingsongLi (changes requested):

paimon-core/src/main/java/org/apache/paimon/table/source/RawFile.java
```diff
@@ -0,0 +1,96 @@
...
+ *
+ * @since 0.6.0
+ */
+@Public
+public class RawFile {
+
+    private final String path;
+    private final long offset;
+    private final long length;
+    private final CoreOptions.FileFormatType format;
```
JingsongLi, line 37: just String? It is not so good to expose `CoreOptions.FileFormatType`.

JingsongLi (approved):
+1

Issue #2227, opened 2023-11-01
[Feature] Kafka cdc supports debezium-json format


sunxiaojian:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/incubator-paimon/issues) and found nothing similar.

### Motivation

Kafka cdc supports debezium-json format

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

sunxiaojian:
I can try to contribute this PR

PR #2321, opened 2023-11-15
[cdc]Pulsar cdc supports debezium-json format

sunxiaojian:
### Purpose

Linked issue: close [#2227](https://github.com/apache/incubator-paimon/issues/2227)


JingsongLi (commented):
Thanks @sunxiaojian for the contribution.
I think by default, all formats will be supported for pulsar too. We don't need to add this test now, otherwise we need to add all format tests for pulsar.

PR #2326, opened 2023-11-16
[core] Check existence for commit files of savepoint recovering

### LINKED ISSUE #2273: [Bug] Streaming writes to a Deduplicate table, flink job can't recover for FileNotFoundException
### Search before asking

### Paimon version

0.6-SNAPSHOT

### Compute Engine

Flink 1.16.1

### Minimal reproduce step

1. DDL 
`CREATE TABLE IF NOT EXISTS paimon_catalog.dim.event_log_aggregation(
  order_no string,
  new_status int,
  creator string ,
  id bigint,
  create_time timestamp COMMENT '日志创建时间',
  version bigint COMMENT '与 create_time 反相关',
  PRIMARY KEY (order_no, new_status) NOT ENFORCED
) with (
    'merge-engine' = 'Deduplicate',
    'sequence.field' = 'version'
);`
2. starting a flink job (consume kafka and write into paimon)
3. waiting some time, stop flink job with a savepoint 
4. recover flink job with savepoint, flink job throw exception
image
5. restart flink job without savepoint, but same exception is occur

Even though I restarted the flink task stateless to write into paimon, the same error occurred:
FileNotFoundException

### Anything else?

_No response_

JingsongLi:
### Purpose

If a user restores a Flink job from an old savepoint and the table has been written by another job for a long time, it means that the files in the savepoint may have been deleted.
At present, the commit will still succeed, which results in the table containing deleted files, which is a serious problem.

This PR aims to check the recovered file, and if it does not exist, report an error to the user.

Fix #2273

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/io/DataFileMeta.java
+++ b/paimon-core/src/main/java/org/apache/paimon/io/DataFileMeta.java
@@ -21,6 +21,7 @@ package org.apache.paimon.io;
 import org.apache.paimon.CoreOptions;
 import org.apache.paimon.data.BinaryRow;
 import org.apache.paimon.data.Timestamp;
+import org.apache.paimon.fs.Path;
 import org.apache.paimon.stats.BinaryTableStats;
 import org.apache.paimon.stats.FieldStatsArraySerializer;
 import org.apache.paimon.types.ArrayType;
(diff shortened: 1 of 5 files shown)


yuzelin (approved):
+1

PR #2328, opened 2023-11-16
[spark] Fix docs, fix bug that spark sort for order should be non-global

leaves12138:
### Purpose

Fix docs, remove location parameter.

Disable non-sort compact by Spark CompactProcedure( it cost too much)


JingsongLi (approved):
+1

PR #2329, opened 2023-11-16
[core] Add one more possible cause in commit conflict exception message

tsreaper:
### Purpose

If users use `BEGIN STATEMENT SET` and `END STATEMENT SET` to execute multiple `INSERT` statements into the same Paimon table, commit conflicts will also occur. This PR adds this possible cause into the exception message.

### Tests

No need for tests.

### API and Format

No.

### Documentation

No.


JingsongLi (approved):
+1

PR #2330, opened 2023-11-16
[cdc] Fix that initializing decimal type that missing precision will throw exception

yuzelin:
### Purpose

The decimal type will validate that precision is between 1-38.

### API and Format

`KafkaCanalSyncTableActionITCase#testMissingDecimalPrecision`


JingsongLi (approved):
+1

JingsongLi (approved):
+1

PR #2336, opened 2023-11-17
[core] Introduce executeFilter to TableRead

JingsongLi:
### Purpose

Computing engines such as Trino need to convert a row into Trino's own data structure after reading it, which is costly. We can actually perform filtering before returning to Trino to reduce the amount of data that needs to be converted.

Trino Connector can just use `TableRead.executeFilter()` to enable this feature.

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/predicate/CompoundPredicate.java
+++ b/paimon-common/src/main/java/org/apache/paimon/predicate/CompoundPredicate.java
@@ -55,7 +55,7 @@ public class CompoundPredicate implements Predicate {
 
     @Override
     public boolean test(InternalRow row) {
-        return false;
+        return function.test(row, children);
     }
 
     @Override
--- a/paimon-common/src/main/java/org/apache/paimon/predicate/CompoundPredicate.java
+++ b/paimon-common/src/main/java/org/apache/paimon/predicate/CompoundPredicate.java
@@ -18,6 +18,7 @@
 
 package org.apache.paimon.predicate;
 
+import org.apache.paimon.data.InternalRow;
 import org.apache.paimon.format.FieldStats;
 
 import java.io.Serializable;
@@ -53,6 +54,11 @@ public class CompoundPredicate implements Predicate {
     }
 
     @Override
+    public boolean test(InternalRow row) {
+        return false;
+    }
+
+    @Override
     public boolean test(long rowCount, FieldStats[] fieldStats) {
         return function.test(rowCount, fieldStats, children);
     }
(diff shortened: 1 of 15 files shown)


paimon-common/src/main/java/org/apache/paimon/predicate/CompoundPredicate.java
```diff
@@ -52,6 +53,11 @@ public boolean test(Object[] values) {
         return function.test(values, children);
     }
 
+    @Override
+    public boolean test(InternalRow row) {
+        return false;
```
Aitozi, line 58: this should be `funciton.test(row, children)` ?

JingsongLi: you are right

paimon-core/src/test/java/org/apache/paimon/table/AppendOnlyFileStoreTableTest.java
```diff
@@ -117,6 +118,61 @@ public void testBatchFilter() throws Exception {
                                 "2|21|201|binary|varbinary|mapKey:mapVal|multiset"));
     }
 
+    @Test
+    public void testBatchFilterWithExecution() throws Exception {
```
yuzelin, line 122: Is it necessary to add a same test in `PrimaryKeyFileStoreTableTest`?

JingsongLi: I think it is the same, so I think it is OK to be just here.

paimon-common/src/main/java/org/apache/paimon/predicate/PredicateProjectionConverter.java
```diff
@@ -0,0 +1,47 @@
...
+public class PredicateProjectionConverter implements PredicateReplaceVisitor {
+
+    private final Map<Integer, Integer> reversed;
+
+    public PredicateProjectionConverter(int[] projection) {
+        this.reversed = new HashMap<>();
+        for (int i = 0; i < projection.length; i++) {
+            reversed.put(projection[i], i);
+        }
+    }
```
tsreaper, line 35: Is it possible that `projection` contains the same integer? Consider a table with three columns `a`, `b` and `c`, if I query `SELECT a, c, c, b FROM T`, then `projection` might be `[0, 2, 2, 1]`?

paimon-core/src/test/java/org/apache/paimon/table/AppendOnlyFileStoreTableTest.java
yuzelin: Alright.

yuzelin (approved):
+1

paimon-common/src/main/java/org/apache/paimon/predicate/PredicateProjectionConverter.java
JingsongLi: It is OK to just pick a last column reference.

PR #2354, opened 2023-11-20
[Hive] HiveCatalog support hive 1.x

humengyu2012:
Add 1.x HiveMetastore support. It may be useful for presto and trino due to presto/trino using 1.2.2 for hive connector.

Support Method:

`org.apache.hadoop.hive.metastore.RetryingMetaStoreClient#getProxy(org.apache.hadoop.hive.conf.HiveConf, org.apache.hadoop.hive.metastore.HiveMetaHookLoader, java.util.Map<java.lang.String,java.lang.Long>, java.lang.String)`

image
image

### DIFF
--- a/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
+++ b/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
@@ -558,53 +558,62 @@ public class HiveCatalog extends AbstractCatalog {
             ImmutableMap.of(
                     // for hive 1.x
                     new Class<?>[] {
-                                HiveConf.class,
-                                HiveMetaHookLoader.class,
-                                ConcurrentHashMap.class,
-                                String.class
-                            },
-                            (getProxyMethod, hiveConf, clientClassName) ->
-                                    (IMetaStoreClient)
-                                            getProxyMethod.invoke(
-                                                    null,
-                                                    hiveConf,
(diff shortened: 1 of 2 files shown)


JingsongLi (commented):
Can you extract a class for client creation? for example `HmsClientFactory`?

And other thing is testing, can we have any testing to cover this modification?

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
```diff
@@ -551,33 +555,97 @@ private Lock lock(Identifier identifier) {
...
+                                    (IMetaStoreClient)
+                                            getProxyMethod.invoke(
+                                                    null,
+                                                    hiveConf,
+                                                    (HiveMetaHookLoader) (tbl -> null),
+                                                    new ConcurrentHashMap<>(),
+                                                    clientClassName,
+                                                    true))
+                    .build();
+    // If clientClassName is HiveMetaStoreClient,
```
JingsongLi, line 613: Leave a empty line above.

humengyu2012: done

humengyu2012:
> Can you extract a class for client creation? for example `HmsClientFactory`?
> 
> And other thing is testing, can we have any testing to cover this modification?

I added a class named `RetryingMetaStoreClientFactory`, can you take a look again? @JingsongLi

JingsongLi (approved):
+1

PR #2359, opened 2023-11-21
[core] Introduce read-optimized system table

tsreaper:
### Purpose

OLAP systems have high demands on reading efficiency. This PR introduces a new system table called the "read-optimized table". This table is optimized for reading by avoiding merging files.

For primary key tables, this system table only scans files on top level. For append only tables, as all files can be read without merging, this system table does nothing special.

### Tests

* `PrimaryKeyFileStoreTableTest#testReadOptimizedTable`.
* `CatalogTableITCase#testReadOptimizedTable`.

### API and Format

No.

### Documentation

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/system/ReadOptimizedTable.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/system/ReadOptimizedTable.java
@@ -1,0 +1,153 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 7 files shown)


paimon-core/src/main/java/org/apache/paimon/table/system/ReadOptimizedTable.java
```diff
@@ -0,0 +1,153 @@
...
+    public List<String> primaryKeys() {
+        return dataTable.primaryKeys();
+    }
+
+    @Override
+    public SnapshotReader newSnapshotReader() {
+        if (dataTable.schema().primaryKeys().size() > 0) {
+            return dataTable
+                    .newSnapshotReader()
+                    .withLevelFilter(level -> level == coreOptions().numLevels() - 1);
```
JingsongLi, line 94: we can just read max level in a split. not just table max level.

you can add test for this.

JingsongLi (approved):
+1

PR #2362, opened 2023-11-21
[core] Support bulk load mode for partial-update merge-engine

JingsongLi:
### Purpose

Now only default mode supports bulk load mode for cross partition update.
This PR:
1. Throws exception for sequence-field definition for cross partition update, this is hard to support, we cannot know old sequence number for the record in previous partition.
2. Extract `ExistingProcessor` to support bulk load mode for partial-update and others merge-engine.


leaves12138 (approved):
+1

PR #2365, opened 2023-11-22
[spark] Merge into support update nested column

Zouxxyy:
### Purpose

Merge into support update nested column


paimon-spark/paimon-spark-common/src/main/scala/org/apache/spark/sql/catalyst/analysis/AssignmentAlignmentHelper.scala
```diff
@@ -35,24 +36,32 @@ trait AssignmentAlignmentHelper extends SQLConfHelper {
...
-  def toRefSeq(expr: Expression): Seq[String] = expr match {
+  protected def alignAssignments(
+      attrs: Seq[Attribute],
+      assignments: Seq[Assignment]): Seq[Assignment] = {
+    generateAlignedExpressions(attrs, assignments).zip(attrs).map {
+      case (expression, field) => Assignment(field, expression)
+    }
+  }
+
+  private def toRefSeq(expr: Expression): Seq[String] = expr match {
```
YannByron, line 64: can we move this method to `ExpressionHelper`~

YannByron:
+1

PR #2366, opened 2023-11-22
[core] Use Flink Managed Memory for RocksDB in cross partition update

JingsongLi:
### Purpose

Default block cache of RocksDB is 8MB. It is too small to performance.

We can use Flink managed memory for RocksDB.


leaves12138 (approved):
+1

PR #2367, opened 2023-11-22
[core][spark] Supports to push down limit

YannByron:
### Purpose
Support to push down limit to accelerate query.

Linked issue: close #xxx


paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/SparkScanBuilder.scala
```diff
@@ -0,0 +1,88 @@
...
+
+  override def pruneColumns(requiredSchema: StructType): Unit = {
+    val pruneFields = requiredSchema.fieldNames
+    val fieldNames = table.rowType.getFieldNames
+    val projected = pruneFields.map(field => fieldNames.indexOf(field))
+    this.projectedIndexes = Some(projected)
+  }
+
+  override def pushLimit(limit: Int): Boolean = {
+    if (table.isInstanceOf[AppendOnlyFileStoreTable]) {
```
JingsongLi, line 81: No need to `if append table`? Paimon-core scan should take care about this.

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/SparkScanBuilder.scala
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  override def pruneColumns(requiredSchema: StructType): Unit = {
+    val pruneFields = requiredSchema.fieldNames
+    val fieldNames = table.rowType.getFieldNames
+    val projected = pruneFields.map(field => fieldNames.indexOf(field))
+    this.projectedIndexes = Some(projected)
+  }
+
+  override def pushLimit(limit: Int): Boolean = {
```
JingsongLi, line 80: Does Spark SQL will not push limit if there is filter?

paimon-spark/paimon-spark-common/src/test/scala/org/apache/paimon/spark/sql/PaimonSQLPerformanceTest.scala
```diff
@@ -0,0 +1,88 @@
...
+import org.apache.paimon.table.source.DataSplit
+
+import org.apache.spark.sql.Row
+import org.apache.spark.sql.connector.read.{ScanBuilder, SupportsPushDownLimit}
+import org.apache.spark.sql.util.CaseInsensitiveStringMap
+import org.junit.jupiter.api.Assertions
+
+import scala.collection.JavaConverters._
+
+class PaimonSQLPerformanceTest extends PaimonSparkTestBase {
```
JingsongLi, line 30: What this class for? It looks like we can not find performance result from these tests?

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/SparkScanBuilder.scala
YannByron: Here, a `boolean` value is needed, which is used in SparkSQL. Or, paimon-core needs to provide an api to be called.

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/SparkScanBuilder.scala
YannByron: yes. SparkSQL can guarantee this.

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/SparkScanBuilder.scala
JingsongLi: We can just return false here? Best effort pushdown?

YannByron: As a temporary solution, it's ok.

YannByron:
These failed UTs are not related to this pr.

YannByron:
https://github.com/apache/incubator-paimon/issues/2404

paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreScan.java
```diff
@@ -199,6 +202,11 @@ public FileStoreScan withMetrics(ScanMetrics metrics) {
         return this;
     }
 
+    @Override
+    public FileStoreScan withLimit(int limit) {
```
JingsongLi, line 206: Can limit implementation just in `AppendOnlyFileStoreScan`?

JingsongLi:
Another solution: we can make this limit pushdown generic. We can do this limit in `InnerTableScanImpl`, and we can just check `Split.convertToRawFiles`, check the rowcount, because rawFiles don't need to be merged.

paimon-core/src/main/java/org/apache/paimon/table/source/RawFile.java
```diff
@@ -76,12 +80,18 @@ public long schemaId() {
...
+        return rowCount;
+    }
+
     public void serialize(DataOutputView out) throws IOException {
         out.writeUTF(path);
         out.writeLong(offset);
         out.writeLong(length);
         out.writeUTF(format);
         out.writeLong(schemaId);
+        out.writeLong(rowCount);
```
JingsongLi, line 94: Increment `DataSplit.serialVersionUID` too.

paimon-spark/paimon-spark-3.1/src/main/scala/org/apache/paimon/spark/SparkScanBuilder.scala
```diff
@@ -0,0 +1,107 @@
...
+import org.apache.paimon.table.Table
+
+import org.apache.spark.internal.Logging
+import org.apache.spark.sql.connector.read.{Scan, ScanBuilder, SupportsPushDownFilters, SupportsPushDownRequiredColumns}
+import org.apache.spark.sql.sources.Filter
+import org.apache.spark.sql.types.StructType
+
+import scala.collection.mutable
+
+class SparkScanBuilder(table: Table)
```
JingsongLi, line 30: Can we have a `SparkBaseScanBuilder` to reuse code?

JingsongLi, line 30: You need to add class for spark-3.2 too? And add some itcase for spark-3.2 too (For ITCase, we also can have some abstraction to reduce code replication).

paimon-core/src/main/java/org/apache/paimon/table/source/InnerTableScanImpl.java
```diff
@@ -68,9 +81,96 @@ public TableScan.Plan plan() {
...
+        }
+    }
+
+    private DataSplit composeDataSplit(DataSplit split, long rowCountRequired) {
+        List<DataFileMeta> dataFiles = new ArrayList<>();
+        List<RawFile> rawFiles = new ArrayList<>();
+        long scannedRowCount = 0;
+
+        List<DataFileMeta> originalDataFiles = split.dataFiles();
+        List<RawFile> originalRawFiles = split.convertToRawFiles().get();
```
JingsongLi, line 160: Just get? Exception when it is empty?

paimon-core/src/main/java/org/apache/paimon/table/source/InnerTableScanImpl.java
```diff
@@ -68,9 +81,96 @@ public TableScan.Plan plan() {
...
+                    break;
+                }
+
+                DataSplit split = splits.get(i);
+                long splitRowCount = getRowCountForSplit(split);
+                if (scannedRowCount + splitRowCount <= pushDownLimit) {
+                    limitedSplits.add(split);
+                    scannedRowCount += splitRowCount;
+                } else {
+                    DataSplit newSplit = composeDataSplit(split, pushDownLimit - scannedRowCount);
```
JingsongLi, line 108: I think we don't need to introduce `composeDataSplit`.

The reason we introduce limit pushdown is to reduce split number. We don't need to reduce files in split.

paimon-spark/paimon-spark-3.1/src/main/scala/org/apache/paimon/spark/SparkScanBuilder.scala
YannByron: For UT, just copy codes for now. Let refine it in another pr.

JingsongLi (approved):
+1

PR #2373, opened 2023-11-23
[flink] Limit max split while continuously scanning to avoid JobManager OOM

leaves12138:
### Purpose

close: https://github.com/apache/incubator-paimon/issues/2370 

### Tests

Added test.

### API and Format

No other influence to user

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
@@ -217,11 +217,13 @@ public class CoreOptions implements Serializable {
                     .defaultValue(Duration.ofSeconds(10))
                     .withDescription("The discovery interval of continuous reading.");
 
-    public static final ConfigOption<Integer> SCAN_SPLIT_MAX_PER_TASK =
-            key("scan.split-max-per-task")
+    public static final ConfigOption<Integer> SCAN_MAX_SPLITS_PER_TASK =
+            key("scan.max-splits-per-task")
                     .intType()
                     .defaultValue(10)
-                    .withDescription("Max split size should be cached for one task.");
+                    .withDescription(
+                            "Max split size should be cached for one task while scanning. "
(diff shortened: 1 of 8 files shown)


paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -219,6 +219,12 @@ public class CoreOptions implements Serializable {
                     .defaultValue(Duration.ofSeconds(10))
                     .withDescription("The discovery interval of continuous reading.");
 
+    public static final ConfigOption<Integer> SCAN_SPLIT_MAX_PER_TASK =
+            key("scan.split-max-per-task")
```
JingsongLi, line 223: `scan.max-splits-per-task`?
And the description should be more...

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/ContinuousFileSplitEnumerator.java
```diff
@@ -85,7 +87,8 @@ public ContinuousFileSplitEnumerator(
             @Nullable Long nextSnapshotId,
             long discoveryInterval,
             StreamTableScan scan,
-            BucketMode bucketMode) {
+            BucketMode bucketMode,
+            Integer splitMaxPerTask) {
```
JingsongLi, line 91: why this is Integer instead of int?

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
leaves12138: done

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/ContinuousFileSplitEnumerator.java
leaves12138: used int instead

JingsongLi (approved):
+1

PR #2379, opened 2023-11-23
[core] Remove Path.getPath

JingsongLi:
### Purpose

"oss://bucket/tmp/1" => Path.getPath => "/tmp/1".

It is dangerous to use this method, we should remove it for avoiding bugs.


leaves12138 (approved):
+1

PR #2381, opened 2023-11-23
[Core]: Support value filter push down in max file level for keyed table

### LINKED ISSUE #1731: [Feature] Add value filter for primary key table
### Search before asking

### Motivation

Adding a value filter can improve pruning efficiency and thus improve query performance

### Solution

consider this case:
```
L(Level 0f LSM)3: insert key = 1, value = a
L2: update key = 1, value = b
filter: value = a
```
If we perform filter push down on value, L3 file will be chosen , but L2 file will be ignored, and the final result will be  key = 1, value = a while the correct is empty. We can't simple push down filters to all file.We can refine the cases and push down some files accordingly.

**Case 1:**
If the user has not configured sequence.field, then the highest LSM level 'Sorted Run' under the bucket can be pushed down.
Consider above case, if L3 is the highest level in current bucket, if we only filter L3, the final result will be empty, it's correct!

**Case 2:**
Consider this case：
```
L4: insert key = 2, value = c
L3: update key = 1, value = a
L2: update key = 1, value = b
filter: key = 1, value = a
```
 User has not configured sequence.field. If key filter pushdown filters out all files in the highest level L4, then value filter pushdown can be applied to L3. The result obtained by following this step in the above example will be empty, it's correct.

### Anything else?

_No response_

shidayang:
### Purpose
Linked issue: close #1731 

### Tests
org.apache.paimon.operation.KeyValueFileStoreScanTest#testWithValueFilterForMaxLevel


paimon-core/src/main/java/org/apache/paimon/operation/KeyValueFileStoreScan.java
```diff
@@ -97,23 +103,50 @@ protected boolean filterByStats(ManifestEntry entry) {
...
-                                                entry.file().schemaId()),
-                                        entry.file().rowCount()))) {
-                    return true;
-                }
+        if (canSkipWholeBucket(entries)) {
+            return Collections.emptyList();
+        }
+
+        // If there is no custom sequence field, we can filter the whole bucket by the max level.
+        if (!hasCustomSequenceField) {
```
JingsongLi, line 118: I think the bool should be more complicated.
1. minSequenceNumber in non-max levels should larger than maxSequenceNumber in max levels.
2. Merge Engine is deduplicate.

paimon-core/src/main/java/org/apache/paimon/operation/KeyValueFileStoreScan.java
```diff
@@ -97,23 +103,50 @@ protected boolean filterByStats(ManifestEntry entry) {
...
+                        .collect(Collectors.toList());
+            }
+        }
+
+        return entries;
+    }
+
+    private boolean canSkipWholeBucket(List<ManifestEntry> entries) {
+        for (ManifestEntry entry : entries) {
+            if (applyValueFilter(entry)) {
```
JingsongLi, line 136: Can we cache this value, and reuse this value for max level filter?

JingsongLi (changes requested):
Thanks @shidayang for the contribution, left two comments.

paimon-core/src/main/java/org/apache/paimon/operation/KeyValueFileStoreScan.java
JingsongLi: You can add some ITCase to cover this.

JingsongLi:
Hi @shidayang .
I still  think it is very dangarous to do filter on max level. Alternatively, we can determine whether the current bucket does not need to be merged in order to perform a filter pushdown. See #2922

You can re-open this if you have more questions.

PR #2386, opened 2023-11-24
[hotfix] Fix variable warehouse of type org.apache.paimon.fs.Path

zhuangchong:
### Purpose

Linked issue: close #xxx

PR #2387, opened 2023-11-24
[Test] hotfix call of Path.getPath()

schnappi17:
### Purpose

Hotfix call of Path.getPath() in compaction unit test.


zhuangchong:
#2386

JingsongLi:
Already fixed.

PR #2391, opened 2023-11-25
[Core]Supports asynchronous manifest merge

hililiwei:
### Purpose

If there are many manifest files for tables, synchronous merging can be slow.


JingsongLi (approved):
+1

PR #2396, opened 2023-11-27
[Feature] support meter-type metric

### LINKED ISSUE #2395: [Feature] Support meter-type metric
### Search before asking

### Motivation

support meter-type metric which is used for measuring throughput.

### Solution

_No response_

### Anything else?

_No response_

zhoulii:
### Purpose

Linked issue: close #2395


JingsongLi:
I think it is better to add meter-type if we have requirement to use it.

PR #2401, opened 2023-11-28
[hotfix] Fix append table's default value for spillable parameter on streaming writes is not as expected.

zhuangchong:
… writes is not as expected.

### Purpose

add method:

    public boolean bufferSpillable() {
        return options.writeBufferSpillable(fileIO.isObjectStore(), isStreamingMode);
    }


JingsongLi:
I cannot understand what are you fixing.

zhuangchong:
> I cannot understand what are you fixing.

I'll take a closer look. I'll close this PR first.

PR #2408, opened 2023-11-28
[hive] In hiveCatalog we do not need catalog.warehouse config

### LINKED ISSUE #2407: [Feature]  In hiveCatalog we do not need catalog.warehouse config
### Search before asking

### Motivation

In Spark, when we set metastore to hive, command like blow:
```
spark-sql --jars /root/ljh/paimon-spark-3.3-0.6-20231122.093342-69.jar,/root/ljh/paimon-oss-0.6-20231122.093342-76.jar \
         --conf spark.sql.catalog.paimon=org.apache.paimon.spark.SparkCatalog \
         --conf spark.sql.extensions=org.apache.paimon.spark.extensions.PaimonSparkSessionExtensions \
                   --verbose \
                   --conf spark.sql.catalog.paimon.metastore=hive \
                   --conf spark.sql.catalog.paimon.warehouse=xxx \
                   --conf spark.sql.catalog.paimon.uri=xxx \
                   --conf spark.sql.catalog.paimon.fs.oss.endpoint=xxx \
```
We must set spark.sql.catalog.paimon.warehouse=xxx, but in hive catalog we do not need warehouse, it makes confused.
But if warehouse not been set, Error: Paimon warehouse path must be set' will be thrown.

### Solution

We need warehouse, because we need FileIO accordingly. So we can just get warehouse from hive metastore, so we can get FileIO accordingly，such as OSSFileIO,S3FileIO.

### Anything else?

_No response_

monkeyboy123:
### Purpose

Linked issue: close #2407

In hiveCatalog we do not need catalog.warehouse config, it makes confused

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/catalog/CatalogFactory.java
+++ b/paimon-core/src/main/java/org/apache/paimon/catalog/CatalogFactory.java
@@ -23,7 +23,8 @@ import org.apache.paimon.factories.Factory;
 import org.apache.paimon.factories.FactoryUtil;
 import org.apache.paimon.fs.FileIO;
 import org.apache.paimon.fs.Path;
-import org.apache.paimon.utils.Preconditions;
+
+import javax.annotation.Nullable;
 
 import java.io.IOException;
 import java.io.UncheckedIOException;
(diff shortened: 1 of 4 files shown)


monkeyboy123:
gentle ping @JingsongLi

JingsongLi (changes requested):
Can you modify `CatalogFactory.createCatalog` too? warehousePath and FileIO can be nullable.

And if the warehouse is null, we can get warehouse from `HiveConf`?

JingsongLi:
I think the correct way is `hiveConf.get(HiveConf.ConfVars.METASTOREWAREHOUSE.varname)`.

JingsongLi:
And we need to enrich warehouse path to default scheme of conf if the path does not contain scheme.

monkeyboy123:
> I think the correct way is `hiveConf.get(HiveConf.ConfVars.METASTOREWAREHOUSE.varname)`.

It is good choice, but if we do it like this, we should let user set hive-conf-dir or set env HIVE_CONF_DIR.

monkeyboy123:
> And we need to enrich warehouse path to default scheme of conf if the path does not contain scheme.

I will do it later.

monkeyboy123:
done @JingsongLi

monkeyboy123:
> Hi @monkeyboy123 , can you take a look to #2432 , is that satisfy your requirement?
I will take a look later.

PR #2409, opened 2023-11-28
[core] Fix that tag deletion might delete used data files

yuzelin:
### Purpose

Currently, when we delete a tag, the data manifest entries are not merged, so the result may contains data files that should not be deleted. For example, a tag base is (+A, -A, +B), if we don't merge it, we will get candidate deleted file set (A, B).
This PR fix it.

### Tests

`UncleanedFileStoreExpireTest#testMixedSnapshotAndTagDeletion`


JingsongLi (approved):
+1

PR #2410, opened 2023-11-28
[core]Introduce pub/sub messaging model

### LINKED ISSUE #1811: [Feature] Add a listener mechanism for Paimon
### Search before asking

### Motivation

In the process of integrating services like the Amoro Lakehouse management system with Paimon, it is necessary to timely perceive the state changes of Paimon's tables, such as whether there is data commit, the current compaction status of the write task, and so on. Therefore, a Listener mechanism is introduced to synchronize the state to the outside world.

### Solution

_No response_

### Anything else?

_No response_

hililiwei:
### Purpose

Linked issue: close #1811

Introduce pub/sub messaging model

### Tests

org.apache.paimon.event.TestListeners

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/events/Listener.java
+++ b/paimon-core/src/main/java/org/apache/paimon/events/Listener.java
@@ -1,0 +1,26 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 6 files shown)


hililiwei:
ping @JingsongLi PTAL, thx.

JingsongLi:
Hi @hililiwei , thanks for the contribution, but now we don't have requirement for listeners.

ArthurSXL8:
> Hi @hililiwei , thanks for the contribution, but now we don't have requirement for listeners.

Hi @JingsongLi we encounter a similar case, which is, when two flink jobs write the data into Paimon, how to update the files from other side when lookup join?Could you please explain it in general?

PR #2417, opened 2023-11-29
[oss] FileSystem should be cached by authority

JingsongLi:
### Purpose

Maybe there are oss and oss-hdfs at the same time.


leaves12138 (approved):
+1

PR #2424, opened 2023-11-30
[Improve] Keep a single META-INF directory at the root level for paimon-s3

pegasas:
…paimon-s3

### Purpose

Linked issue: close https://github.com/apache/incubator-paimon/issues/2413


JingsongLi:
Sorry, we need to put classes into sub directory. Just copy META-INFOs

PR #2427, opened 2023-11-30
[Flink] Add log for migration and optimize process if failed

leaves12138:
### Purpose

1 Add log for migration
2 If migration failed, drop the created table.


JingsongLi (approved):
+1

PR #2428, opened 2023-11-30
[Flink] Fix Table Doesn't Exist Exception when Reading System Tables with Flink Generic Catalog

zhuangchong:
### Purpose

Fix Table Doesn't Exist Exception when Reading System Tables with Flink Generic Catalog

image


JingsongLi:
+1

PR #2429, opened 2023-11-30
[filesystem] Copy META-INF directory of oss and s3 to the root of jar

tsreaper:
### Purpose

Release 0.6 RC1 is canceled because META-INF directory of oss and s3 does not appear in the root of jar. This PR fixes the issue.

### Tests

Tested by hand.

### API and Format

No.

### Documentation

No.


zhuangchong (approved):
LGTM.

PR #2432, opened 2023-11-30
[hive] HiveCatalog infer warehouse from hadoop Conf

JingsongLi:
### Purpose

1. Warehouse can be empty in HiveCatalog, it can infer warehouse from hadoop Conf
2. If the scheme of warehouse is empty, get Default Uri from conf to create FileIO


leaves12138 (approved):
+1

PR #2435, opened 2023-12-01
[oss] Use old oss sdk because jindo sdk contains closed-source resources

tsreaper:
### Purpose

This PR reverts oss sdk back to the old one, because jindo sdk contains closed-source resources.

### Tests

Tested by hand.

### API and Format

No.

### Documentation

No.


JingsongLi (approved):
+1

PR #2436, opened 2023-12-01
[fix] fix migration action and procedure name

leaves12138:
### Purpose

Change the action and procedure name of migration


Zouxxyy:
+1

JingsongLi (approved):
+1

PR #2439, opened 2023-12-01
[docs] Fix docs change "-" to "_"

leaves12138:
Fix doc replace all --a-b-c to --a_b_c


yuzelin:
I think we should keep value format.

JingsongLi (approved):
+1

PR #2440, opened 2023-12-01
[flink] All argument keys in help messages of action factory use underline instead of hyphen

yuzelin:
### Purpose

Follow #2437


JingsongLi (approved):
+1

PR #2441, opened 2023-12-03
[spark] Fix insert overwrite with compact error (#2425)

### LINKED ISSUE #2425: [Bug] For primary key table, when use spark insert overwrite some partition and the compact behavior is triggered, IllegalStateException error occurred
### Search before asking

### Paimon version

release-0.5

### Compute Engine

Spark

### Minimal reproduce step

1. create table paimon.paimon_test.test_table(id bigint, data string, dt string) partitioned by (dt) TBLPROPERTIES('num-sorted-run.compaction-trigger' = '1', 'primary-key' = 'dt,id');
2. insert into  paimon.paimon_test.test_table values (1, 'test', '2023-11-30');
3. insert overwrite paimon.paimon_test.test_table partition (dt = '2023-11-30') select id,data from paimon.paimon_test.test_table  where dt = '2023-11-30';

Exception throw:
Caused by: java.lang.IllegalStateException: Trying to delete file data-44539a97-5060-4b5f-9ea1-9f9119452f50-0.parquet which is not previously added. Manifest might be corrupted.
	at org.apache.paimon.utils.Preconditions.checkState(Preconditions.java:204)
	at org.apache.paimon.manifest.ManifestEntry.assertNoDelete(ManifestEntry.java:154)
	at org.apache.paimon.operation.FileStoreCommitImpl.noConflictsOrFail(FileStoreCommitImpl.java:810)

**the conflict file is the one which added by step two**

### Anything else?

_No response_

Xiangakun:
### Purpose
Fix bug: #2425

Linked issue: close #2425

### DIFF
--- a/paimon-spark/paimon-spark-3.3/src/test/scala/org/apache/paimon/spark/sql/InsertOverwriteTest.scala
+++ b/paimon-spark/paimon-spark-3.3/src/test/scala/org/apache/paimon/spark/sql/InsertOverwriteTest.scala
@@ -35,8 +35,7 @@ class InsertOverwriteTest extends PaimonSparkTestBase {
               ""
             }
 
-            spark.sql(
-              s"""
+            spark.sql(s"""
                  |CREATE TABLE T (a INT, b INT, c STRING)
                  |TBLPROPERTIES ($primaryKeysProp 'bucket'='$bucket', 'file.format'='parquet')
                  |""".stripMargin)
@@ -265,8 +264,7 @@ class InsertOverwriteTest extends PaimonSparkTestBase {
 
             spark.sql(s"""
                          |CREATE TABLE T (a INT, b STRING, pt1 STRING, pt2 INT)
-                         |TBLPROPERTIES ($primaryKeysProp 'bucket'='$bucket',
-                         |'num-sorted-run.compaction-trigger'='1')
+                         |TBLPROPERTIES ($primaryKeysProp 'bucket'='$bucket')
                          |PARTITIONED BY (pt1, pt2)
                          |""".stripMargin)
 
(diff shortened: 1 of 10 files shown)


paimon-spark/paimon-spark-3.3/src/test/scala/org/apache/paimon/spark/sql/InsertOverwriteTest.scala
```diff
@@ -37,7 +37,8 @@ class InsertOverwriteTest extends PaimonSparkTestBase {
 
             spark.sql(s"""
                          |CREATE TABLE T (a INT, b INT, c STRING)
-                         |TBLPROPERTIES ($primaryKeysProp 'bucket'='$bucket')
+                         |TBLPROPERTIES ($primaryKeysProp 'bucket'='$bucket',
+                         |'num-sorted-run.compaction-trigger'='1')
```
YannByron, line 41: Do not use existing test cases to test your cases. Add a new one for it.

Zouxxyy:
@Xiangakun see https://github.com/apache/incubator-paimon/pull/2445

Xiangakun:
And, when use parquet format, seems some jar conflicts:

image

Job aborted due to stage failure: Task 0 in stage 3.0 failed 1 times, most recent failure: Lost task 0.0 in stage 3.0 (TID 3) (10.10.128.57 executor driver): java.lang.NoSuchMethodError: org.apache.parquet.Preconditions.checkArgument(ZLjava/lang/String;Ljava/lang/Object;)V

Xiangakun:
Updated, thanks~

Zouxxyy:
@Xiangakun

> since the writeBuilder in bucketMode match { } is not the same with the one after collect

Can you explain it in more detail?

> And your test case does not cover the real compact action, it will not do a compaction since only one file in each bucket

In fact, the trigger of this bug is that when inserting overwrite partition pk table, restoreFiles + new file can reach `num-sorted-run.compaction-trigger` (no need compaction to be done before), my test can just cover it.

JingsongLi (approved):
+1

PR #2450, opened 2023-12-04
[core] Refactory code about MutableObjectIterator and remove  method: E next(E reuse)

leaves12138:
### Purpose

Refactory code about MutableObjectIterator. And remove method:
```code
    /**
     * Gets the next element from the collection. The contents of that next element is put into the
     * given reuse object, if the type is mutable.
     *
     * @param reuse The target object into which to place next element if E is mutable.
     * @return The filled object or <code>null</code> if the iterator is exhausted.
     * @throws IOException Thrown, if a problem occurred in the underlying I/O layer or in the
     *     serialization / deserialization logic
     */
    E next(E reuse) throws IOException;
```

Reason:
* This method is confusing. After code `row = next(row)`, we don't know exactly what `row` is point to. Does it stay itself, or something else.
* Every `MutableObjectIterator` deserves its own reuse object, it is simple to understand and maintain.
* Sometimes `next(reuse)` throws UnSupportException, sometimes `next()` throws that, behavior confusing.

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/data/BinaryRow.java
+++ b/paimon-common/src/main/java/org/apache/paimon/data/BinaryRow.java
@@ -26,6 +26,7 @@ import org.apache.paimon.types.DecimalType;
 import org.apache.paimon.types.LocalZonedTimestampType;
 import org.apache.paimon.types.RowKind;
 import org.apache.paimon.types.TimestampType;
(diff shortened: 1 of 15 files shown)


JingsongLi:
We should do this optimization very carefully, but we don't have time to finish this now.

PR #2451, opened 2023-12-05
[spark] Add 'Paimon' prefix for Call Procedure

YannByron:
### Purpose

Linked issue: https://github.com/apache/incubator-paimon/issues/2449


YannByron:
This is just for `call procedure`. I will comb though the entire Spark subproject to make the package names and class names more reasonable.

JingsongLi (commented):
Thanks @YannByron 
We can rename all classes with Paimon prefix in `org.apache.spark.sql`.

JingsongLi (approved):
+1

PR #2452, opened 2023-12-05
[core] Introduce scan.file-creation-time-millis

JingsongLi:
### Purpose

Old "scan.timestamp-millis":
`Optional timestamp used in case of \"from-timestamp\" scan mode. If there is no snapshot earlier than this time, the earliest snapshot will be chosen.`
The biggest problem is this rely on snapshots, but snapshot only live for 1 hour by default.

Introduce "scan.file-creation-time-millis" for streaming read.
`After configuring this time, only the data files created after this time will be read. It is independent of snapshots, but it is imprecise filtering (depending on whether or not compaction occurs).`


yuzelin (approved):
+1

PR #2461, opened 2023-12-06
[CDC] Refactor cdc to reduce redundant code and set default checkpoint interval

yuzelin:
### Purpose
1. Introduce `SynchronizationActionBase` to abstract synchronization job `build()` pipeline and some common util methods of all synchronization actions.

2 .Introduce `SyncJobHandler` which collects common methods for different cdc source type, so we can reduce many abstract method (we can even delete `MessageQueueSyncDatabaseActionBase`).

3. If user forget to set cp, set default cp interval to 3 minutes.


JingsongLi (approved):
+1

PR #2469, opened 2023-12-07
[flink]Added the function of generating labels when Flink Sink is completed …

siyangzeng:
### Purpose

We need to enhance paimon's capabilities in flink batch processing scenarios, such as generating tags after flink batch processing is completed. This function has a very good practical effect when combined with changelog internally.


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkConnectorOptions.java
```diff
@@ -301,6 +301,13 @@ public class FlinkConnectorOptions {
                     .withDescription(
                             "Sink committer memory to control heap memory of global committer.");
 
+    public static final ConfigOption<Boolean> SINK_FINISH_GENERATAR_TAG =
+            ConfigOptions.key("sink.finish.generatar-tag")
```
JingsongLi, line 305: Can this feature be a part of `tag.automatic-creation`? We can introduce `BATCH` TagCreationMode.
The downside of this is that it cannot be orthogonal to the automatic creation of tags in streaming writing.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/SinkFinishGeneratorTagOperator.java
```diff
@@ -0,0 +1,213 @@
...
+import java.time.Instant;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * Commit {@link Committable} for snapshot using the {@link CommitterOperator}.
+ * When the task is completed, the corresponding tag is generated.
+ */
+public class SinkFinishGeneratorTagOperator<CommitT, GlobalCommitT> implements
```
JingsongLi, line 47: Can you create a ITCase for this feature?

JingsongLi, line 47: Also ut case.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkConnectorOptions.java
siyangzeng: Yes, we hope to use tags in batch processing scenarios and generate corresponding tags after the current task ends.

Do you mean to add a batch mode in tag.automatic-creation without adding new configuration? If that's the case, I think it's okay

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/SinkFinishGeneratorTagOperator.java
siyangzeng: Okay, I'll provide it later

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkConnectorOptions.java
JingsongLi: yes, `tag.automatic-creation=batch`

siyangzeng: Please take another look

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/SinkFinishGeneratorTagOperator.java
```diff
@@ -0,0 +1,216 @@
...
+/**
+ * Commit {@link Committable} for snapshot using the {@link CommitterOperator}. When the task is
+ * completed, the corresponding tag is generated.
+ */
+public class SinkFinishGeneratorTagOperator<CommitT, GlobalCommitT>
+        implements OneInputStreamOperator<CommitT, CommitT>,
+                SetupableStreamOperator,
+                BoundedOneInput {
+
+    private static final String SINK_FINISH_TAG_PREFIX = "sinkFinish-";
```
JingsongLi, line 53: Only concern is here, better name?
Maybe `batch-write-2023-12-13`?

siyangzeng: yes, updated

JingsongLi:
Compile failed?

JingsongLi:
+1 and merged

PR #2472, opened 2023-12-08
[core] Add partition type check

shidayang:
### Purpose

Partition does not support the four types: Map, Array, Row, and Multiset.


JingsongLi (approved):
+1

PR #2473, opened 2023-12-08
[core]Fix partition path error for binary and varbinary partition type

shidayang:
### Purpose
I encounter FileNotFoundException when quering table with varbinary partition.
This is cause by generating file paths directly using the toString of the byte[].

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/utils/StringFormatter.java
+++ b/paimon-common/src/main/java/org/apache/paimon/utils/StringFormatter.java
@@ -1,0 +1,88 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


paimon-core/src/main/java/org/apache/paimon/utils/RowDataPartitionComputer.java
```diff
@@ -52,7 +55,19 @@ public LinkedHashMap<String, String> generatePartValues(InternalRow in) {
 
         for (int i = 0; i < partitionFieldGetters.length; i++) {
             Object field = partitionFieldGetters[i].getFieldOrNull(in);
```
JingsongLi, line 57: Can we introduce a StringFormatter to replace FieldGetter?

shidayang: Good idea

paimon-common/src/main/java/org/apache/paimon/utils/StringFormatter.java
```diff
@@ -0,0 +1,84 @@
...
+            DataType dataType, Function<T, Object> dataGenerator) {
+        switch (dataType.getTypeRoot()) {
+            case BINARY:
+            case VARBINARY:
+                return row -> {
+                    byte[] bytes = (byte[]) dataGenerator.apply(row);
+                    if (bytes == null) {
+                        return null;
+                    }
+                    return StandardCharsets.ISO_8859_1
```
JingsongLi, line 53: I'm not sure if this is the standard to string method, is your PR from a real business? Can we simply prohibit partition fields from being bytes type in the SchemeValidation?

shidayang: This is not from actual business. Not allowing binary to be a partition is a good way to handle it.

PR #2474, opened 2023-12-08
[doc] Add write and write buffer metrics

schnappi17:
### Purpose

Document paimon write and write buffer metrics.


JingsongLi (approved):
+1

PR #2476, opened 2023-12-11
[core] Introduce 'rowkind.field' to define rowkind from data

JingsongLi:
### Purpose

By default, the primary key table determines the row kind according to the input row. You can also define the 
`'rowkind.field'` to use a field to extract row kind.

The valid row kind string should be `'+I'`, `'-U'`, `'+U'` or `'-D'`.


yuzelin (approved):
+1

wqlsdb:
+1

PR #2477, opened 2023-12-11
[flink][spark] Check Exists before create database to avoid Permission issue

JingsongLi:
### Purpose

Create database directly may throwing Permission exception from Hive Metastore.


yuzelin (approved):
+1

PR #2479, opened 2023-12-11
[spark] define PaimonStatistics for the following CBO optimization

YannByron:
### Purpose

Linked issue: https://github.com/apache/incubator-paimon/issues/2404


JingsongLi:
So this is just a refactor PR?

JingsongLi:
> > So this is just a refactor PR?
> 
> yep. And `case class` has a default `equals` method that is needed.

Got it, it is better to explain in the description.

JingsongLi (approved):
+1

PR #2480, opened 2023-12-11
[doc] Added metadata_column parameter usage example in mysql cdc action

zhuangchong:
### Purpose

Added metadata_column parameter usage example in mysql cdc action

Before modification: 
In mysql database cdc, the metadata column is separated by commas; 
In mysql table cdc, the metadata column directly obtains multiple parameters.


paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/action/cdc/CdcActionCommonUtils.java
```diff
@@ -61,7 +61,7 @@ public class CdcActionCommonUtils {
     public static final String PARTITION_KEYS = "partition_keys";
     public static final String PRIMARY_KEYS = "primary_keys";
     public static final String COMPUTED_COLUMN = "computed_column";
-    public static final String METADATA_COLUMN = "metadata_column";
+    public static final String METADATA_COLUMNS = "metadata_columns";
```
JingsongLi, line 64: Are you change API?

zhuangchong: done.

JingsongLi:
API is something we need to focus on considering compatibility, and behavior is a part of API. Unless absolutely necessary, we should not compromise API compatibility.

JingsongLi (approved):
+1

PR #2482, opened 2023-12-11
[spark] Dynamically adjust the parallelism of scan

YannByron:
### Purpose

Linked issue: https://github.com/apache/incubator-paimon/issues/2404

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/ScanHelper.scala
+++ b/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/ScanHelper.scala
@@ -1,0 +1,131 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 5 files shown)


YannByron:
This optimization is similar to `FilePartition.getFilePartitions` in DataSource V1, which adjust the parallelism based on available cores.

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/ScanHelper.scala
```diff
@@ -0,0 +1,131 @@
...
+        case _ => false
+      }
+      val reshuffled = reshuffleSplits0(toReshuffle.collect { case ds: DataSplit => ds })
+      reshuffled ++ reserved
+    } else {
+      splits
+    }
+  }
+
+  private def reshuffleSplits0(splits: Array[DataSplit]): Array[DataSplit] = {
```
JingsongLi, line 56: Can you add a test for this method?

YannByron: OK~

JingsongLi (approved):
+1

PR #2484, opened 2023-12-11
[Flink] kafka adds sasl and other parameters to create consumers when kafka obtains messages and generates schema.

zhuangchong:
…btains messages and generates schema.

### Purpose

kafka adds sasl and other parameters to create consumers when kafka obtains messages and generates schema.

Caused by: org.apache.flink.kafka.shaded.org.apache.kafka.common.errors.TimeoutException: Timeout expired while fetching topic metadata


JingsongLi (approved):
+1

PR #2485, opened 2023-12-12
[cdc] Fix bug that computed columns work abnormally with case-insensitive

yuzelin:
### Purpose

We cannot retrieve schema with case sensitive because the computed columns rely on the original schema. 
I refactor the flow: 
1. case-sensitive is not involved in the validation, schema retrieving and computed columns building phases; 
2. building Paimon schema uses case-sensitive;
3. when parsing records, first we parse all field and evaluate all computed columns, and then we apply case-sensitive conversion.

### Tests

`KafkaCanalSyncTableActionITCase#testComputedColumnWithCaseInsensitive`
`MongoDBSyncTableActionITCase#testComputedColumnWithCaseInsensitive`
`MySqlSyncTableActionITCase#testComputedColumnWithCaseInsensitive`
`MySqlSyncTableActionITCase#testSpecifyKeysWithCaseInsensitive`


JingsongLi (approved):
+1

PR #2487, opened 2023-12-12
[flink] Fix compaction multiple times in compact action.

leaves12138:
### Purpose

Now, compact times depend on the split size. If one partition bucket has multiple split, then Compact Operator will trigger
multiple compactions. This is not correct, and may cause bugs in the future.


JingsongLi (approved):
+1

Issue #2404, opened 2023-11-28
[umbrella][Feature] Improve the performance of querying paimon table on spark 


YannByron:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/incubator-paimon/issues) and found nothing similar.

### Motivation

Improve the performance of querying paimon table on spark.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #2488, opened 2023-12-12
[spark] Add hashcode for PaimonBatch

Zouxxyy:
### Purpose

In `AdaptiveSparkPlanExec`  spark will use TrieMap to cache sparkplan, we need to add hashcode and equals for `PaimonBatch` for plan reuse


JingsongLi (approved):
+1

Zouxxyy:
#2404

PR #2496, opened 2023-12-12
[improvement] Load AWS credentials via provider chain 

### LINKED ISSUE #2495: [Feature] AWS certification should be optional
### Search before asking

### Motivation

In some production scenarios, IAM certifications are controled by external system(eg. [IAM Role for Service Account (IRSA)](https://aws.amazon.com/blogs/containers/diving-into-iam-roles-for-service-accounts/) and [Vault](https://developer.hashicorp.com/vault/docs/auth/aws)). So create catalog with fixed IAM account is limited in some cases.

### Solution

1. optional `s3.access-key` and `s3.screct-key`
2. load AWS certifications via `DefaultAWSCredentialsProviderChain`

### Anything else?

_No response_

XBaith:
### Purpose

Linked issue: close #2495


paimon-filesystems/paimon-s3/src/main/java/org/apache/paimon/s3/S3Loader.java
```diff
@@ -52,14 +49,6 @@ public String getScheme() {
         return "s3";
     }
 
-    @Override
-    public List<String[]> requiredOptions() {
-        List<String[]> options = new ArrayList<>();
```
JingsongLi, line 57: Can here just check `AWSCredentials awsCredentials = DefaultAWSCredentialsProviderChain.getInstance().getCredentials();`?
If there are awsCredentials in env, return no required option. If there is no awsCredentials, return options.

XBaith: Make sense

paimon-filesystems/paimon-s3/pom.xml
```diff
@@ -39,6 +39,12 @@
             <scope>provided</scope>
         </dependency>
 
+        <dependency>
+            <groupId>com.amazonaws</groupId>
```
JingsongLi, line 43: We should not rely on asw in paimon-s3.
All s3 dependencies should just be in paimon-s3-impl.

XBaith: Then how do we check argurments via `AWSCredentials`?

PR #2498, opened 2023-12-13
[spark] Add task metrics

YannByron:
### Purpose

without this:
image

with this:
image


JingsongLi (commented):
Nice improvement.

Can we have UT case to test the metrics from Reader and the Paimon custom metrics?

YannByron:
> Nice improvement.
> 
> Can we have UT case to test the metrics from Reader and the Paimon custom metrics?

hard to add some UTs for this.

JingsongLi (approved):
+1

JingsongLi (approved):
+1

PR #2500, opened 2023-12-13
[flink] Do not create default database if ddl in it is disable

### LINKED ISSUE #2499: [Feature] Do not create default database if ddl in it is disable
### Search before asking

### Motivation

Do not create default database if `disable-create-table-in-default-db` is configured to disable operations in it

### Solution

_No response_

### Anything else?

_No response_

FangYongs:
### Purpose

Linked issue: close #2499 

If `disable-create-table-in-default-db` is true, flink catalog should not create default database

### Tests

1. Added test case in `FlinkCatalogTest`

### API and Format

no

### Documentation

no

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkCatalog.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkCatalog.java
@@ -148,10 +148,12 @@ public class FlinkCatalog extends AbstractCatalog {
         this.logStoreAutoRegister = options.get(LOG_SYSTEM_AUTO_REGISTER);
         this.logStoreAutoRegisterTimeout = options.get(REGISTER_TIMEOUT);
         this.disableCreateTableInDefaultDatabase = options.get(DISABLE_CREATE_TABLE_IN_DEFAULT_DB);
-        if (!catalog.databaseExists(defaultDatabase)) {
-            try {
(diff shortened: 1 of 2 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkCatalog.java
```diff
@@ -148,7 +148,7 @@ public FlinkCatalog(
         this.logStoreAutoRegister = options.get(LOG_SYSTEM_AUTO_REGISTER);
         this.logStoreAutoRegisterTimeout = options.get(REGISTER_TIMEOUT);
         this.disableCreateTableInDefaultDatabase = options.get(DISABLE_CREATE_TABLE_IN_DEFAULT_DB);
-        if (!catalog.databaseExists(defaultDatabase)) {
+        if (!disableCreateTableInDefaultDatabase && !catalog.databaseExists(defaultDatabase)) {
```
JingsongLi, line 151: Maybe here we can just create default database for `FileSystemCatalog`?

JingsongLi (approved):
+1

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkCatalog.java
JingsongLi: ignore me

PR #2501, opened 2023-12-13
[cdc] Support ignore delete records

yuzelin:


JingsongLi:
We can support this in paimon-core for deduplicate merge-engine.

JingsongLi (approved):
+1

PR #2503, opened 2023-12-13
[docs] Fix docs confusing point about catalog create

leaves12138:
### Purpose

Doc code is confusing, there is no need to create catalog in two different methods in two different ways.


JingsongLi (approved):
+1

PR #2505, opened 2023-12-13
[spark] optimize description of scan

YannByron:
### Purpose

image

image

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonBaseScan.scala
+++ b/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonBaseScan.scala
@@ -17,6 +17,7 @@
  */
 package org.apache.paimon.spark
 
+import org.apache.paimon.predicate.{Predicate, PredicateBuilder}
 import org.apache.paimon.spark.sources.PaimonMicroBatchStream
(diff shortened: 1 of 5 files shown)


paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonBaseScan.scala
```diff
@@ -72,4 +95,23 @@ abstract class PaimonBaseScan(table: Table, readBuilder: ReadBuilder, desc: Stri
...
+      ""
+    }
+    val runtimeFiltersStr = if (runtimeFilters.nonEmpty) {
+      ", RuntimeFilters: " + runtimeFilters.mkString(",")
+    } else {
+      ""
+    }
+    super.description() + "["
+    "Fields: " + requiredFieldNameStr + pushedFiltersStr +
+      runtimeFiltersStr +
```
Zouxxyy, line 113: maybe will we don't need requiredFieldNameStr and runtimeFiltersStr

YannByron: They are not always same in case querying nested column or unsupported runtime filter.

Zouxxyy:
+1

JingsongLi (approved):
+1

PR #2506, opened 2023-12-13
[core] Lazy deserialize hive conf in metastoreClientFactory

Zouxxyy:
### Purpose

As distributed open read API, `ReadBuilder` supports serialization to each task, and `Table` inside it also support serialization.

During tpcds test, we found that the `hiveConf`'s serialization in `Table/CatalogEnvironment/metastoreClientFactory` was very time-consuming. 

`CatalogEnvironment` is only used by table commit (driver side), in theory it does not need to be serialized, but flink relies it now.

Since there is currently no good way to separate it from the `Table`,  this PR proposes a lazy serialization solution, with it, the spark tpcds result increased by 15%+


JingsongLi (approved):
+1

Zouxxyy:
#2404

PR #2507, opened 2023-12-14
[cdc] In schema building phase of mysql cdc, the connection should handle jdbc properties

yuzelin:
### Purpose

If we don't handle jdbc driver, the connection might throw exceptions. For example, not handling `useSSL=false` will cause connection refused.


JingsongLi (approved):
+1

PR #2510, opened 2023-12-14
[core]Support create SecurityContext by CatalogContext

shidayang:
### Purpose

Trino uses SecurityContext, but due to the complexity of the environment, the loaded configuration files are often incorrect. Therefore, we should support using user-supplied configuration.


shidayang:
@JingsongLi PTAL

JingsongLi (approved):
+1

PR #2512, opened 2023-12-15
[core] Add auto increase sequence padding

### LINKED ISSUE #2471: [Bug] spark deletes dynamic bucket table, the result is incorrect
### Search before asking

### Paimon version

0.7

### Compute Engine

spark

### Minimal reproduce step

image

For dynamic bucket tables, flink does not support delet.  But spark can do it, but the result is incorrect. 
image
I think the error should be reported in advance or it can be repaired and the result will be correct.

### Anything else?

_No response_

Zouxxyy:
### Purpose

Linked issue: close #2471

Add a new seq pad mode `inc-seq`, pads the sequence field with auto increase sequence, to solve the following problems: 

When user defines `sequence.field`
1. For the same seq (for example, `sequence.field` is timestamp of second, and records in the same second), the order of records cannot be determined.
2. When performing row level changes, `sequence.field` must be the same, and its order cannot be guaranteed.

A better solution is to introduce a new comparison field, this PR is a trade-off, it reuses the `_SEQUENCE_NUMBER` and divides it into two parts: one part stores the auto-increment sequence (upper 32-bits) and the other part stores the user-defined `sequence.field` (transform to long) (lower 32-bits).

Notice !!!
1. Since we only use 32-bits to store user seq, it will be truncated to 32 bits. This will result in loss of accuracy, e.g. for timestamp, only second accuracy is supported, and supports up to 2106-02-07T06:28:15
2. The upper limit of the auto-increment sequence length is 32 bits.

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
@@ -459,7 +459,9 @@ public class CoreOptions implements Serializable {
                                             text(
                                                     "3. \"millis-to-micro\": Pads the sequence field that indicates time with precision of milli-second to micro-second."),
                                             text(
-                                                    "4. Composite pattern: for example, \"second-to-micro,row-kind-flag\"."))
+                                                    "4. \"inc-seq\": Pads the sequence field with auto increase sequence."),
+                                            text(
+                                                    "5. Composite pattern: for example, \"second-to-micro,row-kind-flag\"."))
                                     .build());
 
     public static final ConfigOption<StartupMode> SCAN_MODE =
@@ -1338,6 +1340,10 @@ public class CoreOptions implements Serializable {
         return Arrays.asList(padding.split(","));
     }
 
+    public boolean incSeqPadding() {
+        return SequenceAutoPadding.INC_SEQ.value.equals(options.get(SEQUENCE_AUTO_PADDING));
+    }
+
     public boolean writeOnly() {
         return options.get(WRITE_ONLY);
     }
@@ -1933,7 +1939,8 @@ public class CoreOptions implements Serializable {
                 "Pads the sequence field that indicates time with precision of seconds to micro-second."),
         MILLIS_TO_MICRO(
                 "millis-to-micro",
-                "Pads the sequence field that indicates time with precision of milli-second to micro-second.");
+                "Pads the sequence field that indicates time with precision of milli-second to micro-second."),
+        INC_SEQ("inc-seq", "Pads the sequence field with auto increase sequence.");
 
         private final String value;
         private final String description;
--- a/paimon-core/src/main/java/org/apache/paimon/mergetree/MergeTreeWriter.java
+++ b/paimon-core/src/main/java/org/apache/paimon/mergetree/MergeTreeWriter.java
@@ -50,6 +50,9 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import static org.apache.paimon.table.sink.SequenceGenerator.INC;
+import static org.apache.paimon.table.sink.SequenceGenerator.INC_SEQ_MASK;
+
 /** A {@link RecordWriter} to write records and generate {@link CompactIncrement}. */
 public class MergeTreeWriter implements RecordWriter<KeyValue>, MemoryOwner {
 
@@ -72,10 +75,11 @@ public class MergeTreeWriter implements RecordWriter<KeyValue>, MemoryOwner {
     private final LinkedHashSet<DataFileMeta> compactAfter;
     private final LinkedHashSet<DataFileMeta> compactChangelog;
 
-    private long newSequenceNumber;
+    private long nextIncSequenceNumber;
     private WriteBuffer writeBuffer;
 
-    private WriterMetrics writerMetrics;
+    private final WriterMetrics writerMetrics;
+    private final boolean incSeqPadding;
 
     public MergeTreeWriter(
             boolean writeBufferSpillable,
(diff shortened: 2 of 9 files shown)


paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -459,7 +459,9 @@ public class CoreOptions implements Serializable {
                                             text(
                                                     "3. \"millis-to-micro\": Pads the sequence field that indicates time with precision of milli-second to micro-second."),
                                             text(
-                                                    "4. Composite pattern: for example, \"second-to-micro,row-kind-flag\"."))
+                                                    "4. \"inc-seq\": Pads the sequence field with auto increase sequence."),
```
JingsongLi, line 462: So for inc-seq, int, long must be a second? You should document this.

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -459,7 +459,9 @@ public class CoreOptions implements Serializable {
...
+                                            text(
+                                                    "5. Composite pattern: for example, \"second-to-micro,row-kind-flag\"."))
```
JingsongLi, line 464: `inc-seq, row-kind-flag` is supported? I think maybe this should be most popular composite pattern

paimon-core/src/main/java/org/apache/paimon/mergetree/MergeTreeWriter.java
```diff
@@ -89,14 +93,22 @@ public MergeTreeWriter(
...
         this.ioManager = ioManager;
         this.keyType = writerFactory.keyType();
         this.valueType = writerFactory.valueType();
         this.compactManager = compactManager;
-        this.newSequenceNumber = maxSequenceNumber + 1;
+        this.incSeqPadding = incSeqPadding;
+        if (maxSequenceNumber == -1) {
+            this.nextIncSequenceNumber = 0;
+        } else if (incSeqPadding) {
+            this.nextIncSequenceNumber = (maxSequenceNumber & INC_SEQ_MASK) + INC;
```
JingsongLi, line 108: We should merge this into `SequenceGenerator`. There should be some refactor.

paimon-spark/paimon-spark-common/src/test/scala/org/apache/paimon/spark/sql/DeleteFromTableTest.scala
```diff
@@ -180,4 +180,24 @@ class DeleteFromTableTest extends PaimonSparkTestBase {
...
+
+    spark.sql("INSERT INTO T VALUES (1, 'a', CAST('2023-12-15 10:00:00' AS TIMESTAMP))")
+
+    spark.sql("INSERT INTO T VALUES (2, 'b', CAST('2023-12-15 10:00:01' AS TIMESTAMP))")
+
+    spark.sql("INSERT INTO T VALUES (3, 'c', CAST('2023-12-15 10:00:02' AS TIMESTAMP))")
+
+    spark.sql("DELETE FROM T WHERE id = 1")
+
+    val rows1 = spark.sql("SELECT * FROM T").collectAsList()
```
YannByron, line 199: please use `checkDataset` or `checkAnswer` to replace `collectAsList + assertThat`.

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
Zouxxyy: `inc-seq` + `row-kind-flag` will further reduce the accuracy, will there be a situation where +U come first, and then -U come, at the same second?

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
JingsongLi, line 462: seconds-and-inc

hekaifei:
@Zouxxyy  Hello, if you have time, please kindly solve this problem, I urgently need, thank you

hekaifei:
@Zouxxyy  If the amount of data is relatively large, the number of data items reaches 5 billion, and auto-increment is only 32 bits. What will happen if it is not enough?

Zouxxyy:
> @Zouxxyy If the amount of data is relatively large, the number of data items reaches 5 billion, and auto-increment is only 32 bits. What will happen if it is not enough?

System will crash, the current design only supports a single bucket with 2,147,483,647 items. You have 5 billion one bucket or total?

hekaifei:
@Zouxxyy   right now not a bucket of 5 billion, just a few million,  but the job keeps running and hasn't restarted, which is hard to say.

I think we should put userSeq on (upper 32-bits) and auto-increment on (lower 32-bits).

hekaifei:
@Zouxxyy right now not a bucket of 5 billion, just a few million, but the job keeps running and the data is updated frequently, which is hard to say.

I think we should put userSeq on (upper 32-bits) and auto-increment on (lower 32-bits).

hekaifei:
@Zouxxyy A few million per bucket right now.  Worry about job keeps running and the data is updated frequently, 32 bit is not enough

hekaifei:
> > @Zouxxyy If the amount of data is relatively large, the number of data items reaches 5 billion, and auto-increment is only 32 bits. What will happen if it is not enough?
> 
> @hekaifei System will crash, the current design only supports a single bucket with 2,147,483,647 items. You have 5 billion one bucket or total?

@Zouxxyy A few million per bucket right now. Worry about job keeps running and the data is updated frequently, 32 bit is not enough

hekaifei:
@Zouxxyy   There are several tables with more than 5 billion data, divided into 1,000 to 2,000 buckets

hekaifei:
@Zouxxyy  I think userSeq should be upper 32-bits, otherwise the same key inserted later and userSeq is smaller than before, Then the before data will be overwritten, This is not expected

hekaifei:
@Zouxxyy  I think userSeq should be upper 32-bits, otherwise the same key inserted later and userSeq is smaller than before, Then the before data will be overwritten, This is not expected

Zouxxyy:
@hekaifei 
> A few million per bucket right now. Worry about job keeps running and the data is updated frequently, 32 bit is not enough

yes, frequent modifications will also cause the solution to collapse

> otherwise the same key inserted later and userSeq is smaller than before, Then the before data will be overwritten

when you want to use row level update, you must use 'merge-engine' = 'deduplicate', means that the larger userSeq overrides the smaller one

hekaifei:
@Zouxxyy  
> when you want to use row level update, you must use 'merge-engine' = 'deduplicate', means that the larger userSeq overrides the smaller one
In the inc-seq solution, the upper 32 bits are auto-increasing sequence, which will bring about  the smaller userSeq overrides the larger one

hekaifei:
@Zouxxyy 
> when you want to use row level update, you must use 'merge-engine' = 'deduplicate', means that the larger userSeq overrides the smaller one

In the inc-seq solution, the upper 32 bits are auto-increasing sequence, which will bring about the smaller userSeq overrides the larger one

JingsongLi:
The final solution should be https://github.com/apache/incubator-paimon/pull/2811

I will close this PR, this PR is a temporary solution.

PR #2513, opened 2023-12-15
[flink] fix bug for `show partitions` with date and time type

zhangjun0x01:
### Purpose

now ,`show partitions` with date type partition will show number, we should show correct date format.


JingsongLi (commented):
I know yes, it's bad here, but due to compatibility issues we kept the date display, not so good to change it here to show it this way, which will cause some experiential issues.

JingsongLi:
Thanks @zhangjun0x01 , but I think it is better to think whole story here.

PR #2515, opened 2023-12-15
[core] Fix creation failure caused by savepoint tag

MonsterChenzhuo:
### Purpose
Fix creation failure caused by savepoint tag

4191702609269_ pic


JingsongLi:
Can you provide a test? And you need to fix tag expire too.

MonsterChenzhuo:
@JingsongLi PTAL,thanks.

JingsongLi (approved):
+1

PR #2516, opened 2023-12-15
[filesystem] Correct the gson dependency after rolling back from jindo sdk to oss sdk

carp84:
### Purpose

When rolling back oss sdk from jindo sdk to the old one through #2435 we missed to roll back the gson related dependency, and this PR aims to fix this up. More details refer to #2290 

### Tests

Manually test and make sure the `META-INF/NOTICE` file correctly reflect the dependencies of the `paimon-filesystems/paimon-oss-impl` and `paimon-filesystems/paimon-hadoop-shaded` modules.

### API and Format

No API or storage format change involved.

### Documentation

No documentation required.


JingsongLi (approved):
Looks good to me! Thanks @carp84

PR #2517, opened 2023-12-15
[Spark] Merge Into: When Not Matched By Source

YannByron:
### Purpose

Linked issue: close #xxx


JingsongLi (commented):
Before this PR, can we create an abstraction for `MergeIntoTableTest`. Just like `TestTrinoITCase`, every spark module has a class to extend this, test merge into feature for every Spark versions.

JingsongLi:
https://github.com/apache/incubator-paimon-trino/blob/main/paimon-trino-common/src/test/java/org/apache/paimon/trino/TestTrinoITCase.java
And 
https://github.com/apache/incubator-paimon-trino/blob/main/paimon-trino-388/src/test/java/org/apache/paimon/trino/TestTrino388ITCase.java

JingsongLi (approved):
+1

PR #2521, opened 2023-12-18
[spark] Support 'show partitions' sql

XuQianJin-Stars:
### Purpose

fix 'show partitions' sql's bug

image

1.  create table with partition's field type is not the same
```
CREATE TABLE T (a VARCHAR(10), b CHAR(10),c BIGINT,dt LONG,hh VARCHAR(4))
            PARTITIONED BY (dt, hh)
            TBLPROPERTIES ('primary-key'='a,b,dt,hh', 'bucket'='3')
```
2. insert some data
```
INSERT INTO T VALUES('a','b',1,20230816,'1132')
INSERT INTO T VALUES('a','b',1,20230816,'1133')
INSERT INTO T VALUES('a','b',1,20230816,'1134')
```
3. show partitions xxxx PARTITION(xxxx)
```
show partitions T PARTITION (hh='1132')
```

### Tests

`PaimonPartitionManagementTest`


paimon-spark/paimon-spark-common/src/test/java/org/apache/paimon/spark/SparkWriteITCase.java
```diff
@@ -237,4 +237,36 @@ public void testNonnull() {
             spark.sql("DROP TABLE IF EXISTS S");
         }
     }
+
+    @Test
+    public void testShowPartitionsPartitioned() {
```
YannByron, line 242: Maybe put these in a new test class named `ShowCommandTest` which will contain case related to `SHOW` commands.

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/catalyst/analysis/PaimonIncompatiblePHRRules.scala
```diff
@@ -43,6 +43,11 @@ case class PaimonIncompatiblePHRRules(session: SparkSession) extends Rule[Logica
         }.toMap
         PaimonTruncateTableCommand(table, partitionSpec)
 
+      case t @ ShowPartitions(PaimonRelation(table), specOpt, _) if t.resolved =>
+        PaimonShowTablePartitionsCommand(
+          table,
+          specOpt.map(s => s.asInstanceOf[ResolvedPartitionSpec]))
```
YannByron, line 49: try this, maybe it can work too:
```
import org.apache.spark.sql.execution.datasources.v2.DataSourceV2Implicits._
specOpt.asResolvedPartitionSpecs
```

XuQianJin-Stars: > asResolvedPartitionSpecs

This is useful for `Seq[PartitionSpec]`.

YannByron:
+1 for nice catch.

PR #2522, opened 2023-12-18
[core] Add "snapshot.expire.clean-empty-directories" to determine whether to clean empty directories when expiring snapshots

zqw142857:
Add a parameter named "data-directories.ignore-clean" to determine whether to ignore the clear data directories.

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
@@ -982,6 +982,12 @@ public class CoreOptions implements Serializable {
                     .defaultValue(MemorySize.ofMebiBytes(10))
                     .withDescription("The threshold for read file async.");
 
+    public static final ConfigOption<Boolean> DATA_DIRECTORIES_IGNORE_CLEAN =
+            key("data-directories.ignore-clean")
+                    .booleanType()
+                    .defaultValue(false)
+                    .withDescription("Whether to ignore clean data directories.");
+
     private final Options options;
 
     public CoreOptions(Map<String, String> options) {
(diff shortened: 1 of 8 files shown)


paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -982,6 +982,12 @@ public class CoreOptions implements Serializable {
                     .defaultValue(MemorySize.ofMebiBytes(10))
                     .withDescription("The threshold for read file async.");
 
+    public static final ConfigOption<Boolean> DATA_DIRECTORIES_IGNORE_CLEAN =
+            key("data-directories.ignore-clean")
```
yuzelin, line 986: Please using "snapshot.expire.ignore-clean-directories"

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -982,6 +982,12 @@ public class CoreOptions implements Serializable {
...
+                    .booleanType()
+                    .defaultValue(false)
+                    .withDescription("Whether to ignore clean data directories.");
```
yuzelin, line 989: 1. Please use key "snapshot.expire.clean-empty-directories" and default value "true".
2. Please make description more clearer, for example, you can use "Whether to try to clean empty directories when expiring snapshots. Note that trying to clean directories might throw exceptions in filesystem, but in most cases it won't cause problems."

yuzelin (commented):
And please add a test in `FileDeletionTest` to test that empty directories won't be deleted when setting this option.

yuzelin:
Fixed checkstyle problem.

yuzelin:
Generator doc.

yuzelin (approved):
+1

PR #2536, opened 2023-12-19
[core] Optimize HashBucketAssigner to more parallilism with less initialized bucket

JingsongLi:
### Purpose

Assuming we need to write multiple partitions simultaneously:

- At present, setting the dynamic bucket assignment parallelism option too large can result in a large number of buckets.
- Setting it too small will pose a risk to memory and make it prone to OOM

This PR aims to decouple initialized buckets and total parallelism of assigners to avoid the aforementioned issues

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/DynamicBucketSink.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/DynamicBucketSink.java
@@ -58,22 +58,17 @@ public abstract class DynamicBucketSink<T> extends FlinkWriteSink<Tuple2<T, Inte
 
         // Topology:
         // input -- shuffle by key hash --> bucket-assigner -- shuffle by partition & bucket -->
-        // writer -->
-        // committer
+        // writer --> committer
 
         // 1. shuffle by key hash
-        Integer numAssigners = table.coreOptions().dynamicBucketAssignerParallelism();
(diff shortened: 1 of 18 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/DynamicBucketSink.java
```diff
@@ -62,17 +62,23 @@ public DataStreamSink<?> build(DataStream<T> input, @Nullable Integer parallelis
...
         }
+
+        Integer totalAssigners = numAssigners;
+        if (parallelism != null && parallelism > totalAssigners) {
+            totalAssigners = parallelism;
+        }
+
         DataStream<T> partitionByKeyHash =
-                partition(input, channelComputer1(), assignerParallelism);
+                partition(input, assignerChannelComputer(numAssigners), totalAssigners);
```
liming30, line 76: This is a little weried for me. From the name, `dynamic-bucket.assigner-parallelism` should control the parallelism of the assigner operator, but it is not anymore. If I want to increase the parallelism of the assigner operator, then I can only increase the parallelism of the sink? 

How about introducing a new configuration to control the initial number of buckets?

JingsongLi: I agree with you, your point is better.

liming30 (approved):
+1

PR #2537, opened 2023-12-20
[Flink] remove schemaFileExists check for showing tables  when using HiveCatalog

wg1026688210:
Reduce the number of accesses to Hive and filesystem for showing tables  when using HiveCatalog


JingsongLi (approved):
+1

PR #2539, opened 2023-12-20
[flink] Optimize Flink Lookup Join: Bulk load for initialization

JingsongLi:
### Purpose

When initializing, it is necessary to scan the entire table and then write one by one to RocksDB. This process is too time-consuming because RocksDB writes slowly.

We can optimize it to bulkload mode, sort it first before writing it during full initialization.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/lookup/BulkLoader.java
+++ b/paimon-core/src/main/java/org/apache/paimon/lookup/BulkLoader.java
@@ -86,7 +86,9 @@ public class BulkLoader {
             }
 
             if (files.size() > 0) {
-                db.ingestExternalFile(columnFamily, files, new IngestExternalFileOptions());
+                IngestExternalFileOptions ingestOptions = new IngestExternalFileOptions();
+                db.ingestExternalFile(columnFamily, files, ingestOptions);
+                ingestOptions.close();
             }
         } catch (RocksDBException e) {
             throw new RuntimeException(e);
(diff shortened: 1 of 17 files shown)


paimon-core/src/main/java/org/apache/paimon/lookup/BulkLoader.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    public void finish() {
+        try {
+            if (writer != null) {
+                writer.finish();
+            }
+
+            if (files.size() > 0) {
+                db.ingestExternalFile(columnFamily, files, new IngestExternalFileOptions());
```
Aitozi, line 89: Should we release the `IngestExternalFileOptions` after ingest or just reuse one object?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/FileStoreLookupFunction.java
```diff
@@ -145,9 +154,41 @@ private void open() throws Exception {
                         options.get(LOOKUP_CACHE_ROWS));
         this.nextLoadTime = -1;
         this.streamingReader = new TableStreamingReader(table, projection, this.predicate);
+        bulkLoad(options);
+    }
 
-        // do first load
-        refresh();
+    private void bulkLoad(Options options) throws Exception {
+        BinaryExternalSortBuffer bulkLoadSorter =
```
Aitozi, line 161: Is this sort necessary to the primary key index lookup? Since, it's already sorted

Aitozi (approved):
+1

PR #2541, opened 2023-12-20
[core] remove Dynamic Bucket Row

leaves12138:
### Purpose

Remove DynamicBucketRow, add another interfacre `write(row, bucket)` in replace

Reason:
* DynamicBucketRow wrapper now is a per-record action, it has a poor performance, per-record `new DynamicBucketRow(row)` will slow down inserting
* DynamicBucketRow is an intrusion of current InternalRow system. The sub-class of InternalRow should not hide information itself. After serialize and deserialize, the bucket information in DynamicBucketRow will lose.


paimon-core/src/main/java/org/apache/paimon/table/sink/TableWriteImpl.java
```diff
@@ -139,6 +150,15 @@ private SinkRecord toSinkRecord(InternalRow row) {
                 row);
     }
 
+    private SinkRecord toSinkRecord(InternalRow row, int bucket) {
```
JingsongLi, line 153: `private SinkRecord toSinkRecord(InternalRow row)` should use this method.

JingsongLi (approved):
+1

PR #2542, opened 2023-12-20
[doc]Change to Trino documentation regarding installation

shidayang:
### Purpose
We need modify Trino doc regarding installation methods after this pr https://github.com/apache/incubator-paimon-trino/pull/50  merged.


JingsongLi (approved):
+1

PR #2543, opened 2023-12-20
[flink] Fix that schema validation fails when using 'scan.x' options to do time travel on schema changed tables

yuzelin:
### API and Format

`BatchFileStoreITCase#testScanFromOldSchema`
`testScanFromOldSchema#testScanFromOldSchema`


JingsongLi (approved):
+1

PR #2545, opened 2023-12-20
[core] The agg table model supports replace types

MonsterChenzhuo:
### Purpose

The agg table model supports replace types


MonsterChenzhuo:
@JingsongLi  PATL,thanks.

paimon-core/src/main/java/org/apache/paimon/mergetree/compact/aggregate/FieldReplaceAgg.java
```diff
@@ -0,0 +1,41 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.paimon.mergetree.compact.aggregate;
+
+import org.apache.paimon.types.DataType;
+
+/** replace aggregate a field of a row. */
+public class FieldReplaceAgg extends FieldAggregator {
```
Aitozi, line 24: It's same to `last_value` ?

PR #2550, opened 2023-12-21
[core] Fix that some tag creations haven't handle tag callbacks

yuzelin:
### Purpose

Let TagManager#createTag to handle callbacks.


paimon-core/src/main/java/org/apache/paimon/table/AbstractFileStoreTable.java
```diff
@@ -284,7 +281,8 @@ public TableCommitImpl newCommit(String commitUser) {
     }
 
     private List<CommitCallback> createCommitCallbacks() {
-        List<CommitCallback> callbacks = new ArrayList<>(loadCommitCallbacks());
+        List<CommitCallback> callbacks =
+                new ArrayList<>(CallbackUtils.loadCommitCallbacks(coreOptions()));
```
JingsongLi, line 285: Why not just using AbstractFileStore.createTagCallbacks?

JingsongLi (approved):
+1

PR #2553, opened 2023-12-21
[doc] Aggregation model function document optimization

MonsterChenzhuo:
### Purpose

Aggregation model function document optimization.


MonsterChenzhuo:
@JingsongLi PTAL,thanks.

JingsongLi (approved):
+1

PR #2554, opened 2023-12-21
[metric]Add running compaction count metric

shidayang:


shidayang:
@schnappi17 PTAL

JingsongLi:
Thanks @shidayang for the contribution, there is only one thread, so the running compaction should be 1 or 0... This metric looks not so solid.

PR #2555, opened 2023-12-21
[flink] build schema after match in the sync action

zhangjun0x01:
### Purpose

We should build the schema after we do the monitor table check

Otherwise, if there are thousands of tables but we only want to synchronize a few of them, it will build thousands of schemas, which is useless and affects performance, and may even throw exceptions during the process of building schemas


JingsongLi (approved):
+1

PR #2556, opened 2023-12-22
[flink] add limit partition request for query

zhangjun0x01:
### Purpose

Add a parameter to limit the number of partitions for user queries to avoid full table scan when the table is too large

refer to hive `hive.metastore.limit.partition.request`

https://cwiki.apache.org/confluence/display/hive/configuration+properties#ConfigurationProperties-hive.metastore.limit.partition.request


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkConnectorOptions.java
```diff
@@ -301,6 +301,14 @@ public class FlinkConnectorOptions {
                     .withDescription(
                             "Sink committer memory to control heap memory of global committer.");
 
+    public static final ConfigOption<Integer> LIMIT_PARTITION_REQUEST =
```
JingsongLi, line 304: Can this be a generic option for all compute engines?

zhangjun0x01: ok。I will add it to `CoreOptions`, and support spark.

zhangjun0x01: I updated it

JingsongLi:
This has been removed hive 3.x https://cwiki.apache.org/confluence/display/Hive/Configuration+Properties#ConfigurationProperties-hive.limit.query.max.table.partition

PR #2565, opened 2023-12-22
[core] The agg table model supports count types

MonsterChenzhuo:
### Purpose

The agg table model supports count types.


JingsongLi (approved):
+1

PR #2572, opened 2023-12-26
[Core]Support product function for agg table

TaoZex:
### Purpose

Support product function for agg table


TaoZex:
Flink CDC Tests of CI failed. It doesn't seem to be the cause of the code and needs to be rerun.
image

JingsongLi (approved):
+1

PR #2577, opened 2023-12-26
[doc] Doc agg table function description unified format

MonsterChenzhuo:
### Purpose

Doc agg table function description unified format.


JingsongLi (approved):
+1

PR #2580, opened 2023-12-27
[flink] Open write-buffer default for batch mode

leaves12138:
### Purpose

Open writer-buffer for append batch mode by default to avoid out-of-memory issues.

### API and Format

No influence to user.


JingsongLi:
This will result to large number of small files.

PR #2584, opened 2023-12-27
[flink] Fix delete exception with lookup changelog-producer and pk as delete key(push down to client).

leaves12138:
### Purpose
Table a is a primary key table with lookup changelog-producer
Delete from table a where pk = xxx will cause an error
Fix this problem


paimon-core/src/main/java/org/apache/paimon/table/TableUtils.java
```diff
@@ -49,7 +53,10 @@ public static long deleteWhere(Table table, List<Predicate> filters) {
         long hit = 0;
         try (RecordReader<InternalRow> reader = readBuilder.newRead().createReader(splits);
                 BatchTableWrite write = writeBuilder.newWrite();
+                // we create temp io manager to for writer
```
yuzelin, line 56: fix comment

yuzelin (approved):
+1

JingsongLi (approved):
+1

PR #2585, opened 2023-12-27
[core] Introduce consumer.ignore-progress

JingsongLi:
### Purpose

Consumer-id:

When previous job is stopped, the newly started job can continue to consume from the previous progress without resuming from the state. The newly reading will start reading from next snapshot id found in consumer files. If you don’t want this behavior, you can set 'consumer.ignore-progress' to true.


yuzelin (approved):
+1

PR #2586, opened 2023-12-27
[core] Primary key table only max level can convert to raw files

shidayang:
### Purpose
1. Append only table always can convert to raw files
2. Primary key table only max level can convert to raw files, Because other level may contain data with RowKind not INSERT, other engines cannot natively read it.

The main change compared to before is that individual level 0 files can not be convert to raw files


shidayang:
@JingsongLi PTAL, I am not sure if this change is as expected. If this change is not added, other engines will need to handle the _VALUE_KIND field.

JingsongLi (approved):
+1

JingsongLi:
@shidayang Good catch, I think this is a real problem.

PR #2592, opened 2023-12-28
[flink] Throws exception if there is a SinkMaterializer before Paimon sink

tsreaper:
### Purpose

Flink's `SinkMaterializer` is an undesired feature for Paimon, because it will keep all the keys in operator's state and also filter out delete records, which may cause incorrect result.

This PR throws an exception if there is a `SinkMaterializer` before Paimon sink. Thus the user will not forget to remove `SinkMaterializer`.

### Tests

* `PartialUpdateITCase#testNoSinkMaterializer`.

### API and Format

No.

### Documentation

No.


JingsongLi (approved):
+1

PR #2595, opened 2023-12-28
[hive] Improve hive database is exists yet create/drop database

XuQianJin-Stars:
### Purpose

Improve hive database is exists yet create/drop database.

### DIFF
--- a/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
+++ b/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
@@ -216,7 +216,7 @@ public class HiveCatalog extends AbstractCatalog {
     }
 
     @Override
-    protected void createDatabaseImpl(String name) {
+    public void createDatabaseImpl(String name) throws DatabaseAlreadyExistException {
         try {
             Database database = getDatabase(name);
             if (database == null) {
@@ -227,6 +227,8 @@ public class HiveCatalog extends AbstractCatalog {
                 database.setName(name);
                 locationHelper.specifyDatabaseLocation(databasePath, database);
                 client.createDatabase(database);
+            } else {
+                throw new DatabaseAlreadyExistException(name);
             }
         } catch (TException | IOException e) {
             throw new RuntimeException("Failed to create database " + name, e);
@@ -234,7 +236,7 @@ public class HiveCatalog extends AbstractCatalog {
     }
 
     @Override
-    protected void dropDatabaseImpl(String name) {
+    public void dropDatabaseImpl(String name) {
         try {
             Database database = getDatabase(name);
             if (database != null) {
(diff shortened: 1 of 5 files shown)


paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
```diff
@@ -221,13 +218,16 @@ protected boolean databaseExistsImpl(String databaseName) {
...
         try {
-            Path databasePath = newDatabasePath(name);
-            locationHelper.createPathIfRequired(databasePath, fileIO);
-
-            Database database = new Database();
-            database.setName(name);
-            locationHelper.specifyDatabaseLocation(databasePath, database);
-            client.createDatabase(database);
+            Database database = getDatabase(name);
+            if (database == null) {
```
JingsongLi, line 222: If database not null, here should throw `DatabaseAlreadyExistException`?
We can add test in `CatalogTestBase`.

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
```diff
@@ -208,43 +208,58 @@ public List<String> listDatabases() {
...
-            Database database = client.getDatabase(name);
-            String location = locationHelper.getDatabaseLocation(database);
-            locationHelper.dropPathIfRequired(new Path(location), fileIO);
-            client.dropDatabase(name, true, false, true);
+            Database database = getDatabase(name);
+            if (database != null) {
+                String location = locationHelper.getDatabaseLocation(database);
+                locationHelper.dropPathIfRequired(new Path(location), fileIO);
+                client.dropDatabase(name, true, false, true);
+            }
```
JingsongLi, line 246: else throw `DatabaseNotExistException`.

JingsongLi (commented):
Thanks @XuQianJin-Stars for your update, left comments.

JingsongLi (approved):
Looks good to me!

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
Zouxxyy: No need to add   Database database = getDatabase(name);  if (database == null) {...} here,

see 

```java
    public void createDatabase(String name, boolean ignoreIfExists, Map<String, String> properties)
        if (isSystemDatabase(name)) {
            throw new ProcessSystemDatabaseException();
        }
        if (databaseExists(name)) {
            if (ignoreIfExists) {
                return;
            }
            throw new DatabaseAlreadyExistException(name);
        }
        createDatabaseImpl(name, properties);
    }
```
user should use createDatabase instead of createDatabaseImpl

Zouxxyy: this pr can be revert actually

PR #2599, opened 2023-12-29
[core] Introduce product IT case

MonsterChenzhuo:
### Purpose

Introduce product IT case

### DIFF
--- a/paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/PreAggregationITCase.java
+++ b/paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/PreAggregationITCase.java
@@ -853,6 +853,157 @@ public class PreAggregationITCase {
         }
     }
 
+    /** ITCase for product aggregate function. */
+    public static class ProductAggregation extends CatalogITCaseBase {
+        @Override
+        protected List<String> ddl() {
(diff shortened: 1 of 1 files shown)


paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/PreAggregationITCase.java
```diff
@@ -853,6 +853,157 @@ public void testStreamingRead() {
         }
     }
 
+    /** ITCase for product aggregate function. */
+    public static class ProductAggregation extends CatalogITCaseBase {
+        @Override
+        protected List<String> ddl() {
+            return Collections.singletonList(
+                    "CREATE TABLE IF NOT EXISTS T1 ("
+                            + "j INT, k INT, "
```
yuzelin, line 862: Why not just set a single primary key?

MonsterChenzhuo: 图片
Refer to the implementation of sum

yuzelin: Alright.

yuzelin (approved):
+1

PR #2601, opened 2023-12-29
[flink] upgrade to Junit5 in flink module

zhangjun0x01:
### Purpose

Linked issue: close #xxx


paimon-flink/paimon-flink-1.14/src/test/java/org/apache/paimon/flink/BatchFileStoreITCase.java
```diff
@@ -21,7 +21,7 @@
 import org.apache.paimon.CoreOptions;
 
 import org.apache.flink.types.Row;
-import org.junit.Test;
+import org.junit.jupiter.api.Test;
```
JingsongLi, line 24: Can you test in your ide?
Here should be `java.lang.NullPointerException`.

And no test will be run in 1.14.

JingsongLi (changes requested):

PR #2602, opened 2023-12-29
[hotfix] Fix unstable test ProductAggregation#testMergeInMemory caused by disorder and scale

yuzelin:
### Purpose

This is because if insert values have `CAST`, it will cause disorder (cannot be solved by setting parallelism to 1); And in this test, the old DECIMAL(4, 2) will truncate result, so the final result will be affected by the order of input.


leaves12138 (approved):
+1

PR #2605, opened 2023-12-30
[flink] Support Flink look up based on LSM

### LINKED ISSUE #1812: [Feature] Support Flink look up based on LSM
### Search before asking

### Motivation

Currently in paimon, flink look up join is leveraging rocksdb's backend storage capabilities. This is a conversion from the original memory to back-end storage, and the timeliness is not inferior.
However, each taskmansger, that is, each process, loads one piece of dimension table data, and if the dimension table is at the level of billions and the task requires dozens or more Taskmanagers to process the data, then dozens or more taskmanager dimension table data will need to be loaded repeatedly.  
With the ordered storage capabilities of the current LSM, can this problem of repeated loading be solved while also meeting the timeliness requirements in general?

### Solution

_No response_

### Anything else?

_No response_

Aitozi:
### Purpose

Linked issue: close #1812

This PR is meant to introduce a LRU cache mode for Lookup Join, which will avoid loading all the data to TaskManager eagerly.

Currently limitation:

- Only support on Primary Key table and join key should equal to the primary key
- Only support on the fixed bucket table
- Only support the table with `'changelog-producer' = 'lookup'`


JingsongLi (commented):
Thanks @Aitozi for the contribution!

I realized that switching to Lookup queries would result in losing the projection and filter.
I may should spend more time to think about this.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/PrimaryKeyLRULookupTable.java
```diff
@@ -0,0 +1,226 @@
...
+
+            if (lookupLevels == null) {
+                Preconditions.checkArgument(
+                        before.isEmpty(), "The before file should be empty for the initial phase.");
+                Levels levels = new Levels(keyComparatorSupplier.get(), after, options.numLevels());
+
+                KeyValueFileReaderFactory factory = readerFactoryBuilder.build(partition, bucket);
+
+                lookupLevels =
+                        new LookupLevels(
```
JingsongLi, line 172: Can you add an interface for lookup? For example:
1. Add `FileStoreTable.newQuery(@Nullable int[] projection)`.
2. `PrimaryKeyFileStoreTable` override `newQuery`.

```
public interface TableQuery extends Closeable {

    void refreshFiles(
            BinaryRow partition,
            int bucket,
            List<DataFileMeta> beforeFiles,
            List<DataFileMeta> dataFiles);

    @Nullable
    InternalRow lookup(BinaryRow partition, int bucket, InternalRow key) throws IOException;
}
```

This can be a separate PR.

Aitozi: Updated in the latest commit.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/PrimaryKeyLRULookupTable.java
```diff
@@ -0,0 +1,117 @@
...
+            List<DataFileMeta> before = ((DataSplit) split).beforeFiles();
+            List<DataFileMeta> after = ((DataSplit) split).dataFiles();
+
+            tableQuery.refreshFiles(partition, bucket, before, after);
+        }
+    }
+
+    @Override
+    public Predicate<InternalRow> recordFilter() {
+        throw new UnsupportedOperationException();
```
JingsongLi, line 89: So many `UnsupportedOperationException`s, you could add an interface `FullCacheLookupTable`?

paimon-core/src/main/java/org/apache/paimon/table/query/TableQueryImpl.java
```diff
@@ -0,0 +1,197 @@
...
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Supplier;
+
+import static org.apache.paimon.CoreOptions.LOOKUP_CACHE_MAX_MEMORY_SIZE;
+
+/** Implementation for {@link TableQuery}. */
+public class TableQueryImpl implements TableQuery {
```
JingsongLi, line 57: Can you add some tests in `PrimaryKeyFileStoreTableTest`?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkConnectorOptions.java
```diff
@@ -280,6 +280,13 @@ public class FlinkConnectorOptions {
                     .defaultValue(16)
                     .withDescription("The thread number for lookup async.");
 
+    @ExcludeFromDocumentation("Not finish yet")
```
JingsongLi, line 283: `Not finish yet`? What is the next?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkConnectorOptions.java
```diff
@@ -280,6 +280,13 @@ public class FlinkConnectorOptions {
...
+    public static final ConfigOption<LookupCacheMode> LOOKUP_CACHE_MODE =
+            ConfigOptions.key("lookup.cache.mode")
```
JingsongLi, line 285: See `https://cwiki.apache.org/confluence/display/FLINK/FLIP-221%3A+Abstraction+for+lookup+source+cache+and+metric`.

lookup.cache => Enum of NONE, PARTIAL and FULL

We are better to be consistent with Flink.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/PrimaryKeyLRULookupTable.java
Aitozi: yes, it looks better

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkConnectorOptions.java
Aitozi: Fixed

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkConnectorOptions.java
Aitozi: Updated

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/PrimaryKeyPartialLookupTable.java
```diff
@@ -22,35 +22,54 @@
...
 import java.util.List;
 
 /** Lookup table for primary key which supports to read the LSM tree directly. */
-public class PrimaryKeyPartialLookupTable implements PartialCacheLookupTable {
+public class PrimaryKeyPartialLookupTable implements LookupTable {
 
-    private static final Logger LOG = LoggerFactory.getLogger(PrimaryKeyPartialLookupTable.class);
-
-    private final FixedBucketKeyExtractor extractor;
+    private final FixedBucketRowKeyExtractor extractor;
```
Aitozi, line 44: `FixedBucketRowKeyExtractor` is project bucket key in the whole record, but in the lookup table, we should project bucket key in the primary key, can this meet the requirement ?

JingsongLi: Oh, we need a test to cover this.

JingsongLi:
Still has one unstable case, need to take a look

Aitozi:
> Still has one unstable case, need to take a look

Yes, I will look into this case

PR #2606, opened 2023-12-30
[flink]Fix the problem of mongo primary key field

MonsterChenzhuo:
### Purpose

Fix the problem of mongo primary key field.


JingsongLi (approved):
+1

PR #2607, opened 2023-12-30
[core]Append tags system table ut

MonsterChenzhuo:
### Purpose

append tags system table ut.

### DIFF
--- a/paimon-core/src/test/java/org/apache/paimon/table/system/TagsTableTest.java
+++ b/paimon-core/src/test/java/org/apache/paimon/table/system/TagsTableTest.java
@@ -1,0 +1,105 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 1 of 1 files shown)


paimon-core/src/test/java/org/apache/paimon/table/system/TagsTableTest.java
```diff
@@ -0,0 +1,105 @@
...
+                new ManifestCommittable(
+                        1,
+                        Timestamp.fromLocalDateTime(LocalDateTime.parse("2023-07-19T12:00:01"))
+                                .getMillisecond()));
+        tagsTable = (TagsTable) catalog.getTable(identifier("T$tags"));
+        tagManager = table.store().newTagManager();
+    }
+
+    @Test
+    public void testSchemasTable() throws Exception {
```
JingsongLi, line 83: name

MonsterChenzhuo: done.

JingsongLi (approved):
+1

PR #2612, opened 2024-01-01
[spark] Decoupling of initialBuckets and assignerParallelism for dynamic bucket write

Zouxxyy:
### Purpose

For write dynamic bucket with spark

- support set `dynamic-bucket.initial-buckets`
- only load indexes belonging to the certain bucket


YannByron:
+1

PR #2614, opened 2024-01-02
[flink] Introduce parallelism execution for lookup join

JingsongLi:
### Purpose

The current lookup join initialization loading is too slow, so we introduce parallelism execution to accelerate it.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/io/SplitsParallelReadUtil.java
+++ b/paimon-core/src/main/java/org/apache/paimon/io/SplitsParallelReadUtil.java
@@ -1,0 +1,132 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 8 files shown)


paimon-core/src/main/java/org/apache/paimon/io/SplitsParallelReadUtil.java
```diff
@@ -0,0 +1,132 @@
...
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+/**
+ * An util class to wrap {@link ParallelExecution} to parallel execution for {@link Split} reader.
```
yuzelin, line 42: `An util` -> `A util` or just `Util`

JingsongLi: Why not an?

JingsongLi: image

yuzelin (commented):
+1

PR #2615, opened 2024-01-02
[core] generate doc and source jar when compile 

zhangjun0x01:
### Purpose

When I debug my code with the official paimon jar in my IDE , I want to download the paimon jar source , but I found it failed, so I think we should build the source jar and upload to maven repository.

We can add an optional parameter to the profile, and we will generate and upload the source jar in every released, user can skip the operation when compile the code in their computer.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -319,6 +319,47 @@ under the License.
                 </plugins>
             </build>
         </profile>
+
+        <profile>
+            <id>docs-and-source</id>
+            <activation>
+                <property>
+                    <name>docs-and-source</name>
+                </property>
+            </activation>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-source-plugin</artifactId>
+                        <version>3.2.1</version>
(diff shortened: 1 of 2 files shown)


JingsongLi:
Thanks @zhangjun0x01 for the contribution.

Have you verified it? It is just a profile...

zhangjun0x01:
> Thanks @zhangjun0x01 for the contribution.
> 
> Have you verified it? It is just a profile...

Yes, I tested it in my computer, when we compile the paimon, we can use command like `mvn clean install -Pdocs-and-source -DskipTests -T8`

when I debug my code with the official paimon jar in my IDE , I want to download the paimon jar source , but I found it failed,  so I think we should build the source jar and upload to maven repository.

 We can add an optional parameter to the profile, and we will generate and upload the  source jar in every released, user can skip the operation when compile the code in their computer.

zhangjun0x01:
> Thanks @zhangjun0x01 for the contribution.
> 
> Have you verified it? It is just a profile...

I add it to the mvn command.

JingsongLi (approved):
+1

PR #2622, opened 2024-01-03
[flink] Fix inconsistency problem when lookup join table with sequence field

### LINKED ISSUE #2621: [Bug] Lookup Join should respect the sequence.field option of primary key table
### Search before asking

### Paimon version

0.6

### Compute Engine

Flink

### Minimal reproduce step

```java
        sql(
                "CREATE TABLE DIM_2 (i INT PRIMARY KEY NOT ENFORCED, j INT, k1 INT, k2 INT) WITH"
                        + " ('continuous.discovery-interval'='1 ms', 'sequence.field' = 'j')");
        sql("INSERT INTO DIM_2 VALUES (1, 11, 111, 1111), (2, 22, 222, 2222)");

        String query =
                "SELECT T.i, D.j, D.k1, D.k2 FROM T LEFT JOIN DIM_2 for system_time as of T.proctime AS D ON T.i = D.i";
        BlockingIterator<Row, Row> iterator = BlockingIterator.of(sEnv.executeSql(query).collect());

        sql("INSERT INTO T VALUES (1), (2), (3)");
        List<Row> result = iterator.collect(3);
        assertThat(result)
                .containsExactlyInAnyOrder(
                        Row.of(1, 11, 111, 1111),
                        Row.of(2, 22, 222, 2222),
                        Row.of(3, null, null, null));

        sql("INSERT INTO DIM_2 VALUES (2, 11, 444, 4444), (3, 33, 333, 3333)");
        Thread.sleep(2000); // wait refresh
        sql("INSERT INTO T VALUES (1), (2), (3), (4)");
        result = iterator.collect(4);
        assertThat(result)
                .containsExactlyInAnyOrder(
                        Row.of(1, 11, 111, 1111),
                        Row.of(2, 22, 222, 2222), // not change
                        Row.of(3, 33, 333, 3333),
                        Row.of(4, null, null, null));

        iterator.close()
```

when we refresh with data of old sequence, the date in rocksdb cache should not be refreshed, otherwise, the join result is unstable.

.

### Anything else?

_No response_

Aitozi:
### Purpose

Linked issue: close #2621

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/types/RowType.java
+++ b/paimon-common/src/main/java/org/apache/paimon/types/RowType.java
@@ -177,6 +177,13 @@ public final class RowType extends DataType {
         }
     }
 
+    public RowType appendDataField(String name, DataType type) {
+        List<DataField> newFields = new ArrayList<>(fields);
+        int newId = currentHighestFieldId(fields) + 1;
+        newFields.add(new DataField(newId, name, type));
+        return new RowType(newFields);
+    }
+
     public static RowType of(DataType... types) {
         final List<DataField> fields = new ArrayList<>();
         for (int i = 0; i < types.length; i++) {
--- a/paimon-common/src/main/java/org/apache/paimon/utils/FunctionWithIOException.java
+++ b/paimon-common/src/main/java/org/apache/paimon/utils/FunctionWithIOException.java
@@ -1,0 +1,24 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.	See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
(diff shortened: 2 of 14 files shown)


Aitozi:
The conflict have be resolved , please help take a look @JingsongLi

JingsongLi:
I have appended a commit [bee1cee](https://github.com/apache/incubator-paimon/pull/2622/commits/bee1ceebef314ea183668d1cfeaf5542affbbc6b), the key point is to make the behavior of the surrounding tool classes not so strange. For example, when passing a rowType to `LookupTable`, when the sequence is enabled, a `BIGINT` field needs to be added, which makes its logic more complex and difficult to understand.

Aitozi:
@JingsongLi Thanks for your improvements, it looks much better, I push a latest commit to fix the serializer mismatch when async lookup join is enabled.

PR #2624, opened 2024-01-03
[hotfix][cdc] Uncheck 'port' of mysql-conf; Fix record null check

yuzelin:
### Purpose

1. port has default value.
2. null value in Json will be parsed into NullNode.


JingsongLi (approved):
+1

PR #2626, opened 2024-01-03
[test] Fix unstable test of parallelism set error.

leaves12138:
### Purpose

Fix unstable test of parallelism check. Parallelism should be greater than 1.


yuzelin (approved):
+1

PR #2629, opened 2024-01-03
[core] Introduce serializeToString and deserializeFromString for internal data structure

Zouxxyy:
### Purpose

Introduce serializeToString and deserializeFromString for internal data structure, support data structures:

BinaryString, Boolean, Byte, Decimal, Double, Float, Int, Long, Shot, Timestamp

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/data/serializer/TimestampSerializer.java
+++ b/paimon-common/src/main/java/org/apache/paimon/data/serializer/TimestampSerializer.java
@@ -21,9 +21,9 @@ package org.apache.paimon.data.serializer;
 import org.apache.paimon.data.Timestamp;
 import org.apache.paimon.io.DataInputView;
 import org.apache.paimon.io.DataOutputView;
+import org.apache.paimon.utils.DateTimeUtils;
 
 import java.io.IOException;
-import java.time.LocalDateTime;
 
 /**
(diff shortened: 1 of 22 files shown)


paimon-common/src/main/java/org/apache/paimon/data/serializer/TimestampSerializer.java
```diff
@@ -73,6 +74,16 @@ public Timestamp deserialize(DataInputView source) throws IOException {
         }
     }
 
+    @Override
+    public String serializeToString(Timestamp record) {
+        return record.toLocalDateTime().toString();
```
JingsongLi, line 79: Use `DateTimeUtils.formatTimestamp`

paimon-common/src/main/java/org/apache/paimon/data/serializer/TimestampSerializer.java
```diff
@@ -73,6 +74,16 @@ public Timestamp deserialize(DataInputView source) throws IOException {
...
+    }
+
+    @Override
+    public Timestamp deserializeFromString(String s) {
+        return Timestamp.fromLocalDateTime(LocalDateTime.parse(s));
```
JingsongLi, line 84: BinaryStringUtils.toTimestamp

JingsongLi (approved):
+1

PR #2632, opened 2024-01-04
[test] Use sql for schema evolution on CatalogTableITCase

XuQianJin-Stars:
### Purpose
Use sql for schema evolution on CatalogTableITCase.

### Tests

`testFilesTable` and `testPartitionsTable`


JingsongLi (approved):
+1

PR #2630, opened 2024-01-03
[CDC] Support cdc sink to UNAWARE bucket mode table

TaoZex:
### Purpose

Linked issue: close #xxx

PR #2636, opened 2024-01-04
[flink] Refactor FlinkSink to clarify the relationships between different bucket-mode tables and sinks

yuzelin:
### Purpose

In addition, #2630 can extend cdc unaware-bucket sink more easy.


leaves12138 (approved):
+1

PR #2637, opened 2024-01-04
[core] If writers size greater than 5, enable buffer and spill for append-only writers in batch mode

leaves12138:
### Purpose

In batch write to append-only table mode, if the partition-bucket writers number is greater than 5, we forcedly to enable write-buffer-for-append and write-buffer-spillable to avoid out of memory error.

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
@@ -327,8 +327,8 @@ public class CoreOptions implements Serializable {
                     .withDescription(
                             "This option only works for append-only table. Whether the write use write buffer to avoid out-of-memory error.");
 
-    public static final ConfigOption<Integer> WRITER_NUMBER_MAX_FORCE_BUFFER =
-            key("writer-number-max-force-buffer")
+    public static final ConfigOption<Integer> WRITE_MAX_WRITERS_TO_SPILL =
+            key("write-max-writers-to-spill")
                     .intType()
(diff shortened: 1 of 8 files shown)


paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -327,6 +327,13 @@ public class CoreOptions implements Serializable {
                     .withDescription(
                             "This option only works for append-only table. Whether the write use write buffer to avoid out-of-memory error.");
 
+    public static final ConfigOption<Integer> WRITER_NUMBER_MAX_FORCE_BUFFER =
+            key("writer-number-max-force-buffer")
```
JingsongLi, line 331: write-max-writers-to-spill

paimon-core/src/main/java/org/apache/paimon/append/AppendOnlyWriter.java
```diff
@@ -210,6 +211,14 @@ public void close() throws Exception {
         sinkWriter.close();
     }
 
+    public void toBufferedWriter() throws Exception {
+        flush(false, false);
```
JingsongLi, line 215: add a `trySyncLatestCompaction(true);`?

JingsongLi, line 215: if it is a buffered sink write, add the spillable is true.

We don't need to new writer.

JingsongLi (approved):
+1

PR #2644, opened 2024-01-05
[core] Optimize lookups on Manifest's FileMeta with binary search

XuQianJin-Stars:
### Purpose

Optimize lookups on Manifest's FileMeta with binary search.

### Tests

`ManifestFileMetaTest`

### API and Format

None.

### Documentation

None.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/manifest/ManifestFileMeta.java
+++ b/paimon-core/src/main/java/org/apache/paimon/manifest/ManifestFileMeta.java
@@ -285,7 +285,7 @@ public class ManifestFileMeta {
                         new FieldStatsArraySerializer(partitionType);
                 int searchResult =
                         binarySearchWithFilter(
-                                base, 0, base.size() - 1, predicate, fieldStatsArraySerializer);
+                                base, j, base.size() - 1, predicate, fieldStatsArraySerializer);
                 if (searchResult != -1) {
                     result.addAll(base.subList(0, searchResult));
                 }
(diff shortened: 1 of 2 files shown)


paimon-core/src/main/java/org/apache/paimon/manifest/ManifestFileMeta.java
```diff
@@ -283,16 +283,11 @@ public static Optional<List<ManifestFileMeta>> tryFullCompaction(
...
-                    if (predicate.test(
-                            file.numAddedFiles + file.numDeletedFiles,
-                            file.partitionStats.fields(fieldStatsArraySerializer))) {
-                        break;
-                    } else {
-                        result.add(file);
-                    }
+                int searchResult =
+                        binarySearchWithFilter(
+                                base, 0, base.size() - 1, predicate, fieldStatsArraySerializer);
```
JingsongLi, line 288: low should not be 0, low should be `j`.

We need to more tests to cover here. The reason for not using binary search at the beginning was that there was not enough energy to write enough tests to cover it.

paimon-core/src/main/java/org/apache/paimon/manifest/ManifestFileMeta.java
```diff
@@ -283,16 +283,11 @@ public static Optional<List<ManifestFileMeta>> tryFullCompaction(
...
-                    if (predicate.test(
-                            file.numAddedFiles + file.numDeletedFiles,
-                            file.partitionStats.fields(fieldStatsArraySerializer))) {
-                        break;
-                    } else {
-                        result.add(file);
-                    }
+                int searchResult =
+                        binarySearchWithFilter(
+                                base, j, base.size() - 1, predicate, fieldStatsArraySerializer);
```
JingsongLi, line 288: I don't remember the specific logic here, but the current modifications are still very dangerous, for example, `j` has not been added here.

JingsongLi (commented):
Thanks @XuQianJin-Stars for your update, but I still have concern here.

JingsongLi:
Thanks @XuQianJin-Stars for the contribution.
It looks like this is not active now.

PR #2646, opened 2024-01-05
[core] snapshot create idle timeout

MonsterChenzhuo:
### Purpose

When a source remains idle for the specified timeout duration in the context of watermarking. It triggers the advancement of snapshots, thereby promoting tag creation.

### Tests

TagAutoCreationTest#testWatermarkIdleTimeoutForceCreatingSnapshot


MonsterChenzhuo:
@yuzelin PTAL,thanks.

MonsterChenzhuo:
@JingsongLi PTAL,thanks.

paimon-core/src/main/java/org/apache/paimon/tag/TagAutoCreation.java
```diff
@@ -96,13 +102,29 @@ private boolean isAutoTag(String tag) {
...
+            return checkIdleTimeout();
+        }
         return timeExtractor instanceof ProcessTimeExtractor
                 && (nextTag == null
                         || isAfterOrEqual(
                                 LocalDateTime.now().minus(delay),
                                 periodHandler.nextTagTime(nextTag)));
     }
 
+    private boolean checkIdleTimeout() {
```
JingsongLi, line 115: inline this method into `forceCreatingSnapshot`, it is hard to understand from this method name `checkIdleTimeout `

paimon-core/src/main/java/org/apache/paimon/tag/TagAutoCreation.java
```diff
@@ -96,13 +102,29 @@ private boolean isAutoTag(String tag) {
     }
 
     public boolean forceCreatingSnapshot() {
+        if (timeExtractor instanceof WatermarkExtractor && idlenessTimeout != null) {
+            return checkIdleTimeout();
+        }
         return timeExtractor instanceof ProcessTimeExtractor
```
JingsongLi, line 108: if (timeExtractor instanceof WatermarkExtractor) {
   ....
}
else if (timeExtractor instanceof ProcessTimeExtractor) {
  ....
}

MonsterChenzhuo:
@JingsongLi  PTAL,thanks.

paimon-core/src/main/java/org/apache/paimon/tag/TagAutoCreation.java

paimon-core/src/main/java/org/apache/paimon/tag/TagAutoCreation.java

MonsterChenzhuo:
@JingsongLi  PTAL,thanks.

PR #2655, opened 2024-01-08
[core][service] Introduce TableQuery and apply to KvQueryServer

JingsongLi:
### Purpose

1. Introduce TableQuery to service and lookup join
2. Apply it to KvQueryServer


Aitozi (approved):
LGTM

PR #2657, opened 2024-01-08
[Spark] Support mergeScalarSubqueries for paimon

YannByron:
Support mergeScalarSubqueries for paimon

Linked issue: close #xxx


YannByron:
https://github.com/apache/incubator-paimon/issues/2404

YannByron:
Based on TpcDS dataset, give a sql like this:
```
select
case when (
	select count(*) from date_dim where d_year between 2000 and 2010) > 4000
then (select avg(d_same_day_ly) from date_dim where d_year between 2000 and 2010)
else (select max(d_same_day_ly) from date_dim where d_year between 2000 and 2010)
end;
```

without this optimization: 

image

with this optimzation:
image

JingsongLi (approved):
+1

Issue #2634, opened 2024-01-04
[Feature] Compatible with upper and lower case when use hivesql queries paimon external table


gnailJC:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/incubator-paimon/issues) and found nothing similar.

### Motivation

1. use sparksql wirte data into a paimon table(has upper case fields) by filesystem-catalog
2. use hivesql to create a paimon external table
3. use hivesql `select upperCaseField(or uppercasefield) from paimon external table` can work

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

zhuangchong:
I can fix it

PR #2659, opened 2024-01-08
[hive] Fix hive to be compatible with uppercase fields when using predicates to query paimon external tables.

zhuangchong:
…

### Purpose

Related issue: #2634


JingsongLi (approved):
+1

PR #2662, opened 2024-01-09
[core] Delete useless schema when expire snapshot

Zouxxyy:
### Purpose

When a schema is not referenced by any snapshot or tag, we can safely delete it

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/AbstractFileStore.java
+++ b/paimon-core/src/main/java/org/apache/paimon/AbstractFileStore.java
@@ -185,6 +185,7 @@ public abstract class AbstractFileStore<T> implements FileStore<T> {
                 options.snapshotNumRetainMin(),
                 options.snapshotNumRetainMax(),
                 options.snapshotTimeRetain().toMillis(),
+                schemaManager,
                 snapshotManager(),
                 newSnapshotDeletion(),
                 newTagManager(),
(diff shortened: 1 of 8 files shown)


zhongyujiang:
IIUC, even if a schema is not referenced by any valid snapshots, there may still be data files associated with it, in this case although the schema won't be used by readers, it can be used for debugging. In addition, these schema files can also be used for auditing since they record the history of  schema changes.
So I'd be cautious about this deleting behaviour, I prefer not to couple the cleaning of schema files with the expiration of snapshots because users may want different retention periods for the two. For example, I might want a longer schema history retention period than snapshot, after all these schema files occupy very little storage.

JingsongLi:
> IIUC, even if a schema is not referenced by any valid snapshots, there may still be data files associated with it, in this case although the schema won't be used by readers, it can be used for debugging. In addition, these schema files can also be used for auditing since they record the history of schema changes. So I'd be cautious about this deleting behaviour, I prefer not to couple the cleaning of schema files with the expiration of snapshots because users may want different retention periods for the two. For example, I might want a longer schema history retention period than snapshot, after all these schema files occupy very little storage.

I agree with you, @zhongyujiang 

The schema file is not too many, and it is very useful for business, and it has no burden. It is just a file, and it is worth keeping forever.
But we can actually provide a CALL command to delete the schema file, but it can only be manually deleted by the user.

JingsongLi:
Close this PR now.

PR #2665, opened 2024-01-09
[core] Don't delete file out of Snapshot expire

JingsongLi:
### Purpose

Remove dead code, here should never come in.


yuzelin (approved):
+1

PR #2668, opened 2024-01-10
[core] Fix index file checkFilesExistence in TableCommitImpl

JingsongLi:
### Purpose

The path of index file is wrong. Here will throw exception. 

This bug is quite serious. For the dynamic bucket mode, long term backpressure is severe, accumulating a lot of checkpoints. Once fails, it is easy to encounter this problem.


yuzelin (approved):
+1

PR #2676, opened 2024-01-11
[core] Optimize CacheManager to reduce disk access

JingsongLi:
### Purpose

Remove `file.length()` for every cache page.


Aitozi (approved):
LGTM

PR #2681, opened 2024-01-11
[flink] Introduce RemoteTableQuery for Flink Lookup join

JingsongLi:
### Purpose

Introduce a new mode Remote Lookup to query service.

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/FileStoreLookupFunction.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/FileStoreLookupFunction.java
@@ -61,6 +61,7 @@ import java.util.stream.IntStream;
 
 import static org.apache.paimon.CoreOptions.CONTINUOUS_DISCOVERY_INTERVAL;
 import static org.apache.paimon.flink.FlinkConnectorOptions.LOOKUP_CACHE_MODE;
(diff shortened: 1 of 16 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/FileStoreLookupFunction.java
```diff
@@ -139,10 +140,17 @@ private void open() throws Exception {
 
         if (options.get(LOOKUP_CACHE_MODE) == LookupCacheMode.AUTO
                 && new HashSet<>(table.primaryKeys()).equals(new HashSet<>(joinKeys))) {
-            try {
+            if (isRemoteServiceAvailable(storeTable)) {
```
Aitozi, line 143: Do you think we could add explicit lookup cache mode eg: `partial_local`, `partial_remote`. It will do good in the test. Otherwise, in IT test, we do not know whether it's  fallback to the `FULL` mode cache

JingsongLi: I would like to keep here simple now, until we have requirement to support specific modes.

I added a test to verify the mode choosing: [d14516f](https://github.com/apache/incubator-paimon/pull/2681/commits/d14516f8c1ddbffa03be096c1cea24381a5fcd71)

Aitozi (approved):
LGTM

PR #2683, opened 2024-01-11
[hotfix] fix NoSuchFiedError:SCHEME

MonsterChenzhuo:
### Purpose

fix NoSuchFiedError:SCHEME


JingsongLi (changes requested):
Have you verify this?

PR #2685, opened 2024-01-12
[hive] Support 'timestamp with local time zone' for Hive

yuzelin:
### Purpose

Linked issue: close #xxx


JingsongLi (commented):
Can we also support Hive2?

yuzelin:
> Can we also support Hive2?

Sure we can. Just convert to timestamp without timezone type

JingsongLi (approved):
+1

PR #2691, opened 2024-01-13
[flink] Add validation for parallelism of the SortOperator.

Aitozi:
### Purpose

Add validation for parallelism of the SortOperator to avoid potential issues with skewed range partitioning.


JingsongLi (approved):
+1

PR #2693, opened 2024-01-14
Revert "[hive] Improve hive database is exists yet create/drop database (#2595)"

Zouxxyy:
This reverts commit b3fd931e0c24202e605489cfcd085314a0b0c470, see https://github.com/apache/incubator-paimon/pull/2595#discussion_r1451752055

### Purpose

Linked issue: close #xxx

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
@@ -119,7 +119,7 @@ public abstract class AbstractCatalog implements Catalog {
                 Collections.singletonList(partitionSpec), BatchWriteBuilder.COMMIT_IDENTIFIER);
     }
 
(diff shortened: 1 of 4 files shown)


JingsongLi:
Hi @Zouxxyy , can you explain more about why we need to revert this commit? test error? performance?

Zouxxyy:
> Hi @Zouxxyy , can you explain more about why we need to revert this commit? test error? performance?

Check alreadly in `createDatabase`, such as whether it is a system db or whether the db exists. `CreateDatabaseImpl` does not require additional verification.

```java
    public void createDatabase(String name, boolean ignoreIfExists)
        if (isSystemDatabase(name)) {
            throw new ProcessSystemDatabaseException();
        }
        if (databaseExists(name)) {
            if (ignoreIfExists) {
                return;
            }
            throw new DatabaseAlreadyExistException(name);
        }

        createDatabaseImpl(name);
    }
```

JingsongLi (approved):
+1

PR #2696, opened 2024-01-15
[core] Introduce bloom filter for local lookup cache file

Aitozi:
### Purpose

Introduce bloom filter for local lookup cache file to speed up the local file lookup


JingsongLi:
Thanks @Aitozi for your contribution.

Can you provide some benchmarks? For example, for high update rates, low update rates, and first-row.

Aitozi:
> Thanks @Aitozi for your contribution.
> 
> Can you provide some benchmarks? For example, for high update rates, low update rates, and first-row.

Yes, I'm still doing some tests, I will share some data here later.

paimon-common/src/main/java/org/apache/paimon/lookup/hash/HashLookupStoreReader.java
```diff
@@ -87,6 +94,15 @@ public class HashLookupStoreReader
             int keyLengthCount = dataInputStream.readInt();
             // Max key length
             int maxKeyLength = dataInputStream.readInt();
+            // BloomFilter
+            bloomFilterEnabled = dataInputStream.readBoolean();
+            if (bloomFilterEnabled) {
+                this.bloomFilter =
```
JingsongLi, line 100: Can this BF into cache?

We have an implementation in Flink: `HashTableBloomFilter`.

BF can base on memory segment too.

Aitozi:
> > Thanks @Aitozi for your contribution.
> > Can you provide some benchmarks? For example, for high update rates, low update rates, and first-row.
> 
> Yes, I'm still doing some tests, I will share some data here later.

Some tests data for low updates of first-row. The input throughput can be doubled compared to the non-bloomfilter version. One of the bottlenecks in the non-bloomfilter is key lookup.

image
image
image

paimon-common/src/main/java/org/apache/paimon/lookup/hash/HashLookupStoreReader.java
Aitozi: I'll take a look

paimon-common/src/main/java/org/apache/paimon/lookup/bloom/BloomFilter.java
```diff
@@ -0,0 +1,148 @@
...
+ * increase accordingly.
+ *
+ * <p>Internally, this implementation of bloom filter uses MemorySegment to store BitSet,
+ * BloomFilter and BitSet are designed to be able to switch between different MemorySegments, so
+ * that Flink can share the same BloomFilter/BitSet object instance for different bloom filters.
+ *
+ * <p>Part of this class refers to the implementation from Apache Hive project
+ * https://github.com/apache/hive/blob/master/common/src/java/org/apache/hive/common/util/BloomFilter.java
+ */
+public class BloomFilter {
```
JingsongLi, line 46: Can you add some ut for these classes?

Aitozi: OK

Aitozi:
Results of highMatch
```
Apple M1 Pro
lookup:                                                                                              Best/Avg Time(ms)    Row Rate(M/s)      Per Row(ns)   Relative
OPERATORTEST_lookup_bf-disabled-0B-value                                                                    16 /   19              6.2            160.8       1.0X
OPERATORTEST_lookup_bf-enabled-0B-value                                                                     20 /   20              5.1            196.7       0.8X
OPERATORTEST_lookup_bf-disabled-500B-value                                                                  27 /   36              3.7            272.4       0.6X
OPERATORTEST_lookup_bf-enabled-500B-value                                                                   22 /   35              4.6            215.6       0.7X
OPERATORTEST_lookup_bf-disabled-1000B-value                                                                 45 /   53              2.2            453.7       0.4X
OPERATORTEST_lookup_bf-enabled-1000B-value                                                                  46 /   53              2.2            464.5       0.3X
OPERATORTEST_lookup_bf-disabled-2000B-value                                                                 68 /   76              1.5            677.6       0.2X
OPERATORTEST_lookup_bf-enabled-2000B-value                                                                  62 /   74              1.6            624.0       0.3X
```

Results of low match
```
OpenJDK 64-Bit Server VM 1.8.0_292-b10 on Mac OS X 10.16
Apple M1 Pro
lookup:                                                                                              Best/Avg Time(ms)    Row Rate(M/s)      Per Row(ns)   Relative
OPERATORTEST_lookup_bf-disabled-0B-value                                                                    41 /   44              2.4            409.6       1.0X
OPERATORTEST_lookup_bf-enabled-0B-value                                                                      6 /    8             16.0             62.7       6.5X
OPERATORTEST_lookup_bf-disabled-500B-value                                                                  37 /   38              2.7            367.5       1.1X
OPERATORTEST_lookup_bf-enabled-500B-value                                                                    5 /    6             18.2             54.9       7.5X
OPERATORTEST_lookup_bf-disabled-1000B-value                                                                 37 /   38              2.7            366.0       1.1X
OPERATORTEST_lookup_bf-enabled-1000B-value                                                                   5 /    5             19.2             52.1       7.9X
OPERATORTEST_lookup_bf-disabled-2000B-value                                                                 37 /   40              2.7            368.7       1.1X
OPERATORTEST_lookup_bf-enabled-2000B-value                                                                   5 /    5             19.0             52.6       7.8X
```

The benefit is clearly seen in the low match result, and for the high match, the result is almost identical in two versions. Do you think it's alright to have this feature enabled by default? @JingsongLi

JingsongLi:
@Aitozi Can you explain what is 0B 500B 1000B?

Aitozi:
> @Aitozi Can you explain what is 0B 500B 1000B?

It's the size of the value. 0B represents the first_row case. The size of the value will affect the cache usage.

JingsongLi:
Hi @Aitozi Thanks for your update!
Left two comments:
1. There is no need to use segments when writing BloomFilter, just use byte[] instead
2. It is best to use standard Segments (default 64KB) when reading BloomFilter to avoid modifying the Cache management system

JingsongLi:
See https://github.com/apache/incubator-paimon/pull/2773

JingsongLi:
Hi @Aitozi , I append this commit https://github.com/apache/incubator-paimon/pull/2696/commits/443a30584b209181653d7e33f21984bb21a60e32 to refactor

Aitozi:
> Hi @Aitozi , I append this commit [443a305](https://github.com/apache/incubator-paimon/commit/443a30584b209181653d7e33f21984bb21a60e32) to refactor

+1, It looks much natural, thanks for your update

PR #2697, opened 2024-01-15
[e2e] fix ContainerLaunchException

MonsterChenzhuo:
### Purpose

图片
The issue is caused by a timeout resulting from the .withLogConsumer("jobmanager_1", new LogConsumer(LOG)) in paimon-e2e-tests/src/test/java/org/apache/paimon/tests/E2eTestBase.java matching logs. This can be resolved by setting a timeout duration. The withStartupTimeout feature is supported from version 1.19.1, so it is necessary to upgrade the testcontainers dependency version.


MonsterChenzhuo:
@yuzelin PTAL,thanks.

yuzelin (approved):
+1

yuzelin:
Good! Lets see whether it can reduce the occurrence of the timeout problem.

PR #2699, opened 2024-01-15
[flink] Add random suffix for lots of data sort to avoid data skew

leaves12138:
### Purpose

Run sort compact action in flink job may cause a data skew with lots of hot data.

This pull request is mainly to resolved this problem, if you config `sort.random-bytes-suffix` = <number>, <number> greater than zero, then we will add a <number>-size long random bytes behind the selected sort columns.

We can avoid hot data all goes to one task by random suffix.


JingsongLi:
Maybe we can have a better range sort to handle data skew problem.

For example, when there are too many simultaneous sampling data, the same data should be evenly distributed to multiple tasks.

leaves12138:
Fixed in https://github.com/apache/incubator-paimon/pull/2705

PR #2705, opened 2024-01-15
[flink] Random choose task while sorting to avoid data skew

leaves12138:
### Purpose

Run sort compact action in flink job may cause a data skew with lots of hot data.

No influence to user.


JingsongLi (approved):
+1

PR #2711, opened 2024-01-16
[core] Introduce compaction.optimization-interval

JingsongLi:
### Purpose

Introduce `compaction.optimization-interval`, it is async option for doing full-compaction. Combined with `read-optimized system table`, it can be better used to users.


yuzelin (approved):
+1

Issue #2708, opened 2024-01-16
[Docs] Add a Chinese Version Documentation for Paimon


Waterkin:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/incubator-paimon/issues) and found nothing similar.

### Motivation

Support Chinese developers with native documentation.

### Solution

1. New branch? (I guess, I am a newbie)
2. Revise docs/config.toml
3. Translate the English docs

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

Waterkin:
I am wondering how to Add a Chinese Version Documentation for Paimon.

PR #2718, opened 2024-01-17
[doc] Add a Chinese version for the documentation

Waterkin:
### Purpose

Linked issue: open #2708 

Add a Chinese version for the documentation, while new Chinese files (stored in `docs/content.zh`) are still the same in English (stored in `docs/content`), needing further translation.


JingsongLi (changes requested):
Hi @Waterkin 

Do not just copy all documentation files. We should do translation one by one.

Waterkin:
Hi @JingsongLi 

I've added an entrance to the new Chinese version at the sidebar, so people can simply focus on the translations of English content at `docs/content.zh`. For example, people can just revise these contents and submit a PR online, making translations more convenient. The convenience will make translation, which is a huge work, easier.

JingsongLi:
> Hi @JingsongLi
> 
> I've added an entrance to the new Chinese version at the sidebar, so people can simply focus on the translations of English content at `docs/content.zh`. For example, people can just revise these contents and submit a PR online, making translations more convenient. The convenience will make translation, which is a huge work, easier.

Do not copy without translation which makes maintenance very difficult, and the result is not yet any benefit.

Waterkin:
@JingsongLi Do you think it ok that I delete the copyed files and only translate some of them?

JingsongLi:
> @JingsongLi Do you think it ok that I delete the copyed files and only translate some of them?

Yes, that is a better way.

Waterkin:
@JingsongLi PTAL, thanks!

JingsongLi:
Hi @Waterkin , very thanks for your contribution.

But it is really hard to maintain two documentation. It is recommended to use translator to read english documentation.

PR #2719, opened 2024-01-17
[core] Rename first_not_null_value agg function to first_non_null_value

tsreaper:
### Purpose

Currently we have two aggregate functions for the `aggregation` merge engine: `last_non_null_value` and `first_not_null_value`. These two names should be unified to prevent users from typo.

This PR renames `first_not_null_value` to `first_non_null_value`. The legacy name is also kept for compatibility.

### Tests

* `PreAggregationITCase#FirstValueAggregation`.

### API and Format

No.

### Documentation

Yes. Document is also changed.


JingsongLi (approved):
+1

PR #2722, opened 2024-01-17
[core] Ensure enough tasks for scan.manifest.parallelism

JingsongLi:

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/utils/FileUtils.java
+++ b/paimon-core/src/main/java/org/apache/paimon/utils/FileUtils.java
@@ -39,18 +39,32 @@ public class FileUtils {
 
     public static final ForkJoinPool COMMON_IO_FORK_JOIN_POOL;
 
-    // if we want to name threads in the fork join pool we need all these
-    // see https://stackoverflow.com/questions/34303094/
+    private static volatile ForkJoinPool scanIoForkJoinPool;
+
     static {
+        COMMON_IO_FORK_JOIN_POOL =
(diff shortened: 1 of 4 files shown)


paimon-core/src/main/java/org/apache/paimon/utils/FileUtils.java
```diff
@@ -39,18 +39,32 @@ public class FileUtils {
...
                     return worker;
                 };
-        COMMON_IO_FORK_JOIN_POOL =
-                new ForkJoinPool(Runtime.getRuntime().availableProcessors(), factory, null, false);
+        return new ForkJoinPool(parallelism, factory, null, false);
+    }
+
+    public static synchronized ForkJoinPool getScanIoForkJoinPool(int parallelism) {
+        if (scanIoForkJoinPool == null || parallelism > scanIoForkJoinPool.getParallelism()) {
+            scanIoForkJoinPool = createForkJoinPool("paimon-scan-io", parallelism);
```
leaves12138, line 65: If scanIoForkJoinPool is not null, and parallelism > scanIoForkJoinPool.getParallelism,
redirect scanIoForkJoinPool may cause thread leak?

leaves12138: Should we wait scanIoForJoinPool finished its tasks, then call its shudown method, then redirect scanIoForkJoinPoll to a new ForkJoinPool?

JingsongLi:
Thanks @leaves12138 for your commit!

PR #2566, opened 2023-12-22
[core] Support tag date formatter

MonsterChenzhuo:
### Purpose

Support tag date yyyyMMdd formatter.

### Tests

TagAutoCreationTest#testTagDatePeriodFormatter
TagAutoCreationTest#testTagHourlyPeriodFormatter

PR #2724, opened 2024-01-17
[core] Introduce 'scan.version' to refactor spark time travel

yuzelin:
### Purpose

After #2566, a tag name can be a number, so spark time travel to a number version string may be confused (the version string is a tag or a snapshot ?)
This PR fix it.


JingsongLi (changes requested):
We may need to introduce a new option `scan.version`.

JingsongLi (approved):
+1

PR #2730, opened 2024-01-17
[flink] Introduce dbz cdc dataType

MonsterChenzhuo:
### Purpose

Introduce dbz cdc dataType.


yuzelin (approved):
+1

TaoZex:
Add doc for users may be better. @MonsterChenzhuo

yuzelin:
> Add doc for users may be better. @MonsterChenzhuo

You're right. I'm considering to do some refactor. I will add docs then.

TaoZex:
> > Add doc for users may be better. @MonsterChenzhuo
> 
> You're right. I'm considering to do some refactor. I will add docs then.

Look forward to this work.

PR #2736, opened 2024-01-18
[core] Fix bug: zindexer should deal well with column `null` value.

leaves12138:
### Purpose

If a column is null, ZIndexer will throws an exception. It shouldn't get the null row:

```sql
                -- it is not correct to get the string before check if it is null.
                BinaryString binaryString = row.getString(fieldIndex);

                return row.isNullAt(fieldIndex)
                        ? NULL_BYTES
                        : ZOrderByteUtils.byteTruncateOrFill(
                                        MemorySegmentUtils.getBytes(
                                                binaryString.getSegments(),
                                                binaryString.getOffset(),
                                                Math.min(
                                                        varTypeSize,
                                                        binaryString.getSizeInBytes())),
                                        varTypeSize,
                                        reuse)
                                .array();
```


yuzelin (approved):
+1

PR #2739, opened 2024-01-19
[format]Fix orc and parquet writer about timestamp not contains [local timezone] and [is_adjust_to_utc]

leoyy0316:
### Purpose
Now, paimon timestamp type lost some metadata
orc file Type meta not contains timestamp with local zone, just timestamp type
parquet file file schema meta not contains adjust to UTC   

Linked issue: close #xxx


JingsongLi (approved):
+1

PR #2744, opened 2024-01-19
[doc] Use underscore instead of hyphen for rollback_to in doc.

liming30:
### Purpose

[doc] Use underscore instead of hyphen for rollback_to in doc.


JingsongLi (approved):
+1

PR #2746, opened 2024-01-20
[hotfix] The new column exception shows the specific column name in the cdc database synchronization action.

zhuangchong:
…

### Purpose

The new column exception shows the specific column name in the cdc database synchronization action.


JingsongLi (approved):
+1

PR #2750, opened 2024-01-20
[WIP][core] Introduce stats system table

MonsterChenzhuo:
### Purpose

 Introduce stats system table.


Zouxxyy:
we don't need an exta stats system table, just add a new `statistis` field (json) to snapshot system table

PR #2757, opened 2024-01-22
[core]: fix flaky test `testSavepointTag`.

zhongyujiang:
### Purpose
`testSavepointTag` is flaky,  this test will fail on CentOS. 
The reason is that when using  `tagManager.tags()`in L243 to get tags, the result order of `listVersionedFileStatus` is undetermined, depending on different operating systems.  
When running this test on CentOS, the result of `listVersionedFileStatus` is `tag-2023-07-18 11  tag-2023-07-18 13  tag-savepoint-11`. Because `tag-2023-07-18 11`  and `tag-savepoint-11` belongs to the same snapshot, the later `tag-savepoint-11` tag will overwride the previous `tag-2023-07-18 11` tag. The result of `tagManager.tags()` will be {<s1, tag-savepoint-11>, <s2, tag-2023-07-18 13>} and thus won't pass the validation.

This fix filters out the savepoint tag when getting tags to fix the issue.

### DIFF
--- a/paimon-core/src/test/java/org/apache/paimon/tag/TagAutoCreationTest.java
+++ b/paimon-core/src/test/java/org/apache/paimon/tag/TagAutoCreationTest.java
@@ -33,6 +33,7 @@ import org.junit.jupiter.api.Test;
 import java.time.Duration;
 import java.time.LocalDateTime;
 import java.time.ZoneId;
+import java.util.Objects;
 
 import static org.apache.paimon.CoreOptions.SINK_WATERMARK_TIME_ZONE;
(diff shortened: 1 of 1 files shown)


yuzelin:
You're right. And the `tags()` method are not reasonable because more than 1 tags can be created based on the same snapshot. I think we should fix it.

zhongyujiang:
> And the tags() method is not reasonable because more than 1 tags can be created based on the same snapshot. I think we should fix it.

Yeah, I think it can be changed to return a map of `tag name` to `snapshot` since `tag name` is unique, and I feel using tag name as key is more reasonable since users usually use `TagManager` to retrieve tags instead of snapshots. What do you think?

PR #2762, opened 2024-01-22
[flink] Fix the problem that flink job for reading paimon won't recov…

### LINKED ISSUE #2761: [Bug] Flink can't recover from exception as 'OutOfRangeException' when sometime the paimon snapshot is expired
### Search before asking

### Paimon version

0.6

### Compute Engine

flink

### Minimal reproduce step

1. Flink job for reading paimon is running for a while and left some checkpoints
2. Flink job try to failover for some exception，and the snapshot of paimon to read is expired.
3. Flink job can't reading latest data on paimon，bu only log "Failed to enumerate files，org.apache.paimon.table.source OutOfRangeException.... ", won't try to cancel job an restart。

like this:
image

When catch `OutOfRangeException`，flink job should try to failover。

### Anything else?

_No response_

bridgeDream:
…er when catching snapshot OutOfRangeException

### Purpose

Linked issue: close #2761

Throw a runtime exception in flink job to try to failover, when reading paimon found a OutOfRangeException

### Tests

None

### API and Format

None
### Documentation

None

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/ContinuousFileSplitEnumerator.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/ContinuousFileSplitEnumerator.java
@@ -25,6 +25,7 @@ import org.apache.paimon.flink.source.assigners.SplitAssigner;
 import org.apache.paimon.table.BucketMode;
 import org.apache.paimon.table.source.DataSplit;
 import org.apache.paimon.table.source.EndOfScanException;
+import org.apache.paimon.table.source.OutOfRangeException;
 import org.apache.paimon.table.source.SnapshotNotExistPlan;
 import org.apache.paimon.table.source.StreamTableScan;
 import org.apache.paimon.table.source.TableScan;
@@ -210,6 +211,8 @@ public class ContinuousFileSplitEnumerator
                 LOG.debug("Catching EndOfStreamException, the stream is finished.");
                 finished = true;
                 assignSplits();
+            } else if (error instanceof OutOfRangeException) {
+                throw (RuntimeException) error;
             } else {
                 LOG.error("Failed to enumerate files", error);
             }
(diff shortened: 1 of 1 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/ContinuousFileSplitEnumerator.java
```diff
@@ -210,6 +211,8 @@ protected void processDiscoveredSplits(
                 LOG.debug("Catching EndOfStreamException, the stream is finished.");
                 finished = true;
                 assignSplits();
+            } else if (error instanceof OutOfRangeException) {
+                throw (RuntimeException) error;
             } else {
                 LOG.error("Failed to enumerate files", error);
```
JingsongLi, line 217: Maybe we need to throw all exceptions here?

bridgeDream: I agree with that. Thank you

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/ContinuousFileSplitEnumerator.java
```diff
@@ -212,6 +212,7 @@ protected void processDiscoveredSplits(
                 assignSplits();
             } else {
                 LOG.error("Failed to enumerate files", error);
+                throw (RuntimeException) error;
```
JingsongLi, line 215: throw new RuntimeException(error)?

bridgeDream: > throw new RuntimeException(error)?

Done, FYI !

JingsongLi:
+1

PR #2765, opened 2024-01-22
[core] Fix that TagManager doesn't handle case where more than 1 tags are created based on the same snapshot

yuzelin:
### Purpose

Also fix #2757

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/utils/TagManager.java
+++ b/paimon-core/src/main/java/org/apache/paimon/utils/TagManager.java
@@ -19,6 +19,7 @@
 package org.apache.paimon.utils;
 
 import org.apache.paimon.Snapshot;
+import org.apache.paimon.annotation.VisibleForTesting;
 import org.apache.paimon.fs.FileIO;
 import org.apache.paimon.fs.FileStatus;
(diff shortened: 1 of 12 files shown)


paimon-core/src/main/java/org/apache/paimon/utils/TagManager.java
```diff
@@ -221,14 +257,40 @@ public SortedMap<Snapshot, String> tags(Predicate<String> filter) {
...
+                            try {
+                                return fileIO.getFileStatus(p);
+                            } catch (IOException e) {
+                                throw new RuntimeException(e);
+                            }
+                        })
+                .sorted(Comparator.comparingLong(FileStatus::getModificationTime))
+                .map(FileStatus::getPath)
+                .map(Path::getName)
+                .map(name -> name.substring(TAG_PREFIX.length()))
```
TaoZex, line 285: How about use `.map(fileStatus -> fileStatus.getPath().getName().substring(TAG_PREFIX.length()))`

yuzelin: Sure.

JingsongLi (approved):
+1

PR #2773, opened 2024-01-23
[core] Let CacheManager page unaware

JingsongLi:
### Purpose

- The Cache system should be page independent and can use a custom pageSize.
- Introduce a new Page size for cache, can be smaller.

### API and Format

cache-page-size default 16 KB


Aitozi (approved):
+1

PR #2779, opened 2024-01-24
[core] Make AbstractFileStoreTable package private

tsreaper:
### Purpose

This PR makes `AbstractFileStoreTable` a package private class. Developers should use `FileStoreTable` interface instead.

### Tests

This PR is a refactor without tests.

### API and Format

No.

### Documentation

No.


JingsongLi (approved):
+1

PR #2781, opened 2024-01-24
[core] Introduce UpgradeStrategy to simplify logical

JingsongLi:
### Purpose

Continue https://github.com/apache/incubator-paimon/pull/2751

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/FirstRowMergeTreeCompactRewriter.java
+++ b/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/FirstRowMergeTreeCompactRewriter.java
@@ -36,9 +36,9 @@ import java.io.UncheckedIOException;
 import java.util.Comparator;
 import java.util.List;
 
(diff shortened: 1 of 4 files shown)


paimon-core/src/main/java/org/apache/paimon/mergetree/compact/FirstRowMergeTreeCompactRewriter.java
```diff
@@ -77,8 +80,19 @@ protected boolean rewriteChangelog(
     }
 
     @Override
-    protected boolean upgradeChangelog(int outputLevel, DataFileMeta file) {
-        return file.level() == 0;
+    protected UpgradeStrategy upgradeChangelog(int outputLevel, DataFileMeta file) {
```
Zouxxyy, line 83: maybe this overwrite is no need

JingsongLi: this class is not extended lookup rewriter.

Zouxxyy:
+1

PR #2789, opened 2024-01-25
[core] Introduce privilege system for catalog based on FileSystem

tsreaper:
### Purpose

This PR introduces an identity-based privilege system for catalogs. Catalogs can now be updated into a privileged catalogs, where privileged users can be created and granted privileges to tables, databases or the whole catalog.

### Tests

* `PrivilegeManagerTest`
* `PrivilegeProcedureITCase`

### API and Format

Yes. This PR introduces two new system tables: user and privilege.

User table is the table which stores all user information. The schema of user table is:
* user (string): user name (primary key)
* sha256 (bytes): sha256 of password

Privilege table is the table storing what privileges each user have. Its schema is:
* name (string): user or role name (primary key)
* entity_type (string): user or role (primary key)
* identifier (string): identifier of object (primary key)
* privilege (string): name of privilege (primary key), see `PrivilegeType`

### Documentation

Yes. Document is also added.

### DIFF
--- a/docs/content/maintenance/manage-privileges.md
+++ b/docs/content/maintenance/manage-privileges.md
@@ -94,9 +94,9 @@ This is the default user if no username and password is provided when creating t
 ## Enable Privileges
 
 Paimon currently only supports file-based privilege system.
-Only catalogs with `'metastore' = 'filesystem'` (the default value) support such privilege system.
+Only catalogs with `'metastore' = 'filesystem'` (the default value) or `'metastore' = 'hive'` support such privilege system.
 
-To enable the privilege system on a filesystem catalog, do the following steps.
+To enable the privilege system on a filesystem / Hive catalog, do the following steps.
 
 {{< tabs "enable-privileges" >}}
 
--- a/docs/content/maintenance/manage-privileges.md
+++ b/docs/content/maintenance/manage-privileges.md
@@ -1,0 +1,246 @@
+---
+title: "Manage Privileges"
+weight: 10
+type: docs
+aliases:
+- /maintenance/manage-privileges.html
+---
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+# Manage Privileges
+
+Paimon provides a privilege system on catalogs.
+Privileges determine which users can perform which operations on which objects,
+so that you can manage table access in a fine-grained manner.
+
+Currently, Paimon adopts the identity-based access control (IBAC) privilege model.
+That is, privileges are directly assigned to users.
+
+{{< hint warning >}}
+This privilege system only prevents unwanted users from accessing tables through catalogs.
+It does not block access through temporary table (by specifying table path on filesystem),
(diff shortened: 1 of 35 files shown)


JingsongLi:
Please finish documentation in this PR.

paimon-core/src/main/java/org/apache/paimon/privilege/PrivilegedFileStore.java
```diff
@@ -0,0 +1,189 @@
...
+import org.apache.paimon.utils.FileStorePathFactory;
+import org.apache.paimon.utils.SnapshotManager;
+import org.apache.paimon.utils.TagManager;
+
+import org.jetbrains.annotations.Nullable;
+
+import java.util.List;
+
+/** {@link FileStore} with privilege checks. */
+public class PrivilegedFileStore<T> implements FileStore<T> {
```
JingsongLi, line 50: We should not wrap FileStore, all things should be finished in Table interface.

paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
```diff
@@ -54,32 +61,43 @@
 import static org.apache.paimon.utils.Preconditions.checkArgument;
 
 /** Common implementation of {@link Catalog}. */
-public abstract class AbstractCatalog implements Catalog {
+public abstract class AbstractCatalog implements Catalog, PrivilegedCatalog {
```
JingsongLi, line 64: Maybe we should wrap a catalog just like `PrevilegedFileStoreTable`.

JingsongLi (commented):
Can you describe storage format for privilege system in the PR description?

paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
JingsongLi: Some catalogs may not use FileSystem based Previlege system.
For example, RestCatalog will register users in server not filesystem.

zhangjun0x01:
hi, @tsreaper , will there be plan to integrate  with apache ranger  in the future? like this: https://doris.apache.org/docs/dev/admin-manual/privilege-ldap/ranger/

docs/content/maintenance/manage-privileges.md
```diff
@@ -0,0 +1,246 @@
...
+* The `root` user,
+which is identified by the provided root password when enabling the privilege system.
+This user always has all privileges in the catalog.
+* The `anonymous` user.
+This is the default user if no username and password is provided when creating the catalog.
+
+## Enable Privileges
+
+Paimon currently only supports file-based privilege system.
+Only catalogs with `'metastore' = 'filesystem'` (the default value) support such privilege system.
```
JingsongLi, line 97: Can we support filesystem privilege for hive catalog too?

paimon-core/src/main/java/org/apache/paimon/privilege/FileBasedPrivilegeManager.java
```diff
@@ -0,0 +1,469 @@
...
+                    "Privilege system is not enabled in warehouse " + warehouse + ".");
+        } else {
+            return null;
+        }
+    }
+
+    private void createUserTable() {
+        Options options = new Options();
+        options.set(CoreOptions.BUCKET, 1);
+        Path tableRoot = new Path(warehouse, USER_TABLE_DIR);
```
JingsongLi, line 399: Can we just use Catalog API here?

paimon-core/src/main/java/org/apache/paimon/privilege/FileBasedPrivilegeManager.java
```diff
@@ -0,0 +1,469 @@
...
+ * <ul>
+ *   <li>name (string): user or role name (primary key)
+ *   <li>entity_type (string): user or role (primary key)
+ *   <li>identifier (string): identifier of object (primary key)
+ *   <li>privilege (string): name of privilege (primary key), see {@link PrivilegeType}
+ * </ul>
+ */
+public class FileBasedPrivilegeManager implements PrivilegeManager {
+
+    private static final String USER_TABLE_DIR = "user.sys";
```
JingsongLi, line 84: Maybe db is privilege, and table is user?

JingsongLi (approved):
+1

PR #2791, opened 2024-01-25
[core] Make PrimaryKeyFileStoreTable and AppendOnlyFileStoreTable package private

tsreaper:
### Purpose

This PR makes `PrimaryKeyFileStoreTable` and `AppendOnlyFileStoreTable` package private classes. Developers should use `FileStoreTable` interface instead and check if the table has primary keys.

### Tests

This PR is a refactor without tests.

### API and Format

No.

### Documentation

No.


JingsongLi:
+1

PR #2793, opened 2024-01-25
[cdc] Added database action factory base class.

zhuangchong:
### Purpose

Added database action factory base class.


paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/action/cdc/SyncDatabaseActionFactoryBase.java
```diff
@@ -0,0 +1,59 @@
...
+import static org.apache.paimon.flink.action.cdc.CdcActionCommonUtils.TABLE_PREFIX;
+import static org.apache.paimon.flink.action.cdc.CdcActionCommonUtils.TABLE_SUFFIX;
+import static org.apache.paimon.flink.action.cdc.CdcActionCommonUtils.TYPE_MAPPING;
+
+/** Base {@link ActionFactory} for synchronizing into database. */
+public abstract class SyncDatabaseActionFactoryBase<T extends SyncDatabaseActionBase>
+        extends SynchronizationActionFactoryBase<T> {
+
+    protected String warehouse;
+    protected String database;
```
TaoZex, line 38: How about use `final`?

zhuangchong: The subclass of ActionFactory is an instance generated through dynamic discovery through spi. This ActionFactory can only have a parameterless constructor, so the variable here does not use `final`.

TaoZex: Thanks for your responce. Overall look good to me.

yuzelin (approved):
+1

zhuangchong:
Thanks. @yuzelin

PR #2796, opened 2024-01-26
[core] Remove Predicate.test(Object[]) to use InternalRow

JingsongLi:
### Purpose

We should use BinaryRow instead of Object[] to reduce memory.


leaves12138 (approved):
+1

PR #2800, opened 2024-01-27
[core] Optimize BinaryTableStats to real binary

JingsongLi:
### Purpose

BinaryTableStats contains many objects cache, this will make JVM heap memory fragile.

In this PR, make BinaryTableStats to real binary, can support tables on the scale of 100TB.


yuzelin (approved):
+1

PR #2801, opened 2024-01-27
[CDC]Improve pg sync table cdc

TaoZex:
### Purpose

Port is not a required parameter in sync pg table.


TaoZex:
PTAL @zhuangchong

zhuangchong (approved):
+1

PR #2808, opened 2024-01-29
[core] Support filter pushdown in FilesTable

JingsongLi:
### Purpose

For large tables, querying the files table is very slow, and we need to support some filter down acceleration queries.

This PR pushdown for three fields:
1. partition
2. bucket
3. level

These three fields related conditions are the most important query conditions.


XuQianJin-Stars:
Overall it looks like a good implementation.

yuzelin (approved):
+1

PR #2810, opened 2024-01-29
[hive] alter paimon table options cannot be synchronized to hive par

herefree:
### Purpose

Fix alter paimon table options cannot be synchronized to hive parameters.

### Tests
org.apache.paimon.hive.HiveCatalogTest.testAlterHiveTableParameters()


JingsongLi:
Looks reasonable to me.

PR #2811, opened 2024-01-29
[core] Fix the issue in `sequence.auto-padding` where `second-to-micro` and `millis-to-micro` are not meeting the requirements.

### LINKED ISSUE #2041: [Bug] The configurations of `second-to-micro` and `millis-to-micro` within `sequence.auto-padding` does not meet the expected behavior.
### Search before asking

### Paimon version

<=master

### Compute Engine

flink <=master

### Minimal reproduce step

The configurations of `second-to-micro` and `millis-to-micro` within `sequence.auto-padding` can lead to issues with nanotime generation (similar to mathematical carry problems). This can result in a combined sequence in later incoming data being smaller than that in historical data when they share the same `sequence.field`. Consequently, this might lead to the data being discarded, even though, theoretically, incoming data should always be the latest, provided they fall under the same `sequence.field`.

See https://github.com/apache/incubator-paimon/pull/1247#issuecomment-1670885102.

### Anything else?

_No response_

hzjhjjyy:
### Purpose

Linked issue: close #2041

By adding `ReorderFunction`(impl: `SequenceFieldReorderFunction`), it performs reordering based on `sequence.field` and the internal incrementing sequence number before data merging. This truly achieves the functionality users desire with custom `sequence.field`.

PS: Perhaps, in the end, achieving optimal performance could involve modifying the underlying storage format, as outlined in https://github.com/facebook/rocksdb/wiki/User-defined-Timestamp.
### Tests

`SequenceFieldReorderFunctionTest`
`SortBufferWriteBufferTestBase`
`PrimaryKeyFileStoreTableTest`

### DIFF
--- a/docs/content/concepts/primary-key-table/sequence-rowkind.md
+++ b/docs/content/concepts/primary-key-table/sequence-rowkind.md
@@ -61,12 +61,7 @@ option to automatically pad the sequence field for you.
    (the time that the change was made in the database) in Mysql Binlog. It is recommended to use the automatic
    padding for row kind flag, which will automatically distinguish between -U (-D) and +U (+I).
 
-2. Insufficient precision: If the provided `sequence.field` doesn't meet the precision, like a rough second or
-   millisecond, you can set `sequence.auto-padding` to `second-to-micro` or `millis-to-micro` so that the precision
-   of sequence number will be made up to microsecond by system.
-
-3. Composite pattern: for example, "second-to-micro,row-kind-flag", first, add the micro to the second, and then
-   pad the row kind flag.
+(Note: If the provided `sequence.field` doesn't meet the precision, like a rough second or millisecond, system will combine `sequence.field` with an internal incrementing sequence number to ensure that even if the same `sequence.field` appears, the built-in incrementing sequence number can distinguish between different records.)
 
 ## Row Kind Field
 
--- a/docs/layouts/shortcodes/generated/core_configuration.html
+++ b/docs/layouts/shortcodes/generated/core_configuration.html
@@ -531,7 +531,7 @@ This config option does not affect the default filesystem metastore.</td>
             <td><h5>sequence.auto-padding</h5></td>
             <td style="word-wrap: break-word;">(none)</td>
             <td>String</td>
-            <td>Specify the way of padding precision, if the provided sequence field is used to indicate "time" but doesn't meet the precise.<ul><li>You can specific:</li><li>1. "row-kind-flag": Pads a bit flag to indicate whether it is retract (0) or add (1) message.</li><li>2. "second-to-micro": Pads the sequence field that indicates time with precision of seconds to micro-second.</li><li>3. "millis-to-micro": Pads the sequence field that indicates time with precision of milli-second to micro-second.</li><li>4. Composite pattern: for example, "second-to-micro,row-kind-flag".</li></ul></td>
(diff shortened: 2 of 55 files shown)


hzjhjjyy:
Could you spare some time to help with a review?  @JingsongLi

JingsongLi:
Hi @hzjhjjyy , you raised a good point.
Should we separate user define timestamp and our incremental sequence?

I don't have an answer now. Perhaps we need significant adjustments in our code.

https://github.com/apache/incubator-paimon/pull/2895 can have some relief.

JingsongLi:
I think that this solution is the ultimate solution, but the current changes are quite significant and uncontrollable. Is it possible to break it down into parts of the code and push it forward?

hzjhjjyy:
Thank you for your review @JingsongLi. Do you think it's better to go for an approach similar to the implementation of RocksDB, or should we keep the current approach in this PR? I've tried the former, but I found that there were many hardcoded data structure operations at the lower level, which made modifications cumbersome and might require extensive testing to ensure compatibility. Therefore, I switched to the approach in the current PR, which adds a reordering of (use seq, auto seq) pairs before data merge without changing the underlying data structure. Of course, the latter approach may sacrifice some performance as a trade-off.

If we choose the latter approach, I can break down the current PR into different commits based on modules and then resubmit a new PR. This would make the review process easier. On the other hand, if we opt for the former approach, it might require more detailed refactoring work and take longer to implement. As of now, the current PR seems to be running smoothly in my production environment.

PS: The PR has been running smoothly in my production environment and achieving the desired results so far.

JingsongLi:
> Thank you for your review @JingsongLi. Do you think it's better to go for an approach similar to the implementation of RocksDB, or should we keep the current approach in this PR? I've tried the former, but I found that there were many hardcoded data structure operations at the lower level, which made modifications cumbersome and might require extensive testing to ensure compatibility. Therefore, I switched to the approach in the current PR, which adds a reordering of (use seq, auto seq) pairs before data merge without changing the underlying data structure. Of course, the latter approach may sacrifice some performance as a trade-off.
> 
> If we choose the latter approach, I can break down the current PR into different commits based on modules and then resubmit a new PR. This would make the review process easier. On the other hand, if we opt for the former approach, it might require more detailed refactoring work and take longer to implement.
> 
> PS: The PR has been running good in my production environment and achieving the desired results so far.

Thank you for your attempt and research. I think we can follow the concept of RocksDB and move forward, but we don't need to modify the underlying data structure.

When comparing, we can first compare the user-defined sequenceField (We don't need to extract long now, just compare) and then compare our self increasing sequenceId, which does not seem to cause serious compatibility issues and does not require modifying the storage structure.

What do you think?

If you feel the need to define behavior clearly, you can also consider creating a PIP to describe it. This PIP does not require voting, but only records our design.

[1] https://cwiki.apache.org/confluence/display/PAIMON/Paimon+Improvement+Proposals

JingsongLi:
I am working on the first step: https://github.com/apache/incubator-paimon/pull/2936

JingsongLi:
> > I am working on the first step: #2936
> 
> That's fantastic! I'm also working on modifying parts similar to what you've changed, after all, that's the core of the code. My current pr doesn't touch the deepest layers yet, mainly to avoid making too many changes that could be difficult to implement. Looking forward to seeing the final implementation from the author. @JingsongLi

You can take a review on https://github.com/apache/incubator-paimon/pull/2936 , and then you can create pull request to modify more parts if you want to do this.

PR #2812, opened 2024-01-29
[core] Introduce more compressions for spill

JingsongLi:
### Purpose

Sometimes, local disk does not have enough space, we need to introduce more compression algorithms to reduce disk usage.


yuzelin (approved):
+1

PR #2814, opened 2024-01-29
[common] Fix the method for obtaining the authority of a URI on Windows.

hzjhjjyy:
### Purpose

Currently, when constructing a path on Windows using something like "file://C:\temp", the value of the authority is "C:\temp" instead of an empty string "".


paimon-common/src/main/java/org/apache/paimon/fs/Path.java
```diff
@@ -173,6 +173,11 @@ public Path(String pathString) {
         // parse uri authority, if any
         if (pathString.startsWith("//", start)
```
JingsongLi, line 174: This is copied from Flink and Hadoop, can you check work in your machine?

hzjhjjyy: For example, when testing in IncrementalStartingScannerTest#testScan.
before fix:
屏幕截图 2024-02-02 170630
after fix:
屏幕截图 2024-02-02 170906

PR #2816, opened 2024-01-30
[core] Agg 'collect' and 'merge_map' support retract

yuzelin:
### Purpose

The retract isn't correct when:
1. If the retract field arrives but accumulator doesn't, won't retract;
2. The retract field from one input stream will retract the result merged from multiple input streams.
for example, using merge_map, acc (1, A) + input (1, B) will get (1, B), and if retract (1, A) later, the result (1, B) will be deleted from result map.

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkTableSinkBase.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkTableSinkBase.java
@@ -83,7 +83,8 @@ public abstract class FlinkTableSinkBase
                 return requestedMode;
             }
 
-            if (options.get(MERGE_ENGINE) == MergeEngine.AGGREGATE) {
+            MergeEngine mergeEngine = options.get(MERGE_ENGINE);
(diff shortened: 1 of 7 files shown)


JingsongLi (approved):
+1 Please update documentation too.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkTableSinkBase.java
```diff
@@ -83,7 +83,8 @@ public ChangelogMode getChangelogMode(ChangelogMode requestedMode) {
                 return requestedMode;
             }
 
-            if (options.get(MERGE_ENGINE) == MergeEngine.AGGREGATE) {
+            MergeEngine mergeEngine = options.get(MERGE_ENGINE);
+            if (mergeEngine == MergeEngine.AGGREGATE || mergeEngine == MergeEngine.PARTIAL_UPDATE) {
```
JingsongLi, line 87: check if agg func is defined

JingsongLi:
+1

PR #2817, opened 2024-01-30
[core] Introduce Context for LookupStoreFactory to avoid writing metadata

JingsongLi:
### Purpose

The previous approach made the file structure too complex, which is not conducive to adding file compression in the future.

We can interact solely through memory.


yuzelin (approved):
+1

PR #2819, opened 2024-01-30
[Bug] Fix the protobuf conflict in paimon-micro-benchmarks

Aitozi:
### Purpose

Fix the exception when running test in IDE.

image


JingsongLi (approved):
+1

PR #2820, opened 2024-01-30
[core] Integrate compression to Hash Lookup

JingsongLi:
### Purpose

At present, Hash Lookup does not have any compression, which may result in high local disk usage and high disk IO. This PR introduces compression and is configurable.

### API and Format

1. Page 16KB -> 64KB, for compression.
2. Default compression: lz4.

### DIFF
--- a/docs/layouts/shortcodes/generated/core_configuration.html
+++ b/docs/layouts/shortcodes/generated/core_configuration.html
@@ -46,7 +46,7 @@ under the License.
         </tr>
         <tr>
             <td><h5>cache-page-size</h5></td>
-            <td style="word-wrap: break-word;">16 kb</td>
+            <td style="word-wrap: break-word;">64 kb</td>
             <td>MemorySize</td>
             <td>Memory page size for caching.</td>
         </tr>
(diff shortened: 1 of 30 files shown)


paimon-common/src/main/java/org/apache/paimon/io/UncompressedPageFileInput.java
```diff
@@ -0,0 +1,79 @@
...
+        int length = (int) Math.min(pageSize, fileLength - position);
+        byte[] result = new byte[length];
+        file.readFully(result);
+        return result;
+    }
+
+    @Override
+    public byte[] readPosition(long position, int length) throws IOException {
+        file.seek(position);
+        byte[] result = new byte[length];
```
Aitozi, line 70: Check the length is not exceed the file length ?

Aitozi (approved):
It looks very nice, only left two minor comments

docs/layouts/shortcodes/generated/core_configuration.html
```diff
@@ -46,7 +46,7 @@
         </tr>
         <tr>
             <td><h5>cache-page-size</h5></td>
-            <td style="word-wrap: break-word;">16 kb</td>
+            <td style="word-wrap: break-word;">64 kb</td>
```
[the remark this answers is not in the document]
JingsongLi: 64 kb has large compression ratio.

paimon-common/src/main/java/org/apache/paimon/io/UncompressedPageFileInput.java
JingsongLi: Here should throw exception for this situation, readFully will throw exception.

PR #2828, opened 2024-01-31
[hive] Support HiveCatalog for Hive 3.x and certain metastore client classes with constructors only for 2.x

tsreaper:
### Purpose

Some metastore client classes (for example, aliyun DLF) only provides constructor for Hive 2.x (with arguments `HiveConf, HiveMetaHookLoader, Boolean`). However metastore protocols are compatible between Hive 2.x and 3.x so these metastore client classes can also support Hive 3.x, just that Hive 3.x requires the constructor arguments to be `Configuration, HiveMetaHookLoader, Boolean` by default.

This PR explicitly specify constructor arguments in `RetryingMetaStoreClientFactory` to support such scenario.

### Tests

This PR is tested by hand.

### API and Format

No.

### Documentation

No.


miomiocat:
+1

JingsongLi (approved):
+1

PR #2829, opened 2024-01-31
[core] Adjust default value of bucket to -1

JingsongLi:
### Purpose

The behavior of defaulting to bucket 1 is not good. It forcibly limits the single parallelism processing and is not distributed processing at all, which is not conducive to users using it out of the box.

So this PR modifies this default value to -1. But we should pay attention to compatibility issues.

### API and Format

For table without bucket option:

- New table: 
     - New Read&Write: Default bucket will be -1.
     - Old Read&Write: There will be problems. 
- Old table: 
     - New Read&Write: Introduce version to TableSchema, old version schema will be added bucket = 1.
     - Old Read&Write: Nothing change.

### DIFF
--- a/docs/layouts/shortcodes/generated/core_configuration.html
+++ b/docs/layouts/shortcodes/generated/core_configuration.html
@@ -34,7 +34,7 @@ under the License.
         </tr>
         <tr>
             <td><h5>bucket</h5></td>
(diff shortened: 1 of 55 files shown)


yuzelin (approved):
+1

PR #2830, opened 2024-01-31
[flink] Fix that sink.parallelism doesn't work in compact_database action divided mode

yuzelin:
### Purpose

In divided mode, the dynamic conf isn't applied.


JingsongLi (approved):
+1

PR #2831, opened 2024-01-31
[core] Fix deserialization of sequenceNumber in lookup

Zouxxyy:
### Purpose

When lookup, we use `DataOutputSerializer` to serialize sequenceNumber by `writeLong`,  which is BigEndian default, so when read, we need use `getLongBigEndian`


JingsongLi:
+1

PR #2832, opened 2024-02-01
[flink][common] Support to ignore delete/update_before records in flink lookup table

bridgeDream:
### Purpose

#2833
Support to ignore delete/update_before records in flink lookup table.

### Tests
None

### API and Format
None

### Documentation
None


bridgeDream:
@JingsongLi 
FYI, Thanks!

bridgeDream:
@JingsongLi  Can you take a look？

JingsongLi:
Hi @bridgeDream , now delete/ub records will be ignored before writing. So we don't need this feature now.

PR #2834, opened 2024-02-01
[Benchmark] Add table read benchmark

Aitozi:
### Purpose

Add a benchmark to measure the table read performance


JingsongLi (approved):
+1

PR #2836, opened 2024-02-01
[core] Add statistics to snapshot system table

Zouxxyy:
### Purpose

Linked issue: close #xxx


paimon-core/src/main/java/org/apache/paimon/table/system/SnapshotsTable.java
```diff
@@ -260,7 +267,12 @@ private InternalRow toRow(Snapshot snapshot, FileStoreTable dataTable) {
                     snapshot.changelogRecordCount(),
                     plan.files(FileKind.ADD).size(),
                     plan.files(FileKind.DELETE).size(),
-                    snapshot.watermark());
+                    snapshot.watermark(),
+                    statsFileHandler
+                            .readStats(snapshot)
+                            .map(Statistics::toString)
```
JingsongLi, line 273: Maybe it is not a good idea to make string for a whole statistics... To large.

Maybe we can have some other system table to show full statistics?

YannByron: +1.

Zouxxyy:
closed is for #3172 has already implemente this

PR #2837, opened 2024-02-01
[core] Fix calculation of sort-spill-threshold conf

Zouxxyy:
### Purpose

Linked issue: close #xxx


JingsongLi (changes requested):
Can you explain this fixing?

PR #2839, opened 2024-02-02
[core] Forbid to change bucket number through dynamic options

yuzelin:
### Purpose

Linked issue: close #xxx


paimon-core/src/main/java/org/apache/paimon/table/AbstractFileStoreTable.java
```diff
@@ -195,6 +195,11 @@ private void checkImmutability(Map<String, String> dynamicOptions) {
                 (k, v) -> {
                     if (!Objects.equals(v, options.get(k))) {
                         SchemaManager.checkAlterTableOption(k);
+
+                        if (CoreOptions.BUCKET.key().equalsIgnoreCase(k)) {
```
JingsongLi, line 199: should check equality between old and new

yuzelin: Have checked:
```
if (!Objects.equals(v, options.get(k))) {
   xxx
}
```

JingsongLi (approved):
+1

PR #2840, opened 2024-02-02
[hotfix] Add incrementSafely in MathUtils

Zouxxyy:
### Purpose

Add a no overflow version of `Math.incrementExact` for int increment sately

Math.java (jdk)

```
    public static int incrementExact(int a) {
        if (a == Integer.MAX_VALUE) {
            throw new ArithmeticException("integer overflow");
        }

        return a + 1;
    }
```

MathUtils.java (paimon)

```
    public static int incrementSafely(int a) {
        if (a == Integer.MAX_VALUE) {
            return a;
        }
        return a + 1;
    }
```

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
@@ -1347,7 +1347,9 @@ public class CoreOptions implements Serializable {
         // By default, this ensures that the compaction does not fall to level 0, but at least to
         // level 1
         Integer numLevels = options.get(NUM_LEVELS);
-        numLevels = numLevels == null ? numSortedRunCompactionTrigger() + 1 : numLevels;
+        if (numLevels == null) {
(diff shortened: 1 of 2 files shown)


paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -1338,16 +1338,18 @@ public Duration optimizedCompactionInterval() {
     public int numSortedRunStopTrigger() {
         Integer stopTrigger = options.get(NUM_SORTED_RUNS_STOP_TRIGGER);
         if (stopTrigger == null) {
-            stopTrigger = numSortedRunCompactionTrigger() + 1;
+            stopTrigger = MathUtils.incrementSafely(numSortedRunCompactionTrigger());
         }
-        return Math.max(numSortedRunCompactionTrigger(), stopTrigger);
```
JingsongLi, line 1343: keep max here

JingsongLi (approved):
+1

PR #2845, opened 2024-02-02
[orc] Optimize ensureSize for orc ColumnVector

tsreaper:
### Purpose

Currently `ColumnVector#ensureSize` will resize to the required size each time. The complexity of this implementation is $\mathcal{O}(n^2)$.

Instead, this PR multiplies the array size by $2$ each time. Thus optimizing the complexity to $\mathcal{O}(n\log n)$.

### Tests

### API and Format

No.

### Documentation

No.


JingsongLi (approved):
+1

PR #2846, opened 2024-02-04
[core] Remove the duplicated ContinuousCompactorFollowUpScanner

Aitozi:
### Purpose

The `ContinuousCompactorFollowUpScanner` is same with `DeltaFollowUpScanner`, So I think we can reuse the `DeltaFollowUpScanner`


Aitozi:
CC @tsreaper

JingsongLi (approved):
+1

PR #2856, opened 2024-02-04
[Bug] Ignore the consumer-id in batch job

### LINKED ISSUE #2855: [Bug] The scan result is empty when run with consumer-id in batch mode
### Search before asking

### Paimon version

0.6.0

### Compute Engine

Flink

### Minimal reproduce step

consumer with a consumer-id in batch mode

the result is empty

### Anything else?

_No response_

Aitozi:
### Purpose

The consumer id should work for stream mode only, otherwise the result of the batch scan is empty. close #2855


JingsongLi (approved):
+1

PR #2857, opened 2024-02-05
[core] Avoid read the snapshot file twice

Aitozi:
### Purpose

Directly pass the Snapshot to reader to avoid read the snapshot file twice


JingsongLi:
Can you rebase master?

Aitozi:
> Can you rebase master?

updated

JingsongLi (approved):
+1

PR #2859, opened 2024-02-06
[core] Use overwrite instead of delete-and-write when commit hint file

Aitozi:
### Purpose

Currently, we use delete-and-write to overwrite the hint file, it's better to run with the atomic overwrite, so there is no intermediate state for the hint file.


JingsongLi:
+1

PR #2863, opened 2024-02-07
[Core]Clean orphan files for branch

TaoZex:
### Purpose
issue:https://github.com/apache/incubator-paimon/issues/2769
subtask of https://github.com/apache/incubator-paimon/issues/1795


paimon-core/src/main/java/org/apache/paimon/utils/SnapshotManager.java
```diff
@@ -253,6 +256,17 @@ public List<Snapshot> safelyGetAllSnapshots() throws IOException {
                         .map(this::snapshotPath)
                         .collect(Collectors.toList());
```
schnappi17, line 258: Maybe we can enrich the `BranchManager::branches()` to get all the branch set info? You can rebase master to see the implementation.

TaoZex: Good idea！I will fix it.

schnappi17:
+1 for this.

PR #2866, opened 2024-02-07
[Core] Support jdbc catalog

sunxiaojian:
Support jdbc catalog

Linked issue : https://github.com/apache/incubator-paimon/issues/841


sunxiaojian:
@JingsongLi PTAL

sunxiaojian:
@FangYongs @yuzelin @JingsongLi  If you have time, please help review it, thx

JingsongLi:
Thanks @sunxiaojian for your contribution.

This PR looks good! I will take a look this week~

docs/content/how-to/creating-catalogs.md
```diff
@@ -175,3 +176,185 @@ Using the table option facilitates the convenient definition of Hive table param
 Parameters prefixed with `hive.` will be automatically defined in the `TBLPROPERTIES` of the Hive table. 
 For instance, using the option `hive.table.owner=Jon` will automatically add the parameter `table.owner=Jon` to the table properties during the creation process.
 
+
+
+## Creating a Catalog with Filesystem Metastore
```
JingsongLi, line 181: Why copy this?

docs/content/how-to/creating-catalogs.md
```diff
@@ -175,3 +176,185 @@ Using the table option facilitates the convenient definition of Hive table param
...
+
+```sql
+CREATE CATALOG my_jdbc WITH (
+    'type' = 'paimon',
+    'metastore' = 'jdbc',
+    -- 'uri' = 'jdbc:mysql://<host>:<port>/<databaseName>',
+    -- 'jdbc.user' = '...', 
+    -- 'jdbc.password' = '...', 
+    -- 'initialize-catalog-tables'='true'
+    -- 'warehouse' = 'hdfs:///path/to/warehouse',
```
JingsongLi, line 347: warehouse should be required.

docs/content/how-to/creating-catalogs.md
```diff
@@ -175,3 +176,185 @@ Using the table option facilitates the convenient definition of Hive table param
...
+| postgres      | postgresql           | [Download](https://mvnrepository.com/artifact/org.postgresql/postgresql)   |
+
+```sql
+CREATE CATALOG my_jdbc WITH (
+    'type' = 'paimon',
+    'metastore' = 'jdbc',
+    -- 'uri' = 'jdbc:mysql://<host>:<port>/<databaseName>',
+    -- 'jdbc.user' = '...', 
+    -- 'jdbc.password' = '...', 
+    -- 'initialize-catalog-tables'='true'
```
JingsongLi, line 346: Do we need this one? What is the scene?

docs/content/how-to/creating-catalogs.md
```diff
@@ -175,3 +176,185 @@ Using the table option facilitates the convenient definition of Hive table param
...
+    -- 'warehouse' = 'hdfs:///path/to/warehouse',
+);
+
+USE CATALOG my_jdbc;
+```
+You can define any connection parameters for a database with the prefix "jdbc.".
+
+You can define any default table options with the prefix `table-default.` for tables created in the catalog.
+
+Also, you can create [FlinkGenericCatalog]({{< ref "engines/flink" >}}).
```
JingsongLi, line 356: No, `FlinkGenericCatalog` must use HiveCatalog.

paimon-core/src/main/java/org/apache/paimon/jdbc/JdbcUtils.java
```diff
@@ -0,0 +1,646 @@
...
+                    + LOCK_NAME
+                    + " VARCHAR(1000) NOT NULL,"
+                    + ACQUIRED_AT
+                    + " TIMESTAMP NULL DEFAULT NULL,"
+                    + "PRIMARY KEY ("
+                    + LOCK_NAME
+                    + ")"
+                    + ")";
+
+    static final String DISTRIBUTED_LOCK_ACQUIRE_SQL =
```
JingsongLi, line 289: Can you explain this? How to lock?

coolderli:
@sunxiaojian Thanks for your work. This is really what we want!

paimon-core/src/main/java/org/apache/paimon/jdbc/JdbcUtils.java
```diff
@@ -0,0 +1,646 @@
...
+import java.util.Map;
+import java.util.Properties;
+import java.util.Set;
+import java.util.function.Consumer;
+import java.util.stream.Stream;
+
+/** Util for jdbc catalog. */
+public class JdbcUtils {
+    private static final Logger LOG = LoggerFactory.getLogger(JdbcUtils.class);
+    public static final String METADATA_LOCATION_PROP = "metadata_location";
```
zhongyujiang, line 43: If I understand correctly, Paimon considers the commit successful once the snapshot file is successfully committed to the file system.  Storing the current metadata location in the catalog metastore will cause consistency problems because the two operations of committing the snapshot file to the file system and committing it to the metastore are not atomic. So there seems to be no point in storing this metadata location？ @JingsongLi Can you please review this? Thanks.

paimon-core/src/main/java/org/apache/paimon/jdbc/JdbcUtils.java
```diff
@@ -0,0 +1,646 @@
...
+                            connection.prepareStatement(DISTRIBUTED_LOCK_RELEASE_SQL)) {
+                        preparedStatement.setString(1, lockName);
+                        return preparedStatement.executeUpdate() > 0;
+                    }
+                });
+    }
+
+    private static int tryClearExpireLock(JdbcClientPool connections, String lockName, long timeout)
+            throws SQLException, InterruptedException {
+        long expirationTimeMillis = System.currentTimeMillis() - timeout * 1000;
```
zhongyujiang, line 634: I'm not sure I understand this, could you please explain?

paimon-core/src/main/java/org/apache/paimon/jdbc/JdbcUtils.java
```diff
@@ -0,0 +1,646 @@
...
+        if (affectedRows > 0) {
+            LOG.debug("Successfully cleared " + affectedRows + " lock records");
+        }
+        return connections.run(
+                connection -> {
+                    try (PreparedStatement preparedStatement =
+                            connection.prepareStatement(DISTRIBUTED_LOCK_ACQUIRE_SQL)) {
+                        preparedStatement.setString(1, lockName);
+                        preparedStatement.setTimestamp(
+                                2, new Timestamp(System.currentTimeMillis()));
```
zhongyujiang, line 611: So we are using timestamp of clients as `acquired_at`? I think this makes no sense since clients can be distributed on different machines, which have no consistent system time. Since `acquired_at` is the timestamp of acquiring the lock, I think we should use the system time of the metastore.

paimon-core/src/main/java/org/apache/paimon/jdbc/JdbcUtils.java
```diff
@@ -0,0 +1,646 @@
...
+
+        String values = String.join(",", Collections.nCopies(size, String.valueOf('?')));
+        sqlStatement.append("(").append(values).append(")");
+        return sqlStatement.toString();
+    }
+
+    public static boolean acquire(JdbcClientPool connections, String lockName, long timeout)
+            throws SQLException, InterruptedException {
+        // Check and clear expire lock
+        int affectedRows = tryClearExpireLock(connections, lockName, timeout);
```
zhongyujiang, line 601: I think `timeout` is a property that should be set when acquiring the lock, not when the lock expires. Because different clients may configure different timeouts.

paimon-core/src/main/java/org/apache/paimon/jdbc/JdbcUtils.java
JingsongLi: @zhongyujiang You understanding is very correct! Yes, Paimon only consider `Snapshot` files as source of truth.
`metadata_location` is come from Iceberg, we don't need this.

paimon-core/src/main/java/org/apache/paimon/jdbc/JdbcUtils.java
sunxiaojian: > So we are using timestamp of clients as `acquired_at`? I think this makes no sense since clients can be distributed on different machines, which have no consistent system time. Since `acquired_at` is the timestamp of acquiring the lock, I think we should use the system time of the metastore.

@zhongyujiang There may indeed be the issue you mentioned, but the database used for the meta storage system is different, and the generation system time function used is also different. May require separate compatibility

sunxiaojian: The initial time setting was for metastore, but it was not compatible with all databases, so this version of the modification was made. In fact, specific database specific time functions can be set for each database

docs/content/how-to/creating-catalogs.md
sunxiaojian: > Do we need this one? What is the scene?

@JingsongLi When creating a JdbcCatalog, it will be determined based on this parameter whether to automatically create a table or not

paimon-core/src/main/java/org/apache/paimon/jdbc/JdbcUtils.java
sunxiaojian: > Can you explain this? How to lock?

@JingsongLi First, create a table for distributed locks and write a data entry to the database table using database.table as the lockId. If successful, the lock acquisition is considered successful. If the primary key conflicts, it is considered a failure

sunxiaojian:
@JingsongLi @zhongyujiang Thank you for the review,   I have fixed the issues raised above.

docs/content/how-to/creating-catalogs.md
JingsongLi: When does it not need to be initialized? You added a public API. Please explain why, how to use.

JingsongLi:
Hi @sunxiaojian , you should make sure that every comment is resolved.

docs/content/how-to/creating-catalogs.md
JingsongLi: Or you can remove this option.

sunxiaojian:
> Hi @sunxiaojian , you should make sure that every comment is resolved.

@JingsongLi All have been resolved

docs/content/how-to/creating-catalogs.md
```diff
@@ -175,3 +176,43 @@ Using the table option facilitates the convenient definition of Hive table param
...
+    'catalog-name'='jdbc'
+    'initialize-catalog-tables'='true'
+    'warehouse' = 'hdfs:///path/to/warehouse'
+);
+
+USE CATALOG my_jdbc;
+```
+You can define any connection parameters for a database with the prefix "jdbc.".
+
+You can define the "initialize-catalog-tables" configuration to automatically create tables required for the initial JdbcCatalog. If it is true, tables will be automatically created when initializing JdbcCatalog. If it is false, tables will not be automatically created and you must manually create them.
```
JingsongLi, line 210: > If it is false, tables will not be automatically created and you must manually create them.

Why is there such a need? I don't understand.

docs/content/how-to/creating-catalogs.md
```diff
@@ -175,3 +176,43 @@ Using the table option facilitates the convenient definition of Hive table param
...
+| postgres      | postgresql           | [Download](https://mvnrepository.com/artifact/org.postgresql/postgresql)   |
+
+```sql
+CREATE CATALOG my_jdbc WITH (
+    'type' = 'paimon',
+    'metastore' = 'jdbc',
+    'uri' = 'jdbc:mysql://<host>:<port>/<databaseName>',
+    'jdbc.user' = '...', 
+    'jdbc.password' = '...', 
+    'catalog-name'='jdbc'
```
JingsongLi, line 201: Here we already have the concept of a catalog name, which is `my_jdbc`.

Maybe this option can be renamed to `store-key` or something else?

docs/content/how-to/creating-catalogs.md
```diff
@@ -175,3 +176,43 @@ Using the table option facilitates the convenient definition of Hive table param
...
+    'initialize-catalog-tables'='true'
```
JingsongLi, line 202: you forgot `,`

paimon-core/src/main/java/org/apache/paimon/ClientPoolImpl.java
```diff
@@ -0,0 +1,155 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.io.Closeable;
+import java.util.ArrayDeque;
+import java.util.Deque;
+
+import static org.apache.paimon.utils.Preconditions.checkState;
+
+/** Source: [core/src/main/java/org/apache/iceberg/ClientPoolImpl.java]. */
+public abstract class ClientPoolImpl<C, E extends Exception>
```
JingsongLi, line 31: Move them to `paimon-common`.

docs/content/how-to/creating-catalogs.md
JingsongLi: at least, `jdbc.catalog-name`, you need documentation to explain that it is designed to store multiple catalogs, and it is different from the catalog name you use in Flink SQL or Spark SQL.

docs/content/how-to/creating-catalogs.md
sunxiaojian: > Why is there such a need? I don't understand.

Some online business databases do not allow table creation through jdbc and must be submitted for approval in advance

sunxiaojian: The final decision is not to keep this configuration. It is recommended that users enable automatic table creation permissions for independent databases, which would be better

JingsongLi: > Some online business databases do not allow table creation through jdbc and must be submitted for approval in advance

You can document this.

sunxiaojian: > > Some online business databases do not allow table creation through jdbc and must be submitted for approval in advance
> 
> You can document this.

Most users may not need it, I will remove this parameter and add it back if there is a strong requirement.

sunxiaojian:
@JingsongLi All have been resolved

paimon-core/src/main/java/org/apache/paimon/jdbc/JdbcCatalogLock.java
```diff
@@ -0,0 +1,137 @@
...
+        this.acquireTimeout = acquireTimeout;
+        this.catalogName = catalogName;
+    }
+
+    @Override
+    public <T> T runWithLock(String database, String table, Callable<T> callable) throws Exception {
+        String lockUniqueName = String.format("%s.%s.%s", catalogName, database, table);
+        lock(lockUniqueName);
+        try {
+            return callable.call();
```
zhongyujiang, line 56: What happens if the lock expires before the commit is completed? Will there be conflicts? Does this break the atomicity of commits?

For example, let's consider the first column as logical timestamp, and the second column is the events that occurred at that time:
1   client-1 successfully aquired a lock of table `a`
2   client-1 starts committing but gets stuck for some reason
5   the lock of table `a ` timeouts
6   client-2 cleaned the expired lock of table `a` and acquired a new lock of table `a`
7   client-2 starts committing
8   client-2 made a successful committment
9   client-1 can start committing

In this case, client-1's commit should fail, but here client-1 will continue to commit. If the underlying file system allows overwriting of files with the same name, then client-1's commit will overwrite client-2's commit.

sunxiaojian: > What happens if the lock expires before the commit is completed? Will there be conflicts? Does this break the atomicity of commits?
> 
> For example, let's consider the first column as logical timestamp, and the second column is the events that occurred at that time: 1 client-1 successfully aquired a lock of table `a` 2 client-1 starts committing but gets stuck for some reason 5 the lock of table `a ` timeouts 6 client-2 cleaned the expired lock of table `a` and acquired a new lock of table `a` 7 client-2 starts committing 8 client-2 made a successful committment 9 client-1 can start committing
> 
> In this case, client-1's commit should fail, but here client-1 will continue to commit. If the underlying file system allows overwriting of files with the same name, then client-1's commit will overwrite client-2's commit.

I understand that this should require checking when writing to ensure the order of submissions, which should not be related to distributed locks, or setting a reasonable timeout time

sunxiaojian: You can start a new thread renewal timeout to solve the problem, but you can also set a reasonable timeout to solve it

JingsongLi: For the current strategy, it is difficult for this situation to occur because currently using lock is only renaming the file, and this operation should be completed quickly.

JingsongLi:
@zhongyujiang Hi, do you have other comments or concerns?

paimon-core/src/main/java/org/apache/paimon/jdbc/JdbcClientPool.java
```diff
@@ -0,0 +1,81 @@
...
+
+    public JdbcClientPool(int poolSize, String dbUrl, Map<String, String> props) {
+        super(poolSize, SQLNonTransientConnectionException.class, true);
+        properties = props;
+        this.dbUrl = dbUrl;
+        Matcher matcher = PROTOCOL_PATTERN.matcher(dbUrl);
+        if (matcher.matches()) {
+            this.protocol = matcher.group(1);
+        } else {
+            throw new RuntimeException("Valid Jdbc url failure: " + dbUrl);
```
zhongyujiang, line 48:
```suggestion
            throw new RuntimeException("Invalid Jdbc url: " + dbUrl);
```

docs/content/how-to/creating-catalogs.md
```diff
@@ -175,3 +176,40 @@ Using the table option facilitates the convenient definition of Hive table param
 Parameters prefixed with `hive.` will be automatically defined in the `TBLPROPERTIES` of the Hive table. 
 For instance, using the option `hive.table.owner=Jon` will automatically add the parameter `table.owner=Jon` to the table properties during the creation process.
 
+## Creating a Catalog with JDBC Metastore
+
+By using the Paimon JDBC catalog, changes to the catalog will be directly stored in relational databases such as MySQL, postgres, etc.
```
zhongyujiang, line 181: `postgres` is not supported for now, should we remove it?

zhongyujiang (commented):
Left two minor comments, I have no other worries except the above one, Thanks @sunxiaojian

docs/content/how-to/creating-catalogs.md
sunxiaojian: > `postgres` is not supported for now, should we remove it?

it's just that locking is not supported

docs/content/how-to/creating-catalogs.md
```diff
@@ -175,3 +176,40 @@ Using the table option facilitates the convenient definition of Hive table param
...
+| postgres      | postgresql           | [Download](https://mvnrepository.com/artifact/org.postgresql/postgresql)   |
+
+```sql
+CREATE CATALOG my_jdbc WITH (
+    'type' = 'paimon',
+    'metastore' = 'jdbc',
+    'uri' = 'jdbc:mysql://<host>:<port>/<databaseName>',
+    'jdbc.user' = '...', 
+    'jdbc.password' = '...', 
+    'store-key'='jdbc',
```
JingsongLi, line 201: Can you modify this to `jdbc.catalog-key`?

docs/content/how-to/creating-catalogs.md
JingsongLi: Can you document this? only SQLITE and MYSQL supports catalog lock

docs/content/how-to/creating-catalogs.md
sunxiaojian: > Can you modify this to `jdbc.catalog-key`?

The prefix "jdbc." is used to connect to the database configuration and can be directly parsed and placed in the connection configuration. If it is necessary to replace it, it can be replaced with "catalog-key"?

sunxiaojian:
@JingsongLi The above issues have been resolved, please review again

JingsongLi (approved):
+1 Thanks @sunxiaojian and @zhongyujiang

PR #2867, opened 2024-02-10
[Core]Add filesystem catalog lock

TaoZex:
### Purpose

When a process in file system attempts to acquire a lock, it tries to create a specific file in the file system. The existence of this file represents the state of the lock. If the file does not exist, it indicates that the lock is available, and the process can successfully acquire the lock. If the file already exists, it indicates that the lock has been acquired by another process.

subtask of https://github.com/apache/incubator-paimon/issues/2824


FangYongs:
@TaoZex In fact, what we want is not a filesystem lock, but to optimize the catalog lock interface and support configuring catalog lock in the filesystem catalog. This will be quite different from the implementation of this PR.

TaoZex:
> @TaoZex其实我们想要的不是文件系统锁，而是优化目录锁接口，支持在文件系统目录中配置目录锁。这将与此 PR 的实施有很大不同。

Thanks for your reply, I will close this pr.

TaoZex:
> @TaoZex In fact, what we want is not a filesystem lock, but to optimize the catalog lock interface and support configuring catalog lock in the filesystem catalog. This will be quite different from the implementation of this PR.

Thanks for your reply, I will close this pr.

FangYongs:
@TaoZex You can help to review https://github.com/apache/incubator-paimon/pull/2872 if you are interested in this feature

PR #2869, opened 2024-02-16
[doc] Modify the wrong parameters of flink action in the document

herefree:
### Purpose
Modify the wrong parameters of flink action in the document,create-tag should be changed to create_tag and delete-tag should be changed to delete_tag


herefree:
@JingsongLi help review

JingsongLi (approved):
+1

PR #2871, opened 2024-02-18
[Core][Flink] Flink job can throw exception in streaming reading mode when upstream paimon is deleted

bridgeDream:
### Purpose
Add a feature to check whether ID of the latest snapshot is valid when scanning the next snapshot in `InnerStreamTableScanImpl` , which can prompt whether the paimon table has been deleted or rebuilt.

Especially in the Flink job when reading paimon, when an upstream table is deleted, an exception can be thrown, causing the flink job failover instead of constantly hanging. Although it appears to be running normally, it is actually unable to consume data and users can't receive an alarm.

### Tests
None

### API and Format
None

### Documentation
None


bridgeDream:
@JingsongLi Can you take a look?

JingsongLi (commented):
Can you add a test for this? @bridgeDream

JingsongLi (commented):
I thought about it for a moment, and this situation is still rare, and this check will bring a cost to every stream read, so I think we should close this PR first.

Feel free to re-open this. @bridgeDream

PR #2875, opened 2024-02-19
[core] Support drop delete in rewriteChangelogCompaction

Zouxxyy:
### Purpose

Support drop delete in rewriteChangelogCompaction

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/ChangelogMergeTreeRewriter.java
+++ b/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/ChangelogMergeTreeRewriter.java
@@ -93,7 +93,7 @@ public abstract class ChangelogMergeTreeRewriter extends MergeTreeCompactRewrite
     public CompactResult rewrite(
             int outputLevel, boolean dropDelete, List<List<SortedRun>> sections) throws Exception {
         if (rewriteChangelog(outputLevel, dropDelete, sections)) {
-            return rewriteChangelogCompaction(outputLevel, sections, true);
+            return rewriteChangelogCompaction(outputLevel, sections, dropDelete, true);
         } else {
             return rewriteCompaction(outputLevel, dropDelete, sections);
         }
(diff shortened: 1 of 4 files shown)


paimon-spark/paimon-spark-common/src/test/scala/org/apache/paimon/spark/sql/LookupCompactionTest.scala
```diff
@@ -78,6 +78,22 @@ class LookupCompactionTest extends PaimonSparkTestBase {
     }
   }
 
+  test("Paimon lookup compaction: drop delete") {
```
YannByron, line 81: Suggest that move the related test to paimon-core.

paimon-core/src/main/java/org/apache/paimon/mergetree/compact/ChangelogMergeTreeRewriter.java
```diff
@@ -132,8 +136,9 @@ private CompactResult rewriteChangelogCompaction(
 
             while (iterator.hasNext()) {
                 ChangelogResult result = iterator.next();
-                if (rewriteCompactFile && result.result() != null) {
-                    compactFileWriter.write(result.result());
+                KeyValue keyValue = result.result();
+                if (rewriteCompactFile && keyValue != null && (!dropDelete || keyValue.isAdd())) {
```
Aitozi, line 140: what about also wrap the `ConcatRecordReader.create(sectionReaders)` in `DropDeleteReader` when `dropDelete` is needed?

Zouxxyy: No, this mode we need read delete

Aitozi: Got it

paimon-core/src/main/java/org/apache/paimon/mergetree/compact/ChangelogMergeTreeRewriter.java
```diff
@@ -170,6 +175,7 @@ public CompactResult upgrade(int outputLevel, DataFileMeta file) throws Exceptio
                     outputLevel,
                     Collections.singletonList(
                             Collections.singletonList(SortedRun.fromSingle(file))),
+                    false,
```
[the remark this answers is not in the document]
Zouxxyy: No need, because we don’t need to rewrite when upgrading, even if rewrite is required, this drop delete must be false too (can take a look at the conditions for drop delete and rewrite in upgrade).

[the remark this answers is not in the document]
Aitozi: Thanks for explanation. Looks reasonable, if rewrite is required, the file level always in level 0.

JingsongLi (approved):
+1

PR #2879, opened 2024-02-21
[Flink]Avoid repeatedly generating execution plans

waywtdcc:
### Purpose

Avoid repeatedly generating execution plans


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/StaticFileStoreSource.java
```diff
@@ -87,13 +92,17 @@ public SplitEnumerator<FileStoreSourceSplit, PendingSplitsCheckpoint> restoreEnu
...
         FileStoreSourceSplitGenerator splitGenerator = new FileStoreSourceSplitGenerator();
-        TableScan scan = readBuilder.newScan();
-        // register scan metrics
-        if (context.metricGroup() != null) {
-            ((InnerTableScan) scan)
-                    .withMetricsRegistry(new FlinkMetricRegistry(context.metricGroup()));
+        if (this.splits == null) {
+            TableScan scan = readBuilder.newScan();
+            // register scan metrics
+            if (context.metricGroup() != null) {
```
JingsongLi, line 98: always register metrics

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/DataTableSource.java
```diff
@@ -250,13 +251,16 @@ private DataStream<RowData> configureSource(
                             options.get(FlinkConnectorOptions.INFER_SCAN_MAX_PARALLELISM));
         }
 
-        return sourceBuilder.withParallelism(parallelism).withEnv(env).build();
+        return sourceBuilder
+                .withParallelism(parallelism)
+                .withEnv(env)
+                .withSplits(this.splits)
```
JingsongLi, line 257: only use this way for few splits. `if (files in splits less than 100)`.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/DataTableSource.java
```diff
@@ -93,6 +93,7 @@ public class DataTableSource extends FlinkTableSource {
     private SplitStatistics splitStatistics;
 
     @Nullable private List<String> dynamicPartitionFilteringFields;
+    private List<Split> splits;
```
JingsongLi, line 96: You should set splits to null if there are some filter or limit pushdown.

PR #2883, opened 2024-02-21
[license] Auto add and unify license header for java & scala file

Zouxxyy:
### Purpose

Auto add and unify license header for java & scala file, for reviewer, this PR has three parts (I have alreadly divide them):

1.  modify spotless pom
2. **move other copyright after import** (Because the current correction plan automatically overwrites everything before `package`, **so we need to move other information to avoid being overwritten**)
3. just auto spotless

### DIFF
--- a/copyright.txt
+++ b/copyright.txt
@@ -1,0 +1,18 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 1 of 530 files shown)


JingsongLi (changes requested):
Hi @Zouxxyy 
It is not good to modify all files frequently. Can you just do the format to scale files only? I don't think we necessarily need to adjust the header of Java files.

JingsongLi (approved):
+1

PR #2885, opened 2024-02-21
[core] Fix bug that changelog may not be produced when setting changelog-producer.lookup-wait to false

tsreaper:
### Purpose

In `AbstractFileStoreWrite#prepareCommit`, if a writer does not produce any new `CommitIncrement` and all previous `CommitIncrement`s have been committed, we'll close this writer.

However, this writer may have an ongoing compaction (possibly triggered by lookup changelog producer). If we ignore this compaction and close the writer, we'll interrupt the compaction and thus interrupt changelog producing.

This PR fixes the bug by checking if a writer has an ongoing compaction.

### Tests

* `TableWriteTest#testChangelogWhenNotWaitForCompaction`

### API and Format

No.

### Documentation

No.


JingsongLi (approved):
+1

PR #2890, opened 2024-02-22
[flink] The dim table in lookup join should ignore scan options

yuzelin:
### Purpose

If user passes scan options to dim table through dynamic option, it will cause exception.
This PR fix it by clear the scan options when loading data from dim table.


JingsongLi (approved):
+1

PR #2894, opened 2024-02-23
[core] Last value and nested_update agg function support retract

JingsongLi:
### Purpose

The `last_value` and `last_non_null_value` just set field to null when accept retract messages.

This is good to most cases.


yuzelin (approved):
+1

yuzelin (approved):
+1

PR #2895, opened 2024-02-23
[core] Sequence Auto padding should based on incremental id

JingsongLi:
### Purpose

The system time is very inaccurate and basically useless. We should use self increasing IDs, which will at least have some effect.


Zouxxyy:
+1

PR #2900, opened 2024-02-26
[core] Remove legacy stats FieldStatsArraySerializer.fromBinary

JingsongLi:
### Purpose

We have introduced `FieldStatsArraySerializer.evolution`, we don't need keep another evolution method `fromBinary`. This PR aims to clean some usage in tests code.


yuzelin (approved):
+1

Issue #2898, opened 2024-02-23
[Feature] Introduce deletion vectors for primary key table


Zouxxyy:
### Motivation

Position deletion is a solution to implement the Merge-On-Read (MOR) structure, which has been adopted by other formats such as Iceberg and Delta. By combining with Paimon's LSM tree, we can create a new mode with deletion vectors (bitmap to identity which row id deleted) index file unique to Paimon.

Under this mode, extra overhead (lookup and write deletion vectors index file) will be introduced during writing, but during reading, data can be directly retrieved using "data + filter with deletion vector", avoiding additional merge costs between different files. Furthermore, this mode can be easily integrated into native engine solutions like Spark + Gluten in the future, thereby significantly enhancing read performance.

PIP: https://cwiki.apache.org/confluence/x/Tws4EQ

JingsongLi:
We have two things remaining:
1. Since there is no need to merge when reading, in this mode, we can support filter pushdown of non-PK fields!
2. Documentation for deletion vectors.

JingsongLi:
3. Supports dv with partial-update and aggregate. Looks like current implementation is not work.
4. Supports dv with first-row.

JingsongLi:
Thanks @Zouxxyy , all finished!

PR #2909, opened 2024-02-27
[core][DV] Support obtain row position when reading orc and parquet

Zouxxyy:
### Purpose

This is the first step of #2898, in this PR:

- add new config `deletion-vectors.enabled` which default is false and should only be used with PK table and orc/parquet format currently
- add system col `_ROW_POSITION`
- add `rowPosition` in `KeyValue` and set it when `deletion-vectors.enabled` is true
- support obtain `rowPosition` when reading orc and parquet


JingsongLi (changes requested):
In my opinion, having an additional field for `KeyValue` may have a significant impact, for example, after Merging, this field will be meaningless.

A less invasive approach could be to introduce a new `RecordReader.RecordIterator`, for example, adding the `RecordWithPositionIterator` class:
```
public interface RecordWithPositionIterator<T> extends RecordIterator<T> {
    long currentPosition();
}
```

Zouxxyy:
closed by new implement https://github.com/apache/incubator-paimon/pull/2916

PR #2910, opened 2024-02-27
[core] Optimize partition expire by filtering partitions while scanning

tsreaper:
### Purpose

Currently, `PartitionExpire` will scan all manifest entries and merge them in memory before extracting the partitions. This is very inefficient any may cause OOM.

This PR optimizes `PartitionExpire` by filtering partitions while scanning, thus decreasing the number of manifest entries to be merged.

### Tests

### API and Format

No.

### Documentation

No.


JingsongLi (approved):
+1

PR #2913, opened 2024-02-27
[Bug] Fix the incorrect lookup join result when lookup from level0

### LINKED ISSUE #2912: [Bug] The lookup join result is not correct when search from level0
### Search before asking

### Paimon version

0.6.0

### Compute Engine

Flink

### Minimal reproduce step

Currently, in `LookupUtils` it directly return when there is no match key in level0 files. It's incorrect.

```
if (startLevel == 0) {
     return level0Lookup.apply(key, levels.level0());
}
```

.

### Anything else?

_No response_

Aitozi:
### Purpose

Linked issue: close #2912


JingsongLi (approved):
+1 after test passing

Aitozi:
@JingsongLi Do we need to cherry-pick to release-0.7 branch ?

PR #2916, opened 2024-02-28
[core] Introduce RecordWithPositionIterator interface

Zouxxyy:
### Purpose

This is the first step of https://github.com/apache/incubator-paimon/issues/2898, in this PR:
- introduce RecordWithPositionIterator
- implement it with orc and parquet format

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/reader/RecordReader.java
+++ b/paimon-common/src/main/java/org/apache/paimon/reader/RecordReader.java
@@ -22,6 +22,7 @@ import org.apache.paimon.annotation.Public;
 import org.apache.paimon.utils.CloseableIterator;
 import org.apache.paimon.utils.ConsumerWithIOException;
 import org.apache.paimon.utils.Filter;
+import org.apache.paimon.utils.Pair;
 
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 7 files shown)


paimon-common/src/main/java/org/apache/paimon/reader/RecordReader.java
```diff
@@ -142,6 +143,33 @@ default void forEachRemaining(Consumer<? super T> action) throws IOException {
         }
     }
 
+    /**
+     * Performs the given action for each remaining element with row position in {@link
+     * RecordReader} until all elements have been processed or the action throws an exception.
+     */
+    default void forEachRemainingWithPosition(Consumer<? super Pair<Long, T>> action)
```
JingsongLi, line 150: Use `BiConsumer`

paimon-format/src/main/java/org/apache/paimon/format/parquet/ParquetReaderFactory.java
```diff
@@ -393,7 +393,9 @@ public void recycle() {
         }
 
         public RecordIterator<InternalRow> convertAndGetIterator() {
-            result.set(columnarBatch.getNumRows());
+            result.reset(columnarBatch.getNumRows());
+            // Since we currently don't implement parquet's filter push down, row position is always
```
JingsongLi, line 397: Can we implement this correctly? I think maybe we will implement parquet's filter push down in future. Leaving a hole here can easily lead to bugs.

JingsongLi:
+1

PR #2918, opened 2024-02-28
[core] orc/parquet reader obtain the fileSize from metadata

wg1026688210:
### Purpose

In order to reduce the access of FileSystem, the format reader of orc/parquet use the fileSize from metadata

Linked issue: related #2978 

### Tests
KeyValueFileReadWriteTest#testReaderUseFileSizeFromMetadata

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/format/FormatReaderFactory.java
+++ b/paimon-common/src/main/java/org/apache/paimon/format/FormatReaderFactory.java
@@ -31,6 +31,11 @@ public interface FormatReaderFactory extends Serializable {
 
     RecordReader<InternalRow> createReader(FileIO fileIO, Path file) throws IOException;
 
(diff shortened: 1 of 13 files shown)


paimon-common/src/main/java/org/apache/paimon/format/FormatReaderFactory.java
```diff
@@ -31,6 +31,11 @@ public interface FormatReaderFactory extends Serializable {
 
     RecordReader<InternalRow> createReader(FileIO fileIO, Path file) throws IOException;
```
JingsongLi, line 32: Can you introduce a `Context` for `FormatReaderFactory`?

wg1026688210: done

JingsongLi (approved):
Looks good to me! Thanks @wg1026688210

PR #2922, opened 2024-02-28
[core] Read optimized table supports value filter pushdown

JingsongLi:
### Purpose

The RO table no longer needs to be merged, so we should enable it to have the ability to non-primary key filtering pushdown.


yuzelin (approved):
+1

PR #2927, opened 2024-03-01
[Hive]Pass Table into Split to Avoid Loading from FileSystem when createRecordReader

wg1026688210:
### Purpose

Pass Table to Split to Avoid Loading from File System when createRecordReader ,this can reduce the access of FileSystem.

Linked issue: close #xxx


paimon-hive/paimon-hive-connector-common/src/main/java/org/apache/paimon/hive/mapred/PaimonInputSplit.java
```diff
@@ -73,13 +78,28 @@ public String[] getLocations() {
...
         dataOutput.writeUTF(path);
         DataOutputSerializer out = new DataOutputSerializer(128);
         split.serialize(out);
         dataOutput.writeInt(out.length());
         dataOutput.write(out.getCopyOfBuffer());
+        writeFileStoreTable(dataOutput);
+    }
+
+    private void writeFileStoreTable(DataOutput dataOutput) throws IOException {
+        if (table == null) {
```
JingsongLi, line 96: It should be nonnull

JingsongLi (approved):
+1

PR #2929, opened 2024-03-01
[test-util] Remove redundant AssertionUtils

yuzelin:
### Purpose

There is PaimonAssertions with the same function.


JingsongLi (approved):
+1

PR #2930, opened 2024-03-01
[core] Remove bucket level metrics because metric names will stay in Flink's jobmanager forever and will cause OOM

tsreaper:
### Purpose

We encountered an issue in production that OOM occurs in Flink's jobmanager. After investigation we found that all metric names will be kept forever in jobmanager's `MetricStore`. As the number of partitions written becomes larger, there will be more and more bucket level metrics, and finally `MetricStore` can't hold them all.

This PR removes all bucket level metrics and replace some of them with aggregated metrics.

### Tests

* `AggregatedMetricsTest`.
* `CompactionMetricsTest`.
* `CompactTimerTest`.

### API and Format

No.

### Documentation

This PR affects the list of available metrics. The document is also updated.


paimon-core/src/main/java/org/apache/paimon/operation/metrics/CompactionMetrics.java
```diff
@@ -19,21 +19,34 @@
...
+
 /** Metrics to measure a compaction. */
 public class CompactionMetrics {
 
-    private static final int HISTOGRAM_WINDOW_SIZE = 100;
     private static final String GROUP_NAME = "compaction";
 
+    public static final String MAX_LEVEL0_FILE_COUNT = "maxLevel0FileCount";
+    public static final String AVG_LEVEL0_FILE_COUNT = "avgLevel0FileCount";
+    public static final String MAX_COMPACTION_THREAD_BUSY = "maxCompactionThreadBusy";
```
JingsongLi, line 39: Each task has only one backend thread, so according to this statement, busy should only have one metric.

paimon-core/src/main/java/org/apache/paimon/operation/metrics/MetricUtils.java
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.paimon.operation.metrics;
+
+import org.slf4j.Logger;
+
+/** Utils for metrics. */
+public class MetricUtils {
```
JingsongLi, line 24: I think we should avoid same name classes. Here you can rename `org.apache.paimon.flink.utils.MetricUtils` to `org.apache.paimon.flink.utils.FlinkMetricUtils`.

PR #2936, opened 2024-03-01
[core] Introduce userDefineSeqComparator for MergeSorter

JingsongLi:
### Purpose

We should separate user define sequence fields and incremental sequence id. This way, there will be no more accuracy issues. When comparing, we can first compare the user-defined sequenceField (We don't need to extract long now, just compare) and then compare our self increasing sequenceId.

There are still many codes that need to be modified as a whole, and this is just the first part.

First step of https://github.com/apache/incubator-paimon/pull/2811


yuzelin (approved):
+1

PR #2939, opened 2024-03-03
[Core]add table schema cache for SchemaManager

wg1026688210:
### Purpose
When reading a split ,recordReader will loads the schema of the split from the FileSystem.
The pr is for adding the cache of TableSchema  for SchemaManager to reduce the access of FileSystem.

### Tests
SchemaManagerTest#testCache

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/schema/SchemaManager.java
+++ b/paimon-core/src/main/java/org/apache/paimon/schema/SchemaManager.java
@@ -47,6 +47,7 @@ import org.apache.paimon.utils.Preconditions;
 
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.ThreadSafe;
+
 import java.io.IOException;
 import java.io.Serializable;
 import java.io.UncheckedIOException;
@@ -90,7 +91,7 @@ public class SchemaManager implements Serializable {
         this.fileIO = fileIO;
         this.tableRoot = tableRoot;
         this.cache = cache;
-        //cache
+        // cache
         listAll();
     }
 
--- a/paimon-core/src/main/java/org/apache/paimon/schema/SchemaManager.java
+++ b/paimon-core/src/main/java/org/apache/paimon/schema/SchemaManager.java
@@ -84,15 +84,9 @@ public class SchemaManager implements Serializable {
     private final Map<Long, TableSchema> cache;
 
     public SchemaManager(FileIO fileIO, Path tableRoot) {
-        this(fileIO, tableRoot, new ConcurrentHashMap<>());
-    }
-
-    public SchemaManager(FileIO fileIO, Path tableRoot, Map<Long, TableSchema> cache) {
         this.fileIO = fileIO;
         this.tableRoot = tableRoot;
-        this.cache = cache;
-        // cache
-        listAll();
+        this.cache = new ConcurrentHashMap<>();
     }
 
     public SchemaManager withLock(@Nullable Lock lock) {
@@ -132,11 +126,6 @@ public class SchemaManager implements Serializable {
         } catch (IOException e) {
             throw new UncheckedIOException(e);
         }
-    }
-
-    @VisibleForTesting
-    public Map<Long, TableSchema> getCachedSchema() {
-        return cache;
     }
 
     /** Create a new schema from {@link Schema}. */
@@ -378,9 +367,7 @@ public class SchemaManager implements Serializable {
             try {
                 boolean success = commit(newSchema);
                 if (success) {
-                    if (cache != null) {
-                        cache.put(newSchema.id(), newSchema);
-                    }
+                    cache.put(newSchema.id(), newSchema);
                     return newSchema;
                 }
             } catch (Exception e) {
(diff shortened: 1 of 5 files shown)


paimon-core/src/main/java/org/apache/paimon/schema/SchemaCache.java
```diff
@@ -0,0 +1,73 @@
...
+import java.io.Serializable;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Function;
+
+/**
+ * The class is responsible for providing a schemaManager with a concurrent and serializable schema
+ * cache.
+ */
+public class SchemaCache implements Serializable {
```
JingsongLi, line 38: You don't need to introduce this class, just merge this into `SchemaManager`.

paimon-core/src/main/java/org/apache/paimon/schema/SchemaCache.java
```diff
@@ -0,0 +1,73 @@
...
+    public TableSchema getSchema(Long id, Function<Long, TableSchema> schemaLoader) {
+        return cache.get(id, schemaLoader);
+    }
+
+    public void deleteSchema(Long id) {
+        cache.invalidate(id);
+    }
+
+    private void writeObject(ObjectOutputStream out) throws IOException {
+        Map<Long, TableSchema> map = new HashMap<>(cache.asMap());
```
JingsongLi, line 58: I think we don't need serialize this cache.

paimon-core/src/main/java/org/apache/paimon/table/AbstractFileStoreTable.java
```diff
@@ -97,6 +102,17 @@ protected AbstractFileStoreTable(
...
         this.tableSchema = tableSchema;
         this.catalogEnvironment = catalogEnvironment;
+        schemaCache = loadSchemaCache(fileIO, path);
+    }
+
+    @NotNull
+    private Map<Long, TableSchema> loadSchemaCache(FileIO fileIO, Path path) {
+        Map<Long, TableSchema> schemaCache = new ConcurrentHashMap<>();
+        SchemaManager schemaManager = new SchemaManager(fileIO, path);
+        for (TableSchema schema : schemaManager.listAll()) {
```
JingsongLi, line 112: Why we need to list all at first?

paimon-core/src/main/java/org/apache/paimon/table/AbstractFileStoreTable.java
```diff
@@ -82,6 +85,8 @@ abstract class AbstractFileStoreTable implements FileStoreTable {
     protected final TableSchema tableSchema;
     protected final CatalogEnvironment catalogEnvironment;
 
+    protected final Map<Long, TableSchema> schemaCache;
```
JingsongLi, line 88: or just store `SchemaManager` here?

paimon-core/src/main/java/org/apache/paimon/schema/SchemaManager.java
```diff
@@ -80,9 +80,16 @@ public class SchemaManager implements Serializable {
 
     @Nullable private transient Lock lock;
 
+    private final Map<Long, TableSchema> cache;
+
     public SchemaManager(FileIO fileIO, Path tableRoot) {
+        this(fileIO, tableRoot, null);
```
JingsongLi, line 86: new hash map? just keep cache not null?

paimon-core/src/main/java/org/apache/paimon/table/AbstractFileStoreTable.java
```diff
@@ -97,6 +100,7 @@ protected AbstractFileStoreTable(
         }
         this.tableSchema = tableSchema;
         this.catalogEnvironment = catalogEnvironment;
+        tableSchemaManager = new SchemaManager(fileIO, path, new ConcurrentHashMap<>());
```
JingsongLi, line 103: Just use `public SchemaManager(FileIO fileIO, Path tableRoot)`

paimon-core/src/test/java/org/apache/paimon/schema/SchemaManagerTest.java
```diff
@@ -94,6 +95,27 @@ public void afterEach() {
...
+                new SchemaManager(FileIOFinder.find(path), path, new ConcurrentHashMap<>());
+        // schema-0
+        manager.createTable(schema);
+        // schema-1
+        manager.commitChanges(SchemaChange.setOption("aaa", "bbb"));
+        // schema-2
+        manager.commitChanges(SchemaChange.setOption("ccc", "ddd"));
+
+        Map<Long, TableSchema> cachedSchema = manager.getCachedSchema();
+        assertThat(cachedSchema).hasSize(3);
```
JingsongLi, line 111: You don't need to `getCachedSchema`.
You can just verify same instances.

JingsongLi:
See https://github.com/apache/incubator-paimon/pull/3021

JingsongLi:
At present, this solution is relatively obscure. Generally speaking, it is difficult to accept a solution where the cache is serialized and reused by distributed tasks.

Considering that https://github.com/apache/incubator-paimon/pull/3021 has already been merged and can solve most problems in most cases (without schema changes), I am considering closing this PR.

You can reopen this PR at any time if you have any further needs.

PR #2940, opened 2024-03-04
[core][bug] Write Spillable with memory preemption has bug : preempt memory will cause writer flush memory to disk which should not happen

leaves12138:
Fix bug when open spillable, still many small files generated.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/append/AppendOnlyWriter.java
+++ b/paimon-core/src/main/java/org/apache/paimon/append/AppendOnlyWriter.java
@@ -176,8 +176,8 @@ public class AppendOnlyWriter implements RecordWriter<InternalRow>, MemoryOwner 
         return compactManager.isCompacting();
     }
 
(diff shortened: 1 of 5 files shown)


paimon-core/src/main/java/org/apache/paimon/append/AppendOnlyWriter.java
```diff
@@ -279,7 +279,11 @@ public long memoryOccupancy() {
 
     @Override
     public void flushMemory() throws Exception {
-        flush(false, false);
+        if (sinkWriter.bufferSpillableWriter()) {
```
JingsongLi, line 282: just same to `MergeTreeWriter`?

leaves12138: `MergeTreeWriter` seems don't have this bug.

JingsongLi:
+1

Issue #1773, opened 2023-08-09
[Feature] Support count on files and manifest list name show on snapshots table


schnappi17:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/incubator-paimon/issues) and found nothing similar.

### Motivation

Support count on files and manifest list name show on snapshots table

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

JingsongLi:
This change will make the snapshots table completely unusable:
1. Firstly, it will be very slow, and each snapshot will scan and read all the files.
2. Secondly, this scan does not contain filters for partitions, which is highly likely to cause OOM.

JingsongLi:
CC @schnappi17 @FangYongs

JingsongLi:
And please do not add fields in the middle of the schema as this will compromise previous compatibility.

PR #2941, opened 2024-03-04
[core] Revert added_file_count and delete_file_count in SnapshotsTable

JingsongLi:
### Purpose

Revert part of #1773


FangYongs:
Thanks @JingsongLi , +1, wait for the CI

PR #2942, opened 2024-03-05
[core] Introduce ExpireChangelogImpl to handle the changelog lifecycle

### LINKED ISSUE #2899: [Feature] Decouple the lifecycle of snapshot and changelog
### Search before asking

### Motivation

# Motivation

Currently, the changelog's lifecycle is binding with the snapshot, when snapshot is expired, the changelog is expired. If we want to keep longer changelog for the streaming job to reset consume, we also have to keep the data files in snapshot, this is not necessary, and will waste much space resource.

In traditional database, the binlog's lifecycle is individual. So, In paimon, we could also support to decouple the lifecycle of snapshot and changelog . In this way, users can choose the flexible option to keep the changelog data. eg:  user can keep one latest snapshot and one day's changelog for consuming.

# How to implement

The lifecyle is handled in expiration process. To keep two different lifecycle, we need two group file point to mark the effective changelog and snapshot

- `EARLIEST` and `LATEST` mark the effective datafile
- `EARLIEST_CHANGELOG` and `LATEST` mark the effective changlog 

> The `LATEST` should be same, and the `EARLIEST_CHANGELOG` only present when the changelog is enabled.

image

However, the hint file is not always there. We should have a way to work when the hint file is missing. So we introduce a mark file `expire-snapshot-x` and `expire-changelog-x` when expire the corresponding object. We create the expire file first, and when the two file are created, we can delete the snapshot metadata. After that, the mark file are also deleted. The `EARLIEST` and `EARLIEST_CHANGELOG` always updated after the file created. So, if the `EARLIEST_` file is missing or inaccurate, we can determine the current effective snapshot and changelog by the `expire-snapshot-x` and `expire-changelog-x` file.

# New config 

- `changelog.time-retained`
- `changelog.num-retained.min`
- `changelog.num-retained.max`

The counterparts of the snapshot retained configs.

### Subtask

- [x] Introduce ExpireChangelogImpl handle the changelog expire
- [x] Integrate the InnerStreamScan with the changlog metadata
- [x] Handle the orphan file cleaner with the changelog metadata
- [x] Support decouple the delta files lifecycle
- [ ] Support to merge the small changelog files

### Anything else?

_No response_

Aitozi:
### Purpose

Linked issue: close #2899

In this PR, the `snapshot` will be moved to the `/changelog` directory for decouple the changelog lifecycle. We will always keep one extra metadata file in the `/changelog` directory for the conveniency of streaming scan.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/ExpireSnapshotsImpl.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/ExpireSnapshotsImpl.java
@@ -18,6 +18,7 @@
 
 package org.apache.paimon.table;
 
+import org.apache.paimon.Changelog;
 import org.apache.paimon.Snapshot;
 import org.apache.paimon.annotation.VisibleForTesting;
 import org.apache.paimon.consumer.ConsumerManager;
@@ -45,6 +46,8 @@ public class ExpireSnapshotsImpl implements ExpireSnapshots {
     private final SnapshotDeletion snapshotDeletion;
     private final TagManager tagManager;
     private final boolean cleanEmptyDirectories;
+    /** Whether to clean the changelog or delta files. */
+    private final boolean changelogDecoupled;
 
     private int retainMax = Integer.MAX_VALUE;
     private int retainMin = 1;
@@ -55,13 +58,15 @@ public class ExpireSnapshotsImpl implements ExpireSnapshots {
             SnapshotManager snapshotManager,
             SnapshotDeletion snapshotDeletion,
             TagManager tagManager,
-            boolean cleanEmptyDirectories) {
+            boolean cleanEmptyDirectories,
+            boolean changelogDecoupled) {
         this.snapshotManager = snapshotManager;
         this.consumerManager =
                 new ConsumerManager(snapshotManager.fileIO(), snapshotManager.tablePath());
         this.snapshotDeletion = snapshotDeletion;
         this.tagManager = tagManager;
         this.cleanEmptyDirectories = cleanEmptyDirectories;
+        this.changelogDecoupled = changelogDecoupled;
     }
 
     @Override
@@ -183,7 +188,17 @@ public class ExpireSnapshotsImpl implements ExpireSnapshots {
                 continue;
             }
 
-            snapshotDeletion.cleanUnusedDataFiles(snapshot, skipper);
+            if (changelogDecoupled) {
+                if (snapshot.changelogManifestList() != null
+                        || snapshot.commitKind() != Snapshot.CommitKind.APPEND) {
+                    snapshotDeletion.cleanUnusedDataFiles(snapshot, skipper);
+                }
+            } else {
+                snapshotDeletion.cleanUnusedDataFiles(snapshot, skipper);
+                if (snapshot.changelogManifestList() != null) {
+                    snapshotDeletion.deleteAddedDataFiles(snapshot.changelogManifestList());
+                }
+            }
         }
 
         // data files and changelog files in bucket directories has been deleted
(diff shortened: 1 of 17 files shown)


paimon-core/src/main/java/org/apache/paimon/table/ExpireSnapshotsImpl.java
```diff
@@ -215,10 +204,15 @@ public int expireUntil(long earliestId, long endExclusiveId) {
...
             Snapshot snapshot = snapshotManager.snapshot(id);
-            snapshotDeletion.cleanUnusedManifests(snapshot, skippingSet);
+            snapshotDeletion.cleanUnusedManifests(snapshot, skippingSet, false);
 
-            // delete snapshot last
-            snapshotManager.fileIO().deleteQuietly(snapshotManager.snapshotPath(id));
+            // move snapshot metadata to changelog
+            try {
+                snapshotManager.moveToChangelog(id);
+                snapshotManager.commitLongLivedChangelogLatestHint(id);
```
JingsongLi, line 212: Maybe we should not just move snapshot file.

We could introduce a new JSON file for changelog.

paimon-core/src/main/java/org/apache/paimon/table/ExpireSnapshotsImpl.java
```diff
@@ -215,10 +204,15 @@ public int expireUntil(long earliestId, long endExclusiveId) {
...
 
             Snapshot snapshot = snapshotManager.snapshot(id);
-            snapshotDeletion.cleanUnusedManifests(snapshot, skippingSet);
+            snapshotDeletion.cleanUnusedManifests(snapshot, skippingSet, false);
 
-            // delete snapshot last
-            snapshotManager.fileIO().deleteQuietly(snapshotManager.snapshotPath(id));
+            // move snapshot metadata to changelog
+            try {
+                snapshotManager.moveToChangelog(id);
```
JingsongLi, line 211: Please note that it is best not to affect the original behavior. For example, here we should determine whether there is a separate configuration for the changelog lifecycle. If it is not configured or is the same as the snapshot lifecycle, we should avoid creating files in the changelog folder.

paimon-core/src/main/java/org/apache/paimon/table/ExpireSnapshotsImpl.java
```diff
@@ -186,17 +186,6 @@ public int expireUntil(long earliestId, long endExclusiveId) {
             snapshotDeletion.cleanUnusedDataFiles(snapshot, skipper);
         }
 
-        // delete changelog files
-        for (long id = beginInclusiveId; id < endExclusiveId; id++) {
-            if (LOG.isDebugEnabled()) {
-                LOG.debug("Ready to delete changelog files from snapshot #" + id);
-            }
-            Snapshot snapshot = snapshotManager.snapshot(id);
-            if (snapshot.changelogManifestList() != null) {
```
JingsongLi, line 195: If possible, we should try to ensure that the delta file of the Append snapshot can be stream read when there is no changelog file present.

JingsongLi (changes requested):

paimon-core/src/main/java/org/apache/paimon/table/ExpireSnapshotsImpl.java
Aitozi: Currently, I've not seen the requirement to introduce a new metadata structure for changelog. We can rewrite the content later when we have to later.

paimon-core/src/main/java/org/apache/paimon/table/ExpireSnapshotsImpl.java
Aitozi: Agree, will give a try.

paimon-core/src/main/java/org/apache/paimon/table/ExpireSnapshotsImpl.java
JingsongLi: If we have to do it later, it is hard to keep forward compatibility.

It is better to provide a simple version now.

PR #2945, opened 2024-03-05
[core] Redefine user defined sequence fields

### LINKED ISSUE #2471: [Bug] spark deletes dynamic bucket table, the result is incorrect

JingsongLi:
### Purpose

Linked PR: #2811

Next step of #2936

Fix #2471

Redefine user defined sequence fields to like: https://github.com/facebook/rocksdb/wiki/User-defined-Timestamp

Completely separate User Defined Sequence Fields from self increasing sequenceId.

This way, there will be no more accuracy issues. When comparing, we can first compare the user-defined sequenceField (We don't need to extract long now, just compare) and then compare our self increasing sequenceId.

### API and Format

Since we have resolved the issue of the same User Defined Sequence Fields value, there is no need for us to introduce sequence Padding.

So this PR, will remove all sequence Padding.

Why not keep `row-kind-flag`? Because this option will make `DELETE XX FROM T WHERE pk = xxx` not work, delete records will not work.

### Documentation

Adjusted.


hzjhjjyy:
+1
`row-kind-flag` is a useful configuration, but it seems redundant with the capability of ignoring deleted records in the merge function. At this lowest level of the sorting stage, it feels better to let the data flow in its original form, and leave the data processing to the merge function.

Zouxxyy (approved):
+1

PR #2946, opened 2024-03-05
[Hive] Cache TableSchema into Configuration to avoid loading read scheme file in PaimonSerDe

wg1026688210:
### Purpose
The Map Task will fetch  the tableschema  of paimon. we cache the schema json to configuration to reduce the access of FileSystem

Linked issue: close #xxx

### DIFF
--- a/paimon-hive/paimon-hive-connector-common/src/main/java/org/apache/paimon/hive/HiveSchema.java
+++ b/paimon-hive/paimon-hive-connector-common/src/main/java/org/apache/paimon/hive/HiveSchema.java
@@ -19,6 +19,7 @@
 package org.apache.paimon.hive;
 
 import org.apache.paimon.CoreOptions;
+import org.apache.paimon.annotation.VisibleForTesting;
 import org.apache.paimon.catalog.CatalogContext;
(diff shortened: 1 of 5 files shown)


paimon-hive/paimon-hive-connector-common/src/main/java/org/apache/paimon/hive/HiveSchema.java
```diff
@@ -192,7 +198,16 @@ public static HiveSchema extract(@Nullable Configuration configuration, Properti
         return new HiveSchema(builder.build());
     }
 
-    private static Optional<TableSchema> getExistingSchema(
+    @VisibleForTesting
+    static Optional<TableSchema> getTableSchemaFromCache(Properties properties) {
```
JingsongLi, line 202: Why not just cache `HiveSchema`?

JingsongLi, line 202: `getSchemaFromProperties`

JingsongLi (approved):
+1

PR #2948, opened 2024-03-06
[core] Exception should be thrown after 'catch' in `AppendOnlyWriter`.

leaves12138:
An small bug: Exception have been swallowed if finally throws an exception too.


JingsongLi:
+1

PR #2949, opened 2024-03-06
[ci] Fix paimon flink tests

yuzelin:
### Purpose

Fix that paimon-flink-1.1x tests cannot be touched now (some tests are error in 1.1x module but not be tested)

And replace hadoop-common dep to paimon-flink pom because the tests of paimon-flink-1.1x need it.


JingsongLi:
+1

Issue #2861, opened 2024-02-07
[Feature] Support Flink read / write data branch


schnappi17:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/incubator-paimon/issues) and found nothing similar.

### Motivation

Subtask of #1795

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

ShockleysxX:
Hi, @schnappi17 I am learning pamion recently and I am very interested in this project, I would like to try this if you are willing to assign it to me.

sunxiaojian:
> sorry for busy, you can take it now~

@ShockleysxX ok

schnappi17:
@sunxiaojian Could you help to review the pr https://github.com/apache/incubator-paimon/pull/2950 that fix the bug of compaction when writing on custom branch, and your pr could rebase on this.

sunxiaojian:
> @sunxiaojian Could you help to review the pr #2950 that fix the bug of compaction when writing on custom branch, and your pr could rebase on this.

Ok, I will handle it as soon as possible.

PR #2950, opened 2024-03-06
[core] Fix collecting wrong sorted runs when compacting on custom branch

schnappi17:
### Purpose

Linked issue: subtask #2861 

When writing to a custom branch, the compaction within writer will picking the wrong sorted runs on the 'main' branch to compact, which will cause the commit conflict, like willing to delete files which never been added.


paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreWrite.java
sunxiaojian, line 1: @schnappi17  Does prepareCommit() also need to be processed?

schnappi17: @sunxiaojian You mean the calculation of `latestCommittedIdentifier`?

sunxiaojian: > You mean the calculation of `latestCommittedIdentifier`?

@schnappi17 yes

sunxiaojian:
@schnappi17 If https://github.com/apache/incubator-paimon/pull/3029 Merged, do we no longer need the current processing method ?

schnappi17:
Fixed in https://github.com/apache/paimon/pull/3029, close this.

PR #2951, opened 2024-03-06
[flink] Postpone sink materializer check from compile time to runtime

tsreaper:
### Purpose

Currently we check for sink materializer and throws exception in SQL compile time. However, some cloud computing flatform (for example, aliyun VVP) allows user to change some Flink configurations after compiling SQL but before starting the job.

This PR postpone sink materializer check from compile time to runtime to work around such case.

### Tests

### API and Format

No.

### Documentation

No,


JingsongLi:
+1

PR #2952, opened 2024-03-06
[core]Verify that the number of partition fields should not be the same as the field

herefree:
### Purpose
Verify that the number of partition fields should not be the same as the field，When I create an append-only table and set the number of partitions to be consistent with the number of fields, the data under each partition is the same.

### Tests

org.apache.paimon.schema.TableSchemaTest#testValidatePartitionKeys


herefree:
@JingsongLi  cc

PR #2953, opened 2024-03-06
[core] Redefine the usage of RecordWithPositionIterator

Zouxxyy:
### Purpose

before: rowPosition(), next()
after: next(), rowPositon()

Only in this way can we correctly apply filter to it

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/data/columnar/ColumnarRowIterator.java
+++ b/paimon-common/src/main/java/org/apache/paimon/data/columnar/ColumnarRowIterator.java
@@ -38,8 +38,8 @@ public class ColumnarRowIterator extends RecyclableIterator<InternalRow>
     private final Runnable recycler;
 
     private int num;
-    private int pos;
-    private long rowPosition;
+    private int nextPos;
+    private long nextGlobalPos;
 
     public ColumnarRowIterator(ColumnarRow rowData, @Nullable Runnable recycler) {
         super(recycler);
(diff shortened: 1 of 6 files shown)


paimon-common/src/main/java/org/apache/paimon/reader/RecordWithPositionIterator.java
```diff
@@ -33,8 +33,7 @@
 public interface RecordWithPositionIterator<T> extends RecordReader.RecordIterator<T> {
 
     /**
-     * Get the row position of the row that will be returned by the following call to {@link
-     * RecordReader.RecordIterator#next}.
+     * Get the row position of the row called by {@link RecordReader.RecordIterator#next}.
      *
      * @return the row position from 0 to the number of rows in the file
      */
```
JingsongLi, line 39: rename to `returnedPosition`?

paimon-format/src/main/java/org/apache/paimon/format/orc/OrcReaderFactory.java
```diff
@@ -183,7 +183,7 @@ private RecordIterator<InternalRow> convertAndGetIterator(
             int batchSize = orcBatch.size;
             paimonColumnBatch.setNumRows(batchSize);
             result.reset(batchSize);
-            result.resetRowPosition(rowNumber);
```
JingsongLi, line 186: Just keep it as it is.
We can implement `ColumnarRowIterator.returnedPosition` to `nextGlobalPosition - 1`?

paimon-common/src/main/java/org/apache/paimon/data/columnar/ColumnarRowIterator.java
```diff
@@ -45,7 +45,6 @@ public ColumnarRowIterator(ColumnarRow rowData, @Nullable Runnable recycler) {
         super(recycler);
         this.rowData = rowData;
         this.recycler = recycler;
-        this.rowPosition = 0;
```
JingsongLi, line 48: currently, we have `pos` and `rowPosition`.

Can we rename them to `nextPos` and `nextGlobalPosition`?

paimon-common/src/main/java/org/apache/paimon/data/columnar/ColumnarRowIterator.java
```diff
@@ -90,9 +96,7 @@ public ColumnarRowIterator mapping(
             if (indexMapping != null) {
                 vectors = VectorMappingUtils.createIndexMappedVectors(indexMapping, vectors);
             }
-            ColumnarRowIterator iterator = new ColumnarRowIterator(rowData.copy(vectors), recycler);
-            iterator.reset(num);
```
JingsongLi, line 94: Can we merge these two reset methods? Introduce only one reset?

JingsongLi (changes requested):

JingsongLi (approved):
+1

PR #2958, opened 2024-03-07
[core] Integrate deletion vector to reader and writer

Zouxxyy:
### Purpose

subTask of #2898, in this PR:

- add a new conf `deletion-vectors.enabled` to control whether deletion vector mode is enabled. Limit:
  - pk table
  - `file.format` = `orc` or `parquet`
  - `changelog-producer` = `none` or `lookup`
  - `merge-engine` != `first-row` (implement it in the future)
  
- integrate deletion vector to reader and writer, when `deletion-vectors.enabled` is true:
  - for writer: update dv index when lookup, since drop delete is necessary, rewrite is always required in lookup compaction
  - for reader: introduce `ApplyDeletionVectorReader` for applying DeletionVector to filter record

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkSink.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkSink.java
@@ -54,6 +54,7 @@ import java.util.Set;
 import java.util.UUID;
 
 import static org.apache.flink.configuration.ClusterOptions.ENABLE_FINE_GRAINED_RESOURCE_MANAGEMENT;
+import static org.apache.paimon.CoreOptions.DELETION_VECTORS_ENABLED;
 import static org.apache.paimon.CoreOptions.FULL_COMPACTION_DELTA_COMMITS;
 import static org.apache.paimon.flink.FlinkConnectorOptions.CHANGELOG_PRODUCER_FULL_COMPACTION_TRIGGER_INTERVAL;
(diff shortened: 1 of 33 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkSink.java
```diff
@@ -102,8 +103,9 @@ private StoreSinkWrite.Provider createWriteProvider(
             Options options = table.coreOptions().toConfiguration();
             ChangelogProducer changelogProducer = table.coreOptions().changelogProducer();
             waitCompaction =
-                    changelogProducer == ChangelogProducer.LOOKUP
-                            && options.get(CHANGELOG_PRODUCER_LOOKUP_WAIT);
+                    (changelogProducer == ChangelogProducer.LOOKUP
+                                    && options.get(CHANGELOG_PRODUCER_LOOKUP_WAIT))
+                            || options.get(DELETION_VECTORS_ENABLED);
```
JingsongLi, line 108: Can you add itcase for Flink?

JingsongLi, line 108: If `CHANGELOG_PRODUCER_LOOKUP_WAIT` is true, can we just not wait?

Zouxxyy: Will support deletion vectors with lookup wait in the another PR

JingsongLi (approved):
+1

PR #2959, opened 2024-03-07
[spark] use the internal session catalog in spark generic catalog

YannByron:
### Purpose

Linked issue: close #xxx


paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/SparkConnectorOptions.java
```diff
@@ -24,6 +24,13 @@
 
 /** Options for spark connector. */
 public class SparkConnectorOptions {
+
+    public static final ConfigOption<Boolean> INNER_SESSION_CATALOG =
+            key("catalog.inner-session-catalog")
```
JingsongLi, line 29: `delegate-built-in-session-catalog`?
Default is true.

YannByron: `catalog.delegate-built-in-session-catalog`: true means use the original spark session catalog.

JingsongLi:
+1

PR #2961, opened 2024-03-07
[WIP][core] Fix LookupChangelogMergeFunctionWrapper handle retract message incorrectly

yuzelin:
### Purpose

Linked issue: close #xxx


paimon-core/src/main/java/org/apache/paimon/mergetree/compact/LookupChangelogMergeFunctionWrapper.java
```diff
@@ -85,6 +86,38 @@ public void add(KeyValue kv) {
     public ChangelogResult getResult() {
         reusedResult.reset();
 
+        if (mergeFunction.candidatesAllLevel0()) {
```
JingsongLi, line 89: Means there is no highlevel in mergeFunction?

It seems we should merge these two methods.

JingsongLi (approved):
+1

PR #2964, opened 2024-03-07
[core] New parameters `commit.force-create-snapshot` Used to force a snapshot to be generated when committing

### LINKED ISSUE #2957: [Feature] Make `forceCreatingSnapshot` a toggle option for paimon
### Search before asking

### Motivation

PaimonSink Commit has a logic that if no data is written during a checkpoint cycle, no snapshot will be generated.
Let me give you an example:

If I read a kafka stream and write it to Paimon, there may be filters in the middle, and flink watermarks are generated at the same time. Then each snapshot of Paimon will have watermark information. I can judge my writing progress through the watermark of the latest snapshot. , if the watermark is advancing in a checkpoint interval, but the filter logic filters out the data, a snapshot will not be generated, but the watermark should be advancing.

Currently, forced snapshot generation only takes effect when the automatic tag creation configuration is enabled and certain conditions are met. I think we should make this a switchable option.

### Solution

Make `forceCreatingSnapshot` a toggle option for paimon

### Anything else?

_No response_

huyuanfeng2018:
### Purpose

Linked issue: close #2957

### Tests

`org.apache.paimon.flink.sink.CommitterOperatorTest#testForceCreateSnapshotCommit`

### DIFF
--- a/docs/layouts/shortcodes/generated/core_configuration.html
+++ b/docs/layouts/shortcodes/generated/core_configuration.html
@@ -81,6 +81,12 @@ under the License.
             <td>Whether to force a compaction before commit.</td>
         </tr>
         <tr>
+            <td><h5>commit.force-create-snapshot</h5></td>
+            <td style="word-wrap: break-word;">false</td>
+            <td>Boolean</td>
+            <td>Whether to force create snapshot on commit.</td>
+        </tr>
+        <tr>
             <td><h5>compaction.max-size-amplification-percent</h5></td>
             <td style="word-wrap: break-word;">200</td>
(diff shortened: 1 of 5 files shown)


huyuanfeng2018:
@JingsongLi PTAL. 

thanks.

JingsongLi:
@huyuanfeng2018  Take a look to `paimon-docs/README`.

huyuanfeng2018:
> @huyuanfeng2018 Take a look to `paimon-docs/README`.

done.

JingsongLi (approved):
+1 Thanks @huyuanfeng2018

PR #2968, opened 2024-03-07
[core] add commit.force-expire option

### LINKED ISSUE #2967: [Feature] support an option to force expiration.
### Search before asking

### Motivation

Sort compact only supports write-only mode, during which expiration operations are not performed. When users write data in write-only mode and use sort compact for sorting and merging, expired deletion cannot be executed.

### Solution

Add an option to force expiration.

### Anything else?

_No response_

zhoulii:
### Purpose

Linked issue: close #2967 

Sort compact only supports write-only mode, during which expiration operations are not performed. When users write data in write-only mode and use sort compact for sorting and merging, expired deletion cannot be executed. This PR add an option to force expiration.

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
@@ -378,6 +378,12 @@ public class CoreOptions implements Serializable {
                     .defaultValue(MemorySize.ofMebiBytes(128))
                     .withDescription("Target size of a file.");
 
+    public static final ConfigOption<Boolean> FORCE_EXPIRE =
+            key("commit.force-expire")
+                    .booleanType()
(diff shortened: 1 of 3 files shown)


paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -378,6 +378,12 @@ public class CoreOptions implements Serializable {
                     .defaultValue(MemorySize.ofMebiBytes(128))
                     .withDescription("Target size of a file.");
 
+    public static final ConfigOption<Boolean> FORCE_EXPIRE =
```
JingsongLi, line 381: Maybe it is better to introduce an option: write.responsible-matters, similar option, I'm not good at naming it, for reference only.

In the option, user can specific what the write does.

PR #2970, opened 2024-03-07
[flink] ignore expiration check interval of partition in batch mode.

### LINKED ISSUE #2969: [Feature] ignore expiration check interval of partition in batch mode.
### Search before asking

### Motivation

If the execution time of a batch job is short,  and it's less than the interval for partition expiration checks, the partition expiration operations will not be executed.

### Solution

During the commit phase of batch jobs, ignore expiration check interval of partition.

### Anything else?

_No response_

zhoulii:
### Purpose

Linked issue: close #2969

If the execution time of a batch job is short, and it's less than the interval for partition expiration checks, the partition expiration operations will not be executed. This PR aims to ignore partition expiration check interval for batch job.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/PartitionExpire.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/PartitionExpire.java
@@ -28,6 +28,9 @@ import org.apache.paimon.partition.PartitionTimeExtractor;
 import org.apache.paimon.types.RowType;
 import org.apache.paimon.utils.RowDataToObjectArrayConverter;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.time.Duration;
 import java.time.LocalDateTime;
 import java.util.ArrayList;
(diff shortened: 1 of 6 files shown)


zhoulii:
Hi @Aitozi , could you please take a look at this PR? Thanks.

paimon-core/src/main/java/org/apache/paimon/operation/PartitionExpire.java
```diff
@@ -117,6 +124,10 @@ private List<BinaryRow> readPartitions(LocalDateTime expireDateTime) {
                 .collect(Collectors.toList());
     }
 
+    public void ignoreExpireCheckInterval() {
```
JingsongLi, line 127: Maybe you can add `close` to this class?

zhoulii: Perhaps I didn't fully grasp your point. Are you suggesting that we should change the method name?

PR #2971, opened 2024-03-07
[core] Add cache catalog for paimon

zhangjun0x01:
### Purpose

Linked issue: close https://github.com/apache/incubator-paimon/issues/2136

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/catalog/CacheCatalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/catalog/CacheCatalog.java
@@ -1,0 +1,264 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.paimon.catalog;
+
+import org.apache.paimon.schema.Schema;
+import org.apache.paimon.schema.SchemaChange;
+import org.apache.paimon.schema.TableSchema;
+import org.apache.paimon.table.Table;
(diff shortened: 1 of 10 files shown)


zhangjun0x01:
hi , @JingsongLi  could you help me review ? thanks

paimon-core/src/main/java/org/apache/paimon/catalog/CacheCatalog.java
```diff
@@ -0,0 +1,264 @@
...
+    @Override
+    public Table getTable(Identifier identifier) throws TableNotExistException {
+        Table cached = tableCache.getIfPresent(identifier);
+        if (cached != null) {
+            return cached;
+        }
+
+        Table table = catalog.getTable(identifier);
+        tableCache.put(identifier, table);
+        return table;
```
wg1026688210, line 97:
```
tableCache#get(@NonNull K var1, @NonNull Function<? super K, ? extends V> var2)
```
will be more clear

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkCatalog.java
```diff
@@ -144,7 +147,13 @@ public FlinkCatalog(
             ClassLoader classLoader,
             Options options) {
         super(name, defaultDatabase);
-        this.catalog = catalog;
+        boolean cacheEnabled = options.get(CACHE_CATALOG);
+        long cacheExpirationIntervalMs = options.get(CACHE_CATALOG_EXPIRATION_MS);
+        if (cacheEnabled) {
```
wg1026688210, line 152: Shall we add this in CatalogFactory

wg1026688210 (commented):
Leave some minor comments

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkCatalog.java
zhangjun0x01: If the catalog is only created once, this code can be moved to `CatalogFactory`. However, in some test case of the `flink-cdc` module, they use the same `warehouse` to create different catalogs. If a cache catalog is added, it will result in assert failure.

paimon-core/src/main/java/org/apache/paimon/catalog/CacheCatalog.java
zhangjun0x01: yes, I update it

JingsongLi:
Hi @zhangjun0x01 , considering that this PR has not been active for some time, I plan to redo it. Unlike this PR:
1. This PR also seems to be based on Iceberg's configuration, so we should follow the same configuration as Iceberg.
2. The cachingCatalog should be created in catalogFactory, not Flink or Spark
3. CachingCatalog should be a catalog, not an abstract catalog

PR #2972, opened 2024-03-07
[Feature] Manage main branch for paimon

sunxiaojian:
### Purpose

1. CALL sys.set_main_branch('%s.%s', 'branch_name') : Only switch the main branch without performing file deletion and migration operations
2. CALL sys.clean_main_branch('%s.%s') :  Clear the main branch configuration file and use the default main branch


sunxiaojian:
@FangYongs PTAL

PR #2976, opened 2024-03-08
[Core]Support compiled class cache in CodeGenUtils

### LINKED ISSUE #2955: [Feature] Support compiled class cache in CodeGenUtils
### Search before asking

### Motivation

screenshot-20240306-171507

When we have a table with thousands of columns, the operation to generate a RecordEqualiser class will consume lots of resources. Especially when we are creating multiple Readers from multiple Table instances at the same time, this is a real case when reading the paimon table using Starrocks as external tables. By reusing the generated class, we can avoid this problem.

### Solution

_No response_

### Anything else?

_No response_

xiangyuf:
### Purpose

Linked issue: close #2955 

Reuse the generated class in CodeGeneratorImpl to improve performance.

### Tests

org.apache.paimon.codegen.CodeGeneratorCacheTest

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/codegen/CodeGenUtils.java
+++ b/paimon-core/src/main/java/org/apache/paimon/codegen/CodeGenUtils.java
@@ -21,12 +21,28 @@ package org.apache.paimon.codegen;
 import org.apache.paimon.data.BinaryRow;
 import org.apache.paimon.types.DataType;
 import org.apache.paimon.types.RowType;
+import org.apache.paimon.utils.Pair;
 
+import org.apache.paimon.shade.guava30.com.google.common.cache.Cache;
+import org.apache.paimon.shade.guava30.com.google.common.cache.CacheBuilder;
+
+import java.time.Duration;
+import java.util.Arrays;
 import java.util.List;
+import java.util.Objects;
 import java.util.stream.IntStream;
 
 /** Utils for code generations. */
 public class CodeGenUtils {
+
+    static final Cache<ClassKey, Pair<Class<?>, Object[]>> COMPILED_CLASS_CACHE =
+            CacheBuilder.newBuilder()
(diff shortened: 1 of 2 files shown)


xiangyuf:
Hi @JingsongLi  , PTAL when you have time. Thx.

xiangyuf:
@FangYongs @liming30 Hi, would u kindly help review this pr?

paimon-codegen/src/main/java/org/apache/paimon/codegen/CodeGeneratorImpl.java
```diff
@@ -68,4 +148,40 @@ private SortSpec getAscendingSortSpec(int[] sortFields) {
         }
         return builder.build();
     }
+
+    /** Class to use as key for the {@link #GENERATED_CLASS_CACHE}. */
+    static class ClassKey {
+        private final String name;
```
JingsongLi, line 154: Should contains interface class, for example, `RecordEqualiser.class`.

Please add test for this.

JingsongLi (changes requested):
Can you just add cache to `CodeGenUtils`? In that way, we can cache generated Class.

xiangyuf:
> Can you just add cache to `CodeGenUtils`? In that way, we can cache generated Class.

@JingsongLi Do you mean cache the instance directly?

JingsongLi:
> > Can you just add cache to `CodeGenUtils`? In that way, we can cache generated Class.
> 
> @JingsongLi Do you mean cache the instance directly?

I mea compiled `Class<T>`.

JingsongLi:
Please fix compile

JingsongLi:
merged in 4ec5806428a8d3ddbb0ba8521b8971f0e9de7b17

Looks like your email (fengxiangyu.fxy@bytedance.com) is not bind to your github id.

PR #2979, opened 2024-03-08
[hive] Drop a Hive external table only deletes metadata but not data files.

zhuangchong:
### Purpose

1.Drop a Hive external table only deletes metadata but not data files.
2.Remove uncalled methods: checkIdentifierUpperCase、schemaFileExists
3.Modify wrong words


paimon-core/src/main/java/org/apache/paimon/jdbc/JdbcCatalogOptions.java
```diff
@@ -27,7 +27,7 @@ public final class JdbcCatalogOptions {
     public static final ConfigOption<String> CATALOG_KEY =
             ConfigOptions.key("catalog-key")
                     .stringType()
-                    .defaultValue(null)
+                    .defaultValue("jdbc")
```
JingsongLi, line 30: modify documentation too

JingsongLi (approved):
+1

PR #2986, opened 2024-03-11
[core] Refactor ChangelogMergeTreeRewriter to lookup free

JingsongLi:
### Purpose

`ChangelogMergeTreeRewriter` should get rid of lookup concept.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/ChangelogMergeTreeRewriter.java
+++ b/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/ChangelogMergeTreeRewriter.java
@@ -22,12 +22,10 @@ import org.apache.paimon.CoreOptions.MergeEngine;
 import org.apache.paimon.KeyValue;
 import org.apache.paimon.compact.CompactResult;
 import org.apache.paimon.data.InternalRow;
-import org.apache.paimon.deletionvectors.DeletionVectorsMaintainer;
 import org.apache.paimon.io.DataFileMeta;
 import org.apache.paimon.io.KeyValueFileReaderFactory;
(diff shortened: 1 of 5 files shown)


paimon-core/src/main/java/org/apache/paimon/mergetree/compact/ChangelogMergeTreeRewriter.java
```diff
@@ -173,24 +174,19 @@ private CompactResult rewriteChangelogCompaction(
...
         List<DataFileMeta> before = extractFilesFromSections(sections);
         List<DataFileMeta> after =
-                rewriteCompactFile
+                compactFileWriter != null
                         ? compactFileWriter.result()
                         : before.stream()
                                 .map(x -> x.upgrade(outputLevel))
                                 .collect(Collectors.toList());
 
-        if (deletionVectorsMaintainer != null) {
```
Zouxxyy, line 182: Looks like here is forgot to add to notifyCompactBefore(before)

paimon-core/src/main/java/org/apache/paimon/mergetree/compact/LookupMergeTreeCompactRewriter.java
```diff
@@ -75,17 +75,18 @@ public LookupMergeTreeCompactRewriter(
...
-        }
+                produceChangelog,
+                dvMaintainer != null);
+        this.dvMaintainer = dvMaintainer;
         this.lookupLevels = lookupLevels;
         this.wrapperFactory = wrapperFactory;
     }
 
+    @Override
+    protected void notifyCompactBefore(List<DataFileMeta> files) {
```
Zouxxyy, line 86: this override can remove

PR #2988, opened 2024-03-11
[core] Introduce deletion files to DataSplit

JingsongLi:
### Purpose

Deletion file should be in Split, and the reader should respect deletion files in Split, instead of reading deletion manifest again which is very costly.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/KeyValueFileStoreRead.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/KeyValueFileStoreRead.java
@@ -23,7 +23,6 @@ import org.apache.paimon.KeyValue;
 import org.apache.paimon.KeyValueFileStore;
 import org.apache.paimon.data.BinaryRow;
 import org.apache.paimon.data.InternalRow;
-import org.apache.paimon.deletionvectors.ApplyDeletionVectorReader;
 import org.apache.paimon.deletionvectors.DeletionVector;
 import org.apache.paimon.disk.IOManager;
 import org.apache.paimon.fs.FileIO;
@@ -247,9 +246,11 @@ public class KeyValueFileStoreRead implements FileStoreRead<KeyValue> {
         // Sections are read by SortMergeReader, which sorts and merges records by keys.
         // So we cannot project keys or else the sorting will be incorrect.
         KeyValueFileReaderFactory overlappedSectionFactory =
-                readerFactoryBuilder.build(partition, bucket, false, filtersForKeys);
+                readerFactoryBuilder.build(
+                        partition, bucket, DeletionVector.emptyFactory(), false, filtersForKeys);
         KeyValueFileReaderFactory nonOverlappedSectionFactory =
-                readerFactoryBuilder.build(partition, bucket, false, filtersForAll);
+                readerFactoryBuilder.build(
+                        partition, bucket, DeletionVector.emptyFactory(), false, filtersForAll);
 
         List<ReaderSupplier<KeyValue>> sectionReaders = new ArrayList<>();
         MergeFunctionWrapper<KeyValue> mergeFuncWrapper =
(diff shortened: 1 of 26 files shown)


paimon-core/src/main/java/org/apache/paimon/table/source/DeletionFile.java
```diff
@@ -0,0 +1,139 @@
...
+ *   <li>The first 4 bytes are length, should equal to {@link #length()}.
+ *   <li>Next 4 bytes are the magic number, should be equal to 1581511376.
+ *   <li>The remaining content should be a RoaringBitmap.
+ * </ul>
+ */
+@Public
+public class DeletionFile {
+
+    private final String path;
+    private final long offset;
```
Zouxxyy, line 46: offset and length maybe can be int, because the size of deletionFile will not be too large, currently, the dv ranges in index meta are just int pair now...

paimon-core/src/main/java/org/apache/paimon/operation/KeyValueFileStoreRead.java
```diff
@@ -267,17 +277,34 @@ private RecordReader<KeyValue> batchMergeRead(
...
                         // See comments on DataFileMeta#extraFiles.
                         String fileName = changelogFile(file).orElse(file.fileName());
-                        return readerFactory.createRecordReader(
-                                file.schemaId(), fileName, file.fileSize(), file.level());
+                        RecordReader<KeyValue> reader =
+                                readerFactory.createRecordReader(
+                                        file.schemaId(), fileName, file.fileSize(), file.level());
+                        if (deletionFile != null) {
+                            DeletionVector deletionVector =
+                                    DeletionVector.read(fileIO, deletionFile);
```
Zouxxyy, line 304: Maybe can add this to IndexFileHandler, readDeletionVector contains version and crc verification

    public DeletionVector readDeletionVector(DeletionFile deletionFile) {
        return deletionVectorsIndex.readDeletionVector(
                new Path(deletionFile.path()),
                Pair.of(deletionFile.offset(), deletionFile.length()));
    }

paimon-core/src/main/java/org/apache/paimon/operation/KeyValueFileStoreRead.java
```diff
@@ -267,17 +277,34 @@ private RecordReader<KeyValue> batchMergeRead(
...
+                            reader = ApplyDeletionVectorReader.create(reader, deletionVector);
```
Zouxxyy, line 305: Worry about conflicts with `readerFactoryBuilder.withDeletionVectorSupplier` in the future

Zouxxyy (commented):
Thanks for the modification! leave some questions

paimon-core/src/main/java/org/apache/paimon/table/source/DeletionFile.java
JingsongLi: It is good to keep extensibility here.

paimon-core/src/main/java/org/apache/paimon/operation/KeyValueFileStoreRead.java
JingsongLi: I don't want introduce `IndexFileHandler` here, imagine a C++Reader, which should not rely on any complex classes, the simpler the better.

paimon-core/src/main/java/org/apache/paimon/operation/KeyValueFileStoreRead.java
JingsongLi: Only the ApplyDeletionVectorReader is the same part, and the Deletion is not conflicting and can be stacked.

In the future, we should avoid disturbing `ReaderFactoryBuilder` on the writer side and have independent deletion logic.

Zouxxyy (approved):
+1

Zouxxyy:
#2898

PR #2991, opened 2024-03-12
[core] Expose file io for other compute system like trino.

leaves12138:
I am working on paimon-trino. Paimon wants to use trino file system, but after serialize and deseialize table, file io will lose. (Trino file system can't be serialized, and every worker should use its own session to create file system). So I want to expose file io in table that we can't set trino file system in. 

* If paimon table need file io, it should be a FileIOTable.
* CatalogContext should accept self-defined file io.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/catalog/CatalogFactory.java
+++ b/paimon-core/src/main/java/org/apache/paimon/catalog/CatalogFactory.java
@@ -67,16 +67,6 @@ public interface CatalogFactory extends Factory {
     }
 
     static Catalog createCatalog(CatalogContext context, ClassLoader classLoader) {
-        Options options = context.options();
-        String metastore = options.get(METASTORE);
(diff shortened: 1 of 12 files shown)


paimon-core/src/main/java/org/apache/paimon/catalog/CatalogFactory.java
```diff
@@ -92,6 +82,30 @@ static Catalog createCatalog(CatalogContext context, ClassLoader classLoader) {
             throw new UncheckedIOException(e);
         }
 
+        return createCatalog(context, classLoader, warehousePath, fileIO);
+    }
+
+    static Catalog createCatalog(CatalogContext context, FileIO fileIO) {
```
JingsongLi, line 88: Can we just add `FileIOLoader preferIOLoader` to `CatalogContext`?

PR #2992, opened 2024-03-12
[core] KeyValueTableRead should always project keys

JingsongLi:
### Purpose

In fact, `KeyValueTableRead` only reads the content of the value and does not read the columns of the key. We should default to opening the projection of the key.


Zouxxyy (approved):
+1

PR #2995, opened 2024-03-12
[core] Remove file io from splits and Add preferFileIO for Paimon catalog.

leaves12138:
Reason:
* Split better not contain `FileIO`, because compute system file io (which wrapped as paimon file io) may not be able to serialized.
* We should add `preferFileIO` for testing wrapped file io by compute system (like trino).

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/catalog/CatalogContext.java
+++ b/paimon-common/src/main/java/org/apache/paimon/catalog/CatalogContext.java
@@ -41,14 +41,17 @@ public class CatalogContext {
 
     private final Options options;
     private final Configuration hadoopConf;
+    @Nullable private final FileIOLoader preferIOLoader;
     @Nullable private final FileIOLoader fallbackIOLoader;
 
     private CatalogContext(
             Options options,
             @Nullable Configuration hadoopConf,
+            @Nullable FileIOLoader preferIOLoader,
             @Nullable FileIOLoader fallbackIOLoader) {
         this.options = checkNotNull(options);
         this.hadoopConf = hadoopConf == null ? getHadoopConfiguration(options) : hadoopConf;
+        this.preferIOLoader = preferIOLoader;
         this.fallbackIOLoader = fallbackIOLoader;
     }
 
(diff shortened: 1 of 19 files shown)


paimon-common/src/main/java/org/apache/paimon/catalog/CatalogContext.java
```diff
@@ -84,6 +95,11 @@ public Configuration hadoopConf() {
         return hadoopConf;
     }
 
+    @Nullable
+    public FileIOLoader preIO() {
```
JingsongLi, line 99: preferIO

paimon-common/src/main/java/org/apache/paimon/fs/FileIO.java
```diff
@@ -374,6 +385,10 @@ static FileIO get(Path path, CatalogContext config) throws IOException {
 
         if (loader == null) {
             String fallbackMsg = "";
+            String preMsg = "";
+            if (preIO != null) {
+                preMsg = " " + preIO.getClass().getSimpleName() + " also cannot access this path.";
```
JingsongLi, line 390: Do not name `pre`, in english, there are so many pre..

paimon-core/src/main/java/org/apache/paimon/table/source/DataFilePlan.java
```diff
@@ -20,12 +20,13 @@
...
 import org.apache.paimon.table.source.snapshot.StartingScanner;
 
+import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
 /** Scanning plan containing snapshot ID and input splits. */
-public class DataFilePlan implements TableScan.Plan {
+public class DataFilePlan implements TableScan.Plan, Serializable {
```
JingsongLi, line 29: Why here need to be `Serializable`?

paimon-core/src/main/java/org/apache/paimon/table/source/SimplePlan.java
```diff
@@ -0,0 +1,57 @@
...
+
+import org.apache.paimon.table.source.snapshot.SnapshotReader;
+
+import org.jetbrains.annotations.Nullable;
+
+import java.io.Serializable;
+import java.util.List;
+
+/** A simple implementation of {@link SnapshotReader.Plan}. */
+public class SimplePlan implements SnapshotReader.Plan, Serializable {
```
JingsongLi, line 29: `PlanImpl`?

paimon-core/src/main/java/org/apache/paimon/table/system/AggregationFieldsTable.java
```diff
@@ -129,17 +133,17 @@ private static class AggregationSplit implements Split {
 
         private static final long serialVersionUID = 1L;
 
-        private final FileIO fileIO;
```
JingsongLi, line 132: Can you add a test to verify no fileIO here?

paimon-core/src/main/java/org/apache/paimon/table/source/DataFilePlan.java
leaves12138: Because we need to serialize `FilesSplit`, and we don't put `FileStoreTable `in `FilesSplit`, then we need to put plan in it. `TableScan.Plan` will be serialized when `FilesSplit` serializing.  Same for `PartitionsSplit`.

JingsongLi: No, you just need `List<Split>`.

JingsongLi:
+1

PR #2997, opened 2024-03-13
[flink] Remove MigrateDatabaseProcedure's dependency on Hive

yuzelin:
### Purpose

paimon-flink module should not depend on Hive directly.


JingsongLi:
+1

PR #3005, opened 2024-03-14
[common] Cache the ServiceLoader.

hzjhjjyy:
### Purpose

To avoid the performance overhead caused by multiple invocations of `ServiceLoader#load()`, use `ServiceLoaderUtils` for caching in places where `ServiceLoader#load()` is frequently called, while also considering its thread safety.


paimon-common/src/main/java/org/apache/paimon/fs/FileIO.java
```diff
@@ -61,6 +63,8 @@ public interface FileIO extends Serializable {
 
     Logger LOG = LoggerFactory.getLogger(FileIO.class);
 
+    AtomicReference<Map<String, FileIOLoader>> FILEIOLOADERS = new AtomicReference<>();
```
JingsongLi, line 66: I think we don't need to cache `FileIO`, it is created in catalog creation.

PR #3007, opened 2024-03-14
[core] Fix merge index out of range for sequence field

JingsongLi:
### Purpose

In the reading merging process, if there is a projection inside, we did not consider the existence of the sequence field and its index position.


yuzelin (approved):
+1

PR #3013, opened 2024-03-14
Reduce the length of lock_id

### LINKED ISSUE #3012: [Bug] Why JDBC metastore LOCK_ID so big?
### Search before asking

### Paimon version

This is mysql's `paimon_distributed_locks` table statement. The lock_id size is set to 1000. Does it really need to be that big?
In fact, this id is only a concatenation of `catalogName, database and table` variables. 1000 has exceeded the maximum length of the default index of mysql. Is it unreasonable to modify the mysql configuration for this in the production environment?
```
CREATE TABLE paimon_distributed_locks
(
    lock_id             VARCHAR(1000)                       NOT NULL,
    acquired_at         TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL,
    expire_time_seconds BIGINT    DEFAULT 0                 NOT NULL,
    PRIMARY KEY (lock_id)
)
```

### Compute Engine

flink1.17

### Minimal reproduce step

use mysql 5.7 and create catalog enable lock.enable

reduce lock_id length

### Anything else?

_No response_

gaoyan1998:
Linked issue: close #3012

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/jdbc/MysqlDistributedLockDialect.java
+++ b/paimon-core/src/main/java/org/apache/paimon/jdbc/MysqlDistributedLockDialect.java
@@ -27,7 +27,7 @@ public class MysqlDistributedLockDialect extends AbstractDistributedLockDialect 
                 + JdbcUtils.DISTRIBUTED_LOCKS_TABLE_NAME
                 + "("
                 + JdbcUtils.LOCK_ID
-                + " VARCHAR(1000) NOT NULL,"
+                + " VARCHAR(200) NOT NULL,"
                 + JdbcUtils.ACQUIRED_AT
                 + " TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL,"
                 + JdbcUtils.EXPIRE_TIME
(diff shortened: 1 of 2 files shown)


sunxiaojian:
@gaoyan1998 Cannot only reduce the length of distributed lock tables. LockId is concatenated from catalog_key, database, and table

gaoyan1998:
> @gaoyan1998 The length of the 'lock_id' field in distributed lock tables cannot be reduced on its own. The 'lock_id' is a concatenation of 'catalog_key', 'database_name', and 'table_name'.

@sunxiaojian  Yes, but I think that even these fields together should not need 1000 so long,i cant understand, if there are concerns can also be set to 500 or 700, I think it is also enough, 1000 will lead to mysql table create problems, its exceeded the maximum index limit.

Of course this can be solved by modifying the mysql parameters, but this is a heavier operation, especially in a production environment.

Just a small suggestion when using it, there may be other things I'm not thinking about, and if it really doesn't work I can close this pr

JingsongLi:
I think it is better to provide a option for this length.

gaoyan1998:
> I think it is better to provide a option for this length.

What kind of option？

sunxiaojian:
> > I think it is better to provide a option for this length.
> 
> What kind of option？

refer: https://github.com/apache/incubator-paimon/pull/3069

PR #3014, opened 2024-03-15
[hotfix] Fix unstable testRandomCdcEventsUnawareBucket

yuzelin:
### Purpose

The TestCdcSourceFunction is bounded source, which will cause the downstream to produce committable with ID Long.MAX_VALUE. But the compaction of unaware bucket is unbounded, so it won't call the `endInput()`, thus the final committable may not be handled.

Additionally, revert some test codes, so others tests may be more stable.


JingsongLi (approved):
+1

PR #3018, opened 2024-03-15
[core] Refactor NullFalseLeafFunction

Zouxxyy:
### Purpose

- Currently `In` and `NotIn` predicate are null false, this PR extract a common parent class `NullFalseLeafFunction` for them
- fix when all stats is null, equal(0, null) still should be false

### Tests

- All original UT passed
- New case
```
PredicateBuilder builder = new PredicateBuilder(RowType.of(new IntType()));
Predicate predicate = builder.equal(0, null);
// new
assertThat(test(predicate, 1, new FieldStats[] {new FieldStats(null, null, null)})).isEqualTo(false);
```


Zouxxyy:
close it currently

PR #3021, opened 2024-03-15
[core] Reduce schema read for readers

JingsongLi:
### Purpose

We already store TableSchema in `AbstractFileStoreTable`, but we don't use it.

We can use it to avoid read schema from filesystem.


leaves12138 (approved):
This will speed up "paimon" a lot when storage is "oss"!

+1

PR #3023, opened 2024-03-15
[codegen] EqualiserCodeGenerator supports ARRAY<ROW>

yuzelin:
### Purpose

See the test case:
If we didn't handle ARRAY<ROW>, the `valueEqualiser` in `FullChangelogMergeFunctionWrapper` and `LookupChangelogMergeFunctionWrapper` might throw exception when call `equals`.


JingsongLi (approved):
+1

PR #3024, opened 2024-03-15
[core] Dv table supports value filter pushdown

Zouxxyy:
### Purpose

subTask of #2898, The dv table no longer needs to be merged, so we should enable it to have the ability to non-primary key filtering pushdown.


JingsongLi (approved):
+1

PR #3031, opened 2024-03-17
[core] Use var length encoding for row position

JingsongLi:
### Purpose

Reduce disk space.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/mergetree/LookupLevels.java
+++ b/paimon-core/src/main/java/org/apache/paimon/mergetree/LookupLevels.java
@@ -48,6 +48,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.time.Duration;
+import java.util.Arrays;
 import java.util.Comparator;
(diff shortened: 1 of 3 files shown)


paimon-core/src/main/java/org/apache/paimon/mergetree/LookupLevels.java
```diff
@@ -342,10 +345,10 @@ public byte[] persistToDisk(KeyValue kv, long rowPosition) {
                 segment.put(bytes.length - 1, kv.valueKind().toByteValue());
                 return bytes;
             } else {
-                byte[] bytes = new byte[8];
-                MemorySegment segment = MemorySegment.wrap(bytes);
-                segment.putLong(0, rowPosition);
-                return bytes;
+                // len: min 1 - max 9
```
Zouxxyy, line 348: Maybe we can add two constants to VarLengthIntUtils to represent the min max of encodeLong

Zouxxyy (approved):
+1

Zouxxyy (approved):

PR #3036, opened 2024-03-18
[core] Fix dv table with partial-update and aggregate

Zouxxyy:
### Purpose

Fix dv table with partial-update and aggregate when changelog-producer = none


JingsongLi (approved):
+1

PR #3037, opened 2024-03-18
[core] Introduce BatchTableCommit.truncateTable

JingsongLi:
### Purpose

1. Introduce BatchTableCommit.truncateTable
2. Fix `TableWriterBenchmark`, it should not flush per 10000 rows.


leaves12138 (approved):
+1

PR #3039, opened 2024-03-19
[core] Optimize write performance

JingsongLi:
### Purpose

- Improve performance for `InternalRowSerializer`.
- Improve performance for `RowDataVectorizer`.


leaves12138 (approved):
+1

PR #3040, opened 2024-03-19
[core] Reduce getFileSize for avro reader

JingsongLi:
### Purpose

More optimization after https://github.com/apache/incubator-paimon/pull/2918


leaves12138 (approved):
+1

PR #3042, opened 2024-03-19
[core] support to drop partition when delete tag

### LINKED ISSUE #3041: [Feature] Support to drop partition when delete the tag
### Search before asking

### Motivation

Currently, the tag-to-partition only support create partition to hive metastore when tag is created. We should also drop the partition in metastore when delete the tag.

### Solution

Add `notifyDeletion` in `TagCallback`

### Anything else?

_No response_

Aitozi:
### Purpose

Linked issue: close #3041

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/BatchWriteGeneratorTagOperator.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/BatchWriteGeneratorTagOperator.java
@@ -21,6 +21,7 @@ package org.apache.paimon.flink.sink;
 import org.apache.paimon.Snapshot;
 import org.apache.paimon.operation.TagDeletion;
 import org.apache.paimon.table.FileStoreTable;
+import org.apache.paimon.table.sink.TagCallback;
 import org.apache.paimon.utils.SnapshotManager;
 import org.apache.paimon.utils.TagManager;
 
@@ -64,6 +65,8 @@ public class BatchWriteGeneratorTagOperator<CommitT, GlobalCommitT>
     private final CommitterOperator<CommitT, GlobalCommitT> commitOperator;
 
     protected final FileStoreTable table;
+
+    private transient List<TagCallback> callbacks;
 
     public BatchWriteGeneratorTagOperator(
             CommitterOperator<CommitT, GlobalCommitT> commitOperator, FileStoreTable table) {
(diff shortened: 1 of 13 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/BatchWriteGeneratorTagOperator.java
```diff
@@ -114,15 +117,15 @@ private void createTag() {
         try {
             // If the tag already exists, delete the tag
             if (tagManager.tagExists(tagName)) {
-                tagManager.deleteTag(tagName, tagDeletion, snapshotManager);
+                tagManager.deleteTag(tagName, tagDeletion, snapshotManager, callbacks);
```
JingsongLi, line 120: Be careful here, you need to create new callbacks, the callbacks will be closed here.

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveMetastoreClient.java
```diff
@@ -91,6 +91,22 @@ public void addPartition(LinkedHashMap<String, String> partitionSpec) throws Exc
         }
     }
 
+    @Override
+    public void deletePartition(LinkedHashMap<String, String> partitionSpec) throws Exception {
+        List<String> partitionValues = new ArrayList<>(partitionSpec.values());
+        try {
+            client.getPartition(
```
JingsongLi, line 98: Why we need to get first?

Aitozi: Not necessary, removed.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/BatchWriteGeneratorTagOperator.java
Aitozi: You are right, but I see some other same usage eg: `TagAutoCreation#callbacks`?  I think it's the same error, do i have to fix it in this PR ?

JingsongLi: Yes, you can fix them together.

JingsongLi (approved):
+1

PR #3044, opened 2024-03-19
[core] support create tag based on tag

### LINKED ISSUE #3043: [Feature] Support to create a tag on a exists tag
### Search before asking

### Motivation

When creating a tag on an expired snapshot, it will fail with `snapshot doesn't exist`.
If the tag of the snapshot have been created, and want to retrigger the `TagCallback` by creating the tag again. I think it's reasonable. We can fallback to the `tags` to find the snapshot.

### Solution

_No response_

### Anything else?

_No response_

Aitozi:
### Purpose

Linked issue: close #3043


JingsongLi (approved):
+1

Looks reasonable.

PR #3051, opened 2024-03-19
[core] add initial map size to solve load index slowly.

### LINKED ISSUE #3038: [Bug] load index too slow when restart flink job using dynamic bucket.
### Search before asking

### Paimon version

0.7

### Compute Engine

Flink

### Minimal reproduce step

Can try to write data using dynamic bucket with datagen generate 0 to 500,000,000 keys, stop the job and restart.

This cost around 20 minute to load index, it's too slow.

### Anything else?

_No response_

Stephen0421:
### Purpose

Linked issue: close #3038

When using dynamic bucket, i find out that when restart the job, it will cost long time to load index, and after observing the flame graph, i find out the thread was stuck at :
image
and by watching the code line 249 in Int2ShortOpenHashMap was quoted by 'There's always an unused entry.' 

So i add initial Int2ShortOpenHashMap size to avoid the thread stuck to accelerate the load process.

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/utils/Int2ShortHashMap.java
+++ b/paimon-common/src/main/java/org/apache/paimon/utils/Int2ShortHashMap.java
@@ -29,6 +29,10 @@ public class Int2ShortHashMap {
         this.map = new Int2ShortOpenHashMap();
     }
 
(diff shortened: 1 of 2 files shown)


JingsongLi (approved):
+1
Thanks @Stephen0421 for the contribution!

PR #3052, opened 2024-03-20
[core] Add validation for fields prefix

Zouxxyy:
### Purpose

Add validation for fields prefix to check field exist


JingsongLi (approved):
+1

PR #3055, opened 2024-03-20
[core] Optimize first_row batch read

JingsongLi:
### Purpose

First row merge engine does not need to be merged in batch read, in this PR, we deal with FIRST ROW just like deletion vectors:
1. Scan only read level bigger that 0.
2. Split splits for first_row.
3. Make Raw Files for first_row.
4. Read no merge when raw file exists.


Zouxxyy (approved):
+1

PR #3059, opened 2024-03-20
[core] Introduce SplitGroup for SplitGenerator

Zouxxyy:
### Purpose

- make SplitGenerator return files + rawConvertible
- optimizer split generate

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/source/MergeTreeSplitGenerator.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/source/MergeTreeSplitGenerator.java
@@ -62,7 +62,7 @@ public class MergeTreeSplitGenerator implements SplitGenerator {
 
     @Override
     public List<SplitGroup> splitForBatch(List<DataFileMeta> files) {
-        boolean rawConvertible = files.stream().allMatch(DataFileMeta::rawConvertible);
+        boolean rawConvertible = files.stream().allMatch(this::rawConvertible);
         boolean oneLevel =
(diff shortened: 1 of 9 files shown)


paimon-core/src/main/java/org/apache/paimon/table/source/MergeTreeSplitGenerator.java
```diff
@@ -61,11 +61,17 @@ public MergeTreeSplitGenerator(
     }
 
     @Override
-    public List<List<DataFileMeta>> splitForBatch(List<DataFileMeta> files) {
-        if (deletionVectorsEnabled || mergeEngine == FIRST_ROW) {
+    public List<SplitGroup> splitForBatch(List<DataFileMeta> files) {
+        boolean rawConvertible = files.stream().allMatch(DataFileMeta::rawConvertible);
+        boolean oneLevel =
```
JingsongLi, line 66: oneLevel and not zero level.

You should add test for this.

JingsongLi (commented):
Hi @Zouxxyy , can you add more tests for your modification?

Zouxxyy:
> Hi @Zouxxyy , can you add more tests for your modification?

Added~

JingsongLi:
any idea about above comment: `oneLevel and not zero level.`?

paimon-core/src/main/java/org/apache/paimon/table/source/MergeTreeSplitGenerator.java
Zouxxyy: move rawConvertible to avoid potential problems with the append table

JingsongLi:
+1

PR #3062, opened 2024-03-20
[cdc] Update the latest dynamic options to the table schema file when synchronizing an existing paimon table using the cdc action.

zhuangchong:
### Purpose

When using cdc action to synchronize the existing paimon table, the latest dynamic options are not updated to the table schema file.

You can first create a paimon table, start a cdc action and configure several new options. After the action is started, check the latest schema of the table. The latest schema of the table does not have the latest options.


yuzelin (approved):
+1

PR #3063, opened 2024-03-20
[hotfix] Add project root path to files to facilitate compiling modules in child dirs

Zouxxyy:
### Purpose

Someone may want to compile a child module in the child directory instead of using `-pl`, but after #2883 using this way to compile will raise an error.
Even though I don't agree with compiling like this (doge.jpg), still find a way to fix it


yuzelin (approved):
+1

Issue #3057, opened 2024-03-20
[Umbrella][Feature] Append table supports UPDATE/DELETE/...


YannByron:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/incubator-paimon/issues) and found nothing similar.

### Motivation

Enrich the append table to meet various scenarios.

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

PR #3067, opened 2024-03-21
[spark] minor refactor SparkWriter

YannByron:
### Purpose

Linked issue: close #xxx

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/commands/BucketProcessor.scala
+++ b/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/commands/BucketProcessor.scala
@@ -22,7 +22,7 @@ import org.apache.paimon.data.{InternalRow => PaimonInternalRow}
 import org.apache.paimon.index.HashBucketAssigner
 import org.apache.paimon.spark.SparkRow
 import org.apache.paimon.spark.util.EncoderUtils
-import org.apache.paimon.table.{AbstractFileStoreTable, FileStoreTable}
+import org.apache.paimon.table.FileStoreTable
 import org.apache.paimon.table.sink.RowPartitionKeyExtractor
 
 import org.apache.spark.TaskContext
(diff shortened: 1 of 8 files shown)


JingsongLi:
Hi @YannByron , Please roughly explain why refactoring is necessary.

YannByron:
Currently, all the spark write operations use `WriteIntoPaimonTable` that has two `write data to files` and `commit` steps. For supporting update/delete/... on append tables, need to split these two steps: `write data to files` is extracted out, while `commit` is kept in certain write operation. As thus, like support `update` on append table, we can find the files that has to be rewrote, write the updated data, commit both the new data files and deleted files.

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/commands/BucketProcessor.scala
```diff
@@ -0,0 +1,142 @@
...
+
+case class CommonBucketProcessor(
+    table: FileStoreTable,
+    bucketColIndex: Int,
+    encoderGroup: EncoderSerDeGroup)
+  extends BucketProcessor {
+
+  def processPartition(rowIterator: Iterator[Row]): Iterator[Row] = {
+    val rowType = table.rowType()
+    val rowKeyExtractor = table.asInstanceOf[AbstractFileStoreTable].createRowKeyExtractor()
```
JingsongLi, line 66: If you want this extractor, you could introduce a `createRowKeyExtractor` to `FileStoreTable`.

Subsequent matters related to permission security may require us to wrap a `FileStoreTable` class, so it is best to rely only on the interface rather than the implementation.

YannByron: I wanted to do that at first...OK.

YannByron:
https://github.com/apache/paimon/issues/3057

PR #3068, opened 2024-03-21
[core] File index interface and file format.

leaves12138:
Related to PIP-17 (https://cwiki.apache.org/confluence/display/PAIMON/PIP-17%3A+Introduce+secondary+column+index)

index file format:

image

REDUNDANT field for compatibility of next version.

var bytes (head) = bytes length + real bytes
var bytes (body) = column bytes + column bytes + ...


paimon-common/src/main/java/org/apache/paimon/fileindex/FileIndex.java
```diff
@@ -0,0 +1,87 @@
...
+
+package org.apache.paimon.fileindex;
+
+import org.apache.paimon.types.DataType;
+
+/**
+ * Secondary index filter interface. Return true, means we need to search this file, else means
+ * needn't.
+ */
+public interface FileIndex {
```
JingsongLi, line 27: Maybe it can be a `FunctionVisitor<Boolean>`?

paimon-common/src/main/java/org/apache/paimon/fileindex/FileIndexPredicateUtil.java
```diff
@@ -0,0 +1,172 @@
...
+            if (!tester.test(filePredicate)) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    public static byte[] serializeIndexMap(String type, Map<String, byte[]> indexMap) {
+        try {
+            ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
```
JingsongLi, line 119: You need to define a file structure to separate offsets and index bytes.

paimon-common/src/main/java/org/apache/paimon/fileindex/ObjectToBytesVisitor.java
```diff
@@ -0,0 +1,290 @@
...
+import org.apache.paimon.types.TinyIntType;
+import org.apache.paimon.types.VarBinaryType;
+import org.apache.paimon.types.VarCharType;
+
+import java.util.Arrays;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+
+/** Convert different type object to bytes. */
+public class ObjectToBytesVisitor implements DataTypeVisitor<Function<Object, byte[]>> {
```
JingsongLi, line 56: Maybe we don't need this class now.

paimon-common/src/main/java/org/apache/paimon/fileindex/FileIndexFile.java
```diff
@@ -0,0 +1,336 @@
...
+
+        private byte[] constructBody(
+                Map<String, byte[]> bytesMap, Map<String, Pair<Integer, Integer>> bodyInfo)
+                throws IOException {
+            ByteArrayOutputStream baos = new ByteArrayOutputStream();
+
+            for (Map.Entry<String, byte[]> entry : bytesMap.entrySet()) {
+                Integer startPosition = baos.size();
+                baos.write(entry.getValue());
+                bodyInfo.put(entry.getKey(), Pair.of(startPosition, baos.size() - 1));
```
zyl891229, line 100: Would it be easier to read with offset and length？

leaves12138: Good question, but I think `startPosition` and `endPosition` are more convenient to wrap for `BytesInputWrapper`.

paimon-common/src/main/java/org/apache/paimon/fileindex/FileIndexFile.java
```diff
@@ -0,0 +1,336 @@
...
+import java.io.DataOutput;
+import java.io.DataOutputStream;
+import java.io.EOFException;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+/** File index file format. Put all column and offset in the header. */
```
JingsongLi, line 41: Document your file format.

paimon-common/src/main/java/org/apache/paimon/fileindex/FileIndexFile.java
```diff
@@ -0,0 +1,336 @@
...
+public final class FileIndexFile {
```
JingsongLi, line 42: `LocalIndexFile`?

paimon-common/src/main/java/org/apache/paimon/fileindex/FileIndexFile.java
```diff
@@ -0,0 +1,336 @@
...
+
+        public Writer(OutputStream outputStream) {
+            this.dataOutputStream = new DataOutputStream(outputStream);
+        }
+
+        public void writeColumnIndex(String indexType, Map<String, byte[]> bytesMap)
+                throws IOException {
+
+            Map<String, Pair<Integer, Integer>> bodyInfo = new HashMap<>();
+            byte[] body = constructBody(bytesMap, bodyInfo);
```
JingsongLi, line 84: why can you just write to `dataOutputStream`?

paimon-common/src/main/java/org/apache/paimon/fileindex/FileIndexFile.java
```diff
@@ -0,0 +1,336 @@
...
+    /** Reader for file index file. */
+    public static class Reader implements Closeable {
+
+        private final SeekableInputStream seekableInputStream;
+        // get header and cache it.
+        private final Map<String, Pair<Integer, Integer>> header = new HashMap<>();
+        private final Map<String, DataField> fields = new HashMap<>();
+        private final String type;
+
+        public Map<String, Pair<Integer, Integer>> getHeader() {
```
JingsongLi, line 163: this method should be placed after constructor.

paimon-common/src/main/java/org/apache/paimon/fileindex/FileIndexFile.java
```diff
@@ -0,0 +1,336 @@
...
+        }
+
+        @Override
+        public void close() throws IOException {
+            IOUtils.closeQuietly(seekableInputStream);
+        }
+    }
+
+    /** Wrap the origin stream with offset start and end. */
+    static class BytesInputWrapper extends SeekableInputStream {
```
JingsongLi, line 247: I cannot get why we need this, why not just read bytes directly?

paimon-common/src/main/java/org/apache/paimon/fileindex/FileIndexPredicate.java
```diff
@@ -0,0 +1,165 @@
...
+                });
+    }
+
+    @Override
+    public void close() throws IOException {
+        this.reader.close();
+    }
+
+    /** Predicate test worker. */
+    private static class FileIndexPredicateWorker implements PredicateVisitor<Boolean> {
```
JingsongLi, line 116: `FileIndexFieldPredicate`?

paimon-common/src/main/java/org/apache/paimon/fileindex/FileIndexPredicate.java
```diff
@@ -0,0 +1,165 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/** Utils to check secondary index (e.g. bloom filter) predicate. */
+public class FileIndexPredicate implements Closeable {
+
+    private final FileIndexFile.Reader reader;
+    private final Map<String, FileIndexPredicateWorker> cacheTester = new HashMap<>();
```
JingsongLi, line 48: cacheTester -> fieldPredicates?

paimon-common/src/main/java/org/apache/paimon/fileindex/FileIndexReader.java
```diff
@@ -0,0 +1,109 @@
...
+
+import java.util.List;
+
+/**
+ * Read file index from serialized bytes. Return true, means we need to search this file, else means
+ * needn't.
+ */
+public interface FileIndexReader extends FunctionVisitor<Boolean> {
+
+    FileIndexReader recoverFrom(SeekableInputStream inputStream);
```
JingsongLi, line 33: directly using byte[]?

paimon-common/src/main/java/org/apache/paimon/fileindex/FileIndexFile.java
```diff
@@ -0,0 +1,336 @@
...
+            // writeVersion
+            dataOutputStream.writeInt(Version.V_1.version());
+            // writeHeadLength
+            dataOutputStream.writeInt(headLength);
+            // writeIndexType
+            dataOutputStream.writeUTF(indexType);
+            // writeColumnSize
+            dataOutputStream.writeInt(bodyInfo.size());
+            // writeColumnInfo, offset = headLength
+            for (Map.Entry<String, Pair<Integer, Integer>> entry : bodyInfo.entrySet()) {
```
zyl891229, line 122: can we sort the key when write and use an algorithm (e.g. binary search) to speed up the search when read
in case of the large number of keys

leaves12138: We `must` read all head columns and put them into a hashmap not a array list, so we don't need to sort.

JingsongLi (approved):
+1

zyl891229:
+1

PR #3069, opened 2024-03-21
[Core]JdbcCatalog supports custom configuration for the length of lock table

sunxiaojian:
Linked issue: close (https://github.com/apache/incubator-paimon/issues/3046)

### DIFF
--- a/docs/content/how-to/creating-catalogs.md
+++ b/docs/content/how-to/creating-catalogs.md
@@ -210,7 +210,10 @@ You can configure any connection parameters that have been declared by JDBC thro
 
 You can also perform logical isolation for databases under multiple catalogs by specifying "catalog-key".
 
+Additionally, when creating a JdbcCatalog, you can specify the maximum length for the lock key by configuring "lock-key-max-length," which defaults to 255. Since this value is a combination of {catalog-key}.{database-name}.{table-name}, please adjust accordingly.
+
 You can define any default table options with the prefix `table-default.` for tables created in the catalog.
+
 
 {{< /tab >}}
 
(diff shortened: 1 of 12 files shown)


JingsongLi (commented):
Modify too more... I don't think we need to let all length configurable...

I think we can just modify this for lock...

sunxiaojian:
> Modify too more... I don't think we need to let all length configurable...
> 
> I think we can just modify this for lock...

Okay,  but this way, when other fields exceed the limit, they can only be resolved through the alter table

sunxiaojian:
> > Modify too more... I don't think we need to let all length configurable...
> > I think we can just modify this for lock...
> 
> Ok, but this way, when other fields exceed the limit, they can only be resolved through the alter table

docs/content/how-to/creating-catalogs.md
```diff
@@ -210,8 +210,11 @@ You can configure any connection parameters that have been declared by JDBC thro
 
 You can also perform logical isolation for databases under multiple catalogs by specifying "catalog-key".
 
+Additionally, when creating a JdbcCatalog, you can specify the maximum length for the lock key by configuring "lock-key-max-length," which defaults to 255. Since this value is a combination of {catalog-key}.{database-name}.{table-name}, please adjust accordingly.
```
JingsongLi, line 213: which defaults to 1000

docs/layouts/shortcodes/generated/jdbc_catalog_configuration.html
```diff
@@ -32,5 +32,11 @@
             <td>String</td>
             <td>Custom jdbc catalog store key.</td>
         </tr>
+        <tr>
+            <td><h5>lock-key-max-length</h5></td>
+            <td style="word-wrap: break-word;">1000</td>
```
JingsongLi, line 37: Here is 1000, but others are 255?

sunxiaojian: > Here is 1000, but others are 255?

@JingsongLi fixed

JingsongLi (approved):
+1

PR #3070, opened 2024-03-21
[core] Skip rewrite when delete row count is 0 with dv

Zouxxyy:
### Purpose

When upgrade is possible and delete row count is 0, we can avoid rewrite

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/LookupMergeTreeCompactRewriter.java
+++ b/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/LookupMergeTreeCompactRewriter.java
@@ -41,7 +41,7 @@ import java.util.List;
 
 import static org.apache.paimon.mergetree.compact.ChangelogMergeTreeRewriter.UpgradeStrategy.CHANGELOG_NO_REWRITE;
 import static org.apache.paimon.mergetree.compact.ChangelogMergeTreeRewriter.UpgradeStrategy.CHANGELOG_WITH_REWRITE;
-import static org.apache.paimon.mergetree.compact.ChangelogMergeTreeRewriter.UpgradeStrategy.NO_CHANGELOG;
+import static org.apache.paimon.mergetree.compact.ChangelogMergeTreeRewriter.UpgradeStrategy.NO_CHANGELOG_NO_REWRITE;
 
 /**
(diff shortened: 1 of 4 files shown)


paimon-core/src/main/java/org/apache/paimon/mergetree/compact/LookupMergeTreeCompactRewriter.java
```diff
@@ -101,10 +104,11 @@ protected UpgradeStrategy upgradeChangelog(int outputLevel, DataFileMeta file) {
...
         }
 
-        // TODO In deletionVector mode, since drop delete is required, rewrite is always required.
-        // TODO wait https://github.com/apache/incubator-paimon/pull/2962
-        // TODO but should be careful to not be deleted by DeletionVectorsMaintainer!
-        if (dvMaintainer != null) {
+        // In deletionVector mode, since drop delete is required, when delete row count > 0 rewrite
+        // is required.
+        if (dvMaintainer != null
+                && file.deleteRowCount().isPresent()
```
JingsongLi, line 110: file.deleteRowCount().map(cnt -> cnt > 0).orElse(true)

paimon-core/src/main/java/org/apache/paimon/mergetree/compact/MergeTreeCompactRewriter.java
```diff
@@ -98,5 +98,10 @@ protected <T> RecordReader<T> readerForMergeTree(
                 mergeSorter);
     }
 
+    /**
+     * Hook to handle the files before compact in rewriting stage. Note: In addition to normal
```
JingsongLi, line 102: remove documentation, we already change the name.

paimon-core/src/main/java/org/apache/paimon/mergetree/compact/LookupMergeTreeCompactRewriter.java
```diff
@@ -101,10 +104,11 @@ protected UpgradeStrategy upgradeChangelog(int outputLevel, DataFileMeta file) {
             return NO_CHANGELOG;
```
JingsongLi, line 104: NO_CHANGELOG -> NO_CHANGELOG_NO_REWRITE

JingsongLi (approved):
+1

PR #3076, opened 2024-03-22
[core]Support common jdbc catalog lock for filesystem catalog.

sunxiaojian:
### Purpose

Linked issue: close (https://github.com/apache/incubator-paimon/issues/3032)

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/catalog/LockContextUtils.java
+++ b/paimon-core/src/main/java/org/apache/paimon/catalog/LockContextUtils.java
@@ -1,0 +1,83 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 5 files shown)


sunxiaojian:
@FangYongs  @JingsongLi  PTAL

paimon-core/src/main/java/org/apache/paimon/catalog/LockContextUtils.java
```diff
@@ -0,0 +1,83 @@
...
+
+import java.sql.SQLException;
+import java.util.Optional;
+
+/** Utils for {@link org.apache.paimon.catalog.CatalogLock.LockContext}. */
+public class LockContextUtils {
+
+    private static final Logger LOG = LoggerFactory.getLogger(FileSystemCatalog.class);
+
+    private static JdbcClientPool connections;
```
JingsongLi, line 39: Why use static member here?

sunxiaojian: > Why using static member here?

@JingsongLi The main reason is that I declared all the methods in utils as static, and this place will not be destroyed immediately after use. It can be destroyed together with the process release, so I declared it as static

JingsongLi: I mean why it should be destroyed together with the process release.

JingsongLi (approved):
+1

JingsongLi:
I will revert this PR in https://github.com/apache/paimon/pull/3099

It seems that the current approach is wrong.

PR #3079, opened 2024-03-22
[Improve] Exceptions are not suitable for flow control. Use the catal…

daixiaoyu:
…og api to determine

paimon-flink-cdc

### Purpose

Exceptions are not suitable for flow control. Use the catalog api to determine


JingsongLi (commented):
Hi @daixiaoyu , for some performance reason, we cannot handle branches without using exceptions, which is acceptable.

PR #3082, opened 2024-03-24
[doc]introduce the minimum version of Maven.

wg1026688210:
### Purpose
#3063 used the Maven variable ${maven.multiModuleProjectDirectory}, which was added in versions after Maven 3.3.1. Therefore, it's necessary to declare the minimum required version of Maven.


JingsongLi (approved):
+1

PR #3084, opened 2024-03-24
[hive] Support recreate the hive table

silentxingtian:
### Purpose

Linked issue: close #xxx

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/schema/SchemaManager.java
+++ b/paimon-core/src/main/java/org/apache/paimon/schema/SchemaManager.java
@@ -129,12 +129,10 @@ public class SchemaManager implements Serializable {
     public void checkTableSchema(TableSchema oldSchema, TableSchema newSchema) {
         boolean isCommon =
                 oldSchema.version() == newSchema.version()
-                        && oldSchema.id() == newSchema.id()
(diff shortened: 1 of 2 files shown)


paimon-core/src/main/java/org/apache/paimon/schema/SchemaManager.java
```diff
@@ -124,16 +125,28 @@ public List<Long> listAllIds() {
         }
     }
 
+    /** Check TableScheme is be modified. */
+    public void checkTableSchema(TableSchema oldSchema, TableSchema newSchema) {
+        boolean isCommon =
+                oldSchema.version() == newSchema.version()
+                        && oldSchema.id() == newSchema.id()
```
JingsongLi, line 132: I don't think we need to compare schema id and options.

JingsongLi (changes requested):
Can you add test for this?

paimon-core/src/main/java/org/apache/paimon/schema/SchemaManager.java
silentxingtian: Okay, if there's no change in the entire logic, I'll remove the schema id and options

silentxingtian:
> Can you add test for this?

ok,I will write a test case code and submit it again

PR #3085, opened 2024-03-25
[docs] Apache Paimon became top level project.

leaves12138:
Happy graduation!

### DIFF
--- a/docs/layouts/partials/docs/inject/footer.html
+++ b/docs/layouts/partials/docs/inject/footer.html
@@ -26,7 +26,7 @@ under the License.
     <tbody>
     <tr>
         <td colspan="2">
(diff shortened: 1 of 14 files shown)


docs/layouts/partials/docs/inject/footer.html
```diff
@@ -30,7 +30,7 @@
         </td>
         <td>
             Apache Paimon is an effort undergoing incubation at The Apache Software Foundation (ASF), sponsored by the
-            Apache Incubator.
+            Apache.
```
Zouxxyy, line 33: The entire text should be replaced.

leaves12138: Resolved

JingsongLi (approved):
+1

PR #3088, opened 2024-03-25
[cdc] When the cdc source modifies field comments, all field comments in the existing paimon table will be set to empty.

zhuangchong:
…

### Purpose

When the cdc source modifies field comments, all field comments in the existing paimon table will be set to empty.

Steps:
Version 0.7, flink1.17, mysqlcdc2paimon whole database migration-action package. Modifying the comment of a field in the mysql table will cause the comments of all fields in the paimon table to be cleared.


yuzelin:
Can you add a test for MySQL sync?

yuzelin (approved):
+1

zhuangchong:
Thanks @yuzelin

PR #3093, opened 2024-03-26
[core] Refactor FileStorePathFactory to clean test methods

JingsongLi:

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/system/ManifestsTable.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/system/ManifestsTable.java
@@ -25,12 +25,11 @@ import org.apache.paimon.data.GenericRow;
 import org.apache.paimon.data.InternalRow;
 import org.apache.paimon.disk.IOManager;
 import org.apache.paimon.format.FileFormat;
-import org.apache.paimon.fs.FileIO;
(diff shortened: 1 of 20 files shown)


paimon-core/src/main/java/org/apache/paimon/table/system/ManifestsTable.java
```diff
@@ -157,25 +148,22 @@ public boolean equals(Object o) {
                 return false;
             }
             ManifestsSplit that = (ManifestsSplit) o;
-            return Objects.equals(location, that.location);
+            return Objects.equals(rowCount, that.rowCount);
```
leaves12138, line 151: I think this is dangerous. If I put `ManifestsSplit` in a `HashSet`. The splits those have the same rowCount will be de-duplicated.

leaves12138:
Thanks for @JingsongLi , left one comment.

paimon-core/src/main/java/org/apache/paimon/table/system/ManifestsTable.java
JingsongLi: There is no scenario that requires splitting different tables into one hashmap.

leaves12138 (approved):
+1

PR #3098, opened 2024-03-26
[core] Unify FileRecordReader and reduce file access

JingsongLi:
### Purpose

1. KeyValueDataFileRecordReader should just use FileRecordReader.
2. FileUtils.checkExists should just in exception path to reduce file access.


leaves12138 (approved):
+1

PR #3099, opened 2024-03-27
[catalog] Refactor Catalog Factory and revert jdbc lock in FileSystemCatalog

JingsongLi:
### Purpose

1. Refactor Catalog Factory, extract separate classes for factory and context.
2. Revert jdbc lock in FileSystemCatalog, lock should be supported from `AbstractCatalog.lockFactory`.


sunxiaojian:
+1

PR #3100, opened 2024-03-27
[core] Shade roaringbitmap dependency into paimon-common

Zouxxyy:
### Purpose

subtask of #2898

- wapper RoaringBitma in paimon-common
- Shade roaringbitmap dependency into paimon-common, paimon-common.jar, before 3.3M -> after 3.6M


JingsongLi (approved):
+1

PR #3102, opened 2024-03-27
[filesystems] Supported Tencent Cloud COS filesystem

chenxi0599:
### Purpose

Linked issue: close #xxx

### DIFF
--- a/docs/content/filesystems/cosn.md
+++ b/docs/content/filesystems/cosn.md
@@ -1,0 +1,102 @@
+---
+title: "cosn"
+weight: 3
+type: docs
+aliases:
+- /filesystems/cosn.html
+---
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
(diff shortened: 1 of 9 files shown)


chenxi0599:
I have passed the test on my own computer, using Tencent cos to read and write data

JingsongLi:
Hi @chenxi0599 , can you explain the difference between using s3 interface and using cos interface? Both can support Tencent Cloud COS filesystem.

chenxi0599:
Certainly, most object storage services are compatible with the S3 interface, including Aliyun OSS. However, I believe utilizing each object storage's native API would be a better choice. I have also compared accessing COS using both S3 SDK and COS SDK, and I feel that the performance with COS SDK is slightly better. Additionally, utilizing COS SDK allows for better representation of error messages returned by the server in the logs.

chenxi0599:
> Hi @chenxi0599 , can you explain the difference between using s3 interface and using cos interface? Both can support Tencent Cloud COS filesystem.

I've searched extensively for an answer to this question but haven't found one yet.  
I believe it's preferable to use Tencent Cloud COS's native interface whenever possible, as they may have optimized it for their specific platform.  I've also compared accessing COS using both the S3 SDK and COS SDK, and it seems like the performance is slightly better with the COS SDK.  Additionally, error messages returned by the server are better reflected in the logs when using the COS SDK.

JingsongLi:
Hi @chenxi0599 , I contacted the engineers of tencent cloud, it looks like it is OK to use S3 SDK to connect tencent COS. We can wait for more information.

PR #3103, opened 2024-03-27
[core] Expose VectorizedRecordIterator to accelerate append table read

JingsongLi:
### Purpose

For the integration of computing engines, if the computing engine can perceive that the returned iterator is a columnar batch, then the computing engine has more opportunities to accelerate computation.


yuzelin (approved):
+1

PR #3108, opened 2024-03-28
[core] Add limitation to the buffer spill disk usage

### LINKED ISSUE #3104: [Feature] Add limitation to disk usage of buffer spill
### Search before asking

### Motivation

Currently, in batch job, the buffer spill is enabled by default, but there is no limitation to the disk usage. It will lead to `disk quota exceed` error when write a big table. I think we can add a limitation to the disk usage in buffer spill.

### Solution

_No response_

### Anything else?

_No response_

Aitozi:
### Purpose

Linked issue: close #3104

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/disk/ChannelWithMeta.java
+++ b/paimon-core/src/main/java/org/apache/paimon/disk/ChannelWithMeta.java
@@ -24,13 +24,17 @@ public class ChannelWithMeta {
     private final FileIOChannel.ID channel;
     private final int blockCount;
     private final int numBytesInLastBlock;
-    private final long size;
+    private final long numEstimatedBytes;
 
-    public ChannelWithMeta(FileIOChannel.ID channel, int blockCount, int numBytesInLastBlock) {
+    public ChannelWithMeta(
+            FileIOChannel.ID channel,
+            int blockCount,
+            int numBytesInLastBlock,
+            long numEstimatedBytes) {
         this.channel = channel;
         this.blockCount = blockCount;
         this.numBytesInLastBlock = numBytesInLastBlock;
-        this.size = channel.getPathFile().length();
+        this.numEstimatedBytes = numEstimatedBytes;
     }
 
     public FileIOChannel.ID getChannel() {
@@ -45,7 +49,7 @@ public class ChannelWithMeta {
         return numBytesInLastBlock;
     }
 
-    public long getSize() {
-        return size;
+    public long getNumEstimatedBytes() {
+        return numEstimatedBytes;
     }
 }
--- a/paimon-core/src/main/java/org/apache/paimon/disk/ChannelWithMeta.java
+++ b/paimon-core/src/main/java/org/apache/paimon/disk/ChannelWithMeta.java
@@ -24,11 +24,13 @@ public class ChannelWithMeta {
     private final FileIOChannel.ID channel;
     private final int blockCount;
     private final int numBytesInLastBlock;
+    private final long size;
 
     public ChannelWithMeta(FileIOChannel.ID channel, int blockCount, int numBytesInLastBlock) {
         this.channel = channel;
         this.blockCount = blockCount;
         this.numBytesInLastBlock = numBytesInLastBlock;
+        this.size = channel.getPathFile().length();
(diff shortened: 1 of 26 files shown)


paimon-common/src/test/java/org/apache/paimon/fs/ByteArraySeekableStreamTest.java
```diff
@@ -52,7 +52,7 @@ public void testBasic() throws IOException {
         }
 
         for (int i = 0; i < RANDOM.nextInt(1000); i++) {
-            int position = RANDOM.nextInt(bl);
+            int position = RANDOM.nextInt(bl - 100);
```
JingsongLi, line 55: rebase master

paimon-core/src/main/java/org/apache/paimon/disk/ChannelWithMeta.java
```diff
@@ -24,11 +24,13 @@ public class ChannelWithMeta {
     private final FileIOChannel.ID channel;
     private final int blockCount;
     private final int numBytesInLastBlock;
+    private final long size;
 
     public ChannelWithMeta(FileIOChannel.ID channel, int blockCount, int numBytesInLastBlock) {
         this.channel = channel;
         this.blockCount = blockCount;
         this.numBytesInLastBlock = numBytesInLastBlock;
+        this.size = channel.getPathFile().length();
```
JingsongLi, line 33: pass size from outside instead of access file again.

paimon-core/src/main/java/org/apache/paimon/disk/ExternalBuffer.java
```diff
@@ -90,7 +96,16 @@ public void reset() {
     @Override
     public boolean flushMemory() throws IOException {
         spill();
-        return true;
+        return getDiskUsage() < maxDiskSize.getBytes();
```
JingsongLi, line 99: Should check first before spill?

paimon-core/src/main/java/org/apache/paimon/disk/ExternalBuffer.java
```diff
@@ -147,6 +163,7 @@ private void spill() throws IOException {
                                 ? inMemoryBuffer.getNumBytesInLastBuffer()
                                 : segment.size();
                 writer.writeBlock(Buffer.create(segment, bufferSize));
+                numEstimatedBytes += bufferSize;
             }
             LOG.info(
```
JingsongLi, line 168: You can just cache `writer.getSize()`.

paimon-core/src/main/java/org/apache/paimon/disk/ChannelWithMeta.java
```diff
@@ -24,11 +24,17 @@ public class ChannelWithMeta {
     private final FileIOChannel.ID channel;
     private final int blockCount;
     private final int numBytesInLastBlock;
+    private final long numEstimatedBytes;
```
JingsongLi, line 27: You can get real size, see comments.

paimon-core/src/main/java/org/apache/paimon/sort/BinaryExternalSortBuffer.java
```diff
@@ -154,7 +161,16 @@ public long getOccupancy() {
     @Override
     public boolean flushMemory() throws IOException {
         spill();
-        return true;
+        return getDiskUsage() < maxDiskSize.getBytes();
```
JingsongLi, line 164: ditto, we should check first?

paimon-core/src/main/java/org/apache/paimon/disk/ExternalBuffer.java
Aitozi: Yes, updated

paimon-core/src/main/java/org/apache/paimon/disk/ExternalBuffer.java
Aitozi: Get it, directly use `FileChannel.getSize()` can avoid an extra system call to get the file length, Right ?

JingsongLi:
Looks good to me!

PR #3110, opened 2024-03-28
[core] Introduce ExpireChangelogImpl to decouple the changelog lifecycle

Aitozi:
### Purpose

This PR is the first step for #2899. It's mainly do the following things.

- Add `changelog` related option to define the changlog lifecycle.
- Introduce `ExpireChangelogImpl` to handle the changelog expire

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/Changelog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/Changelog.java
@@ -39,7 +39,7 @@ import java.util.Map;
  */
 public class Changelog extends Snapshot {
 
-    private static final int CURRENT_VERSION = 3;
+    private static final int CURRENT_VERSION = 1;
 
     public Changelog(
             long id,
--- a/paimon-core/src/main/java/org/apache/paimon/Changelog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/Changelog.java
@@ -20,20 +20,15 @@ package org.apache.paimon;
 
 import org.apache.paimon.fs.FileIO;
 import org.apache.paimon.fs.Path;
-import org.apache.paimon.manifest.ManifestFileMeta;
-import org.apache.paimon.manifest.ManifestList;
 import org.apache.paimon.utils.JsonSerdeUtil;
 
 import org.apache.paimon.shade.jackson2.com.fasterxml.jackson.annotation.JsonCreator;
-import org.apache.paimon.shade.jackson2.com.fasterxml.jackson.annotation.JsonGetter;
 import org.apache.paimon.shade.jackson2.com.fasterxml.jackson.annotation.JsonProperty;
 
 import javax.annotation.Nullable;
 
 import java.io.IOException;
-import java.util.Collections;
-import java.util.List;
-import java.util.Objects;
+import java.util.Map;
 
 /**
  * The metadata of changelog. It generates from the snapshot file during expiration. So that the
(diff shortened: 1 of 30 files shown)


Aitozi:
After discuss with @JingsongLi . We decide not to support decouple the delta log in first version. The reason is that, it's not easy to determine whether a `DELETE` file is from an `APPEND` commit when expire a snapshot. One solution is to modify the `ManifestEntry` to mark the file source. We may implement this later, not in this PR.

paimon-core/src/main/java/org/apache/paimon/Changelog.java
```diff
@@ -0,0 +1,132 @@
...
+/**
+ * The metadata of changelog. It generates from the snapshot file during expiration. So that the
+ * changelog of the table can outlive the snapshot's lifecycle. A table's changelog can come from
+ * two source:
+ * <li>The changelog file. Eg: from the changelog-producer = 'input'
+ * <li>The delta files in the APPEND commits when the changelog-producer = 'none'
+ */
+public class Changelog extends Snapshot {
+
+    private static final int CURRENT_VERSION = 3;
```
Aitozi, line 42: The current version is set to 3, equal to the version of snapshot. Otherwise such as `1`, when pass the `Changelog` as `Snapshot` to `org.apache.paimon.operation.AbstractFileStoreScan#readManifests()` it will be recognize as a table_store_02_version.

Aitozi: Set version to 1 now. Refactor logic in `AbstractFileStoreScan#readManifests()`

paimon-core/src/main/java/org/apache/paimon/Changelog.java
```diff
@@ -0,0 +1,132 @@
...
+
+import java.io.IOException;
+import java.util.Map;
+
+/**
+ * The metadata of changelog. It generates from the snapshot file during expiration. So that the
+ * changelog of the table can outlive the snapshot's lifecycle. A table's changelog can come from
+ * two source:
+ * <li>The changelog file. Eg: from the changelog-producer = 'input'
+ * <li>The delta files in the APPEND commits when the changelog-producer = 'none'
```
JingsongLi, line 38: If currently this feature is not work. We can not document it now.

paimon-core/src/main/java/org/apache/paimon/Changelog.java
```diff
@@ -0,0 +1,132 @@
...
+ */
+public class Changelog extends Snapshot {
+
+    private static final int CURRENT_VERSION = 1;
```
JingsongLi, line 42: Why its version is 1? But the version of snapshot is 3?

paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreScan.java
```diff
@@ -386,6 +387,10 @@ private List<ManifestFileMeta> readManifests(Snapshot snapshot) {
             case DELTA:
                 return snapshot.deltaManifests(manifestList);
             case CHANGELOG:
+                if (snapshot instanceof Changelog) {
```
JingsongLi, line 390: We can just set version to 3.

paimon-core/src/main/java/org/apache/paimon/table/ExpireSnapshotsImpl.java
```diff
@@ -215,9 +222,59 @@ public int expireUntil(long earliestId, long endExclusiveId) {
             }
 
             Snapshot snapshot = snapshotManager.snapshot(id);
-            snapshotDeletion.cleanUnusedManifests(snapshot, skippingSet);
-
-            // delete snapshot last
+            if (changelogDecoupled) {
```
JingsongLi, line 225: Maybe you should write changelog first? Image there is a failover here.

Maybe you can add test in `FileStoreExpireTestBase`, to make sure changelog files generated.

paimon-core/src/main/java/org/apache/paimon/table/RollbackHelper.java
```diff
@@ -121,6 +135,38 @@ private List<Snapshot> cleanSnapshotsDataFiles(Snapshot retainedSnapshot) {
...
+            toBeCleaned.add(snapshotManager.changelog(i));
+            fileIO.deleteQuietly(snapshotManager.longLivedChangelogPath(i));
+        }
+
+        // delete data files of changelog
+        for (Changelog changelog : toBeCleaned) {
+            if (changelog.changelogManifestList() != null) {
+                snapshotDeletion.deleteAddedDataFiles(changelog.changelogManifestList());
+            }
+            if (changelog.deltaManifestList() != null) {
```
JingsongLi, line 159: Let's not consider deltaManifest now.

paimon-core/src/main/java/org/apache/paimon/table/RollbackHelper.java
```diff
@@ -33,6 +34,7 @@
 import java.io.UncheckedIOException;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.SortedMap;
```
JingsongLi, line 40: Not only Rollback, you need to consider orphan file clean too.

paimon-core/src/main/java/org/apache/paimon/Changelog.java
Aitozi: Removed

paimon-core/src/main/java/org/apache/paimon/table/ExpireSnapshotsImpl.java
Aitozi: The snapshot path is deleted after changelog files are generated. The test `org.apache.paimon.operation.ExpireSnapshotsTest#testChangelogOutLivedSnapshot` will verify the changelog metadata is generated.

paimon-core/src/main/java/org/apache/paimon/table/RollbackHelper.java
Aitozi: Yes, I think this can be done in another pr ?

JingsongLi: If the workload is not significant, you can consider completing it here

Aitozi: OK

JingsongLi (approved):
Looks good to me!

PR #3113, opened 2024-03-28
[cdc] Fix that kafka message value might be null which causes NPE

yuzelin:
### Purpose

Null value is produced by tombstone-events
https://debezium.io/documentation/reference/2.5/connectors/mysql.html#mysql-tombstone-events


JingsongLi (approved):
+1

PR #3119, opened 2024-03-29
[spark] Compact procedure support where parameter

Zouxxyy:
### Purpose

- support: `CALL sys.compact(table => 't', where => 'p>0 and p<3')`
- bug fix: Non-partition columns should not be in "partitions", e.g  `CALL sys.compact(table => 't', partitions => 'non_partition_col=1')`) which will result in data loss


leaves12138 (approved):
+1

yuzelin:
Should we also modify flink `compact`?

PR #3121, opened 2024-03-29
[flink] Specify restart strategy for SinkSavepointITCase

tsreaper:
### Purpose

Flink 1.19 uses exponential delay as the default restart strategy. It will make `SinkSavepointITCase` prone to timeout.

This PR specifies restart strategy for `SinkSavepointITCase` to fix this issue.

### Tests

No test is needed.

### API and Format

No changes.

### Documentation

Not needed.


JingsongLi:
+1

Issue #3123, opened 2024-03-30
[Feature] Refactor Kafka-CDC tests


yuzelin:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Motivation

This is an umbrella issue.
Kafka cdc tests are complicated. There are many test cases, we can extract common parts to clean the code.
Besides, Kafka cdc contains four formats now, some tests are redundant.

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

PR #3124, opened 2024-03-30
[cdc][refactor] Refactor records writing in Kafka CDC tests

yuzelin:
### Purpose

A step of #3123
Reduce codes for records writing;
Reuse Kafka objects.


JingsongLi:
+1

PR #3128, opened 2024-04-01
[core] Refactor: merge table-level ignore-delete options and filter delete data in writer layer

yuzelin:
### Purpose

Skip delete data in `AbstractFileStoreWrite`.

Additional changes:
1. `DeleteAction` doesn't change merge-engine and ignore-delete now. I think it equals to `DELETE` statement, so it shouldn't have other semantics.
2. `MergeIntoAction` change `ignore-delete` to `false`.

### API and Format
Deprecate:
first-row.ignore-delete
deduplicate.ignore-delete
partial-update.ignore-delete
Add:
ignore-delete

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/PartialUpdateMergeFunction.java
+++ b/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/PartialUpdateMergeFunction.java
@@ -67,7 +67,6 @@ public class PartialUpdateMergeFunction implements MergeFunction<KeyValue> {
     public static final String SEQUENCE_GROUP = "sequence-group";
 
     private final InternalRow.FieldGetter[] getters;
-    private final boolean ignoreDelete;
     private final Map<Integer, SequenceGenerator> fieldSequences;
     private final boolean fieldSequenceEnabled;
     private final Map<Integer, FieldAggregator> fieldAggregators;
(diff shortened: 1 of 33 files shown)


paimon-core/src/main/java/org/apache/paimon/mergetree/compact/PartialUpdateMergeFunction.java
```diff
@@ -104,12 +101,8 @@ public void add(KeyValue kv) {
         // refresh key object to avoid reference overwritten
         currentKey = kv.key();
 
-        // ignore delete?
+        // ignore or retract?
```
JingsongLi, line 104: remove this line? ignore what?

paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreWrite.java
```diff
@@ -122,13 +125,19 @@ public void withCompactExecutor(ExecutorService compactExecutor) {
...
+        }
+
         WriterContainer<T> container = getWriterWrapper(partition, bucket);
         container.writer.write(data);
         if (container.indexMaintainer != null) {
             container.indexMaintainer.notifyNewRecord(data);
         }
     }
 
+    protected abstract boolean skipData(T data);
```
JingsongLi, line 139: ignore in `TableWriteImpl`.

JingsongLi (approved):
+1

PR #3129, opened 2024-04-01
[spark] update for append table

YannByron:
### Purpose

https://github.com/apache/paimon/issues/3057

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/catalyst/analysis/RowLevelOp.scala
+++ b/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/catalyst/analysis/RowLevelOp.scala
@@ -18,10 +18,8 @@
 
 package org.apache.paimon.spark.catalyst.analysis
 
-import org.apache.paimon.AppendOnlyFileStore
 import org.apache.paimon.CoreOptions.{MERGE_ENGINE, MergeEngine}
 import org.apache.paimon.options.Options
-import org.apache.paimon.spark.catalyst.analysis.RowLevelOp.{AppendTable, PrimaryKeyTable}
 import org.apache.paimon.table.Table
 
 sealed trait RowLevelOp {
@@ -30,10 +28,10 @@ sealed trait RowLevelOp {
 
   protected val supportedMergeEngine: Seq[MergeEngine]
 
-  protected val supportedTable: Seq[String]
+  protected val supportAppendOnlyTable: Boolean
 
   def checkValidity(table: Table): Unit = {
-    if (!supportedTable.contains(table.getClass.getSimpleName)) {
+    if (!supportAppendOnlyTable && table.primaryKeys().isEmpty) {
       throw new UnsupportedOperationException(s"Only support to $name table with primary keys.")
     }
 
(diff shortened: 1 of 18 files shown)


paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/catalyst/analysis/RowLevelOp.scala
```diff
@@ -18,28 +18,63 @@
 
 package org.apache.paimon.spark.catalyst.analysis
 
-import org.apache.paimon.CoreOptions.MergeEngine
+import org.apache.paimon.AppendOnlyFileStore
```
JingsongLi, line 21: Unused.

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/catalyst/analysis/RowLevelOp.scala
```diff
@@ -18,28 +18,63 @@
...
-  val supportedMergeEngine: Seq[MergeEngine]
+
+  val name: String = this.getClass.getSimpleName.stripSuffix("$")
+
+  protected val supportedMergeEngine: Seq[MergeEngine]
+
+  protected val supportedTable: Seq[String]
+
+  def checkValidity(table: Table): Unit = {
+    if (!supportedTable.contains(table.getClass.getSimpleName)) {
```
JingsongLi, line 36: Just check it should be a `FileStoreTable`? And with or without primary key? It is better to just rely on public API.

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/catalyst/analysis/RowLevelOp.scala
```diff
@@ -18,28 +18,63 @@
...
+    }
+  }
 }
 
 case object Delete extends RowLevelOp {
-  override def toString: String = "delete"
 
   override val supportedMergeEngine: Seq[MergeEngine] = Seq(MergeEngine.DEDUPLICATE)
+
+  override val supportedTable: Seq[String] = Seq(PrimaryKeyTable)
```
JingsongLi, line 53: `supportAppendTable`?

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/commands/SparkDataFileMeta.scala
```diff
@@ -0,0 +1,70 @@
...
+    partition: BinaryRow,
+    bucket: Int,
+    totalBuckets: Int,
+    dataFileMeta: DataFileMeta
+) {
+
+  def relativePath(fileStorePathFactory: FileStorePathFactory): String = {
+    val partitionPath = fileStorePathFactory.getPartitionString(partition)
+    if (partitionPath.isEmpty) {
+      FileStorePathFactory.BUCKET_PATH_PREFIX + bucket + "/" +
```
JingsongLi, line 38: Use `FileStorePathFactory.bucketPath`.

YannByron: Here, we don't need the root path prefix.

YannByron: Or, I can add a method in `FileStorePathFactory` which return the relative bucket path.

PR #3133, opened 2024-04-01
[flink] Set default recover strategy for TableEnvironment and StreamExecutionEnvironment in tests

tsreaper:
### Purpose

As mentioned in #3121 , Flink 1.19 uses exponential delay as the default restart strategy. It will make failure tests prone to timeout.

This PR specifies the default restart strategy for `TableEnvironment` and `StreamExecutionEnvironment` in tests.

### Tests

This is a refactor for tests. Existing tests should cover the change.

### API and Format

No changes.

### Documentation

Not needed.


JingsongLi (approved):
+1

PR #3139, opened 2024-04-01
[core] Fix that ignore-delete option is not compatible with old delete records and LocalMergeOperator

yuzelin:
### Purpose

Before #3128 , the delete records can be write into data file even when `ignore-delete=true`. So if using jar built after 3128 to read old data, the merge engine might throw exception. This PR fix it by adding filter in reader.

Besides, records passed by `LocalMergeOperator` also should be filtered.

Also remove `changeIgnoreMergeEngine` in merge_into action. Because I think merge_into should also obey the constrains of merge engine.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/io/KeyValueDataFileRecordReader.java
+++ b/paimon-core/src/main/java/org/apache/paimon/io/KeyValueDataFileRecordReader.java
@@ -34,12 +34,18 @@ public class KeyValueDataFileRecordReader implements RecordReader<KeyValue> {
     private final RecordReader<InternalRow> reader;
     private final KeyValueSerializer serializer;
     private final int level;
+    private final boolean ignoreDelete;
 
     public KeyValueDataFileRecordReader(
-            RecordReader<InternalRow> reader, RowType keyType, RowType valueType, int level) {
+            RecordReader<InternalRow> reader,
+            RowType keyType,
+            RowType valueType,
+            int level,
+            boolean ignoreDelete) {
         this.reader = reader;
         this.serializer = new KeyValueSerializer(keyType, valueType);
         this.level = level;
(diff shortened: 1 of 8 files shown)


paimon-core/src/main/java/org/apache/paimon/io/KeyValueDataFileRecordReader.java
```diff
@@ -50,11 +56,15 @@ public RecordIterator<KeyValue> readBatch() throws IOException {
...
-                        internalRow == null
-                                ? null
-                                : serializer.fromRow(internalRow).setLevel(level));
+        RecordIterator<KeyValue> transformed =
+                iterator.transform(
+                        internalRow ->
+                                internalRow == null
+                                        ? null
+                                        : serializer.fromRow(internalRow).setLevel(level));
+        // In older version, the delete records might be written into data file even when
```
JingsongLi, line 65: Document clear: Version 0.7

paimon-core/src/main/java/org/apache/paimon/io/KeyValueFileReaderFactory.java
```diff
@@ -144,7 +144,12 @@ private RecordReader<KeyValue> createRecordReader(
                     new ApplyDeletionVectorReader<>(fileRecordReader, deletionVector.get());
         }
 
-        return new KeyValueDataFileRecordReader(fileRecordReader, keyType, valueType, level);
+        return new KeyValueDataFileRecordReader(
+                fileRecordReader,
+                keyType,
+                valueType,
+                level,
+                CoreOptions.fromMap(schema.options()).ignoreDelete());
```
JingsongLi, line 152: Can be a class member.

JingsongLi:
We must figure out how to deal with `changeIgnoreMergeEngine`.

PR #3144, opened 2024-04-02
[core] Support the rollback and orphan file clean with changelog deco…

Aitozi:
…uple

### Purpose

The third step of #2899

Handle the orphan file cleaner with the changelog metadata

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/OrphanFilesClean.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/OrphanFilesClean.java
@@ -18,8 +18,10 @@
 
 package org.apache.paimon.operation;
 
+import org.apache.paimon.Changelog;
 import org.apache.paimon.FileStore;
 import org.apache.paimon.Snapshot;
+import org.apache.paimon.annotation.VisibleForTesting;
 import org.apache.paimon.fs.FileIO;
 import org.apache.paimon.fs.FileStatus;
 import org.apache.paimon.fs.Path;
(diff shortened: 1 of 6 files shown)


paimon-core/src/main/java/org/apache/paimon/table/AbstractFileStoreTable.java
```diff
@@ -518,9 +519,31 @@ public void rollbackTo(String tagName) {
...
+                Snapshot restored = taggedSnapshot;
+                if (!changelogExists) {
+                    restored =
+                            new Snapshot(
+                                    taggedSnapshot.version(),
+                                    taggedSnapshot.id(),
+                                    taggedSnapshot.schemaId(),
+                                    taggedSnapshot.baseManifestList(),
+                                    taggedSnapshot.deltaManifestList(),
+                                    null,
```
Aitozi, line 532: we need to 'mute' the `changelogManifestList` here, if the changelog and snapshot already not there. Otherwise, the scanner may read on a non-exist manifest list. WDYT ? @JingsongLi

Aitozi: image

It can be reproduce by disable this block and rerun  `testRollbackWithChangelog`

JingsongLi: I see `testRollbackWithChangelog`, but rollbacked table has no changelog, why you want to read changelog?

JingsongLi: Maybe it is a separate issue which is unrelated to changelog decouple.

Aitozi: After discuss with @JingsongLi , it's a separate issue, we could solve it in another pr. So revert this change.

JingsongLi (approved):
+1

PR #3146, opened 2024-04-02
[core] Add compression for ExternalBuffer.

leaves12138:
External buffer spill could add compression.

This will compress bytes and reduce memory copy while read.

Choose lz4, reason:
1. lz4 is faster than zip zstd snappy， write spill need speed
2. lz4 compression ratio is acceptable

see https://indico.fnal.gov/event/16264/contributions/36466/attachments/22610/28037/Zstd__LZ4.pdf

image

image

image


paimon-core/src/main/java/org/apache/paimon/disk/ExternalBuffer.java
```diff
@@ -44,6 +44,9 @@ public class ExternalBuffer implements RowBuffer {
 
     private static final Logger LOG = LoggerFactory.getLogger(ExternalBuffer.class);
 
+    private static final Lz4BlockCompressionFactory compactionFactory =
```
JingsongLi, line 47: You can use `BlockCompressionFactory` to pass `compression` to here.

leaves12138: Resolved

JingsongLi:
+1

PR #3147, opened 2024-04-02
[spark] Minimize the number of data splits need to be scaned when update

YannByron:
### Purpose

Linked issue: close #xxx


paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/catalyst/analysis/expressions/ExpressionHelper.scala
```diff
@@ -138,18 +138,28 @@ trait ExpressionHelper extends PredicateHelper {
...
+          filter
       })
       .toArray
-    val predicates = filters.map(converter.convert)
-    PredicateBuilder.and(predicates: _*)
+
+    if (filters.isEmpty) {
+      None
+    } else {
+      val predicates = filters.map(converter.convert)
```
Zouxxyy, line 160: `converter.convert` will also throw an exception, maybe we can also add ignoreFailure judgment

Zouxxyy (approved):
+1

PR #3152, opened 2024-04-02
Support custom initial flink catalog & env

melin:
### Purpose

Linked issue: close #xxx

ActionBase initializes the catalog and env, and the API calls the Action class, which can be customized for initialization


JingsongLi (approved):
+1

PR #3155, opened 2024-04-03
[core] BranchManager lists branches in the order of create time

yuzelin:
### Purpose

1. TagManager also return tags in order (of snapshot id)
2. Make test stable


zhuangchong (approved):
+1

PR #3156, opened 2024-04-03
[cdc] Calculated columns support constant value.

### LINKED ISSUE #3097: [Feature] Calculated columns support constant value
### Search before asking

### Motivation

paimon table was collected and written into mysql databases of different regions (for example, Hangzhou and Huzhou) using mysql cdc action. The table was partitioned according to days + regions (pt, region), and different action tasks were started for each region.

### Solution

_No response_

### Anything else?

_No response_

zhuangchong:
### Purpose

Linked issue: close #3097


JingsongLi (changes requested):
Thanks @zhuangchong .

Can you investigate other systems?
An intuitive feeling is that we can enrich types, not just strings, for example, strings need to be '11', and 11 is an int.

JingsongLi (approved):
+1

Issue #3157, opened 2024-04-05
[Bug]  query values with the same primary key


zhourui999:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Paimon version

0.8-snapshot

### Compute Engine

spark 3.1
flink-1.17.2

### Minimal reproduce step

A large online table with frequent data writing and updates, with a data volume of approximately 600 million. CDC synchronizes to the payment table, and after running for a period of time, it is found that the data in the table has duplicate primary keys (using SparkSQL and FlinkSQL queries will result in duplicate data). After executing compact, the same primary key will be merged out

![image](https://github.com/apache/paimon/assets/40812600/22b94204-8fe0-4486-91bc-7dc42f9cee70)

### What doesn't meet your expectations?

The query result does not have duplicate primary keys

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

JingsongLi:
Thanks @zhourui999

PR #3158, opened 2024-04-05
[core] Fix duplicate primary keys in query results

zhourui999:
### Purpose

Linked issue: #3157

Because the same builder is used in the for loop of generateSplits in SnapshotReaderImpl, if rawConvertible is true the last time, it will affect the value of the next Datasplit rawFiles. The ultimate result is the need for MergeRedier's Datasplit, as rawFiles were mistakenly set and NoMergeReader was created, resulting in duplicate data during reading

image

image

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/source/snapshot/SnapshotReaderImpl.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/source/snapshot/SnapshotReaderImpl.java
@@ -291,9 +291,10 @@ public class SnapshotReaderImpl implements SnapshotReader {
                 for (SplitGenerator.SplitGroup splitGroup : splitGroups) {
                     List<DataFileMeta> dataFiles = splitGroup.files;
                     builder.withDataFiles(dataFiles);
-                    if (splitGroup.rawConvertible) {
-                        builder.rawFiles(convertToRawFiles(partition, bucket, dataFiles));
-                    }
+                    builder.rawFiles(
+                            splitGroup.rawConvertible
+                                    ? convertToRawFiles(partition, bucket, dataFiles)
+                                    : Collections.emptyList());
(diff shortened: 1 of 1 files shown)


Zouxxyy:
Thanks for pointing it out!

paimon-core/src/main/java/org/apache/paimon/table/source/snapshot/SnapshotReaderImpl.java
```diff
@@ -289,6 +283,12 @@ private List<DataSplit> generateSplits(
...
                 for (SplitGenerator.SplitGroup splitGroup : splitGroups) {
+                    DataSplit.Builder builder =
+                            DataSplit.builder()
+                                    .withSnapshot(snapshotId)
+                                    .withPartition(partition)
+                                    .withBucket(bucket)
+                                    .isStreaming(isStreaming);
                     List<DataFileMeta> dataFiles = splitGroup.files;
                     builder.withDataFiles(dataFiles);
                     if (splitGroup.rawConvertible) {
```
Zouxxyy, line 294: It can also be modified this way. 
I think the current implementation is okay too, what do you think? @JingsongLi 

```
builder.rawFiles(
               splitGroup.rawConvertible
                       ? convertToRawFiles(partition, bucket, dataFiles)
                       : Collections.emptyList());
```

Zouxxyy (approved):

paimon-core/src/main/java/org/apache/paimon/table/source/snapshot/SnapshotReaderImpl.java
zhourui999: Okay,I think your suggestion is better. I'll make some changes

JingsongLi:
+1

PR #3160, opened 2024-04-07
[core] Reduce small files while batch write append only table with a lot of partitions.

leaves12138:
Reduce small files generation in many partitions write job of append only table.


paimon-core/src/main/java/org/apache/paimon/append/AppendOnlyWriter.java
```diff
@@ -209,13 +214,25 @@ public void close() throws Exception {
...
+
+            // rewrite small files
+            try (RecordReaderIterator<InternalRow> reader = bucketFileRead.read(files)) {
+                while (reader.hasNext()) {
+                    sinkWriter.write(reader.next());
+                }
+            }
+
+            // remove small files
+            for (DataFileMeta file : files) {
```
JingsongLi, line 233: move this into try finally?

JingsongLi:
+1

PR #3165, opened 2024-04-08
[Bug] Fix the race condition in OrphanFilesClean

Aitozi:
### Purpose

The `parallelStream` will run in multi thread, it will lead to unstable results of assert delete files number.

This should can resolve the unstable `OrphanFilesCleanTest.testCleanOrphanFilesWithChangelogDecoupled` such as https://github.com/apache/paimon/actions/runs/8587139001/job/23530395569


JingsongLi (approved):
+1

PR #3166, opened 2024-04-08
[Bug] Fix the unstable testRollbackToTagWithChangelogDecoupled

Aitozi:
### Purpose

if the commit times is 5, the changelog-1 is not expired may lead to unexpected file left.

such as
https://github.com/Aitozi/incubator-paimon/actions/runs/8590046982/job/23537660465


JingsongLi:
+1

PR #3168, opened 2024-04-08
[Feature] Support table statistics procedure for paimon

xuzifu666:
Support table statistics procedure for paimon when user need query statistic of paimon table

Linked issue: https://github.com/apache/paimon/issues/3169


xuzifu666:
@JingsongLi PTAL

PR #3170, opened 2024-04-08
[spark] Add paimon prefix to objects under spark sql package

Zouxxyy:
### Purpose

Avoid conflicts with original spark objects


JingsongLi:
+1

PR #3173, opened 2024-04-08
[test] Reduce the number of tests  for BloomFilter64Test.testFunction

Zouxxyy:
### Purpose

see
```
2024-04-08T10:31:53.5273247Z [INFO] Tests run: 10000, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 1,413.912 s - in org.apache.paimon.utils.BloomFilter64Test
```

### DIFF
--- a/paimon-common/src/test/java/org/apache/paimon/utils/BloomFilter64Test.java
+++ b/paimon-common/src/test/java/org/apache/paimon/utils/BloomFilter64Test.java
@@ -19,7 +19,7 @@
 package org.apache.paimon.utils;
 
 import org.assertj.core.api.Assertions;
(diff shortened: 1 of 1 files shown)


paimon-common/src/test/java/org/apache/paimon/utils/BloomFilter64Test.java
```diff
@@ -30,7 +30,7 @@ public class BloomFilter64Test {
 
     private static final Random RANDOM = new Random();
 
-    @RepeatedTest(10000)
+    @RepeatedTest(10)
```
leaves12138, line 33: I think just @Test is ok

PR #3178, opened 2024-04-09
[core] support decouple the delta files lifecycle

Aitozi:
### Purpose

This PR is meant to support decouple the delta files lifecycle #2899 

The basic idea behind this is that:

- Add fileSource in `DatafileMeta` to indicate whether this file is generated as an `APPEND` or `COMPACT` file
- When expire a snapshot, if changelog is decoupled, and it's a none changelog producer, we do not clean the `APPEND` files in data file
- When expire a changelog, we do the final clean for these files
- Due to the manifest file could still be used by delta files, so the `base` and `delta` manifest file for the `none` producer are also postpone to delete

**About why we need `FileSource` in DataFileMeta**

For `none` changelog producer, only `APPEND` commits are required for stream read. In a `COMPACT` commit, some files from the compact or append could be marked as delete. We should delete the files from the compact commit and keep the files from the append commit for further stream read. So we need a flag to distinguish the file source (compact or append).

Linked issue: close #xxx

### API and Format

Introduce `FileSource` in DataFileMeta

### DIFF
--- a/paimon-spark/paimon-spark-common/src/test/scala/org/apache/paimon/spark/procedure/ExpireSnapshotsProcedureTest.scala
+++ b/paimon-spark/paimon-spark-common/src/test/scala/org/apache/paimon/spark/procedure/ExpireSnapshotsProcedureTest.scala
@@ -71,7 +71,8 @@ class ExpireSnapshotsProcedureTest extends PaimonSparkTestBase with StreamTest {
 
             // expire
             checkAnswer(
-              spark.sql("CALL paimon.sys.expire_snapshots(table => 'test.T', retain_max => 2)"),
+              spark.sql(
+                "CALL paimon.sys.expire_snapshots(table => 'test.T', retain_max => 2, retain_min => 1)"),
               Row(1) :: Nil)
 
             checkAnswer(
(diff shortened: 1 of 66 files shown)


JingsongLi:
Perhaps it's best for us to have an abstract mechanism to ensure that we don't keep both the changelog and delta files at the same time, so that we can better understand this set of things.

Aitozi:
> Perhaps it's best for us to have an abstract mechanism to ensure that we don't keep both the changelog and delta files at the same time, so that we can better understand this set of things.

I extract this logic to `SnapshotDeletion` and `ChangelogDeletion` to handle the manifestList and data file clean.

In this PR, I also refactor the `ExpireChangelogImpl` and `ExpireSnapshotImpl` to immutable, So that, we could know whether the changelog decouple is enabled in the expire process.

paimon-spark/paimon-spark-common/src/test/scala/org/apache/paimon/spark/procedure/ExpireSnapshotsProcedureTest.scala
```diff
@@ -71,7 +71,7 @@ class ExpireSnapshotsProcedureTest extends PaimonSparkTestBase with StreamTest {
 
             // expire
             checkAnswer(
-              spark.sql("CALL paimon.sys.expire_snapshots(table => 'test.T', retain_max => 2)"),
+              spark.sql("CALL paimon.sys.expire_snapshots(table => 'test.T', retain_max => 2, retain_min => 1)"),
```
Aitozi: Before, if user not specify the `retain_min`, the default value is 1 in `ExpireSnapshotsImpl`, now the default value is fallback to the `CoreOptions.SNAPSHOT_RETAIN_MIN = 10`, so if max is 2, we should manually specify the `retain_min => 1`. I think the current behavior is more straightforward, I'm not sure whether this will break the compatibility. Please also help check this  cc @JingsongLi

JingsongLi:
Already merged to three PRs.

Issue #2670, opened 2024-01-10
[Feature]  Whole database compression combined mode does not support append unware bucket table


thomasg19930417:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/incubator-paimon/issues) and found nothing similar.

### Motivation

Whole database compression combined mode does not support append unware bucket table。
![f7467db14a7d8d34906f776d5695c4e3](https://github.com/apache/incubator-paimon/assets/20243868/dc60b912-fa0e-4e3e-8712-d55337cfea95)

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

wg1026688210:
assign to me plz

wg1026688210:
Hi~ @thomasg19930417  Can you provide the command ?

thomasg19930417:
> Hi~ @thomasg19930417 Can you provide the command ?

Use the entire database compression command and specify the compression mode --mode combined, and the table is an append table. There are no other special parameters.

PR #2858, opened 2024-02-05
[compact]support compact unaware bucket for combine_mode

wg1026688210:
Refact the compaction of combine mode for supporting compacting unaware bucket table 

### Purpose

Linked issue: close #2670 

### Tests

[CompactDatabaseActionITCase.java](https://github.com/apache/incubator-paimon/pull/2858/files#diff-9eae358dbfd2474f886992026b8840e7a0d2438bef74eae239fa98974462287c) for IT case in stream and batch mode
CompactionTaskSerializerTest.java for the test of serializing compaction task
AppendOnlyMultiTableCompactionWorkerOperatorTest.java for the  test of  AppendOnlyMultiTableCompactionWorkerOperator

PR #3179, opened 2024-04-09
[compaction] Refact the file scan source of combine mode for supporting unaware logic

wg1026688210:
### Purpose
1. reusing the logic of scan table.
2. add scan unaware bucket table logic.

related issue: #2670


wg1026688210:
finished at #2858

PR #3180, opened 2024-04-09
[flink] Balance MonitorFunction data shuffle for partitioned table

wxplovecc:
### Purpose

Balance MonitorFunction data shuffle for partitioned table, because partitioned table with one bucket only shuffle to one subtask

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/operator/MonitorFunction.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/operator/MonitorFunction.java
@@ -19,6 +19,7 @@
 package org.apache.paimon.flink.source.operator;
 
 import org.apache.paimon.flink.utils.JavaTypeInfo;
+import org.apache.paimon.table.sink.ChannelComputer;
 import org.apache.paimon.table.source.DataSplit;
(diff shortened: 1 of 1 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/operator/MonitorFunction.java
```diff
@@ -230,14 +231,21 @@ public static DataStream<RowData> buildSource(
             ReadBuilder readBuilder,
             long monitorInterval,
             boolean emitSnapshotWatermark) {
+        KeySelector<Split, Integer> keySelector =
+                split -> {
+                    if (((DataSplit) split).partition() != null) {
```
JingsongLi, line 236: partition never be null.

JingsongLi, line 236: You can just use `ChannelComputer.select`.

wxplovecc: Updated

JingsongLi:
+1 thanks @wxplovecc

PR #3181, opened 2024-04-09
[flink] Add merge-engine check when executing row level batch update and delete

yuzelin:
### Purpose

1. partial-update and aggregate should not accept batch update because there is no UB and the aggregate result will be wrong.

2. It's not very clear about how to handle row level batch delete with partial update currently. Doris just delete the record. But it's not easy to delete the record because #3128   and  #3139 .

Linked issue: close #xxx

### API and Format

Support batch update merge engines: `deduplicate` and `first-row`.
Support batch delete merge engines: `deduplicate`.

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
@@ -1692,20 +1692,28 @@ public class CoreOptions implements Serializable {
 
     /** Specifies the merge engine for table with primary key. */
     public enum MergeEngine implements DescribedEnum {
-        DEDUPLICATE("deduplicate", "De-duplicate and keep the last row."),
+        DEDUPLICATE("deduplicate", "De-duplicate and keep the last row.", true, true),
 
-        PARTIAL_UPDATE("partial-update", "Partial update non-null fields."),
+        PARTIAL_UPDATE("partial-update", "Partial update non-null fields.", false, false),
 
-        AGGREGATE("aggregation", "Aggregate fields with same primary key."),
+        AGGREGATE("aggregation", "Aggregate fields with same primary key.", false, false),
 
-        FIRST_ROW("first-row", "De-duplicate and keep the first row.");
+        FIRST_ROW("first-row", "De-duplicate and keep the first row.", true, false);
 
         private final String value;
         private final String description;
+        private final boolean supportBatchUpdate;
+        private final boolean supportBatchDelete;
 
-        MergeEngine(String value, String description) {
+        MergeEngine(
+                String value,
+                String description,
+                boolean supportBatchUpdate,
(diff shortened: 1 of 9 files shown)


paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -1692,20 +1692,28 @@ public boolean deletionVectorsEnabled() {
...
+        DEDUPLICATE("deduplicate", "De-duplicate and keep the last row.", true, true),
 
-        PARTIAL_UPDATE("partial-update", "Partial update non-null fields."),
+        PARTIAL_UPDATE("partial-update", "Partial update non-null fields.", false, false),
 
-        AGGREGATE("aggregation", "Aggregate fields with same primary key."),
+        AGGREGATE("aggregation", "Aggregate fields with same primary key.", false, false),
 
-        FIRST_ROW("first-row", "De-duplicate and keep the first row.");
+        FIRST_ROW("first-row", "De-duplicate and keep the first row.", true, false);
```
Zouxxyy, line 1701: first-row does not need to support update

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -1692,20 +1692,28 @@ public boolean deletionVectorsEnabled() {
 
     /** Specifies the merge engine for table with primary key. */
     public enum MergeEngine implements DescribedEnum {
-        DEDUPLICATE("deduplicate", "De-duplicate and keep the last row."),
+        DEDUPLICATE("deduplicate", "De-duplicate and keep the last row.", true, true),
 
-        PARTIAL_UPDATE("partial-update", "Partial update non-null fields."),
+        PARTIAL_UPDATE("partial-update", "Partial update non-null fields.", false, false),
```
Zouxxyy, line 1697: When using partial-update and no col contains agg, update is theoretically supported

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -1692,20 +1692,28 @@ public boolean deletionVectorsEnabled() {
...
+        AGGREGATE("aggregation", "Aggregate fields with same primary key.", false, false),
 
-        FIRST_ROW("first-row", "De-duplicate and keep the first row.");
+        FIRST_ROW("first-row", "De-duplicate and keep the first row.", true, false);
 
         private final String value;
         private final String description;
-
-        MergeEngine(String value, String description) {
+        private final boolean supportBatchUpdate;
```
JingsongLi, line 1705: Please do not modify this, this is public api, we don't need to expose here.

JingsongLi:
+1

PR #3190, opened 2024-04-10
[core] File index format should include all type index and columns.

leaves12138:
Related pr: https://github.com/apache/paimon/pull/3068

New file index format:
image

* REASON
It's better to put all type of index into one file, we don't want many small files.


JingsongLi (approved):
+1

PR #3193, opened 2024-04-10
[core] support PartitionKeyVisitor

YannByron:
### Purpose

Linked issue: close #xxx

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/predicate/PartitionKeyVisitor.java
+++ b/paimon-common/src/main/java/org/apache/paimon/predicate/PartitionKeyVisitor.java
@@ -1,0 +1,200 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


JingsongLi:
+1

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/commands/DeleteFromPaimonTableCommand.scala
```diff
@@ -62,25 +65,26 @@ case class DeleteFromPaimonTableCommand(
...
       }
 
-      // We do not have to scan table if the following three requirements are met:
-      // 1) no other predicate;
-      // 2) partition condition can convert to paimon predicate;
-      // 3) partition predicate can be visit by OnlyPartitionKeyEqualVisitor.
+      // We have to scan table if the following three requirements are met:
+      // 1) non-partitioned table;
+      // 2) no partition predicate;
+      // 3) all the data in candidate partitions should be deleted;
```
JingsongLi, line 85: "all the data in candidate partitions should be deleted" => scan table?

paimon-common/src/main/java/org/apache/paimon/predicate/PartitionKeyVisitor.java
```diff
@@ -0,0 +1,200 @@
...
+
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/** PartitionKeyVisitor. */
+public class PartitionKeyVisitor implements FunctionVisitor<Set<BinaryRow>> {
```
JingsongLi, line 36: What is `PartitionKeyVisitor`? How to use it? Can you add documentation here?

JingsongLi:
If your goal is to optimize the `DELETE` to the `dropPartitions`, we can determine if there are only predicates for partition fields. You can use `PartitionPredicteVisitor`, and then list Partitions, one by one, to test the predicate.

In this way, you don't need to write a `FunctionVisitor` class.

JingsongLi:
Alread fixed in https://github.com/apache/paimon/pull/3263 ? Close this now.

PR #3201, opened 2024-04-11
[flink] support to infer parallelism for system table

Aitozi:
### Purpose

Currently, the `scan.parallelism` and `scan.infer-parallelism` is not work for system table.  We should implement this because we have `ro` / `audit_log` table

Linked issue: close #xxx


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/SystemTableSource.java
```diff
@@ -139,4 +158,34 @@ public void applyDynamicFiltering(List<String> candidateFilterFields) {
     public boolean isStreaming() {
         return isStreamingMode;
     }
+
+    private DataStreamSource<RowData> configSourceParallelism(
```
JingsongLi, line 162: Why not reuse `DataTableSource.configureSource`?

JingsongLi (approved):
+1

PR #3202, opened 2024-04-11
[core] Add ignoreUnknown to Snapshot

JingsongLi:
### Purpose

In order to make the old reader compatible with the new format


wwj6591812:
+1

PR #3205, opened 2024-04-12
[core] Add listDirectories to FileIO

JingsongLi:
### Purpose

To fix https://github.com/apache/paimon-trino/issues/61


leaves12138 (approved):
+1

PR #3207, opened 2024-04-14
[common] Expose columnar data in RowColumnVector/ArrayColumnVector/MapColumnVector

yuzelin:
### Purpose

More API after  #3103


paimon-common/src/main/java/org/apache/paimon/data/columnar/ColumnarArray.java
```diff
@@ -44,6 +44,15 @@ public ColumnarArray(ColumnVector data, int offset, int numElements) {
         this.numElements = numElements;
     }
 
+    public ColumnVector foo() {
```
JingsongLi, line 47: clear this

JingsongLi:
+1

PR #3208, opened 2024-04-15
[core] Unify delete&update&mergeinto for merge engines

JingsongLi:
### Purpose

correct https://github.com/apache/paimon/pull/3181

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/MergeIntoAction.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/MergeIntoAction.java
@@ -45,6 +45,8 @@ import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import static org.apache.paimon.CoreOptions.MergeEngine.DEDUPLICATE;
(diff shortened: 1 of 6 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/MergeIntoAction.java
```diff
@@ -223,20 +225,10 @@ public void validate() {
         }
 
         CoreOptions.MergeEngine mergeEngine = CoreOptions.fromMap(table.options()).mergeEngine();
-        if ((matchedUpsert || notMatchedUpsert) && !mergeEngine.supportBatchUpdate()) {
+        boolean supportMergeInto = mergeEngine == DEDUPLICATE || mergeEngine == PARTIAL_UPDATE;
```
yuzelin, line 228: I think if there is only 'Not Matched Insert' we shouldn't check the merge engine.

JingsongLi: Consistent with Spark Merge Into.

yuzelin (approved):
+1

PR #3209, opened 2024-04-15
[core] Introduce RawFileSplitRead to accelerate batch read for primary key table

JingsongLi:
### Purpose

`KeyValueFileStoreRead` is an implementation for `KeyValueFileStore`, this class handle LSM merging and changelog row kind things, it will force reading fields such as sequence and row_kind. For `COUNT(*)` querying, it is slow.

If in batch mode and reading raw files, it is recommended to use `RawFileSplitRead`.


leaves12138 (approved):
+1

PR #3226, opened 2024-04-17
[core] Expose `IndexFile` in `Split`.

leaves12138:
* REASON
After bloom-filter added to paimon, we should find extra files in `RawFile`.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/source/DataSplit.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/source/DataSplit.java
@@ -207,10 +207,7 @@ public class DataSplit implements Split {
             rawFile.serialize(out);
         }
 
-        out.writeInt(indexFiles.size());
-        for (IndexFile indexFile : indexFiles) {
-            indexFile.serialize(out);
-        }
+        IndexFile.serializeList(out, indexFiles);
     }
 
     public static DataSplit deserialize(DataInputView in) throws IOException {
@@ -243,11 +240,7 @@ public class DataSplit implements Split {
             rawFiles.add(RawFile.deserialize(in));
         }
 
-        int indexFileNum = in.readInt();
-        List<IndexFile> indexFiles = new ArrayList<>();
-        for (int i = 0; i < indexFileNum; i++) {
-            indexFiles.add(IndexFile.deserialize(in));
-        }
+        List<IndexFile> indexFiles = IndexFile.deserializeList(in);
 
         DataSplit.Builder builder =
                 builder()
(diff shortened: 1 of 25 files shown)


JingsongLi:
Maybe just like Deletion Files, expose a method in `Split` class?

JingsongLi:
Let `RawFile` be a real raw file, nothing else.

paimon-core/src/main/java/org/apache/paimon/table/source/IndexFile.java
```diff
@@ -0,0 +1,58 @@
...
+/** Index file for data file. */
+public class IndexFile {
+
+    private final String path;
+
+    public IndexFile(String path) {
+        this.path = path;
+    }
+
+    public void serialize(DataOutputView out) throws IOException {
```
JingsongLi, line 36: Can we just like `DeletionFile`?

paimon-core/src/main/java/org/apache/paimon/table/source/DataSplit.java
```diff
@@ -297,6 +315,11 @@ public Builder rawFiles(List<RawFile> rawFiles) {
             return this;
         }
 
+        public Builder indexFiles(List<IndexFile> indexFiles) {
+            this.split.indexFiles = indexFiles;
```
JingsongLi, line 319: Maybe here we can get index files lazily from datafiles.

paimon-core/src/main/java/org/apache/paimon/table/source/DataSplit.java
```diff
@@ -292,15 +352,27 @@ public Builder isStreaming(boolean isStreaming) {
...
+            this.split.defaultFormat = defaultFormat;
             return this;
         }
 
         public DataSplit build() {
             checkArgument(split.partition != null);
             checkArgument(split.bucket != -1);
             checkArgument(split.dataFiles != null);
+            checkArgument(split.bucketPath != null);
+            checkArgument(split.defaultFormat != null);
```
JingsongLi, line 375: remove this default format. file must be ended with format identifier.

paimon-core/src/main/java/org/apache/paimon/table/source/RawFile.java
```diff
@@ -85,26 +82,6 @@ public long rowCount() {
         return rowCount;
     }
 
-    public void serialize(DataOutputView out) throws IOException {
```
JingsongLi, line 88: keep this method

paimon-core/src/main/java/org/apache/paimon/table/source/IndexFile.java
```diff
@@ -0,0 +1,41 @@
...
+
+/** Index file for data file. */
+public class IndexFile {
+
+    private final String path;
+
+    public IndexFile(String path) {
+        this.path = path;
+    }
+
```
JingsongLi, line 31: add ser and deser

JingsongLi (approved):
+1

PR #3234, opened 2024-04-18
[core] Add numWriters metrics

tsreaper:
### Purpose

This PR adds a `numWriters` metrics to monitor the number of active writers in each parallelism. This metric can help us determine if the memory shortage is caused by too many writers.

### Tests

* `PrimaryKeyWriterOperatorTest#testNumWritersMetric`

### API and Format

No changes in API or format.

### Documentation

Yes. Document is updated.


JingsongLi:
+1

PR #3242, opened 2024-04-19
[spark] Change SQLConfUtils to scala 

Zouxxyy:
### Purpose

`mvn package install` failed for style check for  SQLConfUtils.java:

```java
import org.apache.paimon.catalog.Catalog

import org.apache.spark.sql.internal.SQLConf
```

but mvn spotless:apply did not cache it,  because `SQLConfUtils.java` is a java file but in scala package, may we can enhance the spotless in the future


yuzelin (approved):
+1

PR #3253, opened 2024-04-24
[dep] Remove Hdfs client dependency

JingsongLi:
### Purpose

We don't need to rely on hdfs client jar, we can just load hdfs-default.xml and hdfs-site.xml.


JingsongLi:
ORC depends on Hdfs client, close this PR.

PR #3257, opened 2024-04-24
[core] Introduce ApplyDeletionFileRecordIterator to know underly iterator and dv

JingsongLi:
### Purpose

Engine can know better about paimon underly deletion vectors, and do some optimization.


JingsongLi:
> I think we can delete `reader()` and `deletionVector()` in `ApplyDeletionVectorReader`.

0.8.0 is under released, let's not break this internal API.

yuzelin (approved):
+1

PR #3258, opened 2024-04-25
[core] Optimize partitions table to avoid OOM

JingsongLi:
### Purpose

We don't need to cache all entries in memory, we can just merge partition info.


leaves12138 (approved):
+1

PR #3259, opened 2024-04-25
fix bug of DataFileMeta's to_string()

wwj6591812:
### Purpose

Now if call DataFileMeta's to_string() will report : 
image

so fix it

Linked issue: close #xxx


JingsongLi (approved):
+1

PR #3262, opened 2024-04-25
[core] optimize the first row merge engine

Aitozi:
### Purpose

If we meet high level key during the compact, we could avoid lookup the key in high level in this turn.


JingsongLi (approved):
+1

PR #3263, opened 2024-04-26
[spark] drop partition when all the conditions is partition-related

YannByron:
### Purpose

Linked issue: close #xxx


JingsongLi:
Which Test can cover this modification?

YannByron:
> Which Test can cover this modification?

no new cases. I've tested the case `Paimon Delete: append-only table with partition` in `DeleteFromTableTest` and check the code path when execute.

JingsongLi:
> > Which Test can cover this modification?
> 
> no new cases. I've tested the case `Paimon Delete: append-only table with partition` in `DeleteFromTableTest` and check the code path when execute.

I see, `DELETE FROM T WHERE dt IN ('2026', '2027')` will use partition dropping now.

JingsongLi (approved):
+1

PR #3264, opened 2024-04-26
[core] Add support of sequence field in first row

Aitozi:
### Purpose

This PR is meant to support set `sequence.field` for first row. For example, user may want to keep the first row in Eventtime semantics (Otherwise, the result may not correct). In this mode, we will generate UB message to retract old value.

Also support generating deletion vector for first row with sequence field.

### DIFF
--- a/docs/content/primary-key-table/merge-engine.md
+++ b/docs/content/primary-key-table/merge-engine.md
@@ -322,13 +322,13 @@ is an empty map instead of `+I[1->A]`
 
 ## First Row
 
-By specifying `'merge-engine' = 'first-row'`, users can keep the first row of the same primary key. It differs from the
-`deduplicate` merge engine that in the `first-row` merge engine, it will generate insert only changelog.
+By specifying `'merge-engine' = 'first-row'`, users can keep the first row of the same primary key. You can use the first row merge engine with or without setting `sequence.field`.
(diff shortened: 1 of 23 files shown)


Aitozi:
CC @JingsongLi

Aitozi:
> Hi @Aitozi , thanks for your contribution!
> 
> Semantically speaking, supporting sequence in FIRST-ROW is very good, and it is this semantics.
> 
> However, there are two issues:
> 
> 1. Users will experience a significant performance gap, with significant differences in streaming read, streaming write, and batch reading, which can greatly affect tuning. Therefore, from this perspective, it is better to let them clearly use another approach.
> 2. There are many additional judgments in the code, and the previous assumptions about FISRT-ROW have been broken one by one, which will increase the complexity of the code.
> 
> Here, I recommend another implementation method by introducing an option: `sequence.field.reverse`, flipping the comparison of sequences. The advantage of this is that it is relatively easy to modify. We can remind users of exception when using FIRST-ROW with sequence.
> 
> If there are not many scenes, it is still recommended to use this low-cost approach, but if you think there are many scenes, we can tolerate these changes to FIRST-ROW.

Hi @JingsongLi , thanks for your valuable inputs. `sequence.field.reverse` looks good to me, will try this solution

PR #3267, opened 2024-04-26
[core][spark] Use list partitions with predicate

JingsongLi:
### Purpose

1. Improve PartitionExpire.
2. Improve `DeleteFromPaimonTableCommand`.


YannByron:
+1

PR #3268, opened 2024-04-26
[core] Expose path from IndexFile

leaves12138:
### Purpose

Linked issue: close #xxx


yuzelin (approved):
+1

paimon-core/src/main/java/org/apache/paimon/table/source/IndexFile.java
```diff
@@ -33,6 +33,10 @@ public IndexFile(String path) {
         this.path = path;
     }
 
+    public String getPath() {
```
JingsongLi, line 36: path

PR #3269, opened 2024-04-26
[hive] Reuse table StorageDescriptor info

Askwang:
### Purpose

When we perform alter table operation, `updateHmsTable` method called will lose some StorageDescriptor infos like `bucketCols` variable. It's value changes from empty to null. This will cause an npe exception if we depend on this variable. Actually, there is no need to recreate new StorageDescriptor and SerDeInfo objects when we alter table. Just reuse previous objects and modify them as needed.

call `updateHmsTable` before:

before

call `updateHmsTable` after:

after


JingsongLi:
Thanks @Askwang for the contribution.

Can you add more description? Why must to reuse?

JingsongLi:
+1

PR #3270, opened 2024-04-26
[spark] fix when the delete condition does not match any partitions

YannByron:
### Purpose

Linked issue: close #xxx


JingsongLi (approved):
+1

PR #3281, opened 2024-04-28
[flink][flink-cdc] Add FLOAT type adaptation to HAVE_SCALE_LIST

Lvbm:
### Purpose

Fixed the problem that when using cdc for data synchronization, when the field type of the data source is float, an exception message will appear during synchronization: just like … :java.lang.NumberFormatException: For input string: "10,2"


Lvbm:
@JingsongLi  please review, thanks!

Lvbm:
@JingsongLi  I'm. sorry, I overlooked the getScale method and it has been fixed

PR #3285, opened 2024-04-29
[core] Avoid lookup high level when not contains level0 key

Aitozi:
### Purpose

In this [PR](https://github.com/apache/paimon/pull/2961), the short cut for `!containsLevel0` is removed. I think we still need this to avoid unnecessary lookup

Linked issue: close #xxx


JingsongLi:
If there is no level0 record, so there must be highLevel record, so `lookup` should not be invoked.

Aitozi:
> If there is no level0 record, so there must be highLevel record, so `lookup` should not be invoked.

Yes, you are right 👍🏻

PR #3286, opened 2024-04-29
[core] Fix the merge order in lookup compaction

Aitozi:
### Purpose

```java
mergeFunction2.reset();
if (highLevel != null) {
    mergeFunction2.add(highLevel);
}
candidates.forEach(mergeFunction2::add);
KeyValue result = mergeFunction2.getResult();
```

When we find a high-level record, we now add it to the merge function first. We previously assumed that the high-level record had a smaller sequence number, which is incorrect. To obtain accurate results, we should add the high-level record and candidates in order.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/LookupChangelogMergeFunctionWrapper.java
+++ b/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/LookupChangelogMergeFunctionWrapper.java
@@ -25,11 +25,15 @@ import org.apache.paimon.deletionvectors.DeletionVectorsMaintainer;
 import org.apache.paimon.lookup.LookupStrategy;
 import org.apache.paimon.mergetree.LookupLevels.PositionedKeyValue;
 import org.apache.paimon.types.RowKind;
+import org.apache.paimon.utils.FieldsComparator;
+import org.apache.paimon.utils.UserDefinedSeqComparator;
 
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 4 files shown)


paimon-core/src/main/java/org/apache/paimon/mergetree/compact/LookupChangelogMergeFunctionWrapper.java
```diff
@@ -136,12 +143,7 @@ public ChangelogResult getResult() {
         }
```
JingsongLi, line 143: actually `mergeFunction.candidates()` is the right order? maybe we can use it.

Aitozi: The `mergeFunction.candidates()` are in the correct order, but we need to determine the right position when a new high-level record is found. Currently, it has not been compared using the user-defined sequence field or sequence number.

Aitozi: See this [test](https://github.com/apache/paimon/pull/3286/files#diff-11928292ecbbd8b574d1a2329427e15330667b67d5bb979a151cf4f0843959f1R334). We have higher sequence in the high level value, the L0's seq is lower, if we add the high level value first, the result is wrong

JingsongLi (approved):
+1

PR #3293, opened 2024-05-02
[core] Introduce record-level expire time

JingsongLi:
### Purpose

In compaction, you can configure record-Level expire time to expire records, you should configure:
1. `'record-level.expire-time'`: time retain for records.
2. `'record-level.time-field'`: time field for record level expire, it should be a seconds INT.

Expiration happens in compaction, and there is no strong guarantee to expire records in time.


leaves12138 (approved):
+1

leaves12138:
May cause expired records not correct when select.

JingsongLi:
> May cause expired records not correct when select.

Yes, expiring records may have some unexpected phenomena, but this is inevitable

PR #3294, opened 2024-05-05
[spark] Delete supports all merge engines

JingsongLi:
### Purpose

Just like DELETE append table, we can DELETE compacted partial-update & agg & first-row table using COPY ON WRITE mode too.

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/commands/PaimonCommand.scala
+++ b/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/commands/PaimonCommand.scala
@@ -118,7 +118,7 @@ trait PaimonCommand extends WithFileStoreTable with ExpressionHelper {
     for (split <- candidateDataSplits) {
       if (!split.rawConvertible()) {
         throw new IllegalArgumentException(
-          "Only compacted table can generate touched files, please use 'COMPACT' procedure.");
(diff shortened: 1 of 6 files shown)


paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/commands/PaimonCommand.scala
```diff
@@ -112,6 +114,14 @@ trait PaimonCommand extends WithFileStoreTable with ExpressionHelper {
       sparkSession: SparkSession): Array[String] = {
     import sparkSession.implicits._
 
+    // only raw convertible can generate input_file_name()
+    for (split <- candidateDataSplits) {
+      if (!split.rawConvertible()) {
+        throw new IllegalArgumentException(
+          "Only compacted table can generate touched files, please use 'COMPACT' procedure.");
```
yuzelin, line 121: `please use 'COMPACT' procedure` -> `please use 'COMPACT' procedure before DELETE`

JingsongLi: this method is shared by DELETE and UPDATE.

JingsongLi: add `first`

yuzelin (approved):
+1

PR #3297, opened 2024-05-06
[flink][lookup] Support async refresh of lookup table

### LINKED ISSUE #3295: [Feature] Support async refresh of lookup table
### Search before asking

### Motivation

Currently, the lookup operator checks the time when getting data from lookup table, and reads delta data of table store refreshes to the lookup table, which will block the lookup operation and cause the job to have a jagged lag. This issue aims to add an asynchronous refresh of lookup table to avoid jagged lags in the job.

### Solution

_No response_

### Anything else?

_No response_

FangYongs:
### Purpose

Linked issue: close #3295 

This PR aims to add async refresh of lookup table to avoid lag in job when the lookup operator is blocked by refresh data from table store

### Tests

1. Updated existing test for refresh async in FileStoreLookupFunctionTest
2. Added LookupTableTest#testPKLookupTableRefreshAsync for refresh async

### API and Format

no

### Documentation

no

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/FullCacheLookupTable.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/FullCacheLookupTable.java
@@ -41,6 +41,8 @@ import org.apache.paimon.utils.PartialRow;
 import org.apache.paimon.utils.TypeUtils;
 import org.apache.paimon.utils.UserDefinedSeqComparator;
 
+import org.apache.paimon.shade.guava30.com.google.common.util.concurrent.MoreExecutors;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -123,7 +125,7 @@ public abstract class FullCacheLookupTable implements LookupTable {
                                         String.format(
                                                 "%s-lookup-refresh",
                                                 Thread.currentThread().getName())))
-                        : null;
+                        : MoreExecutors.newDirectExecutorService();
         this.cachedException = new AtomicReference<>();
         this.maxPendingSnapshotCount = options.get(LOOKUP_REFRESH_ASYNC_PENDING_SNAPSHOT_COUNT);
     }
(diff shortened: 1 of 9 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/PrimaryKeyLookupTable.java
```diff
@@ -99,25 +99,35 @@ public void refresh(Iterator<InternalRow> incremental) throws IOException {
         Predicate predicate = projectedPredicate();
         while (incremental.hasNext()) {
             InternalRow row = incremental.next();
-            primaryKeyRow.replaceRow(row);
-            if (userDefinedSeqComparator != null) {
-                InternalRow previous = tableState.get(primaryKeyRow);
-                if (previous != null && userDefinedSeqComparator.compare(previous, row) > 0) {
-                    continue;
+            if (refreshAsync) {
```
liming30, line 102: If this is a common piece of code, I think it could be implemented in `FullCacheLookupTable` and then provide an abstract `refreshRow` interface to subclasses.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/PrimaryKeyLookupTable.java
```diff
@@ -99,25 +99,35 @@ public void refresh(Iterator<InternalRow> incremental) throws IOException {
...
+                synchronized (lock) {
```
liming30, line 103: Is it possible to achieve thread safety in `RocksDBState`, it looks like only the inconsistency may occur when updating the `cache`.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/PrimaryKeyLookupTable.java
FangYongs: DONE

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/PrimaryKeyLookupTable.java
FangYongs: Currently there will be `primaryKeyRow`, `RocksDBSetState` and `RocksDBListState` in `PrimaryKeyLookupTable`,  `SecondaryIndexLookupTable` and `NoPrimaryKeyLookupTable`, they are not thread safe. I think we can keep the lock in `get` method and optimize it when we find that it has become a bottleneck. What do you think?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/FullCacheLookupTable.java
```diff
@@ -91,7 +110,20 @@ public FullCacheLookupTable(Context context) {
...
         this.projectedType = projectedType;
+        this.refreshAsync = options.get(LOOKUP_REFRESH_ASYNC);
+        this.refreshExecutor =
+                this.refreshAsync
+                        ? Executors.newSingleThreadExecutor(
+                                new ExecutorThreadFactory(
+                                        String.format(
+                                                "%s-lookup-refresh",
+                                                Thread.currentThread().getName())))
+                        : null;
```
liming30, line 124: We can use `MoreExecutors.newDirectExecutorService()`, which can unify the code about `refreshExecutor` without having to judge whether it is `null` or `refreshAsync`.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/FullCacheLookupTable.java
```diff
@@ -189,6 +275,9 @@ public Predicate projectedPredicate() {
     @Override
     public void close() throws IOException {
         stateFactory.close();
+        if (refreshExecutor != null) {
+            refreshExecutor.shutdown();
```
liming30, line 279: This prefer to be placed in a `try-finally` block to avoid thread leaks.

liming30 (approved):
+1

PR #3298, opened 2024-05-06
[core]update FileFormatType from Enum to String, to support userdefin…

YanZhangN:
…e FileFormat.

### Purpose
update FileFormatType from Enum to String, to support userdefine fileformat， besides [orc, avro, parquet].

Linked issue: close #xxx

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
@@ -1162,7 +1162,7 @@ public class CoreOptions implements Serializable {
     }
 
     public String formatType() {
-        return options.get(FILE_FORMAT);
+        return options.get(FILE_FORMAT).toLowerCase();
     }
 
     public FileFormat fileFormat() {
@@ -1198,7 +1198,7 @@ public class CoreOptions implements Serializable {
     }
 
     public static FileFormat createFileFormat(Options options, ConfigOption<String> formatOption) {
-        String formatIdentifier = options.get(formatOption).toString();
+        String formatIdentifier = options.get(formatOption);
         return FileFormat.getFileFormat(options, formatIdentifier);
     }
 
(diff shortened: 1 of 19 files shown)


paimon-common/src/main/java/org/apache/paimon/options/FileFormatType.java
```diff
@@ -0,0 +1,63 @@
...
+
+package org.apache.paimon.options;
+
+import org.apache.paimon.options.description.DescribedEnum;
+import org.apache.paimon.options.description.InlineElement;
+
+import static org.apache.paimon.options.description.TextElement.text;
+
+/** Specifies the file format type for store. */
+public class FileFormatType implements DescribedEnum {
```
JingsongLi, line 27: I feel that FileFormatType can be directly removed, just use the "orc" and "avro" strings, which are almost stable to the point where they cannot be modified.

YanZhangN: FileFormatType is removed, use String to represent FileFormatType instead.

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -1157,7 +1161,7 @@ public static Path path(Options options) {
         return new Path(options.get(PATH));
     }
 
-    public FileFormatType formatType() {
+    public String formatType() {
         return options.get(FILE_FORMAT);
```
JingsongLi, line 1165: I think we need to toLowerCase here. Consider:
```
Options options = new Options();
options.setString(CoreOptions.FILE_FORMAT.key(), "AVRO");
System.out.println(options.get(CoreOptions.FILE_FORMAT));
```

Upper case works fine, the change should be compatible to upper case usage.
You can add test in `FileFormatTest`.

YanZhangN: fixed, and and some tests in `FileFormatTest `.

JingsongLi (approved):
+1 Thanks @YanZhangN

PR #3299, opened 2024-05-06
[Flink] Update using partial columns and support aggregation merge-engine for updates.

hzjhjjyy:
### Purpose

FLINK-32001 has been resolved, so partial columns can now be used for updates.

Different merge engines have different requirements(aggregation has also been supported for updates): 
1. deduplicate: only full columns can be used.
2. partial-update: in addition to the update columns, appended columns include primary keys, partition keys, sequence field, columns defining the last_value in aggregation, and the sequence-group for update columns.
3. aggregation: compared to partial-update, it lacks the sequence-group for update columns.

### Tests

BatchUpdateWithPartialColumnsITCase (all aggregations, sequence field/group and changelog have been tested)

### DIFF
--- a/docs/content/flink/sql-write.md
+++ b/docs/content/flink/sql-write.md
@@ -176,8 +176,7 @@ PARTITION (k0 = 0, k1 = 0) SELECT v FROM my_table WHERE false;
 {{< hint info >}}
 Important table properties setting:
 1. Only [primary key table]({{< ref "primary-key-table" >}}) supports this feature.
-2. [MergeEngine]({{< ref "primary-key-table/merge-engine" >}}) needs to be [deduplicate]({{< ref "primary-key-table/merge-engine#deduplicate" >}})
-   or [partial-update]({{< ref "primary-key-table/merge-engine#partial-update" >}}) to support this feature.
+2. Every [MergeEngine]({{< ref "primary-key-table/merge-engine" >}}) supports this feature.
 3. Do not support updating primary keys.
 {{< /hint >}}
 
--- a/docs/content/primary-key-table/merge-engine.md
+++ b/docs/content/primary-key-table/merge-engine.md
@@ -54,7 +54,7 @@ Assuming that the first column is the primary key, the final result would be `<1
 
 {{< hint info >}}
 For streaming queries, `partial-update` merge engine must be used together with `lookup` or `full-compaction`
-[changelog producer]({{< ref "primary-key-table/changelog-producer" >}}). ('input' changelog producer is also supported, but only returns input records.)
+[changelog producer]({{< ref "primary-key-table/changelog-producer" >}}). ('input' changelog producer is also supported, but only returns input records related to the updated columns.)
 {{< /hint >}}
 
 {{< hint info >}}
@@ -325,7 +325,7 @@ Current supported aggregate functions and data types are:
 
 {{< hint info >}}
 For streaming queries, `aggregation` merge engine must be used together with `lookup` or `full-compaction`
-[changelog producer]({{< ref "primary-key-table/changelog-producer" >}}). ('input' changelog producer is also supported, but only returns input records.)
+[changelog producer]({{< ref "primary-key-table/changelog-producer" >}}). ('input' changelog producer is also supported, but only returns input records related to the updated columns.)
 {{< /hint >}}
 
 ### Retract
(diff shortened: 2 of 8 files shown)


paimon-core/src/main/java/org/apache/paimon/mergetree/compact/PartialUpdateMergeFunction.java
```diff
@@ -441,20 +439,6 @@ default Long generateNullable(InternalRow row, int i) {
 
         private static class SequenceGeneratorVisitor extends DataTypeDefaultVisitor<Generator> {
 
-            @Override
```
JingsongLi, line 444: Why remove here?

hzjhjjyy: The doc doesn't mention support for STRING, so I removed it. Otherwise, using STRING might throw errors.
> For fields.<field-name>.sequence-group, valid comparative data types include: DECIMAL, TINYINT, SMALLINT, INTEGER, BIGINT, FLOAT, DOUBLE, DATE, TIME, TIMESTAMP, and TIMESTAMP_LTZ.

JingsongLi: What we need to do is supporting string using `UserDefinedSeqComparator` way.

If you are interested in this, you can do it in a separate PR, please let this PR clean.

hzjhjjyy: ok, the code has been restored. will submit a new pr.

JingsongLi:
Hi @hzjhjjyy , is this come from user requirement? I understand this optimization, but I'm hesitant to move forward with it.

hzjhjjyy:
> Hi @hzjhjjyy , is this come from user requirement? I understand this optimization, but I'm hesitant to move forward with it.

I think this is to improve the efficiency and scope of update. Naturally, I’m also willing to decide how to handle this pr based on your advice.

JingsongLi:
> > Hi @hzjhjjyy , is this come from user requirement? I understand this optimization, but I'm hesitant to move forward with it.
> 
> I think this is to improve the efficiency and scope of update. Naturally, I’m also willing to decide how to handle this pr based on your advice.

Hi @hzjhjjyy  for your inputs:
1. efficiency of update: In updates, the biggest consumption is twofold: first, discovering this data from the file; second, rewriting the file or using MOR technology. The optimization effect of some updates is not significant.
2. scope of update: I get this can support FieldLastValueAgg, but the default is FieldLastNonNullValueAgg.

Considering these two points, and the changes made by this PR to the current topology are not very worthwhile.

hzjhjjyy:
Hi @JingsongLi .
My own understanding of this PR:
1. For Paimon, this pr is indeed no optimization regarding the calculation method for partial update. The optimization focuses only on reducing the fields retrieved and transmitted when rewriting sql for updates to selects on the flink side. Of course, this has some optimization for large tables since updates typically don't involve many fields simultaneously.
2. Currently, updates are only provided for deduplication and partial-update. Considering the similarity between partial update and aggregation in agg functions, support for aggregation has been added. In my description above, I specifically mentioned last_value just because of its uniqueness, hence the separate mention of its special treatment in this pr. Last_non_null_value can be implemented without special treatment.
Overall, this pr aims to support the feature of using partial columns in updates in flink. I wonder if my explanation clarifies and captures your intent?

JingsongLi:
> Hi @JingsongLi . My own understanding of this pr:
> 
> 1. For Paimon, this pr is indeed no optimization regarding the calculation method for partial update. The optimization focuses only on reducing the fields retrieved and transmitted when rewriting sql for updates to selects on the flink side. Of course, this has some optimization for large tables since updates typically don't involve many fields simultaneously.
> 2. Currently, updates are only provided for deduplication and partial-update. Considering the similarity between partial update and aggregation in agg functions, support for aggregation has been added (otherwise, full-field updates wouldn't be supported). In my description above, I specifically mentioned last_value just because of its uniqueness, hence the separate mention of its special treatment in this pr. Last_non_null_value can be implemented without special treatment.
> 
> Overall, this pr aims to support the feature of using partial columns in updates in flink. I wonder if my explanation clarifies and captures your intent?

Yes, I got your point, but my point is just "Is our modification worth it", we can wait for these requirements to emerge.

JingsongLi:
Close this now to wait for more requirements.

PR #3300, opened 2024-05-06
[Flink] Introduce new parameters to control whether the committer operator and writer operator are chained together

### LINKED ISSUE #3232: [Bug] After the flink job written to Paimon modifies `sink.parallelism`, the job will not be able to recover from the checkpoint
### Search before asking

### Paimon version

master

### Compute Engine

flink

### Minimal reproduce step

1. First set the parallelism degree of 1 to write Paimon

2. Stop passing the current task

3. Modify the parallelism to be greater than 1

4. Restore from the last checkpoint

The job should be able to resume normally from the last checkpoint or savepoint, even if I change the parallelism of the sink

### Anything else?

_No response_

huyuanfeng2018:
### Purpose
Introduce new parameters to control whether the committer operator and writer operator are chained together

Linked issue: close #3232

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/MultiTablesCompactorSink.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/MultiTablesCompactorSink.java
@@ -35,6 +35,7 @@ import org.apache.flink.streaming.api.environment.CheckpointConfig;
 import org.apache.flink.streaming.api.environment.ExecutionCheckpointingOptions;
 import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
 import org.apache.flink.streaming.api.functions.sink.DiscardingSink;
+import org.apache.flink.streaming.api.operators.ChainingStrategy;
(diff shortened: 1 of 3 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/MultiTablesCompactorSink.java
```diff
@@ -121,7 +122,9 @@ protected DataStreamSink<?> doCommit(
         if (streamingCheckpointEnabled) {
             assertStreamingConfiguration(env);
         }
-
+        if (!options.get(SINK_COMMITTER_OPERATOR_CHAINING)) {
```
JingsongLi, line 125: It is better to pass this bool to `CommitterOperator`. It can set `setChainingStrategy` to `ChainingStrategy.NEVER`.

JingsongLi (changes requested):
Thanks @huyuanfeng2018 , left one comment.

huyuanfeng2018:
> Thanks @huyuanfeng2018 , left one comment.

Thank you for your suggestion,
 PTAL.

JingsongLi (approved):
+1

PR #3302, opened 2024-05-07
[core] declare InternalRow type definitely instead of Pattern <T> in classes related to DV

YannByron:
### Purpose

Linked issue: close #xxx


JingsongLi:
+1

PR #3307, opened 2024-05-08
[spark] supports metadata columns

YannByron:
### Purpose

Linked issue: close #xxx

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonRecordReaderIterator.scala
+++ b/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonRecordReaderIterator.scala
@@ -97,7 +97,11 @@ case class PaimonRecordReaderIterator(
         while (!stop) {
           val dataRow = currentIterator.next()
           if (dataRow != null) {
-            if (needMetadata && isFileRecordIterator) {
+            if (needMetadata) {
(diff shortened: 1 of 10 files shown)


paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonRecordReaderIterator.scala
```diff
@@ -84,7 +95,18 @@ case class PaimonRecordReaderIterator(reader: RecordReader[PaimonInternalRow])
...
         var stop = false
         while (!stop) {
-          currentResult = currentIterator.next
+          val dataRow = currentIterator.next()
+          if (dataRow != null) {
+            if (needMetadata && isFileRecordIterator) {
+              updateMetadataRow(currentIterator.asInstanceOf[FileRecordIterator[PaimonInternalRow]])
+              currentResult = joinedRow.replace(dataRow, metadataRow)
+            } else {
+              currentResult = dataRow
```
JingsongLi, line 104: needMetadata & is not FileRecordIterator
Here should throw exception?

YannByron: I prefer to return null instead of throwing exception.

JingsongLi (approved):
+1

PR #3311, opened 2024-05-09
[Flink]supported schemaEvolution when restarting the paimon cdc job

MOBIN-F:
### Purpose

Restarting the flink-paimon-mysql-cdc job schemaCompatible failed when the data type length was modified between the stop and restart of the flink-paimon-mysql-cdc job


MOBIN-F:
I added some test cases, can you help review this pr? @JingsongLi

JingsongLi:
Test Triggerred.

MOBIN-F:
The pr will add the following jobs:
When restarting the paimon job, if the field length changes, the schemaCompatible check is successful and schemaEvolution is performed at the same time.

paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/action/cdc/postgres/PostgresRecordParser.java
```diff
@@ -225,24 +225,15 @@ private DataType extractFieldType(
...
+                    return DataTypes.VARCHAR(Math.max(oldLength, newLength));
                 } else if (paimonField.type() instanceof CharType) {
                     int oldLength = ((CharType) paimonField.type()).getLength();
-                    int newLength = afterData.get(field.field()).asText().length();
-                    if (oldLength < newLength) {
-                        return DataTypes.CHAR(newLength);
-                    } else {
-                        return DataTypes.CHAR(oldLength);
-                    }
+                    return DataTypes.CHAR(Math.max(oldLength, newLength));
```
MOBIN-F: in case "string", We cannot convert the varchar(n)/char(n) type into the string type, which will cause schemaCompatible to fail if the field changes when restarting the paimon-cdc job, because the string length of paimon is integer.max is always greater than n.
At the same time, when pg table performs column type alter(example: n changes from 10 to 20), because pg-debezium does not have ddl event, we can only get the length value n from the record

paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/action/cdc/CdcActionCommonUtils.java
```diff
@@ -83,8 +83,10 @@ public static boolean schemaCompatible(
         for (DataField field : sourceTableFields) {
             int idx = paimonSchema.fieldNames().indexOf(field.name());
             if (idx < 0) {
-                LOG.info("Cannot find field '{}' in Paimon table.", field.name());
-                return false;
+                LOG.info(
+                        "New fields '{}' found in source table, will be synchronized to Paimon table.",
+                        field.name());
+                return true;
             }
```
MOBIN-F, line 90: When idx is less than 0, it means that a new field has been added. Should schemaEvolution be performed? what you think?
@yuzelin @JingsongLi @zhuangchong

MOBIN-F:
@JingsongLi @yuzelin @zhongyujiang I updated the code and test cases, can you help review them?

MOBIN-F:
I will close the PR and create a new PR for better review of the code

MOBIN-F:
move to #3362

PR #3314, opened 2024-05-10
[cdc]Fixed When the order of the same field differs, it is considered a schema change.

chenxi0599:
… a schema change.

### Purpose

When the order of the same field varies, it is regarded as a schema change. Consequently, the Schema Evolution node receives a large volume of records, impacting performance.

So when comparing, it's necessary to ignore the ID of the field.

Linked issue: close #xxx


JingsongLi:
+1

PR #3317, opened 2024-05-10
[flink] Introduce partition mark done

JingsongLi:
### Purpose

This PR is intended to support a mechanism for marking a partition as finished and subsequently pulling up offline scheduling jobs.

Unlike Flink's Filesystem (which came from my younger self), this uses a "mark partition as finished if there is no new data for a certain amount of time" approach, which is currently only supported for system time, but can be extended to include a watermark approach.

Another question is how to mark the end of the partition, there are several options:
1. create a xx.done partition
2. create a _SUCCESS file
3. or other

### Documentation

For partitioned tables, each partition may need to be scheduled to trigger downstream batch computation. Therefore,
it is necessary to choose this timing to indicate that it is ready for scheduling and to minimize the amount of data
drift during scheduling. We call this process: "Partition Mark Done".

Example to mark done:
```sql
CREATE TABLE my_partitioned_table (
    f0 INT,
    f1 INT,
    f2 INT,
    ...
    dt STRING
) PARTITIONED BY (dt) WITH (
    'partition.timestamp-formatter'='yyyyMMdd',
    'partition.timestamp-pattern'='$dt',
    'partition.time-interval'='1 d',
    'partition.idle-time-to-done'='15 m'
);
```

1. Firstly, you need to define the time parser of the partition and the time interval between partitions in order to
   determine when the partition can be properly marked done.
2. Secondly, you need to define idle-time, which determines how long it takes for the partition to have no new data,
   and then it will be marked as done.
3. Thirdly, by default, partition mark done will create _SUCCESS file, the content of _SUCCESS file is a json, contains
   `creationTime` and `modificationTime`, they can help you understand if there is any delayed data. You can also
   configure other actions.

### DIFF
--- a/docs/content/flink/sql-write.md
+++ b/docs/content/flink/sql-write.md
@@ -206,3 +206,33 @@ DELETE FROM my_table WHERE currency = 'UNKNOWN';
 {{< /tab >}}
 
 {{< /tabs >}}
+
+## Partition Mark Done
+
+For partitioned tables, each partition may need to be scheduled to trigger downstream batch computation. Therefore,
+it is necessary to choose this timing to indicate that it is ready for scheduling and to minimize the amount of data
+drift during scheduling. We call this process: "Partition Mark Done".
+
+Example to mark done:
+```sql
+CREATE TABLE my_partitioned_table (
+    f0 INT,
+    f1 INT,
+    f2 INT,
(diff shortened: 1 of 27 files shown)


yuzelin (approved):
+1

PR #3318, opened 2024-05-10
[doc] Add more previous version docs

Zouxxyy:
### Purpose

After:

Only show first 5 versions in `front page`, for the current config`range $docs := first 5 .Site.Params.PreviousDocs`

image

In `all versions` page we can see all version's docs:

image


JingsongLi:
+1

PR #3320, opened 2024-05-11
[core] Forbidden no bucket key for bucketed append table

JingsongLi:
### Purpose

In the previous behavior, no bucket-key would take the whole line and make it bucket-key, which is a poor performance behavior and we should disable it.

And if there is no bucket-key, why not just using bucket -1 mode? Fixed bucket will bring more performance tuning problem.


leaves12138 (approved):
+1

PR #3322, opened 2024-05-13
Modification of Doris and Trino parts

911432:
The Apache Doris link was updated and the Trino part was modified.

### Purpose

Linked issue: close #xxx


JingsongLi (commented):
+1 thanks @911432

PR #3326, opened 2024-05-13
[spark] Add check for provider when creating table with SparkCatalog

Zouxxyy:
### Purpose

For SparkCatalog, we can only create paimon table now:
`CREATE TABLE paimon_tbl1 (id int) USING paimon`
`CREATE TABLE paimon_tbl1 (id int)`

For other provide we need throw exception
`CREATE TABLE parquet_tbl (id int) USING parquet`
throw: SparkCatalog can only create paimon table, but current provider is parquet


JingsongLi (approved):

JingsongLi:
+1

PR #3331, opened 2024-05-14
[core] Exclude watermark and logOffsets for normal table

JingsongLi:
### Purpose

We don't need to save these two fields in the snapshot file when they're not needed.


JingsongLi:
`FileStoreCommitImpl` must use logOffsets... There will be NPE.

PR #3332, opened 2024-05-14
[spark] Fix can not set bucket number in HiveMigrator

xuzifu666:
Fix can not set bucket number in HiveMigrator，currently in MigrateTableProcedure can only set bucket = -1, user cannot set bucekts，the pr is aim to fix it.

Linked issue: close #xxx


JingsongLi:
@xuzifu666 

> Spark load to paimon table is more fast with setting buckets

What is this come from? Do you have some reproduce steps?

xuzifu666:
> @xuzifu666 
> 
> > Spark load to paimon table is more fast with setting buckets
> 
> What is this come from? Do you have some reproduce steps?
This is in our test spark data to paimon,if bucket is -1,would had a bucket assgine, so user would set bucket number avoid bucket assgine, mainly in the scene.@JingsongLi

xuzifu666:
Keep the default is better，so close the pr first

PR #3339, opened 2024-05-16
[core] Optimize snapshots table when specific snapshot id

JingsongLi:
### Purpose

1. Extract a class: LeafPredicateExtractor
2. Handle push down for snapshots table.


leaves12138 (approved):
+1

PR #3347, opened 2024-05-20
[spark] Fix drop database cascade when database with view with SparkGenericCatalog

Zouxxyy:
### Purpose

Becase `SparkGenericCatalog` doest not implement `ViewCatalog` currently, `listTables` will return views too, so dropTable(view) will raise exception, we need to fallback to use `sessionCatalog` to drop


JingsongLi (approved):
+1

PR #2754, opened 2024-01-22
[core][branch] Add branches in tags table

FangYongs:
### Purpose

Linked issue: close #2753 

Add branches column in tags table

### Tests

Added test in `TagsTableTest` to get branches

### API and Format

no

### Documentation

yes

PR #3350, opened 2024-05-20
[core] Revert branchs in TagsTable

JingsongLi:
### Purpose

Revert #2754

`TagsTableTest.testTagBranchesTable` is an unstable case where the order of branches is uncertain.

The underlying reason is that we already have branches table, which should not be recalculated in the Tags table, forming redundant information.


FangYongs:
Thanks @JingsongLi , +1

PR #3362, opened 2024-05-21
[cdc]supported schemaEvolution when restarting the paimon cdc job

MOBIN-F:
### Purpose

Restarting the flink-paimon-cdc job schemaCompatible failed when the column change/add between the stop and restart of the flink-paimon-cdc job


paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/action/cdc/postgres/PostgresRecordParser.java
```diff
@@ -209,6 +226,18 @@ private DataType extractFieldType(DebeziumEvent.Field field) {
...
+                    return DataTypes.VARCHAR(newLength == 0 ? MIN_LENGTH : newLength);
+                } else if (paimonField.type() instanceof VarCharType) {
+                    int oldLength = ((VarCharType) paimonField.type()).getLength();
+                    return DataTypes.VARCHAR(
+                            Math.max(oldLength, newLength == 0 ? MIN_LENGTH : newLength));
+                } else if (paimonField.type() instanceof CharType) {
+                    int oldLength = ((CharType) paimonField.type()).getLength();
+                    return DataTypes.CHAR(
+                            Math.max(oldLength, newLength == 0 ? MIN_LENGTH : newLength));
+                }
```
MOBIN-F, line 240: in case "string", We cannot convert the varchar(n)/char(n) type into the string type, which will cause schemaCompatible to fail if the field changes when restarting the paimon-cdc job, because the string length of paimon is integer.max is always greater than n.
At the same time, when pg table performs column type alter(example: n changes from 10 to 20), because pg-debezium does not have ddl event, we can only get the length value n from the record

MOBIN-F:
can you help review this pr? tks @JingsongLi

JingsongLi:
CC @yuzelin

yuzelin:
Hello, can you solve the conflicts ?

MOBIN-F:
> Hello, can you solve the conflicts ?

done,please help to review this PR, thanks. @yuzelin

PR #3363, opened 2024-05-21
[core] Introduce FileSource to DataFileMeta

Aitozi:
### Purpose

This PR is the first step of the https://github.com/apache/paimon/pull/3178. Introduce the FileSource to DataFileMeta

### API and Format

Yes, the DataFileMeta is changed, but it will keep compatible with default null value.


paimon-core/src/main/java/org/apache/paimon/io/KeyValueDataFileWriter.java
```diff
@@ -80,7 +82,8 @@ public KeyValueDataFileWriter(
             long schemaId,
             int level,
             String compression,
-            CoreOptions options) {
+            CoreOptions options,
+            boolean isCompact) {
```
JingsongLi, line 86: Maybe just pass `FileSource`? For all places in this PR.

Aitozi: Fixed

JingsongLi:
+1

PR #3368, opened 2024-05-21
[core] introduce deletedIndexFiles in IndexIncrement and allow multi …

YannByron:
…index files in the same partition and bucket

### Purpose

Linked issue: close #xxx

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/deletionvectors/DeletionVectorsIndexFile.java
+++ b/paimon-core/src/main/java/org/apache/paimon/deletionvectors/DeletionVectorsIndexFile.java
@@ -22,6 +22,7 @@ import org.apache.paimon.fs.FileIO;
 import org.apache.paimon.fs.Path;
 import org.apache.paimon.fs.SeekableInputStream;
 import org.apache.paimon.index.IndexFile;
+import org.apache.paimon.index.IndexFileMeta;
 import org.apache.paimon.utils.Pair;
 import org.apache.paimon.utils.PathFactory;
 
@@ -30,6 +31,7 @@ import java.io.DataOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.UncheckedIOException;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.Map;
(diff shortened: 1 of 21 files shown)


paimon-core/src/main/java/org/apache/paimon/manifest/IndexManifestFileHandler.java
```diff
@@ -0,0 +1,150 @@
...
+                    indexEntries.put(entry.indexFile().fileName(), entry);
+                } else {
+                    indexEntries.remove(entry.indexFile().fileName());
+                }
+            }
+            return new ArrayList<>(indexEntries.values());
+        }
+    }
+
+    /** We combine the previous and new index files by {@link IndexManifestEntry#identifier}. */
```
JingsongLi, line 128: Just move `IndexManifestEntry#identifier` to here? Actually it is not identifier now. It is only for non-UnawareBucket table.

paimon-core/src/main/java/org/apache/paimon/manifest/IndexManifestFileHandler.java
```diff
@@ -0,0 +1,150 @@
...
+    static class CommonBucketCombiner implements IndexManifestFileCombiner {
+
+        @Override
+        public List<IndexManifestEntry> combine(
+                List<IndexManifestEntry> prevIndexFiles, List<IndexManifestEntry> newIndexFiles) {
+            Map<IndexManifestEntry.Identifier, IndexManifestEntry> indexEntries =
+                    prevIndexFiles.stream()
+                            .filter(entry -> entry.kind() == FileKind.ADD)
```
JingsongLi, line 136: prevIndexFiles should has not delete files. here just checkArguement?
(maybe you can assert outside)

paimon-core/src/main/java/org/apache/paimon/manifest/IndexManifestFileHandler.java
```diff
@@ -0,0 +1,150 @@
...
+     * with UnawareBucket.
+     */
+    static class UnawareBucketCombiner implements IndexManifestFileCombiner {
+
+        @Override
+        public List<IndexManifestEntry> combine(
+                List<IndexManifestEntry> prevIndexFiles, List<IndexManifestEntry> newIndexFiles) {
+            Map<String, IndexManifestEntry> indexEntries = new LinkedHashMap<>();
+            for (IndexManifestEntry entry : prevIndexFiles) {
+                if (entry.kind() == FileKind.ADD) {
```
JingsongLi, line 112: prevIndexFiles should has not delete files. here just checkArguement?
(maybe you can assert outside)

paimon-core/src/main/java/org/apache/paimon/manifest/IndexManifestFileHandler.java
```diff
@@ -0,0 +1,150 @@
...
+                    indexEntries.put(entry.indexFile().fileName(), entry);
```
JingsongLi, line 113: Why not using `deletionVectorsRanges`?

I think fileName is wrong key.

paimon-core/src/main/java/org/apache/paimon/deletionvectors/DeletionVectorsIndexFile.java
```diff
@@ -48,18 +50,20 @@ public DeletionVectorsIndexFile(FileIO fileIO, PathFactory pathFactory) {
...
      * @return A map where the key represents which file the DeletionVector belongs to, and the
      *     value is the corresponding DeletionVector object.
      * @throws UncheckedIOException If an I/O error occurs while reading from the file.
      */
-    public Map<String, DeletionVector> readAllDeletionVectors(
-            String fileName, LinkedHashMap<String, Pair<Integer, Integer>> deletionVectorRanges) {
+    public Map<String, DeletionVector> readAllDeletionVectors(IndexFileMeta fileMeta) {
+        LinkedHashMap<String, Pair<Integer, Integer>> deletionVectorRanges =
+                fileMeta.deletionVectorsRanges();
+        if (deletionVectorRanges == null || deletionVectorRanges.isEmpty()) {
```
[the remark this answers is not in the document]
YannByron: yep, this IndexFileMeta must be Deletion_Vector Index, and `ranges` is not empty. Can remove this part.

paimon-core/src/main/java/org/apache/paimon/manifest/IndexManifestFileHandler.java
YannByron: yes.  all the `prevIndexFiles` should be FileKind.ADD.

paimon-core/src/main/java/org/apache/paimon/manifest/IndexManifestFileHandler.java
YannByron: It can be used in both Hash and Deletion_Vector index files.

paimon-core/src/main/java/org/apache/paimon/manifest/IndexManifestFileHandler.java
YannByron: this method can remove here.

paimon-core/src/main/java/org/apache/paimon/manifest/IndexManifestFileHandler.java
YannByron: ok.

JingsongLi:
+1

PR #3374, opened 2024-05-22
[core] Support default aggregate function for partial update and aggregate merge function

### LINKED ISSUE #3373: [Feature] Support default aggregate function for partial update
### Search before asking

### Paimon version

Support default aggregate such as last_non_value for partial update with and without sequence group and users can update non-null value with sequence group

### Compute Engine

all

### Minimal reproduce step

Partial update with null value

The null value should not be updated in the old record

### Anything else?

_No response_

FangYongs:
Fix update null value for partial update with sequence-group

### Purpose

Linked issue: close #3373

### Tests

1. Added test unit PartialUpdateMergeFunctionTest#testSequenceGroupDefaultAggFunc
2. Added test unit PartialUpdateITCase#testSequenceGroupWithDefaultAggFunc and PartialUpdateITCase#testPartialUpdateWithDefaultAndFieldAggregation

### API and Format

no

### Documentation

no

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/PartialUpdateMergeFunction.java
+++ b/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/PartialUpdateMergeFunction.java
@@ -361,6 +361,7 @@ public class PartialUpdateMergeFunction implements MergeFunction<KeyValue> {
             List<String> fieldNames = rowType.getFieldNames();
             List<DataType> fieldTypes = rowType.getFieldTypes();
             Map<Integer, FieldAggregator> fieldAggregators = new HashMap<>();
+            String defaultAggFunc = options.fieldsDefaultFunc();
             for (int i = 0; i < fieldNames.size(); i++) {
                 String fieldName = fieldNames.get(i);
                 DataType fieldType = fieldTypes.get(i);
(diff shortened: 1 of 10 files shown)


JingsongLi:
See https://paimon.apache.org/docs/master/primary-key-table/merge-engine/#sequence-group

So we introduce sequence group mechanism for partial-update tables. It can solve:
- Disorder during multi-stream update. Each stream defines its own sequence-groups.
- A true partial-update, not just a non-null update.

True partial-update, means null can update column.

FangYongs:
@JingsongLi As we discussed, I add `fields.default-aggregate-function` for partial update with sequence group, please help to review this PR, thanks cc @liming30

paimon-core/src/main/java/org/apache/paimon/mergetree/compact/PartialUpdateMergeFunction.java
```diff
@@ -379,6 +380,16 @@ private Map<Integer, FieldAggregator> createFieldAggregators(
                                     isPrimaryKey,
                                     options,
                                     fieldName));
+                } else if (defaultAggFunc != null) {
```
JingsongLi, line 383: Can you also modify `AggregateMergeFunction`?

FangYongs: @JingsongLi `AggregatedMergeFunction` is updated too

liming30 (approved):
+1，after adding this, users can avoid manually configuring `agg` functions one by one.

JingsongLi (approved):
+1

PR #3377, opened 2024-05-22
[flink] Adjust the setMaxParallelism api call

Aitozi:
### Purpose

`DataStreamSink#setMaxParallelism` is only available in Flink 1.18+. So for compatibility, we can replace with 

```
sink.getTransformation().setMaxParallelism(1);
```


JingsongLi (approved):
+1

PR #3379, opened 2024-05-22
[format] Fix that when ORC batch contains all-null complex ColumnVector, the nested column vector of it cannot do null check

yuzelin:
### Purpose

1. In some cases, we need to check null of the nested ColumnVector of the complex vector.
2. If the complex vector is all null, null check the nested ColumnVector is wrong.


JingsongLi (changes requested):
-1 Please use ITCase to re-produce this issue.

PR #3383, opened 2024-05-22
[core] Introduce ChangelogDeletion to Expiration

Aitozi:
### Purpose

This PR is the second step of https://github.com/apache/paimon/pull/3178 to simply the clean up logic of snapshot and changelog

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/FileDeletionBase.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/FileDeletionBase.java
@@ -25,6 +25,7 @@ import org.apache.paimon.fs.Path;
 import org.apache.paimon.index.IndexFileHandler;
 import org.apache.paimon.index.IndexFileMeta;
 import org.apache.paimon.manifest.FileEntry;
+import org.apache.paimon.manifest.FileKind;
 import org.apache.paimon.manifest.IndexManifestEntry;
 import org.apache.paimon.manifest.ManifestEntry;
 import org.apache.paimon.manifest.ManifestFile;
@@ -33,6 +34,8 @@ import org.apache.paimon.manifest.ManifestList;
 import org.apache.paimon.stats.StatsFileHandler;
 import org.apache.paimon.utils.FileStorePathFactory;
 import org.apache.paimon.utils.FileUtils;
+import org.apache.paimon.utils.Pair;
+import org.apache.paimon.utils.TagManager;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -57,7 +60,7 @@ import java.util.stream.Collectors;
  * Base class for file deletion including methods for clean data files, manifest files and empty
  * data directories.
  */
-public abstract class FileDeletionBase {
+public abstract class FileDeletionBase<T extends Snapshot> {
 
     private static final Logger LOG = LoggerFactory.getLogger(FileDeletionBase.class);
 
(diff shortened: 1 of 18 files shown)


paimon-core/src/main/java/org/apache/paimon/table/AbstractFileStoreTable.java
```diff
@@ -309,23 +308,14 @@ public TableCommitImpl newCommit(String commitUser) {
         if (!options.writeOnly()) {
             boolean changelogDecoupled = options.changelogLifecycleDecoupled();
             ExpireSnapshots expireChangelog =
-                    newExpireChangelog()
-                            .maxDeletes(options.snapshotExpireLimit())
-                            .retainMin(options.changelogNumRetainMin())
-                            .retainMax(options.changelogNumRetainMax());
+                    newExpireChangelog().config(options.expireConfig().build());
```
JingsongLi, line 311: reuse `options.expireConfig().build()`.

paimon-core/src/main/java/org/apache/paimon/operation/FileDeletionBase.java
```diff
@@ -70,6 +73,13 @@ public abstract class FileDeletionBase {
 
     protected final Map<BinaryRow, Set<Integer>> deletionBuckets;
     protected final Executor ioExecutor;
+    protected boolean changelogDecoupled;
```
JingsongLi, line 76: changelogDecoupled who use this?

paimon-core/src/main/java/org/apache/paimon/operation/FileDeletionBase.java
```diff
@@ -70,6 +73,13 @@ public abstract class FileDeletionBase {
...
+
+    /** Used to record which tag is cached in tagged snapshots list. */
+    private int cachedTagIndex = -1;
+
+    /** Used to cache data files used by current tag. */
+    private final Map<BinaryRow, Map<Integer, Set<String>>> cachedTagDataFiles = new HashMap<>();
```
JingsongLi, line 82: maybe we should introduce a `ExpireDeletion`? `TagDeletion` will not use this field.

paimon-core/src/main/java/org/apache/paimon/operation/FileDeletionBase.java
Aitozi: For example, `SnapshotDeletion` will use this to determine whether to clean changelog manifests. It will also be used in `ChangelogDeletion` in following PR.

JingsongLi: but there is no assignment for it.

maybe you need to add it in next pr.

Aitozi: There is `setChangelogDecoupled` . I feel difficult to make this field `final`, so I add a set method for it.

paimon-core/src/main/java/org/apache/paimon/operation/FileDeletionBase.java
Aitozi: It only be cached when accessed, So in `TagDeletion` it will not be populated

paimon-core/src/main/java/org/apache/paimon/operation/TagDeletion.java
```diff
@@ -88,15 +91,21 @@ public void cleanUnusedManifests(Snapshot taggedSnapshot, Set<String> skippingSe
         cleanUnusedManifests(taggedSnapshot, skippingSet, false);
     }
 
-    public Predicate<ManifestEntry> dataFileSkipper(Snapshot fromSnapshot) throws Exception {
-        return dataFileSkipper(Collections.singletonList(fromSnapshot));
-    }
+    protected Collection<ManifestEntry> readMergedDataFiles(Snapshot snapshot) throws IOException {
```
JingsongLi, line 94: Duplicate codes?

paimon-core/src/main/java/org/apache/paimon/operation/FileDeletionBase.java
JingsongLi: My bad. TagDeletion will use skipper too.

paimon-core/src/main/java/org/apache/paimon/operation/TagDeletion.java
Aitozi: The latest commit is reverted, please take a look again when you are free.

JingsongLi (approved):
+1

PR #3387, opened 2024-05-23
[core] Introduce DeletionVectorIndexFileMaintainer

YannByron:
### Purpose

Linked issue: close #xxx


paimon-core/src/main/java/org/apache/paimon/deletionvectors/DeletionVectorsIndexFile.java
```diff
@@ -83,6 +84,29 @@ public Map<String, DeletionVector> readAllDeletionVectors(IndexFileMeta fileMeta
...
+        Map<String, DeletionVector> deletionVectors = new HashMap<>();
+        if (ddFiles.isEmpty()) {
+            return deletionVectors;
+        }
+
+        String indexFile = ddFiles.get(0).deletionFile().path();
+        try (SeekableInputStream inputStream = fileIO.newInputStream(new Path(indexFile))) {
+            checkVersion(inputStream);
+            for (DeletionFileWithDataFile ddFile : ddFiles) {
+                inputStream.seek(ddFile.deletionFile().offset());
```
JingsongLi, line 98: add a `checkArgument(ddFile.deletionFile equal index)`?

paimon-core/src/main/java/org/apache/paimon/deletionvectors/DeletionVectorIndexFileMaintainer.java
```diff
@@ -0,0 +1,112 @@
...
+import java.util.stream.Collectors;
+
+/** DeletionVectorIndexFileMaintainer. */
+public class DeletionVectorIndexFileMaintainer {
+
+    private final IndexFileHandler indexFileHandler;
+
+    private final Map<String, IndexManifestEntry> indexNameToEntry = new HashMap<>();
+
+    private final Map<String, List<DeletionFileWithDataFile>> indexFileToDataFiles =
```
JingsongLi, line 42: Can we just introduce a `Map<String, DeletionFile>`?

I feel that we have strict regulations, where a data file can correspond to a maximum of one deletion file. So a Map here looks better.

And in this way we don't need to introduce a new class `DeletionFileWithDataFile`.

JingsongLi (commented):
Left minor comments.

JingsongLi (approved):
+1

PR #3391, opened 2024-05-24
[flink] Introduce file index rewriter procedure

leaves12138:
### Purpose

Add / Drop / Modify file index for existing tables. 

This action is strictly idempotent.

Trigger this procedure will:
1. If `ManifestFile` does not contain the target file index column and type, add it
2. If `ManifestFile` contains the index column and type, but table options don't, drop it
3. If `ManifestFile` contains the index column and type, and table options also contain, leave it

### API and Format

Three options:
1. Table name
2. Partition filter

sql
```
CALL sys.file_index_rewrite('default.T'); -- just rewrite all the table

CALL sys.file_index_rewrite('default.T', 'pt=a'); --add partition filter

```

### DIFF
--- a/docs/content/concepts/specification.md
+++ b/docs/content/concepts/specification.md
@@ -247,7 +247,33 @@ Global Index is in the index directory, currently, only two places will use glob
 
 ## Data File Index
 
-Define `file-index.bloom-filter.columns`, Paimon will create its corresponding index file for each file. If the index
-file is too small, it will be stored directly in the manifest, or in the directory of the data file. Each data file
-corresponds to an index file, which has a separate file definition and can contain different types of indexes with
-multiple columns.
+### Concept
+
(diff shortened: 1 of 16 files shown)


JingsongLi (changes requested):
Please write document first.

JingsongLi:
`CALL sys.file_index_rewrite('default.T', 'pt=a', 'file-index.bloom-filter.columns=''k'', xxx=xxx');`

We don't need to support this now.

leaves12138:
> `CALL sys.file_index_rewrite('default.T', 'pt=a', 'file-index.bloom-filter.columns=''k'', xxx=xxx');`
> 
> We don't need to support this now.

Removed

leaves12138:
> Please write document first.

Added document

JingsongLi:
+1

PR #3396, opened 2024-05-26
added greater and less than query optimization to snapshot table queries

ctrlaltdilj:
optimizing queries on snapshot table


paimon-core/src/main/java/org/apache/paimon/table/system/SnapshotsTable.java
```diff
@@ -207,9 +210,29 @@ public InnerTableRead withFilter(Predicate predicate) {
 
             LeafPredicate snapshotPred =
                     predicate.visit(LeafPredicateExtractor.INSTANCE).get("snapshot_id");
-            if (snapshotPred != null && snapshotPred.function() instanceof Equal) {
-                specificSnapshot = (Long) snapshotPred.literals().get(0);
+            if (snapshotPred != null) {
+                if (snapshotPred.function() instanceof Equal) {
```
Zouxxyy, line 214: This will cause performance degradation of equal, because this requires list files

paimon-core/src/main/java/org/apache/paimon/utils/SnapshotManager.java
```diff
@@ -364,6 +365,20 @@ public Iterator<Snapshot> snapshots() throws IOException {
...
     }
 
+    public Iterator<Snapshot> snapshotsWithinRange(Long maxSnapshotId, Long minSnapshotId)
+            throws IOException {
+        // check for specific snapshot
+        if (Objects.equals(maxSnapshotId, minSnapshotId)) {
+            return Collections.singletonList(snapshot(maxSnapshotId)).iterator();
+        }
+
+        return listVersionedFiles(fileIO, snapshotDirectory(), SNAPSHOT_PREFIX)
```
JingsongLi, line 375: We still need to list directory.
Can we directly read one by one?

We can combine maxSnapshotId, minSnapshotId, earliest and latest. Calculate a queryable interval and directly call the `snapshot` method one by one.

JingsongLi (commented):
Thanks @ctrlaltdilj for the contribution!

Left one comment.

paimon-core/src/main/java/org/apache/paimon/utils/SnapshotManager.java
ctrlaltdilj: Removed list call, lmk what you think

paimon-core/src/main/java/org/apache/paimon/table/system/SnapshotsTable.java
ctrlaltdilj: Originally I had preserved the efficiency in `snapshotsWithinRange`, with the new code removing the list call it should still be efficient with the iterating on a single snapshopId

JingsongLi (approved):
+1

PR #3399, opened 2024-05-27
[core] Load FileIndex with ServiceLoader

leaves12138:
### Purpose

Use ServiceLoader to load FileIndexer.

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/fileindex/FileIndexer.java
+++ b/paimon-common/src/main/java/org/apache/paimon/fileindex/FileIndexer.java
@@ -24,8 +24,8 @@ import org.apache.paimon.types.DataType;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.ArrayList;
-import java.util.List;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.ServiceLoader;
 
 /** File index interface. To build a file index. */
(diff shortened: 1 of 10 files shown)


paimon-common/src/main/java/org/apache/paimon/fileindex/FileIndexer.java
```diff
@@ -18,25 +18,50 @@
...
-import static org.apache.paimon.fileindex.bloomfilter.BloomFilterFileIndex.BLOOM_FILTER;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.ServiceLoader;
 
 /** File index interface. To build a file index. */
 public interface FileIndexer {
```
JingsongLi, line 32: Introduce `FileIndexerFactory`.

paimon-common/src/main/java/org/apache/paimon/fileindex/FileIndexer.java
```diff
@@ -18,25 +18,45 @@
...
-            case BLOOM_FILTER:
-                return new BloomFilterFileIndex(dataType, options);
-            default:
-                throw new RuntimeException("Doesn't support filter type: " + type);
+
+        ServiceLoader<FileIndexerFactory> serviceLoader =
+                ServiceLoader.load(FileIndexerFactory.class);
+
+        List<FileIndexerFactory> factories = new ArrayList<>();
+        for (FileIndexerFactory indexerFactory : serviceLoader) {
```
JingsongLi, line 46: Consider this is invoked in the critical path. Can we just cache `factories`?

paimon-common/src/main/java/org/apache/paimon/fileindex/FileIndexer.java
```diff
@@ -18,25 +18,56 @@
...
-                return new BloomFilterFileIndex(dataType, options);
-            default:
-                throw new RuntimeException("Doesn't support filter type: " + type);
+
+        FileIndexerFactory fileIndexerFactory = FileIndexerLoadFactory.load(type);
+        return fileIndexerFactory.create(dataType, options);
+    }
+
+    /** Load factory to load FileIndexerFactory. */
+    class FileIndexerLoadFactory {
```
JingsongLi, line 47: We don't need to use this class FileIndexLoadFactory, it doesn't make much sense. Static variables and methods are directly placed in FileIndexer

JingsongLi: You can just create `FileIndexerFactoryUtils`.

JingsongLi (approved):
+1

PR #3402, opened 2024-05-27
[core] Introduce DeletionVectorIndexFileWriter

YannByron:
### Purpose

to support merge and write multi deletion vector within same partition and bucket to one index file.

Linked issue: close #xxx

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/deletionvectors/DeletionVectorIndexFileWriter.java
+++ b/paimon-core/src/main/java/org/apache/paimon/deletionvectors/DeletionVectorIndexFileWriter.java
@@ -1,0 +1,131 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.paimon.deletionvectors;
+
+import org.apache.paimon.fs.FileIO;
+import org.apache.paimon.fs.Path;
+import org.apache.paimon.index.IndexFileMeta;
+import org.apache.paimon.options.MemorySize;
+import org.apache.paimon.table.BucketMode;
+import org.apache.paimon.utils.Pair;
+import org.apache.paimon.utils.PathFactory;
+import org.apache.paimon.utils.Preconditions;
+
+import java.io.DataOutputStream;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+import static org.apache.paimon.deletionvectors.DeletionVectorsIndexFile.DELETION_VECTORS_INDEX;
+import static org.apache.paimon.deletionvectors.DeletionVectorsIndexFile.VERSION_ID_V1;
+import static org.apache.paimon.deletionvectors.DeletionVectorsIndexFile.calculateChecksum;
+
+public class DeletionVectorIndexFileWriter {
+
+    private final PathFactory indexPathFactory;
+    private final FileIO fileIO;
+    private final boolean isWrittenToMulitFiles;
+    private final long targetSizeInBytes;
+
+    private boolean written = false;
+
+    private long writtenSizeInBytes = 0L;
+
+    public DeletionVectorIndexFileWriter(
+            FileIO fileIO,
+            PathFactory pathFactory,
+            BucketMode bucketMode,
+            MemorySize targetSizePerIndexFile) {
+        this.indexPathFactory = pathFactory;
+        this.fileIO = fileIO;
+        this.isWrittenToMulitFiles = BucketMode.BUCKET_UNAWARE == bucketMode;
+        this.targetSizeInBytes = targetSizePerIndexFile.getBytes();
+    }
+
+    public List<IndexFileMeta> write(Map<String, DeletionVector> input) throws IOException {
+        List<IndexFileMeta> result = new ArrayList<>();
+        SingleIndexFileWriter writer = createWriter();
+
+        for (Map.Entry<String, DeletionVector> entry : input.entrySet()) {
+            String dataFile = entry.getKey();
+            byte[] valueBytes = entry.getValue().serializeToBytes();
+            int currentSize = valueBytes.length;
+
+            if (isWrittenToMulitFiles
+                    && written
+                    && writtenSizeInBytes + currentSize > targetSizeInBytes) {
+                result.add(writer.closeWriter());
+                writer = createWriter();
+            }
+
+            writer.write(dataFile, valueBytes);
+            written = true;
+            writtenSizeInBytes += currentSize;
+        }
+        result.add(writer.closeWriter());
+        return result;
+    }
+
+    private SingleIndexFileWriter createWriter() throws IOException {
+        written = false;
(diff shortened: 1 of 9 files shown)


paimon-core/src/main/java/org/apache/paimon/deletionvectors/DeletionVectorIndexFileWriter.java
```diff
@@ -0,0 +1,132 @@
...
+
+/** Writer for deletion vector index file. */
+public class DeletionVectorIndexFileWriter {
+
+    private final PathFactory indexPathFactory;
+    private final FileIO fileIO;
+    private final boolean isWrittenToMulitFiles;
+    private final long targetSizeInBytes;
+
+    private boolean written = false;
```
JingsongLi, line 49: This class looks a bit strange, why does it need variables? It only provides one method, which appears to not require any state preservation.

paimon-core/src/main/java/org/apache/paimon/deletionvectors/DeletionVectorIndexFileWriter.java
```diff
@@ -0,0 +1,132 @@
...
+                    && writtenSizeInBytes + currentSize > targetSizeInBytes) {
+                result.add(writer.closeWriter());
+                writer = createWriter();
+            }
+
+            writer.write(dataFile, valueBytes);
+            written = true;
+            writtenSizeInBytes += currentSize;
+        }
+        result.add(writer.closeWriter());
```
JingsongLi, line 84: It is better to use `try (resource)` to avoid resource leak.

paimon-core/src/main/java/org/apache/paimon/deletionvectors/DeletionVectorIndexFileWriter.java
```diff
@@ -0,0 +1,132 @@
...
+        return result;
+    }
+
+    private SingleIndexFileWriter createWriter() throws IOException {
+        written = false;
+        writtenSizeInBytes = 0L;
+        return new SingleIndexFileWriter(fileIO, indexPathFactory.newPath());
+    }
+
+    static class SingleIndexFileWriter {
```
JingsongLi, line 94: private, and maybe can be an inner class instead of static.
In this way, you can inline `createWriter` to the `SingleIndexFileWriter` constructor.

paimon-core/src/main/java/org/apache/paimon/deletionvectors/DeletionVectorIndexFileWriter.java
```diff
@@ -0,0 +1,157 @@
...
+            throws IOException {
+        SingleIndexFileWriter writer =
+                new SingleIndexFileWriter(fileIO, indexPathFactory.newPath());
+        try {
+            while (iterator.hasNext()) {
+                Map.Entry<String, DeletionVector> entry = iterator.next();
+                long currentSize = writer.write(entry.getKey(), entry.getValue());
+
+                if (isWrittenToMulitFiles
+                        && !writer.hasWritten()
```
JingsongLi, line 85: `!writer.hasWritten()` always return true?

Maybe you should remove it.

paimon-core/src/main/java/org/apache/paimon/deletionvectors/DeletionVectorIndexFileWriter.java
```diff
@@ -0,0 +1,157 @@
...
+import static org.apache.paimon.deletionvectors.DeletionVectorsIndexFile.DELETION_VECTORS_INDEX;
+import static org.apache.paimon.deletionvectors.DeletionVectorsIndexFile.VERSION_ID_V1;
+import static org.apache.paimon.deletionvectors.DeletionVectorsIndexFile.calculateChecksum;
+
+/** Writer for deletion vector index file. */
+public class DeletionVectorIndexFileWriter {
+
+    private final PathFactory indexPathFactory;
+    private final FileIO fileIO;
+    private final boolean isWrittenToMulitFiles;
```
JingsongLi, line 49: typo: `isWrittenToMultiFiles`

paimon-core/src/main/java/org/apache/paimon/deletionvectors/DeletionVectorIndexFileWriter.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    private List<IndexFileMeta> emptyIndexFile() throws IOException {
+        try (SingleIndexFileWriter writer =
+                new SingleIndexFileWriter(fileIO, indexPathFactory.newPath())) {
+            return Collections.singletonList(writer.writtenIndexFile());
+        }
+    }
+
+    class SingleIndexFileWriter implements Closeable {
```
JingsongLi, line 103: private

paimon-core/src/main/java/org/apache/paimon/deletionvectors/DeletionVectorIndexFileWriter.java
```diff
@@ -0,0 +1,157 @@
...
+
+        private final Path path;
+
+        private final DataOutputStream dataOutputStream;
+
+        private final LinkedHashMap<String, Pair<Integer, Integer>> dvRanges;
+
+        private long writtenSizeInBytes = 0L;
+
+        public SingleIndexFileWriter(FileIO fileIO, Path path) throws IOException {
```
JingsongLi, line 114:
```
public SingleIndexFileWriter() throws IOException {
     this.path = indexPathFactory.newPath();
     .....
}
```

paimon-core/src/main/java/org/apache/paimon/deletionvectors/DeletionVectorIndexFileWriter.java
```diff
@@ -0,0 +1,157 @@
...
+
+    private final PathFactory indexPathFactory;
+    private final FileIO fileIO;
+    private final boolean isWrittenToMulitFiles;
+    private final long targetSizeInBytes;
+
+    public DeletionVectorIndexFileWriter(
+            FileIO fileIO,
+            PathFactory pathFactory,
+            BucketMode bucketMode,
```
JingsongLi, line 55: Maybe we don't need `bucketMode`, here we can just pass a `long targetSizePerIndexFile`. If it is `BUCKET_UNAWARE`, `targetSizePerIndexFile` can be `Long.MAX`.

Same to `DeletionVectorsIndexFile`.

paimon-core/src/main/java/org/apache/paimon/deletionvectors/DeletionVectorIndexFileWriter.java
```diff
@@ -0,0 +1,157 @@
...
+            MemorySize targetSizePerIndexFile) {
+        this.indexPathFactory = pathFactory;
+        this.fileIO = fileIO;
+        this.isWrittenToMulitFiles = BucketMode.BUCKET_UNAWARE == bucketMode;
+        this.targetSizeInBytes = targetSizePerIndexFile.getBytes();
+    }
+
+    public List<IndexFileMeta> write(Map<String, DeletionVector> input) throws IOException {
+        if (input.isEmpty()) {
+            return emptyIndexFile();
```
JingsongLi, line 65: If this is possible? Why we need to return a empty Index file?

paimon-core/src/main/java/org/apache/paimon/deletionvectors/DeletionVectorIndexFileWriter.java
YannByron: Fine. That's what I thought at first. But I don't think it's a little serious.

paimon-core/src/main/java/org/apache/paimon/deletionvectors/DeletionVectorIndexFileWriter.java
YannByron: This empty index file is needed to overwrite the previous one when `compact`.
I am going to think it again when to support `compact` operation on deletion vector.

YannByron: here just aligned to the original logic.

JingsongLi: This may only be to overwrite the old deletion file, as there is no proactive message to delete the deletion file for the primary key table.

But for append table, maybe here no need to generate empty deletion file?

JingsongLi (approved):
+1

PR #3406, opened 2024-05-27
[core] Allow FileIndexer to seek and skip if they need to.

leaves12138:
### Purpose

Give `SeekablePositionedMappingInputStream `to FileIndexer, allow them to skip to specified position if they need to.


paimon-common/src/main/java/org/apache/paimon/fileindex/FileIndexer.java
```diff
@@ -29,7 +30,7 @@ public interface FileIndexer {
 
     FileIndexWriter createWriter();
 
-    FileIndexReader createReader(byte[] serializedBytes);
+    FileIndexReader createReader(SeekablePositionedMappingInputStream inputStream);
```
JingsongLi, line 33: Does here need `SeekablePositionedMappingInputStream`? Can it just be `SeekableInputStream`?

paimon-common/src/main/java/org/apache/paimon/fileindex/FileIndexer.java
```diff
@@ -29,7 +31,8 @@ public interface FileIndexer {
 
     FileIndexWriter createWriter();
 
-    FileIndexReader createReader(byte[] serializedBytes);
+    FileIndexReader createReader(
+            SeekableInputStream inputStream, Pair<Integer, Integer> startAndLength);
```
JingsongLi, line 35: int start, int length

JingsongLi:
+1

PR #3409, opened 2024-05-27
[core] close files in CompactRewriter to avoid file leaks.

### LINKED ISSUE #3408: [Bug] File leaks may occur with CompactRewriter.
### Search before asking

### Paimon version

paimon-0.8

### Compute Engine

flink 1.17

### Minimal reproduce step

In https://github.com/apache/paimon/blob/master/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/ChangelogMergeTreeRewriter.java#L156, `iterator` / `compactFileWriter` / `changelogFileWriter` may throw an `Exception`, which will cause the file to fail to be closed and cleaned.

We should close `iterator` / `compactFileWriter` / `changelogFileWriter` and clean the file when an exception occurs.

### Anything else?

_No response_

liming30:
[core] close files in `CompactRewriter` to avoid file leaks.
### Purpose

Linked issue: close #3408 

### Tests

- org.apache.paimon.mergetree.ChangelogMergeTreeRewriterTest#testRewriteFailAndCleanupFiles
- org.apache.paimon.mergetree.ChangelogMergeTreeRewriterTest#testRewriteSuccess

### API and Format

No
### Documentation

No

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/ChangelogMergeTreeRewriter.java
+++ b/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/ChangelogMergeTreeRewriter.java
@@ -30,7 +30,9 @@ import org.apache.paimon.manifest.FileSource;
 import org.apache.paimon.mergetree.MergeSorter;
 import org.apache.paimon.mergetree.SortedRun;
 import org.apache.paimon.utils.CloseableIterator;
+import org.apache.paimon.utils.ExceptionUtils;
 import org.apache.paimon.utils.FieldsComparator;
+import org.apache.paimon.utils.IOUtils;
 
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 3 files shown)


paimon-core/src/main/java/org/apache/paimon/mergetree/compact/ChangelogMergeTreeRewriter.java
```diff
@@ -151,18 +154,22 @@ private CompactResult rewriteOrProduceChangelog(
...
-            if (iterator != null) {
-                iterator.close();
-            }
-            if (compactFileWriter != null) {
-                compactFileWriter.close();
-            }
-            if (changelogFileWriter != null) {
-                changelogFileWriter.close();
+            try {
+                IOUtils.closeAll(iterator, compactFileWriter, changelogFileWriter);
```
JingsongLi, line 161: I do a test:
```
    public static void main(String[] args) throws Exception {
        Exception exception = null;
        try {
            boolean b = ThreadLocalRandom.current().nextBoolean();
            if (b) {
                throw new RuntimeException();
            }
        } catch (Exception e) {
            exception = e;
        } finally {
            System.out.println(exception == null);
            if (exception != null) {
                throw exception;
            }
        }
    }
```

The exception is visible in `finally`.

JingsongLi (approved):
+1

PR #3411, opened 2024-05-28
[Hive] Fix flink not synchronizing to hive metastore when deleting table partitions using hive catalog.

zhuangchong:
### Purpose

Fix flink not synchronizing to hive metastore when deleting table partitions using hive catalog.


JingsongLi (approved):
+1

thanks @zhuangchong !

PR #3421, opened 2024-05-29
[core] [bug] Partition should not be filtered when manifest file predicate and file itself all min/max is null

leaves12138:
### Purpose

Linked issue: close #xxx

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/partition/PartitionPredicate.java
+++ b/paimon-core/src/main/java/org/apache/paimon/partition/PartitionPredicate.java
@@ -121,8 +121,11 @@ public interface PartitionPredicate {
             PredicateBuilder builder = new PredicateBuilder(partitionType);
             for (int i = 0; i < collectors.length; i++) {
                 SimpleColStats stats = collectors[i].result();
-                min[i] = builder.greaterOrEqual(i, stats.min());
-                max[i] = builder.lessOrEqual(i, stats.max());
+                Object minValue = stats.min();
+                Object maxValue = stats.max();
+
+                min[i] = minValue == null ? builder.isNull(i) : builder.greaterOrEqual(i, minValue);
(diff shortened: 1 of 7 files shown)


JingsongLi (commented):
See `PredicateBuilder.equalPartition` too.

JingsongLi:
See `RowDataToObjectArrayConverter.createEqualPredicate`.

JingsongLi:
See `SnapshotReaderImpl.withPartitionFilter`.

leaves12138:
FileStoreLookupFunction.createSpecificPartFilter

leaves12138:
PredicateBuilder.partition(Map<String, String> map, RowType rowType)

paimon-core/src/main/java/org/apache/paimon/partition/PartitionPredicate.java
```diff
@@ -121,8 +121,11 @@ private MultiplePartitionPredicate(
             PredicateBuilder builder = new PredicateBuilder(partitionType);
             for (int i = 0; i < collectors.length; i++) {
                 SimpleColStats stats = collectors[i].result();
-                min[i] = builder.greaterOrEqual(i, stats.min());
-                max[i] = builder.lessOrEqual(i, stats.max());
+                Object minValue = stats.min();
+                Object maxValue = stats.max();
+
+                min[i] = minValue == null ? builder.isNull(i) : builder.greaterOrEqual(i, minValue);
+                max[i] = maxValue == null ? builder.isNull(i) : builder.lessOrEqual(i, maxValue);
```
JingsongLi, line 128: Seems wrong logic here.
Collector should not collect null values.

JingsongLi: I see, if it is null, all values are null.

JingsongLi:
+1

PR #3422, opened 2024-05-29
[core] File index result should be able to process and/or.

leaves12138:
### Purpose

`FileIndexPredicate` return `FileIndexResult`, not `Boolean` any more.

For example, bit map index:

SQL:
```
SELECT * FROM <TABLE> WHERE a1 = 1 and a2 = 2
```

Bitmap index for a1 returns rows 1,2,3
Bitmap index for a2 returns rows 4,5,6.
Then this file could be skipped.

SQL
```
SELECT * FROM <TABLE> WHERE a1=1 or a1=2 or a1=3
```

Bitmap index for a1=1 returns rows null
Bitmap index for a1=2 return rows 1,2,3
Bitmap index for a1=3 return rows 3,4,5
Then this file should be found, and rows should be 1,2,3,4,5 (push down not implement yet)

Generally, file index implement should handle method 'or', whatever the IndexResult is produced by the file index type itself, or other kind of file index.

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/fileindex/FileIndexPredicate.java
+++ b/paimon-common/src/main/java/org/apache/paimon/fileindex/FileIndexPredicate.java
@@ -30,6 +30,9 @@ import org.apache.paimon.predicate.Predicate;
 import org.apache.paimon.predicate.PredicateVisitor;
 import org.apache.paimon.types.RowType;
 
+import org.slf4j.Logger;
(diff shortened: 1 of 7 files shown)


JingsongLi:
Can you write an example about how to use `FileIndexResult`.

leaves12138:
> Can you write an example about how to use `FileIndexResult`.

Added an example about bitmap int pull request description.

JingsongLi:
+1

PR #3423, opened 2024-05-29
[core] Introduce ReadBuilder.withShard

JingsongLi:
### Purpose

Introduce `ReadBuilder.withShard`:

```
    /** Specify the shard to be read, and allocate sharded files to read records. */
    ReadBuilder withShard(int indexOfThisSubtask, int numberOfParallelSubtasks);
```

For some OLAP engines, they do not have a central node to uniformly generate splits, which requires each work to generate the splits that need to be read.

Before this feature, if each worker were to retrieve all the splits, there might be a significant cost, so this interface should minimize the reading of metadata as much as possible.


leaves12138 (approved):
+1

PR #3429, opened 2024-05-29
[spark] Reshuffle splits in PaimonSplitScan

Zouxxyy:
### Purpose

PaimonSplitScan can use `reshuffleSplits` just like PaimonScan


YannByron:
+1.

YannByron (approved):

PR #3430, opened 2024-05-29
[core] Adjust VectoredReadable to better read performance

JingsongLi:
### Purpose

For continuous disk data, if multiple threads are started to read, the performance will be worse than sequential reading.

So the strategy has been adjusted here to ensure that continuous data is read by a single thread as much as possible.


yuzelin (approved):
+1

PR #3437, opened 2024-05-30
[flink] Support FileStoreLookupFunction to only cache required buckets

WencongLiu:
### Purpose

Modifies Paimon's FileStoreLookupFunction to cache only the required buckets during lookup joins.

### Documentation

no


JingsongLi (commented):
Can you take a look to ReadBuilder.withShard?

WencongLiu:
> Can you take a look to ReadBuilder.withShard?

Thanks to @JinsongLi for the reminder.
Assuming the bucket IDs are evenly distributed to each subtask in a round-robin fashion, then the withShard API can retrieve all the data corresponding to the respective subtask.
However, this distribution method cannot support multiple strategies for bucket shuffle, so we need withBucketFilter to provide more flexibility.
I have updated the PR, please take a look when you have time.😄

JingsongLi (approved):
+1

PR #3445, opened 2024-05-30
[core] Fix migrate hive partitioned table with null partition

Zouxxyy:
### Purpose

Fix migrate hive partitioned table with null partition


paimon-common/src/main/java/org/apache/paimon/data/BinaryWriter.java
```diff
@@ -212,5 +212,9 @@ static ValueSetter createValueSetter(DataType elementType) {
     /** Accessor for setting the elements of a binary writer during runtime. */
     interface ValueSetter extends Serializable {
         void setValue(BinaryWriter writer, int pos, Object value);
+
+        default void setNull(BinaryWriter writer, int pos) {
```
JingsongLi, line 216: Don't need to introduce this, here will some performance problem when introduce a default method to a FunctionalInterface.

leaves12138 (approved):
+1

JingsongLi:
+1

PR #3447, opened 2024-05-30
[test] Fix flaky test testLookupDynamicPartition

Zouxxyy:
### Purpose

Fix flaky test testLookupDynamicPartition, https://github.com/apache/paimon/actions/runs/9305326341/job/25611978740

image

### DIFF
--- a/paimon-core/src/test/java/org/apache/paimon/utils/TraceableFileIO.java
+++ b/paimon-core/src/test/java/org/apache/paimon/utils/TraceableFileIO.java
@@ -32,6 +32,7 @@ import org.apache.paimon.fs.local.LocalFileIO;
 import javax.annotation.concurrent.GuardedBy;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


paimon-core/src/test/java/org/apache/paimon/utils/TraceableFileIO.java
```diff
@@ -250,15 +251,13 @@ public String toString() {
     }
 
     public static List<SeekableInputStream> openInputStreams(Predicate<Path> filter) {
-        return OPEN_INPUT_STREAMS.stream()
-                .filter(s -> filter.test(s.file))
-                .collect(Collectors.toList());
+        return new ArrayList<>(OPEN_INPUT_STREAMS)
```
JingsongLi, line 254: It looks like it's also traversed once, is there any difference?

Zouxxyy: copy out to avoid ConcurrentModificationException

JingsongLi (approved):
+1

PR #3452, opened 2024-05-31
[core] Fisrt row merge engine supports none changelog producer

JingsongLi:
### Purpose

We should support this. Just like deletion vectors mode, no need to bind lookup changelog producer.


Zouxxyy (approved):
+1

PR #3453, opened 2024-05-31
[hotfix] Delete redundant symbols in compaction doc

xleoken:
### Purpose

hotfix.

Delete redundant symbols in compaction doc.


JingsongLi (approved):
+1

PR #3455, opened 2024-05-31
[parquet] Parquet use vector IO

JingsongLi:
### Purpose

Next step of https://github.com/apache/incubator-paimon/pull/3369


leaves12138 (approved):
+1

PR #3462, opened 2024-06-03
[spark] Use the value from options ​​instead of default value for 'partition.default-name' 

Zouxxyy:
### Purpose

- Use the value from options ​​instead of default value for 'partition.default-name' 
- fix drop null partition
- minor code refactor for we need to use scala's lazy val


JingsongLi (approved):
+1

PR #3463, opened 2024-06-03
[format] Default compression of file to zstd with level 1

JingsongLi:
### Purpose

After multiple tests, zstd has a higher compression rate than lz4 and snappy at level 1, and the speed difference is not significant.

1. So this PR will adjust the default compression to zstd.
2. The level of zstd is 1 instead of 3, users should adjust `file.compression.zstd-level` value if they want the default level same to zstd lib.
3. Another change is that the compression parameter for the format itself is adjusted to the highest priority, which means that as long as you configure the compression of the format, it will be applied.

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
@@ -152,14 +152,14 @@ public class CoreOptions implements Serializable {
                     .stringType()
                     .defaultValue("zstd")
                     .withDescription(
-                            "Default file compression. It can be overridden by format own configuration.");
+                            "Default file compression. For faster read and write, it is recommended to use LZ4.");
 
     public static final ConfigOption<Integer> FILE_COMPRESSION_ZSTD_LEVEL =
             key("file.compression.zstd-level")
                     .intType()
                     .defaultValue(1)
                     .withDescription(
-                            "Default file compression zstd level. It can be overridden by format own configuration.");
+                            "Default file compression zstd level. For higher compression rates, it can be configured to 9, but the read and write speed will significantly decrease.");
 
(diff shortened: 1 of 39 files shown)


paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -150,10 +150,23 @@ public class CoreOptions implements Serializable {
     public static final ConfigOption<String> FILE_COMPRESSION =
             key("file.compression")
                     .stringType()
+                    .defaultValue("zstd")
+                    .withDescription(
+                            "Default file compression. For faster read and write, it is recommended to use LZ4.");
```
leaves12138, line 155: LZ4 --> zstd

paimon-format/src/main/java/org/apache/paimon/format/parquet/writer/RowDataParquetBuilder.java
```diff
@@ -77,14 +75,10 @@ public ParquetWriter<InternalRow> createWriter(OutputFile out, String compressio
...
-            compressName = compression;
-        } else {
-            compressName =
-                    conf.get(ParquetOutputFormat.COMPRESSION, CompressionCodecName.SNAPPY.name());
+    public String getCompression(String compression) {
+        if (conf.get("parquet.compression") != null) {
+            return conf.get("parquet.compression");
         }
-        return compressName;
+        return compression;
```
leaves12138, line 82: return conf.getString("parquet.compression", compression)?

JingsongLi: you are right... updated

leaves12138 (approved):
+1

PR #3465, opened 2024-06-03
[spark] column nullibility

### LINKED ISSUE #2889: [Bug] Spark doesn't respect column nullibility
### Search before asking

### Paimon version

0.7.0

### Compute Engine

Spark

### Minimal reproduce step

```java 
    @Test
    public void testNullibility() {
        spark.sql(
            "CREATE TABLE T (a INT not null, b STRING) TBLPROPERTIES"
                + " ('primary-key'='a', 'bucket'='4')");

        spark.sql("insert into T VALUES(1, 'aaa'), (null, 'bbb')");
        spark.table("T").printSchema();
        spark.sql("SELECT * FROM T").show(false);
    }
```

And the result is:
```text
root
 |-- a: integer (nullable = true)
 |-- b: string (nullable = true)

+----+---+
|a   |b  |
+----+---+
|NULL|bbb|
|1   |aaa|
+----+---+
```

Should respect column nullibility.

### Anything else?

_No response_

YannByron:
### Purpose

close #2889


JingsongLi (approved):
+1

PR #3467, opened 2024-06-04
[core] Fix that cannot read UPDATE_BEFORE of ignore-delete table

yuzelin:
### Purpose

To maintain compatibility with 0.7, #3139 put `ignoreDelete` in reader, causing that UPDATE_BEFORE cannot be read in streaming reading. So the `ignore-delete` have to be put into merge engines.


JingsongLi (commented):
Add comments to code, just be compatible with 0.7

JingsongLi:
+1

PR #3468, opened 2024-06-04
[parquet] Optimize parquet to use getBytesUnsafe for dictionary

JingsongLi:
### Purpose

Avoid copying byte arrays every time.


leaves12138 (approved):
+1

PR #3470, opened 2024-06-04
[flink] Fix the illegal usage of flink.shaded.guava31 dependency in TableSortInfo

WencongLiu:
### Purpose

Fix the illegal usage of flink.shaded.guava31 dependency in TableSortInfo

### Tests

no

### API and Format

no

### Documentation

no


JingsongLi (approved):
+1

PR #3472, opened 2024-06-04
[core] ReadBuilder.withShard should shuffle files for dv mode and first row

JingsongLi:
### Purpose

dv mode or first row mode, we should shuffle files because we don't need to merge files.


nagisayui:
+1 , we need this feature

PR #3474, opened 2024-06-04
[core] support custom commitUser

wwj6591812:
### Purpose

In our business scenario, the full data are written to the paimon table through bound stream job. Using of bound stream job instead of batch job is to enable downstream build index faster and further reduce the inc data of catchup. When bound stream job failover, the random commitUser will recover from Flink state, this is no problem. 

image

But In our business scenario, we use Apache Airflow schedule Flink jobs. A Airflow node will fail due to lots of reasons. Then user clear the airflow node, a new Flink bound stream job will start with no state, so a new CommitUser will be generated. This will affect downstream judgment on whether to use a new build index processor.

So I submit this pr, we can assign a fixed commitUser to an Airflow node.

image

### Tests

Add ut

### DIFF
--- a/docs/layouts/shortcodes/generated/core_configuration.html
+++ b/docs/layouts/shortcodes/generated/core_configuration.html
@@ -111,6 +111,12 @@ under the License.
             <td>Whether to force create snapshot on commit.</td>
         </tr>
         <tr>
+            <td><h5>commit.user-prefix</h5></td>
+            <td style="word-wrap: break-word;">(none)</td>
+            <td>String</td>
+            <td>Specifies the commit user prefix.</td>
+        </tr>
+        <tr>
             <td><h5>compaction.max-size-amplification-percent</h5></td>
             <td style="word-wrap: break-word;">200</td>
             <td>Integer</td>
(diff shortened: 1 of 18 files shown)


JingsongLi (changes requested):
This is a dangerous option, please write clearly its purpose.

JingsongLi:
It is not so simple... The commitIdentifier should be incremental too.

wwj6591812:
> It is not so simple... The commitIdentifier should be incremental too.

Yes，I change the config from customCommitUser to customCommitUserPrefix.
pls cc，thx

docs/layouts/shortcodes/generated/core_configuration.html
```diff
@@ -182,6 +182,12 @@
             <td>Duration</td>
             <td>The TTL in rocksdb index for cross partition upsert (primary keys not contain all partition fields), this can avoid maintaining too many indexes and lead to worse and worse performance, but please note that this may also cause data duplication.</td>
         </tr>
+        <tr>
+            <td><h5>custom.commit.user.prefix</h5></td>
```
JingsongLi, line 186: commit.user-prefix

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -1302,6 +1309,13 @@ public String fieldsDefaultFunc() {
         return options.get(FIELDS_DEFAULT_AGG_FUNC);
     }
 
+    public static String getCommitUser(Options options) {
```
JingsongLi, line 1312: `createCommitUser`

docs/layouts/shortcodes/generated/core_configuration.html
wwj6591812: done

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
wwj6591812: done

JingsongLi (approved):
+1

PR #3473, opened 2024-06-04
 [flink] supports upgrading from versions below 0.8.0 (excluding 0.8.0) to 0.8.1+.

liming30:
[flink] supports upgrading from versions below 0.8.0 (excluding 0.8.0) to 0.8.1+.

### Purpose

Linked issue: close #3367 

Supports flink job state compatibility when upgrading from versions below 0.8.0 to 0.8.1+.

When upgrading from `0.6/0.7` to `0.8.1+`, users can configure `state.compatible-for-less-than-08` to use the legacy serializer to deserialize the state.

### Tests

- `org.apache.paimon.manifest.ManifestCommittableLegacyV2SerializerCompatibilityTest#testCompatibilityToLegacyVersion2`

### API and Format

No
### Documentation

No

PR #3476, opened 2024-06-05
[core] Make ManifestCommittableSerializer be compatible with Paimon 0.7

JingsongLi:
### Purpose

Another solution for #3473


liming30 (approved):
+1

PR #3487, opened 2024-06-07
[parquet] Introduce filter push down for parquet

JingsongLi:
### Purpose

There are multiple groups within a Parquet, and implementing filter pushdown can have a certain degree of data skipping effect.


leaves12138 (approved):
+1

PR #3500, opened 2024-06-11
[core] Use rawConvertible to determine whether needs to convertToRawFiles

Zouxxyy:
### Purpose

#3226 Add `rawConvertible` in `DataSplit`, and remove the lazy loading mechanism.
We should use `rawConvertible` to determine whether needs to convertToRawFiles to preventing double convert


JingsongLi (approved):
+1

PR #3504, opened 2024-06-11
[spark] make metadata columns passed across shuffles

YannByron:
### Purpose

when there is a big subquery that need to involve shuffle to compute in delete condition, the metadata columns should be the output of `DatasourceV2Relation`, not just in the output of `DataSourceV2ScanRelation`, that can be computed correctly.


JingsongLi:
+1

PR #3505, opened 2024-06-11
[spark] adjust the field order of readSchema

YannByron:
### Purpose

To remove redundant projects requires the project and its child have the same output with order.
So we adjust here.


YannByron:
This pr will eliminate the `ProjectExec` node in cases that the `ProjectExec` output is same with its child output.

JingsongLi:
+1

PR #3508, opened 2024-06-12
[core]remove_orphan_files support dry run and optimize output

MOBIN-F:
### Purpose

1. remove_orphan_files  support dry run，similar to the iceberg
image

2. The current remove_orphan_files (flink/spark) only outputs the number of deleted files after execution. I think it should output the orphan file path to be more transparent to users. as follows

remove_orphan_files procedures output in spark/flink procedures
image

remove_orphan_files  action output in flink action jar
企业微信截图_1718182610243


MOBIN-F:
please help review this pr,tks @JingsongLi @yuzelin

JingsongLi (approved):
+1

PR #3510, opened 2024-06-12
[core] Deletion Vectors mode supports lookup async

JingsongLi:
### Purpose

In the previous implementation, due to concurrent thread safety issues, asynchronous execution could not be enabled for the deletion vectors mode, as there were also an asynchronous compaction thread to write deletion vectors when the main thread materialize them.

We can avoid this situation by handing over the materialization of the deletion vectors to the asynchronous compaction thread. After each compaction is completed, we actively materialize the deletion vectors as a deletion file.

### Tests

`PrimaryKeyFileStoreTableITCase` already coverred this.

### Documentation

`/primary-key-table/deletion-vectors` modified.

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
@@ -1904,7 +1904,7 @@ public class CoreOptions implements Serializable {
         return options.get(RECORD_LEVEL_TIME_FIELD);
     }
 
-    public boolean waitCompaction() {
+    public boolean prepareCommitWaitCompaction() {
         return changelogProducer() == ChangelogProducer.LOOKUP
                 && options.get(CHANGELOG_PRODUCER_LOOKUP_WAIT);
     }
--- a/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
@@ -1194,6 +1194,17 @@ public class CoreOptions implements Serializable {
                     .noDefaultValue()
                     .withDescription("Specifies the commit user prefix.");
 
+    public static final ConfigOption<Boolean> CHANGELOG_PRODUCER_LOOKUP_WAIT =
+            key("changelog-producer.lookup-wait")
+                    .booleanType()
+                    .defaultValue(true)
+                    .withDescription(
+                            "When "
+                                    + CoreOptions.CHANGELOG_PRODUCER.key()
+                                    + " is set to "
+                                    + ChangelogProducer.LOOKUP.name()
+                                    + ", commit will wait for changelog generation by lookup.");
+
     private final Options options;
 
     public CoreOptions(Map<String, String> options) {
@@ -1891,6 +1902,10 @@ public class CoreOptions implements Serializable {
     @Nullable
     public String recordLevelTimeField() {
         return options.get(RECORD_LEVEL_TIME_FIELD);
+    }
+
+    public boolean changelogProducerLookupWait() {
+        return options.get(CHANGELOG_PRODUCER_LOOKUP_WAIT);
     }
 
     /** Specifies the merge engine for table with primary key. */
--- a/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
@@ -1904,8 +1904,9 @@ public class CoreOptions implements Serializable {
         return options.get(RECORD_LEVEL_TIME_FIELD);
     }
 
-    public boolean changelogProducerLookupWait() {
-        return options.get(CHANGELOG_PRODUCER_LOOKUP_WAIT);
+    public boolean waitCompaction() {
+        return changelogProducer() == ChangelogProducer.LOOKUP
+                && options.get(CHANGELOG_PRODUCER_LOOKUP_WAIT);
     }
 
     /** Specifies the merge engine for table with primary key. */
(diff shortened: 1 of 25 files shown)


paimon-core/src/main/java/org/apache/paimon/compact/CompactDeletionFile.java
```diff
@@ -0,0 +1,91 @@
...
+
+    static CompactDeletionFile lazyGeneration(DeletionVectorsMaintainer maintainer) {
+        return new CompactDeletionFile() {
+            @Override
+            public Optional<IndexFileMeta> getOrCompute() {
+                return generateFiles(maintainer).getOrCompute();
+            }
+
+            @Override
+            public CompactDeletionFile mergeOldFile(CompactDeletionFile old) {
```
Zouxxyy, line 81: Maybe we can add a checker here for `CompactDeletionFile old` must have not called `getOrCompute` ever before

paimon-core/src/main/java/org/apache/paimon/mergetree/MergeTreeWriter.java
```diff
@@ -283,15 +286,17 @@ private CommitIncrement drainIncrement() {
...
         newFiles.clear();
         deletedFiles.clear();
         newFilesChangelog.clear();
         compactBefore.clear();
         compactAfter.clear();
         compactChangelog.clear();
+        this.compactDeletionFile = null;
 
-        return new CommitIncrement(dataIncrement, compactIncrement);
+        return new CommitIncrement(dataIncrement, compactIncrement, drainDeletionFile);
```
Zouxxyy, line 299: Can we run `compactDeletionFile.getOrCompute` here, and build a `IndexIncrement`

So that we can make `CommitIncrement` like this:
```
public class CommitIncrement {
    private final DataIncrement dataIncrement;
    private final CompactIncrement compactIncrement;
    private final IndexIncrement indexIncrement;
}
```

paimon-core/src/main/java/org/apache/paimon/compact/CompactDeletionFile.java
```diff
@@ -0,0 +1,91 @@
...
+
+    static CompactDeletionFile generateFiles(DeletionVectorsMaintainer maintainer) {
+        List<IndexFileMeta> files = maintainer.writeDeletionVectorsIndex();
+        if (files.size() > 1) {
+            throw new IllegalStateException(
+                    "Should only generate one compact deletion file, this is a bug.");
+        }
+        Optional<IndexFileMeta> deletionFile =
+                files.isEmpty() ? Optional.empty() : Optional.of(files.get(0));
+        PathFactory pathFactory = maintainer.pathFactory();
```
Zouxxyy, line 44: `DeletionVectorsMaintainer` has `indexFileHandler`, we can use `indexFileHandler.deleteIndexFile` to delete index file directly

Zouxxyy (commented):
Cool implementation, just leaving a little comments

paimon-core/src/main/java/org/apache/paimon/mergetree/MergeTreeWriter.java
```diff
@@ -283,15 +286,17 @@ private CommitIncrement drainIncrement() {
...
                         new ArrayList<>(compactChangelog));
+        CompactDeletionFile drainDeletionFile = this.compactDeletionFile;
 
         newFiles.clear();
         deletedFiles.clear();
         newFilesChangelog.clear();
         compactBefore.clear();
         compactAfter.clear();
         compactChangelog.clear();
+        this.compactDeletionFile = null;
```
Zouxxyy, line 297: Can we run `compactDeletionFile.getOrCompute` here, and build a `IndexIncrement`

So that we can make `CommitIncrement` like this:
```
public class CommitIncrement {
    private final DataIncrement dataIncrement;
    private final CompactIncrement compactIncrement;
    private final IndexIncrement indexIncrement;
}
```

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkSink.java
```diff
@@ -99,10 +98,11 @@ private StoreSinkWrite.Provider createWriteProvider(
         Options options = table.coreOptions().toConfiguration();
         ChangelogProducer changelogProducer = table.coreOptions().changelogProducer();
         boolean waitCompaction;
-        if (table.coreOptions().writeOnly()) {
+        CoreOptions coreOptions = table.coreOptions();
+        if (coreOptions.writeOnly()) {
             waitCompaction = false;
         } else {
-            waitCompaction = prepareCommitWaitCompaction(options);
+            waitCompaction = coreOptions.changelogProducerLookupWait();
```
Zouxxyy, line 105: here not add the judge of changelogProducer, may we need this in coreOptions

```
    public boolean waitCompaction() {
        return changelogProducer().equals(ChangelogProducer.LOOKUP) && changelogProducerLookupWait();
    }
```

Zouxxyy (approved):
+1

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -1893,6 +1904,10 @@ public String recordLevelTimeField() {
         return options.get(RECORD_LEVEL_TIME_FIELD);
     }
 
+    public boolean changelogProducerLookupWait() {
```
Zouxxyy, line 1907: Does it have to be used with lookup?

JingsongLi: good question! I will modify this method to waitCompaction, and add lookup judgement.

JingsongLi: back to name `prepareCommitWaitCompaction`.

PR #3517, opened 2024-06-13
[common] Minor refactor PredicateBuilder to avoid potential stack overflow

yuzelin:
### Purpose

The case is the upstream data changes too many partitions, make the `Predicate` too deep, then error is:
```
Caused by: java.lang.RuntimeException: java.lang.StackOverflowError
	at org.apache.paimon.manifest.ManifestFileMeta.merge(ManifestFileMeta.java:178)
	at org.apache.paimon.operation.FileStoreCommitImpl.tryCommitOnce(FileStoreCommitImpl.java:808)
	... 27 more
Caused by: java.lang.StackOverflowError
	at org.apache.paimon.predicate.CompoundPredicate.test(CompoundPredicate.java:59)
	at org.apache.paimon.predicate.Or.test(Or.java:55)
	at org.apache.paimon.predicate.CompoundPredicate.test(CompoundPredicate.java:59)
	...
```

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/predicate/PredicateBuilder.java
+++ b/paimon-common/src/main/java/org/apache/paimon/predicate/PredicateBuilder.java
@@ -156,9 +156,7 @@ public class PredicateBuilder {
         if (predicates.size() == 1) {
             return predicates.get(0);
         }
-        return predicates.stream()
-                .reduce((a, b) -> new CompoundPredicate(And.INSTANCE, Arrays.asList(a, b)))
-                .get();
+        return new CompoundPredicate(And.INSTANCE, predicates);
     }
 
     @Nullable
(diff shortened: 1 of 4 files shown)


paimon-common/src/main/java/org/apache/paimon/predicate/PredicateBuilder.java
```diff
@@ -156,9 +156,7 @@ public static Predicate and(List<Predicate> predicates) {
         if (predicates.size() == 1) {
             return predicates.get(0);
         }
-        return predicates.stream()
-                .reduce((a, b) -> new CompoundPredicate(And.INSTANCE, Arrays.asList(a, b)))
-                .get();
+        return new CompoundPredicate(And.INSTANCE, predicates);
```
JingsongLi, line 159: Here should be careful... Let me think about this.

JingsongLi:
Finished in https://github.com/apache/paimon/pull/3864

Issue #3340, opened 2024-05-16
[Bug] Partition expiration may conflict with in-flight compaction


schnappi17:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Paimon version

release-0.7

### Compute Engine

flink-1.17

### Minimal reproduce step

1. create partitioned table, set expired = 7d
2. mock data to always generate partition 19700101
3. sometimes it'll failover because compact deletion will conflict with the partition expiration. 
Let's say there is fileA which need to be marked as `delete` in a partition expiration task, and writer suppose the file still exists and pick it in a compaction task. Then compaction task will be finished with committing it as `delete` and will result in deleting a file which is not recorded by the base manifest.

### What doesn't meet your expectations?

This caused a flink job failover

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

PR #3518, opened 2024-06-14
[core] Update possible reasons for commit conflicts

tsreaper:
### Purpose

We recently found that, if data is written into expired partitions, commit conflicts may occur.

It is not easy to completely work around this scenario, so currently we prefer users to filter out expired data by themselves. This PR updates possible reasons for commit conflicts to remind the users.

### Tests

No tests.

### API and Format

No API changes.

### Documentation

No new features.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
@@ -18,6 +18,7 @@
 
 package org.apache.paimon.operation;
 
+import org.apache.paimon.CoreOptions;
 import org.apache.paimon.Snapshot;
 import org.apache.paimon.annotation.VisibleForTesting;
 import org.apache.paimon.data.BinaryRow;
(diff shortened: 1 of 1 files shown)


paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
```diff
@@ -1130,7 +1129,8 @@ private Pair<RuntimeException, RuntimeException> createConflictException(
                         "3. You're recovering from an old savepoint, or you're creating multiple jobs from a savepoint.",
                         "   The job will fail continuously in this scenario to protect metadata from corruption.",
                         "   You can either recover from the latest savepoint, "
-                                + "or you can revert the table to the snapshot corresponding to the old savepoint.");
+                                + "or you can revert the table to the snapshot corresponding to the old savepoint.",
+                        "4. The parallelism of committer, or the parallelism of compact coordinator (for append only scalable table) is not 1.");
```
JingsongLi, line 1133: We don't need this if we have runtime check.

paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
```diff
@@ -1115,12 +1116,10 @@ private Pair<RuntimeException, RuntimeException> createConflictException(
                         "Don't panic!",
                         "Conflicts during commits are normal and this failure is intended to resolve the conflicts.",
                         "Conflicts are mainly caused by the following scenarios:",
-                        "1. Your job is suffering from back-pressuring.",
-                        "   There are too many snapshots waiting to be committed "
-                                + "and an exception occurred during the commit procedure "
-                                + "(most probably due to checkpoint timeout).",
-                        "   See https://paimon.apache.org/docs/master/maintenance/write-performance/ "
-                                + "for how to improve writing performance.",
+                        "1. Data is written into expired partitions.",
```
JingsongLi, line 1119: Maybe we can check delete files and know it is the true reason?

tsreaper:
Also related to #3340.

JingsongLi:
Fixed in https://github.com/apache/paimon/pull/3945

PR #3533, opened 2024-06-17
[flink] Fix source numRecordsIn metrics

tsreaper:
### Purpose

FLIP-27 source has a default implementation of `numRecordsIn` metrics. However this metrics only count the number of batches Paimon source produces, not the number of records.

This PR fixes the metrics by increasing the counter before emitting data to source output.

### Tests

* `SourceMetricsITCase#testNumRecordsIn`

### API and Format

No API changes.

### Documentation

No new feature.

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/FlinkRecordsWithSplitIds.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/FlinkRecordsWithSplitIds.java
@@ -23,6 +23,7 @@ import org.apache.paimon.utils.Reference;
 
 import org.apache.flink.api.common.eventtime.TimestampAssigner;
 import org.apache.flink.api.connector.source.SourceOutput;
+import org.apache.flink.api.connector.source.SourceReaderContext;
 import org.apache.flink.connector.base.source.reader.RecordsWithSplitIds;
 import org.apache.flink.connector.file.src.reader.BulkFormat.RecordIterator;
 import org.apache.flink.connector.file.src.util.RecordAndPosition;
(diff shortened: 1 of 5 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/FlinkRecordsWithSplitIds.java
```diff
@@ -113,8 +115,22 @@ public static void emitRecord(
...
+                context.metricGroup().getIOMetricGroup().getNumRecordsInCounter();
+        boolean firstRecord = true;
+
         RecordAndPosition<RowData> record;
         while ((record = element.next()) != null) {
+            // First record in the iterator is already counted by SourceReaderBase#pollNext.
+            if (firstRecord) {
+                firstRecord = false;
+            } else {
+                numRecordsIn.inc();
```
JingsongLi, line 131: Can we just count here, and invoke `Counter.inc(cnt - 1)` if cnt big than 1.

tsreaper: It is possible that when emitting a batch of records, the source is backpressured and we have to stay in this while loop for some time. If at this moment the metric is queried, `numRecordsIn` will be inconsistent with that of the following operator.

I understand that we must be very careful when adding some code executed with each record. However `numRecordsIn.inc()` is quite simple. It just increases a `long` variable by 1.

JingsongLi (approved):
Can you also modify `ReadOperator`?

JingsongLi (approved):
+1

Issue #3514, opened 2024-06-13
[Bug] When metastore.partitioned-table is enabled, the partition information in hms will not be deleted after the expired partition is automatically deleted.


codeTai:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Paimon version

0.8-SNAPSHOT

### Compute Engine

Flink

### Minimal reproduce step

When metastore.partitioned-table is enabled, the partition information in hms will not be deleted after the expired partition is automatically deleted.

### What doesn't meet your expectations?

Expired partition information in hms is not deleted, causing partition information to accumulate

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

gsralex:
please assign to me

PR #3534, opened 2024-06-17
[core] Fix the expired partition sync to metastore

gsralex:
### Purpose

Linked issue: close [When metastore.partitioned-table is enabled, the partition information in hms will not be deleted after the expired partition is automatically deleted.](https://github.com/apache/paimon/issues/3514)


leaves12138:
Thanks for @gsralex, can you run `mvn spotless:apply` to fix the build issue after adding some tests for expired partition function you have written?

gsralex:
> Thanks for @gsralex, can you run `mvn spotless:apply` to fix the build issue after adding some tests for expired partition function you have written?

OK, I'm very willing.

leaves12138:
Can you write a test for this? You can refer `HiveCatalogITCaseBase.testFlinkWriteAndHiveReadToCompare`. Use flink to write data and expire partition, use `hiveShell` to check if the partition is dropped actually in hive metastore.

leaves12138 (approved):
+1

PR #3536, opened 2024-06-18
[core] Fix FieldCountAgg null value init process

xuzifu666:
Currently we had FieldAggregator to compute all field cout，but not support filter null field count，this pr is aim to support it.

Linked issue: https://github.com/apache/paimon/issues/3537

Null value is filter count by agg operation


paimon-core/src/main/java/org/apache/paimon/mergetree/compact/aggregate/FieldNonNullCountAgg.java
```diff
@@ -0,0 +1,78 @@
...
+            }
+        }
+        return count;
+    }
+
+    @Override
+    public Object retract(Object accumulator, Object inputField) {
+        Object count;
+        if (accumulator == null || inputField == null) {
+            count = (accumulator == null ? (inputField == null ? 0 : 1) : accumulator);
```
JingsongLi, line 62: should be -1?

xuzifu666: Yes, it should be -1，thanks for the reminder, it alse error in FieldCountAgg，would fix it all. @JingsongLi

JingsongLi (commented):
Hi I think there is bug in `FieldCountAgg`, the count should 0 when accumulator is null and inputField is null.

JingsongLi:
actually `FieldCountAgg` is the `FieldNonNullCountAgg`.

xuzifu666:
> actually `FieldCountAgg` is the `FieldNonNullCountAgg`.

Add the logic to the FieldCountAgg and remove FieldNonNullCountAgg.

JingsongLi (approved):
+1

PR #3551, opened 2024-06-19
[cdc] Handle case-insensitive in separate operator instead of record parsers

yuzelin:
### Purpose

Currently, all record parsers have to handle `case-insensitive`. This PR want to move it to a separate operator.

Note that #2485 has already considered handle `case-insensitive` with computed columns, so this PR won't cause problems to computed columns.

### Tests
1. there are already some test cases for table sync with `case-insensitive`.
2. add `MySqlSyncDatabaseActionITCase#testIgnoreCaseCombined`
3. enhance `KafkaCanalSyncDatabaseActionITCase#testCaseInsensitive`


JingsongLi (approved):
+1

PR #3553, opened 2024-06-20
[spark]Support get full table name include database

waywtdcc:
### Purpose

Support get full table name include database
ranger spark auth plugin check auth error

image

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/SparkCatalog.java
+++ b/paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/SparkCatalog.java
@@ -211,7 +211,7 @@ public class SparkCatalog extends SparkBaseCatalog {
     @Override
     public SparkTable loadTable(Identifier ident) throws NoSuchTableException {
         try {
-            return new SparkTable(load(ident));
+            return new SparkTable(load(ident), toIdentifier(ident).getDatabaseName());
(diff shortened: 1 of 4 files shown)


Zouxxyy:
This is the define of `String name()`

```scala
  /**
   * A name to identify this table. Implementations should provide a meaningful name, like the
   * database and table name from catalog, or the location of files for this table.
   */
  String name();
```

I don't think it is a good idea for range to get database through it, because this interface is just a name and users may define it arbitrarily.

However, enriching this name helps display information in spark UI, I remember iceberg is `catalog.db.name` now.

Zouxxyy:
This should be fixed in kyuubi range plugin

waywtdcc:
But it's not unique if there is only one table name, which doesn't feel good.

Zouxxyy:
> But it's not unique if there is only one table name, which doesn't feel good.

Yes, as I said before, this name will eventually be displayed in the UI, it'll be greate to like `catalog.db.name`

The current implementation requires modifying the constructor, it would be nice if the paimonTable could include it.

Zouxxyy:
https://github.com/apache/paimon/pull/3680 focus on the same porblem and merged, sorry for that

JingsongLi:
Fixed in:
https://github.com/apache/paimon/pull/3680
https://github.com/apache/paimon/pull/3686

PR #3554, opened 2024-06-20
[core] Check the validity of scan staring snapshotId

LinMingQiang:
…correct exception.

### Purpose

Linked issue: close #xxx


leaves12138:
This pull request should not effect scan between timestamp...

LinMingQiang:
Based on my understanding ,because of the left open right closed rule of ` IncrementalStartingScanner` that is different from `StaticFromStartingScanner`, we should allow starting snapshotId is equal earliestSnapshotId - 1 like 0.

LinMingQiang:
@JingsongLi Help to review this PR, thanks.

leaves12138 (approved):
+1

PR #3559, opened 2024-06-20
[cdc] Database sync supports specify partition keys and primary keys

yuzelin:
### Purpose

Specified keys if not exist in source table, we will try to set keys from source table or doesn't set keys.

### Tests

`MySqlSyncDatabaseActionITCase#testSpecifyKeys`
`KafkaCanalSyncDatabaseActionITCase#testSpecifyKeys`

### API and Format

database sync supports:
--primary-keys
--partition-keys


JingsongLi (approved):
+1

PR #3562, opened 2024-06-20
[core] speed up tagged snapshot search

askwang:
### Purpose

Linked issue: close #xxx

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/utils/TagManager.java
+++ b/paimon-core/src/main/java/org/apache/paimon/utils/TagManager.java
@@ -355,9 +355,16 @@ public class TagManager {
     }
 
     private int findIndex(Snapshot taggedSnapshot, List<Snapshot> taggedSnapshots) {
-        for (int i = 0; i < taggedSnapshots.size(); i++) {
-            if (taggedSnapshot.id() == taggedSnapshots.get(i).id()) {
-                return i;
+        int left = 0;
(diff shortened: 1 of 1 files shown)


paimon-core/src/main/java/org/apache/paimon/utils/TagManager.java
```diff
@@ -355,9 +355,16 @@ public List<String> allTagNames() {
...
+        int left = 0;
+        int right = taggedSnapshots.size() - 1;
+        while (left <= right) {
+            int mid = left + (right - left) / 2;
+            if (taggedSnapshots.get(mid).id() == taggedSnapshot.id()) {
+                return mid;
+            } else if (taggedSnapshots.get(mid).id() < taggedSnapshot.id()) {
+                left = mid + 1;
+            } else {
+                right = mid - 1;
```
TaoZex, line 367: How about encapsulating the binary search into a general static method?

askwang: good idea. should we add a new util class `ListUtils` or do you have some advice?

JingsongLi (changes requested):
Hi @askwang , I think we don't need to optimize search here. Here is just loop in memory. Very very fast...

askwang:
> Hi @askwang , I think we don't need to optimize search here. Here is just loop in memory. Very very fast...

+1, I close it

PR #3563, opened 2024-06-20
[core] update tag metadata for the same snapshot of the same tag

askwang:
### Purpose
When call create_tag twice for the same snapshot of the same tag, the tag medata should update.
```sql
CALL sys.create_tag(table => 'default.T', tag => 'my_tag', snapshot => 10)
CALL sys.delete_tag(table => 'default.T', tag => 'my_tag', snapshot => 10, time_retained => '1d')
```

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/utils/TagManager.java
+++ b/paimon-core/src/main/java/org/apache/paimon/utils/TagManager.java
@@ -104,14 +104,13 @@ public class TagManager {
                         ? Tag.fromSnapshotAndTagTtl(snapshot, timeRetained, LocalDateTime.now())
                                 .toJson()
                         : snapshot.toJson();
+        Path tagPath = tagPath(tagName);
 
         // update tag metadata into for the same snapshot of the same tag name.
         if (tagExists(tagName)) {
             Snapshot tagged = taggedSnapshot(tagName);
             Preconditions.checkArgument(
                     tagged.id() == snapshot.id(), "Tag name '%s' already exists.", tagName);
-
-            Path tagPath = tagPath(tagName);
             try {
                 fileIO.overwriteFileUtf8(tagPath, content);
             } catch (IOException ignored) {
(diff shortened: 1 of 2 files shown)


paimon-core/src/main/java/org/apache/paimon/utils/TagManager.java
```diff
@@ -94,26 +94,36 @@ public void createTag(
...
         if (tagExists(tagName)) {
             Snapshot tagged = taggedSnapshot(tagName);
             Preconditions.checkArgument(
                     tagged.id() == snapshot.id(), "Tag name '%s' already exists.", tagName);
+
+            Path tagPath = tagPath(tagName);
+            try {
+                fileIO.overwriteFileUtf8(tagPath, content);
+            } catch (IOException ignored) {
+                LOG.info(
```
JingsongLi, line 118: why catch the exception?

paimon-core/src/main/java/org/apache/paimon/utils/TagManager.java
```diff
@@ -94,26 +94,36 @@ public void createTag(
...
+                                .toJson()
+                        : snapshot.toJson();
+
+        // update tag metadata into for the same snapshot of the same tag name.
         if (tagExists(tagName)) {
             Snapshot tagged = taggedSnapshot(tagName);
             Preconditions.checkArgument(
                     tagged.id() == snapshot.id(), "Tag name '%s' already exists.", tagName);
+
+            Path tagPath = tagPath(tagName);
```
JingsongLi, line 114: extract this field to up.

askwang: done

paimon-core/src/main/java/org/apache/paimon/utils/TagManager.java
askwang: method overwriteFileUtf8 maybe throw IOException and createTag does not process it. we just print the info when overwriteFileUtf8 failed.

JingsongLi: So why not throw RuntimeException?

askwang: got. you are right. we should mark it failed.

JingsongLi (approved):
+1

PR #3566, opened 2024-06-21
[spark] Fix update append table with spark3.3-

Zouxxyy:
### Purpose

Fix update table with append table with spark 3.3-
Add a todo: support update appedn table with Exists/In subqueries with spark3.3-


leaves12138 (approved):
+1

Issue #3567, opened 2024-06-21
[Feature] Generate changelog file by copying data file when they are equal


yunfengzhou-hub:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Motivation

There are certain cases in which the changelog file and the data file of a bucket shares the same content. For example, it happens when a paimon job is used to synchronize data from a full snapshot of a database into paimon (which means no two records have the same primary key and no merge would be performed), and the job uses input as changelog producer. In such cases, instead of writing duplicated content twice, we can generate the changelog files by copying the data file and vice versa. This optimization can help reduce the IO overhead spent on Paimon sinks and improve the throughput of related jobs.

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

PR #3568, opened 2024-06-21
[core] Generate changelog by copying data when records are insert-only

yunfengzhou-hub:
### Purpose

Linked issue: #3567

- Implements custom copy in local filesystem and OSS
- Adds the optimization to generate changelog by copying data file
- Triggers the optimization when the following conditions are met
  - The filesystem implements copy
  - The changelog producer is INPUT
  - all records received since last flush are insert-only

### API and Format

This PR does not affect API or storage format.

### Documentation

This PR does not add a new feature that requires document.


yunfengzhou-hub:
Hi @JingsongLi could you please take a look at this PR?

paimon-core/src/main/java/org/apache/paimon/privilege/PrivilegedFileStore.java
```diff
@@ -128,9 +129,12 @@ public FileStoreWrite<T> newWrite(String commitUser) {
...
 
     @Override
-    public FileStoreWrite<T> newWrite(String commitUser, ManifestCacheFilter manifestFilter) {
+    public FileStoreWrite<T> newWrite(
+            String commitUser,
+            ManifestCacheFilter manifestFilter,
+            RecordAttributeManager recordAttributeManager) {
         privilegeChecker.assertCanInsert(identifier);
-        return wrapped.newWrite(commitUser, manifestFilter);
+        return wrapped.newWrite(commitUser, manifestFilter, null);
```
Sxnan, line 137: Should we pass the `recordAttributeManager` instead of null?

paimon-core/src/main/java/org/apache/paimon/privilege/PrivilegedFileStoreTable.java
```diff
@@ -265,9 +266,12 @@ public TableWriteImpl<?> newWrite(String commitUser) {
...
 
     @Override
-    public TableWriteImpl<?> newWrite(String commitUser, ManifestCacheFilter manifestFilter) {
+    public TableWriteImpl<?> newWrite(
+            String commitUser,
+            ManifestCacheFilter manifestFilter,
+            RecordAttributeManager recordAttributeManager) {
         privilegeChecker.assertCanInsert(identifier);
-        return wrapped.newWrite(commitUser, manifestFilter);
+        return wrapped.newWrite(commitUser, manifestFilter, null);
```
Sxnan, line 274: Same as above

paimon-core/src/main/java/org/apache/paimon/table/AppendOnlyFileStoreTable.java
```diff
@@ -135,15 +136,17 @@ public RecordReader<InternalRow> reader(Split split) throws IOException {
...
     @Override
     public TableWriteImpl<InternalRow> newWrite(
-            String commitUser, ManifestCacheFilter manifestFilter) {
+            String commitUser,
+            ManifestCacheFilter manifestFilter,
+            RecordAttributeManager recordAttributeManager) {
         // if this table is unaware-bucket table, we skip compaction and restored files searching
         AppendOnlyFileStoreWrite writer =
-                store().newWrite(commitUser, manifestFilter).withBucketMode(bucketMode());
+                store().newWrite(commitUser, manifestFilter, null).withBucketMode(bucketMode());
```
Sxnan, line 149: Same as above

paimon-filesystems/paimon-oss-impl/pom.xml
```diff
@@ -94,6 +94,12 @@
             <!-- packaged as an optional dependency that is only accessible on Java 11+ -->
             <scope>provided</scope>
         </dependency>
+        <dependency>
```
Sxnan, line 97: Is it used?

Sxnan (commented):
@yunfengzhou-hub Thanks for the PR. I left some comments below.

paimon-core/src/main/java/org/apache/paimon/utils/BucketRecordAttributeManager.java
```diff
@@ -0,0 +1,62 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.paimon.utils;
+
+/**
+ * Manager of the attributes of the internal records during runtime.
+ *
+ * <p>Different from {@link RecordAttributeManager}, this class manages the attributes need to be
```
JingsongLi, line 24: Why it need to be bucket's granularity?

paimon-core/src/main/java/org/apache/paimon/io/KeyValueFileWriterFactory.java
```diff
@@ -123,6 +124,21 @@ public void deleteFile(String filename, int level) {
         fileIO.deleteQuietly(formatContext.pathFactory(level).toPath(filename));
     }
 
+    public void copyFile(String sourceFileName, String targetFileName, int level)
+            throws IOException {
+        Path sourcePath = formatContext.pathFactory(level).toPath(sourceFileName);
+        Path targetPath = formatContext.pathFactory(level).toPath(targetFileName);
+        fileIO.copyFileUtf8(sourcePath, targetPath);
```
JingsongLi, line 131: Please add some test, you will find `copyFileUtf8` is incorrect...

paimon-core/src/test/java/org/apache/paimon/mergetree/MergeTreeTestBase.java
```diff
@@ -367,6 +368,35 @@ public void testWriteMany() throws Exception {
...
+                        Collections.emptyList(),
+                        createCompactManager(service, Collections.emptyList()),
+                        ChangelogProducer.INPUT,
+                        null);
+
+        doTestWriteReadWithChangelog(8, 200, false);
+    }
+
+    @Test
+    public void testChangelogFromCopyingData() throws Exception {
```
JingsongLi, line 384: Please also add test for non-copy file io too.

paimon-common/src/main/java/org/apache/paimon/fs/local/LocalFileIO.java
```diff
@@ -217,6 +217,19 @@ public boolean rename(Path src, Path dst) throws IOException {
         }
     }
 
+    @Override
+    public boolean copyFileUtf8(Path sourcePath, Path targetPath) throws IOException {
```
JingsongLi, line 221: We can have a copyFile method instead of copyFileUtf8.

paimon-core/src/main/java/org/apache/paimon/utils/BucketRecordAttributeManager.java
yunfengzhou-hub: Because the mergeTreeWriter and its flushing behavior is performed at bucket's granularity. From the Flink operator's perspective, it is not sure whether a batch of data flushed in MergeTreeWriter is all insert-only or not.

yunfengzhou-hub:
Thanks for the comments @JingsongLi. I have updated the PR according to the comments.

paimon-core/src/main/java/org/apache/paimon/table/AppendOnlyFileStoreTable.java
```diff
@@ -135,15 +138,18 @@ public RecordReader<InternalRow> reader(Split split) throws IOException {
...
-        return newWrite(commitUser, null);
+        return newWrite(commitUser, null, null);
     }
 
     @Override
     public TableWriteImpl<InternalRow> newWrite(
-            String commitUser, ManifestCacheFilter manifestFilter) {
+            String commitUser,
+            ManifestCacheFilter manifestFilter,
+            @Nullable RecordAttributeManager recordAttributeManager) {
```
JingsongLi, line 148: I prefer adding a `TableWriteImpl.withWriteOnly(boolean writeOnly)`.

paimon-core/src/main/java/org/apache/paimon/utils/BucketRecordAttributeManager.java
yunfengzhou-hub: According to the latest comment, we agree to introduce `withInsertOnly` instead of this class. Thus this comment is not applicable now.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkSinkBuilder.java
```diff
@@ -221,7 +222,7 @@ public FlinkSinkBuilder clusteringIfPossible(
     /** Build {@link DataStreamSink}. */
     public DataStreamSink<?> build() {
         input = trySortInput(input);
-        DataStream<InternalRow> input = MapToInternalRow.map(this.input, table.rowType());
+        DataStream<InternalRow> input = mapToInternalRow(this.input, table.rowType());
```
JingsongLi, line 225: Why introduce this?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/DynamicBucketSink.java
```diff
@@ -54,6 +55,16 @@ public DynamicBucketSink(
     protected abstract SerializableFunction<TableSchema, PartitionKeyExtractor<T>>
             extractorFunction();
 
+    protected HashBucketAssignerOperator<T> createHashBucketAssignerOperator(
```
JingsongLi, line 58: Why introduce this?

paimon-core/src/main/java/org/apache/paimon/mergetree/MergeTreeWriter.java
```diff
@@ -133,6 +139,23 @@ public MergeTreeWriter(
             compactChangelog.addAll(increment.compactIncrement().changelogFiles());
             updateCompactDeletionFile(increment.compactDeletionFile());
         }
+
+        // TODO: Verify the performance of the default implementation of copy,
+        //  and remove the following code block if proved to be of no regression.
+        boolean isCopyMethodOverridden;
+        try {
+            isCopyMethodOverridden =
```
JingsongLi, line 147: Can we just introduce a new FileIO interface?
```
interface NativeCopyFileIO extends FileIO {
    void nativeCopy(Path src, Path dst);
}
```

paimon-filesystems/paimon-oss-impl/src/main/java/org/apache/paimon/oss/OSSFileIO.java
```diff
@@ -70,6 +74,28 @@ public class OSSFileIO extends HadoopCompliantFileIO {
 
     private Options hadoopOptions;
 
+    @Override
+    public boolean copyFile(Path sourcePath, Path targetPath) throws IOException {
+        org.apache.hadoop.fs.Path srcPath = path(sourcePath);
+        org.apache.hadoop.fs.Path dstPath = path(targetPath);
+
+        try (FileSystem fs = createFileSystem(srcPath)) {
```
JingsongLi, line 82: Don't close FileSystem...

JingsongLi, line 82: Maybe we don't need to implement copy in OSSFileIO now.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkSinkBuilder.java
yunfengzhou-hub: According to offline discussion, I'll submit another PR to modify the accessibility of these methods.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/DynamicBucketSink.java
yunfengzhou-hub: Same as above.

paimon-core/src/main/java/org/apache/paimon/mergetree/MergeTreeWriter.java
yunfengzhou-hub: According to offline discussion, I benchmarked the throughput of the default implementation introduced in FileIO in Aliyun, and it seems that the default implementation has been enough to bring performance improvement. Given this result, I'll directly add the copy method to FileIO, instead of introducing the new NativeCopyFileIO interface.

paimon-core/src/main/java/org/apache/paimon/mergetree/MergeTreeWriter.java
```diff
@@ -276,6 +287,32 @@ public void sync() throws Exception {
         trySyncLatestCompaction(true);
     }
 
+    @Override
+    public void withInsertOnly(boolean insertOnly) {
+        this.isInsertOnly = insertOnly;
+        if (writeBuffer == null || writeBuffer.size() == 0) {
```
JingsongLi, line 293: we can just throw exception when insertOnly && writeBuffer not empty

paimon-core/src/main/java/org/apache/paimon/mergetree/MergeTreeWriter.java
```diff
@@ -276,6 +287,32 @@ public void sync() throws Exception {
...
+            areAllRecordsInsertOnlySinceLastFlush = isInsertOnly;
+        } else if (!isInsertOnly) {
+            areAllRecordsInsertOnlySinceLastFlush = false;
+        }
+    }
+
+    private List<DataFileMeta> getChangelogFileMetaFromDataFile(
```
JingsongLi, line 300: Just inline this method?

JingsongLi (approved):
+1

liming30:
@yunfengzhou-hub hi, when will `StoreSinkWrite#withInsertOnly` be called? I only see manual calls in the test code, which does not seem to be automatically inferred by the engine.

https://github.com/apache/paimon/blob/master/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/StoreSinkWrite.java#L49

PR #3573, opened 2024-06-21
[spark] SparkSql UT use version check instead of interfaces

xuzifu666:
SparkSql uts use version check instead of interfaces avoid multiple methods to control test logic

Linked issue: close #xxx


Zouxxyy (approved):
+1

PR #3579, opened 2024-06-22
[core] Remove static LOCAL_TZ

Zouxxyy:
### Purpose

`TimeZone.getDefault()` can't be static final, otherwise we will not be able to react to time zone changes

However, worried about performance degradation of timestamp


JingsongLi (approved):
+1

PR #3588, opened 2024-06-24
[test] Exclude orc dependency to avoid test fails in IDEA

yuzelin:
### Purpose
As far as I know, `FileStoreLookupFunctionTest` will throws `NoSuchMethodError` in IDEA currently.

Linked issue: close #xxx


JingsongLi (approved):
+1

Issue #3593, opened 2024-06-24
[Bug] PartitionExpire drop partition atomicity when sync to hms


xuzifu666:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Paimon version

master

### Compute Engine

spark3.2.0
flink1.17.2

### Minimal reproduce step

CREATE TABLE bdsp_test.paimon_mi_24(
  `user_id` STRING COMMENT '任务用到的库',
  `used_table` STRING COMMENT '任务用到的表',
  `day` STRING COMMENT '按天进行分区'
)USING paimon
PARTITIONED BY (day);

insert into bdsp_test.paimon_mi_24
select 'a', 'a_table', '2024-04-22';

insert into bdsp_test.paimon_mi_24
select 'b', 'a_table', '2024-09-22';

select * from bdsp_test.paimon_mi_24 limit 10;

CALL sys.expire_partitions(table => 'bdsp_test.paimon_mi_24', expiration_time => '1 d', timestamp_formatter => 'yyyy-MM-dd');

If hive version is low which cannot compatible with paimon（such as follow），hms sync would error but paimon drop partition is ok
![1719229921801.png](https://github.com/apache/paimon/assets/10645422/f746c33c-db46-47ee-96a7-e55ec47f0614)

### What doesn't meet your expectations?

Drop partition automicly, If hms failed, paimon also failed.

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

PR #3594, opened 2024-06-24
[core] PartitionExpire drop partitions consisitent with hms drop

xuzifu666:
### Purpose
PartitionExpire drop partition atomicity when sync to hms.
CREATE TABLE bdsp_test.paimon_mi_24(
user_id STRING COMMENT '任务用到的库',
used_table STRING COMMENT '任务用到的表',
day STRING COMMENT '按天进行分区'
)USING paimon
PARTITIONED BY (day);

insert into bdsp_test.paimon_mi_24
select 'a', 'a_table', '2024-04-22';

insert into bdsp_test.paimon_mi_24
select 'b', 'a_table', '2024-09-22';

select * from bdsp_test.paimon_mi_24 limit 10;

CALL sys.expire_partitions(table => 'bdsp_test.paimon_mi_24', expiration_time => '1 d', timestamp_formatter => 'yyyy-MM-dd');

If hive version is low which cannot compatible with paimon（such as follow），hms sync would error but paimon drop partition is ok

Linked issue: https://github.com/apache/paimon/issues/3593

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/PartitionExpire.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/PartitionExpire.java
@@ -108,10 +108,10 @@ public class PartitionExpire {
             LOG.info("Expire Partition: " + partition);
         }
         if (expired.size() > 0) {
-            commit.dropPartitions(expired, commitIdentifier);
             if (metastoreClient != null) {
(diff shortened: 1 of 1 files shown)


JingsongLi (approved):
+1

PR #3602, opened 2024-06-25
[core] Support delete by default in partial updates

### LINKED ISSUE #3048: [Feature] support delete record when merge engine is partial-update
### Search before asking

### Motivation

for now it's not allowed to delete records in a partial-update table. even though a retract with sequence group is ok, but it is not a real delete action. suppose i want to execute a batch sql like `delete from xxx_table where id = xx`, it will throw an exception. i wonder if this delete feature is able to be complemented or not. if it can be achieved, will it be on the roadmap or if it can not be achieved can anyone explains a little bit why. thank you.

### Solution

_No response_

### Anything else?

_No response_

yunfengzhou-hub:
### Purpose
Linked issue: close #3048

This pull request supports dealing with deletion records in partial update merge engine.

### Tests
Unit tests are added in PrimaryKeyFileStoreTableTest to verify the changes in this PR

### API and Format
This pull request affects the default behavior of partial-update merge engine in cases where deletion records exist. Given that the existing behavior is to throw exception in this case, the changes in this PR are backward compatible.

### Documentation
This PR requires to update the document about merge engine, which has been included in the commit.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/PartialUpdateMergeFunction.java
+++ b/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/PartialUpdateMergeFunction.java
@@ -29,6 +29,7 @@ import org.apache.paimon.types.DataType;
 import org.apache.paimon.types.RowKind;
 import org.apache.paimon.types.RowType;
 import org.apache.paimon.utils.FieldsComparator;
+import org.apache.paimon.utils.Preconditions;
 import org.apache.paimon.utils.Projection;
 import org.apache.paimon.utils.UserDefinedSeqComparator;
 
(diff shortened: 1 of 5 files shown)


yunfengzhou-hub:
Hi @JingsongLi Could you please take a look at this PR?

paimon-core/src/main/java/org/apache/paimon/mergetree/compact/PartialUpdateMergeFunction.java
```diff
@@ -357,25 +388,22 @@ public MergeFunction<KeyValue> create(@Nullable int[][] projection) {
                                                 newRowType, newSequenceFields));
                             }
                         });
-                for (int i = 0; i < projects.length; i++) {
```
JingsongLi, line 360: Why removing this code?

yunfengzhou-hub: Oh it is a mistake when I attempted to remove the codes related to primaryKeyMask. I've added it back.

yunfengzhou-hub:
Thanks for the comments @JingsongLi. I have updated the PR according to the comments.

JingsongLi (approved):
+1

PR #3608, opened 2024-06-25
[core] Skip pushing down partition filter to file reader

Zouxxyy:
### Purpose

Skip pushing down partition filter to file reader,  for the reasons:

- The partition filters have been applied to the splits generation, so pushing them to file has little effect.
- In some scenarios, the data file may not contain partition fields (such as migrate table, or our internal implementation that does not write partition fields). Then, if we push down the partition filter to the parquet reader, we will obtain no data. (The reason behind this is that parquet reader will treat all non-existent fields ​​as null, and if we push down `is not null`, reader will reject everything).

e.g. `SELECT * FROM migrate_tbl WHERE pt = 1` always return nothing

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/RawFileSplitRead.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/RawFileSplitRead.java
@@ -57,10 +57,12 @@ import javax.annotation.Nullable;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
+import static org.apache.paimon.predicate.PredicateBuilder.excludePredicateWithFields;
 import static org.apache.paimon.predicate.PredicateBuilder.splitAnd;
 
 /** A {@link SplitRead} to read raw file directly from {@link DataSplit}. */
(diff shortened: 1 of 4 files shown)


Zouxxyy:
Worried about performance regression, maybe we can add a configuration to control this behavior

paimon-core/src/main/java/org/apache/paimon/operation/RawFileSplitRead.java
```diff
@@ -181,9 +185,19 @@ private RawFileBulkFormatMapping createBulkFormatMapping(FormatKey key) {
                         ? filters
                         : SchemaEvolutionUtil.createDataFilters(
                                 tableSchema.fields(), dataSchema.fields(), filters);
+        List<Predicate> nonPartitionFilters = dataFilters;
 
         Pair<int[], RowType> partitionPair = null;
         if (!dataSchema.partitionKeys().isEmpty()) {
+            // Skip pushing down partition filters to reader
+            if (dataFilters != null && !dataFilters.isEmpty()) {
```
JingsongLi, line 193: extract a method to be shared

JingsongLi:
> Worried about performance regression, maybe we can add a configuration to control this behavior

no need, there is no performance regression

JingsongLi (approved):
+1

PR #3610, opened 2024-06-26
[parquet] Parquet supports column index filter

JingsongLi:
### Purpose

- Use column index filter api in parquet reader.
- Add tests.


JingsongLi:
https://issues.apache.org/jira/browse/SPARK-34859

JingsongLi:
Don't not merge this. It seems we need to have more works to do.

JingsongLi:
Also should consider row position of returned iterator.

PR #3613, opened 2024-06-26
[spark] Optimize sort compact procedure by submitting job by partition

Zouxxyy:
### Purpose

Currently, spark sort compact will sort all the filtered data globally, and then write them by dynamic partition overwrite, which will lead to the following problems:

1. If the partition is not specified, data will be sorted globally, which is an unnecessary overhead
2. because it is a full range shuffle, small files are uncontrollable, for all reducers may contain all partitions

Therefore, this PR make a single partition as a sort compact group and then submit it to spark, a new conf ~`max_order_threads`~ `max_concurrent_jobs ` is introduced in sort compact procedure, the maximum number of concurrent job submissions, default is 15.

### Tests

Test on 1T tpc-ds, sort compact web_sales on ws_bill_customer_sk, before 233s -> after 108s, and the target file size can be controled by `spark.sql.shuffle.partitions`

### DIFF
--- a/docs/content/spark/procedures.md
+++ b/docs/content/spark/procedures.md
@@ -46,7 +46,7 @@ This section introduce all available spark procedures about paimon.
             <li>where: partition predicate. Left empty for all partitions. (Can't be used together with "partitions")</li>          
             <li>order_strategy: 'order' or 'zorder' or 'hilbert' or 'none'. Left empty for 'none'.</li>
             <li>order_columns: the columns need to be sort. Left empty if 'order_strategy' is 'none'.</li>
-            <li>max_order_threads: when sort compact is used, files in one partition are grouped and submitted as a single spark job. This parameter controls the maximum number of jobs that can be submitted simultaneously. The default value is 15.</li>
+            <li>max_concurrent_jobs: when sort compact is used, files in one partition are grouped and submitted as a single spark compact job. This parameter controls the maximum number of jobs that can be submitted simultaneously. The default value is 15.</li>
       </td>
       <td>
          SET spark.sql.shuffle.partitions=10; --set the compact parallelism <br/>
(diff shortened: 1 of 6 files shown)


docs/content/spark/procedures.md
```diff
@@ -46,6 +46,7 @@ This section introduce all available spark procedures about paimon.
             <li>where: partition predicate. Left empty for all partitions. (Can't be used together with "partitions")</li>          
             <li>order_strategy: 'order' or 'zorder' or 'hilbert' or 'none'. Left empty for 'none'.</li>
             <li>order_columns: the columns need to be sort. Left empty if 'order_strategy' is 'none'.</li>
+            <li>max_order_threads: when sort compact is used, files in one partition are grouped and submitted as a single spark job. This parameter controls the maximum number of jobs that can be submitted simultaneously. The default value is 15.</li>
```
JingsongLi, line 49: max_concurrent_jobs ?

paimon-spark/paimon-spark-common/src/test/scala/org/apache/paimon/spark/procedure/CompactProcedureTestBase.scala
```diff
@@ -247,6 +249,48 @@ abstract class CompactProcedureTestBase extends PaimonSparkTestBase with StreamT
...
   }
 
+  test("Paimon Procedure: sort compact with max_order_threads") {
+    Seq("order", "zorder").foreach {
+      orderStrategy =>
+        {
+          withTable("T") {
+            spark.sql(s"""
+                         |CREATE TABLE T (id INT, pt STRING)
+                         |TBLPROPERTIES ('bucket'='-1')
```
JingsongLi, line 259: remove `'bucket'='-1'`, it is default value

JingsongLi (approved):
+1

PR #3616, opened 2024-06-26
[spark] Integrate paimon scan metrics into spark scan

Zouxxyy:
### Purpose

Integrate paimon scan metrics into spark scan, so that we can get more information on the UI. Note: these new metrics are only available from Spark 3.4, because `reportDriverMetrics` was added in 3.4.

e.g.

```sql
select count(*) from inventory where inv_date_sk = 2450822
```

image


Zouxxyy:
Hi, @schnappi17 can you help with a review~ Not sure if this is the correct way to use `MetricRegistry`, because the metrics of spark and flink are different.

YannByron (approved):

JingsongLi (approved):
+1 I think it is the correct way to register metrics.

PR #3617, opened 2024-06-27
[flink] Change the access modifier of sink and operators before sink

yunfengzhou-hub:
### Purpose

This PR modifies the access of classes and methods related to flink sink.


JingsongLi (approved):
+1

PR #3618, opened 2024-06-27
[core] Remove replace branch to reduce IO

JingsongLi:
### Purpose

The replace_branch incurs an expensive IO overhead for most operations in the normal code path. For HDFS, it is a namenode access, and for object storage, it is a separate billing. And for object storage, every IO cost 10ms+. It is very costly.

This is difficult to accept, and if replace_branch is not as useful, we should remove this operation.

Detail is here: https://lists.apache.org/thread/q46clxx38fz7n1xw0sgscmcslo3qrp5c


schnappi17:
+1 for this.

Issue #2706, opened 2024-01-16
[Feature][cdc] Handle Kafka CDC Debezium/Canal JSON from Flink SQL


JingsongLi:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/incubator-paimon/issues) and found nothing similar.

### Motivation

At present, some Kafka CDC data comes from Flink SQL, and the CDC data produced by Flink SQL lacks a lot of information (even the data from Flink CDC lacks PK information). There is no schema or PK, and we should also have the ability to handle such scenarios in our Paimon Kafka CDC.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

Pandas886:
I have the same problem.First, I use mysqlcdc to kafka topic.

![image](https://github.com/apache/incubator-paimon/assets/123344357/33f848df-ce9e-457e-9bc7-1f2a8ba76e58)

Then I use kafka cdc action to paimon.I found that the table that automatically builds paimon does not correctly identify the primary key.

the record in topic like that:
```
{
  "schema": {
    "type": "struct",
    "fields": [
      {
        "type": "struct",
        "fields": [
          {
            "type": "int32",
            "optional": false,
            "field": "id"
          },
          {
            "type": "string",
            "optional": true,
            "field": "user_name"
          },
          {
            "type": "string",
            "optional": true,
            "field": "user_password"
          },
          {
            "type": "int16",
            "optional": true,
            "field": "user_type"
          },
          {
            "type": "string",
            "optional": true,
            "field": "email"
          },
          {
            "type": "string",
            "optional": true,
            "field": "phone"
          },
          {
            "type": "int32",
            "optional": true,
            "field": "tenant_id"
          },
          {
            "type": "string",
            "optional": true,
            "name": "com.darcytech.debezium.datetime.string",
            "field": "create_time"
          },
          {
            "type": "string",
            "optional": true,
            "name": "com.darcytech.debezium.datetime.string",
            "field": "update_time"
          },
          {
            "type": "string",
            "optional": true,
            "field": "queue"
          },
          {
            "type": "int16",
            "optional": true,
            "field": "state"
          },
          {
            "type": "string",
            "optional": true,
            "field": "time_zone"
          },
          {
            "type": "int32",
            "optional": true,
            "field": "ttt"
          }
        ],
        "optional": true,
        "name": "mysql_binlog_source.dolphinscheduler3.t_ds_user.Value",
        "field": "before"
      },
      {
[...]

PR #3621, opened 2024-06-27
[cdc] Database sync initializes the set of existed tables

yuzelin:
### Purpose

One of #2706 
The scenario is user might create tables before submitting cdc job, so if we doesn't collect them, the job may fail when processing records of the existed tables.


JingsongLi:
+1

PR #3624, opened 2024-06-27
[core] check null scanSnapshotId to avoid npe

wwj6591812:
add check to avoid npe

### Purpose

Linked issue: close #xxx


JingsongLi (approved):
+1

PR #3634, opened 2024-06-29
[core] Fix read row position with parquet when filter push down 

Zouxxyy:
### Purpose

Parquet write & expose row index in block metadata since 1.12.3 (paimon use 1.13.1) see https://github.com/apache/parquet-java/pull/945, we can use it to computer row position when blocks are filtered.  When row index was not found in block metadata, skip applying filter in order to get the correct row position

### Tests

- testReadRowPositionWithRandomFilter()
- test("Paimon deletionVector: select with format filter push down")


JingsongLi:
+1

PR #3636, opened 2024-06-29
[snapshot] Optimize watermark check in SnapshotManager which can reduce FileIO with Filesystem

xuzifu666:
### Purpose

Optimize watermark check in SnapshotManager which can earlier quit reduce FileIO with Filesystem:
we can find many watermark in snapshot could be Long.MIN firstly which should not be compute.

Linked issue: close https://github.com/apache/paimon/issues/3637


JingsongLi (approved):
+1

PR #3639, opened 2024-06-30
[core] Should not ignore return value of FileIO.writeFileUtf8

JingsongLi:
### Purpose

Old writeFileUtf8 should only be used by commit, others should throw exception when file existing. So this PR introduces a new writeFile.


yuzelin (approved):
+1

PR #3646, opened 2024-07-01
[core] Read deletion indexes at once to reduce file IO in splits generation

Zouxxyy:
### Purpose

 Read deletion indexes at once to reduce file IO in splits generation

### Tests

Test on tpcds q23a

before
```shell
grep FsStats spark-thrift-server.log | grep web_sales | wc
   7437   81807 1770826
```

after
```shell
grep FsStats spark-thrift-server.log | grep web_sales | wc
     67     737   16257
```


YannByron (approved):

JingsongLi (approved):
+1

PR #3650, opened 2024-07-01
[core] Fix that FileDeletionBase#dataFileSkipper might skip tag when deleting tag

yuzelin:
### Purpose

Scenario: If a stream writing job is running and expiring snapshots, and then deleting tags around the expiring snapshots, might lead to the skipping set is wrong and the reading job will throw FileNotFoundException.

The problem is，the `taggedSnapshots` may be changed. 
TODO: maybe should check delete tags.


JingsongLi:
+1

PR #3657, opened 2024-07-02
[core] Fix skippedTableFiles in scan metrics

Zouxxyy:
### Purpose

startDataFiles should be `numAddedFiles() - f.numDeletedFiles` instead of `numAddedFiles() + f.numDeletedFiles`


JingsongLi (approved):
+1

PR #3663, opened 2024-07-03
[core] Parquet becomes default file format

JingsongLi:
### Purpose

At present, Parquet is relatively excellent in terms of compression rate and analytical performance. It is time to use parquet by default.

### Tests

All tests cover default format.

### API and Format

Old tables use orc too, old new table use parquet.


yuzelin (approved):
+1

PR #3664, opened 2024-07-03
[core] Add an expiration strategy that comparing with partition updat…

LinMingQiang:
…e time.

### Purpose
Add an expiration strategy that comparing with partition update time.

Using `partition.expiration-strategy` option to specifies the expiration strategy for partition expiration.

Possible values: 

VALUES-TIME  (default): A partition expiration policy that compares the time extracted from the partition value with the current time.

UPDATE-TIME :  A partition expiration policy that compares the last update time of the partition with the current time.

What is the value of this feature ：

- If your partition value is not date formatted, you can use `UPDATE-TIME` as your partition expiration policy.

- If you only want to keep data that has been updated in the last n days/months/years, you can use `UPDATE-TIME` as your partition expiration policy.

Linked issue: close #xxx

### DIFF
--- a/docs/content/flink/expire-partition.md
+++ b/docs/content/flink/expire-partition.md
@@ -29,7 +29,7 @@ under the License.
 You can set `partition.expiration-time` when creating a partitioned table. Paimon streaming sink will periodically check
 the status of partitions and delete expired partitions according to time.
 
-How to determine whether a partition has expired: compare the time extracted from the partition with the current
+How to determine whether a partition has expired: compare the time extracted from the partition or the last update time of the partition with the current
 time to see if survival time has exceeded the `partition.expiration-time`.
 
 {{< hint info >}}
@@ -42,8 +42,9 @@ An example for single partition field:
 ```sql
 CREATE TABLE t (...) PARTITIONED BY (dt) WITH (
     'partition.expiration-time' = '7 d',
-    'partition.expiration-check-interval' = '1 d',
-    'partition.timestamp-formatter' = 'yyyyMMdd'
+    'partition.expiration-check-interval' = '1 d', -- this is required in `values-time` strategy.
+    'partition.timestamp-formatter' = 'yyyyMMdd'   -- this is required in `values-time` strategy.
+    -- ,'partition.expiration-strategy' = 'update-time'
 );
 ```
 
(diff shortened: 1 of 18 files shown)


paimon-core/src/main/java/org/apache/paimon/partition/PartitionExpireStrategy.java
```diff
@@ -0,0 +1,73 @@
...
+    public Object[] convertPartition(BinaryRow partition) {
+        return toObjectArrayConverter.convert(partition);
+    }
+
+    public abstract PartitionPredicate createPartitionPredicate(LocalDateTime expirationTime);
+
+    public abstract List<PartitionEntry> filterPartitionEntry(
+            List<PartitionEntry> partitionEntries, LocalDateTime expirationTime);
+
+    public static PartitionExpireStrategy createPartitionPredicate(
```
JingsongLi, line 62: createPartitionExpireStrategy

paimon-core/src/main/java/org/apache/paimon/operation/PartitionExpire.java
```diff
@@ -133,41 +122,11 @@ private void deleteMetastorePartitions(List<Map<String, String>> partitions) {
...
-        public boolean test(
-                long rowCount,
-                InternalRow minValues,
-                InternalRow maxValues,
-                InternalArray nullCounts) {
-            return true;
-        }
+    private List<PartitionEntry> readPartitions(LocalDateTime expireDateTime) {
+        return partitionExpireStrategy.filterPartitionEntry(
+                scan.withPartitionFilter(
```
JingsongLi, line 127: For update-time, you should use `withDataFileTimeMills` to filter partitions.

docs/content/flink/expire-partition.md
```diff
@@ -42,8 +42,9 @@ An example for single partition field:
 ```sql
 CREATE TABLE t (...) PARTITIONED BY (dt) WITH (
     'partition.expiration-time' = '7 d',
-    'partition.expiration-check-interval' = '1 d',
-    'partition.timestamp-formatter' = 'yyyyMMdd'
+    'partition.expiration-check-interval' = '1 d', -- this is required in `values-time` strategy.
```
JingsongLi, line 45: check-interval is required in `values-time` strategy?

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -681,6 +681,26 @@ public class CoreOptions implements Serializable {
                             "Whether only overwrite dynamic partition when overwriting a partitioned table with "
                                     + "dynamic partition columns. Works only when the table has partition keys.");
 
+    public static final ConfigOption<PartitionExpireStrategy> PARTITION_EXPIRATION_STRATEGY =
+            key("partition.expiration-strategy")
+                    .enumType(PartitionExpireStrategy.class)
+                    .defaultValue(PartitionExpireStrategy.VALUES_TIME)
```
JingsongLi, line 687: maybe this default value can be auto, without partition.timestamp-formatter, just use update_time.

JingsongLi:
Some discussion: If the partition conversion time fails, I feel like we can throw an exception and remind user to configure the formatter or use update-time mode.

LinMingQiang:
Changing the expiration strategy for a table with data is risky, may be we could provide a procedure to expire the specified wrong partition.

JingsongLi (approved):
Looks good to me!

PR #3668, opened 2024-07-03
[doc] Update spark procedure and dynamic overwrite docs

Zouxxyy:
### Purpose

- Fix sheet misalignment in `procedure` doc
- Make dynamic overwrite description clearer
- Remove `spark.sql.extensions` in other pages, because this is already a necessary configuration in `quick-start` doc


JingsongLi (approved):
+1

PR #3670, opened 2024-07-04
[hotfix] fix the spark generic catalog to execute externalCatalog() creates a new hive metastore connection instance.

zhuangchong:
### Purpose

The SparkSession.active().sharedState().externalCatalog().unwrapped() method will create a new hive metastore connection

image

image


JingsongLi (approved):
+1

PR #3672, opened 2024-07-04
[spark] CompactProcedure add options parameter

askwang:
### Purpose

`call sys.compact` can add some custom configs.


JingsongLi (approved):
+1

PR #3682, opened 2024-07-05
[doc] Document sequence.field supports fields of all data types

zhuangchong:
### Purpose

Document sequence.field supports fields of all data types


JingsongLi (approved):
+1

PR #3684, opened 2024-07-05
[procedure] Support multiple tags and branches delete

xuzifu666:
### Purpose
Support multiple tags and branches delete，when user create many tags or branches such as auto-create tag/branch，then need delete them in a quick way，so support multiple mode.

Linked issue: close [#xxx](https://github.com/apache/paimon/issues/3685)


JingsongLi (approved):
+1

PR #3686, opened 2024-07-05
[core] Get table name and full table name from identifier

Zouxxyy:
### Purpose

User may customize the location of database, so we tend to get the database name from identity return by catalog.

Similarly, get tablename from identity too, to adapt when we can customize the table location in the future (or now?)


JingsongLi (approved):
+1

PR #3691, opened 2024-07-07
[hive] Support multiple repair database or table in action and procedure

xuzifu666:
### Purpose

Add repairDatabasesOrTables in catalog for support multiple repair database or table. Currently repair table only can do one table in a job，so we can improve it with multiple way.

Linked issue: close [#xxx](https://github.com/apache/paimon/issues/3692)

### DIFF
--- a/docs/content/flink/procedures.md
+++ b/docs/content/flink/procedures.md
@@ -275,6 +275,8 @@ All available procedures are listed below.
          CALL sys.repair('databaseName')<br/><br/>
          -- repair a table<br/>
          CALL sys.repair('databaseName.tableName')<br/><br/>
+         -- repair database and table in a string If you specify multiple tags, delimiter is ','<br/>
+         CALL sys.repari('databaseName01,database02.tableName01,database03')
       </td>
       <td>
          Synchronize information from the file system to Metastore. Argument:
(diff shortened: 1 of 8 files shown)


docs/content/flink/procedures.md
```diff
@@ -275,6 +275,8 @@ All available procedures are listed below.
          CALL sys.repair('databaseName')<br/><br/>
          -- repair a table<br/>
          CALL sys.repair('databaseName.tableName')<br/><br/>
+         -- repair database and table in a string if you specify multiple tags, delimiter is ','<br/>
+         CALL sys.repari('databaseName01,database02.tableName01,database03')
```
JingsongLi, line 279: repair

paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
```diff
@@ -262,6 +263,12 @@ default void repairCatalog() {
         throw new UnsupportedOperationException();
     }
 
+    /** Repair databases or tables separated by commas. */
+    @Experimental
+    default void repairDatabasesOrTables(String databaseOrTables) throws TableNotExistException {
```
JingsongLi, line 268: Can you just extract a util method for `repairDatabasesOrTables`?
I don't want to modify API of catalog again.

xuzifu666: OK，would extract the logic to a util method instead of modify catalog. Thanks @JingsongLi

JingsongLi (approved):
+1

PR #3693, opened 2024-07-08
[core] Fix that FileDeletionBase#dataFileSkipper might skip tag when …

chenxi0599:
Purpose
The problem is，the taggedSnapshots may be changed.

This PR refers to [#3650](https://github.com/apache/paimon/pull/3650) on the master branch to modify the 0.8 branch.

Tests
API and Format
Documentation


JingsongLi (approved):
+1

PR #3694, opened 2024-07-08
[common] Use Paimon's FileIO to access hadoop-conf-dir with scheme

tsreaper:
### Purpose

Currently `hadoop-conf-dir` only supports local path. This PR uses Paimon's FileIO to access `hadoop-conf-dir` with scheme.

### Tests

* Unit test.

### API and Format

No changes.

### Documentation

No new feature.


JingsongLi (approved):
+1

Issue #3697, opened 2024-07-08
[Feature] Type handle check need more details about type and classname for user handle


xuzifu666:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Motivation

Type handle check need more details about type and classname for user dig
Currently when type not matched only throw out IllegalArgumentException without any details which is litter hard for user to dig their job，so add more detail in IllegalArgumentException maybe better.

### Solution

Add more detail in IllegalArgumentException about type info.

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

PR #3698, opened 2024-07-08
[core] Add type and classname in TypeHandle condition when throw IllegalArgumentException

xuzifu666:
### Purpose
Type handle check need more details about type and classname for user dig
Currently when type not matched only throw out IllegalArgumentException without any details which is litter hard for user to dig their job，so add more detail in IllegalArgumentException maybe better.

Linked issue: close https://github.com/apache/paimon/issues/3697


JingsongLi (approved):
+1

PR #3699, opened 2024-07-08
[flink] Fix multi-add conflict when two batch jobs are writing into the same partition at the same time

tsreaper:
### Purpose

In new versions of Flink, if a batch job fails, it might restart from some operator in the middle.

If the job is restarted from the commit operator, endInput will be called again, and the same commit messages will be committed again.

So when `endInput` is called, we must check if these commit messages are already committed.

### Tests

* IT case.

### API and Format

No changes.

### Documentation

No new feature.

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/CommitterOperator.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/CommitterOperator.java
@@ -156,8 +156,8 @@ public class CommitterOperator<CommitT, GlobalCommitT> extends AbstractStreamOpe
     @Override
     public void processWatermark(Watermark mark) throws Exception {
         super.processWatermark(mark);
-        // Do not consume END_INPUT_CHECKPOINT_ID watermark in case of batch or bounded stream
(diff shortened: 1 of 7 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/CommitterOperator.java
```diff
@@ -188,28 +189,37 @@ public void endInput() throws Exception {
...
-                committer.commit(Collections.singletonList(commit));
-            }
+        if (checkpointId == END_INPUT_CHECKPOINT_ID) {
+            // In new versions of Flink, if a batch job fails, it might restart from some operator
+            // in the middle.
+            // If the job is restarted from the commit operator, endInput will be called again, and
+            // the same commit messages will be committed again.
+            // So when `endInput` is called, we must check if these commit messages are already
+            // committed.
+            committer.filterAndCommit(committables);
```
JingsongLi, line 218: Here should filter commit, but should not check add files. The latter has a relatively high cost.

JingsongLi (approved):
+1

PR #3701, opened 2024-07-08
[core] Introduce newWriteSelector to WriteBuilder

JingsongLi:
### Purpose

Compute engines require the ability to write when integrating Paimon, and currently we only have one example of single point writing. We need to add an example of distributed writing.

### API and Format

new Experimental API: `WriteBuilder.newWriteSelector`:
```
    /**
     * Create a {@link WriteSelector} to partition records before writers.
     *
     * @return empty if no data distribution is required.
     * @throws UnsupportedOperationException if this table mode does not support building writes
     *     through upper level APIs.
     */
    @Experimental
    Optional<WriteSelector> newWriteSelector();
```

WriteSelector:
```
/**
 * The {@link WriteSelector} determines to which logical downstream writers a record should be
 * written to.
 */
public interface WriteSelector extends Serializable {

    /** Returns the logical writer index, to which the given record should be written. */
    int select(InternalRow row, int numWriters);
}
```

### DIFF
--- a/docs/content/program-api/java-api.md
+++ b/docs/content/program-api/java-api.md
@@ -536,6 +536,10 @@ public class BatchWrite {
         GenericRow record2 = GenericRow.of(BinaryString.fromString("Bob"), 5);
         GenericRow record3 = GenericRow.of(BinaryString.fromString("Emily"), 18);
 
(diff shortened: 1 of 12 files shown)


shaoxiong2019:
+1 Nice work， write-friendly for the MPP engine

PR #3707, opened 2024-07-08
[core] Disable forceBufferSpill when io-manager is null.

leaves12138:
io-manager is null, should not invoke spill.

### Purpose

Linked issue: close #xxx


JingsongLi (commented):
You can just adjust `forceBufferSpill`.

JingsongLi:
+1

PR #3714, opened 2024-07-09
[spark] Support StringEndsWith filter in SparkFilterConverter

xuzifu666:
### Purpose

Linked issue: close https://github.com/apache/paimon/issues/3715

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/predicate/EndsWith.java
+++ b/paimon-common/src/main/java/org/apache/paimon/predicate/EndsWith.java
@@ -48,11 +48,7 @@ public class EndsWith extends NullFalseLeafBinaryFunction {
             Object max,
             Long nullCount,
             Object patternLiteral) {
-        BinaryString minStr = (BinaryString) min;
-        BinaryString maxStr = (BinaryString) max;
-        BinaryString pattern = (BinaryString) patternLiteral;
-        return (minStr.endsWith(pattern) || minStr.compareTo(pattern) <= 0)
-                && (maxStr.endsWith(pattern) || maxStr.compareTo(pattern) >= 0);
+        return true;
     }
 
     @Override
--- a/paimon-common/src/main/java/org/apache/paimon/predicate/EndsWith.java
+++ b/paimon-common/src/main/java/org/apache/paimon/predicate/EndsWith.java
@@ -1,0 +1,67 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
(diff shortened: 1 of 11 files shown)


paimon-common/src/main/java/org/apache/paimon/predicate/EndsWith.java
```diff
@@ -0,0 +1,67 @@
...
+            DataType type,
+            long rowCount,
+            Object min,
+            Object max,
+            Long nullCount,
+            Object patternLiteral) {
+        BinaryString minStr = (BinaryString) min;
+        BinaryString maxStr = (BinaryString) max;
+        BinaryString pattern = (BinaryString) patternLiteral;
+        return (minStr.endsWith(pattern) || minStr.compareTo(pattern) <= 0)
```
JingsongLi, line 54: Really works? Can you add more tests?

Endswith does not support range filtering very well.

xuzifu666: OK，had add tests to test it. @JingsongLi

JingsongLi (commented):
You should add tests in `PredicateTest`.

paimon-spark/paimon-spark-common/src/test/java/org/apache/paimon/spark/SparkFilterConverterTest.java
```diff
@@ -192,6 +210,11 @@ public void testDate() {
 
     @Test
     public void testIgnoreFailure() {
+        testStartsWith();
```
Zouxxyy, line 213: No need modify this, this is only to test `convertIgnoreFailure`, regardless of the filter used

paimon-common/src/main/java/org/apache/paimon/predicate/EndsWith.java
Zouxxyy: +1 with @JingsongLi, does `endWith` really have any effect on range filtering?

Zouxxyy:
By the way, since endwith has no range filtering capability, even we pass it to paimon, it has no filtering effect...

Zouxxyy:
can add test for partition filering

paimon-spark/paimon-spark-common/src/test/java/org/apache/paimon/spark/SparkFilterConverterTest.java
xuzifu666: OK，removed the ut

JingsongLi (approved):
+1

PR #3716, opened 2024-07-09
[hotfix] Improve spark procedures document format and duplicate code calls

zhuangchong:
### Purpose

Improve spark procedures document format and duplicate code calls

current:
image


JingsongLi (approved):
+1

PR #2519, opened 2023-12-17
upgrade maven-jar-plugin: fixes Reproducible Builds

hboutemy:
### Purpose

fix Reproducible Builds issue caused by very old Maven Jar Plugin usage and Git branch property in MANIFEST.MF

### Tests

Reproducible Builds local testing is described in https://maven.apache.org/guides/mini/guide-reproducible-builds.html#how-to-test-my-maven-build-reproducibility

in short:
```
mvn clean install
mvn clean verify artifact:compare
```

### API and Format

should have no impact on paimon features, it's purely a build enhancement

### Documentation

https://github.com/jvm-repo-rebuild/reproducible-central/tree/master/content/org/apache/paimon/README.md

PR #3718, opened 2024-07-10
upgrade maven-jar-plugin

hboutemy:
### Purpose

upgrade maven-jar-plugin to non-obsolete version

### Tests

want to see how E2E CI reacts to that simple change (extracted from #2519), and investigate if anything goes wrong


hboutemy:
ok, basic compilation issue "Error:  Failed to execute goal org.apache.maven.plugins:maven-jar-plugin:3.4.2:jar (default) on project paimon-s3-impl: You have to use a classifier to attach supplemental artifacts to the project instead of replacing them. -> [Help 1]"
I can work on it...

PR #3719, opened 2024-07-10
upgrade maven-jar-plugin: fixes Reproducible Builds issue

hboutemy:
rework on #2519, in 2 steps: upgrade jar plugin first, then we'll see on removing the Git branch name configuration later

### Purpose

Linked issue: close #xxx

### DIFF
--- a/paimon-filesystems/paimon-oss-impl/pom.xml
+++ b/paimon-filesystems/paimon-oss-impl/pom.xml
@@ -102,13 +102,6 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-jar-plugin</artifactId>
-                <executions>
-                    <execution>
-                        <goals>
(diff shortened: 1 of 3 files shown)


hboutemy:
upgrading the jar plugin works: no problem found by CI
but dropping the Git branch name makes 2 tests fail:
- End to End Tests Flink 1.17
- Flink CDC Tests

I'm reverting the Git branch name removal: let's merge the good first step, that will solve one key prerequisite for full Reproducible Builds (sources-jar reproducibility should be ok, for example, but not yet normal jar files, that have the branch name in META-INF/MANIFEST)

hboutemy:
ok, CI is flaky: sometimes, no failures, sometimes failure with timeout
the problem is neither plugin upgrade nor branch name removal: it's just if I'm lucky or not

JingsongLi:
Thanks @hboutemy , +1 for merging.

PR #3721, opened 2024-07-10
[core] modify the default value of "target-file-size"

wwj6591812:
### Purpose

In our production practice, a large amount of data has already entered the lake. But during the production process, we found that the default parameter "target-file-size" is too small, which has caused us two inconveniences:
(1) The default parameter "target-file-size" is too small, resulting in too many ORC files, which often requires manual adjustment of the memory resources of the last GlobalCommittee node.
(2) The default parameter "target-file-size" is too small, which makes it easier to reach the threshold of num-sorted-run.stop-trigger, that resulting in faster triggering of the stopping of writes. This will affect the write throughput of the write job.

In order to reduce the difficulty of parameter tuning for new Paimon users, we would like to increase the default value of this parameter.
We investigated the default values of similar parameters in other data lake systems and found that Paimon has the smallest default value. Compared to Iceberg, it is 512 M (write.target-file-size-bytes), and Aliyun MaxComputer has a threshold of 500 M (odps.merge.maxmerged.filesize.threshold) when merging small files.

Linked issue: close #xxx


paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -438,7 +438,7 @@ public class CoreOptions implements Serializable {
     public static final ConfigOption<MemorySize> TARGET_FILE_SIZE =
             key("target-file-size")
                     .memoryType()
-                    .defaultValue(MemorySize.ofMebiBytes(128))
+                    .defaultValue(MemorySize.ofMebiBytes(256))
```
JingsongLi, line 441: Maybe we can set primary table to 128MB, but set append table to 256MB

wwj6591812: done

JingsongLi:
+1

Issue #3651, opened 2024-07-01
[Bug] spark fixed bucket write causes task data process unevenly for pk table


askwang:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Paimon version

0.8

### Compute Engine

spark

### Minimal reproduce step

When we use fixed bucket to write data for pk table, we set 1000 buckets, and want to process data with 1000 partitions evenly. But we found some tasks process no data and some tasks process many data. This will cause the overall running time of the task to be longer.
We found the root cause is `repartitionByExpression` will perform hash calculation again based on the bucket column. That will result in uneven distribution of partitions.
our some configs:
```sql
primary-key = uuid
bucket = 1000
spark.sql.shuffle.partitions=1000
```
![image](https://github.com/apache/paimon/assets/135721692/8e35fa86-ade6-4817-ad0b-d5699d5b2055)

### What doesn't meet your expectations?

partitions data distribution.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #3726, opened 2024-07-11
[WIP][spark] Use paimon hash function to repartition when writing

Zouxxyy:
### Purpose

to #3651


Zouxxyy:
This PR can avoid data skew, but it will introduce the following two problems,  close it now:
1. the writing performance is several times slower with the same resources.
2. due to the use of rdd, the spark UI cannot display sql correctly

Perhaps we should look for a solution to this problem by sticking with dataframe.

PR #3728, opened 2024-07-11
[benchmark] Fix the problem that parquet-related benchmark test cases call the wrong method in `TableReadBenchmark` class

### LINKED ISSUE #3727: [Bug] Parquet-related benchmark test cases call the wrong method in `TableReadBenchmark` class
### Search before asking

### Paimon version

Latest

### Compute Engine

Spark

### Minimal reproduce step

From the code

The parquet benchmark results are not right.

### Anything else?

_No response_

wayneguow:
### Purpose
Fix the problem that parquet-related benchmark test cases call the wrong method in `TableReadBenchmark` class.

Linked issue: close #3727


wayneguow:
cc @JingsongLi. I haven’t updated the benchmark results in `TableReadBenchmark ` class, because my computer environment is different.

I suggest that we should use GA's unified environment to verify the benchmark results, if necessary, I can work on it.

JingsongLi (approved):
+1

PR #3730, opened 2024-07-11
[spark] Fix spark query with filter timestamp

xuzifu666:
### Purpose
 Fix spark query with filter timestamp

before fix:
spark-sql> select * from bdsp_prd.z_paimon_pk_table where dt='2024-04-11 11:01:00' limit 10;
OK
pk	v1	v2	dt
Time taken: 1.198 seconds
spark-sql> 24/07/11 16:44:13 WARN SparkContext: Successfully stopped SparkContext

after fix，user can query by filter timestamp type：
Spark master: yarn, Application Id: application_1719458929416_3440055
spark-sql> select * from bdsp_prd.z_paimon_pk_table where dt='2024-07-09 12:22:21' limit 10;
24/07/11 16:34:03 WARN TaskSetManager: Finished tasks (1/1) in stage 0.0
OK
pk	v1	v2	dt
2	12.3	21	2024-07-09 12:22:21
Time taken: 18.313 seconds, Fetched 1 row(s)

Linked issue: close #xxx

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/predicate/PredicateBuilder.java
+++ b/paimon-common/src/main/java/org/apache/paimon/predicate/PredicateBuilder.java
@@ -279,7 +279,6 @@ public class PredicateBuilder {
                 int scale = decimalType.getScale();
                 return Decimal.fromBigDecimal((BigDecimal) o, precision, scale);
             case TIMESTAMP_WITHOUT_TIME_ZONE:
-            case TIMESTAMP_WITH_LOCAL_TIME_ZONE:
                 Timestamp timestamp;
                 if (o instanceof java.sql.Timestamp) {
                     timestamp = Timestamp.fromSQLTimestamp((java.sql.Timestamp) o);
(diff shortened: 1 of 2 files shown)


paimon-common/src/main/java/org/apache/paimon/predicate/PredicateBuilder.java
```diff
@@ -278,8 +279,20 @@ public static Object convertJavaObject(DataType literalType, Object o) {
...
                 return Decimal.fromBigDecimal((BigDecimal) o, precision, scale);
             case TIMESTAMP_WITHOUT_TIME_ZONE:
-            case TIMESTAMP_WITH_LOCAL_TIME_ZONE:
                 Timestamp timestamp;
+                if (o instanceof java.sql.Timestamp) {
+                    timestamp = Timestamp.fromSQLTimestamp((java.sql.Timestamp) o);
+                } else if (o instanceof Instant) {
+                    timestamp =
+                            Timestamp.fromInstant(
+                                    ((Instant) o).plusMillis(TimeUnit.HOURS.toMillis(8)));
```
JingsongLi, line 288: Don't hard code, please take a look to `DateTimeUtils`.

JingsongLi (changes requested):
Please add tests.

paimon-common/src/main/java/org/apache/paimon/predicate/PredicateBuilder.java
xuzifu666: OK

JingsongLi (approved):
+1

PR #3731, opened 2024-07-11
[core] Introduce IcebergCommitCallback to create Iceberg metadata after commit

tsreaper:
### Purpose

Iceberg is a popular data lake format. In order to have more users try out Paimon while keeping their existing Iceberg workflow, we introduce `IcebergCommitCallback` in this PR to create Iceberg metadata after commit.

With this feature, Iceberg readers can read Paimon's `RawFile`s.

### Tests

* Unit tests

### API and Format

No changes in Paimon format.

### Documentation

Yes. Document will be added later.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/iceberg/IcebergPathFactory.java
+++ b/paimon-core/src/main/java/org/apache/paimon/iceberg/IcebergPathFactory.java
@@ -1,0 +1,93 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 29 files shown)


paimon-hive/paimon-hive-connector-common/src/main/java/org/apache/paimon/hive/SearchArgumentToPredicateConverter.java
```diff
@@ -145,23 +145,24 @@ private Predicate convertLeaf(PredicateLeaf leaf) {
         DataType columnType = columnTypes.get(idx);
         switch (leaf.getOperator()) {
             case EQUALS:
-                return builder.equal(idx, toLiteral(columnType, leaf.getLiteral()));
```
JingsongLi, line 148: Please create a sparate PR for this.

paimon-core/src/main/java/org/apache/paimon/iceberg/IcebergPathFactory.java
```diff
@@ -0,0 +1,93 @@
...
+
+import org.apache.paimon.fs.Path;
+import org.apache.paimon.utils.PathFactory;
+
+import java.util.UUID;
+
+/** Path factory for Iceberg metadata files. */
+public class IcebergPathFactory {
+
+    private final Path root;
```
JingsongLi, line 29: just store `metadataDirectory`? to avoid exposing root to the entire class

JingsongLi (approved):
Looks good to me!

PR #3736, opened 2024-07-12
[flink] Consumer expiration time must be specified when using consumer-id

discivigour:
### Purpose

Linked issue: close #xxx

### Tests

- IT case: CatalogTableITCase


herefree:
Do we have to add this verification to the code? When we upgrade the paimon version, this verification will cause historical flink tasks restart fail. @tsreaper @discivigour

PR #3744, opened 2024-07-12
 [spark] refactor delete with deletion vector

YannByron:
### Purpose

Linked issue: close #xxx


JingsongLi:
Hi @YannByron , Can you explain why the refactoring was done?

YannByron:
The old logic is to write the unchanged deletion vector in driver side, which has a bad performance. Now, we can do this in parallel by the granularity of `partition` and `bucket`.

JingsongLi (approved):
Looks good to me!

PR #3746, opened 2024-07-13
[spark] Support MarkPartitionDoneProcedure in Spark

xuzifu666:
### Purpose

Currently Spark not support MarkPartitionDoneProcedure for users，some users may only support spark in table manager，so need support it in spark 

Linked issue: close https://github.com/apache/paimon/issues/3747


paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/procedure/actions/AddDonePartitionAction.java
```diff
@@ -0,0 +1,60 @@
...
+import org.apache.paimon.shade.guava30.com.google.common.collect.Iterators;
+
+import java.io.IOException;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+import static org.apache.paimon.utils.PartitionPathUtils.extractPartitionSpecFromPath;
+
+/** A {@link PartitionMarkDoneAction} which add ".done" partition. */
+public class AddDonePartitionAction implements PartitionMarkDoneAction {
```
JingsongLi, line 33: Please reuse classed in Flink. You should move classes to core and reuse them.

xuzifu666: OK，would change it as the advice. Thanks @JingsongLi

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/procedure/MarkPartitionDoneProcedure.java
```diff
@@ -0,0 +1,165 @@
...
+                }
+            } catch (Exception e) {
+                throw new RuntimeException(e);
+            }
+        }
+    }
+
+    static List<PartitionMarkDoneAction> createActions(
+            FileStoreTable fileStoreTable, CoreOptions options) {
+        return Arrays.asList(
```
JingsongLi, line 128: Please reuse method too.

paimon-core/src/main/java/org/apache/paimon/partition/actions/PartitionMarkDoneAction.java
```diff
@@ -16,31 +16,38 @@
...
+    String DEFAULT_MARK_DONE_ACTION = "success-file";
+
     void markDone(String partition) throws Exception;
 
     static List<PartitionMarkDoneAction> createActions(
             FileStoreTable fileStoreTable, CoreOptions options) {
-        return Arrays.asList(options.toConfiguration().get(PARTITION_MARK_DONE_ACTION).split(","))
-                .stream()
+        String markdoneAction =
+                StringUtils.isBlank((options.toConfiguration().get(PARTITION_MARK_DONE_ACTION)))
```
JingsongLi, line 46: Why we need to handle blank?

xuzifu666: Because in Flink handle logic would process in default value in PARTITION_MARK_DONE_ACTION conf if it is not set，but the config is in Flink module not in core or spark moudle，so keep the common logic like this way.

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -1028,6 +1028,25 @@ public class CoreOptions implements Serializable {
                             "Parameter string for the constructor of class #. "
                                     + "Callback class should parse the parameter by itself.");
 
+    public static final ConfigOption<String> PARTITION_MARK_DONE_ACTION =
```
JingsongLi, line 1031: Regenerate doc?

xuzifu666: OK，would move it to core_configuration.html

JingsongLi (approved):
+1

PR #3749, opened 2024-07-15
[core] Disable snapshot.clean-empty-directories by default

JingsongLi:
### Purpose

Object store delete directory will list files for directory, it is very costly. We can not delete directory for every snapshot.


yuzelin (approved):
+1

PR #3750, opened 2024-07-15
[core] add default implementation to FileIndexResult and/or operation.

hang8929201:
### Purpose

add default implementation to FileIndexResult and/or operation.


JingsongLi (approved):
+1

PR #3751, opened 2024-07-15
[core] Introduce delete-file.thread-num

JingsongLi:
### Purpose

For object storage, the larger the concurrency to delete files, the better the overall deletion throughput.


yuzelin (approved):
+1

PR #3755, opened 2024-07-16
[core] fix the collect of overlapped snapshots

askwang:
### Purpose

When call `findOverlappedSnapshots(List<Snapshot> taggedSnapshots, long beginInclusive, long endExclusive)`.
If beginInclusive is 10, endExclusive is 15, the range is [10, 15), and the taggedSnapshots is [8, 9, 11, 12, 15].
In previous way, the overlapped snapshots is [9, 11, 12]. Actually is only [11, 12].


JingsongLi:
Hi @askwang , can you add test cases?

askwang:
> Hi @askwang , can you add test cases?

test added.

askwang:
@JingsongLi can you help take a look, ths

PR #3757, opened 2024-07-16
[core] Support using table branches with $branch_ and support different schemas for different branches

tsreaper:
### Purpose

Schema of branches are stored in separated directories. However currently Paimon does not support different schemas for different branches.

This PR supports this feature, and also support using table branches with `<table_name>$branch_<branch_name>`. This usage is similar to Iceberg's branches.

We recommend this usage instead of `/*+ OPTIONS('branch' = '<branch_name>) */` in Flink. Because if we use dynamic option and the target branch has different schema from the main branch, then Flink sink and catalog will have different schemas and it cannot pass Flink SQL's validation.

### Tests

IT cases.

### API and Format

No changes.

### Documentation

Yes.

### DIFF
--- a/docs/content/maintenance/manage-branches.md
+++ b/docs/content/maintenance/manage-branches.md
@@ -117,10 +117,10 @@ You can read or write with branch as below.
 
 ```sql
 -- read from branch 'branch1'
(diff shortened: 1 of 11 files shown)


JingsongLi (approved):
+1

PR #3760, opened 2024-07-16
[common] Add some methods to RoaringBitmap32

hang8929201:
### Purpose

Add and/or operation to RoaringBitmap32.

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/utils/RoaringBitmap32.java
+++ b/paimon-common/src/main/java/org/apache/paimon/utils/RoaringBitmap32.java
@@ -40,7 +40,7 @@ public class RoaringBitmap32 {
         this.roaringBitmap = new RoaringBitmap();
     }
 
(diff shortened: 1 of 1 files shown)


hang8929201:
add more methods

paimon-common/src/main/java/org/apache/paimon/utils/RoaringBitmap32.java
```diff
@@ -39,6 +40,10 @@ public RoaringBitmap32() {
         this.roaringBitmap = new RoaringBitmap();
     }
 
+    public RoaringBitmap32(RoaringBitmap roaringBitmap) {
```
JingsongLi, line 43: We should not support this, `RoaringBitmap` should be hided.

hang8929201: okay, can I just change it to private?

JingsongLi (approved):
+1

PR #3761, opened 2024-07-16
[rename] Extract assertMainBranch to AbstractCatalog

yuzelin:
### Purpose

`assertMainBranch` can be a common method.


JingsongLi (approved):
+1

PR #3765, opened 2024-07-16
[hotfix] Fix unstable Hive connector ut

xuzifu666:
### Purpose

Fix unstable Hive connector ut

hms port is 9083 in RepairActionITCase not stable because other test also use it，so change it to 9082.

Linked issue: close #xxx


JingsongLi:
+1

PR #3768, opened 2024-07-17
 Compatible with Spark4

awol2005ex:
Compatible with Spark4


JingsongLi (changes requested):
Hi @awol2005ex 
We should avoid copying files.

awol2005ex:
> Hi @awol2005ex We should avoid copying files.

主要是要改的scala太多了

JingsongLi:
> > Hi @awol2005ex We should avoid copying files.
> 
> 主要是要改的scala太多了

Please use English (This is Apache's rule)

JingsongLi:
You can discuss this in dev mail list. We can figure out how to support Spark 4.0.

JingsongLi:
The changes are so huge... We should find a easy way.

PR #3769, opened 2024-07-17
[cdc] Adjust-CdcJsonDeserializationSchema

MOBIN-F:
### Purpose

The serialization entry of cdc kafka_table/database_sync is CdcJsonDeserializationSchema, but CdcJsonDeserializationSchema#deserialize returns CdcSourceRecord(JsonNode). When users extend other format, there will be limitations. Most data serialization is based on byte[], not JsonNode. 
For example: 
Users need to use other serialization protocols to serialize data based on byte[], such as binlog of other databases.


MOBIN-F:
closed, recommended to use flink-cdc Paimon Pipeline Connector first

PR #3770, opened 2024-07-17
[core] Introduce a basic SortLookupStoreFactory

JingsongLi:
### Purpose

Multiple experiments have shown that the performance bottleneck of local files during lookup lies in disk read and write, and file format affects file size. Generally speaking, sorting based implementations can achieve relatively high compression rates.

This PR introduces a sorting implementation that is very similar to the file read and write implementation of levelDB, but with some differences:
1. It does not consider prefix compression for keys, as `zstd` compression automatically considers these situations, and currently our key format (see `RowCompactedSerializer`) is not friendly to prefix compression.
2. Introduced alignment index for Blocks, which can avoid storing the length of each key value.

### API and Format

This PR has not yet fully implemented sorting based files, so it only provides an option and defaults to HASH implementation.


paimon-common/src/main/java/org/apache/paimon/lookup/sort/SortLookupStoreWriter.java
```diff
@@ -0,0 +1,212 @@
...
+    SortLookupStoreWriter(
+            File file,
+            int blockSize,
+            @Nullable BloomFilter.Builder bloomFilter,
+            @Nullable BlockCompressionFactory compressionFactory)
+            throws IOException {
+        this.fileOutputStream = new BufferedOutputStream(Files.newOutputStream(file.toPath()));
+        this.blockSize = blockSize;
+        this.dataBlockWriter = new BlockWriter((int) (blockSize * 1.1));
+        int expectedNumberOfBlocks = 1024;
```
FangYongs, line 79: Currently, the default file size for primary key tables is 128MB. If we use 64KB as the block size, the number of blocks might be around 2048. So, could the expected number of blocks be 2048?

paimon-common/src/main/java/org/apache/paimon/lookup/sort/Footer.java
```diff
@@ -0,0 +1,101 @@
...
+import static org.apache.paimon.lookup.sort.SortLookupStoreWriter.MAGIC_NUMBER;
+import static org.apache.paimon.utils.Preconditions.checkArgument;
+
+/** Footer for a sorted file. */
+public class Footer {
+
+    public static final int ENCODED_LENGTH = 36;
+
+    @Nullable private final BloomFilterHandle bloomFilterHandle;
+    private final BlockHandle indexBlockHandle;
```
FangYongs, line 38: A version field should be reserved here so that we can easily upgrade the version number in the future and be compatible with old versions?

JingsongLi: Is magic number something like version? Maybe we can add version field too.

paimon-common/src/main/java/org/apache/paimon/lookup/sort/BlockWriter.java
```diff
@@ -0,0 +1,99 @@
...
+import org.apache.paimon.utils.IntArrayList;
+
+import java.io.IOException;
+
+import static org.apache.paimon.lookup.sort.BlockAlignedType.ALIGNED;
+import static org.apache.paimon.lookup.sort.BlockAlignedType.UNALIGNED;
+
+/** Writer to build a Block. */
+public class BlockWriter {
+
```
FangYongs, line 32: How about adding a header for Block that contains BlockType and onDiskSize. 
For BlockType, we can easily support multiple level block index like HFile. 
For onDiskSize, we can scan block data directly without reading it from block index for each block.

paimon-common/src/main/java/org/apache/paimon/lookup/sort/Footer.java
FangYongs: It seems a little different. Currently, `magic_number` is unique and unchanged, mainly used to verify whether the data file is correct; `version` may have multiple versions as we upgrade the storage structure, and different data processing will be performed according to different `version`s.

JingsongLi: Sounds good to me.

paimon-common/src/main/java/org/apache/paimon/lookup/sort/SortLookupStoreWriter.java
JingsongLi: Perhaps even larger, currently it is a column storage file converted to row storage

paimon-common/src/main/java/org/apache/paimon/lookup/sort/BlockWriter.java
JingsongLi: Currently, there is already a `BlockAligndType` at the end of the block. But it is indeed extensible.

FangYongs: Do you imply that we can extend `BlockAlignType` to `DATA_ALIGN`, `DATA_UNALIGN`, `INDEX`, etc.? Cool! How about write this `byte` in header of block with `onDiskSize`?

FangYongs:
Thanks @JingsongLi , big +1 for this PR. We can continue to improve and support the key-value format on it 👍🏻

paimon-common/src/main/java/org/apache/paimon/lookup/sort/BlockWriter.java
JingsongLi: > Do you imply that we can extend BlockAlignType to DATA_ALIGN, DATA_UNALIGN, INDEX, etc.?

Yes, but first we should figure out what is multiple level block index, and what is benifit.

> How about write this byte in header of block with onDiskSize?

I cannot understand good about `we can scan block data directly without reading it from block index for each block`.

> footer or header

I don't have any bias, they all seem the same as long as they don't affect the writing process

PR #3771, opened 2024-07-17
[core] Add latestSnapshotId to Table

JingsongLi:
### Purpose

API users have requirement to take a look to latest snapshot id. And maybe use scan.snapshot-id to read data with idempotent.

### API and Format

Add Table.latestSnapshotId public interface.


leaves12138:
+1

PR #3777, opened 2024-07-18
[WIP][core] Fix thread safety bug in Partial-update with agg

JingsongLi:
### Purpose

Partial-update with nested-update agg has thread safety bug.
The nested-update agg has internal state in Projection class, but it is shared by multiple Partial-update function.

### Tests

TODO


leaves12138 (approved):
+1

PR #3779, opened 2024-07-18
[core] fix CommitterOperator chain value when sink.committer-operator-chaining is false

wwj6591812:
when sink.committer-operator-chaining'='false',  an extra node appeared in our flink job.
This is a bug, fix it.

image

### Purpose

Linked issue: close #xxx


JingsongLi:
+1

PR #3780, opened 2024-07-18
[hive] migrate_file support save source table

xuzifu666:
### Purpose
Currently migrate_file would delete source table after migrate，but some condition user hope save origin table.The pr is aim to support it.

Linked issue: close https://github.com/apache/paimon/issues/3781


JingsongLi (approved):
+1

PR #3782, opened 2024-07-18
[core] Lookup wait should affect deletion vectors mode and first row engine

JingsongLi:
### Purpose

1. Rename `changelog-producer.lookup-wait` to `lookup-wait`.
2. Lookup wait should affect deletion vectors mode and first row engine.
3. `AsyncLookupSinkWrite` should be used to deletion vectors mode and first row engine.


leaves12138 (approved):
+1

PR #3785, opened 2024-07-19
[core] Remove duplicate ConfigOption.withDeprecatedKeys

JingsongLi:
### Purpose

In ConfigOption, withDeprecatedKeys is very similar with withFallbackKeys. We can just keep only one.


yuzelin (approved):
+1

PR #3787, opened 2024-07-20
[core] Cleaning snapshots or deleting tags should skip those referenced by branches.

LinMingQiang:
### Purpose

Next part for #3808

The purpose of this pr : 

The purpose of the branch is to asynchronously recover data and overwrite the main branch 
[paimon branch](https://cwiki.apache.org/confluence/display/PAIMON/PIP-9%3A+Support+Branch ) . If the recovery operation takes too long, the snapshot referenced by the branch may be expired and deleted, which will make the branch unavailable.
Let's say we created tag1 and branch1 based on snapshot-1, when snapshot-1 expires then the related `manifest file` will be deleted. However ,if the `manifest file` is referenced by the some tag, it will not be cleaned up.
But if the `manifest file` is still referenced by branch-1, it will still be cleaned, this led to query branch1 will throw `manifets file` not found exception.

Solution :

1: Skip cleanup :   Merge snapshots referenced by branch or tag in the code of `ExpireSnapshotsImpl#expireUntil` then skip the files associated with these snapshots.
2: When to clean : Referring to the logic of `delete tag`, when we delete the branch, we also clean up the expired snapshot files (manifest files and data files).
3: delete tag/branch :  Now, when deleting a tag or branch, we need to consider whether the snapshot involved is still referenced by other tags/branches.

Linked issue: close #xxx

### API and Format

`TagManager#deleteTag` :  Add logical , whether the snapshot is referenced by other branch, if so, the `doClean` of the snapshot should be skipped.

`BranchManager#deleteBranch` :  Add logical , If the snapshot has expired and is not referenced by another branch or tag, trigger `doClean` to delete manifest&data file.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/FileDeletionBase.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/FileDeletionBase.java
@@ -344,6 +344,19 @@ public abstract class FileDeletionBase<T extends Snapshot> {
         return entry -> false;
     }
 
+    public Predicate<ManifestEntry> dataFileSkipper(Snapshot skippingSnapshot) throws Exception {
+        return dataFileSkipper(Collections.singletonList(skippingSnapshot));
+    }
+
+    public Predicate<ManifestEntry> dataFileSkipper(List<Snapshot> skippingSnapshots)
+            throws Exception {
+        Map<BinaryRow, Map<Integer, Set<String>>> skipped = new HashMap<>();
+        for (Snapshot snapshot : skippingSnapshots) {
+            addMergedDataFiles(skipped, snapshot);
+        }
+        return entry -> containsDataFile(skipped, entry);
+    }
+
     /**
      * It is possible that a job was killed during expiration and some manifest files have been
      * deleted, so if the clean methods need to get manifests of a snapshot to be cleaned, we should
--- a/paimon-core/src/main/java/org/apache/paimon/operation/FileDeletionBase.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/FileDeletionBase.java
@@ -488,4 +488,21 @@ public abstract class FileDeletionBase<T extends Snapshot> {
             throw new RuntimeException(e);
         }
     }
+
+    public Set<Path> extractDataFilePaths(
+            Collection<ManifestEntry> manifestEntries, Predicate<ManifestEntry> skipper) {
+        Set<Path> dataFileToDelete = new HashSet<>();
+        for (ManifestEntry entry : manifestEntries) {
+            if (!skipper.test(entry)) {
+                Path bucketPath = pathFactory.bucketPath(entry.partition(), entry.bucket());
+                dataFileToDelete.add(new Path(bucketPath, entry.file().fileName()));
+                for (String file : entry.file().extraFiles()) {
+                    dataFileToDelete.add(new Path(bucketPath, file));
+                }
+
+                recordDeletionBuckets(entry);
+            }
+        }
+        return dataFileToDelete;
+    }
 }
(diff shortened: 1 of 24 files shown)


JingsongLi:
If the snapshot is not deleted here, when will it be cleared?

BsoBird:
> If the snapshot is not deleted here, when will it be cleared?

@JingsongLi @LinMingQiang  Maybe we both need to merge our PRs.

paimon-core/src/main/java/org/apache/paimon/operation/FileDeletionBase.java
```diff
@@ -329,7 +329,7 @@ protected void cleanUnusedManifests(
 
     public Predicate<ManifestEntry> dataFileSkipper(
             List<Snapshot> taggedSnapshots, long expiringSnapshotId) throws Exception {
-        int index = TagManager.findPreviousTag(taggedSnapshots, expiringSnapshotId);
+        int index = TagManager.findPreviousSnapshot(taggedSnapshots, expiringSnapshotId);
```
JingsongLi, line 332: Why rename this to snapshot?

JingsongLi:
Please rebase master.

LinMingQiang:
> Please rebase master.

done.

paimon-core/src/main/java/org/apache/paimon/utils/SnapshotManager.java
```diff
@@ -696,6 +698,50 @@ private static int findPreviousOrEqualSnapshot(
...
+        if (index + 1 < snapshotList.size()) {
+            Snapshot rightSnapshot = snapshotList.get(index + 1);
+            nearestRight = nearestRight.id() < rightSnapshot.id() ? nearestRight : rightSnapshot;
+        }
+        skippedSnapshots.add(nearestRight);
+        return skippedSnapshots;
+    }
+
+    public static List<Snapshot> mergeTreeSetToList(
+            Collection<Snapshot> set1, Collection<Snapshot> set2) {
```
JingsongLi, line 733: Where is the tree set?

sortToList?

paimon-core/src/main/java/org/apache/paimon/utils/BranchManager.java
```diff
@@ -316,4 +388,45 @@ public List<String> branches() {
...
+        for (String branchName : branches()) {
+            Snapshot branchCreateSnapshot =
+                    snapshotManager.copyWithBranch(branchName).earliestSnapshot();
+            // Ignore the empty branch.
+            if (branchCreateSnapshot != null) {
+                FileStoreTable branchTable =
+                        FileStoreTableFactory.create(
+                                fileIO, new Path(branchPath(tablePath, branchName)));
+                SortedMap<Snapshot, List<String>> snapshotTags = branchTable.tagManager().tags();
+                Snapshot earliestSnapshot = branchTable.snapshotManager().earliestSnapshot();
```
JingsongLi, line 406: What is difference betweend branchCreateSnapshot and this earliestSnapshot?

paimon-core/src/main/java/org/apache/paimon/utils/BranchManager.java
```diff
@@ -316,4 +388,45 @@ public List<String> branches() {
...
+                            .computeIfAbsent(earliestSnapshot, s -> new ArrayList<>())
+                            .add(branchName);
+                } else {
+                    Snapshot snapshot = snapshotTags.firstKey();
+                    // current branch is create from tag.
+                    if (earliestSnapshot.id() == snapshot.id()) {
+                        List<String> tags = snapshotTags.get(snapshot);
+                        checkArgument(tags.size() == 1);
+                        sortedSnapshots
+                                .computeIfAbsent(snapshot, s -> new ArrayList<>())
```
JingsongLi, line 419: Why get snapshot from tags?

paimon-core/src/main/java/org/apache/paimon/utils/SnapshotManager.java
LinMingQiang: Change method name to  'mergeSetToSortedList'

JingsongLi:
@LinMingQiang Can we remove this feature? Branches can only be created through tags, or completely new branch (without inheriting data from the main branch).

PR #3792, opened 2024-07-22
[Feature][Flink] Make FlinkCatalogFactory pluggable

yunfengzhou-hub:
### Purpose

This PR makes FlinkCatalogFactory pluggable.

### Tests

Unit tests are added in CompactionITCase to cover the introduced code.

### API and Format

This change does not affect format, and the introduced APIs are not for public usage for now.

### Documentation

This change introduces changes that does not need document.

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkCatalogOptions.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkCatalogOptions.java
@@ -27,6 +27,11 @@ import java.time.Duration;
 
 /** Options for flink catalog. */
 public class FlinkCatalogOptions {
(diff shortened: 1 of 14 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkCatalogOptions.java
```diff
@@ -27,6 +27,11 @@
 
 /** Options for flink catalog. */
 public class FlinkCatalogOptions {
+    @ExcludeFromDocumentation("Only for internal use")
+    public static final ConfigOption<String> CATALOG_TYPE =
```
JingsongLi, line 31: We should avoid introducing this option.

JingsongLi:
Thanks @yunfengzhou-hub , it seems we don't have requirement to do this abstraction.

PR #3798, opened 2024-07-23
[core] Optimize MergeSorter to avoid spill all data

JingsongLi:
### Purpose

In the previous behavior, once the number of readers exceeds the threshold, it will cause all the data from the readers to be stored on the disk, which may result in a significant amount of disk usage.

We can optimize this logic by simply spilling the extra readers onto the disk.


leaves12138 (approved):
+1

PR #3805, opened 2024-07-24
[core] Support commit isolation level

BsoBird:
### Purpose
For the file system catalog, there are many situations where we need to support stricter commit policies.This is especially true for third-party databases that use paimon as their underlying storage.

We define the commit isolation level option to control the behavior.

Linked issue: none

1. Disable retry after failed commits in strict mode
2. In strict mode, reject and clean up potentially dirty commits.
3. Categorized possible exceptions and adjusted the exception handling logic.
4. Supports concurrency control.
5. In strict mode, do not use any hint file.

### Tests

See FileStoreCommitTest.java.

### API and Format

None

#### Glossary:
- Dirty-commit: commitSnapshotId < CurrentLatestSnapshotId  - snapshot.num-retained.max && commit-success=True && snapshotManager.exists(commitSnapshotId)=True && (commitSnapshotId not belong any tag/branch)

#### Flow chat (use strict mode):
```mermaid
    graph TD
START[startCommit] -->USE-LOCK{USE-LOCK?}
USE-LOCK -->|YES| COMMIT_WITH_LOCK{COMMIT_WITH_LOCK_SUCCESS}
USE-LOCK -->|NO| CHECK-BEFORE{CHECK-BEFORE-SUCCESS?}
CHECK-BEFORE -->|NO| COMMIT_FAILED
CHECK-BEFORE -->|YES| COMMIT{COMMIT_SUCCESS?}
COMMIT --> |commit state unknown| COMMIT_FAILED
COMMIT --> |not success| COMMIT_FAILED
COMMIT --> |YES| IS_DIRTY_COMMIT{IS_DIRTY_COMMIT?}
IS_DIRTY_COMMIT --> |YES| COMMIT_FAILED
IS_DIRTY_COMMIT --> |NO| COMMIT_SUCCESS
COMMIT_WITH_LOCK --> |failed-normal| RETRY
COMMIT_WITH_LOCK --> |failed - strict mode| COMMIT_FAILED
COMMIT_WITH_LOCK --> |success| COMMIT_SUCCESS
RETRY --> END
COMMIT_SUCCESS --> END
COMMIT_FAILED --> END
```

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/fs/FileIO.java
+++ b/paimon-common/src/main/java/org/apache/paimon/fs/FileIO.java
@@ -238,6 +238,9 @@ public interface FileIO extends Serializable {
         try {
             writeFile(tmp, content, false);
             success = rename(tmp, path);
+        } catch (IOException e) {
+            // try check once
+            success = exists(path) && !exists(tmp);
         } finally {
             if (!success) {
                 deleteQuietly(tmp);
--- a/paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
@@ -18,10 +18,14 @@
 
 package org.apache.paimon.operation;
 
+import org.apache.paimon.CoreOptions;
 import org.apache.paimon.Snapshot;
 import org.apache.paimon.annotation.VisibleForTesting;
 import org.apache.paimon.data.BinaryRow;
 import org.apache.paimon.data.InternalRow;
+import org.apache.paimon.exception.CommitFailedException;
+import org.apache.paimon.exception.CommitStateUnknownException;
+import org.apache.paimon.exception.DirtyCommitException;
 import org.apache.paimon.fs.FileIO;
 import org.apache.paimon.fs.Path;
 import org.apache.paimon.io.DataFileMeta;
@@ -52,12 +56,14 @@ import org.apache.paimon.utils.FileStorePathFactory;
 import org.apache.paimon.utils.Pair;
 import org.apache.paimon.utils.Preconditions;
 import org.apache.paimon.utils.SnapshotManager;
+import org.apache.paimon.utils.StringUtils;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nullable;
 
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -65,6 +71,7 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -73,6 +80,8 @@ import java.util.Set;
 import java.util.concurrent.Callable;
 import java.util.stream.Collectors;
 
+import static org.apache.paimon.CoreOptions.COMMIT_ISOLATION_LEVEL;
+import static org.apache.paimon.CoreOptions.SNAPSHOT_NUM_RETAINED_MAX;
 import static org.apache.paimon.deletionvectors.DeletionVectorsIndexFile.DELETION_VECTORS_INDEX;
 import static org.apache.paimon.index.HashIndexFile.HASH_INDEX;
 import static org.apache.paimon.partition.PartitionPredicate.createPartitionPredicate;
@@ -130,6 +139,7 @@ public class FileStoreCommitImpl implements FileStoreCommit {
     private final StatsFileHandler statsFileHandler;
 
     private final BucketMode bucketMode;
+    private final Map<String, String> tableOptions;
 
     public FileStoreCommitImpl(
             FileIO fileIO,
@@ -152,7 +162,8 @@ public class FileStoreCommitImpl implements FileStoreCommit {
             String branchName,
             StatsFileHandler statsFileHandler,
             BucketMode bucketMode,
-            @Nullable Integer manifestReadParallelism) {
+            @Nullable Integer manifestReadParallelism,
+            Map<String, String> tableOptions) {
         this.fileIO = fileIO;
         this.schemaManager = schemaManager;
         this.commitUser = commitUser;
@@ -178,6 +189,7 @@ public class FileStoreCommitImpl implements FileStoreCommit {
         this.commitMetrics = null;
         this.statsFileHandler = statsFileHandler;
         this.bucketMode = bucketMode;
+        this.tableOptions = tableOptions;
     }
 
     @Override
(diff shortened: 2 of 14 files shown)


BsoBird:
@JingsongLi  Hi. Can you check this? Tks.

paimon-common/src/main/java/org/apache/paimon/fs/FileIO.java
```diff
@@ -238,6 +238,9 @@ default boolean tryToWriteAtomic(Path path, String content) throws IOException {
         try {
             writeFile(tmp, content, false);
             success = rename(tmp, path);
+        } catch (IOException e) {
+            // try check once
+            success = exists(path) && !exists(tmp);
```
BsoBird, line 243: After throwing an IO exception, it's a good idea to check it again.Currently, this checking strategy is not fully applicable to object storage systems and may result in false positives.However, as the industry evolves, object storage systems are starting to support mutex operations.In the long run, I think we need to add check operations.

paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
```diff
@@ -956,46 +986,46 @@ public boolean tryCommitOnce(
...
+                    oldMetas,
+                    newMetas,
+                    changelogMetas);
+            throw e;
+        } catch (Exception e) {
+            // If the commit succeeds, we ignore all exceptions thrown by the successful commit;
+            // otherwise, we throw an exception.
+            if (!success && useSerializableIsolation) {
+                throw new CommitFailedException(e.getMessage(), e);
+            }
```
BsoBird, line 1017: If the rename operation succeeds, we should ignore any exceptions thrown after that.

paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
```diff
@@ -1274,4 +1308,101 @@ static ConflictCheck noConflictCheck() {
...
+            Snapshot newSnapshot,
+            boolean useSerializableIsolation,
+            boolean useLock) {
+        try {
+            boolean committed = tryAtomicCommitNewSnapshot(newSnapshotPath, newSnapshot);
+            if (committed) {
+                if (useSerializableIsolation && !useLock) {
+                    // If the useSerializableIsolation is turned on,And we not use lock service,
+                    // we will remove all HINTs,as it may provide incorrect information.
+                    snapshotManager.removeSnapshotEarliestHint();
```
BsoBird, line 1324: If Serializable is turned on, and the locking service is not used, we should delete all HINT files, as they are always unreliable.

paimon-core/src/main/java/org/apache/paimon/utils/BranchManager.java
```diff
@@ -284,7 +284,7 @@ public void fastForward(String branchName) {
                             .collect(Collectors.toList());
 
             // Delete latest snapshot hint
-            snapshotManager.deleteLatestHint();
+            snapshotManager.removeSnapshotLatestHint();
```
BsoBird, line 287: The previous method wasn't clear enough, so I've replaced it with

paimon-core/src/test/java/org/apache/paimon/TestFileStore.java
```diff
@@ -363,7 +363,9 @@ public List<Snapshot> commitDataImpl(
 
         List<Snapshot> snapshots = new ArrayList<>();
         for (long id = snapshotIdBeforeCommit + 1; id <= snapshotIdAfterCommit; id++) {
-            snapshots.add(snapshotManager.snapshot(id));
+            if (snapshotManager.snapshotExists(id)) {
+                snapshots.add(snapshotManager.snapshot(id));
+            }
```
BsoBird, line 368: If a dirty commit occurs, the previous processing logic throws an exception, which affects the running of the test case.

paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
```diff
@@ -1274,4 +1308,101 @@ static ConflictCheck noConflictCheck() {
...
+                    snapshotManager.removeSnapshotLatestHint();
+                    snapshotManager.removeChangeLogEarliestHint();
+                    snapshotManager.removeChangeLogLatestHint();
+                } else {
+                    snapshotManager.commitLatestHint(newSnapshotId);
+                }
+            }
+            return committed;
+        } catch (Exception e) {
+            throw new CommitStateUnknownException(e.getMessage(), e);
```
BsoBird, line 1334: If any exception occurs during the commit process, we should throw a CommitStateUnknownException and not clean up any data.

paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
```diff
@@ -956,46 +986,46 @@ public boolean tryCommitOnce(
...
-                            newSnapshotId,
-                            newSnapshotPath,
-                            commitUser,
-                            identifier,
-                            commitKind.name()),
-                    e);
+            throw e;
+        } catch (DirtyCommitException e) {
+            // We need to clean up all the metadata information generated by this commit.
+            cleanUpTmpManifests(
```
BsoBird, line 1003: Maybe cleaning up dirty commits will interfere with the creation of branch/tag. But since we are always cleaning up snapshots that exceed `snapshot.num-retained.max`.This shouldn't have much of an impact. After all, if we keep a snapshot that matches the characteristics of a dirty commit, we may never find the right time to delete it.

paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
```diff
@@ -1274,4 +1308,101 @@ static ConflictCheck noConflictCheck() {
...
+            maxSaveSnapshotNum = Integer.parseInt(maxSaveSnapshotsNumStr);
+        }
+        long latestSnapshotIdAfterCommit =
+                Optional.ofNullable(snapshotManager.latestSnapshotIdWithOutHint())
+                        .orElse(Snapshot.FIRST_SNAPSHOT_ID);
+        long waterMark =
+                latestSnapshotIdAfterCommit > maxSaveSnapshotNum
+                        ? latestSnapshotIdAfterCommit - maxSaveSnapshotNum
+                        : -1;
+        fastFailIfDirtyCommit(newSnapshotId, waterMark, latestSnapshotIdAfterCommit);
```
BsoBird, line 1372: Here we need to add the logic of whether the snapshot belongs to a branch or tag.PR 3787 provides this functionality, I'm not sure if I need to merge it with that PR?

BsoBird: For now I'm starting by porting some of the code from 3787 to the current PR.

JingsongLi:
Thanks @BsoBird ! I need time to look this, next week will give your feedback.

paimon-common/src/main/java/org/apache/paimon/fs/FileIO.java
JingsongLi, line 243: This can be a separate PR?

JingsongLi: This can be a separate PR?

JingsongLi (changes requested):
Hi @BsoBird , I have no idea why the changes is so big...

In my imagination, we only need one branch to complete the new commit mode.
Please note that this is the core code, and any modifications may result in serious consequences.

paimon-common/src/main/java/org/apache/paimon/fs/FileIO.java
BsoBird: > This can be a separate PR?

sure

JingsongLi:
Close this PR because long it is no longer active. Feel free to open a discussion in dev mail list.

PR #3806, opened 2024-07-24
[core] Refactor exception handling of PartitionTimeExtractor

yuzelin:
### Purpose

When using `partition markdone`, if some partitions are non-date formatted, currently `PartitionTimeExtractor` will returns `null` and omits the exception. This PR refactor it for better trouble shooting.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/partition/PartitionTimeExtractor.java
+++ b/paimon-core/src/main/java/org/apache/paimon/partition/PartitionTimeExtractor.java
@@ -18,11 +18,6 @@
 
 package org.apache.paimon.partition;
 
-import org.apache.paimon.CoreOptions;
-
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import javax.annotation.Nullable;
 
 import java.time.LocalDate;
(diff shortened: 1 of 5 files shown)


paimon-core/src/main/java/org/apache/paimon/partition/PartitionTimeExtractor.java
```diff
@@ -112,22 +107,31 @@ public LocalDateTime extract(List<String> partitionKeys, List<?> partitionValues
...
-                            + " use 'update-time' expiration strategy by set {}, the strategy support non-date formatted partition.",
-                    partitionInfos,
-                    this.formatter,
-                    CoreOptions.PARTITION_EXPIRATION_STRATEGY.key());
+            String message =
+                    String.format(
+                            "Can't extract datetime from partition '%s' with formatter '%s' and pattern '%s'.",
+                            partitionInfos, this.formatter, this.pattern);
+            if (ignoreException) {
+                LOG.warn(
```
JingsongLi, line 121: Move this to partition expire.

JingsongLi (approved):
+1

PR #3808, opened 2024-07-24
[core] Clarify the purpose of method findOverlappedSnapshots and move…

LinMingQiang:
… to snapshotManager.

### Purpose

Clarify the purpose of method `findOverlappedSnapshots` is to find the overlapping snapshots between snapshots that still referenced by tags/branches and range of [beginInclusive, endExclusive), not only tagSnapshots.

Linked issue: close #xxx

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/FileDeletionBase.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/FileDeletionBase.java
@@ -35,7 +35,7 @@ import org.apache.paimon.stats.StatsFileHandler;
 import org.apache.paimon.utils.FileDeletionThreadPool;
 import org.apache.paimon.utils.FileStorePathFactory;
 import org.apache.paimon.utils.Pair;
-import org.apache.paimon.utils.TagManager;
+import org.apache.paimon.utils.SnapshotManager;
 
 import org.slf4j.Logger;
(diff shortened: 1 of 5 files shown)


paimon-core/src/main/java/org/apache/paimon/operation/FileDeletionBase.java
```diff
@@ -78,10 +78,11 @@ public abstract class FileDeletionBase<T extends Snapshot> {
     protected boolean changelogDecoupled;
 
     /** Used to record which tag is cached. */
```
JingsongLi, line 80: modify comment too

paimon-core/src/main/java/org/apache/paimon/operation/FileDeletionBase.java
```diff
@@ -78,10 +78,11 @@ public abstract class FileDeletionBase<T extends Snapshot> {
...
-    private long cachedTag = 0;
+    private long cachedSnapshotId = 0;
 
     /** Used to cache data files used by current tag. */
```
JingsongLi, line 83: ditto

LinMingQiang:
done

LinMingQiang:
These unstable tests can be ignored, it was successful in my workflow.

JingsongLi (approved):
+1

PR #3814, opened 2024-07-25
[core] Add hint in local lookup file tracking with remote file

Aitozi:
### Purpose

Add the hint in local file name to track the remote the file for debugging purpose.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/KeyValueFileStoreWrite.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/KeyValueFileStoreWrite.java
@@ -104,6 +104,7 @@ public class KeyValueFileStoreWrite extends MemoryFileStoreWrite<KeyValue> {
     private final FileIO fileIO;
     private final RowType keyType;
     private final RowType valueType;
+    private final RowType partitionType;
     @Nullable private final RecordLevelExpire recordLevelExpire;
     @Nullable private Cache<String, LookupFile> lookupFileCache;
 
@@ -112,6 +113,7 @@ public class KeyValueFileStoreWrite extends MemoryFileStoreWrite<KeyValue> {
             SchemaManager schemaManager,
             TableSchema schema,
             String commitUser,
+            RowType partitionType,
             RowType keyType,
             RowType valueType,
             Supplier<Comparator<InternalRow>> keyComparatorSupplier,
@@ -136,6 +138,7 @@ public class KeyValueFileStoreWrite extends MemoryFileStoreWrite<KeyValue> {
                 deletionVectorsMaintainerFactory,
                 tableName);
         this.fileIO = fileIO;
+        this.partitionType = partitionType;
         this.keyType = keyType;
         this.valueType = valueType;
         this.udsComparatorSupplier = udsComparatorSupplier;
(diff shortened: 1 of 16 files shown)


paimon-core/src/main/java/org/apache/paimon/operation/KeyValueFileStoreWrite.java
```diff
@@ -356,13 +361,26 @@ private <T> LookupLevels<T> createLookupLevels(
...
+        for (InternalRow.FieldGetter getter : getters) {
+            Object part = getter.getFieldOrNull(partition);
+            if (part != null) {
+                builder.append(part);
+            } else {
+                builder.append("null");
+            }
+            builder.append("-");
+        }
+        String prefix = String.format("%s-%s", builder, bucket);
```
JingsongLi, line 375: I am concerned that it may be too long. Can we use its hash directly as a prefix?

JingsongLi: We can know files are come from the same partition.

Aitozi: I think the length is acceptable, here is an actual case. Hash prefix is not intuitive.

image

JingsongLi: Can we cut it off? For example, if it exceeds 20 characters, do not add any more

paimon-core/src/main/java/org/apache/paimon/operation/KeyValueFileStoreWrite.java
```diff
@@ -356,13 +361,26 @@ private <T> LookupLevels<T> createLookupLevels(
                         cacheManager,
                         new RowCompactedSerializer(keyType).createSliceComparator());
         Options options = this.options.toConfiguration();
+        InternalRow.FieldGetter[] getters = partitionType.fieldGetters();
+        StringBuilder builder = new StringBuilder();
+        for (InternalRow.FieldGetter getter : getters) {
```
JingsongLi, line 366: Extract a common method, maybe a static method in `InternalRowPartitionComputer`.

paimon-core/src/main/java/org/apache/paimon/operation/KeyValueFileStoreWrite.java
Aitozi: Sure, it's safer

paimon-core/src/test/java/org/apache/paimon/mergetree/LookupFileTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+    @Test
+    public void testLocalFilePrefix() {
+        assertThat(
+                        LookupFile.localFilePrefix(
+                                "2024073105",
+                                10,
+                                "data-ccbb95e7-8b8c-4549-8ca9-f553843d67ad-3.orc",
+                                100))
+                .isEqualTo("2024073105-10-data-ccbb95e7-8b8c-4549-8ca9-f553843d67ad-3.orc");
```
JingsongLi, line 36: I think it is better to add suffix, I mistakenly thought it was an ORC file

Aitozi: The final name will include the IOManager's file suffix `.channel`

JingsongLi (approved):
+1

PR #3816, opened 2024-07-25
[core] Support reading partition from fallback branch when not found in current branch

tsreaper:
### Purpose

In this PR we have supported reading partitions from fallback branch.

You can set the table option `scan.fallback-branch` so that when a batch job reads from the current branch, if a partition does not exist, the reader will try to read this partition from the fallback branch. For streaming read jobs, this feature is currently not supported, and will only produce results from the current branch.

What's the use case of this feature? Say you have created a Paimon table partitioned by date. You have a long-running streaming job which inserts records into Paimon, so that today's data can be queried in time. You also have a batch job which runs at every night to insert corrected records of yesterday into Paimon, so that the preciseness of the data can be promised.

When you query from this Paimon table, you would like to first read from the results of batch job. But if a partition (for example, today's partition) does not exist in its result, then you would like to read from the results of streaming job. In this case, you can create a branch for streaming job, and set `scan.fallback-branch` to this streaming branch.

Let's look at an example.

```sql
-- create Paimon table
CREATE TABLE T (
    dt STRING NOT NULL,
    name STRING NOT NULL,
    amount BIGINT
) PARTITIONED BY (dt);

-- create a branch for streaming job
CALL sys.create_branch('default.T', 'test');

-- set primary key and bucket number for the branch
ALTER TABLE `T$branch_test` SET (
    'primary-key' = 'dt,name',
    'bucket' = '2'
);

-- set fallback branch
ALTER TABLE T SET (
    'scan.fallback-branch' = 'test'
);

-- set changelog producer for the streaming branch, in case a streaming job would like to read from it in the future
ALTER TABLE `T$branch_test` SET (
    'changelog-producer' = 'lookup'
);

-- write records into the streaming branch
INSERT INTO `T$branch_test` VALUES ('20240725', 'apple', 4), ('20240725', 'peach', 10), ('20240726', 'cherry', 3), ('20240726', 'pear', 6);

-- write records into the default branch
INSERT INTO T VALUES ('20240725', 'apple', 5), ('20240725', 'banana', 7);

SELECT * FROM T;
/*
+------------------+------------------+--------+
|               dt |             name | amount |
+------------------+------------------+--------+
|         20240725 |            apple |      5 |
|         20240725 |           banana |      7 |
|         20240726 |           cherry |      3 |
|         20240726 |             pear |      6 |
+------------------+------------------+--------+
*/

-- reset fallback branch
ALTER TABLE T RESET ( 'scan.fallback-branch' );

-- now it only reads from default branch
SELECT * FROM T;
/*
+------------------+------------------+--------+
|               dt |             name | amount |
+------------------+------------------+--------+
|         20240725 |            apple |      5 |
|         20240725 |           banana |      7 |
+------------------+------------------+--------+
*/
```

### Tests

IT cases.

### API and Format

No changes in format.

### Documentation

### DIFF
--- a/docs/content/maintenance/manage-branches.md
+++ b/docs/content/maintenance/manage-branches.md
@@ -192,9 +192,13 @@ CREATE TABLE T (
 ) PARTITIONED BY (dt);
 
 -- create a branch for streaming job
--- you can even specify primary keys and bucket number, even if the original branch has no primary key
--- in this example, we create a new branch `test`, uses `dt` and `name` as primary keys, has a bucket number of 2, and will copy the table options from the original branch
-CALL sys.create_branch('default.T', 'test', 'dt, name', 2, true);
+CALL sys.create_branch('default.T', 'test');
+
+-- set primary key and bucket number for the branch
+ALTER TABLE `T$branch_test` SET (
+    'primary-key' = 'dt,name',
+    'bucket' = '2'
+);
 
 -- set fallback branch
 ALTER TABLE T SET (
--- a/docs/content/maintenance/manage-branches.md
+++ b/docs/content/maintenance/manage-branches.md
@@ -207,7 +207,7 @@ ALTER TABLE `T$branch_test` SET (
 );
 
 -- write records into the streaming branch
-INSERT INTO `T$branch_test` VALUES ('20240725', 'apple', 4), ('20240726', 'cherry', 3), ('20240726', 'pear', 6);
+INSERT INTO `T$branch_test` VALUES ('20240725', 'apple', 4), ('20240725', 'peach', 10), ('20240726', 'cherry', 3), ('20240726', 'pear', 6);
 
 -- write records into the default branch
 INSERT INTO T VALUES ('20240725', 'apple', 5), ('20240725', 'banana', 7);
--- a/docs/content/maintenance/manage-branches.md
+++ b/docs/content/maintenance/manage-branches.md
@@ -223,6 +223,20 @@ SELECT * FROM T;
 |         20240726 |             pear |      6 |
 +------------------+------------------+--------+
 */
+
+-- reset fallback branch
+ALTER TABLE T RESET ( 'scan.fallback-branch' );
+
+-- now it only reads from default branch
+SELECT * FROM T;
+/*
++------------------+------------------+--------+
+|               dt |             name | amount |
++------------------+------------------+--------+
+|         20240725 |            apple |      5 |
+|         20240725 |           banana |      7 |
++------------------+------------------+--------+
+*/
 ```
 
 {{< /tab >}}
--- a/docs/content/maintenance/manage-branches.md
+++ b/docs/content/maintenance/manage-branches.md
@@ -197,17 +197,13 @@ CALL sys.create_branch('default.T', 'test');
 -- set primary key and bucket number for the branch
 ALTER TABLE `T$branch_test` SET (
     'primary-key' = 'dt,name',
-    'bucket' = '2'
+    'bucket' = '2',
+    'changelog-producer' = 'lookup'
 );
 
 -- set fallback branch
 ALTER TABLE T SET (
     'scan.fallback-branch' = 'test'
-);
-
--- set changelog producer for the streaming branch, in case a streaming job would like to read from it in the future
-ALTER TABLE `T$branch_test` SET (
-    'changelog-producer' = 'lookup'
 );
 
 -- write records into the streaming branch
--- a/docs/content/maintenance/manage-branches.md
+++ b/docs/content/maintenance/manage-branches.md
@@ -159,3 +159,72 @@ Run the following command:
 {{< /tab >}}
 
 {{< /tabs >}}
+
+### Batch Reading from Fallback Branch
+
+You can set the table option `scan.fallback-branch`
+so that when a batch job reads from the current branch, if a partition does not exist,
+the reader will try to read this partition from the fallback branch.
+For streaming read jobs, this feature is currently not supported, and will only produce results from the current branch.
+
+What's the use case of this feature? Say you have created a Paimon table partitioned by date.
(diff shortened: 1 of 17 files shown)


docs/content/maintenance/manage-branches.md
```diff
@@ -159,3 +159,72 @@ Run the following command:
...
+) PARTITIONED BY (dt);
+
+-- create a branch for streaming job
+-- you can even specify primary keys and bucket number, even if the original branch has no primary key
+-- in this example, we create a new branch `test`, uses `dt` and `name` as primary keys, has a bucket number of 2, and will copy the table options from the original branch
+CALL sys.create_branch('default.T', 'test', 'dt, name', 2, true);
+
+-- set fallback branch
+ALTER TABLE T SET (
+    'scan.fallback-branch' = 'test'
```
JingGe, line 201: How to reset it? 'scan.fallback-branch' = null ?

paimon-core/src/main/java/org/apache/paimon/table/FileStoreTableFactory.java
```diff
@@ -89,6 +91,27 @@ public static FileStoreTable create(
...
+            branchOptions.set(CoreOptions.SCAN_FALLBACK_BRANCH, "");
+            FileStoreTable fallbackTable =
+                    FileStoreTableFactory.create(
+                            fileIO,
+                            tablePath,
+                            new SchemaManager(fileIO, tablePath, fallbackBranch).latest().get(),
+                            branchOptions,
+                            catalogEnvironment);
+
+            Preconditions.checkArgument(!(table instanceof FallbackReadFileStoreTable));
```
JingGe, line 110: Does it make sense to move the two Preconditions checks to the constructor of `FallbackReadFileStoreTable` ?

JingGe (commented):
Thanks for driving it. Just left some comments, PTAL

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/BranchSqlITCase.java
```diff
@@ -75,12 +76,90 @@ public void testAlterTable() throws Exception {
...
+                        "+I[apple, 10]", "+I[banana, 20]", "+I[tiger, 10]", "+I[wolf, 20]");
+        assertThat(collectResult(tEnv, "SELECT v, k FROM t WHERE pt = 1"))
+                .containsExactlyInAnyOrder("+I[apple, 10]", "+I[banana, 20]");
+        assertThat(collectResult(tEnv, "SELECT v, k FROM t WHERE pt = 2"))
+                .containsExactlyInAnyOrder("+I[tiger, 10]", "+I[wolf, 20]");
+
+        tEnv.executeSql("INSERT INTO `t$branch_pk` VALUES (2, 10, 'lion')").await();
+        assertThat(collectResult(tEnv, "SELECT v, k FROM t"))
+                .containsExactlyInAnyOrder(
+                        "+I[apple, 10]", "+I[banana, 20]", "+I[lion, 10]", "+I[wolf, 20]");
```
JingGe, line 108: Nit: additional test when `scan.fallback-branch` is not set, i.e. only return data from main branch

JingGe (commented):
Thanks for driving it. Just left some comments, PTAL

docs/content/maintenance/manage-branches.md
tsreaper: `ALTER TABLE T RESET ('scan.fallback-branch')`, see [Paimon docs](https://paimon.apache.org/docs/master/flink/sql-alter/#removing-table-properties).

docs/content/maintenance/manage-branches.md
```diff
@@ -159,3 +159,86 @@ Run the following command:
...
+CREATE TABLE T (
+    dt STRING NOT NULL,
+    name STRING NOT NULL,
+    amount BIGINT
+) PARTITIONED BY (dt);
+
+-- create a branch for streaming job
+-- you can even specify primary keys and bucket number, even if the original branch has no primary key
+-- in this example, we create a new branch `test`, uses `dt` and `name` as primary keys, has a bucket number of 2, and will copy the table options from the original branch
+CALL sys.create_branch('default.T', 'test', 'dt, name', 2, true);
```
JingsongLi, line 197: With the subsequent development, different branches may have significant differences and require a lot of options modifications. Can we consider directly supporting the modification of all options of an empty table?

docs/content/maintenance/manage-branches.md
```diff
@@ -159,3 +159,90 @@ Run the following command:
...
+    'bucket' = '2'
+);
+
+-- set fallback branch
+ALTER TABLE T SET (
+    'scan.fallback-branch' = 'test'
+);
+
+-- set changelog producer for the streaming branch, in case a streaming job would like to read from it in the future
+ALTER TABLE `T$branch_test` SET (
```
JingsongLi, line 209: put two alter together.

paimon-core/src/main/java/org/apache/paimon/table/FileStoreTableFactory.java
```diff
@@ -89,6 +90,24 @@ public static FileStoreTable create(
...
                                 fileIO, tablePath, tableSchema, catalogEnvironment);
-        return table.copy(dynamicOptions.toMap());
+        table = table.copy(dynamicOptions.toMap());
+
+        Options options = new Options(table.options());
+        String fallbackBranch = options.get(CoreOptions.SCAN_FALLBACK_BRANCH);
+        if (!StringUtils.isNullOrWhitespaceOnly(fallbackBranch)) {
+            Options branchOptions = new Options();
+            branchOptions.set(CoreOptions.BRANCH, fallbackBranch);
+            branchOptions.set(CoreOptions.SCAN_FALLBACK_BRANCH, "");
```
JingsongLi, line 100: You can introduce private method to `createNoWrappedTable`.

JingsongLi (approved):
+1

PR #3824, opened 2024-07-28
[core] Introduce DelegateCatalog

JingsongLi:
### Purpose

- Extract `DelegateCatalog` to more catalog implementations, for example, `CachedCatalog`.


Zouxxyy (approved):
+1

PR #3825, opened 2024-07-28
[spark] Support global options via SQL conf

Zouxxyy:
### Purpose

Users can configure paimon option globally via `spark.paimon.x`, e.g.

`set spark.paimon.file.block-size=512M` to control the file block size globally

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonOptionHelper.scala
+++ b/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonOptionHelper.scala
@@ -43,7 +43,7 @@ object PaimonOptionHelper extends SQLConfHelper {
     mergedOptions
   }
 
-  def withDynamicOptions[T <: Table](table: T, extraOptions: JMap[String, String] = null): T = {
+  def withDynamicOptions[T <: Table](table: T, extraOptions: JMap[String, String]): T = {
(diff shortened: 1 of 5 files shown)


JingsongLi (commented):
Why not just using `paimon.xxx`?

Zouxxyy:
> Why not just using `paimon.xxx`?

Spark will only recognize sql conf starting with `spark.`

Zouxxyy:
For example, if we want to configure hadoop conf to spark, we need to set like `spark.hadoop.x`

JingsongLi (commented):
+1 after test pass.

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonOptionHelper.scala
```diff
@@ -0,0 +1,54 @@
...
+
+import org.apache.paimon.table.Table
+
+import org.apache.spark.sql.catalyst.SQLConfHelper
+
+import java.util.{HashMap => JHashMap, Map => JMap}
+
+import scala.collection.JavaConverters._
+
+object PaimonOptionHelper extends SQLConfHelper {
```
YannByron, line 29: `Helper` typically represents a trait, not a object.

JingsongLi (approved):
Looks good to me!

YannByron (approved):

PR #3829, opened 2024-07-29
[core] Introduce CachingCatalog

JingsongLi:
### Purpose

Introduce a caching catalog, it is enabled by default, and its expiration interval is 30 s.

This can effectively reduce interaction with the metastore and lower the pressure on the metastore.

This is based on https://github.com/apache/paimon/pull/3824

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/catalog/CachingCatalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/catalog/CachingCatalog.java
@@ -1,0 +1,194 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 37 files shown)


paimon-core/src/main/java/org/apache/paimon/catalog/CachingCatalog.java
```diff
@@ -0,0 +1,194 @@
...
+            LOG.debug("Evicted {} from the table cache ({})", tableIdentifier, cause);
+            if (RemovalCause.EXPIRED.equals(cause)) {
+                if (!isSpecifiedSystemTable(tableIdentifier)) {
+                    tableCache.invalidateAll(allSystemTables(tableIdentifier));
+                }
+            }
+        }
+    }
+
+    private void invalidateTable(Identifier identifier) {
```
Zouxxyy, line 182: We can expose the method. For example, Spark `TableCatalog` has an interface which should uses it
```scala
  /**
   * Invalidate cached table metadata for an {@link Identifier identifier}.
   * <p>
   * If the table is already loaded or cached, drop cached data. If the table does not exist or is
   * not cached, do nothing. Calling this method should not query remote services.
   *
   * @param ident a table identifier
   */
  default void invalidateTable(Identifier ident) {
  }
```

JingsongLi: Good point, fixed.

Zouxxyy (approved):
Look good to me!  just wonder whether 30s is a bit short

PR #3830, opened 2024-07-29
[core] FileStore should not be serializable

JingsongLi:
### Purpose

For example, `SegmentsCache` in `AbstractFileStore` is not serializable.


yuzelin (approved):
+1

PR #3831, opened 2024-07-29
[core] Introduce SYSTEM_TABLE_LOADERS to SystemTableLoader

JingsongLi:
### Purpose

Introduce a map to `SystemTableLoader`, so outside can get all system tables from it.


yuzelin (approved):
+1

PR #3834, opened 2024-07-29
[core] Move commit callbacks from TableCommit into FileStoreCommit

tsreaper:
### Purpose

Currently `CommitCallback`s are called in `TableCommitImpl`. However as many places directly use `FileStoreCommitImpl` instead of `TableCommitImpl`, commit callbacks won't be called for these usages.

In this PR we move commit callbacks from `TableCommit` into `FileStoreCommit`, so in all scenarios commit callbacks can be called.

### Tests

This is a code refactor. Existing tests should cover this change.

### API and Format

No API / format changes.

### Documentation

No new feature.


JingsongLi (approved):
+1

PR #3836, opened 2024-07-29
[core] Make ObjectsCache thread safe

Zouxxyy:
### Purpose

Make ObjectsCache thread safe


paimon-core/src/main/java/org/apache/paimon/utils/ObjectsCache.java
```diff
@@ -36,11 +37,11 @@
 import java.util.function.BiFunction;
 
 /** Cache records to {@link SegmentsCache} by compacted serializer. */
+@ThreadSafe
 public class ObjectsCache<K, V> {
 
     private final SegmentsCache<K> cache;
     private final ObjectSerializer<V> serializer;
-    private final InternalRowSerializer rowSerializer;
```
JingsongLi, line 43: Introduce a threadlocal serializers?

JingsongLi:
+1

PR #3837, opened 2024-07-29
[flink] Support the watermark function of auditlog table

discivigour:
### Purpose

Support the watermark function of auditlog table

### Tests

- WatermarkITCase

### API and Format

No changes.


JingsongLi (approved):
+1

PR #3842, opened 2024-07-30
[spark] check column nullability when write

YannByron:
### Purpose

Linked issue: close #xxx

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/SparkConnectorOptions.java
+++ b/paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/SparkConnectorOptions.java
@@ -39,6 +39,13 @@ public class SparkConnectorOptions {
                     .withDescription(
                             "If true, allow to merge data types if the two types meet the rules for explicit casting.");
 
+    public static final ConfigOption<Boolean> IGNORE_NULLABLE_CHECK =
+            key("write.check.ignoreNullability")
+                    .booleanType()
+                    .defaultValue(false)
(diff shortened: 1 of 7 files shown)


paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/SparkConnectorOptions.java
```diff
@@ -39,6 +39,13 @@ public class SparkConnectorOptions {
                     .withDescription(
                             "If true, allow to merge data types if the two types meet the rules for explicit casting.");
 
+    public static final ConfigOption<Boolean> IGNORE_NULLABLE_CHECK =
+            key("write.check.ignoreNullability")
+                    .booleanType()
+                    .defaultValue(false)
```
JingsongLi, line 45: Why not just be true?

YannByron: I want to have this check by default to guarantee the nullability consistency.

And, Delta has `constraint` feature (eg: NOT NULL) which can check the incoming data when write. When paimon has this similar feature, we can remove this config and do not check nullability any more.

JingsongLi: Does this option affect compilation? I thinking about checking for null directly at runtime, but checking at compile time is too difficult to use.

JingsongLi (approved):
Looks good to me!

Issue #2619, opened 2024-01-02
[Feature] Support create db/table using location and external table


Zouxxyy:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/incubator-paimon/issues) and found nothing similar.

### Motivation

Support create db/table using location and external table

### Solution

`create db location 'xxx'` or `create table location 'xxx'` 

- For fileSystem catalog, when location is define, an exception is thrown
- For hive catalog, location sync to HMS

`create external table location 'xxx'`

- For fileSystem catalog, an exception is thrown
- For hive catalog, table_type (external) sync to HMS (spark, hive)

Other

- Drop external table, will not delete actural data
- Removed some confusing configuration

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #3843, opened 2024-07-30
[hive][spark] Support create table with specified location using hive catalog

zhongyujiang:
### Purpose

~~This relocates the 'path' property in the table options to 'location' for better presentation.~~

- Adapt 'path' to Spark's 'location' in table props ( the 'path' is still preserved in table props)
- Support the customization of the table location when creating a table(only allowed for creating external table using hive catalog)

Spark uses the reserved property `location` to indicate the location of the table, and in the output of `DESC EXTENDED`, the location information will be displayed under the "# Detailed Table Information" section for better visibility.

```text
+----------------------------+-----------------------------------------------------------------------------------------------------+-------+
|col_name                    |data_type                                                                                            |comment|
+----------------------------+-----------------------------------------------------------------------------------------------------+-------+
|a                           |bigint                                                                                               |NULL   |
|b                           |varchar(10)                                                                                          |NULL   |
|c                           |char(10)                                                                                             |NULL   |
|                            |                                                                                                     |       |
|# Metadata Columns          |                                                                                                     |       |
|__paimon_file_path          |string                                                                                               |       |
|__paimon_row_index          |bigint                                                                                               |       |
|                            |                                                                                                     |       |
|# Detailed Table Information|                                                                                                     |       |
|Name                        |default.testTableAs                                                                                  |       |
|Type                        |MANAGED                                                                                              |       |
|Location                    |file:/var/folders/2r/v_2n6mbj41v7q14m8f3j9q4w0000gn/T/junit3802231298897594813/default.db/testTableAs|       |
|Provider                    |paimon                                                                                               |       |
|Owner                       |zhongyujiang                                                                                         |       |
|Table Properties            |[file.format=parquet]                                                                                |       |
+----------------------------+-----------------------------------------------------------------------------------------------------+-------+
```

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/clone/PickFilesForCloneOperator.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/clone/PickFilesForCloneOperator.java
@@ -18,14 +18,20 @@
 
 package org.apache.paimon.flink.clone;
 
+import org.apache.paimon.CoreOptions;
 import org.apache.paimon.catalog.Catalog;
 import org.apache.paimon.catalog.Identifier;
 import org.apache.paimon.flink.FlinkCatalogFactory;
 import org.apache.paimon.fs.Path;
 import org.apache.paimon.options.Options;
 import org.apache.paimon.schema.Schema;
+import org.apache.paimon.schema.TableSchema;
 import org.apache.paimon.table.FileStoreTable;
 import org.apache.paimon.utils.Preconditions;
+
+import org.apache.paimon.shade.guava30.com.google.common.collect.ImmutableList;
+import org.apache.paimon.shade.guava30.com.google.common.collect.ImmutableMap;
+import org.apache.paimon.shade.guava30.com.google.common.collect.Iterables;
 
 import org.apache.flink.api.java.tuple.Tuple2;
 import org.apache.flink.streaming.api.operators.AbstractStreamOperator;
@@ -37,6 +43,7 @@ import org.slf4j.LoggerFactory;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 
 /**
  * Pick the files to be cloned of a table based on the input record. The record type it produce is
@@ -77,7 +84,7 @@ public class PickFilesForCloneOperator extends AbstractStreamOperator<CloneFileI
         FileStoreTable sourceTable = (FileStoreTable) sourceCatalog.getTable(sourceIdentifier);
         targetCatalog.createDatabase(targetIdentifier.getDatabaseName(), true);
         targetCatalog.createTable(
-                targetIdentifier, Schema.fromTableSchema(sourceTable.schema()), true);
+                targetIdentifier, newSchemaFromTableSchema(sourceTable.schema()), true);
 
         List<CloneFileInfo> result =
                 toCloneFileInfos(
@@ -93,6 +100,18 @@ public class PickFilesForCloneOperator extends AbstractStreamOperator<CloneFileI
         for (CloneFileInfo info : result) {
             output.collect(new StreamRecord<>(info));
         }
+    }
+
+    private static Schema newSchemaFromTableSchema(TableSchema tableSchema) {
+        return new Schema(
+                ImmutableList.copyOf(tableSchema.fields()),
+                ImmutableList.copyOf(tableSchema.partitionKeys()),
+                ImmutableList.copyOf(tableSchema.primaryKeys()),
+                ImmutableMap.copyOf(
+                        Iterables.filter(
+                                tableSchema.options().entrySet(),
+                                entry -> !Objects.equals(entry.getKey(), CoreOptions.PATH.key()))),
+                tableSchema.comment());
     }
 
     private List<CloneFileInfo> toCloneFileInfos(
--- a/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
+++ b/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
@@ -963,11 +963,12 @@ public class HiveCatalog extends AbstractCatalog {
                 }
             } catch (TableNotExistException e) {
                 // hive table does not exist.
-                Table finalNewTable =
-                        newTable == null
-                                ? createHiveTable(
-                                        identifier, tableSchema, location, usingExternalTable())
-                                : newTable;
+                if (newTable == null) {
+                    newTable =
+                            createHiveTable(
+                                    identifier, tableSchema, location, usingExternalTable());
+                }
+                Table finalNewTable = newTable;
                 clients.execute(client -> client.createTable(finalNewTable));
             }
 
--- a/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
+++ b/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
@@ -48,6 +48,7 @@ import org.apache.paimon.table.FormatTable;
 import org.apache.paimon.types.DataField;
 import org.apache.paimon.types.DataTypes;
 import org.apache.paimon.types.RowType;
+import org.apache.paimon.utils.Pair;
 import org.apache.paimon.utils.Preconditions;
 import org.apache.paimon.view.View;
 import org.apache.paimon.view.ViewImpl;
@@ -209,10 +210,18 @@ public class HiveCatalog extends AbstractCatalog {
         return getTableLocation(identifier, table);
     }
 
-    private Path initialTableLocation(Map<String, String> tableOptions, Identifier identifier) {
-        return tableOptions.containsKey(CoreOptions.PATH.key())
-                ? new Path(tableOptions.get(CoreOptions.PATH.key()))
-                : getTableLocation(identifier, null);
+    private Pair<Path, Boolean> initialTableLocation(
+            Map<String, String> tableOptions, Identifier identifier) {
+        boolean externalTable;
+        Path location;
+        if (tableOptions.containsKey(CoreOptions.PATH.key())) {
+            externalTable = true;
+            location = new Path(tableOptions.get(CoreOptions.PATH.key()));
+        } else {
+            externalTable = usingExternalTable();
+            location = getTableLocation(identifier, null);
+        }
+        return Pair.of(location, externalTable);
     }
 
     private Path getTableLocation(Identifier identifier, @Nullable Table table) {
@@ -639,9 +648,9 @@ public class HiveCatalog extends AbstractCatalog {
                         options,
                         schema.comment());
         try {
-            Map<String, String> tableOptions = schema.options();
-            boolean externalTable = createExternalTable(tableOptions);
-            Path location = initialTableLocation(tableOptions, identifier);
+            Pair<Path, Boolean> pair = initialTableLocation(schema.options(), identifier);
+            Path location = pair.getLeft();
+            boolean externalTable = pair.getRight();
             Table hiveTable = createHiveFormatTable(identifier, newSchema, location, externalTable);
             clients.execute(client -> client.createTable(hiveTable));
         } catch (Exception e) {
@@ -650,10 +659,7 @@ public class HiveCatalog extends AbstractCatalog {
         }
     }
 
-    private boolean createExternalTable(Map<String, String> tableOptions) {
-        if (tableOptions.containsKey(CoreOptions.PATH.key())) {
-            return true;
-        }
+    private boolean usingExternalTable() {
         CatalogTableType tableType =
                 OptionsUtils.convertToEnum(
                         hiveConf.get(TABLE_TYPE.key(), CatalogTableType.MANAGED.toString()),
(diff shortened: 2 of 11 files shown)


zhongyujiang:
cc @Zouxxyy @YannByron Can you help review this? Thanks!

JingsongLi (changes requested):
-1, It is best for Spark to be unified with other engines, as using a new option may harm the interoperability between engines

zhongyujiang:
Hi @JingsongLi This is not introducing a new option, but rather better adapting to the Spark engine. Because Spark has always used it to represent [the location of the table](https://github.com/apache/spark/blob/08943104fad9f5ea274e022d75135aedc55a566f/sql/catalyst/src/main/java/org/apache/spark/sql/connector/catalog/TableCatalog.java#L51).

> the interoperability between engines

Are you suggesting that you want users to retrieve the table location information from the `path` option within the options when using the `DESC` command in different engines?

However, I would like to point out that the meaning of the DESC command inherently varies across different engines, closely related to the functionality of the engine. For instance, the DESC command in Trino and Flink does not even display table options information, but only column information. Yet, Flink can display primary key attribute of columns because primary keys are part of the Flink specification (this is not the case in Spark).

Therefore, I believe we should also adapt this to Spark, which would be more in line with the usage habits of Spark users.

BTW, for Iceberg and Hive tables, Spark's DESC EXTENDED DDL command displays the location information through the location field. However, for Paimon tables, the location information is hidden within the options under the path field, making it less convenient to users. This is the motivation behind this PR.

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/SparkTable.scala
```diff
@@ -60,6 +60,8 @@ case class SparkTable(table: Table)
           properties.put(CoreOptions.PRIMARY_KEY.key, String.join(",", table.primaryKeys))
         }
         properties.put(TableCatalog.PROP_PROVIDER, SparkSource.NAME)
+        val location = properties.remove(CoreOptions.PATH.key())
```
Zouxxyy, line 63: This interface (`default Map<String, String> properties()`) is only used in show create or desc table, add the `TableCatalog.PROP_LOCATION` is ok, but wondering whether we need to remove `path`.

Besides, another point worth noting is that paimon currently does not support `create table x location('')` actually?  The new result of `show create table` may cause misunderstanding,  or we need support it first.

JingsongLi: Hi @zhongyujiang , we can discuss in this thread, @Zouxxyy raised two very good points:
1. We could keep `path` option.
2. We should also support execution for `show create table`.

zhongyujiang: >  We could keep path option.

Sure

> We should also support execution for show create table

I've tested this and found that the location info passed by Spark DDL will be ignored right now

IIUC, the table location is closely related to the warehouse address, and of course, different catalogs have different implementations:

- AbstractCatalog calculates the table location based on the warehouse, database, and table name when loading a table, and both FileSystemCatalog and JdbcCatalog follow this behavior
- HiveCatalog retrieves the table location information from the HMS (Hive Metastore)

Therefore, if we want to support this, we need to differentiate the catalog being used. This is feasible with HiveCatalog, but not with other catalogs. What do you think? @Zouxxyy @JingsongLi

JingsongLi: For other catalogs, we can check the location, if not equal, throw exception.

Zouxxyy: can see #https://github.com/apache/paimon/issues/2619

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
```diff
@@ -470,7 +471,18 @@ protected void createTableImpl(Identifier identifier, Schema schema) {
         // if changes on Hive fails there is no harm to perform the same changes to files again
         TableSchema tableSchema;
         try {
-            tableSchema = schemaManager(identifier).createTable(schema, usingExternalTable());
+            Path tableRoot;
+            if (schema.options().containsKey(CoreOptions.PATH.key())) {
+                checkArgument(
+                        Objects.equals(createTableType(), TableType.EXTERNAL),
+                        "The HiveCatalog only supports specifying location when creating an external table");
+                tableRoot = new Path(schema.options().get(CoreOptions.PATH.key()));
```
zhongyujiang, line 479: Location specification is only allowed when creating an external table.

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
```diff
@@ -767,7 +781,11 @@ private void updateHmsTable(Table table, Identifier identifier, TableSchema sche
         }
 
         // update location
-        locationHelper.specifyTableLocation(table, getDataTableLocation(identifier).toString());
+        String location =
+                schema.options().containsKey(CoreOptions.PATH.key())
+                        ? schema.options().get(CoreOptions.PATH.key())
+                        : getDataTableLocation(identifier).toString();
+        locationHelper.specifyTableLocation(table, location);
```
zhongyujiang, line 788: If provided, use the table path provided by the user.

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/SparkTable.scala
zhongyujiang: I have updated this to allow customizing the table location when using the Hive catalog and creating an external table. Please take a look when you have time. Thanks!

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/clone/PickFilesForCloneOperator.java
```diff
@@ -95,6 +102,18 @@ public void processElement(StreamRecord<Tuple2<String, String>> streamRecord) th
...
+    private static Schema newSchemaFromTableSchema(TableSchema tableSchema) {
+        return new Schema(
+                ImmutableList.copyOf(tableSchema.fields()),
+                ImmutableList.copyOf(tableSchema.partitionKeys()),
+                ImmutableList.copyOf(tableSchema.primaryKeys()),
+                ImmutableMap.copyOf(
+                        Iterables.filter(
+                                tableSchema.options().entrySet(),
+                                entry -> !Objects.equals(entry.getKey(), CoreOptions.PATH.key()))),
+                tableSchema.comment());
```
zhongyujiang, line 114: 目前 clone 表创建新的 schema 时，会把原表中所有的 table options 都直接传入新的 schema 中，包括 source 表的 table path，然后在创建表时忽略它，这看起来不正确，并且在 FileSystemCatalog 中禁止传入自定义的 table path 后， clone table 的单侧会失败，所以我这里做了修复。

并且原来的 fromTableSchema 方法构建新 schema 时直接使用了 source schema 中的 partitionKeys 和 primaryKeys  实例，这并不安全，所以我这里做了拷贝。

同时我把这个方法从 Schema 类中移到了这里，这是唯一需要使用它的地方，如果你们觉得它足够通用，我可以把它放回去。

Currently, when creating a new schema for the clone table, all table options from the original table are directly passed into the new schema, including the table path of the source table. This is then ignored during table creation, which seems incorrect. Additionally, after prohibiting the passing of custom table paths in the FileSystemCatalog, the clone table's unit tests would fail. Therefore, I have made a fix here.

Moreover, the original `fromTableSchema` method, when constructing the new schema, directly used the `partitionKeys` and `primaryKeys` instances from the source schema, which is not safe. So, I have made a copy here.

I have also moved this method from the `Schema` class to here, as this is the only place where it is needed. If you feel it is sufficiently general, I can put it back.

zhongyujiang, line 114: Currently, when creating a new schema for the clone table, all table options from the original table are directly passed into the new schema, including the table path of the source table. This is then ignored during table creation, which seems incorrect. Additionally, after prohibiting the passing of custom table paths in the FileSystemCatalog, the clone table's unit tests would fail. Therefore, I have made a fix here.

Moreover, the original fromTableSchema method, when constructing the new schema, directly used the partitionKeys and primaryKeys instances from the source schema, which is not safe. So, I have made a copy here.

I have also moved this method from the Schema class to here, as this is the only place where it is needed. If you feel it is sufficiently general, I can put it back.

JingsongLi: Sounds good to me.

paimon-core/src/main/java/org/apache/paimon/catalog/FileSystemCatalog.java
```diff
@@ -117,6 +119,9 @@ protected void dropTableImpl(Identifier identifier) {
 
     @Override
     public void createTableImpl(Identifier identifier, Schema schema) {
+        checkArgument(
+                !schema.options().containsKey(CoreOptions.PATH.key()),
```
JingsongLi, line 123: Can we just relax this check? For example, If it is the same path, we should allow it.

I am quite concerned that SHOW CREATE TABLE may not run properly.

JingsongLi:
Thanks @zhongyujiang for update. Left some comments.

paimon-core/src/main/java/org/apache/paimon/catalog/FileSystemCatalog.java
zhongyujiang: Thanks for reviewing. I would like to share my thoughts on the prohibition of setting table location in `FileSystemCatalog`. 

I think the `SHOW CREATE TABLE` command is usually used to create a test table with the same schema as the source table. Users typically use `SHOW CREATE TABLE` to get the DDL of a table and then create a test table with a different name for testing purposes. In this scenario, the location information in the DDL will inevitably mismatch the path assigned by the `FileSystemCatalog` for the new table, requiring users to modify the DDL in order to successfully create the table.

So even if the restrictions were relaxed, I believe it would still be somewhat confusing for users, as success would only be guaranteed when the specified location matches the one assigned by the catalog. If that's the case, why bother passing the location at all?

Therefore, instead of relaxing the checks here, I suggest we optimize the documentation by clearly stating this restriction in the documentation. We can declare the location management restrictions of `FileSystemCatalog` and the recommended usage of Spark DDL. What do you think?

zhongyujiang: This is also our current approach to managing Iceberg tables in our platform. We host the location for all Iceberg tables on behalf of the users and strongly advise against specifying a location when creating tables. So far, this has been working well.

JingsongLi: I still think a relax check will be better. "create a test table with a different name for testing purposes", we can have very clear exception for this case.

Zouxxyy (commented):
I see the new commits provide the support for external tables, thanks. Can you add the following test:
For hive catalog (can add in DDLWithHiveCatalogTestBase)
1. Create an internal table, drop it, and check if the files are deleted
2. Create an external table using location, drop it, and check if the files are deleted

For filesystem catalog (can add to DDLTestBase)
All tables should be internal tables

Zouxxyy:
Update:
1. Support `CREATE TABLE x LOCATION 'xxx'` when using hive catalog  (for other catalog, like file system catalog, an exception will be throw). The table created this way will be treated as external table.
2. Drop a managed table,  the data files will be deleted.
3. Drop a external table,  the data files will not be deleted.

paimon-core/src/main/java/org/apache/paimon/catalog/FileSystemCatalog.java
Zouxxyy: > Therefore, instead of relaxing the checks here, I suggest we optimize the documentation by clearly stating this restriction in the documentation. We can declare the location management restrictions of `FileSystemCatalog` and the recommended usage of Spark DDL. What do you think?

I'm +1 for this

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/SparkCatalog.java
```diff
@@ -411,6 +411,12 @@ private Schema toInitialSchema(
         }
         normalizedProperties.remove(PRIMARY_KEY_IDENTIFIER);
         normalizedProperties.remove(TableCatalog.PROP_COMMENT);
+        if (normalizedProperties.containsKey(TableCatalog.PROP_LOCATION)) {
+            String path = normalizedProperties.remove(TableCatalog.PROP_LOCATION);
+            normalizedProperties.put(CoreOptions.PATH.key(), path);
+            // For v2 table, as long as it has specified the location, treat it as external
+            normalizedProperties.put(Catalog.EXTERNAL_PROP, "true");
```
JingsongLi, line 418: We don't need this option, hive catalog can know if it is a external table.

JingsongLi (approved):
+1

zhongyujiang:
@Zouxxyy @JingsongLi Thanks for updating this, I'm sorry that I am not following up on this PR in time.

PR #3845, opened 2024-07-30
[core] Immutable table options can now be changed on an empty table

tsreaper:
### Purpose

Currently we have some immutable table options (for example, merge-engine and primary-key). Once the table is created, they cannot be changed.

However for an empty table, changing these options does not affect any data. Also ,we've supported different schemas for different branches in #3757. Users can set primary keys to a newly created branch with this feature.

### Tests

* Unit tests.

### API and Format

No API / format changes.

### Documentation

No new feature.


JingsongLi (approved):
+1

PR #3853, opened 2024-07-30
[core] Manage the lookup file disk cache in task granularity

### LINKED ISSUE #3848: [Feature] Manage the lookup file disk cache in task granularity
### Search before asking

### Motivation

The LookupLevels create a cache for each bucket in the partition. Therefore, if a task has multiple buckets/partitions to write to, the total disk usage can become almost unlimited. This could lead to exceeding disk quotas errors in Kubernetes environments. As a solution, I propose managing the disk cache at the task level, similar to how CacheManager operates.

### Solution

_No response_

### Anything else?

_No response_

Aitozi:
### Purpose

Linked issue: close #3848

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/mergetree/LookupLevels.java
+++ b/paimon-core/src/main/java/org/apache/paimon/mergetree/LookupLevels.java
@@ -120,10 +120,14 @@ public class LookupLevels<T> implements Levels.DropFileCallback, Closeable {
         return lookupFiles;
     }
 
+    @VisibleForTesting
+    Set<String> cachedFiles() {
+        return cachedFiles;
+    }
+
     @Override
     public void notifyDropFile(String file) {
         lookupFiles.invalidate(file);
-        cachedFiles.remove(file);
     }
 
     @Nullable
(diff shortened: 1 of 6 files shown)


paimon-core/src/main/java/org/apache/paimon/mergetree/LookupLevels.java
```diff
@@ -207,12 +211,8 @@ private LookupFile createLookupFile(DataFileMeta file) throws IOException {
         return new LookupFile(localFile, file, lookupStoreFactory.createReader(localFile, context));
     }
 
-    @Override
-    public void close() throws IOException {
```
JingsongLi, line 211: Should we clean all files when closing?

JingsongLi: But by default, there is no max size configuration.

Aitozi: Evict by the retention time?

Aitozi: Before the LookupLevels's cache file will be cleared when the writer is idled (no data written to in the latest snapshot). 

Now, the cache file will still be there after the idle writer closed and evict by the retention time by default.

JingsongLi: Yes, it is better to remove all files when closing this writer.

JingsongLi: You can manage a collection in `LookupLevels`.

Aitozi: > You can manage a collection in `LookupLevels`.

Thanks, Make sense to me.

JingsongLi (approved):
+1

PR #3854, opened 2024-07-30
[core] Support TimestampToNumericPrimitiveCastRule in casting

xuzifu666:
### Purpose
Follow up https://github.com/apache/paimon/pull/3832 had support NumericPrimitive  to Timestamp  casting，this pr support Timestamp to NumericPrimitive  casting.

Linked issue: close #xxx

### Tests

CastExecutorTest##testTimestampToNumeric

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/casting/TimestampToNumericPrimitiveCastRule.java
+++ b/paimon-core/src/main/java/org/apache/paimon/casting/TimestampToNumericPrimitiveCastRule.java
@@ -44,12 +44,15 @@ public class TimestampToNumericPrimitiveCastRule extends AbstractCastRule<Timest
 
     @Override
     public CastExecutor<Timestamp, Number> create(DataType inputType, DataType targetType) {
-        if (inputType.is(DataTypeRoot.TIMESTAMP_WITHOUT_TIME_ZONE)) {
-            return value -> DateTimeUtils.unixTimestamp(value.getMillisecond());
-        } else if (inputType.is(DataTypeRoot.TIMESTAMP_WITH_LOCAL_TIME_ZONE)) {
-            return value ->
-                    DateTimeUtils.unixTimestamp(
-                            Timestamp.fromLocalDateTime(value.toLocalDateTime()).getMillisecond());
+        if (targetType.is(DataTypeRoot.BIGINT)) {
+            if (inputType.is(DataTypeRoot.TIMESTAMP_WITHOUT_TIME_ZONE)) {
+                return value -> DateTimeUtils.unixTimestamp(value.getMillisecond());
+            } else if (inputType.is(DataTypeRoot.TIMESTAMP_WITH_LOCAL_TIME_ZONE)) {
+                return value ->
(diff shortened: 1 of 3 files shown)


paimon-core/src/main/java/org/apache/paimon/casting/TimestampToNumericPrimitiveCastRule.java
```diff
@@ -0,0 +1,56 @@
...
+                        .input(DataTypeRoot.TIMESTAMP_WITHOUT_TIME_ZONE)
+                        .input(DataTypeRoot.TIMESTAMP_WITH_LOCAL_TIME_ZONE)
+                        .target(DataTypeFamily.NUMERIC)
+                        .build());
+    }
+
+    @Override
+    public CastExecutor<Timestamp, Number> create(DataType inputType, DataType targetType) {
+        if (inputType.is(DataTypeRoot.TIMESTAMP_WITHOUT_TIME_ZONE)) {
+            return value -> DateTimeUtils.unixTimestamp(value.getMillisecond());
```
JingsongLi, line 48: Please add a case cast to and cast from.

xuzifu666: @JingsongLi Thanks， had added case for target type.

xuzifu666: OK，had added cast to and from in test case @JingsongLi

paimon-core/src/main/java/org/apache/paimon/casting/TimestampToNumericPrimitiveCastRule.java
```diff
@@ -0,0 +1,56 @@
...
+        } else if (inputType.is(DataTypeRoot.TIMESTAMP_WITH_LOCAL_TIME_ZONE)) {
+            return value ->
+                    DateTimeUtils.unixTimestamp(
```
JingsongLi, line 51: For int, short, byte types, should not return long.

xuzifu666: Make sense，due to the number is hundreds million at least that Interge.Max also cannot hold the size，so I just handle Bigint condition and others return null. @JingsongLi

JingsongLi: integer is OK?

JingsongLi: And `CastRulePredicate` should be only target to int and bigint

xuzifu666: Yes，should support int and bigint，I would change it latter. Thanks for the reminder. @JingsongLi

JingsongLi (approved):
+1

PR #3856, opened 2024-07-30
[core] Introduce cache.manifest-max-memory to CachingCatalog

JingsongLi:
### Purpose

Next part for https://github.com/apache/paimon/pull/3829

Avoid costly manifest reading for table in CachingCatalog. Reusing `writeManifestCache` mechanism in `AbstractFileStore`.


Zouxxyy (approved):
+1

PR #3857, opened 2024-07-31
[core] Introduce data structure for timestamp with local zone

JingsongLi:
### Purpose

Due to the fact that both types share the same Timestamp, and the default toString for Timestamp is implemented for TimestampWithoutZone, the usage is very disjointed and difficult for users to understand.

We can specifically introduce a separate data structure for TimestampLocalZone, and this PR introduces this data structure without modifying its usage in the code.


yuzelin (approved):
+1

PR #3862, opened 2024-07-31
[core] Identifier can now recognize branch and system tables

tsreaper:
### Purpose

In # we allow visit table branches by `$branch_`. Now an identifier may represent a data table, or a system table, or a table branch.

In this PR, we add some methods in `Identifier` so that it can recognize branches and system tables. Catalogs will need these methods to get schemas, communicate with metastores, etc.

### Tests

This is a refactor. Existing tests should cover the changes.

### API and Format

`Identifier` is a public API. This PR adds some methods to the class.

### Documentation

No new feature.


JingsongLi (approved):
+1

PR #3864, opened 2024-07-31
[core] Optimize manifest-full-compact and parts overwrite by PartitionPredicate

JingsongLi:
### Purpose

The case is the upstream data changes too many partitions, make the Predicate too deep, then error is:
```
Caused by: java.lang.RuntimeException: java.lang.StackOverflowError
	at org.apache.paimon.manifest.ManifestFileMeta.merge(ManifestFileMeta.java:178)
	at org.apache.paimon.operation.FileStoreCommitImpl.tryCommitOnce(FileStoreCommitImpl.java:808)
	... 27 more
Caused by: java.lang.StackOverflowError
	at org.apache.paimon.predicate.CompoundPredicate.test(CompoundPredicate.java:59)
	at org.apache.paimon.predicate.Or.test(Or.java:55)
	at org.apache.paimon.predicate.CompoundPredicate.test(CompoundPredicate.java:59)
	...
```


yuzelin (approved):
+1

PR #3865, opened 2024-07-31
[core] fix bug of watermark override with append only table

wwj6591812:
### Purpose

1、When write date to append only table, the watermark will be override by the CompactionCoordinator with Long.MIN_VALUE. This is a bug, now fix it.
2、Format some code.

Linked issue: close #xxx


paimon-common/src/main/java/org/apache/paimon/predicate/LeafPredicate.java
```diff
@@ -69,7 +69,7 @@ public DataType type() {
         return type;
     }
 
-    public int index() {
+    public int fieldIndex() {
```
JingsongLi, line 72: do not modify this, compute engines rely on this api.

wwj6591812: done

wwj6591812:
see https://github.com/apache/paimon/pull/3872

PR #3867, opened 2024-08-01
[hive] Partition is dropped from metastore only when all branches do not contain this partition

tsreaper:
### Purpose

Currently, when a partition is dropped from any branch, it will also be deleted from the metastore. This is inconvenient for the users because this partition may exist in other branches, including the main branch.

This PR fixes this issue. Paritition is now dropped from metastore only when all branches do not contain this partition.

### Tests

* Unit tests.

### API and Format

No changes in format.

### Documentation

No new feature.


paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
```diff
@@ -304,15 +323,38 @@ && new CoreOptions(tableSchema.options()).partitionedTableInMetastore()) {
...
+    private boolean partitionExistsInOtherBranches(
+            Identifier identifier, Map<String, String> partitionSpec)
+            throws TableNotExistException {
+        FileStoreTable mainTable =
+                (FileStoreTable)
+                        getTable(
+                                new Identifier(
+                                        identifier.getDatabaseName(), identifier.getTableName()));
+        List<String> branchNames =
+                mainTable.branchManager().branches().stream()
```
JingsongLi, line 335: It seems that we can optimize `branches` method, it is really costly.

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
```diff
@@ -304,15 +323,38 @@ && new CoreOptions(tableSchema.options()).partitionedTableInMetastore()) {
...
+                    (FileStoreTable)
+                            getTable(
+                                    new Identifier(
+                                            identifier.getDatabaseName(),
+                                            identifier.getTableName(),
+                                            branchName,
+                                            null));
+            if (!table.newScan().withPartitionFilter(partitionSpec).plan().splits().isEmpty()) {
+                return true;
+            }
```
LinMingQiang, line 355: Can we add a method to BranchManager to get the partition information?

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
```diff
@@ -304,15 +322,35 @@ && new CoreOptions(tableSchema.options()).partitionedTableInMetastore()) {
...
+
+            FileStoreTable table =
+                    (FileStoreTable)
+                            getTable(
+                                    new Identifier(
+                                            identifier.getDatabaseName(),
+                                            identifier.getTableName(),
+                                            branchName,
+                                            null));
+            if (!table.newScan().withPartitionFilter(partitionSpec).plan().splits().isEmpty()) {
```
JingsongLi, line 349: Use `TableScan.listPartitions()`.

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
```diff
@@ -304,15 +322,35 @@ && new CoreOptions(tableSchema.options()).partitionedTableInMetastore()) {
...
+        branchNames.add(DEFAULT_MAIN_BRANCH);
+
+        for (String branchName : branchNames) {
+            if (branchName.equals(identifier.getBranchNameOrDefault())) {
+                continue;
+            }
+
+            FileStoreTable table =
+                    (FileStoreTable)
+                            getTable(
```
JingsongLi, line 343: Construct branch table from original table to improve performance.

JingsongLi (approved):
+1

PR #3872, opened 2024-08-02
[core] fix bug of watermark override with append only table

wwj6591812:
### Purpose

When write data to append only table, the watermark will be override by the CompactionCoordinator with Long.MIN_VALUE. This is a bug, now fix it.

Linked issue: close #xxx


wwj6591812:
Now the fail case is not caused by my modifications.
@JingsongLi 
please cc, thx

JingsongLi (approved):
+1

PR #3875, opened 2024-08-02
[core] Introduce AsyncPositionOutputStream

JingsongLi:
### Purpose

This PR introduces asynchronous file writing to increase data throughput.

In order to avoid the problem of byte array reuse, all written arrays are copied once, which is the overhead brought by this PR. However, compared to writing files, copying a byte array seems to be less expensive.

### Tests

`AsyncPositionOutputStreamTest`

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/fs/AsyncPositionOutputStream.java
+++ b/paimon-common/src/main/java/org/apache/paimon/fs/AsyncPositionOutputStream.java
@@ -42,14 +42,12 @@ public class AsyncPositionOutputStream extends PositionOutputStream {
     private final AtomicReference<Throwable> exception;
     private final Future<?> future;
 
-    private volatile boolean isClosed;
     private long position;
 
     public AsyncPositionOutputStream(PositionOutputStream out) {
         this.out = out;
         this.eventQueue = new LinkedBlockingQueue<>();
         this.exception = new AtomicReference<>();
-        this.isClosed = false;
         this.position = 0;
         this.future = EXECUTOR_SERVICE.submit(this::execute);
     }
@@ -65,7 +63,7 @@ public class AsyncPositionOutputStream extends PositionOutputStream {
 
     private void doWork() throws InterruptedException, IOException {
         try {
-            while (!isClosed) {
+            while (true) {
                 AsyncEvent event = eventQueue.poll(AWAIT_TIMEOUT_SECONDS, TimeUnit.SECONDS);
                 if (event == null) {
                     continue;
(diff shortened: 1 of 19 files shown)


paimon-common/src/main/java/org/apache/paimon/fs/AsyncPositionOutputStream.java
```diff
@@ -0,0 +1,189 @@
...
+        putEvent(new EndEvent());
+        try {
+            this.future.get();
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new RuntimeException(e);
+        } catch (ExecutionException e) {
+            throw new RuntimeException(e);
+        } finally {
+            isClosed = true;
```
tsreaper, line 148: What's the usage of this `isClosed`? The thread needs an `EndEvent` or `isClosed == true` to exit. However you use `future.get()` to wait for thread and then set `isClosed = true`, meaning that only `EndEvent` can stop the thread.

paimon-common/src/main/java/org/apache/paimon/utils/ReuseByteArrayOutputStream.java
```diff
@@ -0,0 +1,37 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.paimon.utils;
+
+import java.io.ByteArrayOutputStream;
+
+/** A {@link ByteArrayOutputStream} which can reuse byte array. */
+public class ReuseByteArrayOutputStream extends ByteArrayOutputStream {
```
tsreaper, line 24: What's the usage of this class? I don't see it is used anywhere.

paimon-common/src/main/java/org/apache/paimon/fs/AsyncPositionOutputStream.java
```diff
@@ -0,0 +1,229 @@
...
+                Thread.currentThread().interrupt();
+                throw new RuntimeException(e);
+            }
+        }
+    }
+
+    @Override
+    public void close() throws IOException {
+        checkException();
+        flushBuffer();
```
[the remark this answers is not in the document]
JingsongLi: It is flushBuffer, close will also do inner flush too.

paimon-common/src/main/java/org/apache/paimon/utils/ReuseByteArrayOutputStream.java
JingsongLi: Just for util usage.

tsreaper (approved):

PR #3876, opened 2024-08-02
[spark] Fix partition transform converter

Zouxxyy:
### Purpose

For DDL like
```
 CREATE TABLE T (id STRING, name STRING, pt STRING)
 PARTITIONED BY (substr(pt, 1, 2))
```
should throw an exception since we do not support hidden partitions


JingsongLi (approved):
+1

PR #3883, opened 2024-08-02
[parquet] Child vector of complex type should arrange elements compactly (like orc)

yuzelin:
### Purpose

For orc, if we have 4 rows of `ARRAY<ARRAY<STRING>: [[0, null], [1], [], null], [2], [], null]`. The data layout is:

image

Currently, the parquet format layout is not same. For example, the inner `IntVector` will have more than one null value (that is ,the empty and null array will be set as null value in inner vector). This PR make parquet to be consistent with orc.


Stephen0421:
+1

JingsongLi:
+1

PR #3888, opened 2024-08-04
[spark] Support show columns in spark

xuzifu666:
### Purpose
Support show columns syntax in paimon，currently spark not support v2 in show columns(fix in spark https://github.com/apache/spark/pull/47568 ) but paimon side not support yet，would error out like：

spark-sql> show COLUMNS from bdsp_test.paimon_mi_31;
Error in query: SHOW COLUMNS is not supported for v2 tables.

Add spark analyzer extensions to capture it，after fix:

Linked issue: https://github.com/apache/paimon/issues/3889


paimon-spark/paimon-spark-common/src/test/scala/org/apache/paimon/spark/sql/PaimonShowColumnsTestBase.scala
```diff
@@ -0,0 +1,47 @@
...
+
+    spark.sql(
+      s"""
+         |INSERT INTO T SELECT '1', 'aa', 12, 22
+         |""".stripMargin
+    )
+
+    assertDoesNotThrow(
+      () => {
+        spark.sql("SHOW COLUMNS FROM T")
```
JingsongLi, line 42: Why not check its result?

xuzifu666: UT had added for check its result and it like picture in discription. @JingsongLi

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/commands/PaimonShowColumnsCommand.scala
```diff
@@ -0,0 +1,37 @@
...
+
+import org.apache.paimon.spark.SparkTable
+import org.apache.paimon.spark.leafnode.{PaimonLeafCommand, PaimonLeafRunnableCommand}
+import org.apache.paimon.table.FileStoreTable
+
+import org.apache.spark.sql.{Row, SparkSession}
+import org.apache.spark.sql.catalyst.expressions.{Attribute, AttributeReference}
+import org.apache.spark.sql.types.{BinaryType, Metadata, StringType}
+
+case class PaimonShowColumnsCommand(v2Table: SparkTable)
```
YannByron, line 29: this can extend `PaimonLeafRunnableCommand` directly, then we don't need `PaimonShowColumnsExec`.

paimon-spark/paimon-spark-common/src/test/scala/org/apache/paimon/spark/sql/PaimonShowColumnsTestBase.scala
```diff
@@ -0,0 +1,52 @@
...
+                 |TBLPROPERTIES ('primary-key'='id')
+                 |""".stripMargin)
+
+    spark.sql(
+      s"""
+         |INSERT INTO T SELECT '1', 'aa', 12, 22
+         |""".stripMargin
+    )
+
+    assertDoesNotThrow(
```
YannByron, line 41: if check its result below, we don't need this assert.

xuzifu666: OK，had check result，I remove the assert @YannByron

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/commands/PaimonShowColumnsCommand.scala
xuzifu666: Thanks，had change it as your suggestion. @YannByron

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/commands/PaimonShowColumnsCommand.scala
```diff
@@ -0,0 +1,41 @@
...
+
+import org.apache.spark.sql.{Row, SparkSession}
+import org.apache.spark.sql.catalyst.expressions.{Attribute, AttributeReference, GenericRow}
+import org.apache.spark.sql.types.{Metadata, StringType}
+import org.apache.spark.unsafe.types.UTF8String
+
+case class PaimonShowColumnsCommand(v2Table: SparkTable)
+  extends PaimonLeafRunnableCommand
+  with WithFileStoreTable {
+  override def table: FileStoreTable = v2Table.getTable.asInstanceOf[FileStoreTable]
```
YannByron, line 33: maybe this command can also apply to the system table, not just `FileStoreTable`. if yes, modify here and add a UT for that.

xuzifu666: Thanks for comments，but I had a try but seems not find a way to compatible with system table and v2table due to their not use same interface. Maybe we can just support FileStoreTable firstly？

YannByron (approved):

PR #3890, opened 2024-08-04
[core] Add writing different data for keys in lookup file store benchmark

### LINKED ISSUE #3861: [Feature] Add writing different data for keys in lookup file store benchmark
### Search before asking

### Motivation

Subtask of #3827 
Currently hash lookup file store will not write value data if it is the same as previous key, which cause the performance gap between hash and sort lookup file store. We need to add benchmark that writes different values between the latest two keys.

### Solution

_No response_

### Anything else?

_No response_

FangYongs:
### Purpose

Linked issue: close #3861 

### Tests

no

### API and Format

no

### Documentation

no


JingsongLi:
+1

PR #3893, opened 2024-08-05
[avro] Do not flush on every block for DataFileWriter

JingsongLi:
### Purpose

Improve performance:
No need to flush every block for avro writer.


leaves12138 (approved):
+1

PR #3896, opened 2024-08-05
[core] Change expire parition log

bknbkn:
### Purpose
Make expire partition log more readable

LOG changed before: ```Expire Partition: org.apache.paimon.data.BinaryRow@b2d8cadd``` 
LOG changed after: ```Expire Partition: {dt=20240710, hour=4}```


JingsongLi (approved):
+1

PR #3897, opened 2024-08-05
[core] Increase write-max-writers-to-spill and logging

JingsongLi:
### Purpose

`write-max-writers-to-spill` default value is 5, which is too small. Generally, at least 10 are acceptable.


leaves12138 (approved):
+1

PR #3898, opened 2024-08-05
[core] add interface in Catalog for CloneAction adapter CachingCatalog

wwj6591812:
Now CloneActionITCase is unstabitily, because we use Catalog#getTable which means the schema file must be copied completely, but in a distributed environment, this cannot be guaranteed.
So we add interface in Catalog for CloneAction adapter CachingCatalog.

### Purpose

Linked issue: close #xxx


JingsongLi:
+1 Thanks @wwj6591812

PR #3899, opened 2024-08-05
[spark] Support specifying time-pattern in ExpairePartition procedure

bknbkn:
### Purpose

Support time-pattern options in spark expire partition procedure


JingsongLi (approved):
+1

PR #3903, opened 2024-08-06
[core] Adjust compaction.max.file-num of bucketed append table to 5

JingsongLi:
### Purpose

Currently, the default value is 50, which means 50 small files in a bucket, it will be large number of small files...

Adjust this to 5.


yuzelin (approved):
+1

Issue #3904, opened 2024-08-06
[Bug] when use `from-timestamp` read data from snapshot, if  time < earliestSnapshot time , no data return


chunji96:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Paimon version

commit id : f2bf3523975a206c069ace411f247f275a161985

### Compute Engine

spark,trino...

### Minimal reproduce step

use UT method `org.apache.paimon.spark.SparkTimeTravelITCase#testTravelToTimestampString`,and when `anchor` < snapshot1 timestamp , UT will fail。the test case following:
```
    @Test
    public void testTravelToTimestampString() throws Exception {
        spark.sql("CREATE TABLE t (k INT, v STRING)");

        // snapshot 1
        writeTable(
                "t",
                GenericRow.of(1, BinaryString.fromString("Hello")),
                GenericRow.of(2, BinaryString.fromString("Paimon")));

        String anchor = LocalDateTime.now().minusDays(1).toString();
//        String anchor = LocalDateTime.now().toString();

        // snapshot 2
        writeTable(
                "t",
                GenericRow.of(3, BinaryString.fromString("Test")),
                GenericRow.of(4, BinaryString.fromString("Case")));

//        assertThat(spark.sql("SELECT * FROM t").collectAsList().toString())
//                .isEqualTo("[[1,Hello], [2,Paimon], [3,Test], [4,Case]]");

        // time travel to snapshot 1
        assertThat(
                        spark.sql(String.format("SELECT * FROM t TIMESTAMP AS OF '%s'", anchor))
                                .collectAsList()
                                .toString())
                .isEqualTo("[[1,Hello], [2,Paimon]]");
    }
```
result is : 
```
org.opentest4j.AssertionFailedError: 
expected: "[[1,Hello], [2,Paimon]]"
 but was: "[]"
Expected :"[[1,Hello], [2,Paimon]]"
Actual   :"[]"
<Click to see difference>


	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at org.apache.paimon.spark.SparkTimeTravelITCase.testTravelToTimestampString(SparkTimeTravelITCase.java:91)
[...]

PR #3910, opened 2024-08-06
[core] when use from-timestamp read data from snapshot, if time < earliestSnapshot time , no data return (#3904)

chunji96:
…liestSnapshot time , no data return (#3904)

### Purpose

Linked issue: close #xxx


JingsongLi:
Can you add unit test and solve test failure?

chunji96:
> Can you add unit test and solve test failure?

ok,i will resolve it,thks @JingsongLi

chunji96:
in #3904 ,method org.apache.paimon.utils.SnapshotManager#earlierOrEqualTimeMills ,if earliestSnapShot.timeMillis() > timestampMills , will return earliestSnapShot. it will lead to  earliestSnapShot excluded when use `incremental-between-timestamp`. i had fix it ,please take a look,thks @JingsongLi

JingsongLi (approved):
+1

PR #3911, opened 2024-08-06
[core] Fix offset bug of bitmap index

hang8929201:
I found a bug in the process of serializing the bitmap index


JingsongLi (approved):
+1

PR #3913, opened 2024-08-06
[flink] create table validate options

MonsterChenzhuo:
### Purpose
A wrong table parameter can also be successfully built. I think it need validate options.
wecom-temp-79456-8b30745bfde073a039835c4e0678b158

### Tests
SchemaChangeITCase#testCreateTableWithInvalidOption


MonsterChenzhuo:
image
In executing the test cases for Hive3, the aforementioned problem was encountered.
cc @JingsongLi

JingsongLi:
Hi @MonsterChenzhuo , it is really hard to validate options.

1. We have so many options.
2. Some options have fallback options.
3. Issue of version compatibility, in some cases, older versions ignore some optimization options of newer versions.

MonsterChenzhuo:
> Hi @MonsterChenzhuo , it is really hard to validate options.
> 
> 1. We have so many options.
> 2. Some options have fallback options.
> 3. Issue of version compatibility, in some cases, older versions ignore some optimization options of newer versions.

😂

PR #3914, opened 2024-08-06
[core] Refactor ForkJoinPool to cached thread pool

JingsongLi:
### Purpose

- Use Cached thread pool to replace ForkJoinPool, Idle threads will be released.
- In `AbstractFileStoreScan.readAndMergeFileEntries`, do filter first, it is in memory filter. This can improve the parallelism of reading the manifest.


yuzelin (approved):
+1

PR #3915, opened 2024-08-06
[doc] Add Auxiliary Statements for spark

Zouxxyy:
### Purpose

Add Spark Auxiliary Statements page for paimon, refer to https://spark.apache.org/docs/latest/sql-ref-syntax.html#auxiliary-statements


JingsongLi:
+1

PR #3918, opened 2024-08-08
[spark] Support push down StringContains

ulysses-you:
### Purpose

This pr makes Spark engine supports push down `StringContains`, e.g., `col like '%xx%'`, and create a new predicate function `Contains` in Paimon to represent it.

### Tests

add test

### API and Format

no


ulysses-you:

JingsongLi (approved):
+1

PR #3919, opened 2024-08-08
[pom] Replaces the maven property with project property.

LinMingQiang:
### Purpose

In my production environment I get compilation errors : 
```
[ERROR] Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:2.13.0:check (spotless-check) on project paimon-parent: Execution spotless-check of goal com.diffplug.spotless:spotless-maven-plugin:2.13.0:check failed: Unable to locate file with path: ${maven.multiModuleProjectDirectory}/.scalafmt.conf: Could not find resource '${maven.multiModuleProjectDirectory}/.scalafmt.conf'. -> [Help 1]
```

Parameter ${maven.multiModuleProjectDirectory}  is not a recommended parameter. Use ${project.basedir} instead of it. 

References ：
https://stackoverflow.com/questions/29778262/what-is-maven-multimoduleprojectdirectory-used-for

https://github.com/apache/iotdb/pull/12982

Linked issue: close #xxx

### DIFF
--- a/.mvn/readme.txt
+++ b/.mvn/readme.txt
@@ -1,4 +1,0 @@
-The .mvn directory is needed to be able to use the ${maven.multiModuleProjectDirectory} property, since git cannot
-commit an empty directory, add this file.
-
-Once we do not use ${maven.multiModuleProjectDirectory}, we can remove this file and .mvn directory.
--- a/pom.xml
+++ b/pom.xml
@@ -828,19 +828,18 @@ under the License.
                             </importOrder>
 
                             <licenseHeader>
-                                <!-- replace it with ${project.rootDirectory} after maven 4.0.0, see MNG-7038 -->
-                                <file>${maven.multiModuleProjectDirectory}/copyright.txt</file>
+                                <file>copyright.txt</file>
                                 <delimiter>${spotless.delimiter}</delimiter>
                             </licenseHeader>
                         </java>
                         <scala>
                             <scalafmt>
                                 <version>3.4.3</version>
-                                <file>${maven.multiModuleProjectDirectory}/.scalafmt.conf</file>
+                                <file>.scalafmt.conf</file>
                             </scalafmt>
 
                             <licenseHeader>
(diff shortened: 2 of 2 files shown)


Zouxxyy:
The reason why introduce `${maven.multiModuleProjectDirectory} `  was that some users like to cd sub-module and then execute mvn, this work well with maven 3.3.1+. ${project.basedir} may not solve this.

A better way is to replace it with `${project.rootDirectory}` after maven 4.0.0, see MNG-7038 (The version is too high, so this solution is not considered)

LinMingQiang:
In this scenario, we should recommend using ```mvn -pl ```. But in many cases parameter ```${maven.multiModuleProjectDirectory}``` would not be set , which causes the entire project to fail to compile. I think production stability is more important. WDYT.

Zouxxyy:
> In this scenario, we should recommend using mvn -pl 

Agree, https://github.com/apache/paimon/pull/3063 is just added for users who don't know how to use it (see my comment in it). Actually it can be reverted.

Zouxxyy:
> But in many cases parameter ${maven.multiModuleProjectDirectory} would not be set 

Can you give more details, is maven 3.3.1- or a very high version?

LinMingQiang:
> > But in many cases parameter ${maven.multiModuleProjectDirectory} would not be set
> 
> Can you give more details, is maven 3.3.1- or a very high version?

My case is maven 3.2.5, this forced me to modify my pom.

LinMingQiang:
I think production stability is more important than this feature https://github.com/apache/paimon/pull/3063, because production failure has already happened to me , WDYT.

LinMingQiang:
Revert https://github.com/apache/paimon/pull/3063

JingsongLi:
What are the current issue? 

Why is your compilation errors? @LinMingQiang (In my env, it is fine)

JingsongLi:
I think better choice is upgrade your maven version.

PR #3920, opened 2024-08-08
[spark][core] spark compact with deletion vector

YannByron:
### Purpose

to support the compaction on append table with Deletion Vector, then once we do some updates/deletes in the Deletion Vector mode and write out some deletion vector index files, we can merge them to the normal data files by compaction.


paimon-core/src/main/java/org/apache/paimon/operation/AppendOnlyFileStoreWrite.java
```diff
@@ -137,6 +168,7 @@ protected RecordWriter<InternalRow> createWriter(
                         : new AppendOnlyCompactManager(
                                 compactExecutor,
                                 restoredFiles,
+                                dvIndexFileMaintainer,
```
JingsongLi, line 171: Only when skipCompaction is false, we can create `DeletionVectorIndexFileMaintainer`. And bucketMode must not be `BucketMode.BUCKET_UNAWARE`.

YannByron: I think so, but I can't confirm this bucket mode is not `BucketMode.BUCKET_UNAWARE`, so I still use the `bucketMode != BucketMode.BUCKET_UNAWARE` instead of `true`.

paimon-core/src/main/java/org/apache/paimon/operation/AppendOnlyFileStoreWrite.java
```diff
@@ -170,69 +197,109 @@ protected RecordWriter<InternalRow> createWriter(
...
-                            options.asyncFileWrite());
-            try {
-                rewriter.write(bucketReader(partition, bucket).read(toCompact));
-            } catch (Exception e) {
-                collectedExceptions = e;
-            } finally {
+
+            @Override
+            public List<DataFileMeta> rewrite(
+                    List<DataFileMeta> toCompact, List<DeletionFile> deletionFiles)
```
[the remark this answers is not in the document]
JingsongLi: Refactor for this: https://github.com/apache/paimon/pull/3963

PR #3923, opened 2024-08-08
[flink] Add test case lookup join update after predicate filter

discivigour:
### Purpose

Support lookup join update after predicate filter.

### Tests

- LookupJoinITCase


JingsongLi:
@discivigour Can you rebase master?

JingsongLi:
+1

PR #3924, opened 2024-08-08
[core] Fix FieldCountAgg should return the same data type as fieldType.

LinMingQiang:
### Purpose
```
CREATE  TABLE testAggregation ( 
id STRING  
,name STRING 
,dt STRING 
,PRIMARY KEY(id, dt) NOT ENFORCED 
,agg_count bigint 
) PARTITIONED BY (dt) WITH ( 
'bucket' = '1' 
,'auto-create' = 'false' 
,'fields.agg_count.aggregate-function' = 'count' 
,'connector' = 'paimon' 
,'merge-engine' = 'aggregation' 
);

insert datas to trigger compaction.
```
throw ClassCastException ：

image

FieldType is Bigint type, but return Integer type.

image

Eventually, an error occurs when executing method `getLong` :

image


JingsongLi (approved):
+1

PR #3927, opened 2024-08-09
[spark] Report output metric when writing

Zouxxyy:
### Purpose

Report output metric when writing

image

image


JingsongLi (approved):
+1

PR #3928, opened 2024-08-09
[spark] Support auto disable bucketed scan

ulysses-you:
### Purpose

This pr adds a new rule `DisableUnnecessaryPaimonBucketedScan` to support auto disable bucketed scan if the bucket scan is not actually effective i.e., there is no shuffle exchange been removed. This change is to avoid performance regression since the bucketed scan may have smaller parallelism than normal scan.

For example: a table with bucket key x but user join/group-by/partition-by on column y.

Note, this rule is inspired from Spark `DisableUnnecessaryBucketedScan` but work for v2 scan. 

### Tests

Add test.

### API and Format

no

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonScan.scala
+++ b/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonScan.scala
@@ -37,13 +37,13 @@ case class PaimonScan(
     filters: Seq[Predicate],
     reservedFilters: Seq[Filter],
     pushDownLimit: Option[Int],
-    disableBucketedScan: Boolean = false)
+    bucketedScanDisabled: Boolean = false)
   extends PaimonBaseScan(table, requiredSchema, filters, reservedFilters, pushDownLimit)
   with SupportsRuntimeFiltering
   with SupportsReportPartitioning {
 
-  def withDisabledBucketedScan(): PaimonScan = {
-    copy(disableBucketedScan = true)
+  def disableBucketedScan(): PaimonScan = {
+    copy(bucketedScanDisabled = true)
   }
 
   @transient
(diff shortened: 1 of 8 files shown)


JingsongLi:
It seems spark test failed.

ulysses-you:
@JingsongLi thank you for the reminder, it took me a while to find the root cause...

ulysses-you:
@JingsongLi @YannByron  do you have to take a look ? thank you

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonScan.scala
```diff
@@ -36,35 +36,53 @@ case class PaimonScan(
...
     reservedFilters: Seq[Filter],
-    pushDownLimit: Option[Int])
+    pushDownLimit: Option[Int],
+    disableBucketedScan: Boolean = false)
   extends PaimonBaseScan(table, requiredSchema, filters, reservedFilters, pushDownLimit)
   with SupportsRuntimeFiltering
   with SupportsReportPartitioning {
 
-  override def outputPartitioning(): Partitioning = {
+  def withDisabledBucketedScan(): PaimonScan = {
```
JingsongLi, line 45: disableBucketedScan

JingsongLi (commented):
Looks good to me! @ulysses-you 

Just left one minor comment.

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonScan.scala
ulysses-you: good naming! but it seems conflict with line40..

JingsongLi: bucketedScanDisabled and disableBucketedScan?

ulysses-you: +1, addressed

JingsongLi:
+1 Thanks @ulysses-you for the contribution. Merging...

PR #3929, opened 2024-08-09
[core] support deletion vector with input ChangelogProducer

wwj6591812:
1、Now input ChangelogProducer and DeletionVector are decoupled, so change the validate of DeletionVectors.
2、Format some code.

### Purpose

Linked issue: close #xxx


paimon-core/src/main/java/org/apache/paimon/table/source/snapshot/DeltaFollowUpScanner.java
```diff
@@ -33,6 +33,7 @@ public class DeltaFollowUpScanner implements FollowUpScanner {
     @Override
     public boolean shouldScanSnapshot(Snapshot snapshot) {
         if (snapshot.commitKind() == Snapshot.CommitKind.APPEND) {
+            System.out.println("here! : " + snapshot.id());
```
JingsongLi, line 36: remove this

JingsongLi:
+1

PR #3930, opened 2024-08-09
[core] support user define delimiter with list agg

wwj6591812:
1、In my company, user want set field delimiter when use list agg.
2、Eliminate todo left over from two years ago.

### Purpose

Linked issue: close #xxx


JingsongLi (approved):
+1

PR #3932, opened 2024-08-10
[fix] Remove count function

MonsterChenzhuo:
### Purpose
Remove count function.


LinMingQiang:
+1

JingsongLi (approved):
+1

JingsongLi:
Discussed with @LinMingQiang :

Assuming the base count value is 100 and delta is 100, the expected result should be 101, but the result is 2.

The reason is that when merging, it always starts from null, which is not a problem in scenarios such as sum. In the case of base merge delta in count, if the merge base starts with null and the result is 1, it means that the base value value will never be passed as the first accumulate in count aggregation.

PR #3933, opened 2024-08-10
[core] support check partition expire when batch mode or bounded stream

wwj6591812:
In my company, some paimon user have reported the following issue :
Everyday the batch jobs have a running time lower than config partition.expiration-check-interval, so many partitions that should have expired have not been expired.
Now add a conf, support check partition expire used in case of batch mode or bounded stream.

### Purpose

Linked issue: close #xxx


LinMingQiang:
You can trigger expiration by setting parameter `partition.expiration-check-interval = 0`

JingsongLi (approved):
+1

JingsongLi:
This option (default false) also looks good to me.

PR #3934, opened 2024-08-11
[core] Remove the feature that creates branches from snapshots.

LinMingQiang:
### Purpose

Relation PR : https://github.com/apache/paimon/pull/3787

Creating branches from snapshots will no longer be supported, branches can only be created through tags.


JingsongLi:
Please fix the test failure.

JingsongLi:
+1

PR #3937, opened 2024-08-12
[flink] Support Flink adaptive parallelism by default

JingsongLi:
### Purpose

Only write manifest cache is conflict with Flink adaptive parallelism, we can only assert it when enabling write manifest cache.


leaves12138 (approved):
+1

PR #3948, opened 2024-08-13
[core] Remove bounded check for append compact topology

JingsongLi:
### Purpose

#3936 hang the compact after the writer, making its lifecycle very easy to control.

So in the bounded streaming mode, we can have a compact topology.


leaves12138 (approved):
+1

PR #3950, opened 2024-08-13
[core] Refactor AppendDeletionFileMaintainer to separate bucketed and unaware

JingsongLi:
### Purpose

1. Rename to `AppendDeletionFileMaintainer`.
2. Separate bucketed and unaware implementations.


YannByron:
+1

PR #3952, opened 2024-08-13
[flink] When reading DataSplit, calculate fetch event time lag with earliest file creation time instead of latest

tsreaper:
### Purpose

Currently when reading `DataSplit`, we calculate `currentFetchEventTimeLag` with the latest file creation time in this split. This is incorrect because fetch event time lag should be defined by the oldest data in this split.

This PR changes this behavior by using the earliest file creation time to calculate fetch event time lag.

### Tests

Tested by hand.

### API and Format

No format changes.

### Documentation

No new feature.


JingsongLi:
+1

PR #3953, opened 2024-08-13
[flink] support history partition compaction in dedicated compaction job.

LsomeYeah:
### Purpose

Linked issue: close #xxx

If a partition has not received any new data writes for an extended period, we consider it to be a history partition. To further reduce the number of small files, it is necessary to perform a full compaction on history partitions.

This pr is to support doing a full compaction for history partition with a dedicated compaction job. Compared to dedicated compaction before:
you can set `history_time`, if a partition don't receive any new data for `history_time`, it will be full compacted.

**note: this feature is only supported in batch mode.**

### Tests

`CompactDatabaseActionITCase#testHistoryPartitionCompact`


JingsongLi:
1. Document this.
2. Support table compact too.
3. Support Spark SQL too.
4. Maybe we should find a better name for the new arguement.

JingsongLi (approved):
+1

PR #3955, opened 2024-08-13
[core] fix skew bug of stream read unware bucket table

wwj6591812:
Now when stream read a unware bucket table with comsume_id and consumer.mode = exactly-once. SKEW will happy, this is a bug, now fix it.
image

### Purpose

Linked issue: close #xxx

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/operator/MonitorFunction.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/operator/MonitorFunction.java
@@ -19,6 +19,7 @@
 package org.apache.paimon.flink.source.operator;
 
 import org.apache.paimon.flink.utils.JavaTypeInfo;
+import org.apache.paimon.table.BucketMode;
 import org.apache.paimon.table.sink.ChannelComputer;
 import org.apache.paimon.table.source.DataSplit;
 import org.apache.paimon.table.source.EndOfScanException;
@@ -38,6 +39,7 @@ import org.apache.flink.runtime.state.FunctionInitializationContext;
 import org.apache.flink.runtime.state.FunctionSnapshotContext;
 import org.apache.flink.streaming.api.checkpoint.CheckpointedFunction;
 import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.streaming.api.datastream.SingleOutputStreamOperator;
 import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
 import org.apache.flink.streaming.api.functions.source.RichSourceFunction;
 import org.apache.flink.streaming.api.watermark.Watermark;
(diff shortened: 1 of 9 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/operator/MonitorFunction.java
```diff
@@ -230,23 +234,52 @@ public static DataStream<RowData> buildSource(
...
-                                return ChannelComputer.select(key.f0, key.f1, numPartitions);
-                            }
-                            return ChannelComputer.select(key.f1, numPartitions);
-                        },
-                        split -> {
-                            DataSplit dataSplit = (DataSplit) split;
-                            return Tuple2.of(dataSplit.partition(), dataSplit.bucket());
-                        })
-                .transform(name + "-Reader", typeInfo, new ReadOperator(readBuilder));
+        if (shuffleByPartition) {
```
JingsongLi, line 259: You really want this? The `streaming-read.shuffle-by-partition` is true by default. It means?

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/AppendOnlyTableITCase.java
```diff
@@ -85,6 +85,22 @@ public void testReadWrite() {
         assertThat(rows).containsExactlyInAnyOrder(Row.of("AAA"), Row.of("BBB"));
     }
 
+    @Test
+    public void testReadWriteWithUnwareBucket() throws Exception {
```
JingsongLi, line 89: What is this test for?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/operator/MonitorFunction.java
wwj6591812: In my business scenario, I will set `streaming-read.shuffle-by-partition` to false.

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/AppendOnlyTableITCase.java
wwj6591812: Sorry, I have change the method name to testReadUnwareBucketTableWithRebalanceShuffle.

This test is to test that the data read under the following conditions is correct：
1、consumer.mode=exactly-once
2、Streaming read unware bucket table.
3、Send Split from MonitorFunction to ReadOperator is rebalance.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/operator/MonitorFunction.java
JingsongLi: This is bug, you should fix it.

wwj6591812: done

JingsongLi (approved):
Thanks @wwj6591812 , looks good to me!

PR #3958, opened 2024-08-14
[core] rename file_size_in_bytes to total_file_size_in_bytes

MonsterChenzhuo:
### Purpose

rename file_size_in_bytes to total_file_size_in_bytes.


MonsterChenzhuo:
cc @JingsongLi

JingsongLi:
Hi @MonsterChenzhuo , maybe it is better to just use bytes, this is SQL, if we return a string with special format, it is hard to use SQL to handle this type.

MonsterChenzhuo:
> Hi @MonsterChenzhuo , maybe it is better to just use bytes, this is SQL, if we return a string with special format, it is hard to use SQL to handle this type.

Is it reasonable to keep file_size_in_bytes in $files since it's a base table, but add data_size to $partitions?

JingsongLi:
Maybe just rename `file_size_in_bytes` to `total_file_size_in_bytes`?

MonsterChenzhuo:
> Maybe just rename `file_size_in_bytes` to `total_file_size_in_bytes`?
+1

JingsongLi:
You need to modify code not just documentation.

MonsterChenzhuo:
> You need to modify code not just documentation.

done.

> You need to modify code not just documentation.

done.

MonsterChenzhuo:
After thinking about it, I don't think it makes much sense to change the name.

PR #3964, opened 2024-08-15
chore: Update snappy version to 1.1.8.4 for aarch64 compatibility

zuston:
### Purpose

We have used paimon embedded in our spark/flink jobs, and we want to make these jobs run on the aarch64 machines, this PR is to make this  feasible

### Tests

aarch64 cpu arch compatibility: the compiled jar has been passed the alibaba's yoda tool check


zuston:
Minor upgrade, PTAL @JingsongLi

JingsongLi (approved):
+1

PR #3966, opened 2024-08-15
[core] Move schema validation of FallbackReadFileStoreTable from constructor to scanner

tsreaper:
### Purpose

Currently schema of main branch and fallback branch are validated in the constructor of `FallbackReadFileStoreTable`. However if the user mistakenly changed the schema of one branch, he won't be able to drop or alter this table, because the object cannot be created.

This PR fixes this issue by moving the validation before creating `TableScan`.

### Tests

### API and Format

No format changes.

### Documentation

No new feature.

### DIFF
--- a/paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/BranchSqlITCase.java
+++ b/paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/BranchSqlITCase.java
@@ -334,16 +334,23 @@ public class BranchSqlITCase extends CatalogITCaseBase {
     }
 
     @Test
-    public void testDifferentRowTypes() {
+    public void testDifferentRowTypes() throws Exception {
         sql(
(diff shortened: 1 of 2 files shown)


paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/BranchSqlITCase.java
```diff
@@ -342,7 +342,7 @@ public void testDifferentRowTypes() {
         sql("ALTER TABLE `t$branch_pk` ADD (v2 INT)");
         sql("ALTER TABLE t SET ( 'scan.fallback-branch' = 'pk' )");
 
-        assertThatThrownBy(() -> sql("INSERT INTO t VALUES (1, 10, 'apple')"))
+        assertThatThrownBy(() -> sql("SELECT * FROM t"))
```
JingsongLi, line 345: Add a test to alter table. This is the bug which you are fixing.

JingsongLi (approved):
+1

PR #3967, opened 2024-08-15
[core] AsyncPositionOutputStream should not flush after close.

leaves12138:
### Purpose

Add check for AsyncPositionOutputStream, it should not call flush after close is called.

Linked issue: close #xxx

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/fs/AsyncPositionOutputStream.java
+++ b/paimon-common/src/main/java/org/apache/paimon/fs/AsyncPositionOutputStream.java
@@ -52,6 +52,7 @@ public class AsyncPositionOutputStream extends PositionOutputStream {
 
     private int totalBuffers;
     private long position;
+    private boolean closed = false;
 
     public AsyncPositionOutputStream(PositionOutputStream out) {
         this.out = out;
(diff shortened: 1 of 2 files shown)


paimon-common/src/main/java/org/apache/paimon/fs/AsyncPositionOutputStream.java
```diff
@@ -172,6 +173,9 @@ public void write(byte[] b, int off, int len) throws IOException {
 
     @Override
     public void flush() throws IOException {
+        if (closed) {
+            throw new RuntimeException("Already closed");
```
JingsongLi, line 177: IOException

paimon-common/src/main/java/org/apache/paimon/fs/AsyncPositionOutputStream.java
```diff
@@ -204,6 +208,7 @@ public void close() throws IOException {
         } catch (ExecutionException e) {
             throw new RuntimeException(e);
         }
+        this.closed = true;
```
JingsongLi, line 211: close {
   if (closed) return
}

JingsongLi (approved):
+1

PR #3970, opened 2024-08-15
[core] OrphanFilesClean now support branches

tsreaper:
### Purpose

Currently `OrphanFilesClean` cannot handle data files created from branches and will mistakenly delete them. This PR modifies `OrphanFilesClean` so that it can support data from branches.

### Tests

IT cases.

### API and Format

No format changes.

### Documentation

No new feature.


JingsongLi:
+1

PR #3971, opened 2024-08-15
[spark][core] spark compact with deletion vector

YannByron:
### Purpose

to support the compaction on append table with Deletion Vector, then once we do some updates/deletes in the Deletion Vector mode and write out some deletion vector index files, we can merge them to the normal data files by compaction.


YannByron:
@JingsongLi please take a look.

JingsongLi (approved):
+1

PR #3975, opened 2024-08-15
 [core] fix the issue that BinaryExternalMerger spill files may be leaked.

### LINKED ISSUE #3974: [Bug] BinaryExternalMerger spill files may be leaked.
### Search before asking

### Paimon version

When `BinaryExternalMerger` merges spill files, if an exception occurs when writing a file, `ChannelReaderInputView` will not be closed, which means that even if `delete` is called, the file remains on the disk (marked as deleted by the operating system, which can be seen using `lsof -n | grep deleted` ) until the process exits.

### Compute Engine

flink 

### Minimal reproduce step

An `IOException` occurred during `AbstractBinaryExternalMerger#mergeChannels`.

Temporary files on disk should be properly cleaned up after an exception.

### Anything else?

_No response_

liming30:
### Purpose

Linked issue: close #3974 

 [core] fix the issue that BinaryExternalMerger spill files may be leaked.


JingsongLi (approved):
+1

PR #3976, opened 2024-08-16
[spark] Support cross partition write

Zouxxyy:
### Purpose

Support cross partition write for spark

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/commands/BucketProcessor.scala
+++ b/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/commands/BucketProcessor.scala
@@ -19,8 +19,7 @@
 package org.apache.paimon.spark.commands
 
 import org.apache.paimon.crosspartition.{GlobalIndexAssigner, KeyPartOrRow}
-import org.apache.paimon.data.{GenericRow, JoinedRow}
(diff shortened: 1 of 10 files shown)


paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/commands/BucketProcessor.scala
```diff
@@ -123,3 +131,105 @@ case class DynamicBucketProcessor(
...
+          val tuple: (KeyPartOrRow, Array[Byte]) = rowIterator.next()
+          val internalRow = SerializationUtils.deserializeBinaryRow(tuple._2)
+          tuple._1 match {
+            case KeyPartOrRow.KEY_PART => assigner.bootstrapKey(internalRow)
+            case KeyPartOrRow.ROW => assigner.processInput(internalRow)
+            case _ =>
+              throw new UnsupportedOperationException(s"unknown kind ${tuple._1}")
+          }
+        } else if (assigner.inBoostrap()) {
+          assigner.endBoostrap(true)
```
JingsongLi, line 221: Actually, only emit records here... So your heap memory will OOM.

I refactor a lazy method for this:
https://github.com/apache/paimon/pull/3978

JingsongLi (approved):
+1

PR #3981, opened 2024-08-16
[core] Catalog support list `sys` namespace for end user

ulysses-you:
### Purpose

This pr adds `sys` database when calling `listDatabases(includeSystem)` in `Catalog`, so that user can be aware of the Paimon sys database featrue.

However, it seems `listDatabases` is used by Paimon internal, e.g., list all table path, so only support catalog `listNamespaces` to only affects end-user.

### Tests

add test

### API and Format

no

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
@@ -27,6 +27,7 @@ import org.apache.paimon.schema.SchemaChange;
 import org.apache.paimon.table.Table;
 
 import java.io.Serializable;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
@@ -85,6 +86,23 @@ public interface Catalog extends AutoCloseable {
      * Get the names of all databases in this catalog.
      *
      * @return a list of the names of all databases
+     */
+    default List<String> listDatabases(boolean includeSystem) {
+        List<String> userDatabases = listDatabases();
+        if (includeSystem) {
+            List<String> allDatabases = new ArrayList<>(userDatabases.size() + 1);
+            allDatabases.add(SYSTEM_DATABASE_NAME);
+            allDatabases.addAll(userDatabases);
+            return allDatabases;
+        } else {
+            return userDatabases;
+        }
+    }
+
+    /**
+     * Get the names of user databases in this catalog.
+     *
+     * @return a list of the names of user databases
      */
     List<String> listDatabases();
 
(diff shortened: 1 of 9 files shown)


ulysses-you:
cc @JingsongLi thank you

paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
```diff
@@ -84,10 +85,23 @@ default Optional<MetastoreClient.Factory> metastoreClientFactory(Identifier iden
     /**
      * Get the names of all databases in this catalog.
      *
-     * @return a list of the names of all databases
+     * @return a list of the names of user databases
      */
     List<String> listDatabases();
```
JingsongLi, line 90: It seems we should modify the implementation of `listDatabases`?

Hi @ulysses-you 
Could you research whether system database in database systems can be viewed with a list database command, such as in MySQL, PostgreSQL, etc.

ulysses-you: yea, I tried mysql and PostgreSQL, both of them will show the sys databases if user has permission.
```sql
mysql> show databases;
+--------------------+
| Database           |
+--------------------+
| information_schema |
| mysql              |
| performance_schema |
| sys                |
+--------------------+
```

```sql
postgres=# \l
                                            数据库列表
   名称    | 拥有者 | 字元编码 | 校对规则 | Ctype | ICU Locale | Locale Provider |    存取权限
-----------+--------+----------+----------+-------+------------+-----------------+-----------------
 postgres  | cathy  | UTF8     | C        | C     |            | libc            |
 template0 | cathy  | UTF8     | C        | C     |            | libc            | =c/cathy       +
           |        |          |          |       |            |                 | cathy=CTc/cathy
 template1 | cathy  | UTF8     | C        | C     |            | libc            | =c/cathy       +
           |        |          |          |       |            |                 | cathy=CTc/cathy
(3 行记录)

```

JingsongLi: Thanks @ulysses-you 

We should implement `listDatabases` in `AbstractCatalog`.

ulysses-you: @JingsongLi  How about change ` List<String> listDatabases();` to ` List<String> listDatabases(boolean includeSystem);`? I'm afraid always return sys database will break somethings. I beleive the usage of `listDatabases` Paimon internal do not need sys database.

JingsongLi:
Hi @ulysses-you , where is this requirement from? If this requirement doesn't come from the user, I suggest we don't return the sys db. I checked that Iceberg also doesn't return it.

ulysses-you:
I see, leave it as is for now. thank you @JingsongLi

PR #3982, opened 2024-08-16
[core] Introduce Table.snapshot(long snapshotId)

JingsongLi:
### Purpose

We should expose more interfaces to compute engines, there may be various needs, and these data structures inherently need to consider compatibility; We can make them public.
1. Make Snapshot public.
2. Introduce Table snapshot(long snapshotId)


yuzelin (approved):
+1

PR #3991, opened 2024-08-19
[core] Add manifestListReader and manifestFileReader to Table

JingsongLi:
### Purpose

Next PR of #3982

We should expose more interfaces to compute engines, there may be various needs, and these data structures inherently need to consider compatibility; We can make them public.

- Add manifestListReader and manifestFileReader to Table
- Make ManifestFileMeta ManifestEntry DataFileMeta public.


yuzelin (approved):
+1

PR #3995, opened 2024-08-19
[core] Fix value filter pushdown in KeyValueFileStoreScan

JingsongLi:
### Purpose

Currently, there are issues with value filtering pushed down to the primary key table: 
1. We shouldn't apply full bucket filtering to incremental data, as it is meaningless and can lead to erroneous data.
2. Our determination of whether to push down value filtering varies between full and incremental data.


JingsongLi:
This will fix the bug in #3923 too.

tsreaper (approved):

PR #3996, opened 2024-08-19
[arrow] Add reader from arrow to paimon

leaves12138:
### Purpose

Add converter from arrow to paimon, for read arrow data

### DIFF
--- a/paimon-arrow/src/main/java/org/apache/paimon/arrow/converter/Arrow2PaimonBatchConverter.java
+++ b/paimon-arrow/src/main/java/org/apache/paimon/arrow/converter/Arrow2PaimonBatchConverter.java
@@ -1,0 +1,63 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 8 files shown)


paimon-arrow/src/main/java/org/apache/paimon/arrow/writer/ArrowFieldWriters.java
```diff
@@ -487,6 +487,48 @@ protected void doWrite(int rowIndex, DataGetters getters, int pos) {
         }
     }
 
+    /** Writer for TIMESTAMP & TIMESTAMP_LTZ. */
+    public static class TimestampWriterLZ extends ArrowFieldWriter {
```
JingsongLi, line 491: Remove this.

paimon-arrow/src/main/java/org/apache/paimon/arrow/converter/Arrow2PaimonBatchConverter.java
```diff
@@ -0,0 +1,63 @@
...
+import org.apache.paimon.data.columnar.ColumnarRow;
+import org.apache.paimon.data.columnar.VectorizedColumnBatch;
+import org.apache.paimon.types.DataType;
+
+import org.apache.arrow.vector.VectorSchemaRoot;
+
+import java.util.Iterator;
+
+/** Util to convert a {@link VectorSchemaRoot} to paimon rows. */
+public class Arrow2PaimonBatchConverter {
```
JingsongLi, line 32: It is better to introduce a `ArrowBatchReader` to reuse columnVectors.

JingsongLi:
+1

PR #4000, opened 2024-08-19
[migrate] Support migrate table with specific separator 

xuzifu666:
### Purpose

Linked issue: close #xxx

Support migrate table with specific separator

Currently options in migrate table only support separate by "," which cannot hold all conditions，such as if user want to set option like  **'orc.encrypt' = 'pii:id,name'** to do orc encrypt condition，this pr is aim to support with specific separator in options separate when call migrate_table.

ParameterUtils##parseCommaSeparatedKeyValues only separate keyValues with comma.

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/utils/ParameterUtils.java
+++ b/paimon-common/src/main/java/org/apache/paimon/utils/ParameterUtils.java
@@ -35,9 +35,13 @@ public class ParameterUtils {
     }
 
     public static Map<String, String> parseCommaSeparatedKeyValues(String keyValues) {
+        return parseSeparatedKeyValues(keyValues, ",");
+    }
+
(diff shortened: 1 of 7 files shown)


docs/content/migration/migration-from-hive.md
```diff
@@ -63,7 +63,7 @@ We can add our table properties while importing by sys.migrate_table('<database>
 <tableproperties> here should be separated by ",".  For example:
 
 ```sql
-CALL sys.migrate_table('hive', 'my_db.wait_to_upgrate', 'file.format=orc,read.batch-size=2096,write-only=true')
+CALL sys.migrate_table('hive', 'my_db.wait_to_upgrate', 'file.format=orc,read.batch-size=2096,write-only=true', ',')
```
JingsongLi, line 66: maybe it is better to introduce options_map, this arguement is standard SQL map

PR #4002, opened 2024-08-20
[arrow] Remove field vector set row count

leaves12138:
### Purpose

Every record set row count will slow down 10 times


yuzelin:
Fixed by #4012

PR #4005, opened 2024-08-20
[core] Add compaction when creating Iceberg compatible metadata

tsreaper:
### Purpose

Currently, new Iceberg compatible metadata are just appended to the metadata list. Without compaction, this list will become longer and longer, and the query speed will become slower.

This PR introduces compaction when creating Iceberg compatible metadata.

### Tests

Modify existing random tests to fit in this change.

### API and Format

No format changes.

### Documentation

No new feature.


JingsongLi (approved):
+1

PR #4008, opened 2024-08-20
[arrow] Read timestamp vector and process by paimon self

leaves12138:
### Purpose

Read timestamp vector and process by paimon self


yuzelin (approved):
+1

PR #4009, opened 2024-08-20
[core] Dynamic bucket mode cannot work with AQE

JingsongLi:
### Purpose

Dynamic bucket mode should know the distribution strategy to filter own data of a task.

If AQE enabled, there is no fixed distribution strategy, task cannot know the its own bucket index.

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkSink.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkSink.java
@@ -24,6 +24,7 @@ import org.apache.paimon.CoreOptions.TagCreationMode;
 import org.apache.paimon.manifest.ManifestCommittable;
 import org.apache.paimon.options.MemorySize;
 import org.apache.paimon.options.Options;
+import org.apache.paimon.table.BucketMode;
 import org.apache.paimon.table.FileStoreTable;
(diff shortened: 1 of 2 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkSink.java
```diff
@@ -222,8 +225,11 @@ public DataStream<Committable> doWrite(
                                         commitUser))
                         .setParallelism(parallelism == null ? input.getParallelism() : parallelism);
 
-        if (!isStreaming && table.coreOptions().writeManifestCache().getBytes() > 0) {
-            assertBatchAdaptiveParallelism(env, written.getParallelism());
+        boolean writeMCacheEnabled = table.coreOptions().writeManifestCache().getBytes() > 0;
+        boolean hashDynamicMode = table.bucketMode() == BucketMode.HASH_DYNAMIC;
```
Zouxxyy, line 229: what about cross_partition

JingsongLi: cross_partition does rely on reading index, pk index will shuffle with data records.

leaves12138 (approved):
+1

Zouxxyy (approved):
+1

PR #3004, opened 2024-03-13
[hive] add hive client pool

zhuangchong:
### Purpose

add hive client pool

PR #4010, opened 2024-08-20
[hive] bugfix jdbc paimon hive client pool (#3004)

mircodee:
### Purpose
1. Fix the bug linking Paimon to Hive's Catalog:
- The key of the cached item has not overridden the HashCode and Equals methods.
- The value of the cached item, that is, when creating the Hive client, the UGI information was not passed.

2. Add Paimon's options as the cache key to simplify user usage。
```
DROP CATALOG paimon_thrift;
CREATE EXTERNAL CATALOG `paimon_thrift`
PROPERTIES (
  "type"  =  "paimon",
  "paimon.catalog.type"  =  "hive",
  "paimon.catalog.warehouse"  =  "hdfs://localhost:9000/user/hive/warehouse/"
  "hive.metastore.uris"  =  "thrift://localhost:9083",
  'paimon.option.client-pool-cache.keys'='ugi,conf:*'
);
set catalog paimon_thrift;
show databases;
```

### Tests
1. Fix before
image

2. Fix after
image

### DIFF
--- a/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/pool/CachedClientPool.java
+++ b/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/pool/CachedClientPool.java
@@ -84,7 +84,14 @@ public class CachedClientPool implements ClientPool<IMetaStoreClient, TException
         this.clientClassName = clientClassName;
         init();
         // set ugi information to hms client
-        this.clientPool();
+        try {
+            run(client -> null);
+        } catch (TException e) {
+            throw new RuntimeException(e);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new RuntimeException(e);
+        }
     }
 
     @VisibleForTesting
@@ -191,8 +198,7 @@ public class CachedClientPool implements ClientPool<IMetaStoreClient, TException
             switch (type) {
                 case UGI:
                     try {
-                        elements.add(
-                                UserGroupInformationConf.of(UserGroupInformation.getCurrentUser()));
+                        elements.add(UserGroupInformation.getCurrentUser());
                     } catch (IOException e) {
                         throw new UncheckedIOException(e);
                     }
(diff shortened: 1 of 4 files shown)


JingsongLi:
Hi @mircodee , thanks for the contribution!

"`when creating the Hive client, the UGI information was not passed.`", it seems outer compute engine should invoke `UGI.doAs` for every catalog method invoking.

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/pool/CachedClientPool.java
```diff
@@ -77,9 +80,11 @@ public CachedClientPool(Configuration conf, Options options, String clientClassN
         this.conf = conf;
         this.clientPoolSize = options.get(CLIENT_POOL_SIZE);
         this.evictionInterval = options.get(CLIENT_POOL_CACHE_EVICTION_INTERVAL_MS);
-        this.key = extractKey(options.get(CLIENT_POOL_CACHE_KEYS), conf);
+        this.key = extractKey(options.get(CLIENT_POOL_CACHE_KEYS), conf, options);
         this.clientClassName = clientClassName;
         init();
+        // set ugi information to hms client
+        this.clientPool();
```
JingsongLi, line 87:
```
try {
            run(client -> null);
        } catch (TException e) {
            Thread.currentThread().interrupt();
        }
```

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/pool/CachedClientPool.java
```diff
@@ -239,11 +289,58 @@ public String value() {
...
+        public int hashCode() {
+            return Objects.hash(key, value);
+        }
+
         public static ConfElement of(String key, String value) {
             return new ConfElement(key, value);
         }
     }
 
+    static class UserGroupInformationConf {
```
JingsongLi, line 314: Why need to introduce `UserGroupInformationConf`? Revert this?

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/pool/HiveClientPool.java
```diff
@@ -45,6 +45,12 @@ public HiveClientPool(int poolSize, Configuration conf, String clientClassName)
         this.hiveConf = new HiveConf(conf, HiveClientPool.class);
         this.hiveConf.addResource(conf);
         this.clientClassName = clientClassName;
+        // set ugi information to hms client
+        try {
+            this.run(client -> null);
```
JingsongLi, line 50: remove this

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/pool/CachedClientPool.java
mircodee: It seems that "UserGroupInformation" returns different object instances for the same user, which cannot be used as a cache key.
image

JingsongLi:
+1

PR #4011, opened 2024-08-20
[core] Expose method to get PartitionEntries from TableScan

reswqa:
### Purpose

Expose method to get PartitionEntries from TableScan

### Tests

No need, we just expose a forwarding method.

### API and Format

Add new method to `TableScan` interface.

### Documentation

No need.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/FallbackReadFileStoreTable.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/FallbackReadFileStoreTable.java
@@ -25,6 +25,7 @@ import org.apache.paimon.data.InternalRow;
 import org.apache.paimon.disk.IOManager;
 import org.apache.paimon.manifest.ManifestEntry;
 import org.apache.paimon.manifest.ManifestFileMeta;
+import org.apache.paimon.manifest.PartitionEntry;
 import org.apache.paimon.metrics.MetricRegistry;
 import org.apache.paimon.predicate.Predicate;
 import org.apache.paimon.reader.RecordReader;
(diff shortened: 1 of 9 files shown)


paimon-core/src/main/java/org/apache/paimon/table/source/TableScan.java
```diff
@@ -38,6 +39,8 @@ public interface TableScan {
     /** List partitions. */
     List<BinaryRow> listPartitions();
 
+    List<PartitionEntry> getPartitionEntries();
```
JingsongLi, line 42: `listPartitionEntries`?

paimon-core/src/main/java/org/apache/paimon/table/source/TableScan.java
```diff
@@ -38,6 +39,8 @@ public interface TableScan {
     /** List partitions. */
     List<BinaryRow> listPartitions();
```
JingsongLi, line 40: this method can be a default method, to wrap `listPartitionEntries`.

paimon-core/src/main/java/org/apache/paimon/table/source/TableScan.java
JingsongLi, line 42: You should make `PartitionEntry` public.

paimon-core/src/main/java/org/apache/paimon/table/FallbackReadFileStoreTable.java
```diff
@@ -301,6 +302,15 @@ public List<BinaryRow> listPartitions() {
             partitions.addAll(fallbackScan.listPartitions());
             return new ArrayList<>(partitions);
         }
+
+        @Override
+        public List<PartitionEntry> listPartitionEntries() {
+            List<PartitionEntry> partitionEntries = mainScan.listPartitionEntries();
+            if (partitionEntries.isEmpty()) {
```
JingsongLi, line 309: if partition not exists in `mainScan`, add it into `partitionEntries`.

Please add test case too.

JingsongLi (approved):
+1

PR #4012, opened 2024-08-20
[arrow] Remove field vector set row count to avoid performance issue

yuzelin:
### Purpose

FieldVecotor.setValueCount() takes much time


leaves12138 (approved):
+1

PR #4014, opened 2024-08-20
[doc] Improve partition mark done action type documentation

zhuangchong:
### Purpose

Improve partition mark done action type documentation

### DIFF
--- a/docs/layouts/shortcodes/generated/core_configuration.html
+++ b/docs/layouts/shortcodes/generated/core_configuration.html
@@ -549,7 +549,7 @@ This config option does not affect the default filesystem metastore.</td>
             <td><h5>partition.mark-done-action</h5></td>
             <td style="word-wrap: break-word;">"success-file"</td>
             <td>String</td>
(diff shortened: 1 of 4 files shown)


JingsongLi (changes requested):
Hi @zhuangchong , we still don't know the usage of `mark-event`, I think it is good to keep it not in documentation.

zhuangchong:
> Hi @zhuangchong , we still don't know the usage of `mark-event`, I think it is good to keep it not in documentation.

The `mark-event` is to mark the partition completion by the `markPartitionForEvent` method provided by the hive metastore client.

 Then use the hive metastore client's isPartitionMarkedForEvent method to query whether the partition is completed.

image

JingsongLi (approved):
+1

JingsongLi:
> > Hi @zhuangchong , we still don't know the usage of `mark-event`, I think it is good to keep it not in documentation.
> 
> The `mark-event` is to mark the partition completion by the `markPartitionForEvent` method provided by the hive metastore client.
> 
> Then use the hive metastore client's isPartitionMarkedForEvent method to query whether the partition is completed.
> 
> image

Make sense, merging...

PR #4015, opened 2024-08-20
[spark] Support migrate_table with options_map 

xuzifu666:
### Purpose

Linked issue: close #xxx

Support migrate_table with options_map 

Support migrate table with specific separator

Currently options in migrate table only support separate by "," which cannot hold all conditions，such as if user want to set option like 'orc.encrypt' = 'pii:id,name' to do orc encrypt condition，this pr is aim to add a options_map to support it.

ParameterUtils##parseCommaSeparatedKeyValues only separate keyValues with comma.
image

### Tests

MigrateTableProcedureTest.scala


JingsongLi (approved):
+1

PR #4016, opened 2024-08-20
[core] MergeTreeWriter clear writer buffer at finally

ulysses-you:
### Purpose

`writeBuffer.clear()` should be put at finally block in case failure.

### Tests

Pass CI

### API and Format

no


JingsongLi (approved):
+1

PR #4023, opened 2024-08-21
[flink] FlinkCatalog should assert different path for creating table

JingsongLi:
### Purpose

Otherwise, it would be strange to create a table with a different path while the data is in another path.


zhuangchong (approved):
+1

PR #4026, opened 2024-08-21
[spark] Support report scan ordering

ulysses-you:
### Purpose

This pr makes `PaimonScan` support `SupportsReportOrdering`. For primary key table, we will do a sorted run during write so the primary keys in one file are always ordered, and skip optimization if the contains a rawConvertible split since we will not do merge sort read.

This can help eliminate local sort for smj/sorted agg etc..

### Tests

add tests

### API and Format

no

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonScan.scala
+++ b/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonScan.scala
@@ -23,8 +23,8 @@ import org.apache.paimon.table.{BucketMode, FileStoreTable, Table}
 import org.apache.paimon.table.source.{DataSplit, Split}
 
 import org.apache.spark.sql.PaimonUtils.fieldReference
-import org.apache.spark.sql.connector.expressions.{Expressions, NamedReference, Transform}
-import org.apache.spark.sql.connector.read.{SupportsReportPartitioning, SupportsRuntimeFiltering}
+import org.apache.spark.sql.connector.expressions.{Expressions, NamedReference, SortDirection, SortOrder, Transform}
+import org.apache.spark.sql.connector.read.{SupportsReportOrdering, SupportsReportPartitioning, SupportsRuntimeFiltering}
 import org.apache.spark.sql.connector.read.partitioning.{KeyGroupedPartitioning, Partitioning, UnknownPartitioning}
 import org.apache.spark.sql.sources.{Filter, In}
 import org.apache.spark.sql.types.StructType
@@ -40,7 +40,8 @@ case class PaimonScan(
     bucketedScanDisabled: Boolean = false)
   extends PaimonBaseScan(table, requiredSchema, filters, reservedFilters, pushDownLimit)
   with SupportsRuntimeFiltering
-  with SupportsReportPartitioning {
+  with SupportsReportPartitioning
+  with SupportsReportOrdering {
 
   def disableBucketedScan(): PaimonScan = {
     copy(bucketedScanDisabled = true)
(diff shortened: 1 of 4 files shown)


JingsongLi:
Thanks @ulysses-you for the contribution.

This optimization looks good to me!

But there is some potential issues.
Perhaps we need to add an interface to the ReadBuilder, because even for primary key tables, the reads may not be ordered. For example, if the deletion vectors mode or read-optimized mode is enabled.

paimon-core/src/main/java/org/apache/paimon/table/source/DataSplit.java
```diff
@@ -108,6 +108,17 @@ public boolean rawConvertible() {
...
+    /**
+     * To indicate if this `DataSplit` keep the raw table ordering. For example, for the primary key
+     * table, we will do sorted run during write and do merge read during read, so the data is
+     * sorted by the primary keys. Return `false` means the ordering is broken. If a `DataSplit` is
+     * `rawConvertible` then there is no sort merge read, so only if the data file number less than
+     * 2, we can get the correct data ordering.
+     */
+    public boolean keepOrdering() {
+        return !rawConvertible || dataFiles.size() < 2;
+    }
```
ulysses-you, line 120: cc @JingsongLi , If I get it correctly, this should address your concern ?

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonScan.scala
```diff
@@ -72,17 +73,56 @@ case class PaimonScan(
...
+
+    val primaryKeys = table match {
+      case fileStoreTable: FileStoreTable => fileStoreTable.primaryKeys().asScala
+      case _ => Seq.empty
+    }
+    if (primaryKeys.isEmpty) {
+      return Array.empty
+    }
+
+    val allSplitsKeepOrdering = lazyInputPartitions.toSeq
```
JingsongLi, line 104: Maybe we should make sure there is only one split in a bucket?

ulysses-you: thank you @JingsongLi, please correct me if wrong. Per my understanding, if a partition contains multi-splits using merge file read, then all splits should be ordered and never be overlapped ?

JingsongLi: Merging only occurs within a single split.

paimon-spark/paimon-spark-3.3/src/main/scala/org/apache/paimon/spark/PaimonScan.scala
```diff
@@ -0,0 +1,124 @@
...
+import org.apache.spark.sql.PaimonUtils.fieldReference
+import org.apache.spark.sql.connector.expressions.{Expressions, NamedReference, Transform}
+import org.apache.spark.sql.connector.read.{SupportsReportPartitioning, SupportsRuntimeFiltering}
+import org.apache.spark.sql.connector.read.partitioning.{KeyGroupedPartitioning, Partitioning, UnknownPartitioning}
+import org.apache.spark.sql.sources.{Filter, In}
+import org.apache.spark.sql.types.StructType
+
+import scala.collection.JavaConverters._
+
+case class PaimonScan(
```
JingsongLi, line 34: Is there possibility to extract a base class `PaimonBucketedScan` to reuse codes?

paimon-core/src/main/java/org/apache/paimon/table/source/DataSplit.java
```diff
@@ -108,6 +108,17 @@ public boolean rawConvertible() {
...
 
+    /**
+     * To indicate if this `DataSplit` keep the raw table ordering. For example, for the primary key
+     * table, we will do sorted run during write and do merge read during read, so the data is
+     * sorted by the primary keys. Return `false` means the ordering is broken. If a `DataSplit` is
+     * `rawConvertible` then there is no sort merge read, so only if the data file number less than
+     * 2, we can get the correct data ordering.
+     */
+    public boolean keepOrdering() {
+        return !rawConvertible || dataFiles.size() < 2;
```
JingsongLi, line 119: Maybe we don't need to introduce method here, just use `!rawConvertible || dataFiles.size() < 2` in spark.

ulysses-you:
thank you @JingsongLi , addressed comments

JingsongLi (approved):
Looks good to me!

PR #4027, opened 2024-08-21
[flink] Partition Mark done should not be affected by compaction

JingsongLi:
### Purpose

As long as the data is complete, there is no need to wait for the compaction to finish.

However, there are except cases, in deletion vectors mode, the zero level is ignored during batch reads. In this case, we should only wait for the compaction to complete.

### Tests

Add `PartitionMarkDoneTest`.


leaves12138 (approved):
+1

PR #4031, opened 2024-08-22
[arrow] Arrow reader should check schema match every time

leaves12138:
### Purpose

Maybe `VectorSchemaRoot` schema does not match projectRowType for now, we should make a fit


yuzelin (approved):
+1

PR #4032, opened 2024-08-22
[cdc] Convert timestamp_ltz field contents using local time zone

### LINKED ISSUE #4030: [Bug] Convert timestamp_ltz field contents using local time zone
### Search before asking

### Paimon version

0.8

### Compute Engine

flink

### Minimal reproduce step

1. Create a mysql table containing the timestamp type, and specify the time zone for mysql server to be UTC+8
2. Use flink-cdc to import the data of the table into kafka
3. Create a paimon table, which uses a timestamp_ltz type field to store the value of the timestamp type field in mysql
4. If you query the paimon table in UTC+8, you will find that there are still 8 hours missing

The content of the timestamp_ltz field retrieved from the Paimon table should be the same as in mysql

### Anything else?

I think the root cause is that we use the utc time zone to read the data from kafka, but use the current time zone to write the data to paimon.

image
image

rfyu:
### Purpose

Linked issue: close #4030


JingsongLi:
It seems test still failure...

PR #4033, opened 2024-08-22
[arrow] Add static element converter

leaves12138:
### Purpose

Sometimes, we need to convert a single element to arrow field vector

### DIFF
--- a/paimon-arrow/src/main/java/org/apache/paimon/arrow/vector/StaticElementConverter.java
+++ b/paimon-arrow/src/main/java/org/apache/paimon/arrow/vector/StaticElementConverter.java
@@ -21,6 +21,7 @@ package org.apache.paimon.arrow.vector;
 import org.apache.paimon.arrow.writer.ArrowFieldWriter;
 import org.apache.paimon.arrow.writer.ArrowFieldWriterFactoryVisitor;
 import org.apache.paimon.data.GenericRow;
+import org.apache.paimon.data.InternalRow;
 import org.apache.paimon.types.DataField;
 
 import org.apache.arrow.memory.BufferAllocator;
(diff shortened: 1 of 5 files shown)


paimon-arrow/src/main/java/org/apache/paimon/arrow/vector/StaticElementConverter.java
```diff
@@ -0,0 +1,61 @@
...
+import org.apache.paimon.arrow.writer.ArrowFieldWriterFactoryVisitor;
+import org.apache.paimon.data.InternalRow;
+import org.apache.paimon.types.DataField;
+
+import org.apache.arrow.memory.BufferAllocator;
+import org.apache.arrow.vector.FieldVector;
+
+import static org.apache.paimon.arrow.ArrowUtils.createVector;
+
+/** Convert a static value which contained in InternalRow to a FieldVector. */
```
yuzelin, line 31: Pass the object value and build GenericRow internal?

yuzelin (approved):
+1

paimon-arrow/src/main/java/org/apache/paimon/arrow/vector/StaticElementConverter.java
```diff
@@ -0,0 +1,63 @@
...
+/** Convert a static value to a FieldVector. */
+public class StaticElementConverter {
+
+    private final GenericRow row;
+    private final FieldVector fieldVector;
+    private final ArrowFieldWriter writer;
+
+    private int pos = 0;
+
+    public StaticElementConverter(
```
JingsongLi, line 40: Why this is a class? It seems a static method is OK.

JingsongLi: We can generate a max size static vector?

leaves12138: We can't fill the max size static vector, it too waste of memory

leaves12138: It's better to fill it when needed

PR #4037, opened 2024-08-22
[core] Async record reader should only work with .orc exactly

leaves12138:
### Purpose

Async record reader should only work with .orc exactly

Only orc reader batch could be put in queue.
Others like parquet or other kind of orc, reuse the batch


JingsongLi:
+1

PR #3973, opened 2024-08-15
[hive] Add a option for HiveCatalog to decide sync all properties to hms

xuzifu666:
### Purpose

Linked issue: close #xxx
Add a option for HiveCatalog to decide sync all properties to hms.
Currently HiveCatalog had not sync all tblproperties to hms，this not fitable for platform dev which need collect these metadata，so add option to control it default is false.

before fix:
```
CREATE TABLE paimon_q_02 (
  `user_id` STRING COMMENT '任务用到的库',
  `used_table` STRING COMMENT '任务用到的表',
  `day` STRING COMMENT '按天进行分区')
USING paimon
PARTITIONED BY (day)
TBLPROPERTIES(
  'file.format' = 'parquet',
  'file.compression' = 'ZSTD')
```
then TABLE_PARAMS not contains 'file.format' = 'parquet', 'file.compression' = 'ZSTD':
```
"tableParams":{"owner":"yuydws","totalSize":"623","numRows":"-1","rawDataSize":"-1","provider":"paimon","COLUMN_STATS_ACCURATE":"false","numFiles":"1","transient_lastDdlTime":"1723718372","storage_handler":"org.apache.paimon.hive.PaimonStorageHandler","table_type":"PAIMON"}
```
after fix:
then TABLE_PARAMS contains 'file.format' = 'parquet', 'file.compression' = 'ZSTD':
```
"tableParams":{"owner":"yuydws","totalSize":"623","numRows":"-1","rawDataSize":"-1","provider":"paimon","COLUMN_STATS_ACCURATE":"false","numFiles":"1","transient_lastDdlTime":"1723718372","file.format":"parquet","file.compression":"ZSTD","storage_handler":"org.apache.paimon.hive.PaimonStorageHandler","table_type":"PAIMON"}
```

PR #4038, opened 2024-08-22
[hive] Sync primary-key/partition-key/bucket-id when sync to hms

xuzifu666:
### Purpose

Linked issue: close #xxx

Follow up https://github.com/apache/paimon/pull/3973
currently primary-key partition-key bucket-id not sync to hms, add these info to properties when sync hms.
user condition is to show metadata in platform.

### DIFF
--- a/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
+++ b/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
@@ -521,6 +521,17 @@ public class HiveCatalog extends AbstractCatalog {
             tblProperties = convertToPropertiesPrefixKey(tableSchema.options(), HIVE_PREFIX);
         }
 
+        // add primary-key, partition-key, bucket-id to tblproperties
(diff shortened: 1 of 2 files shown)


paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
```diff
@@ -517,6 +517,18 @@ private Table createHiveTable(Identifier identifier, TableSchema tableSchema) {
         Map<String, String> tblProperties;
         if (syncAllProperties()) {
             tblProperties = new HashMap<>(tableSchema.options());
+
+            // add primary-key, partition-key, bucket-id to tblproperties
+            if (!tableSchema.primaryKeys().isEmpty()) {
+                tblProperties.put("primary-key", String.join(",", tableSchema.primaryKeys()));
```
JingsongLi, line 523: Use CoreOptions keys

xuzifu666: Thanks for reminder，Done @JingsongLi

JingsongLi (approved):
+1

PR #4039, opened 2024-08-22
[flink] Bypass operator should never block checkpoint

JingsongLi:
### Purpose

The previous `AppendBypassCoordinateOperator` spent a lot of time generating compact tasks when there were many files, leading to significant blocking of normal threads and even resulting in checkpoint delays.

Modify it:
1. Scanning in a separate thread.
2. Make sure there is no message in mail box, and send compact tasks to downstream.

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/AppendBypassCoordinateOperator.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/AppendBypassCoordinateOperator.java
@@ -21,17 +21,26 @@ package org.apache.paimon.flink.source;
 import org.apache.paimon.append.UnawareAppendCompactionTask;
 import org.apache.paimon.append.UnawareAppendTableCompactionCoordinator;
 import org.apache.paimon.table.FileStoreTable;
+import org.apache.paimon.utils.ExecutorUtils;
 
-import org.apache.flink.api.common.operators.ProcessingTimeService;
+import org.apache.flink.api.common.operators.MailboxExecutor;
+import org.apache.flink.api.common.operators.ProcessingTimeService.ProcessingTimeCallback;
 import org.apache.flink.streaming.api.operators.AbstractStreamOperator;
 import org.apache.flink.streaming.api.operators.ChainingStrategy;
 import org.apache.flink.streaming.api.operators.OneInputStreamOperator;
 import org.apache.flink.streaming.runtime.streamrecord.StreamRecord;
+import org.apache.flink.streaming.runtime.tasks.ProcessingTimeService;
+import org.apache.flink.streaming.runtime.tasks.mailbox.MailboxExecutorImpl;
 import org.apache.flink.types.Either;
 
 import java.util.List;
+import java.util.concurrent.Executors;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
 
 import static org.apache.paimon.utils.Preconditions.checkArgument;
+import static org.apache.paimon.utils.ThreadUtils.newDaemonThreadFactory;
 
 /**
  * A {@link OneInputStreamOperator} to accept commit messages and send append compact coordinate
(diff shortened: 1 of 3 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/AppendBypassCoordinateOperator.java
```diff
@@ -57,27 +76,49 @@ public void open() throws Exception {
...
-    public void onProcessingTime(long time) {
-        while (true) {
+    public void asyncPlan(UnawareAppendTableCompactionCoordinator coordinator) {
+        while (compactTasks.size() < MAX_PENDING_TASKS) {
             List<UnawareAppendCompactionTask> tasks = coordinator.run();
-            for (UnawareAppendCompactionTask task : tasks) {
-                output.collect(new StreamRecord<>(Either.Right(task)));
-            }
-
+            compactTasks.addAll(tasks);
```
wwj6591812, line 94: Should we first check if tasks.isEmpty()? If it's not empty, then compactTasks.addAll(tasks);

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/AppendBypassCoordinateOperator.java
```diff
@@ -57,27 +76,49 @@ public void open() throws Exception {
...
+                        newDaemonThreadFactory("Compaction Coordinator"));
+        this.executorService.scheduleWithFixedDelay(
+                () -> asyncPlan(coordinator), 0, intervalMs, TimeUnit.MILLISECONDS);
+        this.getProcessingTimeService().scheduleWithFixedDelay(this, 0, intervalMs);
     }
 
-    @Override
-    public void onProcessingTime(long time) {
-        while (true) {
+    public void asyncPlan(UnawareAppendTableCompactionCoordinator coordinator) {
```
wwj6591812, line 91: private

wwj6591812:
Hi, thx for this  @JingsongLi 

I have a doubt that you say generating compact tasks blocking normal threads and even resulting in checkpoint delays.
But In my test, our checkpoint is normal. But the problem is that the first operator parallelism is 1024, and the Compact Coordinator parallelism is 1. This leads to the phenomenon of back pressure.

So, should we support both this new topology and old union topology simultaneously?

image

JingsongLi:
> Hi, thx for this @JingsongLi
> 
> I have a doubt that you say generating compact tasks blocking normal threads and even resulting in checkpoint delays. But In my test, our checkpoint is normal. But the problem is that the first operator parallelism is 1024, and the Compact Coordinator parallelism is 1. This leads to the phenomenon of back pressure.
> 
> So, should we support both this new topology and old union topology simultaneously?
> 
> image
> 
> image

What is your point? You mean you use new code and the job still backpresure?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/AppendBypassCoordinateOperator.java
JingsongLi: Why?

wwj6591812:
> @wwj6591812
> 
> What is your point? You mean you use new code and the job still back pressure? If not, please valid this first.

ok，after merge，I will test soon.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/AppendBypassCoordinateOperator.java
wwj6591812: > Why?

Because if tasks.isEmpty(), then there is no need to execute compactTasks.addAll (tasks);

wwj6591812: Because if tasks.isEmpty(), then there is no need to execute compactTasks.addAll (tasks);

JingsongLi: So what is the benifit?

leaves12138 (approved):
+1

PR #4042, opened 2024-08-23
[arrow] Pass arrow reader last value by serializer to binary row

leaves12138:
### Purpose

When merge, the last value will be referenced by maxKey. If reader close, the maxKey will be useless. We must avoid this situation

### DIFF
--- a/paimon-arrow/src/main/java/org/apache/paimon/arrow/reader/ArrowBatchReader.java
+++ b/paimon-arrow/src/main/java/org/apache/paimon/arrow/reader/ArrowBatchReader.java
@@ -23,6 +23,7 @@ import org.apache.paimon.data.InternalRow;
 import org.apache.paimon.data.columnar.ColumnVector;
 import org.apache.paimon.data.columnar.ColumnarRow;
 import org.apache.paimon.data.columnar.VectorizedColumnBatch;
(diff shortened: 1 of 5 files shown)


paimon-arrow/src/main/java/org/apache/paimon/arrow/reader/ArrowBatchReader.java
```diff
@@ -85,7 +88,9 @@ public boolean hasNext() {
                     public InternalRow next() {
                         columnarRow.setRowId(position);
                         position++;
-                        return columnarRow;
+                        return position == rowCount
```
JingsongLi, line 91: add some comments

paimon-arrow/src/main/java/org/apache/paimon/arrow/vector/ArrowFormatWriter.java
```diff
@@ -92,10 +92,14 @@ public boolean empty() {
         return rowId == 0;
     }
 
-    @Override
-    public void close() {
+    // if c++ code release this, we don't need to release again (release twice may cause a problem)
+    public void release() {
         array.release();
```
JingsongLi, line 97: move array and schema out of this class.

yuzelin (approved):
+1

PR #4043, opened 2024-08-23
[core] Fix caching catalog case sensitive

ulysses-you:
### Purpose

The CacaingCatalog should respect the wrapped catalog case sensitive, otherwise it would case wrong result.

```
create table t (c1 int);
drop table T;
create table t (c1 int);

[TABLE_OR_VIEW_ALREADY_EXISTS] Cannot create table or view `default`.`T` because it already exists.
```

This pr adds a new option `cache.case-sensitive` to choose the caching catalog case sensitive.

### Tests

add test

### API and Format

no

### DIFF
--- a/docs/layouts/shortcodes/generated/catalog_configuration.html
+++ b/docs/layouts/shortcodes/generated/catalog_configuration.html
@@ -39,7 +39,7 @@ under the License.
             <td>Controls whether the catalog will cache databases, tables and manifests.</td>
         </tr>
         <tr>
-            <td><h5>cache-case-sensitive</h5></td>
+            <td><h5>cache.case-sensitive</h5></td>
             <td style="word-wrap: break-word;">true</td>
             <td>Boolean</td>
             <td>Indicates the caching catalog is case sensitive or not.</td>
--- a/docs/layouts/shortcodes/generated/catalog_configuration.html
+++ b/docs/layouts/shortcodes/generated/catalog_configuration.html
@@ -39,6 +39,12 @@ under the License.
             <td>Controls whether the catalog will cache databases, tables and manifests.</td>
         </tr>
         <tr>
+            <td><h5>cache-case-sensitive</h5></td>
+            <td style="word-wrap: break-word;">true</td>
+            <td>Boolean</td>
+            <td>Indicates the caching catalog is case sensitive or not.</td>
+        </tr>
+        <tr>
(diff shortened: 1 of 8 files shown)


paimon-spark/paimon-spark-common/src/test/scala/org/apache/paimon/spark/sql/DDLTestBase.scala
```diff
@@ -456,4 +456,12 @@ abstract class DDLTestBase extends PaimonSparkTestBase {
     }.getMessage
     assert(error.contains("Unsupported partition transform"))
   }
+
+  test("Fix caching catalog case sensitive") {
```
JingsongLi, line 460: I ran this test in the master branch, it passed.

ulysses-you: oh, sorry for the missleading. It seems the cache future is disabled in tests.. I will modify the test.

ulysses-you:
MacOS filesystem is case insensitive, so the FileSystemCatalog becomes insensitive implicitly.. anyway, I add one more option to choose if make caching catalog case insensitive.

JingsongLi:
> MacOS filesystem is case insensitive, so the FileSystemCatalog becomes insensitive implicitly.. anyway, I add one more option to choose if make caching catalog case insensitive.

I see, actually we had this option before, which was later changed to `allow-upper-case`. There's an odd history behind this: since Flink doesn't natively support case-insensitivity, we cannot store uppercase objects in a case-insensitive storage because if Flink requests an object in uppercase, but the storage returns it in lowercase due to its case-insensitive nature, Flink would fail with an error stating that the object cannot be found.

ulysses-you:
@JingsongLi thank you for the context, I sse the option about allow_upper_case which has different default value for different backend engine and catalog, which makes me a bit confused before.. What do you think about this new option ? I think it at least can resovle some issues mannually and I see the iceberg also has a dedicated option for caching catalog.

JingsongLi:
> @JingsongLi thank you for the context, I see the option about allow_upper_case which has different default value for different backend engine and catalog, which makes me a bit confused before.. What do you think about this new option ? I think it at least can resovle some issues mannually and I see the iceberg also has a dedicated option for caching catalog.

If there is no applicable scenario at present, let's maintain the current status quo for now.

PR #4047, opened 2024-08-23
[hive] repair table support sync all properties to hms 

herefree:
### Purpose

Linked issue: close #xxx
Follow up #4038

### DIFF
--- a/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
+++ b/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
@@ -518,20 +518,8 @@ public class HiveCatalog extends AbstractCatalog {
         if (syncAllProperties()) {
             tblProperties = new HashMap<>(tableSchema.options());
 
-            // add primary-key, partition-key, bucket-id to tblproperties
-            if (!tableSchema.primaryKeys().isEmpty()) {
-                tblProperties.put(
-                        CoreOptions.PRIMARY_KEY.key(), String.join(",", tableSchema.primaryKeys()));
-            }
-            if (!tableSchema.partitionKeys().isEmpty()) {
-                tblProperties.put(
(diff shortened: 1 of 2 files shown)


paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
```diff
@@ -819,6 +819,20 @@ private void updateHmsTable(Table table, Identifier identifier, TableSchema sche
     private void updateHmsTablePars(Table table, TableSchema schema) {
         if (syncAllProperties()) {
             table.getParameters().putAll(schema.options());
+            if (!schema.primaryKeys().isEmpty()) {
```
JingsongLi, line 822: Please extract a method to reuse.

JingsongLi:
> Just curious，in real scene Paimon should not support alter pk，partitionkey？ @herefree

Yes

herefree:
> Just curious，in real scene Paimon should not support alter pk，partitionkey？ @herefree
Yes, if you have some history paimon table which was not syn pk or partitionkey to hive meta, you can use repair table.It also need to add some code to support it, just like this.
image

xuzifu666:
> > Just curious，in real scene Paimon should not support alter pk，partitionkey？ @herefree
> 
> Yes, if you have some history paimon table which was not syn pk or partitionkey to hive meta, you can use repair table.It also need to add some code to support it, just like this. image

Get，but the title maybe not fitable，alter primary-key can take a look at SparkSchemaEvolutionITCase##testSetAndRemoveOption which can not support.

herefree:
> > > Just curious，in real scene Paimon should not support alter pk，partitionkey？ @herefree
> > 
> > 
> > Yes, if you have some history paimon table which was not syn pk or partitionkey to hive meta, you can use repair table.It also need to add some code to support it, just like this. image
> 
> Get，but the title maybe not fitable，alter primary-key can take a look at SparkSchemaEvolutionITCase##testSetAndRemoveOption which can not support.

Yes,you are right,i'll change the title.

JingsongLi:
+1

PR #4053, opened 2024-08-23
[core] Introduce Format table to support csv,orc,parquet tables

JingsongLi:
### Purpose

See https://cwiki.apache.org/confluence/display/PAIMON/PIP-27%3A+Introduce+Format+Table+in+Paimon+Catalog


paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
```diff
@@ -198,7 +199,12 @@ public List<String> listTables(String databaseName) throws DatabaseNotExistExcep
             throw new DatabaseNotExistException(databaseName);
         }
 
-        return listTablesImpl(databaseName).stream().sorted().collect(Collectors.toList());
+        List<String> tables =
+                listTablesImpl(databaseName).stream().sorted().collect(Collectors.toList());
+        if (formatTableEnabled()) {
+            tables.addAll(listFormatTables(databaseName));
```
adrian-wang, line 205: Considering format table and normal paimon table are both tables, in most implementation they should be stored together, I prefer providing only one interface to list all tables no matter what kind of tables they are.

adrian-wang:
+1, LGTM, thanks!

PR #4056, opened 2024-08-26
[core] Expire Iceberg metadata when number of snapshots exceeds limit

tsreaper:
### Purpose

Currently all previously generated Iceberg metadata are not deleted, so the number of metadata files will become larger.

This PR implements expiration of Iceberg metadata. When the number of Iceberg snapshots exceeds limit, related files will be removed.

### Tests

Unit tests.

### API and Format

No format changes.

### Documentation

Document will be added later.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/iceberg/AbstractIcebergCommitCallback.java
+++ b/paimon-core/src/main/java/org/apache/paimon/iceberg/AbstractIcebergCommitCallback.java
@@ -18,6 +18,7 @@
 
 package org.apache.paimon.iceberg;
 
+import org.apache.paimon.CoreOptions;
 import org.apache.paimon.Snapshot;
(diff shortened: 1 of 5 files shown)


paimon-core/src/main/java/org/apache/paimon/iceberg/AbstractIcebergCommitCallback.java
```diff
@@ -90,60 +89,33 @@ public abstract class AbstractIcebergCommitCallback implements CommitCallback {
             ConfigOptions.key("metadata.iceberg.compaction.max.file-num")
                     .intType()
                     .defaultValue(50);
+    static final ConfigOption<Integer> SNAPSHOT_NUM_RETAINED =
```
JingsongLi, line 92: Can we just use the snapshot retain from Paimon table?

JingsongLi:
+1

PR #4058, opened 2024-08-26
[spark] Refactor Paimon write by name

ulysses-you:
### Purpose

This pr refactor the write by-name to improve the behavior in some user cases:

- Support fill with null as default value if user specifies columns less than table, e.g.,
```sql
CREATE TABLE t (c1 int, c2 string);

INSERT INTO TABLE t BY NAME SELECT 1 as c1;
```

- provide more user-friendly error msg if fail

  there are several exceptions, e.g., the column name mismatch, one more column names matched, etc.

### Tests

add tests

### API and Format

no


ulysses-you:

YannByron:
+1

PR #4059, opened 2024-08-26
[spark] Adjust ExpireSnapshotsProcedure param type

askwang:
### Purpose

- change param `older_than` type from TimeStampType to StringType, TimeStampType is not easy to understand and use.
- add new param `time_retained`, which can call `expire_snapshots` periodically. `older_than` is suitable for one-time call.

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/procedure/ExpireSnapshotsProcedure.java
+++ b/paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/procedure/ExpireSnapshotsProcedure.java
@@ -20,6 +20,9 @@ package org.apache.paimon.spark.procedure;
 
 import org.apache.paimon.options.ExpireConfig;
 import org.apache.paimon.table.ExpireSnapshots;
+import org.apache.paimon.utils.DateTimeUtils;
+import org.apache.paimon.utils.StringUtils;
+import org.apache.paimon.utils.TimeUtils;
 
 import org.apache.spark.sql.catalyst.InternalRow;
 import org.apache.spark.sql.connector.catalog.Identifier;
@@ -29,10 +32,10 @@ import org.apache.spark.sql.types.StructField;
 import org.apache.spark.sql.types.StructType;
 
 import java.time.Duration;
+import java.util.TimeZone;
 
 import static org.apache.spark.sql.types.DataTypes.IntegerType;
 import static org.apache.spark.sql.types.DataTypes.StringType;
-import static org.apache.spark.sql.types.DataTypes.TimestampType;
 
 /** A procedure to expire snapshots. */
 public class ExpireSnapshotsProcedure extends BaseProcedure {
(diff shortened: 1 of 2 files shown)


askwang:
help cc @JingsongLi

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/procedure/ExpireSnapshotsProcedure.java
```diff
@@ -42,8 +45,9 @@ public class ExpireSnapshotsProcedure extends BaseProcedure {
                 ProcedureParameter.required("table", StringType),
                 ProcedureParameter.optional("retain_max", IntegerType),
                 ProcedureParameter.optional("retain_min", IntegerType),
-                ProcedureParameter.optional("older_than", TimestampType),
-                ProcedureParameter.optional("max_deletes", IntegerType)
+                ProcedureParameter.optional("older_than", StringType),
+                ProcedureParameter.optional("max_deletes", IntegerType),
+                ProcedureParameter.optional("time_retained", StringType)
```
JingsongLi, line 50: older_than and time_retained? Looks not easy to understand, we should just keep one.

JingsongLi: Because there is currently compatibility issue. I recommend using a script to address the issue of periodic calls when invoking from the outer layer.

askwang: > older_than and time_retained? Looks not easy to understand, we should just keep one.

These two parameters can not be used together.

askwang: > Because there is currently compatibility issue. I recommend using a script to address the issue of periodic calls when invoking from the outer layer.

When using a script to calls periodically, we need to adjust older_than parameter each time. I think it is necessary to keep time_retained,  what do you think?

askwang: Thanks. @JingsongLi Removed time_retained

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/procedure/ExpireSnapshotsProcedure.java
```diff
@@ -85,9 +88,21 @@ public InternalRow[] call(InternalRow args) {
                     if (retainMin != null) {
                         builder.snapshotRetainMin(retainMin);
                     }
-                    if (olderThanMills != null) {
-                        builder.snapshotTimeRetain(
-                                Duration.ofMillis(System.currentTimeMillis() - olderThanMills));
+                    if (!StringUtils.isNullOrWhitespaceOnly(olderThanStr)) {
+                        long olderThanMills;
+                        // forward compatibility for timestamp type
```
JingsongLi, line 93: How to forward compatibility for timestamp type? It is a string instead of a timestamp.

askwang: In default, the value of timestamp type is like "1724664000000", it's actually a string that contains numerics. And get it by 
`Long olderThanMills = args.isNullAt(3) ? null : args.getLong(3) / 1000;`

JingsongLi: `change param older_than type from TimeStampType to StringType, TimeStampType is not easy to understand and use.`

Is this related? Can you show the use cases? How to improve the usage.

askwang: got, has only retain StringType

JingsongLi (approved):
+1

PR #4060, opened 2024-08-26
[refactor] Make ArrowUtils#toArrowField public

yuzelin:
### Purpose

This method is useful. For example, this need it:
https://github.com/apache/paimon-python/pull/8/


JingsongLi:
+1

PR #4061, opened 2024-08-26
[spark] Fix bootstrap record reader not close

ulysses-you:
### Purpose

Fix bootstrap record reader not close

### Tests

N/A

### API and Format

no


JingsongLi:
+1

PR #4068, opened 2024-08-27
[core] Add Table.indexManifestFileReader

JingsongLi:
### Purpose

Same to manifestListReader and manifestFileReader. We can also make indexManifestFileReader public.


wangmj17:
+1

PR #4070, opened 2024-08-27
[arrow] Add ArrowBundleRecords for a VectorSchemaRoot records.

leaves12138:
### Purpose

Wrap VectorSchemaRoot as a BundleRecords.


JingsongLi:
+1

PR #4074, opened 2024-08-27
[core] Introduce level0FileCount for partitions table

MonsterChenzhuo:
### Purpose

Linked issue: close #xxx

### Tests

PartitionsTableTest#testLevel0FileCountValue


JingsongLi:
Thanks @MonsterChenzhuo for the contribution.

But what is usage of level0file?

MonsterChenzhuo:
> Thanks @MonsterChenzhuo for the contribution.
> 
> But what is usage of level0file?

In the scenario where a table has the DelVector enabled, users can quickly determine whether data has been written and whether compaction has been completed by checking the level0file, especially when the current partition has no data.

However, if you use $files, you may notice that querying a very large table can be extremely slow, and the results may not be intuitive. This often requires users to perform aggregation to interpret the results.

JingsongLi:
> > Thanks @MonsterChenzhuo for the contribution.
> > But what is usage of level0file?
> 
> In the scenario where a table has the DelVector enabled, users can quickly determine whether data has been written and whether compaction has been completed by checking the level0file, especially when there is no data found for the current partition
> 
> However, if you use $files, the results may not be intuitive. This often requires users to perform aggregation to interpret the results.

But this is depends on per bucket? We should know the maxLevel0FilesInBucket and avgLevel0FilesInBucket, maybe it is better to just let them in metrics.

MonsterChenzhuo:
> DelVector

Metrics reporting currently relies on the Flink jobs that write data into the table. However, in our enterprise environment, some scenarios involve users querying Paimon tables through a general SQL client tool. This tool supports multiple query engines, such as Spark, Flink, etc., and is not aware of the underlying Flink jobs or the corresponding monitoring.

> > > Thanks @MonsterChenzhuo for the contribution.
> > > But what is usage of level0file?
> > 
> > 
> > In the scenario where a table has the DelVector enabled, users can quickly determine whether data has been written and whether compaction has been completed by checking the level0file, especially when there is no data found for the current partition
> > However, if you use $files, the results may not be intuitive. This often requires users to perform aggregation to interpret the results.
> 
> But this is depends on per bucket? We should know the maxLevel0FilesInBucket and avgLevel0FilesInBucket, maybe it is better to just let them in metrics.


Metrics reporting currently relies on the Flink jobs that write data into the table. However, in some scenarios, users query Paimon tables through a general SQL client tool that is not aware of the underlying Flink jobs or the corresponding monitoring.
In these cases, it might be more convenient to use system tables to view the information.

MonsterChenzhuo:
> > > Thanks @MonsterChenzhuo for the contribution.
> > > But what is usage of level0file?
> > 
> > 
> > In the scenario where a table has the DelVector enabled, users can quickly determine whether data has been written and whether compaction has been completed by checking the level0file, especially when there is no data found for the current partition
> > However, if you use $files, the results may not be intuitive. This often requires users to perform aggregation to interpret the results.
> 
> But this is depends on per bucket? We should know the maxLevel0FilesInBucket and avgLevel0FilesInBucket, maybe it is better to just let them in metrics.



> maxLevel0FilesInBucket and avgLevel0FilesInBucket,

For real-time writes to the Paimon table, we use real-time compaction and collect metrics to monitor maxLevel0FilesInBucket and avgLevel0FilesInBucket. However, for scenarios with infrequent updates (such as T+1) that require high throughput and low consumption, using offline compaction to monitor the number of L0 files through metrics feels less convenient compared to using system tables.

JingsongLi:
It seems a specific usage, let's wait future requirements.

PR #4075, opened 2024-08-27
[core] Fix deletion vectors mode with custom sequence field

JingsongLi:
### Purpose

This fix just lookup whole record to compute sequence. There is an opportunity to return sequence fields, but it looks very hard.


Zouxxyy (approved):
+1, thanks!

PR #4082, opened 2024-08-28
[spark] Support v2 write with overwrite

ulysses-you:
### Purpose

This pr supports v2 write with overwrite, e.g.:

```sql
spark.range(2).writeTo("t").overwrite(col("id") === 'b')
```

Before this pr, it would throw: `Only EqualNullSafe should be used when run INSERT OVERWRITE`

This pr only support overwrite on partition column with `Equal` and `EqualNullSafe` to be compatible with v1 write overwrite.

### Tests

add test

### API and Format

no


YannByron:
+1

PR #4083, opened 2024-08-28
[flink] Do not chain for AppendBypassCompactWorkerOperator

JingsongLi:
### Purpose

It should be a separate node, and no chain is good to rescale parallelism.


leaves12138 (approved):
+1

PR #4084, opened 2024-08-28
[spark] support to merge into in the case that primarykey's column data types are different

YannByron:
### Purpose

Linked issue: close #xxx


YannByron:
@Zouxxyy please take a look.

YannByron:
Maybe it should not be supported in this case.

PR #4085, opened 2024-08-28
[spark] Spark package should exclude zstd-jni

ulysses-you:
### Purpose

Different Spark branch uses different zstd-jni version, so it's better to use Spark's zstd-jni to avoid conflict. The reason why Paimon includes zstd-jni is: Flink uses airlift to support zstd so Flink distribution does not include zstd-jni.

### Tests

manually test, after this pr did not include `com/github/luben/zstd`
image

### API and Format

no


JingsongLi (approved):
+1

PR #4090, opened 2024-08-29
[core] Add field statistics in IcebergDataFileMeta

tsreaper:
### Purpose

This PR adds field statistics in `IcebergDataFileMeta`, so Iceberg can filter files when reading our metadata. We can also use statistics to test if the result of `IcebergConversions` can be correctly recognized by Iceberg.

### Tests

Unit tests.

### API and Format

No format changes.

### Documentation

Document will be added later.


JingsongLi (approved):
+1

PR #4097, opened 2024-08-29
[core] Optimize memory of compact coordinator

JingsongLi:
### Purpose

If it collect all files into memory, it is very easy to OOM. We can filter the files by file size (less than compaction file size).

### Tests

`AppendOnlyTableCompactionTest.testScanSkipBigFiles`.


leaves12138 (approved):
+1

PR #4098, opened 2024-08-29
[flink] fix the source data fetcher thread leak

### LINKED ISSUE #4092: [Bug] The Source Data Fetcher thread may not exit during job failover
### Search before asking

### Paimon version

0.9

### Compute Engine

Flink

### Minimal reproduce step

image

If the task block at poll method, it will hung forever. Besides, this lead to the FileSourceSplitReader can not be close which will bring the memory leak

image

We should wakeup from the `poll` method during the shutdown

### Anything else?

_No response_

Aitozi:
### Purpose

Linked issue: close #4092

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/FileStoreSourceSplitReader.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/FileStoreSourceSplitReader.java
@@ -26,6 +26,7 @@ import org.apache.paimon.reader.RecordReader.RecordIterator;
 import org.apache.paimon.table.source.DataSplit;
 import org.apache.paimon.table.source.Split;
 import org.apache.paimon.table.source.TableRead;
+import org.apache.paimon.utils.Pool;
 
 import org.apache.flink.connector.base.source.reader.RecordsWithSplitIds;
 import org.apache.flink.connector.base.source.reader.splitreader.SplitReader;
@@ -33,13 +34,15 @@ import org.apache.flink.connector.base.source.reader.splitreader.SplitsAddition;
 import org.apache.flink.connector.base.source.reader.splitreader.SplitsChange;
 import org.apache.flink.connector.file.src.reader.BulkFormat;
 import org.apache.flink.connector.file.src.util.MutableRecordAndPosition;
-import org.apache.flink.connector.file.src.util.Pool;
 import org.apache.flink.connector.file.src.util.RecordAndPosition;
 import org.apache.flink.table.data.RowData;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nullable;
 
 import java.io.IOException;
+import java.time.Duration;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.LinkedList;
@@ -50,6 +53,8 @@ import java.util.Set;
 /** The {@link SplitReader} implementation for the file store source. */
 public class FileStoreSourceSplitReader
         implements SplitReader<BulkFormat.RecordIterator<RowData>, FileStoreSourceSplit> {
+
+    private static final Logger LOG = LoggerFactory.getLogger(FileStoreSourceSplitReader.class);
 
     private final TableRead tableRead;
 
@@ -65,6 +70,7 @@ public class FileStoreSourceSplitReader
     private RecordIterator<InternalRow> currentFirstBatch;
 
     private boolean paused;
+    private volatile boolean wakeup;
     private final FileStoreSourceReaderMetrics metrics;
 
     public FileStoreSourceSplitReader(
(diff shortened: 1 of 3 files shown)


Aitozi:
CC @JingsongLi

JingsongLi:
Comment in issue.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/FileStoreSourceSplitReader.java
```diff
@@ -86,11 +92,21 @@ public RecordsWithSplitIds<BulkFormat.RecordIterator<RowData>> fetch() throws IO
...
+
         checkSplitOrStartNext();
 
         // pool first, pool size is 1, the underlying implementation does not allow multiple batches
         // to be read at the same time
         FileStoreRecordIterator iterator = pool();
+        if (iterator == null) {
+            LOG.info("Skip waiting for object pool due to wakeup: {}", wakeup);
+            wakeup = false;
+            return new RecordsWithPausedSplit<>();
```
JingsongLi, line 108: Rename it to `EmptyRecordsWithSplitIds`?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/FileStoreSourceSplitReader.java
```diff
@@ -86,11 +92,21 @@ public RecordsWithSplitIds<BulkFormat.RecordIterator<RowData>> fetch() throws IO
             return new RecordsWithPausedSplit<>();
         }
 
+        if (wakeup) {
```
JingsongLi, line 95: Do we need to wakeup here?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/FileStoreSourceSplitReader.java
```diff
@@ -114,12 +130,16 @@ private boolean reachLimit() {
...
 
     private FileStoreRecordIterator pool() throws IOException {
-        try {
-            return this.pool.pollEntry();
-        } catch (InterruptedException e) {
-            Thread.currentThread().interrupt();
-            throw new IOException("Interrupted");
+        FileStoreRecordIterator iterator = null;
+        while (iterator == null && !wakeup) {
+            try {
```
JingsongLi, line 135: if (wakeup) {
   wakeup = false;
}

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/FileStoreSourceSplitReader.java
```diff
@@ -65,6 +70,7 @@ public class FileStoreSourceSplitReader
     private RecordIterator<InternalRow> currentFirstBatch;
 
     private boolean paused;
+    private volatile boolean wakeup;
```
JingsongLi, line 73: An `AtomicBoolean` is better.

JingsongLi (changes requested):

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/FileStoreSourceSplitReader.java
Aitozi: done

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/FileStoreSourceSplitReader.java
Aitozi: I think we need, and it's meant to reset the wakeup flag. If not, we should reset the `wakeup` at the end of this method. And short-cut at the beginning looks better. WDYT?

JingsongLi: Maybe this should be in `pool`.

JingsongLi: And you can change method name to `poll`....

Aitozi: I mean not only `poll` should wakeup. I think the wakeup should work on the next `fetch` call not on the next `poll`. So after the `fetch` the wakeup should reset to normal.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/FileStoreSourceSplitReader.java
Aitozi: done

JingsongLi:
+1

PR #4099, opened 2024-08-29
[core] skip clean tag manifest file if exception

wwj6591812:
### Purpose

Now when clean tag file, skip datafile if exception happen.
Also skip clean manifest if exception happen.

Linked issue: close #xxx

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/utils/TagManager.java
+++ b/paimon-core/src/main/java/org/apache/paimon/utils/TagManager.java
@@ -41,6 +41,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Comparator;
 import java.util.List;
+import java.util.Set;
 import java.util.SortedMap;
 import java.util.TreeMap;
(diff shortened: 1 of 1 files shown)


paimon-core/src/main/java/org/apache/paimon/utils/TagManager.java
```diff
@@ -235,8 +237,21 @@ private void doClean(
         }
 
         // delete manifests
-        tagDeletion.cleanUnusedManifests(
-                taggedSnapshot, tagDeletion.manifestSkippingSet(skippedSnapshots));
+        success = true;
+        Set<String> manifestSkippingSet = new HashSet<>();
```
LinMingQiang, line 241: Variable 'manifestSkippingSet' should been initialized to null.

wwj6591812: ok，done

LinMingQiang:
+1

JingsongLi (approved):
+1

PR #4103, opened 2024-08-30
[core] Introduce spill-compression.zstd-level to control zstd level

JingsongLi:
### Purpose

Now the zstd level of spill compression must be 1, it is good to expose an option to control the zstd level.


yuzelin (approved):
+1

PR #4106, opened 2024-08-30
[flink] Check that cannot set non-existed fallback branch

yuzelin:
### Purpose

If set non-existed fallback branch, user cannot access the table.


JingsongLi (approved):
+1

PR #4108, opened 2024-08-31
[doc] Factor out most recent Flink version to docs config, and bump it

kaygorodov:
Documentation 

### Purpose

The doc page about programmatic Flink API doesn't refer to the latest paimon-flink-_flink_version_ package. This PR factors out the version and bumps it to 1.20.0

### Tests

built the site and tested the page locally

### API and Format

NA

### Documentation

NA

### DIFF
--- a/docs/config.toml
+++ b/docs/config.toml
@@ -43,6 +43,9 @@ pygmentsUseClasses = true
   # The branch for this version of Apache Paimon
   Branch = "master"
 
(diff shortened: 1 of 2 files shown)


kaygorodov:
The affected page with changes when serving hugo locally:

Screenshot 2024-08-31 at 9 50 53 PM

JingsongLi:
+1

PR #4110, opened 2024-09-02
[flink] refactor the code of the lookup and support computing the changelog generated by compact during read time.

### LINKED ISSUE #3868: [Feature] When the MergeEngine of the dimension table is aggregation / partial-update, there is no need to forcibly enable changelog-producer.

liming30:
### Purpose

Linked issue: close #3868 

[flink] refactor the code of the lookup and support computing the changelog generated by compact during read time.

This is used when changelog producer is none, but CoreOptions#needLookup is true and the table is used as a dim table.

### Tests

- org.apache.paimon.flink.lookup.LookupTableTest#testFullCacheLookupTableWithForceLookup
- org.apache.paimon.flink.lookup.LookupTableTest#testPartialLookupTableWithForceLookup

### API and Format

No

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/source/DataTableStreamScan.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/source/DataTableStreamScan.java
@@ -219,7 +219,7 @@ public class DataTableStreamScan extends AbstractDataTableScan implements Stream
         return false;
     }
 
-    private FollowUpScanner createFollowUpScanner() {
+    protected FollowUpScanner createFollowUpScanner() {
         CoreOptions.StreamScanMode type =
                 options.toConfiguration().get(CoreOptions.STREAM_SCAN_MODE);
         switch (type) {
@@ -249,7 +249,7 @@ public class DataTableStreamScan extends AbstractDataTableScan implements Stream
         return followUpScanner;
     }
 
-    private BoundedChecker createBoundedChecker() {
+    protected BoundedChecker createBoundedChecker() {
         Long boundedWatermark = options.scanBoundedWatermark();
         return boundedWatermark != null
                 ? BoundedChecker.watermark(boundedWatermark)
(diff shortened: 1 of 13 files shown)


JingsongLi:
I commented on the issue, and I want you to really think about whether this change adds real value to the business. Is it just to save on generating the changelog? Is that really the main cost involved?

JingsongLi:
Currently, your scenario involves using a partial-update table for dimension tables:
1. 90% of the scenarios are based on primary key lookup joins. The current implementation requires a lookup, which is why you introduced force-lookup.
2. 10% of the scenarios are based on non-primary key lookup joins. The current implementation requires changelog, but you don't want every dimension table to generate changelog, so this PR aims to address that.

I got it, and I am OK with this.

JingsongLi:
Could you consider extracting the implementation from Core? This functionality is only intended for Flink Lookup Join.

liming30:
@JingsongLi hi, please help review it when you have time.

JingsongLi (approved):
+1

PR #4112, opened 2024-09-02
[core] optimize snapshot system table when specifying snapshot_id

xuzifu666:
### Purpose

Linked issue: close #xxx
Currently select * from T$snapshots where snapshot_id=xxx would scan all snapshot file(call snapshotManager##snapshotsWithinRange which not push down), which would cost more IO when snapshot number is large. so add filter predicate to optizime it maybe necessary.

after the pr select * from T$snapshots where snapshot_id=xxx would scan snapshot file only once.
image


JingsongLi (changes requested):
Can you take a look to `SnapshotsRead.withFilter`? Is there any problem?

xuzifu666:
Pushdown is OK in SnapshotReader, so close the pr.

PR #4115, opened 2024-09-03
[core] Interface writeBundle should be with existing statExtractor

leaves12138:
### Purpose

Add limit for writeBundle interface


paimon-core/src/main/java/org/apache/paimon/io/StatsCollectingSingleFileWriter.java
```diff
@@ -73,6 +73,15 @@ public void write(T record) throws IOException {
         }
     }
 
+    @Override
+    public void writeBundle(BundleRecords bundle) throws IOException {
+        Preconditions.checkState(
+                simpleStatsExtractor != null,
```
JingsongLi, line 79: incorrect, should be null or disabled.

JingsongLi (approved):
+1

PR #4116, opened 2024-09-03
[core] Fix 'Trying to access closed classloader' in Flink

JingsongLi:
### Purpose

This fixes:
https://github.com/apache/paimon/actions/runs/10645769652/job/29511901738

```
Caused by: java.lang.IllegalStateException: Trying to access closed classloader. Please check if you store classloaders directly or indirectly in static fields. If the stacktrace suggests that the leak occurs in a third party library and cannot be fixed immediately, you can disable this check with the configuration 'classloader.check-leaked-classloader'.
```

Flink will close previous classloader, so in the async thread pool, we should set thread classloader.

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/utils/ThreadPoolUtils.java
+++ b/paimon-common/src/main/java/org/apache/paimon/utils/ThreadPoolUtils.java
@@ -120,8 +120,14 @@ public class ThreadPoolUtils {
     public static <U> void randomlyOnlyExecute(
             ExecutorService executor, Consumer<U> processor, Collection<U> input) {
         List<Future<?>> futures = new ArrayList<>(input.size());
+        ClassLoader cl = Thread.currentThread().getContextClassLoader();
         for (U u : input) {
-            futures.add(executor.submit(() -> processor.accept(u)));
+            futures.add(
+                    executor.submit(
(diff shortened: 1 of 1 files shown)


paimon-common/src/main/java/org/apache/paimon/utils/ThreadPoolUtils.java
```diff
@@ -120,17 +120,29 @@ private void advanceIfNeeded() {
...
             ExecutorService executor, Consumer<U> processor, Collection<U> input) {
         List<Future<?>> futures = new ArrayList<>(input.size());
+        ClassLoader cl = Thread.currentThread().getContextClassLoader();
         for (U u : input) {
-            futures.add(executor.submit(() -> processor.accept(u)));
+            futures.add(
+                    executor.submit(
+                            () -> {
+                                Thread.currentThread().setContextClassLoader(cl);
+                                processor.accept(u);
```
leaves12138, line 129: After processor.accept(u); set the classloader back to previous?

JingsongLi: No need, the thread pool is only for this usage.

leaves12138 (approved):
+1

PR #4117, opened 2024-09-03
[core] fix query schema throw exception when specifying schema_id does not exist

herefree:
### Purpose

Before getting schema from SchemaManager , we should determine if a schema exists.


JingsongLi (approved):
+1

PR #4118, opened 2024-09-03
[core] FileBasedBloomFilter should not be pass private member to cache manager.

leaves12138:
### Purpose

Cache manager cause out of memory while lookup merge

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/utils/FileBasedBloomFilter.java
+++ b/paimon-common/src/main/java/org/apache/paimon/utils/FileBasedBloomFilter.java
@@ -20,6 +20,7 @@ package org.apache.paimon.utils;
 
 import org.apache.paimon.annotation.VisibleForTesting;
 import org.apache.paimon.io.PageFileInput;
+import org.apache.paimon.io.cache.CacheCallback;
 import org.apache.paimon.io.cache.CacheKey;
 import org.apache.paimon.io.cache.CacheManager;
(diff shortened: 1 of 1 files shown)


paimon-common/src/main/java/org/apache/paimon/utils/FileBasedBloomFilter.java
```diff
@@ -58,12 +58,13 @@ public boolean testHash(int hash) {
...
         // so we introduce a refresh count to reduce refresh
         if (accessCount == REFRESH_COUNT || filter.getMemorySegment() == null) {
+            final BloomFilter bloomFilter = filter;
             MemorySegment segment =
                     cacheManager.getPage(
                             CacheKey.forPosition(input.file(), readOffset, readLength),
                             key -> input.readPosition(readOffset, readLength),
-                            key -> filter.unsetMemorySegment());
-            filter.setMemorySegment(segment, 0);
+                            key -> bloomFilter.unsetMemorySegment());
```
JingsongLi, line 66: add a static inner class to do this unset.

JingsongLi:
+1

Aitozi:
Hi @leaves12138 , why this will lead to out of memory ?

leaves12138:
> Hi @leaves12138 , why this will lead to out of memory ?

If we use lambda, the member "filter" will be indexed. The instance which contains the "filter" will also be referred, cannot gc

PR #4119, opened 2024-09-03
[spark] Analyze size through scan splits instead of listing files

Zouxxyy:
### Purpose

Analyze size through scan splits instead of listing files


JingsongLi (approved):
+1

PR #4123, opened 2024-09-04
[spark] Add tests for buceted scan with partitioned table

ulysses-you:
### Purpose

This pr adds test for buceted scan with partitioned table. Although we do not support bucketed scan with multi-bucket columns, there is a chance to support multi primary keys if  primary keys contains partition column.

### Tests

add test

### API and Format

no

### DIFF
--- a/paimon-spark/paimon-spark-common/src/test/scala/org/apache/paimon/spark/sql/BucketedTableQueryTest.scala
+++ b/paimon-spark/paimon-spark-common/src/test/scala/org/apache/paimon/spark/sql/BucketedTableQueryTest.scala
@@ -25,7 +25,7 @@ import org.apache.spark.sql.execution.SortExec
 import org.apache.spark.sql.execution.adaptive.AdaptiveSparkPlanHelper
 import org.apache.spark.sql.execution.exchange.ShuffleExchangeLike
 
-class c extends PaimonSparkTestBase with AdaptiveSparkPlanHelper {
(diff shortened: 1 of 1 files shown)


ulysses-you:
cc @JingsongLi thank you!

JingsongLi:
re-trigger test

paimon-spark/paimon-spark-common/src/test/scala/org/apache/paimon/spark/sql/BucketedTableQueryTest.scala
```diff
@@ -25,7 +25,7 @@ import org.apache.spark.sql.execution.SortExec
 import org.apache.spark.sql.execution.adaptive.AdaptiveSparkPlanHelper
 import org.apache.spark.sql.execution.exchange.ShuffleExchangeLike
 
-class BucketedTableQueryTest extends PaimonSparkTestBase with AdaptiveSparkPlanHelper {
+class c extends PaimonSparkTestBase with AdaptiveSparkPlanHelper {
```
Zouxxyy, line 28: Is it a typo?

ulysses-you: great catch! Not sure how it happened..

JingsongLi:
+1

PR #4125, opened 2024-09-04
[spark] Create table should not add spark.paimon.xxx config

askwang:
### Purpose
When we set `spark.paimon.xxx` config,  then create new table will add those configs into table properties. Actually, these configs  should only be used for queries and not for creating tables.

```sql
spark-sql> set `spark.paimon.scan.snapshot-id`=`2`;
spark-sql> create table t1(id int, name string) using paimon TBLPROPERTIES('primary-key' = 'id','bucket'='1');

spark-sql> show create table t1;
CREATE TABLE t1 (
  id INT NOT NULL,
  name STRING)
USING paimon
TBLPROPERTIES (
  'bucket' = '1',
  'path' = '/env/warehouse/paimon-warehouse/default.db/t1',
  'primary-key' = 'id',
  'scan.mode' = 'from-snapshot',
  'scan.snapshot-id' = '2')
```

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/SparkCatalog.java
+++ b/paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/SparkCatalog.java
@@ -60,7 +60,6 @@ import static org.apache.paimon.options.CatalogOptions.ALLOW_UPPER_CASE;
 import static org.apache.paimon.spark.SparkCatalogOptions.DEFAULT_DATABASE;
 import static org.apache.paimon.spark.SparkTypeUtils.toPaimonType;
 import static org.apache.paimon.spark.util.OptionUtils.copyWithSQLConf;
-import static org.apache.paimon.spark.util.OptionUtils.mergeSQLConf;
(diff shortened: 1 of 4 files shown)


paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/SparkCatalog.java
```diff
@@ -390,7 +390,7 @@ private static SchemaChange.Move getMove(
 
     private Schema toInitialSchema(
             StructType schema, Transform[] partitions, Map<String, String> properties) {
-        Map<String, String> normalizedProperties = mergeSQLConf(properties);
+        Map<String, String> normalizedProperties = mergeSQLConf(properties, false);
```
Zouxxyy, line 393: We can just remove this line without modifying mergeSQLConf

askwang: good idea~

Zouxxyy (approved):
+1

JingsongLi (approved):
+1

PR #4127, opened 2024-09-04
Remove StringUtils#isBlank which is same with #isNullOrWhitespaceOnly

yuzelin:
### Purpose

StringUtils#isBlank is redundant, and it's not as clear as isNullOrWhitespaceOnly.


JingsongLi (approved):
+1

PR #4130, opened 2024-09-05
[orc] Move config write.batch-size to CoreOptions

leaves12138:
### Purpose

write.batch-size should be in CoreOptions rather than OrcOptions.

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
@@ -878,7 +878,13 @@ public class CoreOptions implements Serializable {
             key("read.batch-size")
                     .intType()
                     .defaultValue(1024)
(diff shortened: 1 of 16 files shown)


paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -878,7 +878,13 @@ public class CoreOptions implements Serializable {
             key("read.batch-size")
                     .intType()
                     .defaultValue(1024)
-                    .withDescription("Read batch size for orc and parquet.");
+                    .withDescription("Read batch size for any file format if it supports.");
+
+    public static final ConfigOption<Integer> WRITE_BATCH_SIZE =
+            key("write.batch-size")
```
JingsongLi, line 884: fallbackKeys: orc.write.batch-size

JingsongLi (approved):
+1

PR #4135, opened 2024-09-05
[core] Support snapshot system pushdown with CompoundPredicate and fix FileNotFound error when snapshot id not exist

xuzifu666:
### Purpose

Linked issue: close #xxx
The pr had 2 point to support:
1. currently query snapshot system with predicate only support leaf predicate for total push down to SnapshotManager, the pr handle CompoundPredicate to support multiple predicate and all push down to SnapshotManager
when select * from T$snapshots where snapshot_id > 1 and snapshot < 5;

before fix:
image

after fix:
image

2. Snapshot system table not handle with not exist snapshot id query, it would error out with FileNotFould error. this pr handle it.

image


JingsongLi (approved):
+1

PR #4144, opened 2024-09-07
[cdc] Refactor codes cdcTimestampExtractor to separate different data …

LinMingQiang:
…sources.

### Purpose

Currently , we use ```CdcTimestampExtractorFactory``` to create different cdc datasources cdcTimestampExtractor,  and ```CdcTimestampExtractorFactory```  has import all kind cdc datasources. So when I run `MySqlSyncTableAction` and set ```tag.automatic-creation=watermark```, ClassNotFoundException will be thrown, other cdc dataSource class can't be found like ```MongoDBSource/PulsarSource```.

image

```java.lang.ClassNotFoundException ```

image

Linked issue: close #xxx


JingsongLi (approved):
+1

PR #4145, opened 2024-09-08
[core] Manifest table query throw exception with range of snapshot when snapshot id not exist

xuzifu666:
… 

### Purpose
Currently user can only query table manifest info of snapshot from scan.snapshot-id, due to snapshot manager of table, some early snapshot would clean. So maybe need offer user about snapshot range when user query a snapshot id which not exist.

Linked issue: close #xxx

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/system/ManifestsTable.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/system/ManifestsTable.java
@@ -40,7 +40,11 @@ import org.apache.paimon.table.source.TableRead;
 import org.apache.paimon.types.BigIntType;
 import org.apache.paimon.types.DataField;
 import org.apache.paimon.types.RowType;
-import org.apache.paimon.utils.*;
+import org.apache.paimon.utils.FileStorePathFactory;
+import org.apache.paimon.utils.IteratorRecordReader;
+import org.apache.paimon.utils.ProjectedRow;
+import org.apache.paimon.utils.SerializationUtils;
+import org.apache.paimon.utils.SnapshotManager;
 
 import org.apache.paimon.shade.guava30.com.google.common.collect.Iterators;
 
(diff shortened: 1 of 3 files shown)


paimon-core/src/main/java/org/apache/paimon/table/system/ManifestsTable.java
```diff
@@ -195,7 +195,16 @@ private static List<ManifestFileMeta> allManifests(FileStoreTable dataTable) {
...
-        if (snapshotId != null && snapshotManager.snapshotExists(snapshotId)) {
+        if (snapshotId != null) {
+            // reminder user with snapshot id range
+            if (!snapshotManager.snapshotExists(snapshotId)) {
+                Long earliestSnapshotId = snapshotManager.earliestSnapshotId();
+                Long latestSnapshotId = snapshotManager.latestSnapshotId();
+                throw new RuntimeException(
+                        String.format(
+                                "scan.snapshot-id is not exist, you can set it in range from %s to %s",
+                                earliestSnapshotId, latestSnapshotId));
```
LinMingQiang, line 206: better : Specified scan.snapshot-id [%s] is not exist, .....

xuzifu666: Thanks，your advice is nice. @LinMingQiang

paimon-core/src/main/java/org/apache/paimon/table/system/ManifestsTable.java
```diff
@@ -195,7 +195,16 @@ private static List<ManifestFileMeta> allManifests(FileStoreTable dataTable) {
...
+        if (snapshotId != null) {
+            // reminder user with snapshot id range
+            if (!snapshotManager.snapshotExists(snapshotId)) {
+                Long earliestSnapshotId = snapshotManager.earliestSnapshotId();
+                Long latestSnapshotId = snapshotManager.latestSnapshotId();
+                throw new RuntimeException(
+                        String.format(
+                                "Specified scan.snapshot-id %s is not exist, you can set it in range from %s to %s",
+                                snapshotId, earliestSnapshotId, latestSnapshotId));
+            }
```
LinMingQiang, line 207: SnapshotNotExistException

LinMingQiang:
+1

JingsongLi (approved):
+1

PR #4146, opened 2024-09-09
[core] Support schema table push down range query predicate

xuzifu666:
### Purpose

Linked issue: close #xxx
Currently range query with schema table not support range push down, this pr had two points changed:
1. support range query push down to Schema Table, which may improve query with schema in conditions such as schema changed frequently;
2. throw out and reminder user schema id range when user query not exist schema id.

### Tests

CatalogTableITCase##testSchemasTable with adding more case for the method.


JingsongLi (approved):
+1

PR #4150, opened 2024-09-10
[core] Only keep deleted Identifiers in memory for Manifest full merge

JingsongLi:
### Purpose

In Manifest Full Merge, we can just keep deleted Identifiers in memory to reduce OOM.

To do this, we must to add extraFiles to Identifier, because it is a kind of changing.


leaves12138 (approved):
+1

PR #4151, opened 2024-09-10
[core] Remove deleteOnExit method to avoid out of memory

leaves12138:
### Purpose

file.deleteOnExit method will stat in memory forever


JingsongLi:
+1

PR #4126, opened 2024-09-04
[core]Support ignoring specified fields while generating -U, +U changelog for the same record

chenxinwei:
Change-Id: Ida465e4902f13592df552be731da1530fdeaaeb0

### Purpose

Linked issue: close #4095

Support  specifying fields that are ignored for comparison while generating -U, +U changelog for the same record. I will use the changelog-producer.row-deduplicate-ignore-fields to decide which fields are ignored.This configuration is only valid for the changelog-producer.row-deduplicate is true.
### Tests

Add UT in LookupChangelogMergeFunctionWrapperTest.java

PR #4154, opened 2024-09-10
[core] Introduce projection fields to EqualiserCodeGenerator

JingsongLi:
### Purpose

See #4126

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/codegen/CodeGenerator.java
+++ b/paimon-common/src/main/java/org/apache/paimon/codegen/CodeGenerator.java
@@ -48,12 +48,7 @@ public interface CodeGenerator {
     GeneratedClass<RecordComparator> generateRecordComparator(
             List<DataType> inputTypes, int[] sortFields);
 
-    /**
-     * Generate a {@link RecordEqualiser}.
-     *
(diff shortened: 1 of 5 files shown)


paimon-common/src/main/java/org/apache/paimon/codegen/CodeGenerator.java
```diff
@@ -48,12 +48,7 @@ GeneratedClass<NormalizedKeyComputer> generateNormalizedKeyComputer(
...
 
-    /**
-     * Generate a {@link RecordEqualiser}.
-     *
-     * @param fieldTypes Both the input row field types and the sort key field types. Records are *
-     *     compared by the first field, then the second field, then the third field and so on. All *
-     *     fields are compared in ascending order.
-     */
-    GeneratedClass<RecordEqualiser> generateRecordEqualiser(List<DataType> fieldTypes);
+    /** Generate a {@link RecordEqualiser} with fields. */
```
chenxinwei, line 51: Whether it is necessary to add comments for params.

JingsongLi: I think fields (in equal context) is enough to explain.

chenxinwei:
LGTM +1

PR #4159, opened 2024-09-10
[core]delete created_from_snapshot and created_from_snapshot from branchTable

herefree:
### Purpose

`created_from_snapshot` should be the origin of the tag,not the earliest snapshot of the branch.

Linked issue: close #xxx


LinMingQiang:
It is possible that the branches tag does not exist but the snapshot exists?

herefree:
> It is possible that the branches tag does not exist but the snapshot exists?

Yes, if you create empty branch then insert some data to this branch, or you create branch from tag then delete branch's tag. In above two case snapshot will exists,tag will not exists. In this case created_from_tag will be null. 
I do not think we should remain created_from_snapshot,because we only support create_branch from tag. Snapshot will expire quickly, it's hard for us to know whether the earliest snapshot of the branch is from the main branch.

LinMingQiang:
`created_from_snapshot` can be the origin of the tag?

herefree:
> `created_from_snapshot` can be the origin of the tag?

That's okay, I'll modify it.

LinMingQiang:
With the current architecture, we cannot accurately obtain the origin snapshot of the branch. Maybe we should open a discussion thread to discuss this issue.

herefree:
> With the current architecture, we cannot accurately obtain the origin snapshot of the branch. Maybe we should open a discussion thread to discuss this issue.

I create a issue #4166

JingsongLi:
Can we just delete created_from_tag and created_from_snapshot first?

herefree:
> Can we just delete created_from_tag and created_from_snapshot first?

I think it's ok if we can't get the exact tag and snapshot.I'll change it.

JingsongLi (approved):
+1

PR #4165, opened 2024-09-10
[core] Add commit max retries if commit failed

askwang:
### Purpose

We write data with flink stream and run a dedicated compaction with flink. If the compaction commit message is too large and commit time is too long followed, commit will fail and enter an infinite loop.


JingsongLi (approved):
+1

PR #4171, opened 2024-09-11
[spark] Delete with deletion vectors should enable AQE

ulysses-you:
### Purpose

AQE won't break the required distribution and ordering, so it is unnecessary to disable aqe manually.

### Tests

Pass CI

### API and Format

no


ulysses-you:

YannByron:
Disabling AQE will have a better performance when performing delete operations with DV on large amounts of data,
but it can submit more tasks in other cases. Paimon can ignore this, and left to users to tune.

YannByron:
+1

JingsongLi:
+1

PR #4176, opened 2024-09-12
[core] Remove the deprecated core option `full`.

LinMingQiang:
### Purpose

For the new 1.0 version. Remove outdated parameters.


JingsongLi:
Hi @LinMingQiang , If it is not cumbersome to maintain, we don't need to delete expired options. We would only consider cleaning it up if it makes our code very difficult to understand.

LinMingQiang:
You're right, I just wish the code was cleaner in version 1.0, close this.

PR #4177, opened 2024-09-12
[hive] HiveMigrator support set parallelism for procedures

xuzifu666:
### Purpose

Linked issue: close #xxx
Currently migrate table/file/database cannot set thread parallelism, user can not overclock cpu to improve migrate speed or reduce cpu cost for avoid impacting other process in machine , this pr is support parallelism change.


JingsongLi:
Hi @xuzifu666 , parallelism should be a int.

xuzifu666:
> Hi @xuzifu666 , parallelism should be a int.

Thanks, had change parameter parallelism from String to Integer. @JingsongLi

JingsongLi (approved):
+1

PR #4178, opened 2024-09-12
[core] Optimize branches system table when specifying branch_name

herefree:
### Purpose
Optimize branches system table when specifying branch_name

Linked issue: close #xxx


xuzifu666:
Nice catch, but maybe need consider about branch usage in real production.

JingsongLi:
Hi @herefree thanks for the contribution.
But I can hardly imagine a situation with too many branches, so there's no need to optimize it here.

PR #4179, opened 2024-09-12
[avro] Unwrap AvroRuntimeException and throw the real IOException

tsreaper:
### Purpose

Many features, such as `OrphanFilesClean`, in Paimon checks for specific `IOException`s (for example, `FileNotFoundException` or `EOFException`) and deal with them properly. However avro readers currently wraps all `IOException`s into `AvroRuntimeException`, thus we cannot know what the real exception is.

This PR unwraps `AvroRuntimeException` and throws the real `IOException` in avro readers.

### Tests

Unit tests.

### API and Format

No format changes.

### Documentation

No new feature.


JingsongLi (approved):
+1

PR #4181, opened 2024-09-12
[flink] Introduce dedecate job for orphan files clean

JingsongLi:
### Purpose

For large tables, there are a lot of files, and the current single-machine program may lead to OOM (Out of Memory) issues. We should introduce distributed execution to avoid OOM and performance problems.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/LocalOrphanFilesClean.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/LocalOrphanFilesClean.java
@@ -1,0 +1,247 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.paimon.operation;
+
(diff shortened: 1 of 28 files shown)


tsreaper (approved):

paimon-core/src/main/java/org/apache/paimon/operation/OrphanFilesClean.java
```diff
@@ -403,15 +290,15 @@ private void addManifestList(List<String> used, Snapshot snapshot) {
      * {@link FileNotFoundException}, return null. Finally, if retry times reaches the limits,
```
yuzelin, line 290: `return null` -> `defaultValue`

paimon-core/src/main/java/org/apache/paimon/operation/LocalOrphanFilesClean.java
```diff
@@ -0,0 +1,247 @@
...
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+import static org.apache.paimon.utils.Preconditions.checkArgument;
+import static org.apache.paimon.utils.ThreadPoolUtils.createCachedThreadPool;
+import static org.apache.paimon.utils.ThreadPoolUtils.randomlyExecute;
+
+/** Flink {@link OrphanFilesClean}, it will use thread pool to execute deletion. */
```
yuzelin, line 57: Not `Flink`

paimon-core/src/main/java/org/apache/paimon/operation/LocalOrphanFilesClean.java
```diff
@@ -0,0 +1,247 @@
...
+            List<Path> nonSnapshotFiles = snapshotManager.tryGetNonSnapshotFiles(this::oldEnough);
+            nonSnapshotFiles.forEach(fileCleaner);
+            deleteFiles.addAll(nonSnapshotFiles);
+
+            // specially handle the changelog directory
+            List<Path> nonChangelogFiles = snapshotManager.tryGetNonChangelogFiles(this::oldEnough);
+            nonChangelogFiles.forEach(fileCleaner);
+            deleteFiles.addAll(nonChangelogFiles);
+
+            usedFiles.addAll(getUsedFiles(branch));
```
yuzelin, line 103: Why not use `cleanSnapshotDir`?

PR #4183, opened 2024-09-13
[common] Add exception handing in zstd compressor

Shadowell:
### Purpose
Add exception handing when source array is null and add exception info 

Linked issue: close #xxx

### Tests
Existed tests

### API and Format
No

### Documentation
No


JingsongLi:
+1

PR #4185, opened 2024-09-13
[flink] Remove Flink state from write-only writers

tsreaper:
### Purpose

Currently Paimon writer operators in Flink have states, which record the `commitUser` (for all writers) and the list of active buckets (for lookup or full-compaction changelog producer). These states prevent a writer object to be closed too early and cause conflicts.

However for write-only writers, because they only create new files, there is no conflict. So for these writers we can remove Flink state. This optimization is also useful for bounded stream jobs, because when chaining source and write operator together, if some source parallelism have finished, checkpoint cannot be performed if there are union list states in the operators which are still running.

### Tests

### API and Format

No format changes.

### Documentation

No new feature.

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/TableWriteOperator.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/TableWriteOperator.java
@@ -28,8 +28,11 @@ import org.apache.flink.runtime.io.disk.iomanager.IOManager;
 import org.apache.flink.runtime.state.StateInitializationContext;
 import org.apache.flink.runtime.state.StateSnapshotContext;
 
+import javax.annotation.Nullable;
+
 import java.io.IOException;
 import java.util.List;
+import java.util.UUID;
 
 /** An abstract class for table write operator. */
 public abstract class TableWriteOperator<IN> extends PrepareCommitOperator<IN, Committable> {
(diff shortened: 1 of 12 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/TableWriteOperator.java
```diff
@@ -52,37 +55,54 @@ public TableWriteOperator(
         this.initialCommitUser = initialCommitUser;
     }
 
+    private boolean needState() {
+        if (table.coreOptions().writeOnly()) {
+            // Commit user for writers are used to avoid conflicts.
```
wwj6591812, line 60: But users may use the commit user to determine if a table's snapshot is generated by the other job. See : https://github.com/apache/paimon/pull/3474

So, I think you can add a config for users to decide whether they need the commitUser state (default true). And explain in the doc that in bounded stream scenarios, if this configuration is true, there will be side effects(when chaining source and write operator together, if some source parallelism have finished, checkpoint cannot be performed).

tsreaper: This `commitUser` is only for writers. What you have mentioned is the `commitUser` for comitters.

wwj6591812: ok

wwj6591812:
+1

PR #4186, opened 2024-09-13
[core] Changelog expire should check index file existence

leaves12138:
### Purpose

Just two small bugs:

1 Check index file exists before read
2 Try more times for table consumer read (Rename may not be atomic)


JingsongLi (approved):
+1

PR #4158, opened 2024-09-10
[spark] Compaction add parallelize parallelism to avoid small partitions

askwang:
### Purpose

Linked issue:  https://github.com/apache/paimon/issues/4157

PR #4187, opened 2024-09-14
[spark] Fix compact unreasonable log info

askwang:
### Purpose

follow up https://github.com/apache/paimon/pull/4158


JingsongLi (approved):
+1

PR #4192, opened 2024-09-14
[core] disable streaming read of ReadOptimizedTable which has primary-key.

### LINKED ISSUE #4191: [Bug] disable streaming read of ReadOptimizedTable which has primary-key.
### Search before asking

### Paimon version

paimon-0.9 snapshot

### Compute Engine

flink-1.17.2

### Minimal reproduce step

Stream a `ReadOptimizedTable`, which contains the primary key.

`ReadOptimizedTable` with primary key should only provide batch read. In streaming mode, the semantics of incremental data are strange.  For example, if `changelog-producer` is set to `INPUT`, we will not be able to obtain incremental data because we only obtain the top-level files. At the same time, `SystemTable` should only provide `insert-only` data, which means that `ReadOptimizedTable` cannot provide retract messages.

### Anything else?

_No response_

liming30:
### Purpose

Linked issue: close #4191

[core] disable streaming read of ReadOptimizedTable which has primary-key.
### Tests

1. org.apache.paimon.table.PrimaryKeyFileStoreTableTest#testStreamingReadOptimizedTable

### API and Format

No
### Documentation

No

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/system/ReadOptimizedTable.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/system/ReadOptimizedTable.java
@@ -137,6 +137,10 @@ public class ReadOptimizedTable implements DataTable, ReadonlyTable {
 
     @Override
     public StreamDataTableScan newStreamScan() {
+        if (wrapped.schema().primaryKeys().size() > 0) {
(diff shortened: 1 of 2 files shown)


JingsongLi (approved):
+1

Issue #4184, opened 2024-09-13
[Feature] Implement distributed orphan file clean for spark


JingsongLi:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Motivation

For large tables, there are a lot of files, and the current single-machine program may lead to OOM (Out of Memory) issues. We should introduce distributed execution to avoid OOM and performance problems.

In https://github.com/apache/paimon/pull/4181 , we have a distributed orphan file clean for Flink.

We should modify Spark too.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

xuzifu666:
Had submit a pr to support it. https://github.com/apache/paimon/pull/4196

xuzifu666:
Refer Flink logic, support spark distributed for it. https://github.com/apache/paimon/pull/4199

ulysses-you:
pr: https://github.com/apache/paimon/pull/4207

PR #4200, opened 2024-09-17
[spark] Support distributed orphan file clean for spark

xuzifu666:
### Purpose

Linked issue: close https://github.com/apache/paimon/issues/4184

### Tests

RemoveOrphanFilesProcedureTest

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/orphan/SparkOrphanFilesClean.java
+++ b/paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/orphan/SparkOrphanFilesClean.java
@@ -1,0 +1,243 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.paimon.spark.orphan;
+
+import org.apache.paimon.Snapshot;
+import org.apache.paimon.catalog.Catalog;
+import org.apache.paimon.catalog.Identifier;
+import org.apache.paimon.fs.FileStatus;
+import org.apache.paimon.fs.Path;
+import org.apache.paimon.manifest.ManifestEntry;
+import org.apache.paimon.manifest.ManifestFile;
+import org.apache.paimon.manifest.ManifestFileMeta;
+import org.apache.paimon.operation.OrphanFilesClean;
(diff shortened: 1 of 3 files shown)


paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/orphan/SparkOrphanFilesClean.java
```diff
@@ -0,0 +1,250 @@
...
+
+        List<String> branches = validBranches();
+
+        // snapshot and changelog files are the root of everything, so they are handled specially
+        // here, and subsequently, we will not count their orphan files.
+        AtomicLong deletedInLocal = new AtomicLong(0);
+        cleanSnapshotDir(branches, p -> deletedInLocal.incrementAndGet());
+
+        // branch and manifest file
+        CollectionAccumulator<Pair<String, String>> pairsAccumulator =
```
JingsongLi, line 123: Why we need a `CollectionAccumulator`? Why not just use RDD?

xuzifu666: Thanks @JingsongLi       I consider 2 points:
1. Want to reuse OrphanFilesClean##collectWithoutDataFile method and Spark API seems not contain side output stream like Flink  
2. If use rdd maybe need split another rdd. CollectionAccumulator work as side output.

xuzifu666: Here CollectionAccumulator change to Rdd is better, had changed it and do some change in parent method for more common calling. @JingsongLi

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/orphan/SparkOrphanFilesClean.java
```diff
@@ -0,0 +1,260 @@
...
+                                String branch = (String) pair.getLeft();
+
+                                Snapshot snapshot = Snapshot.fromJson((String) pair.getRight());
+                                Consumer<ManifestFileMeta> manifestConsumer =
+                                        manifest -> {
+                                            Pair<String, String> pair0 =
+                                                    Pair.of(branch, manifest.fileName());
+                                            collect.add(pair0);
+                                        };
+                                collectWithoutDataFile(branch, snapshot, null, manifestConsumer);
```
JingsongLi, line 153: It is better to execute this once.

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/orphan/SparkOrphanFilesClean.java
```diff
@@ -0,0 +1,260 @@
...
+                    "Only FileStoreTable supports remove-orphan-files action. The table type is '%s'.",
+                    table.getClass().getName());
+
+            Long cleanNums =
+                    new SparkOrphanFilesClean(
+                                    (FileStoreTable) table,
+                                    olderThanMillis,
+                                    fileCleaner,
+                                    parallelism)
+                            .doOrphanClean(ctx);
```
JingsongLi, line 102: It is better to return a RDD or Dataset, in that way, all tables can be executed together to allow for greater utilization of all resources.

xuzifu666: OK，would changed it latter.

xuzifu666:
https://github.com/apache/paimon/pull/4207 dataset code style is more graceful than rdd style, so want to use 4207 version to support the issue @ulysses-you @JingsongLi

PR #4203, opened 2024-09-18
[cdc] Unaware bucket cdc sink should not chain

JingsongLi:
### Purpose

rebalance it to make sure schema change work to avoid infinite loop.

### Tests

No test, it is hard to reproduce in the unit test.


leaves12138 (approved):
+1

PR #4204, opened 2024-09-18
[core] System files table distributed execute remold

leaves12138:
### Purpose

Flink search files table takes too long, files collect by single executor may be out-of-memory too. 
It's better to execute distributedly.


paimon-core/src/main/java/org/apache/paimon/table/system/FilesTable.java
```diff
@@ -181,14 +204,17 @@ private static class FilesSplit extends SingletonSplit {
         @Nullable private final LeafPredicate partitionPredicate;
```
JingsongLi, line 204: You should merge partitionPredicate and partition.

paimon-core/src/main/java/org/apache/paimon/table/system/FilesTable.java
```diff
@@ -170,23 +179,56 @@ public InnerTableScan withFilter(Predicate pushdown) {
...
     private static class FilesSplit extends SingletonSplit {
 
-        @Nullable private final LeafPredicate partitionPredicate;
+        @Nullable private final BinaryRow partition;
         @Nullable private final LeafPredicate bucketPredicate;
         @Nullable private final LeafPredicate levelPredicate;
 
         private FilesSplit(
-                @Nullable LeafPredicate partitionPredicate,
+                @Nullable BinaryRow partition,
```
JingsongLi, line 228: Use `@Nullable Map<String, String> partSpec`. You can get this from `InternalRowPartitionComputer`?

JingsongLi:
Use:
```
SnapshotReader snapshotReader = storeTable.newSnapshotReader();
if (partitionPredicate != null) {
    List<BinaryRow> partitions = snapshotReader.withPartitionFilter(partitionPredicate).partitions();
}
```

JingsongLi:
+1

PR #4206, opened 2024-09-18
[core] Optimize IncrementalStartingScanner to plan with thread pool

JingsongLi:
### Purpose

When there are 100+ snapshots, `IncrementalStartingScanner` will be very slow, we can optimize it to execute in thread pool.

In this PR, introduced `ManifestsReader` for `SnapshotReader` to read manifest files, and expose `SnapshotReader.readManifest` to public. Outside can use these two interfaces to muti-threads execution.


leaves12138 (approved):
+1

PR #4208, opened 2024-09-18
[core] Support async refresh for PrimaryKeyPartialLookupTable

### LINKED ISSUE #3386: [Feature] Support async refresh for PrimaryKeyPartialLookupTable
### Search before asking

### Motivation

Async refresh for FullCacheLookupTable is supported by https://github.com/apache/paimon/issues/3295. We should also support async refresh for PrimaryKeyPartialLookupTable.

### Solution

_No response_

### Anything else?

_No response_

xiangyuf:
### Purpose

Linked issue: close #3386

Support async refresh for PrimaryKeyPartialLookupTable. Async refresh files generated by new snapshots which are intervened with expired cached files. The total size of new generated lookup during one refresh will not greatly exceed the total size of expired lookup files.
### Tests

1. Added UT LookupTableTest#testAsyncRefreshPartialLookupTable 
2. Added UT LookupLevelsTest#testRefreshFiles
3. Added IT FileStoreLookupFunctionTest#testAsyncRefreshLookupTableNoError

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/AsyncRefreshLookupTable.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/AsyncRefreshLookupTable.java
@@ -1,0 +1,141 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.paimon.flink.lookup;
+
+import org.apache.paimon.annotation.VisibleForTesting;
+import org.apache.paimon.options.Options;
+import org.apache.paimon.table.FileStoreTable;
+import org.apache.paimon.utils.ExecutorThreadFactory;
+import org.apache.paimon.utils.ExecutorUtils;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nullable;
+
+import java.io.IOException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+
+import static org.apache.paimon.flink.FlinkConnectorOptions.LOOKUP_REFRESH_ASYNC;
+import static org.apache.paimon.flink.FlinkConnectorOptions.LOOKUP_REFRESH_ASYNC_PENDING_SNAPSHOT_COUNT;
+
+/** A {@link LookupTable} supports async refresh. */
+public abstract class AsyncRefreshLookupTable implements LookupTable {
+    private static final Logger LOG = LoggerFactory.getLogger(AsyncRefreshLookupTable.class);
+    private final FileStoreTable table;
+
+    private final int maxPendingSnapshotCount;
+
+    @Nullable private ExecutorService refreshExecutor;
+
+    private final AtomicReference<Exception> cachedException;
+
+    private Future<?> refreshFuture;
+
+    protected final boolean refreshAsync;
+
+    public AsyncRefreshLookupTable(FileStoreTable table) {
+        Options options = Options.fromMap(table.options());
+        this.table = table;
+        this.refreshAsync = options.get(LOOKUP_REFRESH_ASYNC);
+        this.cachedException = new AtomicReference<>();
+        this.maxPendingSnapshotCount = options.get(LOOKUP_REFRESH_ASYNC_PENDING_SNAPSHOT_COUNT);
+    }
+
+    protected void init() throws Exception {
+        this.refreshExecutor =
+                this.refreshAsync
+                        ? Executors.newSingleThreadExecutor(
+                                new ExecutorThreadFactory(
+                                        String.format(
+                                                "%s-lookup-refresh",
+                                                Thread.currentThread().getName())))
+                        : null;
+    }
+
+    @Override
+    public void refresh() throws Exception {
+        if (refreshExecutor == null) {
+            doRefresh(false);
+            return;
+        }
+
+        Long latestSnapshotId = table.snapshotManager().latestSnapshotId();
+        Long nextSnapshotId = nextSnapshotId();
+        if (latestSnapshotId != null
+                && nextSnapshotId != null
+                && latestSnapshotId - nextSnapshotId > maxPendingSnapshotCount) {
+            LOG.warn(
+                    "The latest snapshot id {} is much greater than the next snapshot id {} for {}}, "
+                            + "you may need to increase the parallelism of lookup operator.",
+                    latestSnapshotId,
+                    nextSnapshotId,
+                    maxPendingSnapshotCount);
+            if (refreshFuture != null) {
+                // Wait the previous refresh task to be finished.
+                refreshFuture.get();
+            }
+            doRefresh(false);
+        } else {
+            refreshAsync();
+        }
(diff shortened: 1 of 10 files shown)


xiangyuf:
@FangYongs @JingsongLi  Would you kindly review this when you have time?

xiangyuf:
+ @Aitozi

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/FullCacheLookupTable.java
```diff
@@ -87,6 +84,7 @@ public abstract class FullCacheLookupTable implements LookupTable {
     private Predicate specificPartition;
 
     public FullCacheLookupTable(Context context) {
+        super(context.table);
```
Aitozi, line 87: Some fields of this class  such as `refreshExecutor`, `maxPendingSnapshotCount`.. are extracted to the parent class, so can be removed now.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/FullCacheLookupTable.java
```diff
@@ -249,17 +209,7 @@ private void doRefresh() throws Exception {
...
     public final List<InternalRow> get(InternalRow key) throws IOException {
-        List<InternalRow> values;
-        if (refreshAsync) {
-            synchronized (lock) {
-                values = innerGet(key);
-            }
-        } else {
-            values = innerGet(key);
-        }
-        if (appendUdsFieldNumber == 0) {
```
Aitozi, line 260: this line should be kept ?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/AsyncRefreshLookupTable.java
```diff
@@ -0,0 +1,149 @@
...
+        this.lock = this.refreshAsync ? new Object() : null;
+        this.cachedException = new AtomicReference<>();
+        this.maxPendingSnapshotCount = options.get(LOOKUP_REFRESH_ASYNC_PENDING_SNAPSHOT_COUNT);
+    }
+
+    @Override
+    public List<InternalRow> get(InternalRow key) throws IOException {
+        List<InternalRow> values;
+        if (refreshAsync) {
+            synchronized (lock) {
```
Aitozi, line 81: I'm curious about what this lock to protect, Is this to avoid the race condition between `get` and `refresh` ? Or the parallelism `get` ?  
I think we should  ensure the `QueryExecutor` is `ThreadSafe` Or guarded by lock.

JingsongLi:
CC @liming30 too

liming30:
The purpose of introducing asynchronous refresh in #3297 is to solve the problem that local data refresh will block `lookup`. Is this issue also trying to solve the same problem?  `PrimaryKeyPartialLookupTable.refresh` only refreshes metadata, and it seems that it cannot solve the problem of synchronous refresh of local data.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/AsyncRefreshLookupTable.java
xiangyuf: @Aitozi this is first introduced in `FullCacheLookupTable` scenario. Explained by @FangYongs, 'Currently there will be primaryKeyRow, RocksDBSetState and RocksDBListState in PrimaryKeyLookupTable, SecondaryIndexLookupTable and NoPrimaryKeyLookupTable, they are not thread safe. I think we can keep the lock in get method and optimize it when we find that it has become a bottleneck.' See more in https://github.com/apache/paimon/pull/3297

xiangyuf:
> The purpose of introducing asynchronous refresh in #3297 is to solve the problem that local data refresh will block `lookup`. Is this issue also trying to solve the same problem? `PrimaryKeyPartialLookupTable.refresh` only refreshes metadata, and it seems that it cannot solve the problem of synchronous refresh of local data.

@liming30 Thx for reply. I see your point. Maybe I should split this task into two tasks: 1, Make the refreshAsync options also work for PrimaryKeyPartialLookupTable; 2, Improve the refresh behavior for LocalQueryExecutor, create local LookupFile during refresh according to the before and after DataFileMetas in a DataSplit. WDYT?

xiangyuf:
Hi @liming30 @Aitozi, after second thought I think `AsyncRefreshLookupTable` should not handle the consistence issue between `asyncRefresh` operation and `get` operation for the time being. This can be resolved inside `PrimaryKeyPartialLookupTable` and `FullCacheLookupTable`. WDYT?

JingsongLi:
I overall think that if there is a solution to the data file loading issue, then it would be meaningful.

But the current code changes are only asynchronous metadata loading, which doesn't seem to have much value.

xiangyuf:
> I overall think that if there is a solution to the data file loading issue, then it would be meaningful.
> 
> But the current code changes are only asynchronous metadata loading, which doesn't seem to have much value.

Yes, I can also create new lookup files asynchronously if they are overlapped with previous evicted files.

xiangyuf:
cc: @xiangyuf @Aitozi @yunfengzhou-hub for review

xiangyuf:
Before async refresh feature enabled:
image

After async refresh feature enabled:
image

xiangyuf:
This picture is more clear to show the benefits of async refresh for lookup file.
image

xiangyuf:
Reproduce procedure:

DDL:
`
use catalog paimon;
use paimon_test;

CREATE  TABLE dim_orders (
    `order_id`          INT,
    `order_name`        STRING,
    `order_product_id`  INT,
    `order_customer_id` INT,
    `order_status`      STRING,
    `create_date`       TIMESTAMP,
    `create_ts`         INT,
    PRIMARY             key (`order_id`) NOT ENFORCED
)
WITH (
    'bucket' = '20',
    'bucket-key' = 'order_id'
);
`

Dim update job:

`
CREATE  TABLE datagen_source (
            `order_id`          INT,
            `order_name`        STRING,
            `order_product_id`  INT,
            `order_customer_id` INT,
            `order_status`      STRING
        )
        WITH (
            'connector' = 'datagen',
            'rows-per-second' = '50000',
            'fields.order_name.length' = '10',
            'fields.order_status.length' = '5',
            'fields.order_id.min' = '1',
            'fields.order_id.max' = '100000000',
            'fields.order_product_id.min' = '1',
            'fields.order_product_id.max' = '1000000',
            'fields.order_customer_id.min' = '1',
            'fields.order_customer_id.max' = '1000000'
        );

CREATE  VIEW temp_source AS
SELECT  order_id,
        order_name,
        order_product_id,
        order_customer_id,
        order_status,
        CURRENT_TIMESTAMP AS create_date,
        CAST(UNIX_TIMESTAMP() AS INT) AS create_ts
FROM    datagen_source;

INSERT INTO `paimon`.`paimon_test`.`dim_orders`
SELECT  *
FROM    temp_source;
`

Dim join job:
`
CREATE  TABLE `datagen_source` (
            `product_id`          INTEGER,
            `product_name`        STRING,
            `product_category_id` INTEGER,
            `product_order_id`    INTEGER,
            `product_status`      STRING,
            `create_date`         AS `proctime`()
        )
        WITH (
            'fields.product_category_id.min' = '1',
            'connector' = 'datagen',
            'fields.product_category_id.max' = '1000000',
            'fields.product_name.length' = '10',
            'fields.product_status.length' = '5',
            'rows-per-second' = '48000',
            'fields.product_order_id.max' = '200000000',
            'fields.product_order_id.min' = '1'
        );

CREATE  TABLE `print_sink` (
            `product_id`          INTEGER,
            `product_name`        STRING,
            `product_category_id` INTEGER,
            `product_order_id`    INTEGER,
            `product_status`      STRING,
            `create_date`         TIMESTAMP,
            `order_name`          STRING,
            `orders_customer_id`  INTEGER,
            `order_status`        STRING
        )
        WITH ('connector' = 'print');

CREATE  VIEW `trade_orde_view` AS
SELECT  `gen`.`product_id`,
        `gen`.`product_name`,
        `gen`.`product_category_id`,
        `gen`.`product_order_id`,
        `gen`.`product_status`,
        `gen`.`create_date`,
        `orders`.`order_name`,
        `orders`.`order_customer_id`,
        `order_status`
FROM    `datagen_source` AS `gen`
INNER JOIN
        `paimon`.`paimon_test`.`dim_orders`
        /*+ `OPTIONS`('lookup.cache' = 'AUTO', 'lookup.refresh.async' = 'true', 'lookup.cache-max-disk-size' = '40 gb') */
        FOR SYSTEM_TIME AS OF `gen`.`create_date` AS `orders`
ON      `gen`.`product_order_id` = `orders`.`order_id`;

INSERT INTO `print_sink` (
    SELECT  *
    FROM    `trade_orde_view`
    WHERE   `order_name` IS NOT NULL
);
`

xiangyuf:
cc: @JingsongLi @FangYongs  for review.

PR #4210, opened 2024-09-18
[test] Add test case for format read nested type with purning

Zouxxyy:
### Purpose

Add test case for format read nested type with purning


JingsongLi (approved):
+1

PR #4211, opened 2024-09-19
[hotfix] fix unspecified value parameter when using PaimonMetadataColumn.get method

zhuangchong:
### Purpose

 fix unspecified value parameter when using PaimonMetadataColumn.get method

image

https://github.com/apache/paimon/actions/runs/10933145723/job/30351130554#step:4:4636


JingsongLi:
+1

PR #4213, opened 2024-09-19
[core] Add Table.uuid method

JingsongLi:
### Purpose

Add Table.uuid method for table uniq id, this can be used like some RBAC support.
- for filesystem catalog, uuid is db + tableName + schema-createTime.
- for hive catalog, uuid is db + tableName + createTime.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/AbstractFileStoreTable.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/AbstractFileStoreTable.java
@@ -168,6 +168,14 @@ abstract class AbstractFileStoreTable implements FileStoreTable {
     }
 
     @Override
+    public String uuid() {
+        if (catalogEnvironment.uuid() != null) {
+            return catalogEnvironment.uuid();
+        }
+        return fullName();
+    }
+
+    @Override
     public Optional<Statistics> statistics() {
         Snapshot snapshot = TimeTravelUtil.resolveSnapshot(this);
         if (snapshot != null) {
--- a/paimon-core/src/main/java/org/apache/paimon/table/AbstractFileStoreTable.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/AbstractFileStoreTable.java
@@ -172,7 +172,8 @@ abstract class AbstractFileStoreTable implements FileStoreTable {
         if (catalogEnvironment.uuid() != null) {
             return catalogEnvironment.uuid();
         }
-        return fullName();
+        long earliestCreationTime = schemaManager().earliestCreationTime();
+        return fullName() + "." + earliestCreationTime;
     }
 
     @Override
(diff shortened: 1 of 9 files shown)


paimon-core/src/main/java/org/apache/paimon/table/AbstractFileStoreTable.java
```diff
@@ -164,6 +164,14 @@ public Identifier identifier() {
                 : identifier;
     }
 
+    @Override
+    public String uuid() {
+        if (catalogEnvironment.uuid() != null) {
+            return catalogEnvironment.uuid();
+        }
+        return fullName();
```
adrian-wang, line 172: should at least log warning or add a setting to enabling simply use `fullName()` as `uuid()`

paimon-core/src/main/java/org/apache/paimon/table/AbstractFileStoreTable.java
```diff
@@ -167,6 +167,15 @@ public Identifier identifier() {
                 : identifier;
     }
 
+    @Override
+    public String uuid() {
+        if (catalogEnvironment.uuid() != null) {
+            return catalogEnvironment.uuid();
+        }
+        long earliestCreationTime = schemaManager().earliestCreationTime();
```
adrian-wang, line 175: will this earliestCreationTime change as we do compaction or snapshot expire?

paimon-core/src/test/java/org/apache/paimon/catalog/CatalogTestBase.java
```diff
@@ -948,4 +948,16 @@ public void testFormatTable() throws Exception {
...
     }
+
+    @Test
+    public void testTableUUID() throws Exception {
+        catalog.createDatabase("test_db", false);
+        Identifier identifier = Identifier.create("test_db", "test_table");
+        catalog.createTable(identifier, DEFAULT_TABLE_SCHEMA, false);
+        Table table = catalog.getTable(identifier);
+        String uuid = table.uuid();
+        assertThat(uuid).startsWith(identifier.getFullName() + ".");
```
adrian-wang, line 959: This will fail those non-FileStoreTable, right?

adrian-wang (commented):
Only 2 small questions

paimon-core/src/main/java/org/apache/paimon/table/AbstractFileStoreTable.java
JingsongLi: Never, actually we do not clean schema files, so the first schema file (schema-0) will always exist, and immutable.

So this earliestCreationTime should always be the same.

paimon-core/src/test/java/org/apache/paimon/catalog/CatalogTestBase.java
JingsongLi: Yes

adrian-wang:
LGTM, +1, let's merge this!

PR #4215, opened 2024-09-19
[core] Improve error message for concurrent snapshot expiration

yunfengzhou-hub:
### Purpose

This PR adds a more detailed error message to the snapshot expiration process, showing users a possible cause of the FileNotFoundException they may face and the recommended solution to it.

### Tests

A UT has been added to verify the error message in case of concurrent snapshot expiration.

### API and Format

This PR does not affect API or format.

### Documentation

This change does not need documentation

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/Snapshot.java
+++ b/paimon-core/src/main/java/org/apache/paimon/Snapshot.java
@@ -29,8 +29,12 @@ import org.apache.paimon.shade.jackson2.com.fasterxml.jackson.annotation.JsonIgn
 import org.apache.paimon.shade.jackson2.com.fasterxml.jackson.annotation.JsonInclude;
 import org.apache.paimon.shade.jackson2.com.fasterxml.jackson.annotation.JsonProperty;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import javax.annotation.Nullable;
 
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.util.Map;
 import java.util.Objects;
@@ -61,6 +65,7 @@ import java.util.Objects;
 @Public
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class Snapshot {
+    private static final Logger LOG = LoggerFactory.getLogger(Snapshot.class);
 
     public static final long FIRST_SNAPSHOT_ID = 1;
 
@@ -357,6 +362,16 @@ public class Snapshot {
     public static Snapshot fromPath(FileIO fileIO, Path path) {
         try {
             return Snapshot.fromJson(fileIO.readFileUtf8(path));
+        } catch (FileNotFoundException e) {
+            String errorMessage =
+                    String.format(
+                            "Snapshot file %s does not exist. "
+                                    + "It might have been expired by other jobs operating on this table. "
+                                    + "In this case, you can avoid concurrent modification issues by configuring "
+                                    + "write-only = true and use a dedicated compaction job, or configuring "
+                                    + "different expiration thresholds for different jobs.",
+                            path);
+            throw new RuntimeException(errorMessage, e);
         } catch (IOException e) {
             throw new RuntimeException("Fails to read snapshot from path " + path, e);
         }
--- a/paimon-core/src/test/java/org/apache/paimon/utils/SnapshotManagerTest.java
+++ b/paimon-core/src/test/java/org/apache/paimon/utils/SnapshotManagerTest.java
@@ -369,7 +369,9 @@ public class SnapshotManagerTest {
         localFileIO.deleteQuietly(snapshotManager.snapshotPath(3));
         thread.join();
 
-        assertThat(exception.get()).hasMessageContaining("Fails to read snapshot from path");
+        assertThat(exception.get())
+                .hasMessageFindingMatch("Snapshot file .* does not exist")
+                .hasMessageContaining("dedicated compaction job");
     }
 
     @Test


yunfengzhou-hub:
Hi @tsreaper could you please take a look at this PR?

paimon-core/src/main/java/org/apache/paimon/table/ExpireSnapshotsImpl.java
```diff
@@ -219,6 +219,25 @@ public int expireUntil(long earliestId, long endExclusiveId) {
         return (int) (endExclusiveId - beginInclusiveId);
     }
 
+    private Snapshot getSnapshot(long snapshotId) {
```
JingsongLi, line 222: Just modify `SnapshotManager.snapshot`?

paimon-core/src/main/java/org/apache/paimon/table/ExpireSnapshotsImpl.java
```diff
@@ -240,4 +259,14 @@ private void writeEarliestHint(long earliest) {
...
         return snapshotDeletion;
     }
+
+    @VisibleForTesting
+    public SnapshotManager snapshotManager() {
+        return snapshotManager;
+    }
+
+    @VisibleForTesting
+    public void setSnapshotManager(SnapshotManager manager) {
```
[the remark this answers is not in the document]
yunfengzhou-hub: If snapshot file does not exist from the beginning, its deletion will be skipped and no exception would be thrown. The exception only occurs when ExpireSnapshotImpl found the snapshot file at first, and then failed to read the file.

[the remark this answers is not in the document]
yunfengzhou-hub: I overrode the content of the snapshot file, mocking the situation that ExpireSnapshotImpl can find the snapshot file, but later it failed to read the content of the snapshot file. It also worked in the test case without exposing `setSnapshotManager` in ExpireSnapshotImpl.

paimon-core/src/main/java/org/apache/paimon/utils/SnapshotManager.java
```diff
@@ -118,8 +118,20 @@ public Path snapshotDirectory() {
...
+        } catch (Exception e) {
+            String errorMessage =
+                    String.format(
+                            "Failed to read snapshot file for snapshot %s. "
+                                    + "It might have been expired by other jobs operating on this table. "
+                                    + "In this case, you can avoid concurrent modification issues by configuring "
+                                    + "write-only = true and use a dedicated compaction job, or configuring "
+                                    + "different expiration thresholds for different jobs.",
+                            snapshotId);
+            throw new IllegalStateException(errorMessage, e);
```
JingsongLi, line 133: This is a bad idea, we should not wrap exception here. FileNotException will be hided.

You can still modify expire class.

yunfengzhou-hub: The FileNotFoundException has already been wrapped with RuntimeException in Snapshot#fromPath. I'll modify this method to avoid wrapping FileNotFoundException twice.

paimon-core/src/main/java/org/apache/paimon/Snapshot.java
```diff
@@ -357,6 +362,16 @@ public static Snapshot fromJson(String json) {
     public static Snapshot fromPath(FileIO fileIO, Path path) {
         try {
             return Snapshot.fromJson(fileIO.readFileUtf8(path));
```
wwj6591812, line 364: Here my throw UncheckedIOException, could you add same logic to handle the UncheckedIOException.

paimon-core/src/main/java/org/apache/paimon/Snapshot.java
```diff
@@ -357,6 +362,16 @@ public static Snapshot fromJson(String json) {
...
+        } catch (FileNotFoundException e) {
+            String errorMessage =
+                    String.format(
+                            "Snapshot file %s does not exist. "
+                                    + "It might have been expired by other jobs operating on this table. "
```
wwj6591812, line 369: "by other jobs operating on this table." -> "by other write jobs on this table." ?

JingsongLi (approved):
+1

paimon-core/src/main/java/org/apache/paimon/Snapshot.java
JingsongLi: Where to throw `UncheckedIOException`?

paimon-core/src/main/java/org/apache/paimon/Snapshot.java
JingsongLi: `write` may be not so accurate, maybe just expire procedure.

paimon-core/src/main/java/org/apache/paimon/Snapshot.java
wwj6591812: In Snapshot#fromPath will call Snapshot#fromJson, and then call JsonSerdeUtil.fromJson.
In JsonSerdeUtil.fromJson, IOException will transform to UncheckedIOException which is not IOException.

paimon-core/src/main/java/org/apache/paimon/Snapshot.java
wwj6591812: OK

wwj6591812:
+1

paimon-core/src/main/java/org/apache/paimon/Snapshot.java
JingsongLi: We don't need to take care it, it is true exception, just throw it.

PR #4219, opened 2024-09-19
[flink] Remove state from append only unaware bucket writer

tsreaper:
### Purpose

Currently Paimon writer operators in Flink have states, which record the `commitUser` (for all writers) and the list of active buckets (for lookup or full-compaction changelog producer). These states prevent a writer object to be closed too early and cause conflicts.

However for append only unaware bucket writers, because they only create new files, there is no conflict. So for these writers we can remove Flink state. This optimization is also useful for bounded stream jobs, because when chaining source and write operator together, if some source parallelism have finished, checkpoint cannot be performed if there are union list states in the operators which are still running.

### Tests

IT case.

### API and Format

No format changes.

### Documentation

No new feature.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreWrite.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreWrite.java
@@ -214,7 +214,6 @@ public abstract class AbstractFileStoreWrite<T> implements FileStoreWrite<T> {
                 result.add(committable);
 
                 if (committable.isEmpty()) {
-
                     if (writerCleanChecker.apply(writerContainer)) {
                         // Clear writer if no update, and if its latest modification has committed.
                         //
@@ -246,6 +245,10 @@ public abstract class AbstractFileStoreWrite<T> implements FileStoreWrite<T> {
         return result;
     }
 
+    // This abstract function returns a whole function (instead of just a boolean value),
+    // because we do not want to introduce `commitUser` into this base class.
+    //
+    // For writers with no conflicts, `commitUser` might be some random value.
     protected abstract Function<WriterContainer<T>, Boolean> createWriterCleanChecker();
 
     protected static <T>
--- a/paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreWrite.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreWrite.java
@@ -52,6 +52,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.function.Function;
 
 import static org.apache.paimon.io.DataFileMeta.getMaxSequenceNumber;
 
@@ -64,7 +65,6 @@ public abstract class AbstractFileStoreWrite<T> implements FileStoreWrite<T> {
 
     private static final Logger LOG = LoggerFactory.getLogger(AbstractFileStoreWrite.class);
 
-    private final String commitUser;
     protected final SnapshotManager snapshotManager;
     private final FileStoreScan scan;
     private final int writerNumberMax;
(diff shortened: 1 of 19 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/NoopStoreSinkWriteState.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    @Override
+    public StateValueFilter stateValueFilter() {
+        return stateValueFilter;
+    }
+
+    @Override
+    public @Nullable List<StateValue> get(String tableName, String key) {
+        return Collections.emptyList();
```
JingsongLi, line 45: Should we throw exception for get and put?

paimon-core/src/main/java/org/apache/paimon/operation/AppendOnlyFileStoreWrite.java
```diff
@@ -143,50 +109,38 @@ protected RecordWriter<InternalRow> createWriter(
...
                 fileIndexOptions,
                 options.asyncFileWrite());
     }
 
+    protected abstract CompactManager getCompactManager(
+            BinaryRow partition,
+            int bucket,
+            List<DataFileMeta> restoredFiles,
+            ExecutorService compactExecutor,
+            @javax.annotation.Nullable DeletionVectorsMaintainer dvMaintainer);
```
JingsongLi, line 142: import this

paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreWrite.java
```diff
@@ -266,6 +246,37 @@ public List<CommitMessage> prepareCommit(boolean waitCompaction, long commitIden
...
+        //
+        // Condition 2: No compaction is in progress. That is, no more changelog will be
+        // produced.
+        return writerContainer ->
+                writerContainer.lastModifiedCommitIdentifier < latestCommittedIdentifier
+                        && !writerContainer.writer.isCompacting();
+    }
+
+    protected static <T>
+            Function<WriterContainer<T>, Boolean> createNoConflictAwareWriterCleanChecker() {
```
JingsongLi, line 276: Just inline this method in `AppendOnlyUnawareBucketFileStoreWrite`.

paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreWrite.java
```diff
@@ -266,6 +246,37 @@ public List<CommitMessage> prepareCommit(boolean waitCompaction, long commitIden
         return result;
     }
 
+    protected abstract Function<WriterContainer<T>, Boolean> createWriterCleanChecker();
```
JingsongLi, line 249: Should we do this abstraction? Or just introducing a `boolean doWriterCleanChecker`?

tsreaper: This abstract function returns a whole function (instead of just a boolean value), because we do not want to introduce `commitUser` into this base class.

For writers with no conflicts, `commitUser` might be some random value.

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/UnawareBucketAppendOnlyTableITCase.java
```diff
@@ -346,6 +357,95 @@ public void testFileIndex() {
...
+
+        StreamTableEnvironment tEnv = StreamTableEnvironment.create(env);
+        tEnv.registerCatalog("mycat", sEnv.getCatalog("PAIMON").get());
+        tEnv.executeSql("USE CATALOG mycat");
+        tEnv.createTemporaryView("S", tEnv.fromDataStream(source).as("id"));
+
+        tEnv.executeSql("INSERT INTO append_table SELECT id, 'test' FROM S").await();
+        assertThat(batchSql("SELECT * FROM append_table"))
+                .containsExactlyInAnyOrder(Row.of(1, "test"), Row.of(2, "test"));
+        System.out.println(table.snapshotManager().latestSnapshotId());
```
wwj6591812, line 384: cut System.out.println.

wwj6591812:
+1

JingsongLi (approved):
+1

PR #4220, opened 2024-09-19
[cdc] Avoid duplicate schema file generation

codeTai:
### Purpose

When the Flink CDC job is restarted and the table options have not changed, avoid regenerating the schema file

### Tests

### API and Format

No format changes.

### Documentation

No new feature.


JingsongLi (approved):
+1

PR #4221, opened 2024-09-19
[flink] Set default parallelism for sink in cases where AQE not supported

yunfengzhou-hub:
### Purpose

This PR provides default parallelism configuration for sink operators, in cases where undefined/adaptive parallelism has not been supported.

### Tests

Existing UTs are modified to verify that the cases mentioned above can function correctly now.

### API and Format

No API or format is changed

### Documentation

No documentation needed. The existing unsupported cases had not been documented, so no modification needed for documents.

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkSinkBuilder.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkSinkBuilder.java
@@ -44,12 +44,14 @@ import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nullable;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 
+import static org.apache.flink.configuration.CoreOptions.DEFAULT_PARALLELISM;
 import static org.apache.paimon.flink.FlinkConnectorOptions.CLUSTERING_SAMPLE_FACTOR;
 import static org.apache.paimon.flink.FlinkConnectorOptions.CLUSTERING_STRATEGY;
 import static org.apache.paimon.flink.FlinkConnectorOptions.MIN_CLUSTERING_SAMPLE_FACTOR;
@@ -75,7 +77,7 @@ public class FlinkSinkBuilder {
 
     private DataStream<RowData> input;
     @Nullable protected Map<String, String> overwritePartition;
-    @Nullable protected Integer parallelism;
+    @Nullable private Integer parallelism;
     @Nullable private TableSortInfo tableSortInfo;
 
     // ============== for extension ==============
@@ -208,6 +210,7 @@ public class FlinkSinkBuilder {
 
     /** Build {@link DataStreamSink}. */
     public DataStreamSink<?> build() {
+        parallelism = checkAndUpdateParallelism(input, parallelism);
         input = trySortInput(input);
         DataStream<InternalRow> input = mapToInternalRow(this.input, table.rowType());
         if (table.coreOptions().localMergeEnabled() && table.schema().primaryKeys().size() > 0) {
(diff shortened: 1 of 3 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkSinkBuilder.java
```diff
@@ -282,4 +285,39 @@ private DataStream<RowData> trySortInput(DataStream<RowData> input) {
...
+    private Integer checkAndUpdateParallelism(DataStream<?> input, Integer parallelism) {
+        try {
+            boolean parallelismUndefined = parallelism == null || parallelism == -1;
+            boolean isStreaming = FlinkSink.isStreaming(input);
+            boolean isAdaptiveParallelismEnabled =
+                    AdaptiveParallelism.isEnabled(input.getExecutionEnvironment());
+            boolean writeMCacheEnabled = table.coreOptions().writeManifestCache().getBytes() > 0;
+            boolean hashDynamicMode = table.bucketMode() == BucketMode.HASH_DYNAMIC;
+            if (parallelismUndefined
+                    && isStreaming
```
JingsongLi, line 298: is streaming? not batch?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkSinkBuilder.java
```diff
@@ -282,4 +285,50 @@ private DataStream<RowData> trySortInput(DataStream<RowData> input) {
...
+                    && !isStreaming
+                    && isAdaptiveParallelismEnabled
+                    && (writeMCacheEnabled || hashDynamicMode)) {
+                List<String> messages = new ArrayList<>();
+                if (writeMCacheEnabled) {
+                    messages.add("Write Manifest Cache");
+                }
+                if (hashDynamicMode) {
+                    messages.add("Dynamic Bucket Mode");
+                }
```
wwj6591812, line 307: Add delimiter between "Write Manifest Cache" and "Dynamic Bucket Mode" if writeMCacheEnabled and hashDynamicMode are all true.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkSinkBuilder.java
```diff
@@ -75,7 +77,7 @@ public class FlinkSinkBuilder {
 
     private DataStream<RowData> input;
     @Nullable protected Map<String, String> overwritePartition;
-    @Nullable protected Integer parallelism;
+    @Nullable private Integer parallelism;
```
wwj6591812, line 80: Could change Integer to int.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkSinkBuilder.java
```diff
@@ -282,4 +285,50 @@ private DataStream<RowData> trySortInput(DataStream<RowData> input) {
         }
         return input;
     }
+
+    private Integer checkAndUpdateParallelism(DataStream<?> input, Integer parallelism) {
```
wwj6591812, line 289: Could change Integer to int.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkSinkBuilder.java
yunfengzhou-hub: The delimiter would be added during the message list's toString() method. Currently it looks like this

> Paimon Sink with [Write Manifest Cache, Dynamic Bucket Mode] does not support Flink's Adaptive Parallelism mode. Configuring sink parallelism to `input parallelism` instead. You can also set Paimon `sink.parallelism` manually to override this configuration.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkSinkBuilder.java
yunfengzhou-hub: Different from an Integer, an int cannot be null. The Integer being null or not shows whether the parallelism has been explicitly configured, which might further influences internal implementations. Thus I prefer to keep the existing usage of Integer to better preserve this information.

A similar pattern exists in flink's Transformation, where it needs `boolean parallelismConfigured` along with the `int parallelism`.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkSinkBuilder.java
yunfengzhou-hub: Same as above.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkSinkBuilder.java
wwj6591812: OK

JingsongLi (approved):
+1

PR #4222, opened 2024-09-19
Compatible with Spark4 （upgrade antlr4 to version 4.13.1 Compatible with jdk17  ）

awol2005ex:
Compatible with Spark4 （upgrade antlr4 to version 4.13.1 Compatible with jdk17  ）


JingsongLi:
work in progress in https://github.com/apache/paimon/pull/4325 , the benefit is avoiding codes copy.

PR #4223, opened 2024-09-19
[core][spark]support format table create, and implement format table for SparkCatalog

adrian-wang:
### Purpose

Support format table creation, and add an implementation of format table for SparkCatalog.

### Tests

Added `SparkCatalogWithHiveTest`

### API and Format

No API changed.

### Documentation

https://cwiki.apache.org/confluence/display/PAIMON/PIP-27%3A+Introduce+Format+Table+in+Paimon+Catalog

### DIFF
--- a/paimon-spark/paimon-spark-common/src/test/scala/org/apache/paimon/spark/sql/DDLTestBase.scala
+++ b/paimon-spark/paimon-spark-common/src/test/scala/org/apache/paimon/spark/sql/DDLTestBase.scala
@@ -154,9 +154,7 @@ abstract class DDLTestBase extends PaimonSparkTestBase {
       val error = intercept[Exception] {
         spark.sql(s"CREATE TABLE parquet_tbl (id int) USING parquet")
       }.getMessage
-      assert(
-        error.contains(
-          "SparkCatalog can only create paimon table, but current provider is parquet"))
+      assert(error.contains("does not support format table"))
     }
   }
 


JingsongLi (commented):
Hi @adrian-wang , Thanks for the contribution!

1. Can you add documentation? Maybe in `sql-ddl`?
2. Can you add a test case? Maybe in `SparkGenericCatalogWithHiveTest`?

About how to create a format table, maybe we can just introduce a option `type`? Type = format?

JingsongLi:
See https://github.com/apache/paimon/pull/4243

paimon-spark/paimon-spark-common/src/test/java/org/apache/paimon/spark/SparkCatalogWithHiveTest.java
```diff
@@ -0,0 +1,102 @@
...
+                                "spark.sql.catalog.spark_catalog.warehouse",
+                                warehousePath.toString())
+                        .master("local[2]")
+                        .getOrCreate();
+
+        spark.sql("CREATE DATABASE IF NOT EXISTS my_db1");
+        spark.sql("USE spark_catalog.my_db1");
+        spark.sql(
+                "CREATE TABLE IF NOT EXISTS table_orc (a INT, bb INT, c STRING) USING orc TBLPROPERTIES"
+                        + " ('paimon.format.table'='true')");
```
JingsongLi, line 69: Can we avoid setting this option?

adrian-wang: This is to identify a format table, with provider set as specified format.

adrian-wang: If use `type` in #4243 , I feel like maybe it is not strong enough to tell generic catalog to delegate it to non-generic catalog

adrian-wang:
@melin need to extend jdbc table def before support it.

JingsongLi (approved):
Looks good to me!

Issue #4209, opened 2024-09-18
[Feature] Support nested col pruning push down


Zouxxyy:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Motivation

Paimon's format reader alreadly support read nested struct type with purning, add test for it: https://github.com/apache/paimon/pull/4210

we can try to push down nested projection to format reader

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #4227, opened 2024-09-19
[core] Add immutable id for system field 

Zouxxyy:
### Purpose

Field id should not be changed at any time

- Add immutable id for system field
- Make all fileds immutable even in calculating internally

Preparation for #4209


JingsongLi:
Thanks @Zouxxyy for the contribution.

I just want to figure out what is the impact of this PR? Does it affect the storage?

Zouxxyy:
> I just want to figure out what is the impact of this PR? Does it affect the storage?

no，these system field just used when computer, and not persist in schema(storage)

Zouxxyy:
This PR is prepare for using rowType to replace projection, because I want rowType to be immutable.
And, in theory, the id of the field is unique and cannot be modified once written, because schema evolution is based on id

JingsongLi (approved):
+1

PR #4228, opened 2024-09-20
[core] Reduce unnecessary manifest reading during orphan clean

bknbkn:
### Purpose

Read manifest to find used files will cost time, if the candidates is empty, we don't need to do that


wwj6591812:
+1

PR #4229, opened 2024-09-20
[core] Optimize IncrementalStartingScanner to randomlyExecute

JingsongLi:
### Purpose

Sequence batch should not be used; it only optimizes memory, and we only need to return the results in order.

Randomly execution can improve performance.


wwj6591812:
+1

PR #4235, opened 2024-09-23
[core] Check snapshot existence before expire

leaves12138:
### Purpose

If there is two writing job directly aiming to two different partitions, then there'll come up a conflict while expiring snapshot. This should be avoided.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/ExpireSnapshotsImpl.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/ExpireSnapshotsImpl.java
@@ -32,6 +32,7 @@ import org.apache.paimon.utils.TagManager;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.FileNotFoundException;
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


paimon-core/src/main/java/org/apache/paimon/table/ExpireSnapshotsImpl.java
```diff
@@ -162,21 +163,27 @@ public int expireUntil(long earliestId, long endExclusiveId) {
...
-                        String.format(
-                                "Skip cleaning data files of snapshot '%s' due to failed to build skipping set.",
-                                id),
-                        e);
-                continue;
-            }
-
-            snapshotDeletion.cleanUnusedDataFiles(snapshot, skipper);
+            final long snapshotId = id;
+            quietExecute(
```
JingsongLi, line 167: How about:
```
Snapshot snapshot;
try {
    snapshot = snapshotManager.tryGetSnapshot(id);
} catch (FileNotFoundException e) {
    beginInclusiveId = id + 1;
    continue;
}
```

JingsongLi (approved):
+1

PR #4240, opened 2024-09-23
[core] Fast return current table if target branch is same with current

ulysses-you:
### Purpose

We can just return current table if the target branch of `switchToBranch` is same with current. For example, `RemoveOrphanFiles` will always do `switchToBranch` for main branch.

### Tests

Pass CI

### API and Format

no


JingsongLi (approved):
+1

PR #4109, opened 2024-09-02
[flink] Support Flink Materialized Table

reswqa:
### Purpose

Linked issue: close #4045

Flink has introduced [Materialized Table](https://nightlies.apache.org/flink/flink-docs-release-1.20/docs/dev/table/materialized-table/overview/) in 1.20 release, we can support it in Paimon connector.

### Documentation

No

PR #4243, opened 2024-09-24
[core] Introduce type option

JingsongLi:
### Purpose

No we have multiple tables:
1. normal paimon table
2. format table
3. Flink materialized table

We can have a type option to describe this.

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
@@ -87,6 +87,12 @@ public class CoreOptions implements Serializable {
 
     public static final String COLUMNS = "columns";
 
+    public static final ConfigOption<TableType> TYPE =
+            key("type")
+                    .enumType(TableType.class)
+                    .defaultValue(TableType.TABLE)
+                    .withDescription("Type of the table.");
+
     public static final ConfigOption<Integer> BUCKET =
             key("bucket")
                     .intType()
--- a/paimon-common/src/main/java/org/apache/paimon/TableType.java
+++ b/paimon-common/src/main/java/org/apache/paimon/TableType.java
@@ -54,6 +54,6 @@ public enum TableType implements DescribedEnum {
                 return type;
             }
         }
-        throw new UnsupportedOperationException("Unknown type: " + name);
+        throw new UnsupportedOperationException("Unknown table type: " + name);
     }
 }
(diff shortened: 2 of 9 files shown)


paimon-common/src/main/java/org/apache/paimon/TableType.java
```diff
@@ -0,0 +1,59 @@
...
+package org.apache.paimon;
+
+import org.apache.paimon.options.description.DescribedEnum;
+import org.apache.paimon.options.description.InlineElement;
+
+import static org.apache.paimon.options.description.TextElement.text;
+
+/** Type of the table. */
+public enum TableType implements DescribedEnum {
+    TABLE("table", "Normal Paimon table."),
```
wwj6591812, line 28: where is 'Flink materialized table'?

paimon-common/src/main/java/org/apache/paimon/TableType.java
```diff
@@ -0,0 +1,59 @@
...
+        return text(description);
+    }
+
+    public static TableType fromString(String name) {
+        for (TableType type : TableType.values()) {
+            if (type.value.equals(name)) {
+                return type;
+            }
+        }
+        throw new UnsupportedOperationException("Unknown type: " + name);
```
wwj6591812, line 57: change "Unknown type" to "Unknown table type"?

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -87,6 +87,12 @@ public class CoreOptions implements Serializable {
 
     public static final String COLUMNS = "columns";
 
+    public static final ConfigOption<TableType> TYPE =
+            key("type")
```
wwj6591812, line 91: Change key("type") to key("table-type")?

paimon-common/src/main/java/org/apache/paimon/TableType.java
JingsongLi: Let https://github.com/apache/paimon/pull/4109 add it.

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
JingsongLi: No, table-type has been added for hive, and its value is Paimon.

Another type for table.

adrian-wang:
+1

wwj6591812:
+1

PR #4245, opened 2024-09-24
[format] Format prefix in option keys are processed in each format

tsreaper:
### Purpose

Currently, format prefix in option keys are removed before sending into the constructor of each format. However, most formats (orc, parquet) still recover these prefixes in their code, so it would be better to let the format itself deal with these prefixes.

Also, we're going to introduce a special `compacted changelog` format, which wraps other file formats. However with the current code, each format can only see the options related to itself. If option keys are processed in each format, we can see options for other formats and keep what we need.

This PR process option keys in each format.

### Tests

### API and Format

No format changes.

### Documentation

No new feature.

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/format/FileFormat.java
+++ b/paimon-common/src/main/java/org/apache/paimon/format/FileFormat.java
@@ -28,7 +28,9 @@ import org.apache.paimon.types.RowType;
 import javax.annotation.Nullable;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.ServiceLoader;
 
@@ -106,5 +108,21 @@ public abstract class FileFormat {
         }
 
         return Optional.empty();
+    }
+
+    protected Options getIdentifierPrefixOptions(Options options, boolean keepPrefix) {
+        Map<String, String> result = new HashMap<>();
+        String prefix = formatIdentifier.toLowerCase() + ".";
+        for (String key : options.keySet()) {
+            if (key.toLowerCase().startsWith(prefix)) {
+                String substr = key.substring(prefix.length());
+                if (keepPrefix) {
+                    result.put(prefix + substr, options.get(key));
+                } else {
+                    result.put(substr, options.get(key));
+                }
+            }
+        }
+        return new Options(result);
     }
 }
--- a/paimon-common/src/main/java/org/apache/paimon/format/FileFormat.java
+++ b/paimon-common/src/main/java/org/apache/paimon/format/FileFormat.java
@@ -19,7 +19,6 @@
 package org.apache.paimon.format;
 
 import org.apache.paimon.CoreOptions;
-import org.apache.paimon.annotation.VisibleForTesting;
 import org.apache.paimon.format.FileFormatFactory.FormatContext;
 import org.apache.paimon.options.Options;
 import org.apache.paimon.predicate.Predicate;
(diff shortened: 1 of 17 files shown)


paimon-format/src/main/java/org/apache/paimon/format/avro/AvroFileFormat.java
```diff
@@ -62,11 +62,14 @@ public class AvroFileFormat extends FileFormat {
...
 
-    private final FormatContext context;
+    private final Options options;
+    private final int zstdLevel;
 
     public AvroFileFormat(FormatContext context) {
         super(IDENTIFIER);
-        this.context = context;
+
+        this.options = getIdentifierPrefixOptions(context.options(), false);
```
JingsongLi, line 71: Can you also modify `AVRO_OUTPUT_CODEC` and `AVRO_ROW_NAME_MAPPING` to add prefix?

paimon-common/src/main/java/org/apache/paimon/format/FileFormat.java
```diff
@@ -103,14 +110,19 @@ private static Optional<FileFormat> fromIdentifier(
...
-                        options.get(CoreOptions.WRITE_BATCH_SIZE),
-                        options.get(CoreOptions.FILE_COMPRESSION_ZSTD_LEVEL),
-                        options.get(CoreOptions.FILE_BLOCK_SIZE));
-        return FileFormat.fromIdentifier(formatIdentifier, context);
+    protected Options getIdentifierPrefixOptions(Options options, boolean keepPrefix) {
+        Map<String, String> result = new HashMap<>();
+        String prefix = formatIdentifier.toLowerCase() + ".";
+        for (String key : options.keySet()) {
+            if (key.toLowerCase().startsWith(prefix)) {
+                String substr = key.substring(prefix.length());
```
wwj6591812, line 118: Change substr to another better name?

paimon-format/src/main/java/org/apache/paimon/format/avro/AvroFileFormat.java
```diff
@@ -62,11 +62,14 @@ public class AvroFileFormat extends FileFormat {
     public static final ConfigOption<Map<String, String>> AVRO_ROW_NAME_MAPPING =
```
wwj6591812, line 62: In AvroFileFormat, IDENTIFIER && AVRO_OUTPUT_CODEC && AVRO_ROW_NAME_MAPPING should be private.

paimon-common/src/main/java/org/apache/paimon/format/FileFormat.java
```diff
@@ -74,9 +75,15 @@ public Optional<SimpleStatsExtractor> createStatsExtractor(
         return Optional.empty();
     }
 
-    @VisibleForTesting
     public static FileFormat fromIdentifier(String identifier, Options options) {
-        return fromIdentifier(identifier, new FormatContext(options, 1024, 1024));
+        return fromIdentifier(
```
wwj6591812, line 79: This called method fromIdentifier should be private.

wwj6591812:
Good work!
After download your code in my computer, I have found no big problems.
Already add three small suggest.

paimon-format/src/test/java/org/apache/paimon/format/orc/OrcFileFormatTest.java
```diff
@@ -47,8 +47,8 @@ public void testAbsent() {
     @Test
     public void testPresent() {
         Options options = new Options();
-        options.setString("haha", "1");
-        options.setString("compress", "zlib");
+        options.setString("orc.haha", "1");
```
wwj6591812, line 50: Change haha to other normal orc option?

wwj6591812:
Good work!
After download your code in my computer, I have found no big problems.
Already add four small suggests.

JingsongLi:
+1

PR #4246, opened 2024-09-24
[common][flink] Add support for complex types in kafka debezium avro cdc action

AshishKhatkar:
### Purpose
@umeshdangat has an outstanding PR which was updated after PR by @zhuangchong was merged. That PR: https://github.com/apache/paimon/pull/3323

The above patch allows consuming data from avro data from kafka into paimon but it doesnt support complex avro types. This PR achieves that. The original PR https://github.com/zhuangchong/flink-table-store/pull/1 was pointing to @zhuangchong branch to clearly show the changes only relavant to supporting complex avro types.

This PR fixes the failing E2E tests for PR: https://github.com/apache/paimon/pull/3931

Copying the note from PR: https://github.com/apache/paimon/pull/3931

> One issue is CdcSourceRecord contains Map<String, String> thus the current somewhat tedious approach is to deserialize avro complex types into json strings and then read them back from json strings rather than changing CdcSourceRecord Map<String, Object> to support value as Object. It would be a much larger change looking at the code changes needed.
> 
> I had to update the DataField to add a method for dataFieldEqualsIgnoreId which already existing in RichEventParser. For nested RowType fields this becomes necessary (coming from nested avro records) as when a DataField.type= RowType we cannot simply do equals on all data fields as they contain Id as well and it fails the equality, although there is no schema change.

@JingsongLi @zhuangchong let us know what you think.

cc: @umeshdangat

Linked issue: close xxx

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/utils/BinaryStringUtils.java
+++ b/paimon-common/src/main/java/org/apache/paimon/utils/BinaryStringUtils.java
@@ -336,11 +336,11 @@ public class BinaryStringUtils {
                 break;
             case 6: // microseconds
                 millis = epoch / 1000;
-                nanosOfMillis = (int)((epoch % 1000) * 1000);
+                nanosOfMillis = (int) ((epoch % 1000) * 1000);
                 break;
             case 9: // nanoseconds
                 millis = epoch / 1_000_000;
-                nanosOfMillis = (int)(epoch % 1_000_000);
+                nanosOfMillis = (int) (epoch % 1_000_000);
                 break;
             default:
                 throw new RuntimeException("Unsupported precision: " + precision);
--- a/paimon-common/src/main/java/org/apache/paimon/utils/BinaryStringUtils.java
+++ b/paimon-common/src/main/java/org/apache/paimon/utils/BinaryStringUtils.java
@@ -319,7 +319,7 @@ public class BinaryStringUtils {
         return DateTimeUtils.parseTimestampData(input.toString(), precision, timeZone);
     }
 
-    // Helper method to convert milliseconds to Timestamp with the provided precision.
+    // Helper method to convert epoch to Timestamp with the provided precision.
     private static Timestamp fromMillisToTimestamp(long epoch, int precision) {
         // Calculate milliseconds and nanoseconds from epoch based on precision
         long millis;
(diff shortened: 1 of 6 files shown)


JingsongLi:
CC @yuzelin to take a reivew~

JingsongLi:
Thanks @AshishKhatkar

paimon-common/src/main/java/org/apache/paimon/utils/BinaryStringUtils.java
```diff
@@ -315,6 +322,25 @@ public static Timestamp toTimestamp(BinaryString input, int precision, TimeZone
...
+        long seconds = millis / 1000;
+        int nanos = (int) ((millis % 1000) * 1_000_000);
+
+        // Adjust the nanoseconds to the specified precision
+        if (precision < 9) {
+            nanos =
+                    (int)
+                            (Math.floor(nanos / Math.pow(10, 9 - precision))
+                                    * Math.pow(10, 9 - precision));
+        }
```
yuzelin, line 337: I think this is unnecessary. We can store the data and let the readers truncate the data based on precision.
So we can just use `Timestamp#fromEpochMillis(long milliseconds, int nanosOfMillisecond)`

yuzelin (commented):
Thanks，@AshishKhatkar! I left a comment, please take a look~

paimon-common/src/main/java/org/apache/paimon/utils/BinaryStringUtils.java
AshishKhatkar: I read the documentation for `Timestamp#fromEpochMillis(long milliseconds, int nanosOfMillisecond)` and it expects nanosOfMs to be in range [0, 999_999]. Are you saying we need to default to precision 6 here?

yuzelin: For example, if we want to store `1970-01-01 00:00:00.123456789`, the first argument `milliseconds` is `123` and the second argument `nanosOfMillisecond` is `456789`.

AshishKhatkar: @yuzelin done.

yuzelin: Thanks~

yuzelin (approved):
+1

PR #4256, opened 2024-09-25
[core] Make ClientPool simple to reduce thread safe bugs

JingsongLi:
### Purpose

In previous implementations, there were many places where locks were not used to ensure visibility, which may lead to infinite waiting. We should make the logic of the Pool simple enough to ensure its security.


yuzelin (approved):
+1

PR #4258, opened 2024-09-25
[core] Push bucket filter into manifest entry reader

JingsongLi:
### Purpose

We already have bucket filter in manifest entry reader, we can move all bucket filters into manifest entry reader, and do the validation in outside write.

This can simplify the codes in `AbstractFileStoreScan`.


yuzelin (approved):
+1

PR #4260, opened 2024-09-26
[core] Optimize UnawareAppendTableCompactionCoordinator to avoid OOM

JingsongLi:
### Purpose

If the table is very large and has many partitions, it can lead to putting too many ManifestEntries directly into memory, causing JVM OOM, and this situation is difficult to recover, with almost no means available at present.

So in this PR, the process of Plan is changed to Iterator mode, so that Compact can be safely performed regardless of the size of the table.


leaves12138 (approved):
+1

PR #4262, opened 2024-09-26
[test] Fix unstable tests in PrimaryKeyFileStoreTableITCase

tsreaper:
### Purpose

This PR fixes unstable tests in `PrimaryKeyFileStoreTableITCase` by:

1. Increase write buffer size so not too many files are created during testing.
2. Set `continuous.discovery-interval` to a very small value. Because during testing, Paimon table will produce snapshots very quickly, so consumers of these snapshots should also discover new snapshots very quickly.

### Tests

Existing test should cover this change.

### API and Format

No format changes.

### Documentation

No new feature.


JingsongLi (approved):
+1

PR #4263, opened 2024-09-26
[arrow] Enable upper-case field name in arrow format writer

leaves12138:
### Purpose

* [arrow] Enable upper case in arrow format writer
* [arrow] Fix bug for timestamp type converter
* [arrow] Add log


JingsongLi:
+1

PR #4269, opened 2024-09-26
[spark] Support nested col pruning

Zouxxyy:
### Purpose

to #4209, Support nested col pruning,  e.g.

```sql
CREATE TABLE students (
    name STRING,
    age INT,
    course STRUCT<course_name: STRING, grade: DOUBLE>
) USING paimon;
```

```sql
SELECT course.grade FROM students;
```

will only obtain `course.grade` from colume-storage-format (parquet, orc)

### DIFF
--- a/paimon-format/src/main/java/org/apache/paimon/format/parquet/ParquetReaderFactory.java
+++ b/paimon-format/src/main/java/org/apache/paimon/format/parquet/ParquetReaderFactory.java
@@ -33,7 +33,10 @@ import org.apache.paimon.fs.Path;
 import org.apache.paimon.options.Options;
 import org.apache.paimon.reader.RecordReader;
 import org.apache.paimon.reader.RecordReader.RecordIterator;
+import org.apache.paimon.types.ArrayType;
+import org.apache.paimon.types.DataField;
 import org.apache.paimon.types.DataType;
+import org.apache.paimon.types.MapType;
 import org.apache.paimon.types.RowType;
 import org.apache.paimon.utils.Pool;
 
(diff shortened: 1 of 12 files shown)


JingsongLi (commented):
Are there some configuration to disable nested projection? I am concern about bugs in nested projection, at least, we should have option to disable it.

Zouxxyy:
> Are there some configuration to disable nested projection? I am concern about bugs in nested projection, at least, we should have option to disable it.

yes, spark has a conf ti disable it

```scala
  val NESTED_SCHEMA_PRUNING_ENABLED =
    buildConf("spark.sql.optimizer.nestedSchemaPruning.enabled")
      .internal()
      .doc("Prune nested fields from a logical relation's output which are unnecessary in " +
        "satisfying a query. This optimization allows columnar file format readers to avoid " +
        "reading unnecessary nested column data. Currently Parquet and ORC are the " +
        "data sources that implement this optimization.")
      .version("2.4.1")
      .booleanConf
      .createWithDefault(true)
```

paimon-format/src/main/java/org/apache/paimon/format/parquet/ParquetReaderFactory.java
```diff
@@ -155,13 +156,37 @@ private MessageType clipParquetSchema(GroupType parquetSchema) {
...
 
+    /** Clips `parquetGroup` by `requestedRowType`. */
+    private GroupType clipParquetGroup(GroupType parquetGroup, RowType requestedRowType) {
+        Types.GroupBuilder<GroupType> builder = Types.buildGroup(parquetGroup.getRepetition());
+        List<DataField> fields = requestedRowType.getFields();
+        for (DataField field : fields) {
+            String fieldName = field.name();
+            if (parquetGroup.containsField(fieldName)) {
+                Type type = parquetGroup.getType(fieldName);
+                if (type instanceof GroupType && field.type() instanceof RowType) {
```
JingsongLi, line 178: Can Spark push down nested fields for array and map?

Zouxxyy: Not support, only support  struct col pruning

image

image

Zouxxyy: It support, update

JingsongLi (approved):
+1

PR #4270, opened 2024-09-27
[core] Delete redundant if statement when produce changelog by lookup for first row.

wwj6591812:
### Purpose

When I study the logic of produce changelog, I find a redundant if statement affects my understanding of the code logic.
In order not to affect more people, so now delete it.

Linked issue: close #xxx

### Tests

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/FirstRowMergeFunctionWrapper.java
+++ b/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/FirstRowMergeFunctionWrapper.java
@@ -66,11 +66,7 @@ public class FirstRowMergeFunctionWrapper implements MergeFunctionWrapper<Change
             return reusedResult;
         }
 
-        reusedResult.setResult(result);
-        if (result.level() == 0) {
-            // new record, output changelog
-            return reusedResult.addChangelog(result);
-        }
-        return reusedResult;
+        // new record, output changelog
+        return reusedResult.setResult(result).addChangelog(result);
     }
 }


wwj6591812:
The ut is fail, but not caused by my modify.

paimon-core/src/main/java/org/apache/paimon/mergetree/compact/FirstRowMergeFunctionWrapper.java
```diff
@@ -67,10 +67,10 @@ public ChangelogResult getResult() {
         }
 
         reusedResult.setResult(result);
-        if (result.level() == 0) {
-            // new record, output changelog
-            return reusedResult.addChangelog(result);
-        }
+
+        // mergeFunction.containsHighLevel is false, so new record, output changelog
```
Aitozi, line 71: what about 

```
// new record, output changelog
return reusedResult.setResult(result).addChangelog(result);
```

wwj6591812: done
good, very thanks.

Aitozi (approved):
LGTM

paimon-core/src/main/java/org/apache/paimon/mergetree/compact/FirstRowMergeFunctionWrapper.java
```diff
@@ -66,11 +66,7 @@ public ChangelogResult getResult() {
             return reusedResult;
         }
 
-        reusedResult.setResult(result);
-        if (result.level() == 0) {
```
JingsongLi, line 70: Why removing this `if`? What about merging level 4 and level 5?

Aitozi: if merging with high level, it should return at line 61.

JingsongLi:
+1

PR #4272, opened 2024-09-27
[core] Check existence before expire: more check.

leaves12138:
### Purpose

Two place need to check

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/ExpireSnapshotsImpl.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/ExpireSnapshotsImpl.java
@@ -231,7 +231,6 @@ public class ExpireSnapshotsImpl implements ExpireSnapshots {
             if (e.getCause() == null || !(e.getCause() instanceof FileNotFoundException)) {
                 throw e;
             }
-            LOG.info("Snapshot is deleted while expiring, skip.");
         }
 
         for (long id = beginInclusiveId; id < endExclusiveId; id++) {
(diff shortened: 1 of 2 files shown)


paimon-core/src/main/java/org/apache/paimon/table/ExpireSnapshotsImpl.java
```diff
@@ -213,8 +214,25 @@ public int expireUntil(long earliestId, long endExclusiveId) {
...
+            // the end exclusive snapshot is gone
+            // there is no need to proceed
+            return 0;
+        }
+
+        Set<String> skippingSet = new HashSet<>();
+        try {
+            skippingSet.addAll(snapshotDeletion.manifestSkippingSet(skippingSnapshots));
+        } catch (Exception e) {
+            // mayby snapshot been deleted by other jobs.
```
wwj6591812, line 230: mayby?

paimon-core/src/main/java/org/apache/paimon/table/ExpireSnapshotsImpl.java
```diff
@@ -213,8 +214,25 @@ public int expireUntil(long earliestId, long endExclusiveId) {
...
+            if (e.getCause() == null || !(e.getCause() instanceof FileNotFoundException)) {
+                throw e;
+            }
```
wwj6591812, line 233: Add a log here?
Explain why we catch Exception here and no longer throw them out.

paimon-core/src/main/java/org/apache/paimon/table/ExpireSnapshotsImpl.java
leaves12138: See https://github.com/apache/paimon/pull/4235

yuzelin (approved):
+1

paimon-core/src/main/java/org/apache/paimon/table/ExpireSnapshotsImpl.java
wwj6591812: OK

PR #4275, opened 2024-09-29
[spark] Support push down aggregate with group by partition column

ulysses-you:
### Purpose

If the aggregate all group by keys are from partition columns, we can support push down it. It makes user to get every partition (or with partition filter) row count fast. 

### Tests

add test

### API and Format

no


ulysses-you:
cc @JingsongLi thank you

JingsongLi (approved):
+1

YannByron:
Link to https://github.com/apache/paimon/issues/2404.

PR #4281, opened 2024-09-30
[flink] Emit event time lag should not increase when no new records are read

tsreaper:
### Purpose

Currently for source with consumers, `currentEmitEventTimeLag` will continue to increase even if no new records are read. According to the definition of this metrics, this time lag is the different between the event time and the emit time of the last record.

This PR fixes the issue by sticking to the definition of the metrics.

### Tests

Unit tests.

### API and Format

No format changes.

### Documentation

No new feature.


wwj6591812:
+1

JingsongLi (approved):
+1

PR #4284, opened 2024-10-07
[test] add more test case for schema evolution with schema included in records

zhuyaogai:
### Purpose

Add more test case for schema evolution with schema included in records. Currently, there are only test cases for schema evolution with no schema included in records(debezium-json format), so all the columns are treated as STRING type, and  only cover the schema evolution case of `Add Columns`. So I add more test case to check schema evolution, includes:
1. column type promotion (such as int -> bigint)
2. column type changed ignore (such as bigint -> int)
3. column type covert exception (such as bigint -> string)

Linked issue: close #xxx

### Tests

org.apache.paimon.flink.action.cdc.kafka.KafkaDebeziumSyncTableActionITCase


zhuyaogai:
@JingsongLi hi, master, could you take a look at this when you're free? Thanks!

JingsongLi:
+1

PR #4286, opened 2024-10-08
[core] Optimize multiple commit to reduce conflicts

JingsongLi:
### Purpose

At present, if the number of files is very large and the commit interval is relatively small, and multiple jobs are written simultaneously, there will be serious competition, and even retry failures (more than ten times will fail).

This is because the data files conflicts checking may be triggered at present, which requires a relatively long time to read data files from old snapshot. If other jobs commit at this time, and repeated commit may still fail because of repeated conflicts checking.

This is very wasteful. We can actually reuse the base files, we can just read incremental files and merge it to last time base files.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
@@ -821,6 +821,7 @@ public class FileStoreCommitImpl implements FileStoreCommit {
                 break;
             }
 
+            // TODO optimize OVERWRITE too
             RetryResult retryResult = (RetryResult) result;
             retryResult.cleanAll();
         }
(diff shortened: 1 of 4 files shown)


paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
```diff
@@ -1363,4 +1395,104 @@ static ConflictCheck noConflictCheck() {
...
+    private class RetryResult implements CommitResult {
+
+        private final String deltaManifestList;
+        private final String changelogManifestList;
+
+        private final String oldIndexManifest;
+        private final String newIndexManifest;
+
+        private final List<ManifestFileMeta> mergeBeforeManifests;
+        private final List<ManifestFileMeta> mergeAfterManifests;
```
tsreaper, line 1420: These are logically the same. What we really want to do is to optimize conflict checking.

Let's say we want to commit snapshot $s$, but the latest snapshot is snapshot $(s + k)$. Because we already confirmed that our new changes is compatible to snapshots up to $s$, during next retry we only need to check delta manifests in snapshots $(s + 1)$ to $(s + k)$.

paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
```diff
@@ -876,41 +923,54 @@ boolean tryCommitOnce(
...
+            boolean rewriteIndexManifest = true;
+            if (retryResult != null) {
+                if (Objects.equals(oldIndexManifest, retryResult.oldIndexManifest)) {
+                    rewriteIndexManifest = false;
+                    indexManifest = retryResult.newIndexManifest;
+                    LOG.info("Reusing index manifest {} for retry.", indexManifest);
+                } else {
+                    cleanIndexManifest(retryResult.oldIndexManifest, retryResult.newIndexManifest);
+                }
+            }
```
tsreaper, line 969: NIT: Use a big `if... else...` block, instead of two `?:` and an `if... else...`

tsreaper (approved):

PR #4287, opened 2024-10-08
[core] Filter data file while executing local orphan clean

bknbkn:
### Purpose

Local Remove Orphan Files OOM always occurred when collecting data file meta. Because  data files in tables may be large (could reach tens of millions or more). 

Relatively speaking, the number of candidate files is not large (ten thousand or hundreds of thousands).

So we can just collect data file meta in candidate files to reduce the probability of OOM occurrence.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/LocalOrphanFilesClean.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/LocalOrphanFilesClean.java
@@ -106,7 +106,8 @@ public class LocalOrphanFilesClean extends OrphanFilesClean {
         // delete unused files
         candidateDeletes.removeAll(usedFiles);
         candidateDeletes.stream().map(candidates::get).forEach(fileCleaner);
-        deleteFiles.addAll(candidateDeletes.stream().map(candidates::get).collect(Collectors.toList()));
+        deleteFiles.addAll(
+                candidateDeletes.stream().map(candidates::get).collect(Collectors.toList()));
 
         return deleteFiles;
     }
@@ -161,8 +162,8 @@ public class LocalOrphanFilesClean extends OrphanFilesClean {
                                     dataFiles.add(f.fileName());
                                 }
                                 f.extraFiles().stream()
-                                    .filter(candidateDeletes::contains)
-                                    .forEach(dataFiles::add);
+                                        .filter(candidateDeletes::contains)
+                                        .forEach(dataFiles::add);
                             });
         }
         return dataFiles;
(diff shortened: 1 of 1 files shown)


paimon-core/src/main/java/org/apache/paimon/operation/LocalOrphanFilesClean.java
```diff
@@ -101,10 +104,10 @@ public List<Path> clean() throws IOException, ExecutionException, InterruptedExc
...
 
         // delete unused files
-        Set<String> deleted = new HashSet<>(candidates.keySet());
-        deleted.removeAll(usedFiles);
-        deleted.stream().map(candidates::get).forEach(fileCleaner);
-        deleteFiles.addAll(deleted.stream().map(candidates::get).collect(Collectors.toList()));
+        candidateDeletes.removeAll(usedFiles);
+        candidateDeletes.stream().map(candidates::get).forEach(fileCleaner);
+        deleteFiles.addAll(
+                candidateDeletes.stream().map(candidates::get).collect(Collectors.toList()));
```
wwj6591812, line 111: candidateDeletes.clear();

wwj6591812:
I think this is a good optimization, but two questions.
1、In your production environment, you encountered OOM here? By dump heap and analyze?
2、Since you have done LocalOrphanFilesClean, can you also add this optimization to the distributed OrphanFilesClean?

bknbkn:
I'm not sure whether this optimization can be used in a distributed situation, because that means that the candidate set needs to be broadcast to each executor. At this time, the number of data file meta in each executor is not large, and there is no risk of OOM. 

On the other hand, the process of joining condidate set and used file set actually implicitly includes the process of broadcasting and filtering.

And the analyze in production environment will be given later.

paimon-core/src/main/java/org/apache/paimon/operation/LocalOrphanFilesClean.java
bknbkn: fixed

bknbkn:
Test in table with 7TB size:

**Before this patch, it will OOM when it collect 80 million data file metas:**
490d8adf-0f08-42f8-8453-01e60ba9849f
23436c6d-280d-481c-988c-7beead657e69

After optimization, it only collect 1000+ manifests and 60000+ data file metas, the procedure only cost 20min on local machine:
e46b828e-1013-4e81-86c2-b92e5daae241
c6ff7173-50f9-4e50-ba72-9d02bf712a20

paimon-core/src/main/java/org/apache/paimon/operation/LocalOrphanFilesClean.java
```diff
@@ -114,7 +118,7 @@ private List<String> getUsedFiles(String branch) {
         ManifestFile manifestFile =
                 table.switchToBranch(branch).store().manifestFileFactory().create();
         try {
-            List<String> manifests = new ArrayList<>();
+            Set<String> manifests = new HashSet<>();
```
bknbkn, line 121: While reading multiple snapshots, there will be duplicate manifests. Deduplication can improve efficiency.

JingsongLi (approved):
Good point! Thanks @bknbkn

PR #4251, opened 2024-09-24
[core] Support statistic with time travel

xuzifu666:
### Purpose

Scenario：user query statistic table by snapshot-id and analyzed table multiple times. Such as user can do the analyzed at regular time and collect the statistic metric to monitor.
Currently paimon query statistic system table with snapshot_id would error out due to PaimonAnalyzeTableColumnCommand##commitStatistics not keep the real snapshot id the same with statistic snapshot_id，the pr is aimed to fix it.
before fix:

after fix:

Linked issue: close #xxx

### Tests
AnalyzeTableTestBase "Paimon analyze: test statistic system table with predicate"

PR #1667, opened 2023-07-26
[core] support querying the manifest of the specified snapshot.

liming30:
### Purpose

Linked issue: close #1666 

[core] support querying the manifest of the specified snapshot.

### Tests

Added `org.apache.paimon.table.system.ManifestsTableTest` to verify correctness.

### API and Format

No.

### Documentation

No.

PR #4291, opened 2024-10-08
[core] Manifest and Statistic support query by tag name

xuzifu666:
### Purpose

Linked issue: close #xxx
Follow up https://github.com/apache/paimon/pull/1667 and https://github.com/apache/paimon/pull/4251 
which can only support query by snapshot id, this pr is aimed to support tag name query.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/AbstractFileStoreTable.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/AbstractFileStoreTable.java
@@ -59,13 +59,7 @@ import org.apache.paimon.table.source.snapshot.SnapshotReaderImpl;
 import org.apache.paimon.table.source.snapshot.StaticFromTimestampStartingScanner;
 import org.apache.paimon.table.source.snapshot.StaticFromWatermarkStartingScanner;
 import org.apache.paimon.tag.TagPreview;
-import org.apache.paimon.utils.BranchManager;
-import org.apache.paimon.utils.Preconditions;
-import org.apache.paimon.utils.SegmentsCache;
-import org.apache.paimon.utils.SimpleFileReader;
-import org.apache.paimon.utils.SnapshotManager;
-import org.apache.paimon.utils.SnapshotNotExistException;
-import org.apache.paimon.utils.TagManager;
+import org.apache.paimon.utils.*;
 
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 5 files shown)


paimon-core/src/main/java/org/apache/paimon/table/AbstractFileStoreTable.java
```diff
@@ -170,8 +171,15 @@ public Identifier identifier() {
     public Optional<Statistics> statistics() {
         Snapshot latestSnapshot;
         Long snapshotId = coreOptions().scanSnapshotId();
+        String tagName = coreOptions().scanTagName();
+
```
wwj6591812, line 175: Here add a check, if both snapshotId and tagName are not null, then there is ambiguity here.

paimon-core/src/main/java/org/apache/paimon/table/system/ManifestsTable.java
```diff
@@ -211,7 +213,11 @@ private static List<ManifestFileMeta> allManifests(FileStoreTable dataTable) {
             }
             snapshot = snapshotManager.snapshot(snapshotId);
         } else if (snapshotId == null) {
```
wwj6591812, line 215: here, Condition 'snapshotId == null' is always 'true'

paimon-core/src/main/java/org/apache/paimon/table/system/ManifestsTable.java
```diff
@@ -198,6 +199,7 @@ private static List<ManifestFileMeta> allManifests(FileStoreTable dataTable) {
         CoreOptions coreOptions = CoreOptions.fromMap(dataTable.options());
         SnapshotManager snapshotManager = dataTable.snapshotManager();
         Long snapshotId = coreOptions.scanSnapshotId();
+        String tagName = coreOptions.scanTagName();
```
wwj6591812, line 202: Here add a check, if both snapshotId and tagName are not null, then there is ambiguity here.

xuzifu666: tagName had checked follow line：if (!StringUtils.isEmpty(tagName) && dataTable.tagManager().tagExists(tagName))

paimon-core/src/main/java/org/apache/paimon/table/AbstractFileStoreTable.java
xuzifu666: tagName had checked follow line：if (!StringUtils.isEmpty(tagName) && dataTable.tagManager().tagExists(tagName))，so here may not need add a check.

paimon-core/src/main/java/org/apache/paimon/table/system/ManifestsTable.java
xuzifu666: yes ，change to else

paimon-core/src/main/java/org/apache/paimon/table/AbstractFileStoreTable.java
xuzifu666: If user both set snapshot-id and tag-name would be ambiguity here  would add a warning log for it, if user both set, would use snapshot-id firstly.

LinMingQiang:
+1

wwj6591812:
+1

wwj6591812:
+1

JingsongLi (approved):
+1

PR #4292, opened 2024-10-09
[core] Delete tag not failed if tag not exists

askwang:
### Purpose

when re-create a tag with the same name, we should delete it first and dont want to the job failed if tag not exists, like this:
```
CALL sys.delete_tag(table => 'default.T', tag => 'my_tag');
CALL sys.create_tag(table => 'default.T', tag => 'my_tag');
```
This is like `drop table [if exists] tb` usage.


JingsongLi (approved):
+1

PR #4293, opened 2024-10-09
[spark] Add test case for writing nested array

Zouxxyy:
### Purpose

- `WrappedArray` was removed in Scala 2.13, replaced it by `scala.collection.Seq`
- Optimize seq to list
- Add test case for https://github.com/apache/paimon/pull/1932


JingsongLi:
+1

PR #4296, opened 2024-10-09
[spark] Fix push down limit data correctness issue with deletion vector

ulysses-you:
### Purpose

If we enable the deletion vector, the raw files row count may not equal to the actual row count. We can not push down limit if there are some deletion files. 

### Tests

add test

### API and Format

no


JingsongLi (approved):
+1

PR #4299, opened 2024-10-10
[core][spark] Support push down limit for primary key table

ulysses-you:
### Purpose

For the primary key table, we still have a chance to push down limit if the split is raw convertible. This pr relaxes the push down limit condition to support primary key table.

### Tests

add test

### API and Format

no

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/source/DataTableBatchScan.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/source/DataTableBatchScan.java
@@ -96,6 +96,27 @@ public class DataTableBatchScan extends AbstractDataTableScan {
             long scannedRowCount = 0;
             SnapshotReader.Plan plan = ((ScannedResult) result).plan();
             List<DataSplit> splits = plan.dataSplits();
+            if (splits.isEmpty()) {
+                return result;
+            }
+
+            // We first add the rawConvertible split to avoid merging, and if the row count
+            // is still less than limit number, then add split which is not rawConvertible.
+            splits.sort(
+                    (x, y) -> {
+                        if (x.rawConvertible() && y.rawConvertible()) {
+                            return 0;
+                        } else if (x.rawConvertible()) {
+                            return -1;
+                        } else {
+                            return 1;
+                        }
+                    });
+            // fast return if there is no rawConvertible split
+            if (!splits.get(0).rawConvertible()) {
+                return result;
+            }
+
             List<Split> limitedSplits = new ArrayList<>();
             for (DataSplit dataSplit : splits) {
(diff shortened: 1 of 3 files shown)


paimon-core/src/main/java/org/apache/paimon/table/source/DataTableBatchScan.java
```diff
@@ -96,6 +96,27 @@ private StartingScanner.Result applyPushDownLimit(StartingScanner.Result result)
...
+                    (x, y) -> {
+                        if (x.rawConvertible() && y.rawConvertible()) {
+                            return 0;
+                        } else if (x.rawConvertible()) {
+                            return -1;
+                        } else {
+                            return 1;
+                        }
+                    });
+            // fast return if there is no rawConvertible split
```
wwj6591812, line 115: Why you can ignore pushDownLimit if there is no rawConvertible split?

paimon-core/src/main/java/org/apache/paimon/table/source/DataTableBatchScan.java
```diff
@@ -96,6 +96,27 @@ private StartingScanner.Result applyPushDownLimit(StartingScanner.Result result)
...
+            splits.sort(
+                    (x, y) -> {
+                        if (x.rawConvertible() && y.rawConvertible()) {
+                            return 0;
+                        } else if (x.rawConvertible()) {
+                            return -1;
+                        } else {
+                            return 1;
+                        }
+                    });
```
wwj6591812, line 114: Add a blank line.

wwj6591812:
I think this is a good point!

wwj6591812:
I think this is a good optimize.
Change title from "Spark" to "Core"，because Flink also use optimize.

paimon-core/src/main/java/org/apache/paimon/table/source/DataTableBatchScan.java
ulysses-you: It follows previous behavior, if all splits are not rawConvertible, then we add all splits as the final plan. The semantics of pushDownLimit is just a best effort, the engine will do limit again.

ulysses-you:
thank you @wwj6591812 , addressed comments

ulysses-you:
cc @JingsongLi thank you

paimon-core/src/main/java/org/apache/paimon/table/source/DataTableBatchScan.java
wwj6591812: OK.

JingsongLi:
+1

YannByron:
Link to https://github.com/apache/paimon/issues/2404.

PR #4306, opened 2024-10-10
[core] Sort some system table query result

askwang:
### Purpose

Sort some system table query result, including tags/partitions/buckets tables.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/system/BucketsTable.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/system/BucketsTable.java
@@ -40,6 +40,7 @@ import org.apache.paimon.types.DataField;
 import org.apache.paimon.types.DataTypes;
 import org.apache.paimon.types.RowType;
 import org.apache.paimon.utils.IteratorRecordReader;
+import org.apache.paimon.utils.Pair;
 import org.apache.paimon.utils.ProjectedRow;
 import org.apache.paimon.utils.RowDataToObjectArrayConverter;
 import org.apache.paimon.utils.SerializationUtils;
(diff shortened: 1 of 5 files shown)


wwj6591812:
Good, I think no problem, but ut failed.
@askwang

askwang:
> Good, I think no problem, but ut failed. @askwang

fixed, thanks. personal  repository ut is success

paimon-core/src/main/java/org/apache/paimon/table/system/BucketsTable.java
```diff
@@ -181,9 +184,22 @@ public RecordReader<InternalRow> createReader(Split split) {
                     new RowDataToObjectArrayConverter(
                             fileStoreTable.schema().logicalPartitionType());
 
+            // sorted by partition and bucket
+            Map<Pair<String, Integer>, BucketEntry> pairBucketsMap =
+                    new TreeMap<>(
+                            Comparator.comparing((Pair<String, Integer> o) -> o.getLeft())
+                                    .thenComparing(Pair::getRight));
+            buckets.forEach(
```
JingsongLi, line 192:
```
Function<BucketEntry, String> toPartString = e -> Arrays.toString(converter.convert(e.partition()));
buckets.sort(Comparator.comparing(toPartString).thenComparing(BucketEntry::bucket));
```

askwang: A little difference, change `List<BucketEntry> buckets` to `List<Pair<String, BucketEntry>> bucketList`, to avoid `converter.convert(e.partition())` compute twice.

JingsongLi (approved):
+1

paimon-core/src/main/java/org/apache/paimon/table/system/BucketsTable.java
```diff
@@ -181,9 +184,24 @@ public RecordReader<InternalRow> createReader(Split split) {
...
+                            .map(
+                                    entry ->
+                                            Pair.of(
+                                                    Arrays.toString(
+                                                            converter.convert(entry.partition())),
+                                                    entry))
+                            .sorted(
+                                    Comparator.comparing(
+                                                    (Pair<String, BucketEntry> p) -> p.getLeft())
+                                            .thenComparing(p -> p.getRight().bucket()))
```
jiangjin-f, line 199: .map(pair -> toRow(pair.getLeft(), pair.getRight()))

PR #4309, opened 2024-10-11
[flink] fix compact unaware bucket table in combine mode the writer parallelism is always 1

Tan-JiaLiang:
### Purpose

fix compact unaware bucket table in combine mode the writer parallelism is always 1


wwj6591812:
Good, I think no problem.
+1

JingsongLi:
+1

PR #4312, opened 2024-10-12
[flink] Flink LookupTable support cache row filter

Sxnan:
### Purpose

Flink LookupTable support cache row filter to filter out rows to be stored in cache.

### Tests

- UT added in `LookupTableTest` and `KeyValueFileReadWriteTest`.

### API and Format

No

### Documentation

No

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/FileStoreLookupFunction.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/FileStoreLookupFunction.java
@@ -33,6 +33,7 @@ import org.apache.paimon.table.Table;
 import org.apache.paimon.table.source.OutOfRangeException;
 import org.apache.paimon.types.RowType;
 import org.apache.paimon.utils.FileIOUtils;
+import org.apache.paimon.utils.Filter;
 import org.apache.paimon.utils.RowDataToObjectArrayConverter;
 
 import org.apache.paimon.shade.guava30.com.google.common.primitives.Ints;
@@ -95,6 +96,8 @@ public class FileStoreLookupFunction implements Serializable, Closeable {
     private transient long nextLoadTime;
 
     protected FunctionContext functionContext;
+
+    @Nullable private Filter<InternalRow> cacheRowFilter;
 
     public FileStoreLookupFunction(
             Table table, int[] projection, int[] joinKeyIndex, @Nullable Predicate predicate) {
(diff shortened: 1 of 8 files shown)


paimon-core/src/main/java/org/apache/paimon/io/KeyValueFileReaderFactory.java
```diff
@@ -102,11 +103,23 @@ public RecordReader<KeyValue> createRecordReader(DataFileMeta file) throws IOExc
 
     public RecordReader<KeyValue> createRecordReader(
             long schemaId, String fileName, long fileSize, int level) throws IOException {
+        return createRecordReader(schemaId, fileName, fileSize, level, null);
+    }
+
+    public RecordReader<KeyValue> createRecordReader(
```
JingsongLi, line 109: I think you don't need to modify `paimon-core`, you can do the filter outer.

Sxnan: Good point! I updated the PR so we don't apply the filter in the `KeyValueFileReaderFactory`.  

However, I think we still need to modify the `LocalTableQuery` in the `paimon-core` unless I am missing something.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/FileStoreLookupFunction.java
```diff
@@ -361,4 +367,8 @@ protected Set<Integer> getRequireCachedBucketIds() {
         // TODO: Implement the method when Flink support bucket shuffle for lookup join.
         return null;
     }
+
+    protected void setCacheRowFilter(@Nullable Filter<InternalRow> cacheRowFilter) {
```
wwj6591812, line 371: Where this method be called?

wwj6591812: What business scenarios is this feature used in?

Sxnan: This change can reduce the memory usage of the lookup join function if how the fact table rows are shuffled is known by the lookup join function. We can filter out rows that will not shuffle to this specific subtask.

For this feature to work, Flink will support specifying a custom data distribution of the fact table, please see the [FLIP](https://cwiki.apache.org/confluence/display/FLINK/FLIP-462+Support+Custom+Data+Distribution+for+Input+Stream+of+Lookup+Join).

JingsongLi (approved):
+1

PR #4318, opened 2024-10-14
[core] Support timestamp type in Iceberg compatibility

tsreaper:
### Purpose

`TIMESTAMP` is a widely used data type. This PR supports `TIMESTAMP` and `TIMESTAMP_LTZ` type in Iceberg compatibility.

### Tests

Unit tests and IT cases.

### API and Format

No format changes.

### Documentation

Document is also added.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/iceberg/manifest/IcebergConversions.java
+++ b/paimon-core/src/main/java/org/apache/paimon/iceberg/manifest/IcebergConversions.java
@@ -20,8 +20,12 @@ package org.apache.paimon.iceberg.manifest;
 
 import org.apache.paimon.data.BinaryString;
 import org.apache.paimon.data.Decimal;
+import org.apache.paimon.data.Timestamp;
 import org.apache.paimon.types.DataType;
 import org.apache.paimon.types.DecimalType;
+import org.apache.paimon.types.LocalZonedTimestampType;
+import org.apache.paimon.types.TimestampType;
+import org.apache.paimon.utils.Preconditions;
 
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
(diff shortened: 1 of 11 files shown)


paimon-core/src/main/java/org/apache/paimon/iceberg/manifest/IcebergConversions.java
```diff
@@ -112,6 +131,15 @@ public static Object toPaimonObject(DataType type, byte[] bytes) {
                 DecimalType decimalType = (DecimalType) type;
                 return Decimal.fromUnscaledBytes(
                         bytes, decimalType.getPrecision(), decimalType.getScale());
+            case TIMESTAMP_WITHOUT_TIME_ZONE:
+            case TIMESTAMP_WITH_LOCAL_TIME_ZONE:
+                int timestampPrecision = ((TimestampType) type).getPrecision();
+                long timestampLong =
```
wwj6591812, line 137: Change the order of these two statements, for example:

Preconditions.checkArgument(
                        timestampPrecision > 3 && timestampPrecision <= 6,
                        "Paimon Iceberg compatibility only support timestamp type with precision from 4 to 6.");
                long timestampLong =
                        ByteBuffer.wrap(bytes).order(ByteOrder.LITTLE_ENDIAN).getLong();

paimon-core/src/main/java/org/apache/paimon/iceberg/metadata/IcebergDataField.java
```diff
@@ -137,38 +148,23 @@ private static String toTypeString(DataType dataType) {
                 DecimalType decimalType = (DecimalType) dataType;
                 return String.format(
                         "decimal(%d, %d)", decimalType.getPrecision(), decimalType.getScale());
+            case TIMESTAMP_WITHOUT_TIME_ZONE:
+                int timestampPrecision = ((TimestampType) dataType).getPrecision();
+                Preconditions.checkArgument(
```
wwj6591812, line 153: There are four place check this precision.
Should we extract this to reduce common code?

JingsongLi (approved):
+1

PR #4321, opened 2024-10-14
[flink] Fix FlinkGenericCatalog can not get procedure

Aitozi:
### Purpose

Fix the exception as below when use `FlinkGenericCatalog`

image


JingsongLi:
+1

PR #4328, opened 2024-10-15
[core] Fix the KvQueryTableTest may stuck in loop

Aitozi:
### Purpose

Before, it use the pattern as bellow, which will lead the ConcurrentHashMap stuck in loop 

```
Map<String, String> test = new ConcurrentHashMap<>();
test.computeIfAbsent(
        "a",
        new Function<String, String>() {
            @Override
            public String apply(String s) {
                return test.remove("a");
            }
        });
System.out.println(test);
```

The `KvQueryTableTest` will stuck in the thread as below

image

### DIFF
--- a/paimon-service/paimon-service-client/src/main/java/org/apache/paimon/service/network/NetworkClient.java
+++ b/paimon-service/paimon-service-client/src/main/java/org/apache/paimon/service/network/NetworkClient.java
@@ -143,30 +143,23 @@ public class NetworkClient<REQ extends MessageBody, RESP extends MessageBody> {
                     new IllegalStateException(clientName + " is already shut down."));
         }
 
-        final ServerConnection<REQ, RESP> connection =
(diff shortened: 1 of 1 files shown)


wwj6591812:
1、Does this must appear every time?
2、This is a bug in JDK?

Aitozi:
> 1、Does this must appear every time? 2、This is a bug in JDK?

```
Map<String, String> test = new ConcurrentHashMap<>();
test.computeIfAbsent(
        "a",
        new Function<String, String>() {
            @Override
            public String apply(String s) {
                return test.remove("a");
            }
        });
System.out.println(test);
```

In jdk8 it will stuck, in jdk11 it will throw `IllegalStageException`. I think it's a wrong use case. So, we should avoid to modify the map in the `computeIfAbsent` call.

JingsongLi:
+1

PR #4331, opened 2024-10-16
[core] Rollback action write hint first 

yuzelin:
### Purpose

If we start a job when Rollback is running, it may read latest snapshot id from hint file and get wrong snapshot id.
This PR fix it.


wwj6591812:
+1

JingsongLi (commented):
Please write description.

JingsongLi:
+1

PR #4323, opened 2024-10-14
[core] Support add partition for paimon

xuzifu666:
### Purpose
Currently paimon can not support add partition, due to add partition not need commit transaction，so make hms sync to achieve it.

Linked issue: close #xxx

PR #4334, opened 2024-10-16
[core] Adjust create partition interface

JingsongLi:
### Purpose

Adjust https://github.com/apache/paimon/pull/4323


xuzifu666:
+1 look good to me！thanks @JingsongLi

PR #4336, opened 2024-10-16
[doc] Add read performance doc.

wwj6591812:
### Purpose

Recently, when I read Paimon data through Flink batch job, I found that the job run time is often very long and often OOM.
Lucky, I have solved by adjust some Paimon parameters. 
But in this process, I found that the Paimon lacks a doc which describes how to speed up reading. 
So I initialized a doc related to reading performance, hoping to help more Paimon users, and also welcome more friends in Paimon community to supplement this doc.

Linked issue: close #xxx

### DIFF
--- a/docs/content/maintenance/read-performance.md
+++ b/docs/content/maintenance/read-performance.md
@@ -1,0 +1,68 @@
+---
+title: "Read Performance"
+weight: 2
+type: docs
+aliases:
+- /maintenance/read-performance.html
+---
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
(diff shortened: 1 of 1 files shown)


docs/content/maintenance/read-performance.md
```diff
@@ -0,0 +1,68 @@
...
+specific language governing permissions and limitations
+under the License.
+-->
+
+# Read Performance
+
+There are many ways to improve read performance when reading Paimon data with Flink.
+
+## Acceleration In Flink client
+When use Flink batch job scan Paimon table, if you don't add option `scan.parallelism`, Paimon will infer the parallelism by read manifest files which take lots of time. So you can read Paimon table with the `scan.parallelism` table property.
```
JingsongLi, line 32: I don't think that for most tables it will cause "take lots of time".

docs/content/maintenance/read-performance.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Read Compact Snapshot
+You can read snapshot which commitKind = Compact, Compared to read other commitKind snapshot, reading this snapshot requires fewer merge operations, resulting in higher performance.
```
JingsongLi, line 35: It may only be faster if there is a configuration full-compact.

wwj6591812:
Close this PR.

PR #4340, opened 2024-10-16
[core] Introduce View Support to HiveCatalog

JingsongLi:
### Purpose

This PR aims to support View in Catalog, our metastore behind us, such as Hive metastore, supports views, but our catalog has not done this conversion well. This PR aims to support the definition of views and bridge with HiveCatalog.

#### Definition of view

```
/** Interface for view definition. */
public interface View {

    /** A name to identify this view. */
    String name();

    /** Full name (including database) to identify this view. */
    String fullName();

    /** Returns the row type of this view. */
    RowType rowType();

    /** Returns the view representation. */
    String query();

    /** Optional comment of this view. */
    Optional<String> comment();

    /** Options of this view. */
    Map<String, String> options();

    /** Copy this view with adding dynamic options. */
    View copy(Map<String, String> dynamicOptions);
}
```

#### Catalog interfaces

```
/**
     * Return a {@link View} identified by the given {@link Identifier}.
     *
     * @param identifier Path of the view
     * @return The requested view
     * @throws ViewNotExistException if the target does not exist
     */
    default View getView(Identifier identifier) throws ViewNotExistException {
        throw new ViewNotExistException(identifier);
    }

    /**
     * Drop a view.
     *
     * @param identifier Path of the view to be dropped
     * @param ignoreIfNotExists Flag to specify behavior when the view does not exist: if set to
     *     false, throw an exception, if set to true, do nothing.
     * @throws ViewNotExistException if the view does not exist
     */
    default void dropView(Identifier identifier, boolean ignoreIfNotExists)
            throws ViewNotExistException {
    }

    /**
     * Create a new view.
     *
     * @param identifier path of the view to be created
     * @param view the view definition
     * @param ignoreIfExists flag to specify behavior when a view already exists at the given path:
     *     if set to false, it throws a ViewAlreadyExistException, if set to true, do nothing.
     * @throws ViewAlreadyExistException if view already exists and ignoreIfExists is false
     * @throws DatabaseNotExistException if the database in identifier doesn't exist
     */
    default void createView(Identifier identifier, View view, boolean ignoreIfExists)
            throws ViewAlreadyExistException, DatabaseNotExistException {
    }

    /**
     * Get names of all views under this database. An empty list is returned if none exists.
     *
     * @return a list of the names of all views in this database
     * @throws DatabaseNotExistException if the database does not exist
     */
    default List<String> listViews(String databaseName) throws DatabaseNotExistException {
    }
```

Implemented these methods in `HiveCatalog`.

#### Flink Integration

Integrate to Flink Catalog.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/view/ViewImpl.java
+++ b/paimon-core/src/main/java/org/apache/paimon/view/ViewImpl.java
@@ -1,0 +1,110 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.paimon.view;
+
+import org.apache.paimon.catalog.Identifier;
+import org.apache.paimon.types.RowType;
+
+import javax.annotation.Nullable;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+
+/** Implementation of {@link View}. */
+public class ViewImpl implements View {
+
+    private final Identifier identifier;
+    private final RowType rowType;
+    private final String query;
+    @Nullable private final String comment;
+    private final Map<String, String> options;
+
+    public ViewImpl(
+            Identifier identifier,
+            RowType rowType,
+            String query,
+            @Nullable String comment,
+            Map<String, String> options) {
+        this.identifier = identifier;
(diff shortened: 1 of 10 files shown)


adrian-wang:
maybe also support `listViews` and `alterView`?

paimon-core/src/main/java/org/apache/paimon/view/ViewImpl.java
```diff
@@ -0,0 +1,106 @@
...
+    @Override
+    public View copy(Map<String, String> dynamicOptions) {
+        Map<String, String> newOptions = new HashMap<>(options);
+        newOptions.putAll(dynamicOptions);
+        return new ViewImpl(identifier, rowType, query, comment, newOptions);
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
```
wwj6591812, line 92: NeedBraces: 'if' construct must use '{}'s.

paimon-core/src/main/java/org/apache/paimon/view/ViewImpl.java
```diff
@@ -0,0 +1,106 @@
...
+        if (o == null || getClass() != o.getClass()) return false;
```
wwj6591812, line 93: NeedBraces: 'if' construct must use '{}'s.

JingsongLi:
> maybe also support `listViews` and `alterView`?

Alter view looks invalid to me, it just modify properties, I cannot see the case.

LinMingQiang:
Is there any problem with this PR? I am looking forward to this feature.

adrian-wang:
LGTM

JingsongLi:
> Is there any problem with this PR? I am looking forward to this feature.

Let's merge this.

PR #4343, opened 2024-10-17
[hive] Support syncing Iceberg compatible metadata to Hive, so Paimon table can be accessed from Iceberg Hive catalog

tsreaper:
### Purpose

Iceberg Hive catalog is the most widely used Iceberg catalog. This PR supports syncing Iceberg compatible metadata to Hive, so Paimon table can be accessed from Iceberg Hive catalog.

### Tests

IT cases.

### API and Format

No format changes.

### Documentation

Document is also added.


JingsongLi (approved):
+1

PR #4356, opened 2024-10-21
[Bug] Fix NoSuchElementException when enable partition mark done

Aitozi:
### Purpose

image

when enable the partition mark done, the streaming job will fail with `NoSuchElementException`.


JingsongLi (approved):
+1

PR #4357, opened 2024-10-22
[Bug] Close PartitionMarkDone to prevent potential resource leakage

Aitozi:
### Purpose

During closing, we should also close partition mark done to release the resource such as metastore client in MarkDoneAction.


JingsongLi:
+1

PR #4360, opened 2024-10-22
[hive] Support create formatTable set csv.field-delimiter

herefree:
### Purpose
I found that when using flink to create a format table, flink will throw an exception when reading the format table. This is due to the fact that the field-delimiter of the hive table is not specified. This pr is to support setting csv.field-delimiter.

Linked issue: close #xxx

### DIFF
--- a/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveFormatTableUtils.java
+++ b/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveFormatTableUtils.java
@@ -58,11 +58,13 @@ class HiveFormatTableUtils {
         String comment = options.remove(COMMENT_PROP);
         String location = hiveTable.getSd().getLocation();
         Format format;
+        SerDeInfo serdeInfo = hiveTable.getSd().getSerdeInfo();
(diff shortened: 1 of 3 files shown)


paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveFormatTableUtils.java
```diff
@@ -58,11 +58,15 @@ public static FormatTable convertToFormatTable(Table hiveTable) {
...
         String location = hiveTable.getSd().getLocation();
         Format format;
+        SerDeInfo serdeInfo = hiveTable.getSd().getSerdeInfo();
         if (Options.fromMap(options).get(TYPE) == FORMAT_TABLE) {
             format = Format.valueOf(options.get(FILE_FORMAT.key()).toUpperCase());
-            // field delimiter for csv leaves untouched
+            if (format.equals(Format.CSV)) {
+                options.put(
+                        FIELD_DELIMITER.key(),
+                        serdeInfo.getParameters().getOrDefault(FIELD_DELIM, "\u0001"));
```
wwj6591812, line 67: serdeInfo.getParameters() may null?

JingsongLi (approved):
+1

PR #4362, opened 2024-10-22
[parquet] Add type id to parquet files

tsreaper:
### Purpose

Currently parquet files produced by Paimon have no type id, however other compute engine (for example, Trino) or lake format (for example, Iceberg) rely on type id to project columns.

This PR adds type id to parquet files.

### Tests

Unit tests.

### API and Format

Yes. Special field ids are assigned to array elements and map keys / values. Also type ids are added to parquet files.

### Documentation

No new feature.


paimon-format/src/main/java/org/apache/paimon/format/parquet/ParquetSchemaConverter.java
```diff
@@ -197,8 +196,47 @@ private static Type createTimestampWithLogicalType(
...
+            if (field.type().getTypeRoot() == DataTypeRoot.ARRAY) {
+                GroupType groupType = (GroupType) parquetType;
+                GroupType wrapperType = (GroupType) groupType.getFields().get(0);
+                Type elementTypeWithId =
+                        wrapperType
+                                .getFields()
+                                .get(0)
+                                .withId(SpecialFields.getArrayElementFieldId(field.id()));
+                typeWithId =
+                        ConversionPatterns.listOfElements(
```
JingsongLi, line 211: Can we just use `convertToParquetType(String name, DataType type, Type.Repetition repetition)`?

It is not good to use `ConversionPatterns.listOfElements` twice.

JingsongLi (commented):
I just worry about StarRocks, can you verify that new table can be queried by StarRocks?

paimon-format/src/main/java/org/apache/paimon/format/parquet/ParquetSchemaConverter.java
tsreaper: We need to set id for element type here. By calling `convertToParquetType` we can only get an array type, whose element type has no id.

PR #4373, opened 2024-10-24
[core] Check snapshot when reset consumer

xuzifu666:
### Purpose

Linked issue: close #xxx
Currently ResetConsumerProcedure not check for snapshot id， if user set a snapshot id which is large than the latest snapshot id （it would wait for a long time, and if small than earlist snapshot id is ok which would consumer from the latest snapshot id），this pr is aim to check it if set a not exisit large snapshot id.


paimon-flink/paimon-flink-1.18/src/main/java/org/apache/paimon/flink/procedure/ResetConsumerProcedure.java
```diff
@@ -49,6 +51,15 @@ public String[] call(
             throws Catalog.TableNotExistException {
         FileStoreTable fileStoreTable =
                 (FileStoreTable) catalog.getTable(Identifier.fromString(tableId));
+        SnapshotManager snapshotManager = fileStoreTable.snapshotManager();
+        Long latestSnapshotId = snapshotManager.latestSnapshotId();
+        if (latestSnapshotId != null && nextSnapshotId > latestSnapshotId) {
+            throw new SnapshotNotExistException(
+                    String.format(
+                            "the snapshot id is not exist, you can set it <= %s.",
```
LinMingQiang, line 59: The specified snapshot id %s should be smaller than the latest snapshot id %s.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/ResetConsumerAction.java
```diff
@@ -49,6 +51,17 @@ public ResetConsumerAction withNextSnapshotIds(Long nextSnapshotId) {
     @Override
     public void run() throws Exception {
         FileStoreTable dataTable = (FileStoreTable) table;
+        SnapshotManager snapshotManager = dataTable.snapshotManager();
+        Long latestSnapshotId = snapshotManager.latestSnapshotId();
+        if (nextSnapshotId != null
```
wwj6591812, line 56: Add a public static method, Otherwise, Otherwise, this check code would appear in three different places.

paimon-flink/paimon-flink-1.18/src/main/java/org/apache/paimon/flink/procedure/ResetConsumerProcedure.java
```diff
@@ -49,6 +51,15 @@ public String[] call(
             throws Catalog.TableNotExistException {
         FileStoreTable fileStoreTable =
                 (FileStoreTable) catalog.getTable(Identifier.fromString(tableId));
+        SnapshotManager snapshotManager = fileStoreTable.snapshotManager();
```
JingsongLi, line 54: Do not copy codes 4 times.

paimon-flink/paimon-flink-1.18/src/main/java/org/apache/paimon/flink/procedure/ResetConsumerProcedure.java
```diff
@@ -49,6 +51,15 @@ public String[] call(
...
+        Long latestSnapshotId = snapshotManager.latestSnapshotId();
+        if (latestSnapshotId != null && nextSnapshotId > latestSnapshotId) {
```
JingsongLi, line 56: You can just read the snapshot.

JingsongLi (changes requested):

JingsongLi (approved):
+1

PR #4376, opened 2024-10-25
[core] Fix snapshot privilege check logic

yunfengzhou-hub:
### Purpose

Fix snapshot privilege check logic

### Tests

Introduced `org.apache.paimon.privilege.PrivilegedCatalogTest` to verify the changes.

### API and Format

This change does not affect API or storage format

### Documentation

This change does not introduce a new feature

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/privilege/PrivilegeChecker.java
+++ b/paimon-core/src/main/java/org/apache/paimon/privilege/PrivilegeChecker.java
@@ -24,6 +24,22 @@ import java.io.Serializable;
 
 /** Check if current user has privilege to perform related operations. */
 public interface PrivilegeChecker extends Serializable {
+    default void assertCanSelectOrInsert(Identifier identifier) {
+        try {
+            assertCanSelect(identifier);
+        } catch (NoPrivilegeException e) {
+            try {
+                assertCanInsert(identifier);
+            } catch (NoPrivilegeException e1) {
+                throw new NoPrivilegeException(
+                        e1.getUser(),
+                        e1.getObjectType(),
+                        e1.getIdentifier(),
+                        PrivilegeType.SELECT,
+                        PrivilegeType.INSERT);
+            }
+        }
+    }
 
     void assertCanSelect(Identifier identifier);
 
--- a/paimon-core/src/test/java/org/apache/paimon/privilege/PrivilegedCatalogTest.java
+++ b/paimon-core/src/test/java/org/apache/paimon/privilege/PrivilegedCatalogTest.java
@@ -1,0 +1,91 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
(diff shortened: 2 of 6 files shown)


paimon-core/src/main/java/org/apache/paimon/privilege/PrivilegeChecker.java
```diff
@@ -24,6 +24,18 @@
 
 /** Check if current user has privilege to perform related operations. */
 public interface PrivilegeChecker extends Serializable {
+    default void assertCanSelectOrInsert(Identifier identifier) {
+        try {
+            assertCanSelect(identifier);
+        } catch (Exception e) {
```
tsreaper, line 30: Catch `NoPrivilegeException` only.

paimon-core/src/main/java/org/apache/paimon/privilege/PrivilegeChecker.java
```diff
@@ -24,6 +24,18 @@
...
+            try {
+                assertCanInsert(identifier);
+            } catch (Exception e1) {
+                throw new RuntimeException(
+                        "Need privilege to select or insert to table " + identifier, e);
```
tsreaper, line 35: Add `e1` as cause. Add `e` as suppressed exception.

paimon-core/src/main/java/org/apache/paimon/privilege/PrivilegeChecker.java
```diff
@@ -24,6 +24,25 @@
...
+            } catch (NoPrivilegeException e1) {
+                NoPrivilegeException exception =
+                        new NoPrivilegeException(
+                                e1.getUser(),
+                                e1.getObjectType(),
+                                e1.getIdentifier(),
+                                PrivilegeType.SELECT,
+                                PrivilegeType.INSERT);
+                exception.addSuppressed(e);
+                throw exception;
```
tsreaper, line 42:
```suggestion
                throw new NoPrivilegeException(
                        e1.getUser(),
                        e1.getObjectType(),
                        e1.getIdentifier(),
                        PrivilegeType.SELECT,
                        PrivilegeType.INSERT);
```

Now you only catch `NoProvilegeException`, so the cause of this exception is clear. No need to add suppressed or cause.

paimon-core/src/test/java/org/apache/paimon/privilege/PrivilegedCatalogTest.java
```diff
@@ -0,0 +1,117 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/** Tests for {@link PrivilegedCatalog}. */
+public class PrivilegedCatalogTest extends FileSystemCatalogTest {
```
tsreaper, line 34: Add tests in `PrivilegeProcedureITCase`. No need to create a new test class.

yunfengzhou-hub: PrivilegeProcedureITCase has tests that covered `HashBucketAssignerOperator#initializeState`, where `snapshotManager()` is verified.

I discovered this bug during some direct invocation of Catalog methods, so this test to verify Catalog interface is also needed.

PR #4378, opened 2024-10-25
[cdc] Add flink execution mode check for cdc table sync action.

LinMingQiang:
### Purpose

Linked issue: close #xxx

Flink-cdc do not support batch mode, including flink cdc `snapshot` scan mode . Although it is a bounded source.

https://issues.apache.org/jira/browse/FLINK-36280

https://github.com/apache/flink-cdc/pull/2901

if use set RuntimeExecutionMode = Batch. job will be never finish and never fail.

image

so , we need to check the flink execution mode.


LinMingQiang:
If this change is acceptable, I will complete the other actions / mongodb/Postgres.

wwj6591812:
Do we need to handle RuntimeExecutionMode.AUTOMATIC?

wwj6591812:
Do we need to handle RuntimeExecutionMode.AUTOMATIC?

LinMingQiang:
> Do we need to handle RuntimeExecutionMode.AUTOMATIC?

yes need too, may be we can make STREAMING mode as default and ignore the user's settings.

LinMingQiang:
@yuzelin Help me check this pr?  tks.

JingsongLi (approved):
+1

PR #4381, opened 2024-10-25
[flink] fix false assumption on catalog table passing logic

### LINKED ISSUE #4372: [Bug] Assumption on the Consistency of Flink CatalogTable is Vulnerable
### Search before asking

### Paimon version

87321e13277bb32e909f4949bb0358eaffcf00e0

### Compute Engine

Flink

### Minimal reproduce step

Create a MaterializeTable and try to insert some data to it. When the code enters `FlinkTableFactory#createDynamicTableSink`, check `context.getCatalogTable().getOrigin()`.

The expected value in `context.getCatalogTable().getOrigin()` should be a `CatalogMaterializedTable`, while actually it is a `CatalogTable`.

### Anything else?

The cause of the unexpected behavior above is as follows: When creating a Flink job that inserts into Paimon(Reading from Paimon has the same problem), Flink will first invoke PaimonCatalog's `getTable` method to acquire certain metadata about the table, and then invoke `FlinkTableFactory#createDynamicTableSink` to create a flink sink. Suppose the table that Paimon returned in `getTable` is tableA and the table passed into `FlinkTableFactory#createDynamicTableSink` is tableB, Paimon has wrongly assumed that tableB == tableA or tableB == tableA.copy(...). However, Flink Catalog and TableFactory API has not guaranteed this assumption. During the optimization process of Flink SQL Planner, Flink might create another table and pass it to PaimonCatalog. Flink is not doing it wrong so long as the behaviors of the public APIs(CatalogTable) does not change.

This false assumption resulted in the following codes. You can see that for a table acquired through `context.getCatalogTable().getOrigin()`, Paimon assumes that the table must be a certain child class of CatalogBaseTable.

```java
// in AbstractFlinkTableFactory#createDynamicTableSource

if (origin instanceof SystemCatalogTable) {
    return new SystemTableSource(
            ((SystemCatalogTable) origin).table(),
            isStreamingMode,
            context.getObjectIdentifier());
} else {
```

```java
// in AbstractFlinkTableFactory#buildPaimonTable

if (origin instanceof DataCatalogTable) {
    FileStoreTable fileStoreTable = (FileStoreTable) ((DataCatalogTable) origin).table();
    table = fileStoreTable.copyWithoutTimeTravel(newOptions);
} else {
    table =
            FileStoreTableFactory.create(createCatalogContext(context))
                    .copyWithoutTimeTravel(newOptions);
}
```

Thus it is necessary to get these implementations rid of the false assumption to avoid potential correctness problems.

yunfengzhou-hub:
### Purpose

Linked issue: close #4372 

### Tests
 
The change is verified by FlinkCatalogTest.

### API and Format

This change does not affect API or Format.

### Documentation

This change does not introduce a new feature.

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/AbstractFlinkTableFactory.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/AbstractFlinkTableFactory.java
@@ -22,7 +22,9 @@ import org.apache.paimon.CoreOptions.LogChangelogMode;
 import org.apache.paimon.CoreOptions.LogConsistency;
 import org.apache.paimon.CoreOptions.StreamingReadMode;
 import org.apache.paimon.annotation.VisibleForTesting;
+import org.apache.paimon.catalog.Catalog;
 import org.apache.paimon.catalog.CatalogContext;
+import org.apache.paimon.catalog.Identifier;
 import org.apache.paimon.data.Timestamp;
 import org.apache.paimon.flink.log.LogStoreTableFactory;
 import org.apache.paimon.flink.sink.FlinkTableSink;
@@ -88,6 +90,12 @@ public abstract class AbstractFlinkTableFactory
         implements DynamicTableSourceFactory, DynamicTableSinkFactory {
 
     private static final Logger LOG = LoggerFactory.getLogger(AbstractFlinkTableFactory.class);
+
+    @Nullable private final FlinkCatalog flinkCatalog;
+
+    public AbstractFlinkTableFactory(@Nullable FlinkCatalog flinkCatalog) {
+        this.flinkCatalog = flinkCatalog;
+    }
 
     @Override
     public DynamicTableSource createDynamicTableSource(Context context) {
@@ -227,7 +235,7 @@ public abstract class AbstractFlinkTableFactory
                 Options.fromMap(context.getCatalogTable().getOptions()), new FlinkFileIOLoader());
     }
 
-    static Table buildPaimonTable(DynamicTableFactory.Context context) {
+    Table buildPaimonTable(DynamicTableFactory.Context context) {
         CatalogTable origin = context.getCatalogTable().getOrigin();
         Table table;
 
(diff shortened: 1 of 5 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/AbstractFlinkTableFactory.java
```diff
@@ -243,16 +257,34 @@ static Table buildPaimonTable(DynamicTableFactory.Context context) {
...
         newOptions.putAll(dynamicOptions);
 
-        // notice that the Paimon table schema must be the same with the Flink's
-        if (origin instanceof DataCatalogTable) {
-            FileStoreTable fileStoreTable = (FileStoreTable) ((DataCatalogTable) origin).table();
-            table = fileStoreTable.copyWithoutTimeTravel(newOptions);
+        FileStoreTable fileStoreTable;
+
+        // The following if conditions provide a shortcut to acquire Paimon table.
+        if (origin instanceof FormatCatalogTable) {
```
JingsongLi, line 263: It should not be a `FormatCatalogTable`.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/AbstractFlinkTableFactory.java
```diff
@@ -243,16 +257,34 @@ static Table buildPaimonTable(DynamicTableFactory.Context context) {
...
-                    FileStoreTableFactory.create(createCatalogContext(context))
-                            .copyWithoutTimeTravel(newOptions);
+            // In case the shortcut is not matched, the paimon table can still be acquired from
+            // catalog.
+            Identifier identifier =
+                    Identifier.create(
+                            context.getObjectIdentifier().getDatabaseName(),
+                            context.getObjectIdentifier().getObjectName());
+            try {
+                fileStoreTable = (FileStoreTable) flinkCatalog.catalog().getTable(identifier);
```
JingsongLi, line 279: What do you want to solve? This should be a `CatalogMaterializedTable`?

yunfengzhou-hub: The return type of `DynamicTableFactory.Context#getCatalogTable().getOrigin()` is `CatalogTable`, and `CatalogMaterializedTable` is not assignable from `CatalogTable`.

The problem I want to solve is that in the following code 
```java
if (origin instanceof DataCatalogTable) {
    FileStoreTable fileStoreTable = (FileStoreTable) ((DataCatalogTable) origin).table();
    table = fileStoreTable.copyWithoutTimeTravel(newOptions);
} else {
    table =
            FileStoreTableFactory.create(createCatalogContext(context))
                    .copyWithoutTimeTravel(newOptions);
}
```
Flink Paimon catalog assumed that if the origin table is not a `DataCatalogTable`, then it must support `FileStoreTableFactory#create` (i.e. the scheme of the origin table's path url can be matched with a FileIOLoader). While in practice I encountered a custom FileIO that does not support SPI dynamic loading, and thus would throw exceptions when running into `FileStoreTableFactory#create`.

JingsongLi:
+1

PR #4383, opened 2024-10-25
[flink] Lookup source should ignore scan options

yuzelin:
### Purpose

`lookup.dynamic-partition` will create `DynamicPartitionLoader`, it will use scan options to load data. But dim table should ignore scan options.


wwj6591812:
Please add a description to this PR. Thx
Why submit this PR and what is the background?

JingsongLi:
We need 1. Description 2. Test.

tsreaper (approved):

PR #4385, opened 2024-10-26
[core] Support sequence.field sort order for pk table

xuzifu666:
### Purpose
In production such as compute ealist login time zipper table，user want to get the small record to query when set sequence.field, currently paimon can only get the largest record, user need (Long.max - timestamp) which is not very convenient. This pr is aimed to support user set sequence.field order to get largest or small record with a option sequence.field.sort.is.ascending， default true keep the largest one，with false keep the small one.
eg like ut case.

Linked issue: close #xxx

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
@@ -610,11 +610,11 @@ public class CoreOptions implements Serializable {
                                     + " the sequence number determines which data is the most recent.");
 
     @Immutable
-    public static final ConfigOption<Boolean> SEQUENCE_FIELD_SORT_IS_ASCENDING =
-            key("sequence.field.sort.is.ascending")
-                    .booleanType()
-                    .defaultValue(true)
-                    .withDescription("Specify the order of sequence.field is ascending.");
+    public static final ConfigOption<SortOrder> SEQUENCE_FIELD_SORT_ORDER =
+            key("sequence.field.sort-order")
+                    .enumType(SortOrder.class)
+                    .defaultValue(SortOrder.ASCENDING)
+                    .withDescription("Specify the order of sequence.field.");
 
     @Immutable
     public static final ConfigOption<Boolean> PARTIAL_UPDATE_REMOVE_RECORD_ON_DELETE =
@@ -2050,8 +2050,8 @@ public class CoreOptions implements Serializable {
                 .orElse(Collections.emptyList());
     }
 
-    public Boolean sequenceFieldSortOrder() {
-        return options.get(SEQUENCE_FIELD_SORT_IS_ASCENDING);
+    public boolean sequenceFieldSortOrderIsAscending() {
+        return options.get(SEQUENCE_FIELD_SORT_ORDER) == SortOrder.ASCENDING;
     }
 
     public boolean partialUpdateRemoveRecordOnDelete() {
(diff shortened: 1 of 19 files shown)


paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -609,6 +609,13 @@ public class CoreOptions implements Serializable {
                             "The field that generates the sequence number for primary key table,"
                                     + " the sequence number determines which data is the most recent.");
 
+    @Immutable
+    public static final ConfigOption<Boolean> SEQUENCE_FIELD_SORT_IS_ASCENDING =
+            key("sequence.field.sort.is.ascending")
```
JingsongLi, line 614: sequence.field.sort-order

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -609,6 +609,13 @@ public class CoreOptions implements Serializable {
                             "The field that generates the sequence number for primary key table,"
                                     + " the sequence number determines which data is the most recent.");
 
+    @Immutable
+    public static final ConfigOption<Boolean> SEQUENCE_FIELD_SORT_IS_ASCENDING =
```
JingsongLi, line 613: Introduce SortOrder:
ASCENDING, DESCENDING

paimon-core/src/main/java/org/apache/paimon/sort/BinaryExternalSortBuffer.java
```diff
@@ -120,7 +120,7 @@ public static BinaryExternalSortBuffer create(
             int maxNumFileHandles,
             CompressOptions compression,
             MemorySize maxDiskSize) {
-        RecordComparator comparator = newRecordComparator(rowType.getFieldTypes(), keyFields);
+        RecordComparator comparator = newRecordComparator(rowType.getFieldTypes(), keyFields, true);
```
JingsongLi, line 123: Why here pass true?

paimon-core/src/main/java/org/apache/paimon/operation/MergeFileSplitRead.java
```diff
@@ -338,6 +338,6 @@ private RecordReader<KeyValue> projectOuter(RecordReader<KeyValue> reader) {
     @Nullable
     public UserDefinedSeqComparator createUdsComparator() {
         return UserDefinedSeqComparator.create(
-                readerFactoryBuilder.readValueType(), sequenceFields);
+                readerFactoryBuilder.readValueType(), sequenceFields, true);
```
JingsongLi, line 341: Why here pass true?

JingsongLi (changes requested):

paimon-core/src/main/java/org/apache/paimon/sort/BinaryExternalSortBuffer.java
xuzifu666: Yes, here should keep the same with options. Had change it.

paimon-core/src/main/java/org/apache/paimon/operation/MergeFileSplitRead.java
xuzifu666: the same as upper. change it as options. @JingsongLi

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
xuzifu666: Done

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
xuzifu666: done

xuzifu666:
Addressed, PTAL @JingsongLi

JingsongLi (approved):
+1

PR #4386, opened 2024-10-28
[flink] PartitionMarkDone enables the use of various partition trigge…

Aitozi:
…r strategies.

### Purpose

Linked issue: close #xxx

In our company, we have encountered an issue with the HMS partition statistic being incorrect. This is because during the writing process, we only update the metastore partition when it is first written to.
Therefore, we would like to implement the PartitionMarkDone strategy to update the statistics in HMS after a short idle period for each partition.
We need a separate configuration for PartitionMarkDone due to differing requirements:
- We do not want to rely on extracting accurate timestamps from partition names.
- We aim to store each partition's last access time in state so that it can be reported back to HMS.
To achieve this, we plan on extending `PartitionMarkDone` functionality to support different trigger strategies for partitions. Following this extension, we will be able to introduce custom triggers such as `PartitionHmsReporterTrigger`.


Aitozi:
WDYT? @JingsongLi

JingsongLi:
Hi @Aitozi , can you explain this pr from API level? Or you are modifying current behavior?

Aitozi:
@JingsongLi The behavior is not change. In this PR, I extract the `PartitionTrigger` interface, The default is `PartitionMarkDoneTrigger`.

In `StoreCommitter`, it acts the `PartitionCollector`, it can trigger based on different config such as `partition-mark-done` or `hms-report`

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/partition/PartitionMarkDoneTrigger.java
```diff
@@ -65,21 +78,30 @@ public class PartitionMarkDoneTrigger {
...
+    private final boolean waitCompaction;
+    private final InternalRowPartitionComputer partitionComputer;
 
     public PartitionMarkDoneTrigger(
             State state,
             PartitionTimeExtractor timeExtractor,
             @Nullable Duration timeInterval,
             @Nullable Duration idleTime,
-            boolean markDoneWhenEndInput)
+            boolean markDoneWhenEndInput,
```
JingsongLi, line 90: Can you keep `PartitionMarkDone` as it is?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/partition/PartitionCollector.java
```diff
@@ -0,0 +1,71 @@
...
+
+import org.apache.flink.api.common.state.OperatorStateStore;
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+/** Partition collector. */
+public class PartitionCollector implements Closeable {
```
JingsongLi, line 33: `PartitionListeners`?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/partition/PartitionTrigger.java
```diff
@@ -0,0 +1,32 @@
...
+
+package org.apache.paimon.flink.sink.partition;
+
+import org.apache.paimon.manifest.ManifestCommittable;
+
+import java.io.Closeable;
+import java.util.List;
+
+/** The partition trigger. */
+public interface PartitionTrigger extends Closeable {
```
JingsongLi, line 27: `PartitionListener`?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/partition/PartitionMarkDoneTrigger.java
Aitozi: I reset the commit, and updated in https://github.com/apache/paimon/pull/4398

PR #4283, opened 2024-10-06
[Hive] Add AWS Glue Catalog Support #4282

Moonlight-CL:
### Purpose

Linked issue: close #4282

### Tests

Existing tests should cover this change.

PR #4387, opened 2024-10-28
[hive] Fix wrong construct for dlf metastore

mxdzs0612:
### Purpose

An enhancement introduced by #4283 lead to the following problem for dlf catalog:
```
Failed to create the desired metastore client (class name: com.aliyun.datalake.metastore.hive3.ProxyMetaStoreClient)
```

Because dlf hive3 ProxyMetaStoreClient still uses hive2 construct method.


JingsongLi:
Thanks for the testing!

JingsongLi (approved):
+1

PR #4390, opened 2024-10-28
[spark]: clean up zstd-jni lib files

zhongyujiang:
### Purpose

Linked issue: close #xxx

we already excluded zstd-jni jar from spark runtime jar(#4085), however, the zstd-jni jar contains not only Java files but also references to some native libraries. These files should also be excluded.

image


zhongyujiang:
cc @JingsongLi @Zouxxyy @ulysses-you

JingsongLi:
Thanks @zhongyujiang , +1

ulysses-you:
+1

Issue #3941, opened 2024-08-12
[Feature] [Spark] Remove support for Spark 3.1


YannByron:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Motivation

Paimon has plan to support Spark4.0. To avoid to maintain too many spark versions, We decide to remove the support for spark 3.1. 

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

YannByron:
Do this once paimon 0.9 is released.

PR #4395, opened 2024-10-29
[Flink] Support speculative execution when batch read paimon table.

wwj6591812:
### Purpose

image

In my company's production job, when batch read a paimon table, I receive the exception.
After read flink and paimon code, I know it means paimon use custom SourceEvent ReaderConsumeProgressEvent.
I have come up with two solutions:
1、Implement the interface SupportsHandleExecutionAttemptSourceEvent according to the exception suggestion;
2、Delete ReaderConsumeProgressEvent in StaticFileStoreSplitEnumerator.

Due to the possibility of new other SourceEvent being added in the future, so I chose the solutions-1.

Linked issue: close #xxx

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/StaticFileStoreSplitEnumerator.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/StaticFileStoreSplitEnumerator.java
@@ -28,6 +28,7 @@ import org.apache.flink.api.connector.source.SourceEvent;
 import org.apache.flink.api.connector.source.SplitEnumerator;
 import org.apache.flink.api.connector.source.SplitEnumeratorContext;
 import org.apache.flink.api.connector.source.SplitsAssignment;
+import org.apache.flink.api.connector.source.SupportsHandleExecutionAttemptSourceEvent;
 import org.apache.flink.table.connector.source.DynamicFilteringEvent;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -41,7 +42,8 @@ import static org.apache.paimon.utils.Preconditions.checkNotNull;
 
 /** A {@link SplitEnumerator} implementation for {@link StaticFileStoreSource} input. */
 public class StaticFileStoreSplitEnumerator
-        implements SplitEnumerator<FileStoreSourceSplit, PendingSplitsCheckpoint> {
+        implements SplitEnumerator<FileStoreSourceSplit, PendingSplitsCheckpoint>,
+                SupportsHandleExecutionAttemptSourceEvent {
 
     private static final Logger LOG = LoggerFactory.getLogger(StaticFileStoreSplitEnumerator.class);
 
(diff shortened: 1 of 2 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/StaticFileStoreSplitEnumerator.java
```diff
@@ -41,7 +42,8 @@
 
 /** A {@link SplitEnumerator} implementation for {@link StaticFileStoreSource} input. */
 public class StaticFileStoreSplitEnumerator
-        implements SplitEnumerator<FileStoreSourceSplit, PendingSplitsCheckpoint> {
+        implements SplitEnumerator<FileStoreSourceSplit, PendingSplitsCheckpoint>,
+                SupportsHandleExecutionAttemptSourceEvent {
```
JingsongLi, line 46: You may need to introduce classes for Flink 1.15.

wwj6591812: Could we remove 1.15 like remove spark-3.1(https://github.com/apache/paimon/issues/3941)?

We also has plan to support Flink-2.0, to avoid to maintain too many Flink versions, could we remove the support for Flink-1.15?

@JingsongLi

JingsongLi: No, it is good to keep Flink 1.15 support.

wwj6591812: OK

JingsongLi (commented):
A simpler solution for 1.15, you can just create a class `org.apache.flink.api.connector.source.SupportsHandleExecutionAttemptSourceEvent` in paimon-flink-1.15.

wwj6591812:
> A simpler solution for 1.15, you can just create a class `org.apache.flink.api.connector.source.SupportsHandleExecutionAttemptSourceEvent` in paimon-[flink-1](https://issues.apache.org/jira/browse/FLINK-1).15.

Good idea, done.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/StaticFileStoreSplitEnumerator.java
```diff
@@ -118,6 +120,12 @@ public Snapshot snapshot() {
...
 
+    @Override
+    public void handleSourceEvent(int subtaskId, int attemptNumber, SourceEvent sourceEvent) {
+        // Only recognize events that don't care attemptNumber
+        handleSourceEvent(subtaskId, sourceEvent);
+    }
+
     @Override
     public void handleSourceEvent(int subtaskId, SourceEvent sourceEvent) {
         if (sourceEvent instanceof ReaderConsumeProgressEvent) {
```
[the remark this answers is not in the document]
zhuzhurk: In case some one later added a new kind of event directly to this method, ignoring the method with `attemptNumber`.

zhuzhurk (approved):
LGTM.

JingsongLi:
+1 Thanks @wwj6591812 and @zhuzhurk

PR #4397, opened 2024-10-29
[flink] fix sink parallelism in case of AQE

yunfengzhou-hub:
### Purpose

In #4221, a default parallelism was added to Paimon Sink in AQE cases. It is discovered that the value of Flink's `parallelism.default` configuration could actually be -1, which means Paimon sink might be unintentionally working under adaptive parallelism mode.

### Tests

BatchFileStoreITCase#testAQEWithXXX can be used to verify the changes in this PR and #4221.

### API and Format

No API or Format is changed.

### Documentation

No documentation needed. The existing unsupported cases had not been documented, so no modification needed for documents.


JingsongLi (approved):
+1

PR #4398, opened 2024-10-29
[flink] Introduce PartitionListener to enable custom actions in the F…

Aitozi:
…link committer for partitions.

### Purpose

replace for: https://github.com/apache/paimon/pull/4386

Fix for https://github.com/apache/paimon/issues/4402

1. Extract `PartitionListener`
2. Add `ReportHmsListener` which will report the partition statistic to hms when partition is idle. (enable by default)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/StoreCommitter.java
```diff
@@ -44,6 +45,7 @@ public class StoreCommitter implements Committer<Committable, ManifestCommittabl
     private final TableCommitImpl commit;
     @Nullable private final CommitterMetrics committerMetrics;
     @Nullable private final PartitionMarkDone partitionMarkDone;
```
JingsongLi, line 47: `PartitionMarkDone` can implement the `PartitionListener`.

JingsongLi (approved):
+1

PR #4399, opened 2024-10-29
[core]  add check for make directory for db location when creating database

### LINKED ISSUE #4396: [Bug]  the underlying local dir doesn't created but create schema return succ  if using a wrong localfs path
### Search before asking

### Paimon version

main

### Compute Engine

JavaAPI

### Minimal reproduce step

1. use a filesystem catalog backend
2. use a wrong fs path as warehouse location like `/wrong_path`
3. create schema succ, but the underlying path is not created

create schema failed

### Anything else?

_No response_

FANNG1:
### Purpose

Linked issue: close #4396 

According the API of mkdirs in `FileIO`, it may return false to represent mkdir failed,  we should handle the false condition when creating database.

```java
    boolean mkdirs(Path path) throws IOException;
```


JingsongLi (approved):
+1

PR #4401, opened 2024-10-29
[core] Fix the dead loop issue in CacheManager

### LINKED ISSUE #4400: [Bug] Fix the dead loop issue in CacheManager
### Search before asking

### Paimon version

Currently caffeine2 will remove the added item directly when it was deleted before which will cause dead loop in `CacheManager`

### Compute Engine

ALL

### Minimal reproduce step

Added CI

cache manager

### Anything else?

_No response_

FangYongs:
### Purpose

Linked issue: close #4400 

### Tests

Added test case CacheManagerTest

### API and Format
no

### Documentation

no

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/io/cache/CacheManager.java
+++ b/paimon-common/src/main/java/org/apache/paimon/io/cache/CacheManager.java
@@ -61,21 +61,18 @@ public class CacheManager {
 
     public MemorySegment getPage(CacheKey key, CacheReader reader, CacheCallback callback) {
         CacheValue value =
-                checkNotNull(
-                        cache.get(
-                                key,
-                                k -> {
-                                    this.fileReadCount++;
-                                    try {
-                                        return new CacheValue(
-                                                MemorySegment.wrap(reader.read(k)), callback);
-                                    } catch (IOException e) {
-                                        throw new RuntimeException(e);
-                                    }
-                                }),
-                        String.format("Cache result for key(%s) is null", key));
+                cache.get(
+                        key,
+                        k -> {
+                            this.fileReadCount++;
+                            try {
+                                return new CacheValue(MemorySegment.wrap(reader.read(k)), callback);
+                            } catch (IOException e) {
+                                throw new RuntimeException(e);
+                            }
+                        });
 
-        return value.segment;
+        return checkNotNull(value, String.format("Cache result for key(%s) is null", key)).segment;
     }
 
     public void invalidPage(CacheKey key) {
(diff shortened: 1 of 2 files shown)


FangYongs:
@JingsongLi Please have a look at this issue, THX

paimon-common/src/main/java/org/apache/paimon/io/cache/CacheManager.java
```diff
@@ -59,7 +59,7 @@ public CacheManager(MemorySize maxMemorySize) {
 
     public MemorySegment getPage(CacheKey key, CacheReader reader, CacheCallback callback) {
         CacheValue value = cache.getIfPresent(key);
-        while (value == null || value.isClosed) {
+        while (value == null) {
```
Aitozi, line 62: just use if ?

paimon-common/src/main/java/org/apache/paimon/io/cache/CacheManager.java
```diff
@@ -59,7 +59,7 @@ public CacheManager(MemorySize maxMemorySize) {
 
     public MemorySegment getPage(CacheKey key, CacheReader reader, CacheCallback callback) {
         CacheValue value = cache.getIfPresent(key);
```
JingsongLi, line 61: Can here just use `cache.get(key, Function creator)`?

FangYongs: Done

paimon-common/src/main/java/org/apache/paimon/io/cache/CacheManager.java
```diff
@@ -58,16 +60,21 @@ public CacheManager(MemorySize maxMemorySize) {
...
-            try {
-                this.fileReadCount++;
-                value = new CacheValue(MemorySegment.wrap(reader.read(key)), callback);
-            } catch (IOException e) {
-                throw new RuntimeException(e);
-            }
-            cache.put(key, value);
-        }
+        CacheValue value =
+                checkNotNull(
```
JingsongLi, line 64: remove `checkNotNull`? it looks useless.

paimon-common/src/main/java/org/apache/paimon/io/cache/CacheManager.java
```diff
@@ -95,7 +101,7 @@ private static class CacheValue {
         private final MemorySegment segment;
         private final CacheCallback callback;
 
-        private boolean isClosed = false;
+        private boolean cached = false;
```
JingsongLi, line 104: remove this field?

FangYongs: done

JingsongLi:
+1

PR #4403, opened 2024-10-29
[core] Manage the bloomfilter buffer in cache manager

Aitozi:
### Purpose

The BloomFilter in `SortLookupStoreReader` is not managed by the CacheManager, which will lead to the excessive buffer usage. This PR do these two things:

- Manage the BloomFilter's buffer with the CacheManager
- Release the BloomFilter's buffer when the Reader is closed.

Linked issue: close #xxx


JingsongLi (approved):
+1

PR #4404, opened 2024-10-30
[core] try to get table store for checking table options when creating table

LsomeYeah:
### Purpose

Linked issue: close #xxx

If some table options are set incorrectly, such as `ignore-delete = max`, it may cause an exception when dropping the table. This is because it tries to getTable before dropping.

The implementation of getting table in different catalog is different too. For example:
 `CachingCatalog` will `putTableCache` and this may call `PrimaryKeyFileStoreTable#store()` method, which will create `MergeFunction` through some table options. When trying to read these options, exception occurs, causing getTable failed and leads to dropping table failed.

So, When creating table, we will try to get table store for checking if the relevant table options are set correctly. If not, the table created will be deleted and throw an exception.

### Tests

CatalogTestBase#testCreateTable

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
@@ -279,6 +279,18 @@ public abstract class AbstractCatalog implements Catalog {
         } else {
             createTableImpl(identifier, schema);
         }
+        Table table = null;
+        try {
+            table = getTable(identifier);
+            FileStoreTable fileStoreTable = (FileStoreTable) table;
+            fileStoreTable.store();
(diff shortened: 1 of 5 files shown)


paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
```diff
@@ -279,6 +279,18 @@ public void createTable(Identifier identifier, Schema schema, boolean ignoreIfEx
         } else {
             createTableImpl(identifier, schema);
         }
+        Table table = null;
+        try {
+            table = getTable(identifier);
```
JingsongLi, line 284: I think you should do this in SchemaManager.

JingsongLi (approved):
+1

PR #4409, opened 2024-10-30
[hive] Refactor HiveCatalog to reduce io when getTable

JingsongLi:
### Purpose

There are a large number of useless IO accesses in the current HiveCatalog implementation, which can be reduced to one-fifth of the number of accesses through code refactoring.


yuzelin (approved):
+1

PR #4412, opened 2024-10-30
[core] Space allowed between multiple deleted tags

askwang:
### Purpose

If there are spaces between multiple deleted tags, will throw tag not found exception.


JingsongLi (approved):
+1

PR #4413, opened 2024-10-31
[flink] Lookup join supports time periods black list

yuzelin:
### Purpose

In the specified periods, lookup table cache refreshing is forbidden. This is aimed at avoiding server busy time.


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/FileStoreLookupFunction.java
```diff
@@ -293,10 +338,24 @@ private void reopen() {
...
 
-    private void checkRefresh() throws Exception {
-        if (nextLoadTime > System.currentTimeMillis()) {
+    @VisibleForTesting
+    void checkRefresh() throws Exception {
+        long currentTimeMillis = System.currentTimeMillis();
+        if (nextLoadTime > currentTimeMillis) {
             return;
         }
+
```
JingsongLi, line 347: refreshDynamicPartition should also be intercepted.

JingsongLi (approved):
+1

PR #4420, opened 2024-10-31
[core] Add file compression type in file names

askwang:
### Purpose

Default, the file name written by paimon only has the file format, it can also include the file compression like spark.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/io/DataFilePathFactory.java
+++ b/paimon-core/src/main/java/org/apache/paimon/io/DataFilePathFactory.java
@@ -39,18 +39,21 @@ public class DataFilePathFactory {
     private final String formatIdentifier;
     private final String dataFilePrefix;
     private final String changelogFilePrefix;
+    private final String fileCompression;
 
     public DataFilePathFactory(
             Path parent,
             String formatIdentifier,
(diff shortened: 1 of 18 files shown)


paimon-core/src/main/java/org/apache/paimon/io/DataFilePathFactory.java
```diff
@@ -62,7 +65,8 @@ public Path newChangelogPath() {
     }
 
     private Path newPath(String prefix) {
-        String name = prefix + uuid + "-" + pathCount.getAndIncrement() + "." + formatIdentifier;
+        String extension = "." + fileCompression + "." + formatIdentifier;
```
wwj6591812, line 68: Please add a config option that whether add fileCompression or not.

wwj6591812:
Why you submit this pr?
What are the business requirements for this feature?

askwang:
Has added a config option. thanks for review @JingsongLi @wwj6591812

askwang:
> Why you submit this pr? What are the business requirements for this feature?

It has no business effect, just provide convenience to know the compression type based on the written file name.

JingsongLi (approved):
+1

PR #4425, opened 2024-10-31
 [flink] support setting uid suffix for source/sink to improve state compatibility when jobGraph changes.

### LINKED ISSUE #4424: [Feature] support setting uid suffix for source/sink to improve state compatibility when jobGraph changes.
### Search before asking

### Motivation

In Flink jobs, the source and sink of paimon are both stateful. In our business scenarios, we may modify the parallelism, add sources, and replace source/sink tables, which can easily lead to problems where the state cannot be restored, such as `Cannot map checkpoint/savepoint state for operator xxx to the new program.`

For example, the following two operations cannot be restored from the Savepoint:

1. `insert into table_A (select f1 from table_B)`  -> `insert in table_A (select f1 from table_B union all select f1 from table_C)`

2. `insert into table_A select f1 from table_B` -> `insert into table_C select f1 from table_B`

Therefore I want to set `uid` for source/sink operators to improve state compatibility when restoring from savepoint.

### Solution

Set uid for source/sink operators.

### Anything else?

_No response_

liming30:
### Purpose

Linked issue: close #4424 

 [flink] support setting uid suffix for source/sink to improve state compatibility when jobGraph changes.

### Tests

1. `org.apache.paimon.flink.FlinkJobRecoveryITCase`

### API and Format

No

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkConnectorOptions.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkConnectorOptions.java
@@ -414,6 +414,24 @@ public class FlinkConnectorOptions {
                                     + "in order to compact several changelog files from the same partition into large ones, "
                                     + "which can decrease the number of small files. ");
 
+    public static final ConfigOption<String> SOURCE_OPERATOR_UID_SUFFIX =
+            key("source.operator-uid.suffix")
+                    .stringType()
+                    .noDefaultValue()
+                    .withDescription(
+                            "Set the uid suffix for the source operators. After setting, the uid format is "
+                                    + "${UID_PREFIX}_${TABLE_NAME}_${USER_UID_SUFFIX}. If the uid suffix is not set, flink will "
+                                    + "automatically generate the operator uid, which may be incompatible when the topology changes.");
+
+    public static final ConfigOption<String> SINK_OPERATOR_UID_SUFFIX =
+            key("sink.operator-uid.suffix")
+                    .stringType()
+                    .noDefaultValue()
+                    .withDescription(
+                            "Set the uid suffix for the writer, dynamic bucket assigner and committer operators. The uid format is "
+                                    + "${UID_PREFIX}_${TABLE_NAME}_${USER_UID_SUFFIX}. If the uid suffix is not set, flink will "
+                                    + "automatically generate the operator uid, which may be incompatible when the topology changes.");
+
     public static List<ConfigOption<?>> getOptions() {
         final Field[] fields = FlinkConnectorOptions.class.getFields();
         final List<ConfigOption<?>> list = new ArrayList<>(fields.length);
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkConnectorOptions.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkConnectorOptions.java
@@ -447,6 +447,11 @@ public class FlinkConnectorOptions {
         return list;
     }
 
+    public static String generateCustomUid(
+            String uidPrefix, String tableName, String userDefinedSuffix) {
+        return String.format("%s_%s_%s", uidPrefix, tableName, userDefinedSuffix);
+    }
+
     /** The mode of lookup cache. */
     public enum LookupCacheMode {
         /** Auto mode, try to use partial mode. */
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/DynamicBucketSink.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/DynamicBucketSink.java
@@ -24,24 +24,29 @@ import org.apache.paimon.table.Table;
 import org.apache.paimon.table.sink.ChannelComputer;
 import org.apache.paimon.table.sink.PartitionKeyExtractor;
 import org.apache.paimon.utils.SerializableFunction;
+import org.apache.paimon.utils.StringUtils;
 
 import org.apache.flink.api.common.typeinfo.BasicTypeInfo;
 import org.apache.flink.api.java.tuple.Tuple2;
 import org.apache.flink.api.java.typeutils.TupleTypeInfo;
 import org.apache.flink.streaming.api.datastream.DataStream;
 import org.apache.flink.streaming.api.datastream.DataStreamSink;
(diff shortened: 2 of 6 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/DynamicBucketSink.java
```diff
@@ -88,11 +93,21 @@ public DataStreamSink<?> build(DataStream<T> input, @Nullable Integer parallelis
...
-                        .transform("dynamic-bucket-assigner", rowWithBucketType, assignerOperator)
+                        .transform(
+                                DYNAMIC_BUCKET_ASSIGNER_NAME, rowWithBucketType, assignerOperator)
                         .setParallelism(partitionByKeyHash.getParallelism());
 
+        String uidSuffix = table.options().get(SINK_OPERATOR_UID_SUFFIX.key());
+        if (!StringUtils.isNullOrWhitespaceOnly(uidSuffix)) {
+            bucketAssigned =
+                    bucketAssigned.uid(
+                            String.format(
```
wwj6591812, line 106: Add a public static method in FlinkConnectorOptions?
public static String generateCustomUidSuffix(String xx, String yy, String zz) {
String.format(
                            "%s_%s_%s",
                            xx,
                            yy,
                            zz;
}

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/FlinkJobRecoveryITCase.java
```diff
@@ -0,0 +1,320 @@
...
+                                                                != ExecutionState.RUNNING) {
+                                                            allTaskRunning.set(false);
+                                                        }
+                                                    }))
+                    .get();
+        }
+
+        String checkpointPath = miniCluster.triggerCheckpoint(jobID).get();
+        Snapshot snapshot = waitForNewSnapshot("target_table", -1L);
+        while (snapshot.totalRecordCount() < totalReocrds) {
```
wwj6591812, line 313: Unboxing of 'snapshot.totalRecordCount()' may produce 'NullPointerException'

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/FlinkJobRecoveryITCase.java
```diff
@@ -0,0 +1,320 @@
...
+                        String.format(
+                                "CREATE TABLE IF NOT EXISTS `%s` (k INT, f1 STRING, pt STRING) WITH ('bucket'='-1', 'commit.force-create-snapshot'='true')",
+                                tableName));
+                return;
+            default:
+                throw new IllegalArgumentException("Unsupported bucket mode: " + bucketMode);
+        }
+    }
+
+    private Snapshot waitForNewSnapshot(String tableName, long initialSnapshot)
```
wwj6591812, line 260: Actual value of parameter 'tableName' is always 'target_table' 

Private static final String xxx = "target_table";

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/FlinkJobRecoveryITCase.java
```diff
@@ -0,0 +1,320 @@
...
+        Snapshot snapshot = findLatestSnapshot(tableName);
+        while (snapshot == null || snapshot.id() == initialSnapshot) {
+            Thread.sleep(2000L);
+            snapshot = findLatestSnapshot(tableName);
+        }
+        return snapshot;
+    }
+
+    @SuppressWarnings("unchecked")
+    private <T> T reflectGetField(Object instance, String fieldName)
```
wwj6591812, line 271: Actual value of parameter 'fieldName' is always 'miniCluster' 

Private static final String xxx = "miniCluster";

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/FlinkJobRecoveryITCase.java
liming30: My mistake. I should have used `target_table2` in the second check in `testRestoreFromSavepointWithIgnoreSinkState` . Resolved.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/DynamicBucketSink.java
liming30: resolved.

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/FlinkJobRecoveryITCase.java
liming30: Starting from paimon 0.4, `totalRecordCount` is no longer null. So I added `noinspection DataFlowIssue` here to avoid warnings.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkConnectorOptions.java
```diff
@@ -414,6 +414,24 @@ public class FlinkConnectorOptions {
...
+            key("source.operator-uid.suffix")
+                    .stringType()
+                    .noDefaultValue()
+                    .withDescription(
+                            "Set the uid suffix for the source operators. After setting, the uid format is "
+                                    + "${UID_PREFIX}_${TABLE_NAME}_${USER_UID_SUFFIX}. If the uid suffix is not set, flink will "
+                                    + "automatically generate the operator uid, which may be incompatible when the topology changes.");
+
+    public static final ConfigOption<String> SINK_OPERATOR_UID_SUFFIX =
+            key("sink.operator-uid.suffix")
```
Aitozi, line 427: If a job with a multi insert to same table eg: partial update + partial insert usage, will this lead to the same uid for two committer operator? The same to the source operator

liming30: I prefer to use flink `hints` syntax to set `uid` for different `insert into` statements instead of setting it in table options.

Aitozi: Get it 👍🏻

wwj6591812:
+1

PR #4436, opened 2024-11-04
[core] Optizime IN filter pushdown to snapshot/tag/schema system tables

xuzifu666:
### Purpose

Linked issue: close #xxx

1. Currently IN filter cannot pushdown to snapshot/tag/schema system tables and IN would tramsform to OR(contians all Equal LeafPredicate), if user query with IN filter would cost more unnessary IO in a large number of snapshots/tag/schema;
2. TagTable predicate specified as LeafPredicate which cannot resolve other kind of Predicate, this pr had improved it.
3. Add APIs for get snapshots/tags/schemas which get multiple numbers for a list argument.

such like sql: select snapshot_id, schema_id, commit_user from T$snapshots where snapshot_id in (1, 3);
before the pr would query all snapshots files firstly, then filter with 1 and 3;
after the pr  would query only snapshots file with 1 and 3, not query all snapshots file, which can reduce unnessary IO.
note: if filter contains other fields would handle regress as before( like: select snapshot_id, schema_id, commit_user from T$snapshots where snapshot_id in (1, 3) or/and commit_user='xxx'; ), just work only snapshot_id IN filter.


paimon-core/src/main/java/org/apache/paimon/table/system/TagsTable.java
```diff
@@ -217,18 +217,49 @@ public RecordReader<InternalRow> createReader(Split split) {
                 throw new IllegalArgumentException("Unsupported split: " + split.getClass());
             }
             Path location = ((TagsSplit) split).location;
-            LeafPredicate predicate = ((TagsSplit) split).tagName;
+            Predicate predicate = ((TagsSplit) split).tagPredicate;
             TagManager tagManager = new TagManager(fileIO, location, branch);
+            String leafName = "tag_name";
```
wwj6591812, line 222: first : private static final TAG_NAME = "tag_name";
second : replace all "tag_name" to TAG_NAME in this class

paimon-core/src/main/java/org/apache/paimon/table/system/TagsTable.java
```diff
@@ -217,18 +217,49 @@ public RecordReader<InternalRow> createReader(Split split) {
...
-                if (tagManager.tagExists(equalValue)) {
-                    nameToSnapshot.put(equalValue, tagManager.tag(equalValue));
+                if (predicate instanceof CompoundPredicate) {
+                    CompoundPredicate compoundPredicate = (CompoundPredicate) predicate;
+                    // optimize for IN filter
+                    if ((compoundPredicate.function()) instanceof Or) {
+                        List<Predicate> children = compoundPredicate.children();
+                        for (Predicate leaf : children) {
+                            if (leaf instanceof LeafPredicate
+                                    && (((LeafPredicate) leaf).function() instanceof Equal)
```
wwj6591812, line 244: Why here not judge:
((LeafPredicate) predicate).literals().get(0) instanceof BinaryString
?

paimon-core/src/main/java/org/apache/paimon/table/system/TagsTable.java
```diff
@@ -217,18 +217,49 @@ public RecordReader<InternalRow> createReader(Split split) {
...
+                                    && predicate
+                                                    .visit(LeafPredicateExtractor.INSTANCE)
+                                                    .get(leafName)
+                                            != null) {
+                                String equalValue =
+                                        ((LeafPredicate) leaf).literals().get(0).toString();
+                                predicateMap.put(equalValue, tagManager.tag(equalValue));
```
wwj6591812, line 251: if (tagManager.tagExists(equalValue)) {
predicateMap.put(equalValue, tagManager.tag(equalValue));
}

xuzifu666:
@wwj6591812 Thanks review，had addressed.

wwj6591812:
+1

LinMingQiang:
+1

paimon-core/src/main/java/org/apache/paimon/table/system/SchemasTable.java
```diff
@@ -223,6 +226,22 @@ public InnerTableRead withFilter(Predicate predicate) {
                         handleLeafPredicate(leaf, leafName);
                     }
                 }
+
+                // optimize for IN filter
+                if ((compoundPredicate.function()) instanceof Or) {
```
JingsongLi, line 231: We should extract a class to reduce duplicate codes, this can be done in a new PR.

JingsongLi (approved):
+1

Issue #4439, opened 2024-11-04
[Bug] JDBC catalog may produce duplicated schema names when calling list schema


FANNG1:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Paimon version

0.8

### Compute Engine

JavaAPI

### Minimal reproduce step

1. create paimon JDBC catalog
2. create schema `s1` 
3. create table `t1`
4. call list schema api, will get two `s1`

### What doesn't meet your expectations?

only get one `s1`

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

PR #4440, opened 2024-11-04
[Core] remove duplicated schema names in list schema in JDBC catalog

FANNG1:
### Purpose

fixes: #4439

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/jdbc/JdbcCatalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/jdbc/JdbcCatalog.java
@@ -157,7 +157,6 @@ public class JdbcCatalog extends AbstractCatalog {
                         row -> row.getString(JdbcUtils.DATABASE_NAME),
                         JdbcUtils.LIST_ALL_PROPERTY_DATABASES_SQL,
                         catalogKey));
-
         return databases.stream().distinct().collect(Collectors.toList());
     }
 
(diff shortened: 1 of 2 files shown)


paimon-core/src/main/java/org/apache/paimon/jdbc/JdbcCatalog.java
```diff
@@ -155,7 +156,8 @@ public List<String> listDatabases() {
                         row -> row.getString(JdbcUtils.DATABASE_NAME),
                         JdbcUtils.LIST_ALL_PROPERTY_DATABASES_SQL,
                         catalogKey));
-        return databases;
+
+        return databases.stream().distinct().collect(Collectors.toList());
```
davidradl, line 160: I am curious what the root cause of this is.
Why does databases contain the duplicate - is it ever valid? 
Are there other implications to leaving this list with duplicates? 
Would it not be better to fix this when the duplicate was added?
Also it would be good to add a junit.

wwj6591812: +1 curious

FANNG1: There are two tables in JDBC catalog, `paimon_database_properties` contains database properties, `paimon_tables`  contains table , schema, catalog relations.  The same database will present in these two tables if creating a database and tables because Paimon will automatically add some properties when creating schema.
@davidradl  do you have any suggestion where to place the test?

JingsongLi: Can you add a test case?

FANNG1: sure

FANNG1: @JingsongLi  @davidradl add a test, please help to review again.

wwj6591812:
+1

JingsongLi (approved):
+1

PR #4443, opened 2024-11-04
[core] Show fileSource in the FilesTable

Aitozi:
### Purpose

This could let user know where is the file come from: Append or Compact.


paimon-core/src/test/java/org/apache/paimon/table/system/FilesTableTest.java
```diff
@@ -211,7 +211,8 @@ private List<InternalRow> getExpectedResult(long snapshotId) {
                                             maxCol1, maxKey, partition1, partition2)),
                             file.minSequenceNumber(),
                             file.maxSequenceNumber(),
-                            file.creationTime()));
+                            file.creationTime(),
+                            BinaryString.fromString(file.fileSource().get().toString())));
```
wwj6591812, line 215: Why not : 
BinaryString.fromString(file.fileSource().map(FileSource::toString).orElse(null))
?

Aitozi: fixed

PR #4445, opened 2024-11-04
[core] throw exception if table is recreated when it still being read

LsomeYeah:
### Purpose

Linked issue: close #xxx

A table which is read in streaming read, will not throw exception even if the table is recreated currently. For example, the expected snapshot id is 10001, but the latest snapshot of the recreated table might be null or much less than 10001.

This is because when try to get the next snapshot, if the next snapshot is not exists in filesystem and its id is greater than earliest snapshot id, it will be treat as the next snapshot is still in generating progress.

This feature adds a check in `NextSnapshotFetcher#getNextSnapshot`. If next expected snapshot id is greater than the latest snapshot id plus one, it will throw an exception.

### Tests

`PrimaryKeyFileStoreTableITCase#testRecreateTableWithException`

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/utils/NextSnapshotFetcher.java
+++ b/paimon-core/src/main/java/org/apache/paimon/utils/NextSnapshotFetcher.java
@@ -48,20 +48,16 @@ public class NextSnapshotFetcher {
         Long latestSnapshotId = snapshotManager.latestSnapshotId();
         // No snapshot now
         if (earliestSnapshotId == null || earliestSnapshotId <= nextSnapshotId) {
-            if (earliestSnapshotId == null && nextSnapshotId > 1) {
+            if ((earliestSnapshotId == null && nextSnapshotId > 1)
+                    || (latestSnapshotId != null && nextSnapshotId > latestSnapshotId + 1)) {
                 throw new OutOfRangeException(
                         String.format(
-                                "The earliest snapshot is null now, but the next expected snapshot id is %d. "
-                                        + "Most possible cause might be the table had been recreated.",
-                                nextSnapshotId));
+                                "The next expected snapshot is too big! Most possible cause might be the table had been recreated."
+                                        + "The next snapshot id is %d, while the latest snapshot id is %s",
+                                nextSnapshotId,
+                                latestSnapshotId == null ? "null" : latestSnapshotId));
             }
-            if (latestSnapshotId != null && nextSnapshotId > latestSnapshotId + 1) {
-                throw new OutOfRangeException(
-                        String.format(
-                                "The next expected snapshot with id %d is greater than latest snapshot with id %d plus one. "
-                                        + "Most possible cause might be the table had been recreated.",
-                                nextSnapshotId, latestSnapshotId));
-            }
+
             LOG.debug(
                     "Next snapshot id {} does not exist, wait for the snapshot generation.",
                     nextSnapshotId);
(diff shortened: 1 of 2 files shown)


paimon-core/src/main/java/org/apache/paimon/utils/NextSnapshotFetcher.java
```diff
@@ -45,8 +45,19 @@ public Snapshot getNextSnapshot(long nextSnapshotId) {
...
         // No snapshot now
         if (earliestSnapshotId == null || earliestSnapshotId <= nextSnapshotId) {
+            if ((earliestSnapshotId == null && nextSnapshotId > 1)
+                    || (latestSnapshotId != null && nextSnapshotId > latestSnapshotId + 1)) {
+                throw new OutOfRangeException(
+                        String.format(
+                                "The next expected snapshot is too big! Most possible cause might be the table had been recreated."
+                                        + "The next snapshot id is %d, while the latest snapshot id is %s",
+                                nextSnapshotId,
+                                latestSnapshotId == null ? "null" : latestSnapshotId));
```
tsreaper, line 58:
```suggestion
                                latestSnapshotId));
```

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/PrimaryKeyFileStoreTableITCase.java
```diff
@@ -335,6 +338,77 @@ public void testBatchJobWithConflictAndRestart() throws Exception {
...
+            values.add(String.format("(0, %d, %d)", i, i + 1));
+            values.add(String.format("(1, %d, %d)", i, i + 1));
+        }
+        bEnv.executeSql("INSERT INTO t VALUES " + String.join(", ", values)).await();
+
+        // start a read job
+        CloseableIterator<Row> it = sEnv.executeSql("SELECT * FROM t").collect();
+
+        // wait the read job to read the current table
+        Thread.sleep(10000);
```
tsreaper, line 384: Assert that all snapshots are read by checking the result of `SELECT`. Don't sleep for this long!

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/PrimaryKeyFileStoreTableITCase.java
```diff
@@ -335,6 +338,77 @@ public void testBatchJobWithConflictAndRestart() throws Exception {
...
+    public void testRecreateTableWithException(boolean isReloadData) throws Exception {
+        TableEnvironment bEnv = tableEnvironmentBuilder().batchMode().build();
+        bEnv.executeSql(createCatalogSql("testCatalog", path + "/warehouse"));
+        bEnv.executeSql("USE CATALOG testCatalog");
+        bEnv.executeSql(
+                "CREATE TABLE t ( pt INT, k INT, v INT, PRIMARY KEY (pt, k) NOT ENFORCED ) "
+                        + "PARTITIONED BY (pt) "
+                        + "WITH ("
+                        + "    'bucket' = '2'\n"
+                        + ")");
```
tsreaper, line 353: Set `continuous.discovery-interval` to a smaller value (for example `1s` or `500ms`) so this test can run faster. No need to wait for 10 seconds.

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/PrimaryKeyFileStoreTableITCase.java
```diff
@@ -335,6 +338,88 @@ public void testBatchJobWithConflictAndRestart() throws Exception {
...
+            bEnv.executeSql("INSERT INTO t VALUES " + String.join(", ", values)).await();
+        }
+        assertThatCode(
+                        () -> {
+                            while (true) {
+                                if (it.hasNext()) {
+                                    it.next();
+                                }
+                            }
+                        })
```
tsreaper, line 417:
```suggestion
        assertThatCode(() -> it.next())
```

tsreaper (approved):

PR #4446, opened 2024-11-04
[core] Refactor the Field Aggregator factory to validate data types when creating function

zhuangchong:
### Purpose

1.Validate data types when creating aggregate functions, not at calculating
2.Create aggregate function in factory class

### Tests

org.apache.paimon.mergetree.compact.aggregate.FieldAggregatorTest


JingsongLi (approved):
Thanks @zhuangchong , +1

PR #4450, opened 2024-11-05
[cdc] Fix ambiguous naming in CdcRecord

zhuyaogai:
### Purpose

Linked issue: close #xxx

Currently, the field named `fields` in `CdcRecord` look like `schema` info, but actually represent `data` info, which is ambiguous. And there is also some ambiguity with `RichCdcRecord` field naming. So I rename the the field named `fields` in `CdcRecord` to `data`.

### Tests

Existing

### API and Format

No

### Documentation

No


JingsongLi (approved):
+1

PR #4454, opened 2024-11-05
[core] fix the issue where the historical schemaId could not be found when reading the branch table.

### LINKED ISSUE #4407: [Bug] Branch can only read data in latest schema
### Search before asking

### Paimon version

1.0-SNAPSHOT

### Compute Engine

Flink

### Minimal reproduce step

1. Create table with schema version 1
2. Insert some values
3. Alter table to create schema version 2 (but don't insert any new values)
4. Create tag & branch
5. Select * from branch 

The select job will fail because it only has schema version 2 but will try to read schema version 1.
Example stacktrace: https://gist.github.com/gmdfalk/802eb18c912a4d85e17f206820a0c55a

I expect the branch to be able to read any data, not just in the latest schema.
The branch only knows schema 2 and cannot read entries written in schema 1.

### Anything else?

_No response_

liming30:
### Purpose

Linked issue: close #4407 

[core] fix the issue where the historical schemaId could not be found when reading the branch table.

### Tests

- `org.apache.paimon.flink.BranchSqlITCase#testReadBranchTableWithMultiSchemaIds`

### API and Format

No
### Documentation

No

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/utils/BranchManager.java
+++ b/paimon-core/src/main/java/org/apache/paimon/utils/BranchManager.java
@@ -94,10 +94,7 @@ public class BranchManager {
 
         try {
             TableSchema latestSchema = schemaManager.latest().get();
-            fileIO.copyFile(
-                    schemaManager.toSchemaPath(latestSchema.id()),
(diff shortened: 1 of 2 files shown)


JingsongLi (approved):
+1

PR #4455, opened 2024-11-05
[core] Introduce refreshPartitions interface to CachingCatalog

mxdzs0612:
### Purpose

Among OLAP scenario, we need an interface to refresh cached partition info to assure that if it is necessary to renew a materialized view's partition.

### Tests

added.

### API and Format

Catalog.refreshPartitions

### Documentation

no

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/catalog/CachingCatalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/catalog/CachingCatalog.java
@@ -268,7 +268,6 @@ public class CachingCatalog extends DelegateCatalog {
 
     public void refreshPartitions(Identifier identifier) throws TableNotExistException {
         if (partitionCache != null) {
-            partitionCache.invalidate(identifier);
             List<PartitionEntry> result = wrapped.listPartitions(identifier);
             partitionCache.put(identifier, result);
         }
--- a/paimon-core/src/main/java/org/apache/paimon/catalog/CachingCatalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/catalog/CachingCatalog.java
@@ -267,6 +267,14 @@ public class CachingCatalog extends DelegateCatalog {
     }
 
     @Override
+    public void refreshPartitions(Identifier identifier) throws TableNotExistException {
+        if (partitionCache != null) {
+            partitionCache.invalidate(identifier);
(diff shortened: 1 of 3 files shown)


paimon-core/src/main/java/org/apache/paimon/catalog/CachingCatalog.java
```diff
@@ -266,6 +266,14 @@ public List<PartitionEntry> listPartitions(Identifier identifier)
         return result;
     }
 
+    @Override
+    public void refreshPartitions(Identifier identifier) throws TableNotExistException {
+        if (partitionCache != null) {
+            partitionCache.invalidate(identifier);
```
JingsongLi, line 272: just using wrapped.listPartitions and put to partitionCache?

paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
```diff
@@ -299,6 +299,14 @@ void dropPartition(Identifier identifier, Map<String, String> partitions)
      */
     List<PartitionEntry> listPartitions(Identifier identifier) throws TableNotExistException;
 
+    /**
+     * Refresh cached partitions of the table.
+     *
+     * @param identifier path of the table to refresh partitions
+     * @throws TableNotExistException if the table does not exist
+     */
+    default void refreshPartitions(Identifier identifier) throws TableNotExistException {}
```
JingsongLi, line 308: Do not add this, this is just for caching catalog.

paimon-core/src/main/java/org/apache/paimon/catalog/CachingCatalog.java
```diff
@@ -266,6 +266,14 @@ public List<PartitionEntry> listPartitions(Identifier identifier)
         return result;
     }
 
+    public void refreshPartitions(Identifier identifier) throws TableNotExistException {
+        if (partitionCache != null) {
+            partitionCache.invalidate(identifier);
```
JingsongLi, line 271: Why we need to invalidate it first?

mxdzs0612: Just copied from sr, didn't think through. I'll delete it.

JingsongLi:
+1

PR #4346, opened 2024-10-18
[spark] Replace create existing tag semantic with replace_tag

askwang:
### Purpose
In default way, the create_tag procedure for an existing tag means to update it or fail,  it's semantic is not very accurate, replace it with repalce_tag procedure.

before semantic
```sql
CALL paimon.sys.create_tag(table => 'test.T', tag => 'test_tag', snapshot => 1)")

-- success. update it
CALL paimon.sys.create_tag(table => 'test.T', tag => 'test_tag', snapshot => 1, time_retained => '1 d')")

-- failed
CALL paimon.sys.create_tag(table => 'test.T', tag => 'test_tag', snapshot => 2)")
```

after semantic
```sql
CALL paimon.sys.create_tag(table => 'test.T', tag => 'test_tag', snapshot => 1)")

-- failed
CALL paimon.sys.create_tag(table => 'test.T', tag => 'test_tag', snapshot => 1, time_retained => '1 d')")

-- success. use replace to update
CALL paimon.sys.replace_tag(table => 'test.T', tag => 'test_tag', snapshot => 2, time_retained => '1 d')")
```

PR #4457, opened 2024-11-05
[spark] Add replace tag ddl syntax

askwang:
### Purpose

follow up https://github.com/apache/paimon/pull/4346, add ddl syntax.


JingsongLi (approved):
+1 Thanks @askwang

PR #4461, opened 2024-11-06
[hive] Enable Format Table by default

JingsongLi:
### Purpose

By default, we should enable PaimonCatalog to have default access to Hive tables


yuzelin (approved):
+1

PR #4462, opened 2024-11-06
[spark] Add rollback ddl syntax

askwang:
### Purpose

Add rollback to snapshot/tag/timestamp ddl syntax.
```sql
alter table t rollback to snapshot `2`;
alter table t rollback to tag `test-tag`;
alter table t rollback to timestamp `1730876906134`;
```

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/antlr4/org.apache.spark.sql.catalyst.parser.extensions/PaimonSqlExtensions.g4
+++ b/paimon-spark/paimon-spark-common/src/main/antlr4/org.apache.spark.sql.catalyst.parser.extensions/PaimonSqlExtensions.g4
@@ -74,7 +74,7 @@ statement
     | ALTER TABLE multipartIdentifier createReplaceTagClause                                #createOrReplaceTag
     | ALTER TABLE multipartIdentifier DELETE TAG (IF EXISTS)? identifier                    #deleteTag
     | ALTER TABLE multipartIdentifier RENAME TAG identifier TO identifier                   #renameTag
-    | ALTER TABLE multipartIdentifier ROLLBACK TO kind identifier                           #rollback
+    | ALTER TABLE multipartIdentifier ROLLBACK TO timeTravelKind identifier                 #rollback
   ;
 
 callArgument
@@ -105,7 +105,7 @@ timeUnit
   | MINUTES
   ;
 
-kind
+timeTravelKind
     : SNAPSHOT
     | TAG
     | TIMESTAMP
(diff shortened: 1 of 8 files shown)


LinMingQiang:
Why do we need a new `alter` ddl syntax. Is it the same as the `call`  procedure ?

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/execution/RollbackExec.scala
```diff
@@ -0,0 +1,70 @@
...
+            val snapshot = paimonTable.snapshotManager.earlierOrEqualTimeMills(version.toLong)
+            assert(snapshot != null)
+            paimonTable.rollbackTo(snapshot.id);
+
+          case _ =>
+            throw new UnsupportedOperationException(s"Unsupported rollback kind '$kind'.")
+        }
+      case t =>
+        throw new UnsupportedOperationException(
+          s"Can not delete tag for non-paimon FileStoreTable: $t")
```
wwj6591812, line 66: Can not delete tag?

askwang:
> Why do we need a new `alter` ddl syntax. Is it the same as the `call` procedure ?

SQL syntax support will be provided later, it is the same as procedure call.

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/execution/RollbackExec.scala
askwang: my mistake, updated, ths

LinMingQiang:
Why do we use  `alter table`  instead of `call` if they are the same?  And `alter table` should be used to modify the table metadata information, like alter table add col.

askwang:
> Why do we use `alter table` instead of `call` if they are the same? And `alter table` should be used to modify the table metadata information, like alter table add col.

I think tag/snapshot are also part of the table meta info, maybe we can use alter table to modify it.

paimon-spark/paimon-spark-common/src/main/antlr4/org.apache.spark.sql.catalyst.parser.extensions/PaimonSqlExtensions.g4
```diff
@@ -104,6 +105,12 @@ timeUnit
   | MINUTES
   ;
 
+kind
```
JingsongLi, line 108: timeTravelKind?

askwang: changed.

JingsongLi:
> Why do we use `alter table` instead of `call` if they are the same? And `alter table` should be used to modify the table metadata information, like alter table add col.

I agree with your idea. I have checked and currently no other lake formats have implemented custom syntax here. We can consider temporarily stopping and using CALL is not a bad idea.

@askwang What do you think?

Zouxxyy:
In my mind, rollback should actually generate a new snapshot containing the changes, from this perspective, alter table should not be used. (although the current implementation directly deletes the existing snapshots).

Besides in spark doc description: ALTER TABLE statement changes the schema or properties of a table.

askwang:
> > Why do we use `alter table` instead of `call` if they are the same? And `alter table` should be used to modify the table metadata information, like alter table add col.
> 
> I agree with your idea. I have checked and currently no other lake formats have implemented custom syntax here. We can consider temporarily stopping and using CALL is not a bad idea.
> 
> @askwang What do you think?

ok

PR #4467, opened 2024-11-06
[core] Fast return if rollback verion equals lastest version

askwang:
### Purpose

If the rollback snapshot version is equal to the lastest snapshot version, we should fast return.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/AbstractFileStoreTable.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/AbstractFileStoreTable.java
@@ -512,13 +512,12 @@ abstract class AbstractFileStoreTable implements FileStoreTable {
                 "Rollback snapshot '%s' doesn't exist.",
                 snapshotId);
 
-        Snapshot snapshot = snapshotManager.snapshot(snapshotId);
         // fast return
-        if (snapshot.equals(snapshotManager.latestSnapshot())) {
+        if (new Long(snapshotId).equals(snapshotManager.latestSnapshotId())) {
             return;
         }
 
-        rollbackHelper().cleanLargerThan(snapshot);
+        rollbackHelper().cleanLargerThan(snapshotManager.snapshot(snapshotId));
     }
 
     public Snapshot findSnapshot(long fromSnapshotId) throws SnapshotNotExistException {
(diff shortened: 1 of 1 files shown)


paimon-core/src/main/java/org/apache/paimon/table/AbstractFileStoreTable.java
```diff
@@ -625,14 +630,19 @@ public void rollbackTo(String tagName) {
         checkArgument(tagManager.tagExists(tagName), "Rollback tag '%s' doesn't exist.", tagName);
 
         Snapshot taggedSnapshot = tagManager.taggedSnapshot(tagName);
+        SnapshotManager snapshotManager = snapshotManager();
```
wwj6591812, line 633: Add a blank line.

paimon-core/src/main/java/org/apache/paimon/table/AbstractFileStoreTable.java
```diff
@@ -512,6 +512,11 @@ public void rollbackTo(long snapshotId) {
                 "Rollback snapshot '%s' doesn't exist.",
                 snapshotId);
 
+        // fast return
+        if (new Long(snapshotId).equals(snapshotManager.latestSnapshotId())) {
```
wwj6591812, line 516: I think new Long has a little strange.
Why not:

Long latestSnapshotId = snapshotManager.latestSnapshotId();
if (latestSnapshotId == null) {
    exception or error log
}
if (taggedSnapshot.id() == latestSnapshotId) {
    fast return
}

wwj6591812:
A question:
Assuming that when the user executes the rollback procedure with snapshot = 1, the latestSnapshot Id of the Paimon table is 1, and before the rollback procedure ends, the latestSnapshot Id of the Paimon table changes to 2. After you apply this patch, in the end, the rollback procedure fast return so that the result is not snapshot = 1.
 Is this the result that the user wants?

askwang:
> A question: Assuming that when the user executes the rollback procedure with snapshot = 1, the latestSnapshot Id of the Paimon table is 1, and before the rollback procedure ends, the latestSnapshot Id of the Paimon table changes to 2. After you apply this patch, in the end, the rollback procedure fast return, so the RollbackHelper#cleanLargerThan will not be called. Is this the result that the user wants?

In this case, the result is not what the user wants. But although there does not fast return, this case may also occur. For example, the rollback snapshot is 1 and the latest snapshot of table is also 1, before rollback ends the lastest snapshot of table changes to 2 or 3, the rollback logic gets the lastest shotshot is diffcult to keep always the newest, this will cause some new snapshots not to be processed.

JingsongLi:
Hi @askwang , what is this PR aims to? just for performance?

JingsongLi:
> > Hi @askwang , what is this PR aims to? just for performance?
> 
> yes, juse fast return, no need to execute RollbackHelper#cleanLargerThan.

Do we need to care about the speed of this operation? I think it should be very fast even no optimization.

PR #4468, opened 2024-11-06
[core] Delete useless methods in Catalog

JingsongLi:
### Purpose

Delete follow methods in Catalog because of useless, We recommend using methods such as `getTable` directly to avoid errors in implementation.
```
    /**
     * Get lock factory from catalog. Lock is used to support multiple concurrent writes on the
     * object store.
     */
    Optional<CatalogLockFactory> lockFactory();

    /** Get lock context for lock factory to create a lock. */
    default Optional<CatalogLockContext> lockContext() {
        return Optional.empty();
    }

    /** Get metastore client factory for the table specified by {@code identifier}. */
    default Optional<MetastoreClient.Factory> metastoreClientFactory(Identifier identifier)
            throws TableNotExistException {
        return Optional.empty();
    }
```

Delete follow methods in Catalog because of repetitive:
```
    /**
     * Check if a database exists in this catalog.
     *
     * @param databaseName Name of the database
     * @return true if the given database exists in the catalog false otherwise
     */
    default boolean databaseExists(String databaseName) {
        try {
            loadDatabaseProperties(databaseName);
            return true;
        } catch (DatabaseNotExistException e) {
            return false;
        }
    }

    /**
     * Check if a table exists in this catalog.
     *
     * @param identifier Path of the table
     * @return true if the given table exists in the catalog false otherwise
     */
    default boolean tableExists(Identifier identifier) {
        try {
            return getTable(identifier) != null;
            return false;
        }
    }

    /**
     * Check if a view exists in this catalog.
     *
     * @param identifier Path of the view
     * @return true if the given view exists in the catalog false otherwise
     */
    default boolean viewExists(Identifier identifier) {
        try {
            return getView(identifier) != null;
        } catch (ViewNotExistException e) {
            return false;
        }
    }
```

Modify `loadDatabaseProperties` in Catalog because we can have a better naming for `getDatabase`:
```
    /**
     * Return a {@link Database} identified by the given name.
     *
     * @param name Database name
     * @return The requested {@link Database}
     * @throws DatabaseNotExistException if the requested database does not exist
     */
    Database getDatabase(String name) throws DatabaseNotExistException;
```

### Tests

Existing tests.

### API and Format

Catalog API.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
@@ -98,7 +98,6 @@ public abstract class AbstractCatalog implements Catalog {
         return fileIO;
     }
 
-    @Override
     public Optional<CatalogLockFactory> lockFactory() {
         if (!lockEnabled()) {
             return Optional.empty();
@@ -118,7 +117,6 @@ public abstract class AbstractCatalog implements Catalog {
         return Optional.empty();
     }
 
-    @Override
     public Optional<CatalogLockContext> lockContext() {
         return Optional.of(CatalogLockContext.fromOptions(catalogOptions));
     }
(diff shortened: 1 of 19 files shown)


yuzelin (approved):
+1

PR #4475, opened 2024-11-07
commit failed, add table name to the log

melin:
### Purpose

flink cdc pipeline writes multiple paimon. If write fails, only file name information, no table name.


JingsongLi (approved):
+1

PR #4477, opened 2024-11-07
[core] Call full gc to avoid meta space oom while generating classes.

leaves12138:
### Purpose

To avoid out-of-memory error while generating classes, so we call full gc to collect garbage classloaders.


wwj6591812:
+1

JingsongLi (approved):
+1

PR #4478, opened 2024-11-07
[Core]commit failed, add table name to the log

melin:
### Purpose
flink cdc pipeline writes multiple table. If write failed, only file name information, no table name. It is difficult to find which table failed to write


wwj6591812:
You means FlinkCdcMultiTableSink?
@melin

JingsongLi:
I remember I have merged this PR?....

PR #4479, opened 2024-11-07
[parquet] Support to enable parquet bloomfilter

Aitozi:
### Purpose

This pr is meant to support to enable parquet bloomfilter.

```
// Enable the bloom filter for all columns
conf.set("parquet.bloom.filter.enabled", true);
// Disable the bloom filter for the column 'column.path'
conf.set("parquet.bloom.filter.enabled#column.path", false);
// The bloom filter will be enabled for all columns except 'column.path'
```

Linked issue: close #xxx


JingsongLi (approved):
+1

PR #4480, opened 2024-11-07
[core] Add batch add partitions in MetastoreClient

Zouxxyy:
### Purpose

hiveClient support `client.add_partitions` to add partitions at once, we can use it in AddPartitionCommitCallback

this speeds up the load table a lot when have many partitions to sync.


wwj6591812:
+1

JingsongLi (approved):
+1

PR #4482, opened 2024-11-08
[core] Fix merge schemas equal method issue

harveyyue:
### Purpose

Linked issue: close #xxx
No


wwj6591812:
+1

JingsongLi:
I remember this is not an issue. @harveyyue Can you find the exact case?

JingsongLi:
> > I remember this is not an issue. @harveyyue Can you find the exact case?
> 
> Using spark format `paimon` to write upstream data will generate a new schema-xxx since have no schema changes `dataset.format("paimon").options(options);`

@harveyyue Can you add a test case?

harveyyue:
> > > I remember this is not an issue. @harveyyue Can you find the exact case?
> > 
> > 
> > Using spark format `paimon` to write upstream data will generate a new schema-xxx since have no schema changes `dataset.format("paimon").options(options);`
> 
> @harveyyue Can you add a test case?

Done
We did some changes to reselect column according table schema, so we will trigger the `==` compare method from row type.

JingsongLi (approved):
+1

PR #4483, opened 2024-11-08
[core][spark] Support adding and dropping nested columns in Spark

tsreaper:
### Purpose

Spark SQL supports adding and dropping nested columns, but Paimon currently does not support it. This PR adds support for adding and dropping nested columns.

### Tests

Unit tests and IT cases.

### API and Format

No format changes.

### Documentation

This feature does not need a document.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
@@ -536,7 +536,7 @@ public abstract class AbstractCatalog implements Catalog {
         for (SchemaChange change : changes) {
             if (change instanceof SchemaChange.AddColumn) {
                 SchemaChange.AddColumn addColumn = (SchemaChange.AddColumn) change;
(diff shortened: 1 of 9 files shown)


JingsongLi:
Try to support this in Flink SQL too:
```
Flink SQL> create table s1 (a int, nc row<f1 int, f2 string> ) with ('connector'='datagen', 'number-of-rows'='10');

Flink SQL> alter table s1 modify nc row<f1 int, f2 string, f3 string>;
```

tsreaper:
> Try to support this in Flink SQL too:
> 
> ```
> Flink SQL> create table s1 (a int, nc row<f1 int, f2 string> ) with ('connector'='datagen', 'number-of-rows'='10');
> [INFO] Execute statement succeed.
> 
> Flink SQL> alter table s1 modify nc row<f1 int, f2 string, f3 string>;
> [INFO] Execute statement succeed.
> ```

Let's implement this in another PR.

JingsongLi (commented):
Please add doc too

JingsongLi (approved):
+1

PR #4485, opened 2024-11-08
[Bug] Fix the NPE when report the statistic with DeletionFile

Aitozi:
### Purpose

image

This is due to the `List<DeletionFile>` may contain `null`. I think the hms metadata should only contains the datafile. So, 
remove the DeletionFile related code.


JingsongLi (approved):
+1

PR #4487, opened 2024-11-09
[core] Introduce drop-deleted-record-count compaction metric.

zhuyaogai:
### Purpose

Linked issue: close #xxx

Introduce a metric that records the number of  records(`DELETE(-D)`) that have been dropped by compaction.

### Tests

No

### API and Format

No

### Documentation

No


zhuyaogai:
Updated @wwj6591812

JingsongLi:
Can you explain the real usage in production env? I cannot get why from `Introduce a metric that records the number of records(DELETE(-D)) that have been dropped by compaction.`.

JingsongLi:
Feel free to re-open this if you think it is true requirement.

PR #4489, opened 2024-11-11
[core][spark] Support renaming nested columns in Spark

tsreaper:
### Purpose

Currently Paimon only supports renaming not nested columns. This PR supports renaming nested columns in Paimon, and also add support to Spark.

### Tests

Unit tests and IT cases.

### API and Format

No format changes.

### Documentation

Document will be added in a dedicated PR.


JingsongLi (approved):
+1

PR #4492, opened 2024-11-11
[core][flink] Introduce HashMapLocalMerger to 'local-merge-buffer-size'

JingsongLi:
### Purpose

We previously introduced the Local sorting method for merging, which can reduce a large amount of data when there is a lot of repetition.

However, the cost of sorting is very high. We can refer to Flink's Local Aggregation and introduce Hash, which can greatly improve the performance of Local Merge. (3 times+)


leaves12138 (approved):
+1

PR #4498, opened 2024-11-11
[hive] Make HiveCatalog.listTables Lighter weight

JingsongLi:
### Purpose

Try to avoid create Paimon table in HiveCatalog.listTables, this can accelerate list tables and avoid potential problems (just show tables here, we don't need to access tables).


wwj6591812:
+1

PR #4499, opened 2024-11-12
[flink] Support updating row type to another row type in Flink

tsreaper:
### Purpose

Unlike Spark, Flink does not support directly updating a nested column. So we support updating a whole row type to another row type in Flink to implement this feature.

### Tests

Unit tests and IT cases.

### API and Format

No format changes.

### Documentation

Document will be added later.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/schema/SchemaManager.java
+++ b/paimon-core/src/main/java/org/apache/paimon/schema/SchemaManager.java
@@ -40,6 +40,7 @@ import org.apache.paimon.table.FileStoreTableFactory;
 import org.apache.paimon.types.DataField;
 import org.apache.paimon.types.DataType;
 import org.apache.paimon.types.DataTypeCasts;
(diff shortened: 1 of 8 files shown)


paimon-core/src/main/java/org/apache/paimon/schema/SchemaManager.java
```diff
@@ -596,6 +590,58 @@ private static void assertNotUpdatingPrimaryKeys(
...
+            boolean keepNullability,
+            AtomicInteger highestFieldId,
+            String fullFieldName) {
+        if (keepNullability) {
+            targetType = targetType.copy(field.type().isNullable());
+        }
+
+        DataType oldType = field.type();
+        if (oldType.getTypeRoot() == DataTypeRoot.ROW) {
+            checkState(
```
JingsongLi, line 605: This logical is for Flink, we should move these logical to `FlinkCatalog`.

JingsongLi (approved):
+1

PR #4500, opened 2024-11-12
[core] Make default of 'lookup.local-file-type' to sort

JingsongLi:
### Purpose

Sort has better performance and compression ratio, so we should consider switching to sort by default.


Aitozi:
I prefer to have more benchmark between the sort and hash format before making the sort format as default.

JingsongLi:
> I prefer to have more benchmark between the sort and hash format before making the sort format as default.

Thanks! Expect more input from you~

PR #4501, opened 2024-11-12
[hive] HiveCatalog.listViews method reduces the number of queries for hms metadata.

zhuangchong:
### Purpose

HiveCatalog.listViews method reduces the number of queries for hms metadata.


JingsongLi (approved):
+1

PR #4502, opened 2024-11-12
[core] Introduce bitmap index record reader

Tan-JiaLiang:
### Purpose

Introduction of bitmap index record reader to filter data in advance with bitmap index result.

### Tests

org.apache.paimon.table.AppendOnlyFileStoreTableTest#testBSIAndBitmapIndexInMemory
org.apache.paimon.table.AppendOnlyFileStoreTableTest#testBSIAndBitmapIndexInDisk
org.apache.paimon.table.PrimaryKeyFileStoreTableTest#testDeletionVectorsWithFileIndexInMeta
org.apache.paimon.table.PrimaryKeyFileStoreTableTest#testDeletionVectorsWithBitmapFileIndexInFile

### API and Format

No.

### Documentation

No.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/RawFileSplitRead.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/RawFileSplitRead.java
@@ -24,6 +24,8 @@ import org.apache.paimon.deletionvectors.ApplyDeletionVectorReader;
 import org.apache.paimon.deletionvectors.DeletionVector;
 import org.apache.paimon.disk.IOManager;
 import org.apache.paimon.fileindex.FileIndexResult;
+import org.apache.paimon.fileindex.bitmap.ApplyBitmapIndexRecordReader;
+import org.apache.paimon.fileindex.bitmap.BitmapIndexResult;
 import org.apache.paimon.format.FileFormatDiscover;
 import org.apache.paimon.format.FormatKey;
 import org.apache.paimon.format.FormatReaderContext;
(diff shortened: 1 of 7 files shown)


paimon-core/src/main/java/org/apache/paimon/operation/RawFileSplitRead.java
```diff
@@ -212,14 +214,20 @@ private RecordReader<InternalRow> createFileReader(
...
-        FileRecordReader fileRecordReader =
+        RecordReader<InternalRow> fileRecordReader =
                 new FileRecordReader(
                         bulkFormatMapping.getReaderFactory(),
                         formatReaderContext,
                         bulkFormatMapping.getIndexMapping(),
                         bulkFormatMapping.getCastMapping(),
                         PartitionUtils.create(bulkFormatMapping.getPartitionPair(), partition));
 
+        if (fileIndexResult instanceof BitmapIndexResult) {
```
JingsongLi, line 225: Here should be bug, bitmap in `ApplyBitmapIndexRecordReader` should be merged with `ApplyDeletionVectorReader`, otherwise the latter will encounter problems.

Can you create a test for this?

JingsongLi (changes requested):

paimon-core/src/main/java/org/apache/paimon/operation/RawFileSplitRead.java
Tan-JiaLiang: Could you be more specific for the bug?

As I understand it
The bitmap index reader would filter all the mismatched data
The DV reader would filter all the deleted data
Is there a conflict?

Can this test case be covered?
org.apache.paimon.table.PrimaryKeyFileStoreTableTest#testDeletionVectorsWithFileIndexInMeta
org.apache.paimon.table.PrimaryKeyFileStoreTableTest#testDeletionVectorsWithBitmapFileIndexInFile

JingsongLi: You are right, the magic is in the returned position, it makes chain processing possible.

JingsongLi (approved):
+1

PR #4504, opened 2024-11-12
[doc] fix flink procedure document sql syntax error.

zhuangchong:
### Purpose

fix flink procedure sql syntax error.


JingsongLi (approved):
+1

PR #4506, opened 2024-11-12
[core] Support delete stats in result of scan plan.

wwj6591812:
### Purpose

In my company's production environment, when use Flink session cluster for OLAP scan Paimon, we found the JobManager's memory is always heavy.
So, we will optimize this by two ways:
(1) Delete stats in DataSplit.
(2) When dataSkipping, cut unused stats in ManifestEntry.

This pr is for (1)

Linked issue: close #xxx

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/io/DataFileMeta.java
+++ b/paimon-core/src/main/java/org/apache/paimon/io/DataFileMeta.java
@@ -409,6 +409,27 @@ public class DataFileMeta {
                 valueStatsCols);
     }
 
+    public DataFileMeta copyWithoutStats() {
+        return new DataFileMeta(
+                fileName,
+                fileSize,
+                rowCount,
+                minKey,
+                maxKey,
+                keyStats,
+                EMPTY_STATS,
+                minSequenceNumber,
+                maxSequenceNumber,
+                schemaId,
+                level,
+                extraFiles,
+                creationTime,
+                deleteRowCount,
+                embeddedIndex,
+                fileSource,
+                Collections.emptyList());
+    }
+
     public List<Path> collectFiles(DataFilePathFactory pathFactory) {
         List<Path> paths = new ArrayList<>();
         paths.add(pathFactory.toPath(fileName));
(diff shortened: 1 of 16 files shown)


wwj6591812:
@JingsongLi Hi，Please CC, Thx.

paimon-core/src/main/java/org/apache/paimon/io/DataFileMeta.java
```diff
@@ -409,6 +409,27 @@ public DataFileMeta rename(String newFileName) {
                 valueStatsCols);
     }
 
+    public DataFileMeta withoutStats() {
```
JingsongLi, line 412: copyWithoutStats

paimon-core/src/main/java/org/apache/paimon/io/DataFileMeta.java
```diff
@@ -409,6 +409,27 @@ public DataFileMeta rename(String newFileName) {
...
+        return new DataFileMeta(
+                fileName,
+                fileSize,
+                rowCount,
+                minKey,
+                maxKey,
+                EMPTY_STATS,
```
JingsongLi, line 419: It is better to keep key stats?

paimon-core/src/main/java/org/apache/paimon/io/DataFileMeta.java
```diff
@@ -409,6 +409,27 @@ public DataFileMeta rename(String newFileName) {
...
+                minSequenceNumber,
+                maxSequenceNumber,
+                schemaId,
+                level,
+                extraFiles,
+                creationTime,
+                deleteRowCount,
+                embeddedIndex,
+                fileSource,
+                valueStatsCols);
```
JingsongLi, line 430: empty list

paimon-core/src/main/java/org/apache/paimon/manifest/ManifestEntry.java
```diff
@@ -121,6 +121,10 @@ public Identifier identifier() {
                 file.embeddedIndex());
     }
 
+    public ManifestEntry withoutStats() {
```
JingsongLi, line 124: copyWithoutStats

JingsongLi (changes requested):

wwj6591812:
@JingsongLi 
had addressed, Thanks for review.

JingsongLi (approved):
+1

PR #4508, opened 2024-11-13
[core] Adjust default value of 'snapshot.expire.limit'

askwang:
### Purpose

When we use flink write data and spark compact, we find expiring only 10 snapshots each time in compact, and the snapshot backlog is quite serious. Increase the default value to 50 to avoid this situation.


JingsongLi:
+1

PR #4509, opened 2024-11-13
[doc]  Remove unused config in record-level expire

askwang:
### Purpose

In https://github.com/apache/paimon/pull/4458 removed the config `record-level.time-field-type`, update doc.


JingsongLi:
+1

PR #4511, opened 2024-11-13
[core]partition system table support time travel

zddr:
### Purpose

Linked issue: close #xxx

partition system table support time travel


JingsongLi:
+1

PR #4512, opened 2024-11-13
[core] Make FormatReaderFactory return FileRecordReader to reduce cast

JingsongLi:
### Purpose

In `ApplyBitmapIndexRecordReader` and `ApplyBitmapIndexRecordReader`, they will check the type of iterator which should be `FileRecordIterator`, it is always `FileRecordIterator`, and we should garrantee it by interface.

So we let FormatReaderFactory return FileRecordReader, and all related readers should be `FileRecordReader`s.


yuzelin (approved):
+1

PR #4513, opened 2024-11-13
[parquet] Fix nested array/map has no id in parquet files

tsreaper:
### Purpose

In #4362 we write field ids into parquet files. However that PR fails to add id for nested array/map type (for example `ARRAY(MAP(INT, ARRAY(INT)))` type. This PR fixes the bug.

### Tests

Unit tests.

### API and Format

Field id of array element, map keys and values are changed.

### Documentation

No document is needed.


JingsongLi (approved):
+1

PR #4518, opened 2024-11-13
[doc] Update hll_sketch and theta_sketch example in aggregation.md

hadoopkandy:
### Purpose
Update hll_sketch and theta_sketch example in aggregation.md

### Documentation

Don't add any new feature


JingsongLi (approved):
Thanks @hadoopkandy

PR #4520, opened 2024-11-13
[core] Introduce binlog system table to pack the UB and UA

### LINKED ISSUE #4505: [Feature] Allow to pack the UPDATE_BEFORE and UPDATE_AFTER message when reading the audit table
### Search before asking

### Motivation

Flink breaks the ub and ua into two records, however, user may want to compare the ub and ua message in the udf to find the diff column. So, I proposal to add a mode in aduit_table to support consume the ub and ua message in one row. 

We could make the original column type to an array type.

for example:

```
    public void testAuditTablePackUbUa() throws Exception {
        sql(
                "CREATE TABLE T (a INT, b INT, primary key (a) NOT ENFORCED) with ('changelog-producer' = 'lookup', 'bucket' = '2', 'scan.pack.update-before-after.enabled' = 'true')");
        BlockingIterator<Row, Row> iterator =
                streamSqlBlockIter(
                        "SELECT * FROM T$audit_log /*+ OPTIONS('scan.mode' = 'latest') */");
        sql("INSERT INTO T VALUES (1, 2)");
        sql("INSERT INTO T VALUES (2, 2)");
        List<Row> rows = iterator.collect(3);
        System.out.println(rows);
        iterator.close();
    }
```

This will produce 

```
[+I[+I, [1, null], [2, null]], +I[+U, [1, 1], [2, 3]], +I[+I, [2, null], [2, null]]]
```

If the column has two index, then the first index is update_before value, and the second index is update_after value. 

### Solution

_No response_

### Anything else?

_No response_

Aitozi:
### Purpose

Linked issue: close #4505

### DIFF
--- a/docs/content/maintenance/system-tables.md
+++ b/docs/content/maintenance/system-tables.md
@@ -406,4 +406,23 @@ SELECT * FROM T$statistics;
 1 rows in set
 */
 ```
+### Binlog Table
+
+You can streaming or batch query the binlog through binlog table. In this system table,
+the update before and update after will be packed in one row. 
+
+```
+/*
++------------------+----------------------+-----------------------+
+|     rowkind      |       column_0       |       column_1        |
++------------------+----------------------+-----------------------+
+|        +I        |       [col_0]        |       [col_1]         |
++------------------+----------------------+-----------------------+
+|        +U        | [col_0_ub, col_0_ua] | [col_1_ub, col_1_ua]  |
++------------------+----------------------+-----------------------+
+|        -D        |       [col_0]        |       [col_1]         |
++------------------+----------------------+-----------------------+
+*/
+```
+
 
--- a/paimon-common/src/main/java/org/apache/paimon/reader/PackChangelogReader.java
+++ b/paimon-common/src/main/java/org/apache/paimon/reader/PackChangelogReader.java
@@ -1,0 +1,131 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.paimon.reader;
+
+import org.apache.paimon.data.InternalRow;
+import org.apache.paimon.data.serializer.InternalRowSerializer;
+import org.apache.paimon.types.RowKind;
+import org.apache.paimon.types.RowType;
+
+import javax.annotation.Nullable;
+
+import java.io.IOException;
+import java.util.function.BiFunction;
+
+/** The reader which will pack the update before and update after message together. */
+public class PackChangelogReader implements RecordReader<InternalRow> {
+
+    private final RecordReader<InternalRow> reader;
+    private final BiFunction<InternalRow, InternalRow, InternalRow> function;
+    private final InternalRowSerializer serializer;
+    private boolean initialized = false;
+
+    public PackChangelogReader(
+            RecordReader<InternalRow> reader,
+            BiFunction<InternalRow, InternalRow, InternalRow> function,
+            RowType rowType) {
+        this.reader = reader;
+        this.function = function;
+        this.serializer = new InternalRowSerializer(rowType);
+    }
+
+    @Nullable
+    @Override
+    public RecordIterator<InternalRow> readBatch() throws IOException {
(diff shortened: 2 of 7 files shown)


JingsongLi (commented):
I feel like we can create a new system table instead of an option.

Aitozi:
> I feel like we can create a new system table instead of an option.

Yes, it's better. Updated.

docs/content/maintenance/system-tables.md
```diff
@@ -406,4 +406,23 @@ SELECT * FROM T$statistics;
...
+
+You can streaming or batch query the binlog through binlog table. In this system table,
+the update before and update after will be packed in one row. 
+
+```
+/*
++------------------+----------------------+-----------------------+
+|     rowkind      |     column_0         |     column_1          |
++------------------+----------------------+-----------------------+
+|        +I        |  [col_0, null]       |   [col_1, null]       |
```
JingsongLi, line 419: Why not just one element?

paimon-core/src/main/java/org/apache/paimon/table/system/BinlogTable.java
```diff
@@ -0,0 +1,697 @@
...
+/**
+ * A {@link Table} for reading binlog of table. The binlog format is as below.
+ *
+ * <p>INSERT: [+I, [co1, null], [col2, null]]
+ *
+ * <p>UPDATE: [+U, [co1_ub, col1_ua], [col2_ub, col2_ua]]
+ *
+ * <p>DELETE: [-D, [co1, null], [col2, null]]
+ */
+public class BinlogTable implements DataTable, ReadonlyTable {
```
JingsongLi, line 96: Can we reuse code with `AuditLogTable`? Like extending it to reuse something.

paimon-core/src/main/java/org/apache/paimon/table/system/BinlogTable.java
```diff
@@ -0,0 +1,697 @@
...
+
+    public static final String BINLOG = "binlog";
+
+    public static final PredicateReplaceVisitor PREDICATE_CONVERTER =
```
wwj6591812, line 100: private

paimon-core/src/main/java/org/apache/paimon/table/system/BinlogTable.java
```diff
@@ -0,0 +1,697 @@
...
+    }
+
+    @Override
+    public String name() {
+        return wrapped.name() + SYSTEM_TABLE_SPLITTER + BINLOG;
+    }
+
+    @Override
+    public RowType rowType() {
+        List<DataField> fields = new ArrayList<>();
```
wwj6591812, line 152: init map with capacity

paimon-core/src/main/java/org/apache/paimon/table/system/BinlogTable.java
wwj6591812, line 96: May be can extract some common code with the AuditLogTable system table.

paimon-flink/paimon-flink-common/pom.xml
```diff
@@ -162,6 +162,12 @@ under the License.
             <artifactId>iceberg-data</artifactId>
             <version>${iceberg.version}</version>
             <scope>test</scope>
+            <exclusions>
+                <exclusion>
+                    <artifactId>parquet-avro</artifactId>
```
wwj6591812, line 167: Why do this here?
It seems has no relation with this PR?

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/CatalogTableITCase.java
```diff
@@ -132,6 +132,40 @@ public void testSnapshotsTable() throws Exception {
                         Row.of(3L, 0L, "APPEND"));
     }
 
+    @Test
```
wwj6591812, line 135: I think add a class of BinlogTableTest in package "org.apache.paimon.table.system" may better?

docs/content/maintenance/system-tables.md
Aitozi: The schema should keep same, so this is an array only with one element.

paimon-flink/paimon-flink-common/pom.xml
Aitozi: See the first commit message

paimon-common/src/main/java/org/apache/paimon/reader/PackChangelogReader.java
```diff
@@ -0,0 +1,131 @@
...
+        @Nullable
+        @Override
+        public InternalRow next() throws IOException {
+            InternalRow row1 = nextRow();
+            if (row1 == null) {
+                return null;
+            }
+            InternalRow row2 = null;
+            if (row1.getRowKind() == RowKind.UPDATE_BEFORE) {
+                row1 = serializer.copy(row1);
```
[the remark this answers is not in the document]
Aitozi: The underlay's row fetched from the `nextRow` could be reused.

docs/content/maintenance/system-tables.md
JingsongLi: I mean one element, not one element and a null, two elements.

Aitozi: Get it, updated.

Aitozi:
Addressed the comments, pls take a look again @JingsongLi @wwj6591812

JingsongLi (approved):
+1

PR #4522, opened 2024-11-13
[hive] Improve paimon format table conversion hive table in hive catalog.

zhuangchong:
### Purpose

Improve paimon format table conversion hive table in hive catalog.

1.Remove duplicate codes, such as the setting of serdeParamters of csv
2.Put paimon formate table conversion hive table and hive table conversion paimon formate table into a class. They are related and will be easier to find and modify at the same time in subsequent code changes.
3.Extract public tool class HiveFormat


JingsongLi (changes requested):
Thanks @zhuangchong for the contribution, but I think this actually increases complexity, as can be seen from the number of lines of code. This may increase the risk of some bugs.

zhuangchong:
> Thanks @zhuangchong for the contribution, but I think this actually increases complexity, as can be seen from the number of lines of code. This may increase the risk of some bugs.

@JingsongLi 
I optimized the code and used the FormatTable.Format enumeration class instead of strings. Please help with code review.

JingsongLi (approved):
+1

PR #4523, opened 2024-11-13
[orc] Add type id to orc files

yuzelin:
### Purpose

Same to #4362

### DIFF
--- a/paimon-format/src/main/java/org/apache/paimon/format/orc/reader/OrcSplitReaderUtil.java
+++ b/paimon-format/src/main/java/org/apache/paimon/format/orc/reader/OrcSplitReaderUtil.java
@@ -18,8 +18,10 @@
 
 package org.apache.paimon.format.orc.reader;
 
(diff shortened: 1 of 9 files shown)


paimon-format/src/main/java/org/apache/paimon/format/orc/reader/OrcSplitReaderUtil.java
```diff
@@ -32,23 +34,41 @@
 /** Util for orc types. */
 public class OrcSplitReaderUtil {
 
-    public static TypeDescription toOrcType(DataType type) {
+    public static final String PAIMON_ORC_FIELD_ID_KEY = "paimon.field.id";
```
JingsongLi, line 37: paimon.id?

JingsongLi (commented):
Can you also modify convert orc type to paimon type, and check field id?

JingsongLi (approved):
+1

PR #4528, opened 2024-11-14
[flink] Support updating row type nested in array/map in Flink

tsreaper:
### Purpose

This PR supports updating row type nested in array/map in Flink. For example, we can update a field of type `ARRAY<MAP<INT, ROW<f1 INT, f2 DOUBLE>>>` to type `ARRAY<MAP<INT, ROW<f3 STRING, f1 INT>>>` with the `ALTER TABLE` SQL.

### Tests

Unit tests and IT cases.

### API and Format

No format changes.

### Documentation

Document will be added later.


JingsongLi (approved):
+1

PR #4533, opened 2024-11-14
[parquet] Fix that cannot read parquet ROW<DECIMAL> data

yuzelin:
### Purpose

In `NestedColumnReader#readRow`, we need to know the length of nested vectors. Generally, it can be gotten from `AbstractHeapVector` but `ParquetDecimalVector` is not an `AbstractHeapVector`. This PR fix it.

Also remove two unused classes: `RowColumnReader`, `RowPosition`.


JingsongLi (approved):
+1

PR #4534, opened 2024-11-15
[core] Drop stats in manifest file reading

JingsongLi:
### Purpose

We planned to remove the statistical information when reading the file, but this optimization was hindered by the need for statistical information later for whole bucket filter.

In this PR, we refactor this logical to drop stats and compute filter selected result, in this way, we can keep whole bucket filter.


wwj6591812:
Very thanks for @JingsongLi do this work!
Our company needs this feature now!

wwj6591812:
+1

leaves12138 (approved):
+1

Issue #4521, opened 2024-11-13
[Feature] Determine in advance whether there are changed fields, otherwise the latest schama information will not be obtained.


GangYang-HX:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Motivation

Optimize the logic of org.apache.paimon.flink.sink.cdc.UpdatedDataFieldsProcessFunctionBase#extractSchemaChanges: prioritize whether updatedDataFields is empty to avoid accessing the latest schema information every time

### Solution

prioritize whether updatedDataFields is empty to avoid accessing the latest schema information every time

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

GangYang-HX:
**Background**
![image](https://github.com/user-attachments/assets/b6a22705-1310-438a-b070-c7089448b68c)
![image](https://github.com/user-attachments/assets/a2332d02-d4a3-47cb-922d-40d1ce4b9c7b)
Here, each thread maintains a set of field information separately. If different threads process the same field one after another, the shema change check will be triggered. In this case, the latest schame information will be frequently obtained, resulting in a decrease in the overall throughput of the task, which will cause subsequent exceptions such as checkpoint failure.

**Solution**
Maintain a state cache for the latest shema information to avoid direct access to the file system.

GangYang-HX:
For example, the Paimon table has 1500 fields, the Parallelism of the Write operator is 500, and the task is restarted. In extreme cases, it will trigger 1500*500 calls to the latest schema information. If each call takes 20ms, the total time is: 1500*500*30ms=6.25h. This will greatly affect the throughput of the task.

PR #4535, opened 2024-11-15
[paimon-flink-cdc] Add the latest_schema state at schema evolution operator ，Reduce the latest schema access frequency

GangYang-HX:
### Purpose

Linked issue: [Issue-4521](https://github.com/apache/paimon/issues/4521)

In scenarios where the number of Paimon table fields is large and the Write concurrency is high, reduce the Latest-Schema access frequency to improve the throughput of job cold start

### Tests
Case-1: Observe whether the checkpoint time of schema evolution changes
image
image
image
Conclusion: After optimization, Schema Evolution is basically completed in seconds, or even milliseconds.

Case-2: Observe the log to see if there are still a large number of read schema behaviors
image
Conclusion: From hundreds of thousands to 115 times

### API and Format
org.apache.paimon.flink.sink.cdc.UpdatedDataFieldsProcessFunction#processElement

### Documentation
Before the Schema Evolution operator calls org.apache.paimon.flink.sink.cdc.UpdatedDataFieldsProcessFunctionBase#extractSchemaChanges, add a judgment to confirm whether the field update really needs to be triggered.
1. Add a List variable to determine whether it is an updated column: List<DataField> latestSchemaList
2. Add a state ListState. When the task is restored from the state, it is directly restored from here: ListState<DataField> latestSchemaListState

### DIFF
--- a/paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/sink/cdc/UpdatedDataFieldsProcessFunction.java
+++ b/paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/sink/cdc/UpdatedDataFieldsProcessFunction.java
@@ -54,7 +54,7 @@ public class UpdatedDataFieldsProcessFunction
     private final Identifier identifier;
 
     private ListState<DataField> latestSchemaListState;
-    private List<DataField> latestSchemaList;
+    private final List<DataField> latestSchemaList;
 
     public UpdatedDataFieldsProcessFunction(
             SchemaManager schemaManager, Identifier identifier, Catalog.Loader catalogLoader) {
@@ -79,7 +79,7 @@ public class UpdatedDataFieldsProcessFunction
                 extractSchemaChanges(schemaManager, actualUpdatedDataFields)) {
             applySchemaChange(schemaManager, schemaChange, identifier);
         }
-        actualUpdatedDataFields.forEach(field -> latestSchemaList.add(field));
+        latestSchemaList.addAll(actualUpdatedDataFields);
     }
 
     @Override
@@ -96,10 +96,10 @@ public class UpdatedDataFieldsProcessFunction
                                 new ListStateDescriptor<>(
                                         "latest-schema-list-state", DataField.class));
         if (context.isRestored()) {
-            latestSchemaListState.get().forEach(dataField -> latestSchemaList.add(dataField));
+            latestSchemaListState.get().forEach(latestSchemaList::add);
         } else {
             RowType oldRowType = schemaManager.latest().get().logicalRowType();
-            oldRowType.getFields().forEach(dataField -> latestSchemaList.add(dataField));
+            latestSchemaList.addAll(oldRowType.getFields());
         }
     }
 
--- a/paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/sink/cdc/UpdatedDataFieldsProcessFunction.java
+++ b/paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/sink/cdc/UpdatedDataFieldsProcessFunction.java
@@ -26,8 +26,6 @@ import org.apache.paimon.types.DataField;
 import org.apache.paimon.types.RowType;
 
 import org.apache.commons.collections.CollectionUtils;
-import org.apache.flink.api.common.state.ListState;
-import org.apache.flink.api.common.state.ListStateDescriptor;
 import org.apache.flink.runtime.state.FunctionInitializationContext;
 import org.apache.flink.runtime.state.FunctionSnapshotContext;
 import org.apache.flink.streaming.api.checkpoint.CheckpointedFunction;
@@ -53,7 +51,6 @@ public class UpdatedDataFieldsProcessFunction
 
     private final Identifier identifier;
 
-    private ListState<DataField> latestSchemaListState;
     private final List<DataField> latestSchemaList;
 
     public UpdatedDataFieldsProcessFunction(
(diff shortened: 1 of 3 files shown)


paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/sink/cdc/UpdatedDataFieldsProcessFunction.java
```diff
@@ -37,25 +46,65 @@
...
+            return;
+        }
+        for (SchemaChange schemaChange :
+                extractSchemaChanges(schemaManager, actualUpdatedDataFields)) {
             applySchemaChange(schemaManager, schemaChange, identifier);
         }
+        actualUpdatedDataFields.forEach(field -> latestSchemaList.add(field));
+    }
+
+    @Override
```
wwj6591812, line 85: please add a ut

paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/sink/cdc/UpdatedDataFieldsProcessFunction.java
```diff
@@ -37,25 +46,65 @@
...
-        extends UpdatedDataFieldsProcessFunctionBase<List<DataField>, Void> {
+        extends UpdatedDataFieldsProcessFunctionBase<List<DataField>, Void>
+        implements CheckpointedFunction {
 
     private final SchemaManager schemaManager;
 
     private final Identifier identifier;
 
+    private ListState<DataField> latestSchemaListState;
+    private List<DataField> latestSchemaList;
```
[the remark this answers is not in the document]
GangYang-HX: latestSchemaListState cannot be set to final because it cannot be initialized in the constructor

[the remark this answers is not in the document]
wwj6591812: I mean latestSchemaList 'latestSchemaList' may be 'final', not latestSchemaListState.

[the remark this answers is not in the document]
GangYang-HX: ok,thanks

JingsongLi:
Hi @GangYang-HX , do we really need to put the schema to the state? Maybe just a field in memory is OK?

paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/sink/cdc/UpdatedDataFieldsProcessFunction.java
```diff
@@ -37,25 +44,54 @@
  * be 1.
  */
 public class UpdatedDataFieldsProcessFunction
-        extends UpdatedDataFieldsProcessFunctionBase<List<DataField>, Void> {
+        extends UpdatedDataFieldsProcessFunctionBase<List<DataField>, Void>
+        implements CheckpointedFunction {
```
JingsongLi, line 48: Can you just use `open(Configuration parameters)`?

paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/sink/cdc/UpdatedDataFieldsProcessFunction.java
```diff
@@ -37,25 +44,54 @@
...
 
     private final SchemaManager schemaManager;
 
     private final Identifier identifier;
 
+    private final List<DataField> latestSchemaList;
```
JingsongLi, line 54: @Nullable
private List<DataField> latestSchemaList

paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/sink/cdc/UpdatedDataFieldsProcessFunction.java
```diff
@@ -37,25 +44,54 @@
...
+    }
+
+    @Override
+    public void initializeState(FunctionInitializationContext context) throws Exception {
+        RowType oldRowType = schemaManager.latest().get().logicalRowType();
+        latestSchemaList.addAll(oldRowType.getFields());
+    }
+
+    private boolean dataFieldContainIgnoreId(DataField dataField) {
+        return latestSchemaList.stream()
```
JingsongLi, line 94: if (latestSchemaList == null) {
     RowType rowType = schemaManager.latest().get().logicalRowType();
     latestSchemaList.addAll(rowType.getFields());
}

paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/sink/cdc/UpdatedDataFieldsProcessFunction.java
JingsongLi, line 54: Maybe it can be:
```
Set<FieldIdentifier> latestFields;

FieldIdentifier {
    String name;
    DataType type;
    String description;
}
```

paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/sink/cdc/UpdatedDataFieldsProcessFunction.java
```diff
@@ -43,19 +50,51 @@ public class UpdatedDataFieldsProcessFunction
...
+        for (SchemaChange schemaChange :
+                extractSchemaChanges(schemaManager, actualUpdatedDataFields)) {
             applySchemaChange(schemaManager, schemaChange, identifier);
         }
+        /**
+         * Here, actualUpdatedDataFields cannot be used to update latestFields because there is a
+         * non-SchemaChange.AddColumn scenario. Otherwise, the previously existing fields cannot be
+         * modified again.
+         */
+        updateLatestFields();
```
JingsongLi, line 85: Can we just add updatedDataFields to latestFields?

GangYang-HX: **No**, this is the logic in previous versions, but there are risks in actual testing.
**Reason**: FieldIdentifier only guarantees the uniqueness of <name, type, description>. If any of the attributes is adjusted repeatedly, it will lead to misjudgment.
**For example**: Field A's type is int, and then it is changed to string. If you want to adjust it to int again, it will not work,because latestFields has already saved an element like <A, int, description>.

JingsongLi:
+1

PR #4536, opened 2024-11-15
[Bug] Skip overwriting the partition when no data append for dynamic …

Aitozi:
…partition overwrite mode.

### Purpose

If `dynamic-partition-overwrite` is enabled, the partition should not be overwrite when the input data is empty. Currently, we check the `appendTableFiles`, but it contains the `DELETE` files. We should exclude the `DELETE` files for check.

Linked issue: close #xxx

### DIFF
--- a/paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/BatchFileStoreITCase.java
+++ b/paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/BatchFileStoreITCase.java
@@ -84,6 +84,25 @@ public class BatchFileStoreITCase extends CatalogITCaseBase {
         assertThat(batchSql("SELECT * FROM T")).isEmpty();
     }
 
+    @ParameterizedTest
+    @ValueSource(booleans = {true, false})
+    public void testOverwritePartitionedTableEmpty(boolean dynamicPartitionOverwrite) {
+        batchSql("CREATE TABLE T2 (a INT, b INT, c INT) PARTITIONED BY (a) ");
+        batchSql("INSERT INTO T2 VALUES (1, 11, 111), (2, 22, 222)");
+        assertThat(batchSql("SELECT * FROM T2"))
+                .containsExactlyInAnyOrder(Row.of(1, 11, 111), Row.of(2, 22, 222));
+        batchSql(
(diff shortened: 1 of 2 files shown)


paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/BatchFileStoreITCase.java
```diff
@@ -84,6 +84,25 @@ public void testOverwriteEmpty() {
         assertThat(batchSql("SELECT * FROM T")).isEmpty();
     }
 
+    @ParameterizedTest
+    @ValueSource(booleans = {true, false})
+    public void testOverwritePartitionedTableEmpty(boolean dynamicPartitionOverwrite) {
+        batchSql("CREATE TABLE T2 (a INT, b INT, c INT) PARTITIONED BY (a) ");
```
wwj6591812, line 90: Why use TABLE T2 instead of TABLE T which created in BatchFileStoreITCase#ddl?

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/BatchFileStoreITCase.java
```diff
@@ -84,6 +84,25 @@ public void testOverwriteEmpty() {
...
+        batchSql("INSERT INTO T2 VALUES (1, 11, 111), (2, 22, 222)");
+        assertThat(batchSql("SELECT * FROM T2"))
+                .containsExactlyInAnyOrder(Row.of(1, 11, 111), Row.of(2, 22, 222));
+        batchSql(
+                "INSERT OVERWRITE T2 /*+ OPTIONS('dynamic-partition-overwrite' = '%s') */"
```
wwj6591812, line 95: I don't think this UT covers the situation that appendTableFiles contains some DELETE files.

wwj6591812:
Thanks @Aitozi 
Left two small comments.

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/BatchFileStoreITCase.java
Aitozi: It will cover. But, I take an another thought, I think the current behavior is acceptable, It's consistent for the partitioned and unpartitioned table when insert overwrite with false. So, closing it now.

Issue #4496, opened 2024-11-11
[Feature] View support with Spark


Zouxxyy:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Motivation

View support with Spark

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

PR #4538, opened 2024-11-18
[spark] Introduce View Support to SparkCatalog

Zouxxyy:
### Purpose

To #4496, implement the following basic view usages for SparkCatalog:

```sql
CREATE VIEW v xxx;
CREATE OR REPLACE VIEW v xxx; -- non-atomic, just drop then create
SHOW VIEWS;
SHOW VIEWS LIKE '';
SELECT FROM view;
```


JingsongLi (approved):
Good job!

JingsongLi:
Thanks @Zouxxyy

PR #4539, opened 2024-11-18
[core] Fix that batch unaware bucket compact cannot stop when there is no snapshot

yuzelin:


xuzifu666:
+1，Maybe need a ut case is better.

JingsongLi (approved):
+1

PR #4541, opened 2024-11-18
[flink] change env.fromData to fromElements for compatibility with flink 1.18 and lower versions

LsomeYeah:
### Purpose

Linked issue: close #xxx

In `FlinkOrphanFilesClean#doOrphanClean`, `env.fromData` is used for collecting information. While `env.fromData` was introduced in Flink 1.19, so it'll throw `NoSuchMethodError` when used in Flink 1.18- version. This pr changed `env.fromData` to `env.fromElements` to be compatible with lower flink versions.


JingsongLi (approved):
+1

PR #4544, opened 2024-11-18
[hive] Sync db owner prop to hms when creating the db

Zouxxyy:
### Purpose

Sync db owner prop to hms when creating the db


JingsongLi (approved):
+1

PR #4545, opened 2024-11-19
[Core][Flink][Spark] Add deletedFileTotalLenInBytes in result of OrphanFilesClean

wwj6591812:
### Purpose

In my company, after we run a RemoveOrphanFilesAction, we not only want to know the number of orphan files that have been cleared, but also want to know how much capacity has been cleared.

This pr add deletedFileTotalLenInBytes in result of OrphanFilesClean.

Linked issue: close #xxx

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/LocalOrphanFilesClean.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/LocalOrphanFilesClean.java
@@ -21,12 +21,12 @@ package org.apache.paimon.operation;
 import org.apache.paimon.CoreOptions;
 import org.apache.paimon.catalog.Catalog;
 import org.apache.paimon.catalog.Identifier;
-import org.apache.paimon.fs.FileStatus;
 import org.apache.paimon.fs.Path;
 import org.apache.paimon.manifest.ManifestEntry;
 import org.apache.paimon.manifest.ManifestFile;
 import org.apache.paimon.table.FileStoreTable;
 import org.apache.paimon.table.Table;
+import org.apache.paimon.utils.Pair;
 import org.apache.paimon.utils.SerializableConsumer;
 
 import javax.annotation.Nullable;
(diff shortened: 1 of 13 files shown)


xuzifu666 (commented):
+1

wwj6591812:
@JingsongLi 
Hi, please cc, Thx.
The failed UT was not caused by my modifications.

paimon-core/src/main/java/org/apache/paimon/operation/LocalOrphanFilesClean.java
```diff
@@ -87,16 +90,27 @@ public LocalOrphanFilesClean(
...
         List<String> branches = validBranches();
 
         // specially handle to clear snapshot dir
-        cleanSnapshotDir(branches, deleteFiles::add);
+        cleanSnapshotDir(
+                branches,
+                deleteFiles::add,
+                p -> {
+                    try {
+                        deletedFilesSizeInBytes.addAndGet(fileIO.getFileSize(p));
```
JingsongLi, line 103: getFileSize should always be together with delete, otherwise, I feel there may be issues with execution efficiency, getFileSize should be executed parallelismly too.

wwj6591812: > getFileSize should always be together with delete, otherwise, I feel there may be issues with execution efficiency, getFileSize should be executed parallelismly too.

@JingsongLi Very thanks for your suggestions.

I would like to confirm with you the code improvement plan:
Are you means that getFileStatus execute only once in OrphanFilesClean#createFileCleaner, and fileCleaner can direct return the size of deleted file?

JingsongLi: I just think about it, maybe we don't need to invoke `getFileSize`, the files come from `listStatus`, it already contains file size.

wwj6591812: OK, I will have a try.
Thx

wwj6591812: > I just think about it, maybe we don't need to invoke `getFileSize`, the files come from `listStatus`, it already contains file size.

@JingsongLi 
Hi，please cc, THX.

JingsongLi (approved):
Thanks @wwj6591812 , looks good to me!

Issue #4507, opened 2024-11-13
[Feature] Supports specifying hive environment parameters when altering table


GangYang-HX:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Motivation

When using the Paimon Schema Evolution function, the alertTable method of Hive will be called to add fields, and the statistical information of the Hive table will be updated by default. In fact, Paimon will use its own statistical information instead of Hive's, so it can be turned off in scenarios where there are many fields in the Paimon table and they are updated frequently to avoid a sharp increase in HMS memory.
![image](https://github.com/user-attachments/assets/678ee2fe-a27b-4569-b08f-034a43e43dbc)

### Solution

When alterTable is triggered, the method called is changed to alter_table_with_environmentContext instead of alter_table. In this way, you can control whether to update table stats by setting the DO_NOT_UPDATE_STATS parameter in the EnvironmentContext object

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

PR #4549, opened 2024-11-19
[Hive]To avoid excessive HMS memory usage, when executing AlterTable for a Paimon table containing a large number of fields

GangYang-HX:
### Purpose

When metastore=hive is set, if the Paimon table has many fields (1600+), the MetaStoreUtils.updateTableStatsSlow() method will be called by default to update the table statistics, which will cause a sharp increase in HMS memory usage.
In fact, Paimon has its own statistics and does not rely on Hive's statistics.

So, an option is provided to control whether to update Hive table statistics to avoid excessive HMS memory usage.

Linked issue: [Issue-4507](https://github.com/apache/paimon/issues/4507)

### API and Format

org.apache.paimon.hive.HiveCatalog#alterTableToHms


xuzifu666:
Please change your title 【Hive】to [Hive] and make your code style with format, you can execute 'mvn spotless:apply' for it

GangYang-HX:
> Please change your title 【Hive】to [Hive] and make your code style with format, you can execute 'mvn spotless:apply' for it

Thank you. All the tests can be passed locally, but there are still errors now. It seems that the errors have nothing to do with my changes.

JingsongLi:
+1

PR #4552, opened 2024-11-20
[core] Remove FileStoreScan.withManifestList and fix unstable test

JingsongLi:
### Purpose

- FileStoreScan.withManifestList is only used by test code, we should remove it.
- Fix unstable test ExpireSnapshotsTest.testChangelogOutLivedSnapshot


leaves12138 (approved):
+1

PR #4555, opened 2024-11-20
[core] all catalogs use 'allow-upper-case' to control case-sensitive

LsomeYeah:
### Purpose

Linked issue: close #xxx

FileSystemCatalog now use 'case-sensitive' to control case-sensitive, which is not compatible with other catalog. Make FileSystemCatalog can control case-sensitive with 'allow-upper-case' too.
### Tests

`FileSystemCatalogTest#testCreateTableAllowUpperCase`


JingsongLi (approved):
+1

PR #4558, opened 2024-11-20
[flink] Fix that dim table cannot refresh overwrite changes

yuzelin:
### Purpose

Linked issue: close #xxx

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/source/DataTableStreamScan.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/source/DataTableStreamScan.java
@@ -196,7 +196,7 @@ public class DataTableStreamScan extends AbstractDataTableScan implements Stream
 
             // first try to get overwrite changes
             if (snapshot.commitKind() == Snapshot.CommitKind.OVERWRITE) {
-                SnapshotReader.Plan overwritePlan = tryGetOverwirtePlan(snapshot);
+                SnapshotReader.Plan overwritePlan = handleOverwriteSnapshot(snapshot);
                 if (overwritePlan != null) {
                     nextSnapshotId++;
                     return overwritePlan;
@@ -229,7 +229,7 @@ public class DataTableStreamScan extends AbstractDataTableScan implements Stream
     }
 
     @Nullable
-    protected SnapshotReader.Plan tryGetOverwirtePlan(Snapshot snapshot) {
+    protected SnapshotReader.Plan handleOverwriteSnapshot(Snapshot snapshot) {
         if (supportStreamingReadOverwrite) {
             LOG.debug("Find overwrite snapshot id {}.", nextSnapshotId);
             SnapshotReader.Plan overwritePlan =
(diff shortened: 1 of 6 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/LookupDataTableScan.java
```diff
@@ -58,6 +62,16 @@ public LookupDataTableScan(
         this.lookupScanMode = lookupScanMode;
     }
 
+    @Override
+    @Nullable
+    protected SnapshotReader.Plan tryGetOverwirtePlan(Snapshot snapshot) {
```
wwj6591812, line 67: tryGetOverwritePlan

paimon-core/src/main/java/org/apache/paimon/table/source/DataTableStreamScan.java
```diff
@@ -228,6 +228,18 @@ private boolean shouldDelaySnapshot(long snapshotId) {
         return false;
     }
 
+    @Nullable
+    protected SnapshotReader.Plan tryGetOverwirtePlan(Snapshot snapshot) {
```
wwj6591812, line 232: tryGetOverwritePlan

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/FileStoreLookupFunction.java
```diff
@@ -249,7 +250,7 @@ public Collection<RowData> lookup(RowData keyRow) {
                 rows.add(new FlinkRowData(matchedRow));
             }
             return rows;
-        } catch (OutOfRangeException e) {
+        } catch (OutOfRangeException | DimScanOverwriteException e) {
```
wwj6591812, line 253: Why catch DimScanOverwriteException then reopen can solve this problem.?

JingsongLi, line 253: Create a `ReopenException`.

paimon-core/src/main/java/org/apache/paimon/table/source/DataTableStreamScan.java
JingsongLi, line 232: `handleOverwriteSnapshot`.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/FileStoreLookupFunction.java
yuzelin: Because the StreamScan doesn't read OVERWRITE snapshot (in most cases), so that dim table cannot refresh data when you overwrite the dim table. Here we just reopen the lookup table to reload the full data of dim table.

wwj6591812: very thanks

JingsongLi:
+1

PR #4559, opened 2024-11-20
support aliyun-json

### LINKED ISSUE #4529: [Feature] Support the json format which is used by the data integration of Aliyun Dataworks
### Search before asking

### Motivation

We use aliyun dataworks and it's data integration to collect data from mysql, oracle etc. in realtime, then save the data into paimon with the paimon-flink-action jar.
But the jar does not support the json format which is used by aliyun data integration.

### Solution

Add new cdc json data parser for aliyun data integration.

### Anything else?

_No response_

JackeyLee007:
[flink]
### Purpose

Linked issue: close #4529 

To support the json format of aliyun Data Integration, DI in short.
The data in kafka collected by DI from mysql or oracle, are not in standard format. Not canal-json or debezium json. I would like call it aliyun-json.

We want to sink the data to paimon from kafka directly, with the kafka_sync_database/table action. So the aliyun-json must be supported, and parsed into cdc record.

### Tests

Supplied with the commit.
### API and Format

No.
### Documentation

When used to process the data collected by ID, with paimon-flink-action, it just need to specify the json format as the following:
```shell
... paimon-flink-action-<version>.jar
kafka_sync_database 
...
--kafka_conf value.format=aliyun-json 
...
```

### DIFF
--- a/paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/action/cdc/format/aliyun/AliyunDataFormat.java
+++ b/paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/action/cdc/format/aliyun/AliyunDataFormat.java
@@ -1,0 +1,34 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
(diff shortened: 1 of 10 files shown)


JingsongLi:
Merged into master, close this one.

Issue #4556, opened 2024-11-20
[Bug] 


GangYang-HX:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Paimon version

Paimon-0.8.1

### Compute Engine

Flink-1.18.1

### Minimal reproduce step

1. Start a Spark offline task containing a large number of tasks to read the Paimon table data
2. During the offline task, add a new field
Not necessarily displayed, there is a high probability！！！

### What doesn't meet your expectations?

The alterTable operation is not atomic. When reading the Paimon table data, the Hive field and Paimon latest-schema information will be checked. There is a certain probability that they will not match and eventually cause query exceptions.

`Hive DDL and paimon schema mismatched! It is recommended not to write any column definition as Paimon external table can read schema from the specified location.
There are 1665 fields in Hive DDL: id, sticky_album_id ......
There are 1666 fields in Paimon schema: id, sticky_album_id ......
	at org.apache.paimon.hive.HiveSchema.checkFieldsMatched(HiveSchema.java:249)
	at org.apache.paimon.hive.HiveSchema.extract(HiveSchema.java:165)
	at org.apache.paimon.hive.PaimonStorageHandler.getDataFieldsJsonStr(PaimonStorageHandler.java:89)
	at org.apache.paimon.hive.PaimonStorageHandler.configureInputJobProperties(PaimonStorageHandler.java:84)
	at org.apache.spark.sql.hive.HiveTableUtil$.configureJobPropertiesForStorageHandler(TableReader.scala:438)
	at org.apache.spark.sql.hive.HadoopTableReader$.initializeLocalJobConfFunc(TableReader.scala:468)
	at org.apache.spark.sql.hive.HadoopTableReader.$anonfun$createOldHadoopRDD$1(TableReader.scala:354)
	at org.apache.spark.sql.hive.HadoopTableReader.$anonfun$createOldHadoopRDD$1$adapted(TableReader.scala:354)
	at org.apache.spark.rdd.HadoopRDD.$anonfun$getJobConf$8(HadoopRDD.scala:184)
	at org.apache.spark.rdd.HadoopRDD.$anonfun$getJobConf$8$adapted(HadoopRDD.scala:184)
	at scala.Option.foreach(Option.scala:407)
	at org.apache.spark.rdd.HadoopRDD.$anonfun$getJobConf$6(HadoopRDD.scala:184)
	at scala.Option.getOrElse(Option.scala:189)
[...]

PR #4561, opened 2024-11-21
[Hive] Fix Hive DDL and paimon schema mismatched bug

GangYang-HX:
### Purpose
Linked issue: [issue-4556](https://github.com/apache/paimon/issues/4556)

HiveCatalog add column is divided into two stages: first modify the paimon schema, then modify HMS. There will be a time difference between the two stages. If there is a task read (based on hivecatalog) during this time, the paimon schema and hive schema check will fail in the getDataFieldsJsonStr link.

Solution: Actually, you only need to ensure that the Hive DDL is a subset of the paimon schema.

### Tests
org.apache.paimon.hive.HiveTableSchemaTest#testSupersetColumnNameAndType
org.apache.paimon.hive.HiveTableSchemaTest#testSubsetColumnNameAndType

### API and Format
org.apache.paimon.hive.HiveSchema#checkPartitionMatched

### Documentation
only need to ensure that the Hive DDL is a subset of the paimon schema.

### DIFF
--- a/paimon-hive/paimon-hive-connector-common/src/main/java/org/apache/paimon/hive/HiveSchema.java
+++ b/paimon-hive/paimon-hive-connector-common/src/main/java/org/apache/paimon/hive/HiveSchema.java
@@ -233,9 +233,10 @@ public class HiveSchema {
             }
         }
 
-        if (schemaFieldNames.size() != hiveFieldNames.size()) {
+        // It is OK that hive is a subset of paimon
+        if (schemaFieldNames.size() < hiveFieldNames.size()) {
             throw new IllegalArgumentException(
-                    "Hive DDL and paimon schema mismatched! "
+                    "Hive DDL is a superset of paimon schema! "
                             + "It is recommended not to write any column definition "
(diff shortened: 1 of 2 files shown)


paimon-hive/paimon-hive-connector-common/src/main/java/org/apache/paimon/hive/HiveSchema.java
```diff
@@ -233,9 +233,10 @@ private static void checkFieldsMatched(
             }
         }
 
-        if (schemaFieldNames.size() != hiveFieldNames.size()) {
+        // It is OK that hive is a subset of paimon
+        if (schemaFieldNames.size() < hiveFieldNames.size()) {
```
ZhangChaoming, line 237: Checking schema filed names is a subset of paimon's just by List#size() is not precise. Use Set#containsAll() maybe better. :)

JingsongLi: The following code will be checked one by one, I think it is OK

PR #4571, opened 2024-11-21
[hive] Determine timestamp ltz type based on hive runtime version

Zouxxyy:
### Purpose

Determine timestamp ltz type based on hive runtime version, because we will only shade `paimon-hive-common` into the `paimon-bundle`, so it should be hive version independent

### Tests

Cover by the existing cases


zhuangchong (approved):
+1

JingsongLi:
+1

PR #4574, opened 2024-11-22
[core] Ensure triggerCompaction in RecordWriter.isCompacting

JingsongLi:
### Purpose

RecordWriter.isCompacting: Check if a compaction is in progress, or if a compaction result remains to be fetched.

It is better to try to trigger compact before check if a compaction is in progress. Otherwise, there may be residual potential incomplete compacts.


yuzelin (approved):
+1

PR #4575, opened 2024-11-22
[core] Improve object table for fileIO, Privileged and parent_path

JingsongLi:
### Purpose

Improve object table:
1. FileIO can be customized in AbstractCatalog.
2. Privileged should handle ObjectTable.
3. Add parent_path in Object table schema.


yanbinyang:
+1

Issue #4442, opened 2024-11-04
[Feature] Avoid deprecated APIs removed in Flink 2.0 Preview


yunfengzhou-hub:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Motivation

Flink 2.0 Preview has been released, and Paimon needs to make itself compatible with 2.0 Preview now, to be better prepared for the Flink 2.0 in future.

### Solution

A bunch of public APIs that had been marked as @Deprecated are now removed in Flink 2.0 Preview. Paimon needs to replace usages of these deprecated APIs to the latest alternatives.

### Anything else?

Flink 2.0 still have more work to do with its internal implementation, and only dealt with its public APIs in 2.0 Preview. So if Paimon has been using Flink's public methods that are not marks as @Public or @PublicEvolving, the usages might still be compatible with 2.0 Preview for now, but should also be updated to alternatives as early as possible.

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

yunfengzhou-hub:
Apart from deprecated **public** APIs, Paimon has also been using **internal** public methods or private methods through reflection. These usages bring risks to Paimon's compatibility to Flink as these internal implementations might be modified at any time. They should also be replaced as part of this issue.

wwj6591812:
Good work!

mlegore:
Hi, I believe that there are still references to SinkFunction that can cause  java.lang.NoClassDefFoundError: org/apache/flink/streaming/api/functions/sink/SinkFunction. When you try to use FlinkSink/FlinkSinkFactory, it will have that issue for me.

PR #4567, opened 2024-11-21
[flink] Avoid deprecated APIs removed in Flink 2.0 Preview

yunfengzhou-hub:
### Purpose

Linked issue: close #4442

This PR removes usages of deprecated APIs and make paimon-flink modules better prepared for Flink 2.0 formal release.

PR #4578, opened 2024-11-25
[flink][hotfix] Wait for consumer reset before job close

yunfengzhou-hub:
### Purpose

Linked issue: #4442 

This PR fixes the bug that Consumers might not have been reset yet when the streaming job is closed in test cases. This is because the resetting logic happens in `MonitorFunction#notifyCheckpointComplete`, and Flink does not guarantee the invocation of this method when the job is canceled or stopped.

This bug happens rarely in the current test cases because Flink's legacy source function runs independently in a separated thread. When we use FLIP-27 source for MonitorFunction and checkpoint events may pile up in Flink's Mailbox, this bug will be triggered more frequently.

### Tests

### API and Format

### Documentation

### DIFF
--- a/paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/BranchSqlITCase.java
+++ b/paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/BranchSqlITCase.java
@@ -27,6 +27,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.flink.types.Row;
 import org.apache.flink.util.CloseableIterator;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.Timeout;
 
 import java.io.IOException;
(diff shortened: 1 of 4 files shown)


wwj6591812:
Hi, @yunfengzhou-hub .
You means we will use FLIP-27 source to rewrite MonitorFunction?
If yes, I have not see this change in https://github.com/apache/paimon/pull/4567/files.

yunfengzhou-hub:
> Hi, @yunfengzhou-hub . You means we will use FLIP-27 source to rewrite MonitorFunction? If yes, I have not see this change in https://github.com/apache/paimon/pull/4567/files.

Hi @wwj6591812 . Yes we will rewrite it. In the PR you mentioned, MonitorFunction is renamed to MonitorSource, as it is not a SourceFunction any more.

JingsongLi (approved):
+1

PR #4583, opened 2024-11-25
[flink] Adopt getTaskInfo() when acquiring parallelism info

yunfengzhou-hub:
### Purpose

Linked issue: #4442 

This pull request adopts RuntimeContext#getTaskInfo when acquiring parallelism information.

### Tests

### API and Format

### Documentation


JingsongLi (commented):
If you remove `RuntimeContextUtils` for old flink versions, do some tests fail?

yunfengzhou-hub:
> If you remove `RuntimeContextUtils` for old flink versions, do some tests fail?

Yes some tests will fail. In old flink versions there is no `RuntimeContext#getTaskInfo`, so we need to use `getRuntimeContext().getIndexOfThisSubtask()` instead of `getRuntimeContext().getTaskInfo().getIndexOfThisSubtask()` in these versions.

JingsongLi (approved):
+1

Issue #4586, opened 2024-11-25
[Feature]  In a Paimon primary key table, using ORC offers significantly higher efficiency for point lookups based on the primary key compared to Parquet.


Aiden-Dong:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Basic Information

table

```
{
  "version" : 2,
  "id" : 0,
  "fields" : [ {
    "id" : 0,
    "name" : "f0",
    "type" : "BIGINT NOT NULL"
  }, {
    "id" : 1,
    "name" : "f1",
    "type" : "STRING"
  }, {
    "id" : 2,
    "name" : "f2",
    "type" : "STRING"
  }, {
    "id" : 3,
    "name" : "f4",
    "type" : "FLOAT"
  }, {
    "id" : 4,
    "name" : "f5",
    "type" : "DOUBLE"
  }, {
    "id" : 5,
    "name" : "f6",
    "type" : "BOOLEAN"
  }, {
    "id" : 6,
    "name" : "f7",
    "type" : {
      "type" : "ARRAY",
      "element" : "BIGINT"
    }
  } ],
  "highestFieldId" : 6,
  "partitionKeys" : [ ],
  "primaryKeys" : [ "f0" ],
  "options" : {
    "bucket" : "1",
    "file.format" : "parquet/orc"
  },
  "timeMillis" : 1732796139692
}
```

- RowCount  : `400w`
- bucket : `1`
- FileNumber : `7`
- Ordered Write

The data sample ： 

```
1 : [1, h2,7c17184a-46d3-4565-aa87-b31316af2144, 1.000000, 1.000000, false, ([1,2,3,]))
2 : [2, h1,919fc338-5ef6-474c-a1fa-1b2407512731, 2.000000, 2.000000, true, ([2,3,4,]))
3 : [3, h3,6126ee02-8d74-4c3a-a00e-328af5871e4a, 3.000000, 3.000000, false, ([3,4,5]))
4 : [4, h3,c2b0e039-a468-4853-aaa1-57bcde5a49b6, 4.000000, 4.000000, trye, ([4,5,6]))
```

### The Write Example

```
Table table = TableUtil.getTable(); // PrimaryKeyFileStoreTable
BatchWriteBuilder writeBuilder = table.newBatchWriteBuilder();

String[] items = new String[] {"h1", "h2", "h3"};

BatchTableWrite write = writeBuilder.newWrite(); // TableWriteImpl

[...]

PR #4587, opened 2024-11-25
[Core] Optimize Parquet Reading by Enhancing Predicate Pushdown to the Page Level

Aiden-Dong:
### Purpose

Linked issue: [#4586](https://github.com/apache/paimon/issues/4586)

### DIFF
--- a/paimon-format/src/main/java/org/apache/paimon/format/parquet/ParquetReaderFactory.java
+++ b/paimon-format/src/main/java/org/apache/paimon/format/parquet/ParquetReaderFactory.java
@@ -130,7 +130,7 @@ public class ParquetReaderFactory implements FormatReaderFactory {
                 buildFieldsList(projectedType.getFields(), projectedType.getFieldNames(), columnIO);
 
         return new ParquetReader(
-                reader, requestedSchema, reader.getRecordCount(), poolOfBatches, fields);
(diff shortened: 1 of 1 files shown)


JingsongLi:
This modification is not that simple, it requires us to modify the column reader.

Aiden-Dong:
> 这个修改没那么简单，它需要我们修改列阅读器。

I found that it works in my local tests, and the column pages obtained based on FilterRowGroup are filtered. Is there any error I might have missed?

Aiden-Dong:
> This modification is not that simple, it requires us to modify the column reader.

I found that it works in my local tests, and the column pages obtained based on FilterRowGroup are filtered. Is there any error I might have missed?

Aiden-Dong:
In my local tests, I found that with 4 million rows and 3 columns (3 files), after 30 random reads, the time dropped from 14s to 1.9s, and the correct results were returned.

JingsongLi:
https://github.com/apache/paimon/pull/3610

JingsongLi:
See implementation in Spark.

JingsongLi:
Close this now, feel free to re-open if you have more questions.

Aiden-Dong:
> See implementation in Spark.

ok tks, I'll go check how parquet is implemented when reading in Spark.

PR #4592, opened 2024-11-26
[core] Improve the performance of show tables

xleoken:
### Purpose

When store the catalog in s3 object store, as the number of tables increases, the performance of the `show tables` deteriorates.

Because the `FileUtils#listVersionedFileStatus` method will send many `list` request to the s3 server.

### Tests

**List 100 tables** 

Before patch
```
test_table98
test_table99
Time taken: 68.043 seconds, Fetched 100 row(s)
```

After patch
```
test_table98
test_table99
Time taken: 2.298 seconds, Fetched 100 row(s)
```

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
@@ -653,7 +653,16 @@ public abstract class AbstractCatalog implements Catalog {
     }
 
     protected boolean tableExistsInFileSystem(Path tablePath, String branchName) {
-        return !new SchemaManager(fileIO, tablePath, branchName).listAllIds().isEmpty();
+        SchemaManager schemaManager = new SchemaManager(fileIO, tablePath, branchName);
+
+        // in order to improve the performance, check the schema-0 firstly.
+        boolean schemaZeroExists = schemaManager.schemaExists(0);
+        if (schemaZeroExists) {
+            return true;
+        } else {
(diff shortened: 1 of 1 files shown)


paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
```diff
@@ -653,7 +653,7 @@ protected List<String> listTablesInFileSystem(Path databasePath) throws IOExcept
     }
 
     protected boolean tableExistsInFileSystem(Path tablePath, String branchName) {
-        return !new SchemaManager(fileIO, tablePath, branchName).listAllIds().isEmpty();
+        return new SchemaManager(fileIO, tablePath, branchName).schemaExists(0);
```
JingsongLi, line 656: if schem-0 not exists, still use `!listAllIds().isEmpty()` to check the table.

xleoken: Updated. By the way, in what scenario would `schema-0` will be deleted.

JingsongLi: I don't know, but by design, old useless schema files can be deleted.

xleoken: > I don't know, but by design, old useless schema files can be deleted.

ok.

JingsongLi:
+1

PR #4595, opened 2024-11-26
[spark] Explicitly set spark3 profile in publish snapshot

Zouxxyy:
### Purpose

Explicitly set spark3 profile in publish snapshot, because explicit profile will override the default profile


wwj6591812:
+1

YannByron (approved):

PR #4598, opened 2024-11-27
[core] Drop stats for overwrite commit

JingsongLi:
### Purpose

Overwrite commit will read all entries, and generate DELETE entries.

We should drop stats for these DELETE entries, otherwise memory will be too big.


wwj6591812:
+1

PR #4600, opened 2024-11-27
[core] Add tests for drop stats for overwrite commit

LsomeYeah:
### Purpose

Linked issue: close #xxx

Overwrite commit will read all entries, and generate DELETE entries.

We should drop stats for these DELETE entries, otherwise memory will be too big.

### Tests

`FileStoreCommitTest#testDropStatsForOverwrite`


JingsongLi:
+1

PR #4601, opened 2024-11-27
[core] Cache statistics in AbstractFileStoreTable

JingsongLi:
### Purpose

1. Add Stats cache
2. Fix cache with Table.copy


wwj6591812:
+1

PR #4604, opened 2024-11-27
[core] Avoid extract from file footer when stats mode is none

Aitozi:
### Purpose

This could save one remote IO operation when stats mode is none

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/io/StatsCollectingSingleFileWriter.java
+++ b/paimon-core/src/main/java/org/apache/paimon/io/StatsCollectingSingleFileWriter.java
@@ -25,6 +25,7 @@ import org.apache.paimon.format.SimpleStatsCollector;
 import org.apache.paimon.format.SimpleStatsExtractor;
 import org.apache.paimon.fs.FileIO;
 import org.apache.paimon.fs.Path;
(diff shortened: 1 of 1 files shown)


wwj6591812:
+1

paimon-core/src/main/java/org/apache/paimon/io/StatsCollectingSingleFileWriter.java
```diff
@@ -43,7 +43,7 @@
 public abstract class StatsCollectingSingleFileWriter<T, R> extends SingleFileWriter<T, R> {
 
     @Nullable private final SimpleStatsExtractor simpleStatsExtractor;
-    @Nullable private SimpleStatsCollector simpleStatsCollector = null;
```
JingsongLi, line 46: Can you keep it nullable, and just create a field `isStatsDisabled` from 
```
Arrays.stream(statsCollectors)
                        .allMatch(p -> p instanceof NoneSimpleColStatsCollector)
```

Aitozi: Fixed

JingsongLi (approved):
+1

PR #4606, opened 2024-11-28
[doc] fix the ‘primary-key-table’ link

yangjf2019:
### Purpose

When we click on ‘primary-key-table’, the jump page is not accurate and we need to jump to the exact address.

### Documentation

_docs/content/flink/sql-write.md
docs/content/flink/action-jars.md_

Hi, @JingsongLi please take a look at it in your spare time,Thank you very much!


wwj6591812:
+1

JingsongLi:
+1

PR #4607, opened 2024-11-28
[core] Remove all lineage implementation

JingsongLi:
### Purpose

https://cwiki.apache.org/confluence/display/PAIMON/PIP-5%3A+Paimon+Table+And+Data+Lineage+For+Flink

It has been a year and a half since the creation of this PIP, and we lack awareness of this feature. It is still an unfinished feature, but in the code, it may mislead developers who are new to the Paimon community, leading to some misunderstandings.

So I am considering deleting the current implementation, and if necessary, we can re implement this feature based on future needs, provided that it can be fully implemented within a certain period of time.


liming30:
+1

liming30 (approved):

PR #4611, opened 2024-12-01
[flink] Avoid deprecated usage on TableSchema, DataType and DescriptorProperties

yunfengzhou-hub:
### Purpose

Linked issue: #4442

This PR removes deprecated usages about Flink Table API, mainly including TableSchema and DescriptorProperties.

### Tests

### API and Format

### Documentation

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/DataCatalogTable.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/DataCatalogTable.java
@@ -23,33 +23,55 @@ import org.apache.paimon.table.Table;
 import org.apache.paimon.types.DataField;
 
 import org.apache.flink.table.api.Schema;
-import org.apache.flink.table.api.TableColumn;
-import org.apache.flink.table.api.TableSchema;
-import org.apache.flink.table.api.constraints.UniqueConstraint;
 import org.apache.flink.table.catalog.CatalogBaseTable;
 import org.apache.flink.table.catalog.CatalogTable;
-import org.apache.flink.table.catalog.CatalogTableImpl;
 
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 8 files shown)


JingsongLi (commented):
I am concern about the serialization for watermark spec, can you create some test for compatibility? (You can create schema file from old Paimon, and read it in tests)

yunfengzhou-hub:
> I am concern about the serialization for watermark spec, can you create some test for compatibility? (You can create schema file from old Paimon, and read it in tests)

@JingsongLi `FlinkCatalogPropertiesUtilTest` has recorded the serialization result as follows
```java
Map<String, String> expected = new HashMap<>();
String watermarkPrefix = compoundKey(SCHEMA, WATERMARK, 0);
expected.put(compoundKey(watermarkPrefix, WATERMARK_ROWTIME), "test_time");
expected.put(
        compoundKey(watermarkPrefix, WATERMARK_STRATEGY_EXPR),
        "`test_time` - INTERVAL '0.001' SECOND");
expected.put(compoundKey(watermarkPrefix, WATERMARK_STRATEGY_DATA_TYPE), "TIMESTAMP(3)");
```
This code stands for the serialization result since the older versions of Paimon. This PR does not change this part of code, which means the compatibility should be guaranteed.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/DataCatalogTable.java
```diff
@@ -104,27 +124,24 @@ private Schema toSchema(TableSchema tableSchema, Map<String, String> comments) {
...
-
         return builder.build();
     }
 
     @Override
     public CatalogBaseTable copy() {
         return new DataCatalogTable(
                 table,
-                getSchema().copy(),
+                Schema.newBuilder().fromSchema(schema).build(),
```
JingsongLi, line 144: Why need to copy it? I see there is no copy in `DefaultCatalogTable`.

yunfengzhou-hub: The copy was added because the copy had been made in the existing implementation. But I agree with it that the copy is unnecessary. I'll remove it.

JingsongLi:
+1

PR #4614, opened 2024-12-02
[flink] Replace legacy SourceFunction with v2 Source

yunfengzhou-hub:
### Purpose

Linked issue: #4442

This PR replaces legacy SourceFunction with FLIP-27 Source.

### Tests

### API and Format

### Documentation


paimon-flink/paimon-flink-1.20/src/main/java/org/apache/paimon/flink/source/operator/MonitorSource.java
```diff
@@ -0,0 +1,277 @@
...
+    private transient StreamTableScan scan;
+    private transient SourceContext<Split> ctx;
+
+    private transient ListState<Long> checkpointState;
+    private transient ListState<Tuple2<Long, Long>> nextSnapshotState;
+    private transient TreeMap<Long, Long> nextSnapshotPerCheckpoint;
+
+    public MonitorSource(
+            ReadBuilder readBuilder, long monitorInterval, boolean emitSnapshotWatermark) {
+        this.readBuilder = readBuilder;
```
JingsongLi, line 102: Can you extract implementation in common, let all flink 1.x share codes.

paimon-flink/paimon-flink-1.20/src/main/java/org/apache/paimon/flink/source/operator/MonitorSource.java
```diff
@@ -0,0 +1,277 @@
...
+ *
+ * <p>The splits to be read are forwarded to the downstream {@link ReadOperator} which can have
+ * parallelism greater than one.
+ *
+ * <p>Currently, there are two features that rely on this monitor:
+ *
+ * <ol>
+ *   <li>Consumer-id: rely on this function to do aligned snapshot consumption, and ensure that all
+ *       data in a snapshot is consumed within each checkpoint.
+ *   <li>Snapshot-watermark: when there is no watermark definition, the default Paimon table will
```
JingsongLi, line 76: Can you add comment? Why Flink 2.0 can support this?

paimon-flink/paimon-flink-1.20/src/main/java/org/apache/paimon/flink/source/operator/MonitorSource.java
yunfengzhou-hub: In future paimon-flink-common will rely on flink 2.0, which will not provide SourceFunction as public API. Where can we place the common MonitorSource in this case?

JingsongLi: I mean, you can extract a class implementation without `SourceFunction`, just for reusing codes.

paimon-flink/paimon-flink-1.20/src/main/java/org/apache/paimon/flink/source/operator/MonitorSource.java
yunfengzhou-hub: MonitorFunction have state, and the states of legacy SourceFunction and FLIP-27 Source are not compatible. In order to support existing paimon 0.9 + flink 1.20 users to update their Flink job to paimon 1.0 + flink 1.20 without state loss, this PR chose to keep using SourceFunction for flink 1.20 and earlier versions.

According to offline discussion, I'll update all MonitorFunction into MonitorSource, including for earlier flink versions, and tests against the new source in old flink versions.

paimon-flink/paimon-flink-1.20/src/main/java/org/apache/paimon/flink/source/operator/MonitorSource.java
yunfengzhou-hub: According to the discussion in the comment below, I'll update all legacy source function to FLIP-27 source class. So reusing code for source function is not applicable now.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/operator/MonitorSource.java
```diff
@@ -71,74 +71,88 @@
...
+
+            Long nextSnapshot = this.scan.checkpoint();
+            if (nextSnapshot != null) {
+                results.add(new SimpleSourceSplit(CHECKPOINT_STATE + nextSnapshot));
+                this.nextSnapshotPerCheckpoint.put(checkpointId, nextSnapshot);
+            }
+
+            this.nextSnapshotPerCheckpoint.forEach(
+                    (k, v) ->
+                            results.add(new SimpleSourceSplit(NEXT_SNAPSHOT_STATE + k + ":" + v)));
```
tsreaper, line 136: So you're storing state information in splits? Create a utility class for this, for example:

```java
stateBuilder.storeSingle(CHECKPOINT_STATE, nextSnapshot);
stateBuilder.storeList(NEXT_SNAPSHOT_STATE, this.nextSnapshotPerCheckpoint.stream().map((k, v) -> k + ":" + v).collect(Collectors.asList()));
return stateBuilder.build();
```

And extract state with

```java
stateBuilder.getSingle(CHECKPOINT_STATE);
stateBuilder.getList(NEXT_SNAPSHOT_STATE);
```

paimon-flink/paimon-flink-cdc/src/test/java/org/apache/paimon/flink/sink/cdc/TestCdcSource.java
```diff
@@ -0,0 +1,123 @@
...
+                    new SimpleSourceSplit(Integer.toString(events.size())));
+        }
+
+        @Override
+        public void addSplits(List<SimpleSourceSplit> list) {
+            int count =
+                    list.stream()
+                            .map(x -> Integer.parseInt(x.splitId()))
+                            .reduce(Integer::sum)
+                            .orElse(0);
```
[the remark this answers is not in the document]
yunfengzhou-hub: The implementation has used splitId to store the value in the split, which is actually not good. I'll fix this.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/operator/MonitorSource.java
```diff
@@ -0,0 +1,304 @@
...
+            }
+
+            if (isEmpty) {
+                Thread.sleep(monitorInterval);
+            }
+            return InputStatus.MORE_AVAILABLE;
+        }
+    }
+
+    private static class SplitState<T> {
```
tsreaper, line 206: Make it a public utility class. You also store state values into splits in tests, for example `FiniteTestSource`.

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/source/operator/TestingSourceOperator.java
```diff
@@ -0,0 +1,175 @@
...
+            boolean emitProgressiveWatermarks) {
+
+        this(
+                parameters,
+                reader,
+                watermarkStrategy,
+                timeService,
+                new TestingOperatorEventGateway(),
+                1,
+                5,
```
tsreaper, line 76: What are these magic numbers?

yunfengzhou-hub: `TestingSourceOperator` is introduced in reference to the Flink class with the same name. These numbers should be randomized values used to initialize MockRuntimeContext. I have added reference to that Flink class in the Javadoc.

tsreaper (approved):

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/BucketUnawareCompactSource.java
```diff
@@ -74,76 +72,66 @@ public BucketUnawareCompactSource(
...
-        if (ctx != null) {
-            synchronized (ctx.getCheckpointLock()) {
-                isRunning = false;
+        @Override
+        public InputStatus pollNext(ReaderOutput<UnawareAppendCompactionTask> readerOutput)
+                throws Exception {
+            long sleepTimeMillis = scanInterval - System.currentTimeMillis() + lastFetchTimeMillis;
+            if (sleepTimeMillis > 0) {
+                Thread.sleep(sleepTimeMillis);
+            }
```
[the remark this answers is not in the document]
skdfeitian: Will this modification cause the source to be busy max 100% in the flink UI?
 Please help me check this issue :  https://github.com/apache/paimon/issues/5463

PR #4617, opened 2024-12-02
[core] Make metadata.stats-dense-store default value is true

JingsongLi:
### Purpose

Without `metadata.stats-dense-store`, none `metadata.stats-mode` is meaningless.

So we should enable `metadata.stats-dense-store` by default, and it just affect none `metadata.stats-mode`.


yuzelin (approved):
+1

PR #4618, opened 2024-12-02
[spark] Support changing column types in array<struct> or map<?,struct>

tsreaper:
### Purpose

We have supported changing nested column types in array/map in Flink. This PR supports this feature in Spark.

### Tests

IT cases.

### API and Format

No format changes.

### Documentation

Document will be added later.


JingsongLi (approved):
+1

PR #4619, opened 2024-12-02
[docs] Add document for nested column evolution

tsreaper:
### Purpose

This PR adds document for nested column evolution in Flink and Spark.

### DIFF
--- a/docs/content/flink/sql-alter.md
+++ b/docs/content/flink/sql-alter.md
@@ -78,6 +78,10 @@ If you use object storage, such as S3 or OSS, please use this syntax carefully, 
 
 The following SQL adds two columns `c1` and `c2` to table `my_table`.
 
+{{< hint info >}}
+Flink SQL does not support adding nested columns. To add a column in a row type, see [Changing Column Type](#changing-column-type).
+{{< /hint >}}
+
 ```sql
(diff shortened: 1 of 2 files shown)


docs/content/flink/sql-alter.md
```diff
@@ -78,6 +78,10 @@ If you use object storage, such as S3 or OSS, please use this syntax carefully,
 
 The following SQL adds two columns `c1` and `c2` to table `my_table`.
 
+{{< hint info >}}
+Flink SQL does not support adding nested columns. To add a column in a row type, see [Changing Column Type](#changing-column-type).
```
JingsongLi: If you want to add a column in a row type, see....

docs/content/flink/sql-alter.md
```diff
@@ -99,6 +103,10 @@ otherwise this operation may fail, throws an exception like `The following colum
 ALTER TABLE my_table DROP (c1, c2);
 ```
 
+{{< hint info >}}
+Flink SQL does not support dropping nested columns. To drop a column in a row type, see [Changing Column Type](#changing-column-type).
```
JingsongLi: If you want to drop a column in a row type, see....

JingsongLi (approved):
+1

PR #4620, opened 2024-12-02
[iceberg] Use gzip by default to iceberg avro writer

JingsongLi:
### Purpose

Some Iceberg reader cannot support zstd for manifest files, for example DuckDB.

It is better to use gzip by default.


yuzelin (approved):
+1

PR #4621, opened 2024-12-02
[iceberg] Introduce metadata.iceberg.manifest-legacy-version

JingsongLi:
### Purpose

AWS Athena may use old manifest reader to read Iceberg manifest by names, we should let Paimon producing legacy Iceberg manifest list file, so we introduce: `'metadata.iceberg.manifest-legacy-version'`.

Depends on https://github.com/apache/paimon/pull/4620


yuzelin (approved):
+1

PR #4626, opened 2024-12-03
[core] Support nested types in Iceberg compatible metadata

tsreaper:
### Purpose

Currently, Iceberg compatible metadata has supported all primitive types in Iceberg. This PR adds support for nested types.

### Tests

Unit tests and IT cases.

### API and Format

No format changes.

### Documentation

Document is also added.


JingsongLi (approved):
+1

PR #4627, opened 2024-12-03
[doc] Add metastore.client.class option to hive catalog configuration document.

zhuangchong:
…

### Purpose

[doc] Add metastore.client.class option to hive catalog configuration document.


JingsongLi:
+1

PR #4629, opened 2024-12-04
[core] Drop stats for deleted data files to reduce memory

JingsongLi:
### Purpose

Statistic is only useful for ADD files.


leaves12138 (approved):
+1

JingsongLi:
+1

PR #4631, opened 2024-12-04
[core] Use min/max to fast skip the bsi index comparison

Tan-JiaLiang:
### Purpose

Firstly, When users turn off to collect the column stats, bsi compare would return a wrong result. Because the predicate can not out of the bsi min/max range in oneil compare.

Secondly, using min/max compare can fast return. e.g. in the GT case, if the predicate is less than the min value, we just need to return the existence bitmap.

At last, because of the v1.0 in paimon is not yet released, I didn't do the compatibility work about the bsi index.

### Tests

org.apache.paimon.utils.BitSliceIndexRoaringBitmapTest

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/utils/BitSliceIndexRoaringBitmap.java
+++ b/paimon-common/src/main/java/org/apache/paimon/utils/BitSliceIndexRoaringBitmap.java
@@ -18,11 +18,15 @@
 
 package org.apache.paimon.utils;
 
+import org.apache.paimon.annotation.VisibleForTesting;
+
 import java.io.DataInput;
 import java.io.DataOutput;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.Objects;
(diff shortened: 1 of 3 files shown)


paimon-common/src/main/java/org/apache/paimon/utils/BitSliceIndexRoaringBitmap.java
```diff
@@ -220,6 +309,7 @@ public boolean isNotEmpty() {
         public void serialize(DataOutput out) throws IOException {
             out.writeByte(VERSION_1);
             out.writeLong(min);
+            out.writeLong(max);
```
JingsongLi, line 312: Can you add detailed format to the page: https://paimon.apache.org/docs/master/concepts/spec/fileindex/

You can add an explanation for `bsi serialized`.

JingsongLi (approved):
+1 Thanks @Tan-JiaLiang

PR #4634, opened 2024-12-04
[tests] Fix @Timeout annotation not working in PrimaryKeyFileStoreTableITCase

tsreaper:
### Purpose

`@Timeout` annotation currently does not work in `PrimaryKeyFileStoreTableITCase`, because Flink's `collect` method will ignore `InterruptedException` triggered by junit timeout.

This PR creates another timeout thread to cancel the job, so that we can work-around this issue.


JingsongLi:
+1

PR #4638, opened 2024-12-04
[hive][spark] Support creating external table without schema when the table already exists

Zouxxyy:
### Purpose

- When the table already exists, we do not need to specify the schema when creating an external table (since it must be consistent with the old schema and can easily be written incorrectly).  
- TBLPROPERTIES can be a subset of the existing props, and for cross-engine compatibility, we have omitted the comparison of `owner`.

`CREATE TABLE t USING paimon LOCATION 'xxxx'`

extra updates:

- not save `provider` to props when create table with spark

Deprecated modification ~- Support add new props when creating external tables, for example, when a table written in Flink is used to create an external table in Spark, in this case, new props such as {provider: paimon} need be added~

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/schema/SchemaManager.java
+++ b/paimon-core/src/main/java/org/apache/paimon/schema/SchemaManager.java
@@ -208,24 +208,18 @@ public class SchemaManager implements Serializable {
         return createTable(schema, false);
     }
 
-    public TableSchema createTable(Schema schema, boolean ignoreIfExistsSame) throws Exception {
+    public TableSchema createTable(Schema schema, boolean externalTable) throws Exception {
         while (true) {
             Optional<TableSchema> latest = latest();
             if (latest.isPresent()) {
-                TableSchema oldSchema = latest.get();
-                boolean isSame =
-                        Objects.equals(oldSchema.fields(), schema.fields())
-                                && Objects.equals(oldSchema.partitionKeys(), schema.partitionKeys())
-                                && Objects.equals(oldSchema.primaryKeys(), schema.primaryKeys())
-                                && Objects.equals(oldSchema.options(), schema.options());
-                if (ignoreIfExistsSame && isSame) {
-                    return oldSchema;
+                TableSchema latestSchema = latest.get();
+                if (externalTable) {
+                    checkSchemaForExternalTable(latestSchema, schema);
+                    return latestSchema;
+                } else {
+                    throw new IllegalStateException(
+                            "Schema in filesystem exists, creation is not allowed.");
                 }
-
-                throw new IllegalStateException(
-                        "Schema in filesystem exists, please use updating,"
-                                + " latest schema is: "
-                                + oldSchema);
             }
 
             List<DataField> fields = schema.fields();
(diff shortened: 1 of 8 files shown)


paimon-core/src/main/java/org/apache/paimon/schema/SchemaManager.java
```diff
@@ -254,6 +246,41 @@ public TableSchema createTable(Schema schema, boolean ignoreIfExistsSame) throws
         }
     }
 
+    private TableSchema mergeSchemaForExternalTable(TableSchema oldSchema, Schema newSchema)
```
JingsongLi, line 249: Can you provide some behavior example from other systems? It looks a litter weird.

Zouxxyy: Spark supports automatic inference the schema of external table

```text
spark-sql (default)> create table t0 (id int) using parquet;
Time taken: 0.222 seconds

spark-sql (default)> insert into t0 values 1;
Time taken: 1.514 seconds

spark-sql (default)> create table t0_ext using parquet location '/user/hive/warehouse/t0';
Time taken: 0.231 seconds

spark-sql (default)> select * from t0_ext;
1
Time taken: 0.153 seconds, Fetched 1 row(s)

spark-sql (default)> show create table t0_ext;
CREATE TABLE default.t0_ext (
  id INT)
USING parquet
LOCATION '/hive/warehouse/t0'
```

JingsongLi: automatic inference is just inference, not to update schema of the storage.

How about iceberg and delta?

Zouxxyy: delta not accept new props, and its doc say:

```
When you run `CREATE TABLE` with a `LOCATION` that _already_ contains data stored using <Delta>, <Delta> does the following:

- If you specify _only the table name and location_, for example:


CREATE TABLE default.people1 USING DELTA LOCATION '/tmp/delta/people'`

  the table in the metastore automatically inherits the schema, partitioning, and table properties of the existing data. This functionality can be used to "import" data into the metastore.

- If you specify _any configuration_ (schema, partitioning, or table properties), <Delta> verifies that the specification exactly matches the configuration of the existing data.

  .. important::
    If the specified configuration does not _exactly_ match the configuration of the data, <Delta> throws an exception that describes the discrepancy.

```

here is the test result

```
create table delta2 using delta tblproperties ('k1' = 'v1') location '/user/hive/warehouse/delta1';
[DELTA_CREATE_TABLE_WITH_DIFFERENT_PROPERTY] The specified properties do not match the existing /user/hive/warehouse/delta1.

== Specified ==
k1=v1

== Existing ==
```

iceberg is weird, It has no inheritance at all and even create a new schema at the same path
```
create table ice1 (id int) using iceberg;
desc table extended ice1;
id                      int                                                                                        
# Detailed Table Information                                                
Name                    iceberg_catalog.test.ice1                           
Type                     MANAGED                                     
Location               /user/hive/warehouse/test.db/ice1                     
Provider                iceberg                                     
Owner                   root                                        
Table Properties        [current-snapshot-id=none,format=iceberg/parquet,format-version=2,write.parquet.compression-codec=zstd]


create table ice2 using iceberg tblproperties ('k1' = 'v1') location '/user/hive/warehouse/test.db/ice1';
desc table extended ice1;                                                                              
# Detailed Table Information                                                
Name                    iceberg_catalog.test.ice2                           
Type                    MANAGED                                     
Location              /user/hive/warehouse/test.db/ice1                     
Provider                iceberg                                     
Owner                   root                                        
Table Properties        [current-snapshot-id=none,format=iceberg/parquet,format-version=2,k1=v1,write.parquet.compression-codec=zstd]
```

So let's go back to paimon, If we have a table create by flink paimon, and want to create a new table location 'xxx' by spark paimon,
however spark paimon will add `{provider: paimon}` automaticly when creating table, which conflicts with the existing options in the table. In this case, should we discard this prop or add it (like PR) or throw exception(really don't want it).

JingsongLi: I love the behavior of delta lake, I think it is correct.

PR #4640, opened 2024-12-04
[core] Introduce manifest.delete-file-drop-stats

JingsongLi:
### Purpose

It is better to be compatible to old readers. So this PR introduces an option manifest.delete-file-drop-stats to disable this by default.


wwj6591812:
+1

PR #4641, opened 2024-12-04
[core] Make commit.max-retries behave as it means

Zouxxyy:
### Purpose

`commit.max-retries` represents retry times. If we configure it to 10, it means a maximum of 11 commits. If we configure it to 0, it means no retries are allowed.


wwj6591812:
+1

JingsongLi (approved):
+1

PR #4643, opened 2024-12-04
[flink] Optimizing parallelism for fixed bucekt and non-partitioned t…

LinMingQiang:
…able.

For non-partitioned table, the writerOperator's parallelism will be set to bucketNums if the bucketNums is less than inputOperator's parallelism.

Example :
Default parallelism : 4
Paimon tables bucket nums : 1
Result : There will be 3 tasks that are idle.

Before : 
image

After :

image

### Purpose

Linked issue: close #xxx


xuzifu666:
+1

JingsongLi (approved):
+1

PR #4651, opened 2024-12-06
[core] Trim key field in reading, map it to value field

leaves12138:
### Purpose

The key-value storage (in primary-key table) are redundant.

Read key-value from storage may appear to fetch more fields than we need.
For example, the format of a row is:   _KEY_a, _KEY_b, _FIELD_SEQUENCE, _ROW_KIND, a, b, c, d, e, f, g
The column "a" and "b" have double read. A optimization is:
<pre>
read before:    _KEY_a, _KEY_b, _FIELD_SEQUENCE, _ROW_KIND, a, b, c, d, e, f, g

read after:        a, b, _FIELD_SEQUENCE, _ROW_KIND, c, d, e, f, g
</pre>

While reading, we find the `a`, `b` from value and fill them to key.

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/table/SpecialFields.java
+++ b/paimon-common/src/main/java/org/apache/paimon/table/SpecialFields.java
@@ -72,6 +72,8 @@ public class SpecialFields {
 
     public static final String KEY_FIELD_PREFIX = "_KEY_";
     public static final int KEY_FIELD_ID_START = SYSTEM_FIELD_ID_START;
+    // reserve 1000 for other system fields
+    public static final int KEY_FIELD_ID_END = Integer.MAX_VALUE - 1_000;
 
     public static final DataField SEQUENCE_NUMBER =
             new DataField(Integer.MAX_VALUE - 1, "_SEQUENCE_NUMBER", DataTypes.BIGINT().notNull());
@@ -97,6 +99,10 @@ public class SpecialFields {
 
     public static boolean isSystemField(String field) {
         return field.startsWith(KEY_FIELD_PREFIX) || SYSTEM_FIELD_NAMES.contains(field);
+    }
+
+    public static boolean isKeyField(int id) {
+        return id >= KEY_FIELD_ID_START && id < KEY_FIELD_ID_END;
     }
 
     // ----------------------------------------------------------------------------------------
--- a/paimon-common/src/main/java/org/apache/paimon/table/SpecialFields.java
+++ b/paimon-common/src/main/java/org/apache/paimon/table/SpecialFields.java
@@ -72,8 +72,6 @@ public class SpecialFields {
 
     public static final String KEY_FIELD_PREFIX = "_KEY_";
     public static final int KEY_FIELD_ID_START = SYSTEM_FIELD_ID_START;
-    // reserve 1000 for other system fields
-    public static final int KEY_FIELD_ID_END = Integer.MAX_VALUE - 1_000;
 
     public static final DataField SEQUENCE_NUMBER =
             new DataField(Integer.MAX_VALUE - 1, "_SEQUENCE_NUMBER", DataTypes.BIGINT().notNull());
@@ -101,8 +99,8 @@ public class SpecialFields {
         return field.startsWith(KEY_FIELD_PREFIX) || SYSTEM_FIELD_NAMES.contains(field);
     }
 
-    public static boolean isKeyField(int id) {
-        return id >= KEY_FIELD_ID_START && id < KEY_FIELD_ID_END;
+    public static boolean isKeyField(String field) {
+        return field.startsWith(KEY_FIELD_PREFIX);
     }
 
     // ----------------------------------------------------------------------------------------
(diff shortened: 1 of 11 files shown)


Zouxxyy:
If these fields are not used when reading, can we store them when writing

leaves12138:
> If these fields are not used when reading, can we do not store them when writing

Correct, I am working on trimming them from writing.

wwj6591812:
+1

paimon-core/src/main/java/org/apache/paimon/utils/BulkFormatMapping.java
```diff
@@ -35,18 +36,23 @@
...
 import static org.apache.paimon.predicate.PredicateBuilder.excludePredicateWithFields;
+import static org.apache.paimon.table.SpecialFields.KEY_FIELD_ID_START;
 
 /** Class with index mapping and bulk format. */
 public class BulkFormatMapping {
 
     @Nullable private final int[] indexMapping;
     @Nullable private final CastFieldGetter[] castMapping;
     @Nullable private final Pair<int[], RowType> partitionPair;
+    @Nullable private final int[] trimmedKeyMapping;
```
tsreaper: Now that we have so many mappings, it would be better to add some comments to explain what each mapping means and what columns they store.

paimon-core/src/main/java/org/apache/paimon/utils/BulkFormatMapping.java
```diff
@@ -139,17 +171,51 @@ public BulkFormatMapping build(
...
+            Map<Integer, Integer> positionMap = new HashMap<>();
+
+            for (DataField field : fields) {
+                fieldMap.put(field.id(), field);
+            }
+
+            AtomicInteger index = new AtomicInteger();
+            for (int i = 0; i < fieldsWithoutPartition.size(); i++) {
+                DataField field = fieldsWithoutPartition.get(i);
+                boolean keyField = SpecialFields.isKeyField(field.id());
```
tsreaper: Why checking this with id? In [Paimon spec](https://paimon.apache.org/docs/master/concepts/spec/datafile/) we state that key column names must start with `_KEY_`. Check their name instead.

paimon-common/src/main/java/org/apache/paimon/table/SpecialFields.java
```diff
@@ -72,6 +72,8 @@ public class SpecialFields {
 
     public static final String KEY_FIELD_PREFIX = "_KEY_";
     public static final int KEY_FIELD_ID_START = SYSTEM_FIELD_ID_START;
+    // reserve 1000 for other system fields
+    public static final int KEY_FIELD_ID_END = Integer.MAX_VALUE - 1_000;
```
tsreaper: So there can be at most $10^9$ keys and only $10^3$ system fields? What if I need to add a new group of system fields?

paimon-core/src/main/java/org/apache/paimon/utils/BulkFormatMapping.java
```diff
@@ -139,17 +176,51 @@ public BulkFormatMapping build(
...
+                int id = keyField ? field.id() - KEY_FIELD_ID_START : field.id();
+                // field in data schema
+                DataField f = fieldMap.get(id);
+
+                if (f != null) {
+                    if (positionMap.containsKey(id)) {
+                        map[i] = positionMap.get(id);
+                    } else {
+                        trimmedFields.add(keyField ? f : field);
+                        map[i] = positionMap.computeIfAbsent(id, k -> index.getAndIncrement());
```
tsreaper, line 211:
```suggestion
                        trimmedFields.add(f);
                        map[i] = positionMap.put(id, index.getAndIncrement());
```

paimon-core/src/test/java/org/apache/paimon/utils/BulkFormatMappingTest.java
```diff
@@ -0,0 +1,95 @@
...
+import org.junit.jupiter.api.Test;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/** Test for {@link BulkFormatMapping.BulkFormatMappingBuilder}. */
+public class BulkFormatMappingTest {
+
+    @Test
+    public void tesTtrimKeyFields() {
```
tsreaper, line 36:
```suggestion
    public void testTrimKeyFields() {
```

paimon-core/src/main/java/org/apache/paimon/io/KeyValueFileReaderFactory.java
```diff
@@ -144,7 +144,8 @@ private FileRecordReader<KeyValue> createRecordReader(
                                         fileIO, filePath, fileSize, orcPoolSize),
                         bulkFormatMapping.getIndexMapping(),
                         bulkFormatMapping.getCastMapping(),
-                        PartitionUtils.create(bulkFormatMapping.getPartitionPair(), partition));
+                        PartitionUtils.create(bulkFormatMapping.getPartitionPair(), partition),
+                        bulkFormatMapping.getTrimmedKeyMapping());
```
JingsongLi, line 148: Can we just merge this trimmedKeyMapping into indexMapping?

leaves12138: Sure, we can

paimon-core/src/main/java/org/apache/paimon/utils/BulkFormatMapping.java
leaves12138: This cause bug

tsreaper: Please add some comments about why we can't directly use `f`.

tsreaper (approved):

PR #4652, opened 2024-12-06
[core] Expire partiitons add default delete num

askwang:
### Purpose

When expiring too many partitions using a dedicated expire_partitions job, it will have an impact on manifest merge and flink task stability. So adding a conf to control the expired number of partitions at a time.


JingsongLi (approved):
+1

PR #4654, opened 2024-12-06
[test][spark] Add insert with column list test case

Zouxxyy:
### Purpose

Add insert with column list test case


paimon-spark/paimon-spark-ut/src/test/scala/org/apache/paimon/spark/sql/InsertOverwriteTableTestBase.scala
```diff
@@ -508,4 +508,56 @@ abstract class InsertOverwriteTableTestBase extends PaimonSparkTestBase {
         ) :: Nil
     )
   }
+
+  test("Paimon Insert: insert with column list") {
+    sql("CREATE TABLE T (name String, student_id INT) PARTITIONED BY (address STRING)")
+
+    // insert with a column list
+    sql("INSERT INTO T (name, student_id, address) VALUES ('a', '1', 'Hangzhou')")
```
JingsongLi, line 516: It is better to add documentation too.

JingsongLi:
+1

PR #4658, opened 2024-12-07
[common] Fix FLOAT precision checking. (#4656)

### LINKED ISSUE #4656: [Bug] Precision Loss in `castFromStringInternal` for Valid Float Values (e.g., -0.0009163)
### Search before asking

### Paimon version

paimon-flink-1.18-1.0-SNAPSHOT.jar

### Compute Engine

flink

### Minimal reproduce step

create table in mysql 
``` sql
CREATE TABLE `example_table` (
  `id` bigint NOT NULL,
  `value` float DEFAULT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci
```
and insert data

``` sql
INSERT INTO example_table (id, value) VALUES (1, -0.0009163);
```

``` java
Float f = Float.valueOf(s);
if (f.toString().length() != s.length()) {
} else {
    return f;
}```
Precision checking should not strictly rely on comparing the lengths of the values. As a result, values such as `-0.0009163` and similar ones, which can be safely represented in MySQL FLOAT, are mistakenly rejected.


### Anything else?

_No response_

### PR DESCRIPTION
**Pull request by hunshenshi**
### Purpose

Linked issue: close #4656

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/utils/TypeUtils.java
+++ b/paimon-common/src/main/java/org/apache/paimon/utils/TypeUtils.java
@@ -144,11 +144,13 @@ public class TypeUtils {
                 } else {
                     // Compatible canal-cdc
                     Float f = Float.valueOf(s);
-                    if (f.toString().length() != s.length()) {
-                        throw new NumberFormatException(
-                                s + " cannot be cast to float due to precision loss");
-                    } else {
+                    // Validate precision by comparing the rounded value
+                    if ((double) f == d || Math.abs(f - d) < Math.abs(d * 1e-6)) {
                         return f;
+                    } else {
+                        // Significant precision loss detected
+                        throw new NumberFormatException(
+                                s + " cannot be cast to FLOAT due to precision loss");
                     }
                 }
             case DOUBLE:

### CODE REVIEW

File: paimon-common/src/main/java/org/apache/paimon/utils/TypeUtils.java [hunk 1]
```diff
@@ -144,11 +144,13 @@ public static Object castFromStringInternal(String s, DataType type, boolean isC
                 } else {
                     // Compatible canal-cdc
                     Float f = Float.valueOf(s);
-                    if (f.toString().length() != s.length()) {
-                        throw new NumberFormatException(
-                                s + " cannot be cast to float due to precision loss");
-                    } else {
+                    // Validate precision by comparing the rounded value
```
wwj6591812, line 147: float f = Float.parseFloat(s);

paimon-common/src/main/java/org/apache/paimon/utils/TypeUtils.java
```diff
@@ -144,11 +144,13 @@ public static Object castFromStringInternal(String s, DataType type, boolean isC
...
+                    if ((double) f == d || Math.abs(f - d) < Math.abs(d * 1e-6)) {
```
wwj6591812, line 148: why use "d * 1e-6"?
what about Double#compare?

paimon-common/src/main/java/org/apache/paimon/utils/TypeUtils.java
wwj6591812, line 147: float f = Float.parseFloat(s);

paimon-common/src/main/java/org/apache/paimon/utils/TypeUtils.java
wwj6591812, line 148: why use "d * 1e-6"?
what about Double#compare?

paimon-common/src/main/java/org/apache/paimon/utils/TypeUtils.java
```diff
@@ -144,11 +144,13 @@ public static Object castFromStringInternal(String s, DataType type, boolean isC
...
                         return f;
+                    } else {
+                        // Significant precision loss detected
+                        throw new NumberFormatException(
+                                s + " cannot be cast to FLOAT due to precision loss");
```
wwj6591812, line 153: The exception message need improve. It needs to cover the comparative failure situation.

paimon-common/src/main/java/org/apache/paimon/utils/TypeUtils.java
hunshenshi: 1. The value 1e-6 (which is 0.000001) is typically a reasonable threshold for floating-point comparisons, it's a balance between precision and practical usability.
2. This approach helps avoid rejecting values that are within the acceptable precision for FLOAT types in databases like MySQL, while still accounting for the limitations of floating-point precision.

PR #4662, opened 2024-12-09
[spark] Paimon parser only resolve own supported procedures

askwang:
### Purpose

Spark allows users to configure multiple extensions and each extension is allowed to inject its own SQL parser, conflicts maybe occur if multiple extensions support the`call` command. We know that the `SYSTEM_DATABASE_NAME` of  all paimon procedures are `sys`, so only parsing supported procedures in paimon.

Also, support ``call catalog.`sys`.`rollback` procedure`` with backticks.

like iceberg https://github.com/apache/iceberg/pull/11480


JingsongLi (approved):
+1

PR #4663, opened 2024-12-09
[core] Optimize drop partitions to avoid stack overflow

JingsongLi:
### Purpose

If we expire too many partitions, there may be a stack overflow exception in `FileStoreCommitImpl.dropPartitions`. This is because there are so many partitions in the construction for predicate.

We can optimize it by using `PartitionPredicate.fromMultiple`, if all inputs are full partition fields.


wwj6591812:
+1

tsreaper (approved):

PR #4669, opened 2024-12-09
[hive] Make HiveMetastoreClient.addPartition thread safe

JingsongLi:
### Purpose

We can catch `AlreadyExistsException` to support automatic partition addition. Otherwise, adding partitions for multiple jobs together may cause exceptions.


wwj6591812:
+1

PR #4672, opened 2024-12-10
[core] Introduce data-file.path-directory

JingsongLi:
### Purpose

Specify the path directory of data files.

Duckdb Iceberg connector may rely on files placed in the `root/data` directory, while Paimon is usually placed directly in the `root` directory, so you can configure this parameter for the table to achieve compatibility: `'data-file.path-directory' = 'data'`.


Zouxxyy (approved):
+1

PR #4673, opened 2024-12-10
[hotfix] Fix flaky test of orc tests in ArrowBatchConverterTest

yuzelin:
### Purpose

The orc char reader will trim the string which make the tests assumption wrong.


JingsongLi:
+1

PR #4680, opened 2024-12-10
[Test] Add retun for failed tests

leaves12138:
### Purpose

E2E Tests always get to TimeoutException. We add retry for those failed tests.


JingsongLi:
We should not rerun tests.

PR #4681, opened 2024-12-10
[core] Clean up invalid branch cache and not cache system table in caching catalog

Zouxxyy:
### Purpose

- Clean up invalid branch cache when table or brach id dropped
- Caching catalog not cache system table

Other:

Remove "spark.sql.catalog.paimon.cache-enabled" -> "false" in `PaimonSparkTestBase`. Note: there are still some cases that cannot enable it because it is incompatible with `write.merge-schema` currently.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/catalog/CachingCatalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/catalog/CachingCatalog.java
@@ -201,6 +201,9 @@ public class CachingCatalog extends DelegateCatalog {
             throws TableNotExistException {
         super.dropTable(identifier, ignoreIfNotExists);
         invalidateTable(identifier);
+        if (identifier.isMainTable()) {
+            invalidateAttachedTables(identifier);
+        }
     }
 
     @Override
@@ -228,7 +231,13 @@ public class CachingCatalog extends DelegateCatalog {
         // For system table, do not cache it directly. Instead, cache the origin table and then wrap
         // it to generate the system table.
         if (identifier.isSystemTable()) {
-            Table originTable = getTable(identifier.toOriginTable());
+            Identifier originIdentifier =
+                    new Identifier(
+                            identifier.getDatabaseName(),
+                            identifier.getTableName(),
+                            identifier.getBranchName(),
+                            null);
+            Table originTable = getTable(originIdentifier);
             table =
                     SystemTableLoader.load(
                             checkNotNull(identifier.getSystemTableName()),
@@ -298,7 +307,7 @@ public class CachingCatalog extends DelegateCatalog {
         public void onRemoval(Identifier identifier, Table table, @NonNull RemovalCause cause) {
             LOG.debug("Evicted {} from the table cache ({})", identifier, cause);
             if (RemovalCause.EXPIRED.equals(cause)) {
-                tryInvalidateAttachedTables(identifier);
+                // ignore now
             }
         }
     }
@@ -306,19 +315,16 @@ public class CachingCatalog extends DelegateCatalog {
     @Override
     public void invalidateTable(Identifier identifier) {
         tableCache.invalidate(identifier);
-        tryInvalidateAttachedTables(identifier);
         if (partitionCache != null) {
             partitionCache.invalidate(identifier);
         }
     }
 
-    private void tryInvalidateAttachedTables(Identifier identifier) {
-        if (identifier.isMainTable()) {
-            // invalidate cached branches
-            for (@NonNull Identifier i : tableCache.asMap().keySet()) {
-                if (identifier.getTableName().equals(i.getTableName())) {
-                    tableCache.invalidate(i);
-                }
+    /** invalidate attached tables, such as cached branches. */
+    private void invalidateAttachedTables(Identifier identifier) {
+        for (@NonNull Identifier i : tableCache.asMap().keySet()) {
+            if (identifier.getTableName().equals(i.getTableName())) {
+                tableCache.invalidate(i);
             }
         }
     }
--- a/paimon-core/src/main/java/org/apache/paimon/catalog/CachingCatalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/catalog/CachingCatalog.java
@@ -228,12 +228,7 @@ public class CachingCatalog extends DelegateCatalog {
         // For system table, do not cache it directly. Instead, cache the origin table and then wrap
         // it to generate the system table.
         if (identifier.isSystemTable()) {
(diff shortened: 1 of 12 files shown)


paimon-core/src/main/java/org/apache/paimon/catalog/CachingCatalog.java
```diff
@@ -227,27 +224,13 @@ public Table getTable(Identifier identifier) throws TableNotExistException {
...
             }
-            table =
-                    SystemTableLoader.load(
-                            Preconditions.checkNotNull(identifier.getSystemTableName()),
-                            (FileStoreTable) originTable);
-            if (table == null) {
-                throw new TableNotExistException(identifier);
-            }
-            putTableCache(identifier, table);
-            return table;
```
JingsongLi, line 250: why removing this?

paimon-common/src/main/java/org/apache/paimon/utils/StringUtils.java
```diff
@@ -546,6 +546,10 @@ public static String caseSensitiveConversion(String str, boolean allowUpperCase)
         return allowUpperCase ? str : str.toLowerCase();
     }
 
+    public static String safeToLowerCase(String str) {
```
JingsongLi, line 549: create a method:
```
public static boolean equalsIgnoreCase(String str1, String str2) {
    return str1 != null ? str1.equalsIgnoreCase(str2) : str2 == null;
}
```

paimon-core/src/main/java/org/apache/paimon/catalog/CachingCatalog.java
```diff
@@ -309,32 +292,35 @@ private class TableInvalidatingRemovalListener implements RemovalListener<Identi
...
-        if (!isSpecifiedSystemTable(identifier)) {
-            tableCache.invalidateAll(allSystemTables(identifier));
-        }
-    }
-
-    private static Iterable<Identifier> allSystemTables(Identifier ident) {
-        List<Identifier> tables = new ArrayList<>();
-        for (String type : SYSTEM_TABLES) {
-            tables.add(Identifier.fromString(ident.getFullName() + SYSTEM_TABLE_SPLITTER + type));
+    private void tryInvalidateAttachedTables(Identifier identifier) {
```
JingsongLi, line 309: I'm wondering if the system table doesn't need to be stored in the cache anymore. Just wrap it when retrieving it

paimon-core/src/main/java/org/apache/paimon/catalog/CachingCatalog.java
Zouxxyy: the line 253~254 do the same thing, and this can cover by testInvalidateSysTablesIfBaseTableIsDropped

paimon-core/src/main/java/org/apache/paimon/catalog/CachingCatalog.java
Zouxxyy: I'm ok to cache the system table or not, because identity is our cache key, and the system table is a kind of identity.

But I'm just wondering why do we need to cache the origin table before caching the system table (maybe this step can be skipped)?

paimon-core/src/main/java/org/apache/paimon/catalog/CachingCatalog.java
Zouxxyy: Oh, I see, here we want to reuse the table to wrap the system table

paimon-core/src/main/java/org/apache/paimon/catalog/CachingCatalog.java
Zouxxyy: I checked the previous logic again. Now we reuse the table (in cache) and then wrap it to the system table. In this case, I think we don't need to cache the system table.

Another solution is to only cache the system table (not warp from cache table). 

Which one do you think is more suitable?

JingsongLi: I think we can just cache the original table.

wwj6591812:
+1

paimon-core/src/main/java/org/apache/paimon/catalog/CachingCatalog.java
```diff
@@ -309,32 +298,29 @@ private class TableInvalidatingRemovalListener implements RemovalListener<Identi
...
+        if (identifier.isMainTable()) {
```
JingsongLi, line 316: We don't need to care about branch tables.

paimon-core/src/main/java/org/apache/paimon/catalog/CachingCatalog.java
```diff
@@ -227,26 +225,17 @@ public Table getTable(Identifier identifier) throws TableNotExistException {
...
-                            null);
-            Table originTable = tableCache.getIfPresent(originIdentifier);
-            if (originTable == null) {
-                originTable = wrapped.getTable(originIdentifier);
-                putTableCache(originIdentifier, originTable);
-            }
+        // For system table, do not cache it directly. Instead, cache the origin table and then wrap
+        // it to generate the system table.
+        if (identifier.isSystemTable()) {
+            Table originTable = getTable(identifier.toOriginTable());
```
JingsongLi, line 231: Just keep:
```
Identifier originIdentifier =
                    new Identifier(
                            identifier.getDatabaseName(),
                            identifier.getTableName(),
                            identifier.getBranchName(),
                            null);
```

It is hard to understand `toOriginTable`, with branch, or not with branch?

paimon-core/src/main/java/org/apache/paimon/catalog/CachingCatalog.java
Zouxxyy: Why, when the table is deleted, the cached branch table should also be cleaned up.

JingsongLi: Maybe we can just do this in `invalidateTable`?

Zouxxyy: I see, we can move this logic to drop table

JingsongLi: Alter table with branch is also supported, maybe still in `invalidateTable`?

JingsongLi: But not in `TableInvalidatingRemovalListener`.

JingsongLi (approved):
+1

PR #4682, opened 2024-12-10
[hotfix] CloneAction throw more clear exception when no table in source catalog

yuzelin:
### Purpose

env.fromCollection will throw `Collection must not be null` which is confusing.


JingsongLi (approved):
+1

Issue #4530, opened 2024-11-14
[Feature] Using bitmap index to accelerate the query


Tan-JiaLiang:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Motivation

Currently we have introduced bitmap indexes. There are some optimisations that we can do when the user queries only against the bitmap indexed columns.

Suppose we have a table `usershop_behavior` with a bitmap index on the `gender` column and a bsi index on the `gmv` column.
```sql
CREATE TABLE usershop_behavior (
    uid BIGINT
    gender STRING,
    gmv BIGINT
) WITH (
    'file-index.bitmap.columns' = 'gender',
    'file-index.bsi.columns' = 'gmv'
);
```

The bitmap index and bsi index can be used not only for filtering, but also for some simple aggregation like:
```sql
SELECT 
    gender,
    COUNT(*) AS total,
    SUM(gmv) AS total_gmv,
    AVG(gmv) AS avg_gmv
FROM usershop_behavior
GROUP BY gender

SELECT 
    gender,
    COUNT(*) AS total,
    SUM(gmv) AS total_gmv,
    AVG(gmv) AS avg_gmv
FROM usershop_behavior
WHERE gender='M'
GROUP BY gender
```

BSI index can be useful in topk scenarios
```sql
SELECT *
FROM usershop_behavior
ORDER BY gmv DESC
LIMIT 10;

SELECT *
FROM usershop_behavior
WHERE gender='M'
ORDER BY gmv DESC
LIMIT 10;
```

### Solution

Apache Flink and Apache Spark are already provides some interfaces. e.g.

Apache Flink:
`org.apache.flink.table.connector.source.abilities.SupportsAggregatePushDown`

Apache Spark:
`org.apache.spark.sql.connector.read.SupportsPushDownTopN`
`org.apache.spark.sql.connector.read.SupportsPushDownAggregates`

When queries hit the bitmap index rules, we can rewrite TableScan to BitmapIndexScan.

### Anything else?

Currently our index is designed to be used only for Data Skipping and it is not as reliable as filtering using partitioned keys. (We can't tell the flink&spark engine that filtering with indexes is reliable.)

This is because creating the index is split into several steps.
1. stop the ingesting task
2. using alter table to add index options
3. call rewrite index procedure
4. restart the ingesting task

We need to find a way to make indexes reliable, like partition keys. (e.g. throw exception when index is empty?)
[...]

PR #4684, opened 2024-12-11
[core] Support file index filter push down

Tan-JiaLiang:
### Purpose

Linked issue: #4530

The filter will push down if the index matches the predicate, if the index does not exist at runtime, it will fallback to using the paimon predicate filter.

For now, we support bloom-filter/bitmap/bsi index, and the bitmap/bsi index supports filter push down and they will use `ApplyBitmapIndexRecordReader` to filter by position.

### Tests

org.apache.paimon.table.AppendOnlyFileStoreTableTest#testFileIndexFilterPushDownFallback
org.apache.paimon.table.PrimaryKeyFileStoreTableTest#testDeletionVectorsWithFileIndexFilterPushDownFallback
org.apache.paimon.flink.source.FilterPushDownITCase#testPartitionAndIndexConditionConsuming

### API and Format

No

### Documentation

No


Tan-JiaLiang:
I've found that the performance of the file index filter is not as good as I expected, so I'm marking it as a draft for now.

PR #4685, opened 2024-12-11
[core] Check file size after write bundle of records

leaves12138:
### Purpose

If we write bundle, we can't precisely reach the `CHECK_ROLLING_RECORD_CNT`. So we need to force check target file size.


paimon-core/src/main/java/org/apache/paimon/io/RollingFileWriter.java
```diff
@@ -65,9 +65,9 @@ public long targetFileSize() {
     }
 
     @VisibleForTesting
```
JingsongLi, line 67: remove this VisibleForTesting?

JingsongLi (approved):
+1

PR #250, opened 2022-07-28
[FLINK-28715] Throw better exception when file not found in reading

JingsongLi:
When reading a file, if it is found that the file does not exist, it directly throws a file not found exception, which is often difficult for users to understand.
We can make it more clear in the exception message, e.g.
The file cannot be found, this may be because the read is too slow and the previous snapshot expired, you can configure a larger snapshot.time-retained or speed up your read.

```
Caused by: java.io.FileNotFoundException: File does not exist:
at org.apache.flink.table.store.file.utils.FileUtils.getFileSize(FileUtils.java:94) ~[flink-table-store-dist-0.2.jar:0.2-SNAPSHOT]
at org.apache.flink.table.store.file.data.DataFileReader$DataFileRecordReader.<init>(DataFileReader.java:86) ~[flink-table-store-dist-0.2.jar:0.2-SNAPSHOT]
```

PR #4693, opened 2024-12-12
[core] improve error message when file not found in reading

izhangzhihao:
### Purpose

improve error message when file not found in reading
Linked PR: #250


izhangzhihao:
@JingsongLi pls have a review

paimon-core/src/main/java/org/apache/paimon/utils/FileUtils.java
```diff
@@ -114,8 +114,8 @@ public static void checkExists(FileIO fileIO, Path file) throws IOException {
                             "File '%s' not found, Possible causes: "
                                     + "1.snapshot expires too fast, you can configure 'snapshot.time-retained'"
                                     + " option with a larger value. "
-                                    + "2.consumption is too slow, you can improve the performance of consumption"
-                                    + " (For example, increasing parallelism).",
+                                    + "2.the file was compacted before it was consumed, you can improve the performance of consumption"
```
JingsongLi, line 117: No, not be compacted, it is because of snapshot expired

PR #4699, opened 2024-12-13
[core] Store `cardinality` in the deletion vector meta

Zouxxyy:
### Purpose

Store `cardinality` in the deletion vector meta, new meta like this:

```
 1. f0: the data file name corresponding to this deletion vector.
 2. f1: the starting offset of this deletion vector in the index file.
 3. f2: the length of this deletion vector in the index file.
 4. cardinality: the number of deleted rows.
 ```
new:

`cardinality`: It can be used for optimizing count(*) on the dv table.

compatibility:

upgrate `CommitMessageSerializer` to version 5
upgrate `DataSplit` to version 4


JingsongLi:
Can you explain the usage for them separately?

JingsongLi (approved):
+1

PR #4701, opened 2024-12-13
[core] Retry if snapshot commit hint failed.

leaves12138:
### Purpose

Maybe over-writer file will fail, so it's better to add some retries.

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/fs/FileIOUtils.java
+++ b/paimon-common/src/main/java/org/apache/paimon/fs/FileIOUtils.java
@@ -44,19 +44,4 @@ public class FileIOUtils {
         io.exists(path);
         return fileIO;
     }
-
-    public static void overwriteFileUtf8WithRetry(
-            FileIO fileIO, Path path, String content, int times) throws IOException {
-        int loopTime = times;
-        while (loopTime-- > 0) {
-            try {
-                fileIO.overwriteFileUtf8(path, content);
-                return;
-            } catch (IOException e) {
-                if (loopTime == 0) {
-                    throw e;
-                }
-            }
-        }
-    }
 }
--- a/paimon-common/src/main/java/org/apache/paimon/fs/FileIOUtils.java
+++ b/paimon-common/src/main/java/org/apache/paimon/fs/FileIOUtils.java
@@ -44,4 +44,19 @@ public class FileIOUtils {
         io.exists(path);
         return fileIO;
     }
+
+    public static void overwriteFileUtf8WithRetry(
+            FileIO fileIO, Path path, String content, int times) throws IOException {
+        int loopTime = times;
+        while (loopTime-- > 0) {
+            try {
+                fileIO.overwriteFileUtf8(path, content);
+                return;
+            } catch (IOException e) {
+                if (loopTime == 0) {
+                    throw e;
+                }
+            }
+        }
+    }
 }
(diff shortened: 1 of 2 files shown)


paimon-common/src/main/java/org/apache/paimon/fs/FileIOUtils.java
```diff
@@ -44,4 +44,19 @@ public static FileIOLoader checkAccess(FileIOLoader fileIO, Path path, CatalogCo
         io.exists(path);
         return fileIO;
     }
+
+    public static void overwriteFileUtf8WithRetry(
```
JingsongLi, line 48: inline this in commitHint

paimon-common/src/main/java/org/apache/paimon/fs/FileIOUtils.java
```diff
@@ -44,4 +44,19 @@ public static FileIOLoader checkAccess(FileIOLoader fileIO, Path path, CatalogCo
...
+            FileIO fileIO, Path path, String content, int times) throws IOException {
+        int loopTime = times;
+        while (loopTime-- > 0) {
+            try {
+                fileIO.overwriteFileUtf8(path, content);
+                return;
+            } catch (IOException e) {
+                if (loopTime == 0) {
```
JingsongLi, line 56: Sleep a random value here?

paimon-core/src/main/java/org/apache/paimon/utils/SnapshotManager.java
```diff
@@ -883,6 +884,6 @@ public void commitEarliestHint(long snapshotId) throws IOException {
 
     private void commitHint(long snapshotId, String fileName, Path dir) throws IOException {
         Path hintFile = new Path(dir, fileName);
-        fileIO.overwriteFileUtf8(hintFile, String.valueOf(snapshotId));
+        FileIOUtils.overwriteFileUtf8WithRetry(fileIO, hintFile, String.valueOf(snapshotId), 3);
```
xuzifu666, line 887: Hard code here? Or make it configurable better？

leaves12138: ok

paimon-core/src/main/java/org/apache/paimon/utils/SnapshotManager.java
```diff
@@ -883,6 +885,22 @@ public void commitEarliestHint(long snapshotId) throws IOException {
...
         Path hintFile = new Path(dir, fileName);
-        fileIO.overwriteFileUtf8(hintFile, String.valueOf(snapshotId));
+        int loopTime = 3;
+        while (loopTime-- > 0) {
+            try {
+                fileIO.overwriteFileUtf8(hintFile, String.valueOf(snapshotId));
+                return;
+            } catch (IOException e) {
+                try {
+                    Thread.sleep(RANDOM.nextInt(1000) + 500);
```
JingsongLi, line 895: `ThreadLocalRandom.current()`

paimon-core/src/main/java/org/apache/paimon/utils/SnapshotManager.java
```diff
@@ -883,6 +885,22 @@ public void commitEarliestHint(long snapshotId) throws IOException {
...
+                } catch (InterruptedException ex) {
+                    // throw root cause
+                    throw new RuntimeException(e);
```
JingsongLi, line 898: Thread.currentThread().interrupt();

JingsongLi (approved):
+1

PR #4705, opened 2024-12-13
[core] Fix predicate literals cast in filter pushdown after schema evolution

yuzelin:
### Purpose

Reproduce case:
```
CREATE TABLE T (
  id INT,
  f DECIMAL(5,2),
);
INSERT INTO T VALUES (1, 111.32);
ALTER TABLE T MODIFY (f DECIMAL(6, 3));
SELECT * FROM T WHERE f < 111.321;
```
The result is empty but expected 111.320

There are also same problem for other field cast Scenario.

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/casting/CastExecutors.java
+++ b/paimon-common/src/main/java/org/apache/paimon/casting/CastExecutors.java
@@ -18,17 +18,32 @@
 
 package org.apache.paimon.casting;
 
+import org.apache.paimon.data.Decimal;
+import org.apache.paimon.predicate.Equal;
+import org.apache.paimon.predicate.GreaterOrEqual;
+import org.apache.paimon.predicate.GreaterThan;
+import org.apache.paimon.predicate.In;
+import org.apache.paimon.predicate.LeafFunction;
+import org.apache.paimon.predicate.LeafPredicate;
+import org.apache.paimon.predicate.LessOrEqual;
+import org.apache.paimon.predicate.LessThan;
+import org.apache.paimon.predicate.NotEqual;
+import org.apache.paimon.predicate.NotIn;
 import org.apache.paimon.types.DataType;
 import org.apache.paimon.types.DataTypeFamily;
 import org.apache.paimon.types.DataTypeRoot;
+import org.apache.paimon.types.DecimalType;
 
 import javax.annotation.Nullable;
 
+import java.math.BigDecimal;
 import java.util.HashMap;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.stream.Collectors;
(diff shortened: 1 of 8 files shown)


paimon-common/src/main/java/org/apache/paimon/casting/CastExecutors.java
```diff
@@ -100,6 +115,147 @@ public class CastExecutors {
         return IDENTITY_CAST_EXECUTOR;
     }
 
+    /**
+     * When filter a field witch was evolved from/to a numeric type, we should carefully handle the
+     * precision match and overflow problem.
+     */
+    @Nullable
+    public static List<Object> safelyCastLiteralsWithNumericEvolution(
```
JingsongLi, line 123: Optional<List<Object>>

paimon-common/src/main/java/org/apache/paimon/casting/CastExecutors.java
```diff
@@ -100,6 +115,147 @@ public class CastExecutors {
...
+            LeafPredicate predicate, DataType outputType) {
```
JingsongLi, line 124: List<Object> literals, DataType inputType, DataType outputType

paimon-common/src/main/java/org/apache/paimon/casting/CastExecutors.java
JingsongLi, line 123: castLiteralsWithNumericEvolution

wwj6591812:
+1

BitFunnelComp:
The predicate conversion of decimal/floating type seems to be too complicated and there may be bad cases. Is it possible to set up a whitelist to only allow integer conversion?

yuzelin:
> The predicate conversion of decimal/floating type seems to be too complicated and there may be bad cases. Is it possible to set up a whitelist to only allow integer conversion?

I think this is a good idea. Actually, decimal type and related schema evolution is rarely used in industry.

JingsongLi (commented):
Add a large BIGINT (Bigger than Int.Max) testing.

paimon-common/src/main/java/org/apache/paimon/casting/CastExecutors.java
```diff
@@ -100,6 +102,63 @@ public class CastExecutors {
...
+                        return Optional.of(literals);
+                    }
+
+                    // Parquet filter Int comparator cannot handle long value.
+                    // See org.apache.parquet.schema.PrimitiveType.
+                    // So ignore filter if long literal is out of int scale.
+                    List<Object> newLiterals = new ArrayList<>(literals.size());
+                    for (Object literal : literals) {
+                        long originalValue = (long) literal;
+                        int newValue = (int) originalValue;
```
JingsongLi, line 142: Create a CastExecutor to do this casting?

JingsongLi:
+1

PR #4706, opened 2024-12-14
[spark] Add purge table command to clear data

xuzifu666:
### Purpose

Linked issue: close #xxx
Currently Paimon cannot clear data from table rightly, truncate would do a overwrite snapshot and clear data\manifest... after snapshot expire, but user sometime need clear data quickly to reduce storage pressure, so add a purge command to clear data quickly. 

### Tests
PaimonPurgeTableTestBase

### DIFF
--- a/docs/content/spark/sql-ddl.md
+++ b/docs/content/spark/sql-ddl.md
@@ -258,6 +258,15 @@ CREATE VIEW v1 AS SELECT * FROM t1;
 CREATE OR REPLACE VIEW v1 AS SELECT * FROM t1;
 ```
 
+### Purge Table
+
+Purge table with delete data and metadata directly.
+
+```sql
+-- purge a table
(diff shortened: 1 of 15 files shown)


JingsongLi:
Can you find some similar examples? For example, from other lake formats

xuzifu666:
> Can you find some similar examples? For example, from other lake formats



> Can you find some similar examples? For example, from other lake formats

Hudi truncate table would clear metadata and data directly； iceberg would do like paimon. In some env user want to delete all data quickly only need keep schema info, so add the command. @JingsongLi

JingsongLi:
After offline discussing with @xuzifu666 , we can provide a procedure: purge_files.

xuzifu666:
link this pr to: https://github.com/apache/paimon/pull/4717  and close it first.

PR #4708, opened 2024-12-15
[core] Tolerate the NoSuchObjectException when report the partition s…

Aitozi:
…tatistic

### Purpose

We may report to the hms when the partition has been expired, we should handle this case.

Linked issue: close #xxx


JingsongLi (approved):
+1

PR #4709, opened 2024-12-15
[core][spark] Enable limit pushdown and count optimization for dv table

Zouxxyy:
### Purpose

Since we introduce dv cardinality in #4699, we can enable limit pushdown and count optimization for dv table now


JingsongLi (approved):
+1

PR #4713, opened 2024-12-16
[flink] avoid using 'newHashMapWithExpectedSize' which is internal in flink

LsomeYeah:
### Purpose

Linked issue: close #xxx

`newHashMapWithExpectedSize` is internally used in flink, we should avoid using it.


JingsongLi:
+1

PR #4714, opened 2024-12-16
[spark] Fix relativeBucketPath with data-file.path-directory

Zouxxyy:
### Purpose

Fix relativeBucketPath with data-file.path-directory


JingsongLi:
+1

PR #4497, opened 2024-11-11
[core][format] Optimize manifest reading performance，add pushdown for manifest and orc.

ranxianglei:
### Purpose

English：Optimize the manifest reading performance, optimize the format object creation performance, and reduce the total time spent on the actual test manifest to less than 3ms (of course there is room for optimization to reduce it to less than 1ms). With the orc push-down function turned on, the metadata format is changed to orc, which can handle high-concurrency (qps greater than 10,000) and low-latency (overall rt less than 50ms) scenarios.

Chinese：优化manifest读取性能，优化format对象创建性能，实际测试manifest总耗时降低到3ms以下（当然还有优化空间降低到1ms以下）。配合元数据缓存开启，orc下推功能开启，元数据格式改成orc，可以承接高并发（qps 大于1万）低延迟（整体rt 50ms以下）场景

Linked issue: close #xxx

PR #4716, opened 2024-12-16
[orc] Optimize configuration creating in orc file format

zhangyazhe:
### Purpose

1. This PR optimizes the creation performance of OrcFileFormat, with an overall performance improvement of at least 500 times. Performance tests can be found in the test files.
2. In actual tests, before optimization, creating an OrcFileFormat took about 6ms, with approximately 7 calls per query, and the scan time was about 40ms+. Under high QPS conditions (50+), the scan time could surge to several hundred milliseconds. After optimization, creating an OrcFileFormat is controlled within about 0.2ms, with approximately 7 calls per query, and the scan time is about 2ms or less.
3. Optimization principle: Each time Hadoop config is created and set, it scans system files.

### Tests

paimon-format/src/test/java/org/apache/paimon/format/orc/OrcFileFormatTest.java

Note: This PR is split from https://github.com/apache/paimon/pull/4497


paimon-format/src/main/java/org/apache/paimon/format/orc/OrcFileFormat.java
```diff
@@ -71,13 +74,29 @@ public class OrcFileFormat extends FileFormat {
     private final int writeBatchSize;
     private final boolean deletionVectorsEnabled;
 
+    private static final org.apache.hadoop.conf.Configuration emptyConf =
+            new org.apache.hadoop.conf.Configuration();
+    private static final ObjectCacheManager<Properties, Configuration> configCache =
```
JingsongLi, line 79: Just use `Cache`, `ObjectCacheManager` is too simple, we don't need a class to wrap Cache.

paimon-format/src/main/java/org/apache/paimon/format/orc/OrcFileFormat.java
```diff
@@ -71,13 +74,29 @@ public class OrcFileFormat extends FileFormat {
...
+            ObjectCacheManager.newObjectCacheManager(Duration.ofDays(365), 1000);
```
JingsongLi, line 80: .expireAfterAccess(Duration.ofMinutes(30))?

paimon-format/src/main/java/org/apache/paimon/format/orc/OrcFileFormat.java
```diff
@@ -71,13 +74,29 @@ public class OrcFileFormat extends FileFormat {
...
+
+    static {
+        emptyConf.set("paimon.empty.configuration", "paimon.empty.configuration");
```
JingsongLi, line 83: Why add an emptyConf?

JingsongLi: Maybe we can just use `new Configuration(false)`.

paimon-format/src/main/java/org/apache/paimon/format/orc/OrcFileFormat.java
```diff
@@ -71,13 +75,23 @@ public class OrcFileFormat extends FileFormat {
     private final int writeBatchSize;
     private final boolean deletionVectorsEnabled;
 
+    private static final Cache<Properties, Configuration> configCache =
+            Caffeine.newBuilder().maximumSize(100).expireAfterWrite(Duration.ofMinutes(30)).build();
+
     public OrcFileFormat(FormatContext formatContext) {
         super(IDENTIFIER);
         this.orcProperties = getOrcProperties(formatContext.options(), formatContext);
-        this.readerConf = new org.apache.hadoop.conf.Configuration();
```
JingsongLi, line 77: I just use `new Configuration(false)`, and tests passed. Do you think it is ok to just use this way?

PR #4728, opened 2024-12-17
[flink] make warehouse in Flink action optional

yuzelin:
### Purpose

warehouse is catalog conf, it can be placed into catalog conf.
And some catalog doesn't contains warehouse option.

### Tests

CompactActionITCase


JingsongLi (approved):
+1

PR #4730, opened 2024-12-17
[core] Skip case checking in catalog

Zouxxyy:
### Purpose

Checked with iceberg and delta, they don't check the case of tables or fields with hive in catalog.

So skip case checking in paimon too, and let the catalog implementation do case checking itself.

For example, table names are always lowercase in hive, so when creating a table with the same lowercase with hive catalog, hive will throw an exception:

```sql
create table t;
-- table exists exception
create table T;
```


JingsongLi:
I think it is not enable catalog case-sensitive by default, we should remove checking for `case-sensitive`?

PR #4735, opened 2024-12-18
[hive] Fix sync hms partition with data-file.path-directory

Zouxxyy:
### Purpose

1. Fix sync hms partition with `data-file.path-directory`
2. make `data-file.path-directory` Immutable
3. sync `data-file.path-directory` to hms anyway


JingsongLi (approved):
+1

PR #4737, opened 2024-12-18
[hive] Batch list tables and skip checking table exists in filesystem with hive catalog

Zouxxyy:
### Purpose

For a database containing thousands of tables on OSS, show tables may take several minutes.

It is enough to compare prop to determine whether it is a paimon table when show tables,  we can skip checking table exists in filesystem. And  the batch list api can reduce the metastore call overhead.

cost: 3min -> 4s


JingsongLi (approved):
+1

PR #4745, opened 2024-12-20
[Parquet] Revert parquet patch #3883 which tries to construct column vectors like orc

yuzelin:
### Purpose

#3883 has many bad cases，so revert it.
Note that parquet `ColumnVector` with nested type is different from orc, so related tests are cancelled.
`ArrowBatchConverter` used the `ColumnVector`, so be careful when you use `ArrowBatchConverter` with parquet format (nested type may cause error).


JingsongLi (approved):
+1

PR #4747, opened 2024-12-20
[hive] Ignore path comparison when creating external table

Zouxxyy:
### Purpose

Ignore path comparison when creating external table, because the original table may not have the 'path' attribute


JingsongLi:
+1

PR #4749, opened 2024-12-21
[parquet] parquet reader should not retrun VectorizedRowIterator for nested schema

JingsongLi:
### Purpose

Another solution for https://github.com/apache/paimon/pull/4748


yuzelin (approved):
+1

PR #4758, opened 2024-12-23
[core]Add schema validation for record-level.time-field

### LINKED ISSUE #4724: [Bug] Time field for record-level expire may block compaction
### Search before asking

### Paimon version

0.9.0

### Compute Engine

Flink

### Minimal reproduce step

```sql
CREATE TABLE T (pk INT, a INT, ts INT) 
WITH (
  'bucket'='1', 
  'primary-key'='pk', 
  'full-compaction.delta-commits' = '1', 
  'record-level.expire-time'='1s', 
  'record-level.time-field'='ts')
```

```sql
-- batchsql 1, no compact, "dirty data"
INSERT INTO T VALUES (1, 1, CAST(NULL AS INT));

-- batchsql 2，trigger compact
INSERT INTO T VALUES (2, 2, 2);    -- ERROR: Time field for record-level expire should not be null.
```

When a row of dirty data (the value of time field for record-level expire is null) writed into L0 files, the following compaction will fail. I think it is a dangerous phenomenon, if there is no snapshot to roll back to. 
So should there be stricter constraints on this time field? For example, when creating a table, make sure that the type of this field is not nullable.

### Anything else?

_No response_

herefree:
### Purpose
Add schema validation for record-level.time-field

Linked issue: close #4724

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
+++ b/paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
@@ -28,11 +28,15 @@ import org.apache.paimon.format.FileFormat;
 import org.apache.paimon.options.ConfigOption;
 import org.apache.paimon.options.Options;
 import org.apache.paimon.types.ArrayType;
+import org.apache.paimon.types.BigIntType;
 import org.apache.paimon.types.DataField;
 import org.apache.paimon.types.DataType;
+import org.apache.paimon.types.IntType;
+import org.apache.paimon.types.LocalZonedTimestampType;
 import org.apache.paimon.types.MapType;
 import org.apache.paimon.types.MultisetType;
 import org.apache.paimon.types.RowType;
+import org.apache.paimon.types.TimestampType;
 import org.apache.paimon.types.VarCharType;
 import org.apache.paimon.utils.StringUtils;
 
@@ -44,6 +48,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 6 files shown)


paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
```diff
@@ -176,6 +181,35 @@ public static void validateTableSchema(TableSchema schema) {
             }
         }
 
+        if (schema.options().containsKey(CoreOptions.RECORD_LEVEL_TIME_FIELD.key())) {
```
wwj6591812, line 184: String recordLevelTimeField = options.recordLevelTimeField();
if (recordLevelTimeField != null) {
    xxx;
}

paimon-core/src/test/java/org/apache/paimon/schema/SchemaValidationTest.java
```diff
@@ -103,4 +104,22 @@ public void testFromSnapshotConflict() {
                 .hasMessageContaining(
                         "[scan.snapshot-id] must be null when you set [scan.timestamp-millis,scan.timestamp]");
     }
+
+    @Test
+    public void testRecordLevelTimeField() {
+        Map<String, String> options = new HashMap<>();
```
wwj6591812, line 110: Map<String, String> options = new HashMap<>(2);

paimon-core/src/test/java/org/apache/paimon/table/RecordLevelExpireTest.java
```diff
@@ -19,17 +19,49 @@
...
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 import java.time.Duration;
+import java.util.UUID;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
 class RecordLevelExpireTest extends PrimaryKeyTableTestBase {
+    @Override
```
wwj6591812, line 43: Please add a blank line before @Override.

paimon-core/src/test/java/org/apache/paimon/table/RecordLevelExpireTest.java
```diff
@@ -19,17 +19,49 @@
...
+    @BeforeEach
+    public void beforeEachBase() throws Exception {
```
wwj6591812, line 45: Why add this method?

herefree: Column "col1" datatype is `DataTypes.INT()` in PrimaryKeyTableTestBase, but it should be `DataTypes.INT().notNull()` in RecordLevelExpireTest.

paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
herefree: fixed

paimon-core/src/test/java/org/apache/paimon/schema/SchemaValidationTest.java
herefree: fixed

paimon-core/src/test/java/org/apache/paimon/table/RecordLevelExpireTest.java
herefree: fixed

paimon-core/src/test/java/org/apache/paimon/table/RecordLevelExpireTest.java
wwj6591812: ok

wwj6591812:
+1

JingsongLi (approved):
+1

Issue #4471, opened 2024-11-07
[Feature] Support Variant type


Zouxxyy:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Motivation

Add Variant type to paimon data type

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

YannByron:
This feature is based on Spark 4.x: https://github.com/apache/paimon/issues/3940

Shekharrajak:
Hi,

Do we have any pending open item to support variant ?

Zouxxyy:
Oh, this should be the final PR  https://github.com/apache/paimon/pull/7913 for the Spark variant integration. I will close this issue.

PR #4764, opened 2024-12-24
[spark] Integrate Variant with Spark4

Zouxxyy:
### Purpose

Subtask of #4471

- Integrate Variant with Spark4

```sql
CREATE TABLE t (v VARIANT) USING paimon;

INSERT INTO t VALUES parse_json('{"age":26,"city":"Beijing"}');

SELECT variant_get(v, '$.age', 'int') FROM t;
```


JingsongLi (approved):
+1

PR #4765, opened 2024-12-24
[common] Using a faster deserialization method in RoaringBitmap32

Tan-JiaLiang:
### Purpose

In `RoaringBitmap`, the `deserialize(java.io.DataInput, byte[])` function has a better performance to deserialize the `BitmapContainer`.

### Tests

No.

### API and Format

No.

### Documentation

No.

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/utils/RoaringBitmap32.java
+++ b/paimon-common/src/main/java/org/apache/paimon/utils/RoaringBitmap32.java
@@ -85,8 +85,19 @@ public class RoaringBitmap32 {
         roaringBitmap.serialize(out);
     }
 
+    public byte[] serialize() {
+        roaringBitmap.runOptimize();
+        ByteBuffer buffer = ByteBuffer.allocate(roaringBitmap.serializedSizeInBytes());
+        roaringBitmap.serialize(buffer);
+        return buffer.array();
+    }
+
     public void deserialize(DataInput in) throws IOException {
-        roaringBitmap.deserialize(in);
+        roaringBitmap.deserialize(in, null);
+    }
+
+    public void deserialize(ByteBuffer buffer) throws IOException {
+        roaringBitmap.deserialize(buffer);
     }
 
     @Override
@@ -103,17 +114,6 @@ public class RoaringBitmap32 {
 
     public void clear() {
         roaringBitmap.clear();
-    }
-
-    public byte[] serialize() {
-        roaringBitmap.runOptimize();
-        ByteBuffer buffer = ByteBuffer.allocate(roaringBitmap.serializedSizeInBytes());
-        roaringBitmap.serialize(buffer);
-        return buffer.array();
-    }
-
-    public void deserialize(byte[] rbmBytes) throws IOException {
-        roaringBitmap.deserialize(ByteBuffer.wrap(rbmBytes));
     }
 
     public void flip(final long rangeStart, final long rangeEnd) {
--- a/paimon-core/src/main/java/org/apache/paimon/deletionvectors/DeletionVector.java
+++ b/paimon-core/src/main/java/org/apache/paimon/deletionvectors/DeletionVector.java
@@ -103,7 +103,7 @@ public interface DeletionVector {
             ByteBuffer buffer = ByteBuffer.wrap(bytes);
             int magicNum = buffer.getInt();
             if (magicNum == BitmapDeletionVector.MAGIC_NUMBER) {
-                return BitmapDeletionVector.deserializeFromByteBuffer(buffer.slice());
+                return BitmapDeletionVector.deserializeFromByteBuffer(buffer);
             } else {
                 throw new RuntimeException("Invalid magic number: " + magicNum);
             }
--- a/paimon-core/src/main/java/org/apache/paimon/deletionvectors/DeletionVector.java
+++ b/paimon-core/src/main/java/org/apache/paimon/deletionvectors/DeletionVector.java
@@ -26,9 +26,9 @@ import org.apache.paimon.table.source.DeletionFile;
 
 import javax.annotation.Nullable;
 
-import java.io.ByteArrayInputStream;
 import java.io.DataInputStream;
 import java.io.IOException;
+import java.nio.ByteBuffer;
 import java.util.List;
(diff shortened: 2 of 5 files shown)


JingsongLi:
Can you provide the benchmark result?

Tan-JiaLiang:
I'm trying to understand how to use the benchmark in paimon, and I hope my benchmark using in correct.

bitmap32-deserialize-benchmark | Best/Avg Time(ms) | Row Rate(K/s) | Per Row(ns) | Relative
-------- | ----- |  ----- |  ----- |  -----
bitmap32-deserialize-benchmark_without-buffer | 239 /  259 | 0.4 | 2385027.0 | 1.0X
bitmap32-deserialize-benchmark_with-buffer  | 2 /    2 | 54.1 | 18500.0 | 128.9X

JingsongLi:
Thanks for your information.

paimon-common/src/main/java/org/apache/paimon/utils/RoaringBitmap32.java
```diff
@@ -89,6 +91,10 @@ public void deserialize(DataInput in) throws IOException {
         roaringBitmap.deserialize(in);
     }
 
+    public void deserialize(DataInput in, @Nullable byte[] buffer) throws IOException {
+        roaringBitmap.deserialize(in, buffer);
```
JingsongLi, line 95: if `deserialize(in, null)` is better, can you just modify `RoaringBitmap32.deserialize(DataInput)`?

Tan-JiaLiang: I totally agree with this, but be aware that it may take up 8KB of memory during deserialization, although I think it's reasonable.

paimon-common/src/main/java/org/apache/paimon/utils/RoaringBitmap32.java
```diff
@@ -86,7 +86,7 @@ public void serialize(DataOutput out) throws IOException {
     }
 
     public void deserialize(DataInput in) throws IOException {
-        roaringBitmap.deserialize(in);
+        roaringBitmap.deserialize(in, null);
```
Zouxxyy, line 89: Maybe we can tested `roaringBitmap.deserialize(ByteBuffer bbf)` directly?  We actually know the exact byte count of the bitmap

Tan-JiaLiang:
Add the `deserialize(ByteBuffer)` benchmark

bitmap32-deserialize-benchmark | Best/Avg Time(ms) | Row Rate(K/s) | Per Row(ns) | Relative
-- | -- | -- | -- | --
benchmark_deserialize(DataInput) | 242 / 254 | 0.4 | 2417750.0 | 1.0X
benchmark_deserialize(DataInput, byte[]) | 2 / 3 | 49.8 | 20087.0 | 120.4X
benchmark_deserialize(ByteBuffer) | 1 / 2 | 75.3 | 13287.0 | 182.0X

paimon-common/src/main/java/org/apache/paimon/utils/RoaringBitmap32.java
Tan-JiaLiang: Did you mean adding a benchmark?

Zouxxyy: From the test results, it is more optimal, and I have checked other formats, which also use it. we can just use it.

Tan-JiaLiang: The premise of ByteBuffer is that we need to know the actual size of the Bitmap. This seems to require recording the actual size at serialisation time.

Tan-JiaLiang: We can using it in delection vector bitmap and rbm aggregate function, but the bitmap and bsi index can not for now.

Zouxxyy: > roaringBitmap.deserialize(ByteBuffer bbf)

Does the bitmap index not record the byte size? Would you be willing to integrate `roaringBitmap.deserialize(ByteBuffer bbf)` into a usable place such as dv?

JingsongLi (commented):
Hi @Tan-JiaLiang , can you use null for bitmap index, and use ByteBuffer for deletion vectors?

Tan-JiaLiang:
@JingsongLi @Zouxxyy Can you help to review again?

paimon-core/src/main/java/org/apache/paimon/deletionvectors/DeletionVector.java
```diff
@@ -99,11 +99,11 @@ default boolean checkedDelete(long position) {
...
     static DeletionVector deserializeFromBytes(byte[] bytes) {
-        try (ByteArrayInputStream bis = new ByteArrayInputStream(bytes);
-                DataInputStream dis = new DataInputStream(bis)) {
-            int magicNum = dis.readInt();
+        try {
+            ByteBuffer buffer = ByteBuffer.wrap(bytes);
+            int magicNum = buffer.getInt();
             if (magicNum == BitmapDeletionVector.MAGIC_NUMBER) {
-                return BitmapDeletionVector.deserializeFromDataInput(dis);
+                return BitmapDeletionVector.deserializeFromByteBuffer(buffer.slice());
```
Zouxxyy, line 106: Why is slice needed here

Tan-JiaLiang: This is really just a programming convention to ensure that the bytes passed in are bitmaps only. I take a look that it is already using in `RoaringArray#deserialize`, so it's redundant here.

https://github.com/RoaringBitmap/RoaringBitmap/blob/60dcdb58846037ea8ea712317a0ac887e00fe25b/roaringbitmap/src/main/java/org/roaringbitmap/RoaringArray.java#L544-L550

Tan-JiaLiang:
Seems the CDC e2e test is still having some thread safe problem, I will take a look in next week. Can someone help to trigger the build one more time?

Zouxxyy (approved):
+1

PR #4766, opened 2024-12-24
[core] External Path in DataFileMeta should be the file path

JingsongLi:
### Purpose

Fix after https://github.com/apache/paimon/pull/4761
- External Path in DataFileMeta should be the file path instead of directory.
- And clean interface methods.


neuyilan:
LGTM

PR #4769, opened 2024-12-24
[rest] Fix GetTableResponse should return fields

JingsongLi:
#### Modify GetTableResponse:

```
public class GetTableResponse implements RESTResponse {

    private static final String FIELD_PATH = "path";
    private static final String FIELD_SCHEMA_ID = "schema-id";
    private static final String FIELD_SCHEMA = "schema";

    @JsonProperty(FIELD_PATH)
    private final String path;

    @JsonProperty(FIELD_SCHEMA_ID)
    private final long schemaId;

    @JsonProperty(FIELD_SCHEMA)
    private final Schema schema;
}
```

We should use Schema instead of TableSchema to keep one serialization implementation.

#### UpdateTable to RenameTable

UpdateTable should be refactored, because we need a unified `SchemaChange` to support changes.


wwj6591812:
+1

jerry-024:
+1

PR #4772, opened 2024-12-24
[flink] Introduce max_two_pt for Flink lookup join

yuzelin:
### Purpose

There is a scenario: the Join job is calculating the T day data, and the dim table is writing the (T + 1) day data, we can load two partitions data, so the Join job won't lose result when day is converting from T to T + 1.


JingsongLi (approved):
+1

PR #4773, opened 2024-12-24
[core] optimize the binlog table read performance

Aitozi:
### Purpose

Avoid to construct the rowType for each record.

image


JingsongLi:
+1

PR #4777, opened 2024-12-25
[core] SortLookupStoreWriter should support empty record

JingsongLi:
### Purpose

In some cases, for example, all records in remote file has been marked deleted. It is possible to build an empty local file to lookup. `BlockWriter.finish` will throw a `IllegalStateException`.

We can fix it by supporting empty record in `SortLookupStoreWriter`.


yuzelin (approved):
+1

PR #4779, opened 2024-12-25
Add check of older_than when RemoveOrphanFiles

wwj6591812:
### Purpose

Now when clean OrphanFiles by call RemoveOrphanFilesProcedure or RemoveOrphanFilesAction, If the argument older_than is large then now, dataFiles that are currently being written and not referenced by snapshots will be mistakenly cleaned up.
So add a check of the argument older_than.

Linked issue: close #xxx


JingsongLi (approved):
+1

PR #4781, opened 2024-12-26
[flink] Add a action/procedure to remove unexisting files from manifests

tsreaper:
### Purpose

This PR adds an action/procedure to remove unexisting files from manifests. It has the following use cases:

* There is currently a known case when unexisting data file might be written into manifest. Consider a write-only job (W) plus a dedicated compaction job (C):
    1. W commits a snapshot with file F. Then W constantly fails and restarts, each time before we can retry the commit (or W is stopped, creating Flink savepoint S).
    2. C compacts F into a larger file, so F is now deleted from the manifest.
    3. Before the compact snapshot expires, but after all snapshots created by W expires, W comes back to normal (or restarted from savepoint S). As W cannot find its previous snapshot, it assumes that this snapshot has not been committed, so file F is committed to the manifest once again.
    4. When the compact snapshot expires, file F will be deleted from the file system. Now F is in the manifest, but not on the file system. With this situation, user might want to remove F from the manifest to continue reading the table.
* User deletes a data file by mistake (for example, by incorrectly setting the time threshold
    for orphan files cleaning). If the user can tolerate skipping some records when consuming
    this table, he can also use this action to remove the file from manifest.

Note that user is on his own risk using this procedure, which may cause data loss when used outside from the use cases above.

### Tests

Unit tests and IT cases.

### API and Format

No format changes.

### Documentation

Document is also added.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/ListUnexistingFiles.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/ListUnexistingFiles.java
@@ -19,7 +19,6 @@
 package org.apache.paimon.operation;
 
 import org.apache.paimon.data.BinaryRow;
-import org.apache.paimon.fs.FileIO;
 import org.apache.paimon.fs.Path;
 import org.apache.paimon.io.DataFileMeta;
 import org.apache.paimon.io.DataFilePathFactory;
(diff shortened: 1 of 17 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/RemoveUnexistingFilesAction.java
```diff
@@ -0,0 +1,272 @@
...
+        this.parallelism = parallelism;
+        return this;
+    }
+
+    @Override
+    public void build() throws Exception {
+        buildDataStream();
+    }
+
+    public DataStream<String> buildDataStream() throws Exception {
```
wwj6591812, line 114: This method will not throw any Exception. So we can remove "throws Exception".

wwj6591812, line 114: This method will not throw any Exception. So we can remove "throws Exception".

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/RemoveUnexistingFilesAction.java
```diff
@@ -0,0 +1,272 @@
...
+        return this;
+    }
+
+    public RemoveUnexistingFilesAction withParallelism(int parallelism) {
+        this.parallelism = parallelism;
+        return this;
+    }
+
+    @Override
+    public void build() throws Exception {
```
wwj6591812, line 110: Check ExecutionOptions.RUNTIME_MODE == RuntimeExecutionMode.BATCH ?

paimon-core/src/main/java/org/apache/paimon/operation/ListUnexistingFiles.java
```diff
@@ -0,0 +1,68 @@
...
+                        .withPartitionFilter(Collections.singletonList(partition))
+                        .plan()
+                        .splits();
+        for (Split split : splits) {
+            DataSplit dataSplit = (DataSplit) split;
+            DataFilePathFactory dataFilePathFactory =
+                    pathFactory.createDataFilePathFactory(partition, dataSplit.bucket());
+            for (DataFileMeta meta : dataSplit.dataFiles()) {
+                Path path = dataFilePathFactory.toPath(meta);
+                if (!fileIO.exists(path)) {
```
JingsongLi, line 60: It is better to use a thread pool, this invoking is very slow for object store.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/RemoveUnexistingFilesAction.java
```diff
@@ -0,0 +1,449 @@
...
+        public void open() throws Exception {
+            reuse = new BinaryRow(table.schema().partitionKeys().size());
+            pathFactory = table.store().pathFactory();
+        }
+
+        @Override
+        public void processElement(StreamRecord<byte[]> record) throws Exception {
+            byte[] bytes = record.getValue();
+            reuse.pointTo(MemorySegment.wrap(bytes), 0, bytes.length);
+            Path partitionPath = pathFactory.getPartitionPath(reuse);
```
JingsongLi, line 221: We need to consider external path for table...

I think it is better to still use fileio.exists method...

paimon-core/src/main/java/org/apache/paimon/operation/ListUnexistingFiles.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    public Map<Integer, Map<String, DataFileMeta>> list(BinaryRow partition) throws Exception {
+        Map<Integer, Map<String, DataFileMeta>> result = new ConcurrentHashMap<>();
+        List<Split> splits =
+                table.newScan()
+                        .withPartitionFilter(Collections.singletonList(partition))
+                        .plan()
+                        .splits();
+        ThreadPoolUtils.randomlyOnlyExecute(
```
JingsongLi, line 64: return a List to collect files

JingsongLi (commented):
Can you also add implementation for spark?

JingsongLi (approved):
+1

PR #4782, opened 2024-12-26
[core] CoreOptions.fileFormat is cpu expensive, because FileFormat initalization is costly. As much as we can, reduce the rate.

leaves12138:
### Purpose

Just don't call `options.fileformat` as much as you want.


paimon-common/src/main/java/org/apache/paimon/format/FileFormat.java
```diff
@@ -88,7 +103,7 @@ public static FileFormat fromIdentifier(String identifier, Options options) {
 
     /** Create a {@link FileFormat} from format identifier and format options. */
     public static FileFormat fromIdentifier(String identifier, FormatContext context) {
-        return fromIdentifier(identifier, context, FileFormat.class.getClassLoader())
+        return getFileFormatFromLoadedCache(identifier, context)
```
JingsongLi, line 106: Use `org.apache.paimon.factories.FactoryUtil`

paimon-core/src/main/java/org/apache/paimon/AbstractFileStore.java
```diff
@@ -101,11 +104,13 @@ protected AbstractFileStore(
         this.writeManifestCache =
                 SegmentsCache.create(
                         options.pageSize(), options.writeManifestCache(), Long.MAX_VALUE);
+        this.fileFormat = options.fileFormat();
+        this.pathFactory = pathFactory(fileFormat.getFormatIdentifier());
```
JingsongLi, line 108: Don't create path factory here... It should be created one by one

paimon-common/src/main/java/org/apache/paimon/factories/BaseFactoryUtil.java
```diff
@@ -0,0 +1,72 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import java.util.ServiceLoader;
+
+/** Base Factory Util. */
+public class BaseFactoryUtil {
```
JingsongLi, line 32: Do not introduce this method, it has no abstraction... Just put static method into `FactoryUtil`.

paimon-core/src/main/java/org/apache/paimon/AppendOnlyFileStore.java
```diff
@@ -120,6 +121,7 @@ public AppendOnlyFileStoreWrite newWrite(
                     commitUser,
                     rowType,
                     partitionType,
+                    options.fileFormat(),
```
JingsongLi, line 124: Why we need to modify this?

JingsongLi (approved):
+1

ranxianglei:
extends Factory is more batter

ranxianglei:
This pr is from my pr https://github.com/apache/paimon/pull/4497 , You spared me a private discussion and plagiarized my PR. Without my consent, you quickly merged the PR and then closed my original PR at the same time. 
Then tell me the conclusion of your discussion in this pr https://github.com/apache/paimon/pull/4813 

Why do this? Do you follow the Apache open source spirit? stop your mistakes please  @JingsongLi @leaves12138

JingsongLi:
Even in https://github.com/apache/paimon/pull/4813 , you have never provided a better version or convincing reasons.

Issue #4540, opened 2024-11-18
[Feature] Support REST Catalog


jerry-024:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Motivation

Provide REST Catalog to implement Catalog. By REST Catalog:
- the user could easily access their catalog system
- less dependency
- customize the server's logic

### Solution

[PIP-28: Introduce REST Catalog](https://cwiki.apache.org/confluence/display/PAIMON/PIP-28%3A+Introduce+REST+Catalog)

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

JiaQiTang98:
Hi, is there any plan to support a api like plan in Iceberg, it will return the tasks for scaning table(include data files and delete files).

PR #4786, opened 2024-12-26
[core] Support partition API

jerry-024:
### Purpose

Support partition API
-  createPartition
-  dropPartition(ps: just metadata)
-  listPartitions

Linked issue: #4540 

### Tests
- RESTObjectMapperTest.createPartitionRequestParseTest
- RESTObjectMapperTest.dropPartitionRequestParseTest
- RESTObjectMapperTest.listPartitionsResponseParseTest
- RESTCatalogTest.testCreatePartition
- RESTCatalogTest.testCreatePartitionWhenTableNotExist
- RESTCatalogTest.testCreatePartitionWhenTableNoPermissionException
- RESTCatalogTest.testDropPartition
- RESTCatalogTest.testDropPartitionWhenPartitionNoExist
- RESTCatalogTest.testDropPartitionWhenTableNoPermission
- RESTCatalogTest.testDropPartitionWhenTableNoExist
- RESTCatalogTest.testListPartitionsWhenMetastorePartitionedIsTrue
- RESTCatalogTest.testListPartitionsFromFile
- RESTCatalogTest.convertToPartitionEntryTest


paimon-core/src/main/java/org/apache/paimon/rest/responses/ListPartitionsResponse.java
```diff
@@ -0,0 +1,126 @@
...
+        public static final String FIELD_RECORD_COUNT = "recordCount";
+        public static final String FIELD_FILE_SIZE_IN_BYTES = "fileSizeInBytes";
+        public static final String FIELD_FILE_COUNT = "fileCount";
+        public static final String FIELD_LAST_FILE_CREATION_TIME = "lastFileCreationTime";
+
+        @JsonProperty(FIELD_SPEC)
+        private final Map<String, String> spec;
+
+        @JsonProperty(FIELD_PARTITION_TYPE)
+        private final RowType partitionType;
```
JingsongLi, line 66: Do not add type in this class.

paimon-core/src/main/java/org/apache/paimon/rest/RESTCatalog.java
```diff
@@ -360,17 +376,45 @@ public void dropTable(Identifier identifier, boolean ignoreIfNotExists)
...
     }
 
     @Override
     public void dropPartition(Identifier identifier, Map<String, String> partitions)
-            throws TableNotExistException, PartitionNotExistException {}
+            throws TableNotExistException, PartitionNotExistException {
+        checkNotSystemTable(identifier, "dropPartition");
+        dropPartitionMetadata(identifier, partitions);
+        Table table = getTable(identifier);
+        if (table != null) {
```
JingsongLi, line 400: it is null? I think maybe getTable will throw TableNotExistsException.

paimon-core/src/main/java/org/apache/paimon/rest/RESTCatalogOptions.java
```diff
@@ -71,4 +71,10 @@ public class RESTCatalogOptions {
                     .stringType()
                     .noDefaultValue()
                     .withDescription("REST Catalog auth token provider path.");
+
+    public static final ConfigOption<Boolean> METASTORE_PARTITIONED =
```
JingsongLi, line 75: Use `CoreOptions.metastore.partitioned-table`.

paimon-core/src/main/java/org/apache/paimon/rest/RESTCatalog.java
```diff
@@ -360,17 +376,45 @@ public void dropTable(Identifier identifier, boolean ignoreIfNotExists)
...
     public void createPartition(Identifier identifier, Map<String, String> partitionSpec)
             throws TableNotExistException {
-        throw new UnsupportedOperationException();
+        try {
+            CreatePartitionRequest request = new CreatePartitionRequest(identifier, partitionSpec);
+            client.post(
+                    resourcePaths.partitions(
+                            identifier.getDatabaseName(), identifier.getTableName()),
+                    request,
+                    SuccessResponse.class,
```
JingsongLi, line 385: here should return Partition?

paimon-core/src/main/java/org/apache/paimon/rest/RESTCatalog.java
```diff
@@ -434,6 +528,23 @@ protected GetTableResponse getTableResponse(Identifier identifier)
...
+            throws TableNoPermissionException {
+        try {
+            DropPartitionRequest request = new DropPartitionRequest(partitions);
+            return client.delete(
+                    resourcePaths.partitions(
+                            identifier.getDatabaseName(), identifier.getTableName()),
+                    request,
+                    headers());
+        } catch (NoSuchResourceException ignore) {
+            return new SuccessResponse();
```
JingsongLi, line 542: I think here should throw table not exist exception?

paimon-core/src/main/java/org/apache/paimon/rest/RESTCatalog.java
```diff
@@ -434,6 +528,23 @@ protected GetTableResponse getTableResponse(Identifier identifier)
         }
     }
 
+    protected SuccessResponse dropPartitionMetadata(
+            Identifier identifier, Map<String, String> partitions)
+            throws TableNoPermissionException {
+        try {
+            DropPartitionRequest request = new DropPartitionRequest(partitions);
+            return client.delete(
```
JingsongLi, line 536: we should return Partition?

paimon-core/src/main/java/org/apache/paimon/rest/RESTCatalog.java
jerry-024: when the partition's metadata doesn't exist in metadata, there may still be data in the filesystem. So we ignore this exception when dropping the partition.

JingsongLi: should return false instead of NoSuchResourceException?

paimon-core/src/main/java/org/apache/paimon/rest/RESTCatalog.java
```diff
@@ -434,6 +524,23 @@ protected GetTableResponse getTableResponse(Identifier identifier)
         }
     }
 
+    protected boolean dropPartitionMetadata(Identifier identifier, Map<String, String> partitions)
```
JingsongLi, line 527: Why need these tests invoking this method? Why not invoking catalog methods?

JingsongLi:
+1

Issue #4785, opened 2024-12-26
[Bug][parquet] Read struct(null, null) should return struct(null, null) instead of null


Zouxxyy:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Minimal reproduce step

```sql
CREATE TABLE t (i INT, s STRUCT<f1: INT, f2: INT>);
INSERT INTO t VALUES (1, STRUCT(null, null));
SELECT * FROM t;
```
should return [null, null] instead of null

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #4787, opened 2024-12-26
[spark] Fix writing null struct col

Zouxxyy:
### Purpose

```sql
CREATE TABLE t (i INT, s STRUCT<f1: INT, f2: INT>);
INSERT INTO t VALUES ((1, null), (2, STRUCT(null, null));
SELECT s FROM t;
```

before
```
[null,null], [null,null]
```

after
```
-- orc and avro
null, [null,null]
```

There is still bug with reading parquet, which fix involves core changes, add an issue to track it https://github.com/apache/paimon/issues/4785


JingsongLi (approved):
+1

PR #4797, opened 2024-12-29
[core] Alter table set option with empty key filter

xuzifu666:
### Purpose

Linked issue: close #xxx
Currently paimon set properties(options) not filter empty key:

ALTER TABLE T SET ('' = '2 mb');

then show create table : 
CREATE TABLE `PAIMON`.`default`.`T` (
  `a` INT,
  `b` INT,
  `c` INT
) WITH (
  'scan.infer-parallelism' = 'false',
  'path' = '/var/folders/kb/wgkwnwq156g6kcpgldkyjcf00000gp/T/junit3546018753241373391/a5ced1b0-a37b-4879-b035-c14a351168d3/default.db/T',
 **'' = '2 mb'**
  'write-manifest-cache' = '1 mb'
)
this pr filter the empty option key

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
@@ -45,6 +45,7 @@ import org.apache.paimon.table.system.CatalogOptionsTable;
 import org.apache.paimon.table.system.SystemTableLoader;
 import org.apache.paimon.types.RowType;
 import org.apache.paimon.utils.Preconditions;
+import org.apache.paimon.utils.StringUtils;
 
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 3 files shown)


paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
```diff
@@ -547,6 +549,18 @@ private void validateCustomTablePath(Map<String, String> options) {
         }
     }
 
+    private List<SchemaChange> filterNonEmptyKeyInSchemaChange(List<SchemaChange> changes) {
+        List<SchemaChange> schemaChanges = new ArrayList<>();
+        for (SchemaChange schemaChange : changes) {
+            if (schemaChange instanceof SchemaChange.SetOption
+                    && StringUtils.isEmpty(((SchemaChange.SetOption) schemaChange).key())) {
```
wwj6591812, line 556: What about throw a exception instead of continue?

xuzifu666: Thanks, empty is a non-sense config, maybe filter directly also OK.

xuzifu666:
Communicate with spark community，they suggest spark side also not do the change，this mainly impact in indicate for user， so close it first to keep with spark， related pr：https://github.com/apache/spark/pull/49329

PR #4798, opened 2024-12-29
[flink] Add new TypeMapping options to ignore type change in special cases

### LINKED ISSUE #4783: [Feature] Ignore decimal type evolution in special case
### Search before asking

### Motivation

We use ticdc to collect data from tidb and send to kafka in canal-json format, then sink the data with paimon-flink-action.

But actually only type name are collected and sent to kafka for the metadata part. For example, the decimal fields, only the name 'decimal' provided, without the precision and scale. 

In paimon-flink-action, the missing precision and scale will be set to maximum, which is decimal(38,16). That is too large and not expected.
In this case, the original format is prefer, which means not changing the deciaml type during cdc. 

### Solution

Add a type-mappint option for kafka cdc, like 
```shell
# to ignore only decimal change
--type_mapping decimal-no-change
# to ingore any type change
--type_mapping no_change
``` 
And ignore the type change in $$RichEventParser::parseSchemaChange$$

### Anything else?

Enlarging type also occurs to char/varchar, which is acceptable.

JackeyLee007:
### Purpose

Linked issue: close #4783

### Tests

org.apache.paimon.flink.action.cdc.mysql.MySqlCdcTypeMappingITCase#testDecimalNoChange
org.apache.paimon.flink.action.cdc.mysql.MySqlCdcTypeMappingITCase#testNoChange

### API and Format

No
### Documentation

Yes and doc provided

### DIFF
--- a/paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/sink/cdc/UpdatedDataFieldsProcessFunctionBase.java
+++ b/paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/sink/cdc/UpdatedDataFieldsProcessFunctionBase.java
@@ -219,7 +219,8 @@ public abstract class UpdatedDataFieldsProcessFunctionBase<I, O> extends Process
         }
 
         boolean allowTypeChange =
-                !this.typeMapping.containsMode(TypeMapping.TypeMappingMode.NO_CHANGE);
+                this.typeMapping == null
+                        || !this.typeMapping.containsMode(TypeMapping.TypeMappingMode.NO_CHANGE);
 
         List<SchemaChange> result = new ArrayList<>();
         for (DataField newField : updatedDataFields) {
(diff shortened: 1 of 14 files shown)


paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/sink/cdc/UpdatedDataFieldsProcessFunctionBase.java
```diff
@@ -214,6 +218,10 @@ protected List<SchemaChange> extractSchemaChanges(
             oldFields.put(oldField.name(), oldField);
         }
 
+        boolean allowTypeChange =
+                this.typeMapping == null
+                        || !this.typeMapping.containsMode(TypeMapping.TypeMappingMode.NO_CHANGE);
+
```
wwj6591812, line 224: Where the "DECIMAL_NO_CHANGE" use?

JackeyLee007: Thanks for review, I'll check that again.

JackeyLee007:
Close for now

PR #4805, opened 2024-12-31
[flink] add IT to flink when use RESTCatalog

jerry-024:
### Purpose

Linked issue: #4540 
- update RESTCatalogTest to extend CatalogTestBase
- add RESTCatalogITCase to support IT case


JingsongLi (approved):
Looks good to me!

PR #4807, opened 2024-12-31
[core] Introduce Partition to Catalog.listPartitions

JingsongLi:
### Purpose

Now `Catalog.listPartitions` returns `PartitionEntry`, but in `PartitionEntry`, the partition is `BinaryRow`.

We should not return `BinaryRow` in Catalog interfaces, only `Map<String, String>` can be used for partition spec. `BinaryRow` is really difficult to be used.

 `Catalog.listPartitions` was introduced in Paimon 1.0, we should also modify Paimon 1.0, it is a blocker for Paimon 1.0 releasing.


paimon-core/src/main/java/org/apache/paimon/partition/Partition.java
```diff
@@ -0,0 +1,135 @@
...
+            @JsonProperty(FIELD_LAST_FILE_CREATION_TIME) long lastFileCreationTime) {
+        this.spec = spec;
+        this.recordCount = recordCount;
+        this.fileSizeInBytes = fileSizeInBytes;
+        this.fileCount = fileCount;
+        this.lastFileCreationTime = lastFileCreationTime;
+    }
+
+    @JsonGetter(FIELD_SPEC)
+    public Map<String, String> getSpec() {
```
yuzelin, line 75: The name style of getters should be the same with `Snapshot` (spec).

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/partition/PartitionStatisticsReporter.java
```diff
@@ -22,7 +22,7 @@
 import org.apache.paimon.fs.Path;
 import org.apache.paimon.io.DataFileMeta;
 import org.apache.paimon.metastore.MetastoreClient;
-import org.apache.paimon.metastore.PartitionStats;
```
yuzelin, line 25: I think we can delete class `PartitionStats`.

yuzelin (approved):
+1

PR #4808, opened 2024-12-31
[core] Introduce incremental-to-auto-tag for reading changes between auto tags and let incremental-between be tag-first

yuzelin:
### Purpose

incremental-to: see doc.
tag-first: when using incremental-between, check if the args can be parsed to tags firstly.


paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -1080,6 +1080,13 @@ public class CoreOptions implements Serializable {
                             "Read incremental changes between start timestamp (exclusive) and end timestamp, "
                                     + "for example, 't1,t2' means changes between timestamp t1 and timestamp t2.");
 
+    public static final ConfigOption<String> INCREMENTAL_TO =
+            key("incremental-to")
```
JingsongLi, line 1084: incremental-to-auto-tag

JingsongLi (approved):
+1

PR #4813, opened 2025-01-02
[core][format] op format factory mv to cache .

ranxianglei:
### Purpose

from https://github.com/apache/paimon/pull/4497

better impl for https://github.com/apache/paimon/pull/4782

Linked issue: close #xxx


leaves12138:
I tihink `FileFormatFactory` is a special interface, we don't need to change. Many third-party file formats depend on this service loader. We could just distinguish it from Factory.

ranxianglei:
The significance of unified management is that many methods do not need to be implemented multiple times.
For example, this PR https://github.com/apache/paimon/pull/4818 implements the factory coverage mechanism.

Third-party file formats  just rename `org.apache.paimon.format.FileFormatFactory` to `org.apache.paimon.factories.Factory` then works. 

@leaves12138

leaves12138:
That's not beckoning enough to make this change. If you think FormatFactoryUtils is too redundant, we can re-use more code.

leaves12138:
Can you fix this as what I mentioned？ Still regard FileFormatFactory as an independent interface.

leaves12138:
> The significance of unified management is that many methods do not need to be implemented multiple times. For example, this PR #4818 implements the factory overide mechanism.
> 
> Third-party file formats just rename `org.apache.paimon.format.FileFormatFactory` to `org.apache.paimon.factories.Factory` then works.
> 
> @leaves12138

That need to handle files in META-INFO/services also

ranxianglei:
Sure， I am writing @leaves12138

ranxianglei:
done.Compatible with existing third-party factory implementations .

leaves12138:
> done.Compatible with existing third-party factory implementations .

We don't need to combine META-INF/services/org.apache.paimon.format.FileFormatFactory and META-INF/services/org.apache.paimon.factories.Factory. `FileFormatFactory` could be regarded as a special.

PR #4820, opened 2025-01-02
[core] Refactor TagManager to remove unnecessary tag existence check

yuzelin:
### Purpose

Move all tag existence check into TagManager methods;
Unify get tag. method.


JingsongLi (approved):
+1

PR #4824, opened 2025-01-03
[spark] Clean empty directory after removing orphan files

askwang:
### Purpose

Empty directories can not be cleaned after executing `remove_orphan_files`.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/LocalOrphanFilesClean.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/LocalOrphanFilesClean.java
@@ -70,6 +70,8 @@ public class LocalOrphanFilesClean extends OrphanFilesClean {
 
     private final List<Path> deleteFiles;
 
+    private final boolean dryRun;
+
     private final AtomicLong deletedFilesLenInBytes = new AtomicLong(0);
 
     private Set<String> candidateDeletes;
@@ -79,16 +81,20 @@ public class LocalOrphanFilesClean extends OrphanFilesClean {
     }
 
     public LocalOrphanFilesClean(FileStoreTable table, long olderThanMillis) {
-        this(table, olderThanMillis, path -> table.fileIO().deleteQuietly(path));
+        this(table, olderThanMillis, path -> table.fileIO().deleteQuietly(path), false);
     }
 
     public LocalOrphanFilesClean(
-            FileStoreTable table, long olderThanMillis, SerializableConsumer<Path> fileCleaner) {
+            FileStoreTable table,
+            long olderThanMillis,
+            SerializableConsumer<Path> fileCleaner,
+            boolean dryRun) {
         super(table, olderThanMillis, fileCleaner);
         this.deleteFiles = new ArrayList<>();
         this.executor =
                 createCachedThreadPool(
                         table.coreOptions().deleteFileThreadNum(), "ORPHAN_FILES_CLEAN");
+        this.dryRun = dryRun;
     }
 
     public CleanOrphanFilesResult clean()
@@ -129,7 +135,9 @@ public class LocalOrphanFilesClean extends OrphanFilesClean {
         candidateDeletes.clear();
 
         // clean empty directory
-        cleanEmptyDirectory(deleteFiles);
+        if (!dryRun) {
+            cleanEmptyDirectory(deleteFiles);
+        }
 
         return new CleanOrphanFilesResult(
                 deleteFiles.size(), deletedFilesLenInBytes.get(), deleteFiles);
@@ -230,7 +238,8 @@ public class LocalOrphanFilesClean extends OrphanFilesClean {
             @Nullable String tableName,
             long olderThanMillis,
             SerializableConsumer<Path> fileCleaner,
-            @Nullable Integer parallelism)
+            @Nullable Integer parallelism,
+            boolean dryRun)
             throws Catalog.DatabaseNotExistException, Catalog.TableNotExistException {
         List<String> tableNames = Collections.singletonList(tableName);
         if (tableName == null || "*".equals(tableName)) {
@@ -259,7 +268,7 @@ public class LocalOrphanFilesClean extends OrphanFilesClean {
 
             orphanFilesCleans.add(
                     new LocalOrphanFilesClean(
-                            (FileStoreTable) table, olderThanMillis, fileCleaner));
+                            (FileStoreTable) table, olderThanMillis, fileCleaner, dryRun));
         }
 
         return orphanFilesCleans;
@@ -271,7 +280,8 @@ public class LocalOrphanFilesClean extends OrphanFilesClean {
             @Nullable String tableName,
             long olderThanMillis,
             SerializableConsumer<Path> fileCleaner,
-            @Nullable Integer parallelism)
+            @Nullable Integer parallelism,
+            boolean dryRun)
             throws Catalog.DatabaseNotExistException, Catalog.TableNotExistException {
         List<LocalOrphanFilesClean> tableCleans =
                 createOrphanFilesCleans(
@@ -280,7 +290,8 @@ public class LocalOrphanFilesClean extends OrphanFilesClean {
                         tableName,
                         olderThanMillis,
                         fileCleaner,
-                        parallelism);
+                        parallelism,
+                        dryRun);
 
         ExecutorService executorService =
                 Executors.newFixedThreadPool(Runtime.getRuntime().availableProcessors());
--- a/paimon-core/src/main/java/org/apache/paimon/operation/LocalOrphanFilesClean.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/LocalOrphanFilesClean.java
@@ -136,26 +136,28 @@ public class LocalOrphanFilesClean extends OrphanFilesClean {
 
         // clean empty directory
         if (!dryRun) {
-            cleanEmptyDirectory(deleteFiles);
+            cleanEmptyDataDirectory(deleteFiles);
         }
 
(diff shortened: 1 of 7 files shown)


paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/orphan/SparkOrphanFilesClean.scala
```diff
@@ -137,6 +138,7 @@ case class SparkOrphanFilesClean(
         it =>
           var deletedFilesCount = 0L
           var deletedFilesLenInBytes = 0L
+          val involvedDirectories = new ArrayBuffer[String]()
```
JingsongLi, line 141: Should be a Set to do deduplicating?

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/orphan/SparkOrphanFilesClean.scala
```diff
@@ -145,22 +147,31 @@ case class SparkOrphanFilesClean(
             deletedFilesLenInBytes += fileInfo.getLong(2)
             specifiedFileCleaner.accept(deletedPath)
             logInfo(s"Cleaned file: $pathToClean")
+            involvedDirectories.append(deletedPath.getParent.toUri.toString)
             deletedFilesCount += 1
           }
           logInfo(
             s"Total cleaned files: $deletedFilesCount, Total cleaned files len : $deletedFilesLenInBytes")
-          Iterator.single((deletedFilesCount, deletedFilesLenInBytes))
+          Iterator.single((deletedFilesCount, deletedFilesLenInBytes, involvedDirectories))
```
JingsongLi, line 155: distinct directories and clean them with multi concurrent?

paimon-core/src/main/java/org/apache/paimon/operation/OrphanFilesClean.java
```diff
@@ -91,6 +95,10 @@ public abstract class OrphanFilesClean implements Serializable {
     protected final int partitionKeysNum;
     protected final Path location;
 
+    private static final String THREAD_NAME = "ORPHAN-FILES-CLEAN-THREAD-POOL";
+    private static final ThreadPoolExecutor executorService =
```
JingsongLi, line 99: Do not introduce another thread pool, for `LocalOrphanFilesClean`, you should just use executor in it.

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/orphan/SparkOrphanFilesClean.scala
```diff
@@ -145,22 +148,31 @@ case class SparkOrphanFilesClean(
...
       }
+      .cache()
+
+    // clean empty directories
+    val deletedPaths =
+      deleted.flatMap { case (_, _, paths) => paths }.collect().map(new Path(_)).toSet
+    cleanEmptyDirectory(deletedPaths.asJava)
```
JingsongLi, line 163: You should use Spark distributed RDD or DataFrame to clean them.

paimon-core/src/main/java/org/apache/paimon/operation/OrphanFilesClean.java
```diff
@@ -393,4 +401,27 @@ public static long olderThanMillis(@Nullable String olderThan) {
...
     }
+
+    public void cleanEmptyDirectory(Set<Path> deletedPaths) {
+        if (deletedPaths.isEmpty()) {
+            return;
+        }
+
+        randomlyOnlyExecute(executorService, this::tryDeleteEmptyDirectory, deletedPaths);
+
+        for (int level = 0; level < partitionKeysNum; level++) {
```
wwj6591812, line 412: What the `level` used for?

askwang: When the empty bucket directories are cleaned, the parent partition directories should also cleaned if they are empty.

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/orphan/SparkOrphanFilesClean.scala
```diff
@@ -145,22 +149,47 @@ case class SparkOrphanFilesClean(
...
             logInfo(s"Cleaned file: $pathToClean")
+            involvedDirectories.add(deletedPath.getParent.toUri.toString)
             deletedFilesCount += 1
           }
           logInfo(
             s"Total cleaned files: $deletedFilesCount, Total cleaned files len : $deletedFilesLenInBytes")
-          Iterator.single((deletedFilesCount, deletedFilesLenInBytes))
+          Iterator.single((deletedFilesCount, deletedFilesLenInBytes, involvedDirectories.toSet))
       }
+      .cache()
```
JingsongLi, line 159: Can we avoid caching here?

askwang: This dataset need to use twice, currently cannot be removed, do you have a better way?

paimon-core/src/main/java/org/apache/paimon/operation/OrphanFilesClean.java
```diff
@@ -393,4 +400,20 @@ public static long olderThanMillis(@Nullable String olderThan) {
             return parsedTimestampData.getMillisecond();
         }
     }
+
+    /** Try to clean empty partition directories. */
+    protected void tryCleanPartitionDirectory(Set<Path> deletedPaths) {
+        for (int level = 0; level < partitionKeysNum; level++) {
+            deletedPaths.forEach(this::tryDeleteEmptyDirectory);
```
Zouxxyy, line 407: We can terminate the loop early if the deletion fails

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/orphan/SparkOrphanFilesClean.scala
```diff
@@ -145,22 +149,47 @@ case class SparkOrphanFilesClean(
...
+    (finalDeletedDataset, (usedManifestFiles, deleted))
+  }
+
+  private def cleanEmptyDirectory(deletedPaths: Dataset[String]): Unit = {
+    import spark.implicits._
+
+    val partitionDirectory = deletedPaths
+      .filter(_.contains(BUCKET_PATH_PREFIX))
+      .mapPartitions {
+        iter =>
```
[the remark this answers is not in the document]
Zouxxyy: Another way:

```
cleanEmptyBuckets (The task of deleting the last file in the bucket will succeed) ->
Set(involvedPartitions)  -> 
distinct ->  
map delete involvedPartitions
```

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/orphan/SparkOrphanFilesClean.scala
```diff
@@ -145,22 +149,47 @@ case class SparkOrphanFilesClean(
             deletedFilesLenInBytes += fileInfo.getLong(2)
             specifiedFileCleaner.accept(deletedPath)
             logInfo(s"Cleaned file: $pathToClean")
+            involvedDirectories.add(deletedPath.getParent.toUri.toString)
```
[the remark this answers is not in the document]
askwang: added

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/orphan/SparkOrphanFilesClean.scala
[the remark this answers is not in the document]
askwang: The second is a better way after offline discussion. But has a question, distributed involvedPartitions cannot been distinct if we not aggregate or collect it. So cleaning the involvedPartitions after deleting empty bucket directires.

paimon-core/src/main/java/org/apache/paimon/operation/OrphanFilesClean.java
askwang: filter the partition directories if deletion fails

askwang:
@Zouxxyy Very thanks for you review, has adjusted. You can help review again for your free.

paimon-core/src/main/java/org/apache/paimon/operation/LocalOrphanFilesClean.java
```diff
@@ -127,10 +134,30 @@ public CleanOrphanFilesResult clean()
...
+        // clean empty directory
+        if (!dryRun) {
+            cleanEmptyDirectory(deleteFiles);
+        }
+
         return new CleanOrphanFilesResult(
                 deleteFiles.size(), deletedFilesLenInBytes.get(), deleteFiles);
     }
 
+    private void cleanEmptyDirectory(List<Path> deleteFiles) {
```
Zouxxyy, line 146: 1. Extract a common method that can be used by both core, spark and so on. Note: bucket and partitions logic can be processed together, as long as we set level = partitionKeysNum + 1.
2. Let the entire method be executed through randomlyOnlyExecute.

Zouxxyy (commented):
The overall implementation idea is already quite clear. In each task, clean empty dirs of the involved buckets set. 

And remember to clean up any unused code, such as "ORPHAN-FILES-CLEAN-THREAD-POOL," etc.

Zouxxyy (approved):
+1

PR #4827, opened 2025-01-03
[core] Optimize snapshotManager code eliminate duplicate logic.

herefree:
### Purpose
There are some duplicate binarySearch codes in different method.Extract into a common binarySearch method.

Linked issue: close #xxx

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/utils/SnapshotManager.java
+++ b/paimon-core/src/main/java/org/apache/paimon/utils/SnapshotManager.java
@@ -290,15 +290,11 @@ public class SnapshotManager implements Serializable {
             return earliest - 1;
         }
 
-        while (earliest < latest) {
-            long mid = (earliest + latest + 1) / 2;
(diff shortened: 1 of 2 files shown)


paimon-core/src/main/java/org/apache/paimon/utils/SnapshotManager.java
```diff
@@ -434,55 +376,63 @@ private Snapshot changelogOrSnapshot(long snapshotId) {
...
+                return earliest;
             }
-            if (commitWatermark == null) {
-                earliest = mid + 1;
+            earliest++;
+        }
+        return null;
+    }
+
+    private @Nullable Long binarySearch(
```
wwj6591812, line 412: Could you add a ut to test this method?

herefree: fixed

herefree: > Could you add a ut to test this method?

fixed

PR #4834, opened 2025-01-03
[common] A FileIO API to list files iteratively

### LINKED ISSUE #4791: [Feature] A `FileIO` API to list files iteratively
### Search before asking

### Motivation

Currently the `FileIO` interface only supports listing all files / directories under a given path at a time. As a consequence callers of `FileIO`, e.g. `ObjectRefresh`, can only choose to load the entire catalog of files into memory, which may lead to poor performance and OOM.

### Solution

Introduce paged list API like the following:

```java
Pair<FileStatus[], String> listFilesPaged(
        Path path, boolean recursive, long pageSize, @Nullable String continuationToken)
```

This should allow implementations to take advantage of batched list APIs that are commonly seen in object stores, e.g. ListObjectsV2 with continuation token.

### Anything else?

_No response_

smdsbz:
### Purpose

Linked issue: close #4791

Proposing `FileIO#listFilesPaged` and its non-iterative sibling `FileIO#listFiles` .

### Tests

* `FileIOTest#testListFiles`
* ~`FileIOTest#testListFilesPaged`~

### API and Format

Adding `listFilesPaged` method to `FileIO` interface.

### Documentation

As described in the methods' JavaDocs.

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/fs/FileIO.java
+++ b/paimon-common/src/main/java/org/apache/paimon/fs/FileIO.java
@@ -22,11 +22,13 @@ import org.apache.paimon.annotation.Public;
 import org.apache.paimon.catalog.CatalogContext;
 import org.apache.paimon.fs.hadoop.HadoopFileIOLoader;
 import org.apache.paimon.fs.local.LocalFileIO;
+import org.apache.paimon.utils.Pair;
 
 import org.apache.commons.io.IOUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import javax.annotation.Nullable;
 import javax.annotation.concurrent.ThreadSafe;
 
 import java.io.BufferedReader;
@@ -39,6 +41,7 @@ import java.net.URI;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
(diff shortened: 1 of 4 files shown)


paimon-common/src/main/java/org/apache/paimon/fs/FileIO.java
```diff
@@ -104,6 +107,71 @@ public interface FileIO extends Serializable {
...
+     * @param continuationToken If supplied will list files after this token, otherwise list from
+     *     the beginning. You may acquire this token from the return of this method.
+     * @return A page of statuses of the files in the given path and the continuation token of this
+     *     page. The continuation token will be <code>null</code> if the returned page is the last
+     *     page.
+     */
+    default Pair<FileStatus[], String> listFilesPaged(
+            Path path, boolean recursive, long pageSize, @Nullable String continuationToken)
+            throws IOException {
+        FileStatus[] all = listFiles(path, recursive);
```
wwj6591812, line 154: In https://github.com/apache/paimon/issues/4791, it says that :"As a consequence callers of FileIO, e.g. ObjectRefresh, can only choose to load the entire catalog of files into memory, which may lead to poor performance and OOM."

If apply this pr, here call `listFiles`, It still keep all the results of the `listFiles` in memory, and I understand that it cannot solve the OOM problem. What's your opinion?

smdsbz: The purpose of this PR is to agree on the proposed paged-list API, along with a functionally-correct default implementation. I intend to submit tailored implementations of various stores in separate PRs.

wwj6591812: ok

smdsbz:
UT failed due to a bug described in #4849, which is irrelevant to this PR.

paimon-common/src/main/java/org/apache/paimon/fs/FileIO.java
JingsongLi, line 154: Maybe, we can throw UnsupportedException for this method by default.

Because, it looks like this default implementation has side effects.

smdsbz: Solid point, someone may stumble on it and unconsciously firing full-fat `listFiles`es for each page.

I'm erasing the default impl for `listFilesPaged`.

smdsbz: Also adding `supportsListFilesPaged` for testing the availability of `listFilesPaged`.

paimon-common/src/main/java/org/apache/paimon/fs/FileIO.java
```diff
@@ -104,6 +106,62 @@ public interface FileIO extends Serializable {
...
+     * @param recursive if set to <code>true</code> will recursively list files in subdirectories,
+     *     otherwise only files in the current directory will be listed
+     * @param pageSize maximum size of the page
+     * @param continuationToken If supplied will list files after this token, otherwise list from
+     *     the beginning. You may acquire this token from the return of this method.
+     * @return A page of statuses of the files in the given path and the continuation token of this
+     *     page. The continuation token will be <code>null</code> if the returned page is the last
+     *     page.
+     */
+    default Pair<FileStatus[], String> listFilesPaged(
```
JingsongLi, line 159: Do you think a `listFileIterator` is OK for you?

smdsbz: In that case, I'm thinking on reworking the definition of this API to

```java
default Iterator<FileStatus> listFilesIterative(Path path, boolean recursive) {
  // a default implementation that stores all files in the returned Iterator

  // implementors may return Iterators that internally maintain pages
  // the preferred page size is to be determined when FileIO is configure()-ed
}
```

`listFiles` will be preserved as a handy helper to unpack the result of `listFilesIterative`.

The hinting method `supportsListFilesPaged` will be removed. (Maybe propose something like `FileIO#hasFeature` in the future, but that deserves a separate PR.)

What's your take?

paimon-common/src/main/java/org/apache/paimon/fs/FileIO.java
```diff
@@ -104,6 +107,59 @@ public interface FileIO extends Serializable {
...
+        }
+        return new Iterator<FileStatus>() {
+            @Override
+            public boolean hasNext() {
+                return !files.isEmpty();
+            }
+
+            @Override
+            public FileStatus next() {
+                return files.remove();
```
JingsongLi, line 158: Maybe this can be lazily using `Iterator`?
In hasNext or next, we do really `listStatus` for sub directories.

smdsbz: Adding a `FileStatusIterator` interface to allow `hasNext` and `next` throwing `IOException` while lazy listing.

paimon-common/src/main/java/org/apache/paimon/fs/FileStatusIterator.java
```diff
@@ -0,0 +1,49 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.paimon.fs;
+
+import java.io.Closeable;
+import java.io.IOException;
+
+/** An iterator for lazily listing {@link FileStatus}. */
+public interface FileStatusIterator extends Closeable {
```
JingsongLi, line 25: Just add a class like `org.apache.hadoop.fs.RemoteIterator`?

smdsbz: Templating `FileStatusIterator` to `o.a.p.fs.RemoteIterator<E>`

paimon-common/src/main/java/org/apache/paimon/fs/RemoteIterator.java
```diff
@@ -0,0 +1,49 @@
...
+     * @throws IOException - if failed to list entries lazily
+     */
+    E next() throws IOException;
+
+    /**
+     * Closes the iterator and its associated resources.
+     *
+     * @throws IOException - if failed to close the iterator
+     */
+    void close() throws IOException;
```
JingsongLi, line 48: Do we must have this method?

JingsongLi:
+1

PR #4835, opened 2025-01-05
[core] Fix retract behavior for FieldProductAgg

### LINKED ISSUE #4822: [Bug] Incorrect retract behavior for FieldProductAgg
### Search before asking

### Paimon version

master

### Compute Engine

Paimon Core

### Minimal reproduce step

`assertThat(fieldProductAgg.retract(null, 5)).isEqualTo(5);`

If a null ProductAgg field receives retract message, it should return a null value or a 1/retract value.

### Anything else?

_No response_

herefree:
### Purpose
Fix retract behavior for FieldProductAgg.If a null ProductAgg field receives retract message, return a 1/retract value.

Linked issue: close #4822

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/aggregate/FieldProductAgg.java
+++ b/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/aggregate/FieldProductAgg.java
@@ -89,7 +89,7 @@ public class FieldProductAgg extends FieldAggregator {
         Object product;
 
         if (accumulator == null || inputField == null) {
-            product = (accumulator == null ? inputField : accumulator);
+            product = (accumulator == null ? inverse(inputField) : accumulator);
         } else {
             switch (fieldType.getTypeRoot()) {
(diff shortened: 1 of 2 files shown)


paimon-core/src/main/java/org/apache/paimon/mergetree/compact/aggregate/FieldProductAgg.java
```diff
@@ -89,7 +89,7 @@ public Object retract(Object accumulator, Object inputField) {
         Object product;
 
         if (accumulator == null || inputField == null) {
-            product = (accumulator == null ? inputField : accumulator);
+            product = (accumulator == null ? inverse(inputField) : accumulator);
```
xiangyuf, line 92: I prefer to return null here. Return 0 will lead to following value all equals to 0.

herefree: ok, I'll close it.

PR #4837, opened 2025-01-06
[core] remove nullable check for record-level.time-field

### LINKED ISSUE #4821: [BUG] Retract message for Aggregate table will lead to NPE for record ttl compaction
### Search before asking

### Motivation

For an aggregation Paimon PK table, if it receives a retract message at first, it will inevitably generate a +I message with  a negative value for SumAggField or null value for other agg fields. If this table has enabled record level ttl, this will throw an IllegalArgumentException during compaction if the ttl time field is null.

The first react message may generated from the out-of-order issue during now-primary key hashing in streaming computation.

image

### Solution

Ignore first retract message for aggregate table if its record level TTL is enabled. In most cases, the first retract message is meaningless for an aggregation table.

### Anything else?

_No response_

xiangyuf:
### Purpose

Linked issue: close #4821 

We should not enforce not nullable check for record-level.time-field for two reasons:

1, When creating paimon tables using  Spark SQL or Hive SQL, users cannot specify certain fields are not nullable;

2, In certain cases like #4821, Paimon writer will inevitably write null time field into data files. Not nullable datatype will cause parquet reader fail with following msgs:
image

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
+++ b/paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
@@ -203,12 +203,6 @@ public class SchemaValidation {
                                 "The record level time field type should be one of INT, BIGINT, or TIMESTAMP, but field type is %s.",
                                 dataType));
             }
-            if (dataType.isNullable()) {
-                throw new IllegalArgumentException(
-                        String.format(
-                                "Time field %s for record-level expire should be not null.",
-                                recordLevelTimeField));
-            }
         }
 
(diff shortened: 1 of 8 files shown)


JingsongLi (commented):
Can you just add some tests to nullable field?

xiangyuf:
> Can you just add some tests to nullable field?

@JingsongLi Added tests for this change. More tests for null value record-level.time-field are included in this pr: https://github.com/apache/paimon/pull/4839. I prefer to split these changes into two commits.

JingsongLi (approved):
+1

PR #4838, opened 2025-01-06
[Bug] overwrite is true when commit change log

jerry-024:
### Purpose
when commit change log set overwrite file as true


JingsongLi (approved):
+1

PR #4839, opened 2025-01-06
[core] support null value for record-level.time-field

### LINKED ISSUE #4821: [BUG] Retract message for Aggregate table will lead to NPE for record ttl compaction

xiangyuf:
### Purpose

Return true for RecordLevelExpire reader filter when time field is null.

Linked issue: close #4821 

### Tests

RecordLevelExpireWithNullFieldTest

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/io/RecordLevelExpire.java
+++ b/paimon-core/src/main/java/org/apache/paimon/io/RecordLevelExpire.java
@@ -33,15 +33,14 @@ import org.apache.paimon.types.TimestampType;
 import javax.annotation.Nullable;
 
 import java.time.Duration;
+import java.util.Optional;
 import java.util.function.Function;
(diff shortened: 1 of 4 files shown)


wwj6591812:
+1

JingsongLi (approved):
+1

PR #4841, opened 2025-01-06
[Bug] use snapshot when SnapshotReaderImpl.toChangesPlan

jerry-024:
### Purpose

Fix `incremental-between` when then snapshot id reference snapshot file has deleted bug


JingsongLi (approved):
+1

PR #4846, opened 2025-01-06
 [flink][action] add '`' to the fields of merge into action to avoid exceptions when the field name is an SQL keyword.

### LINKED ISSUE #4845: [Feature] MergeIntoAction supports using fields that are the same as SQL keywords.
### Search before asking

### Motivation

Some tables may use SQL keywords as table fields. When executing `MergeInto`, we need to automatically add  ` for these fields.

### Solution

_No response_

### Anything else?

_No response_

liming30:
### Purpose

Linked issue: close #4845 

[flink][action] add '`' to the fields of merge into action to avoid exceptions when the field name is an SQL keyword.

### Tests

- `org.apache.paimon.flink.action.MergeIntoActionITCase#testSqlWithKeywordCase`

### API and Format

No.
### Documentation

No.


JingsongLi:
+1

PR #4848, opened 2025-01-06
[core] Introduce nested type cast to string

Zouxxyy:
### Purpose

Introduce nested type cast to string. For example, it can be used for partition cast in system table

### DIFF
--- a/paimon-common/src/test/java/org/apache/paimon/casting/CastExecutorTest.java
+++ b/paimon-common/src/test/java/org/apache/paimon/casting/CastExecutorTest.java
@@ -19,17 +19,24 @@
 package org.apache.paimon.casting;
 
 import org.apache.paimon.data.BinaryString;
+import org.apache.paimon.data.GenericArray;
+import org.apache.paimon.data.GenericMap;
+import org.apache.paimon.data.GenericRow;
 import org.apache.paimon.data.Timestamp;
+import org.apache.paimon.types.ArrayType;
 import org.apache.paimon.types.BigIntType;
 import org.apache.paimon.types.BinaryType;
 import org.apache.paimon.types.BooleanType;
 import org.apache.paimon.types.CharType;
+import org.apache.paimon.types.DataTypes;
 import org.apache.paimon.types.DateType;
 import org.apache.paimon.types.DecimalType;
 import org.apache.paimon.types.DoubleType;
 import org.apache.paimon.types.FloatType;
 import org.apache.paimon.types.IntType;
 import org.apache.paimon.types.LocalZonedTimestampType;
+import org.apache.paimon.types.MapType;
+import org.apache.paimon.types.RowType;
 import org.apache.paimon.types.SmallIntType;
 import org.apache.paimon.types.TimeType;
 import org.apache.paimon.types.TimestampType;
@@ -41,6 +48,8 @@ import org.apache.paimon.utils.DecimalUtils;
 
 import org.junit.jupiter.api.Test;
 
+import java.util.HashMap;
+import java.util.Map;
 import java.util.TimeZone;
 
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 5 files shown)


paimon-common/src/test/java/org/apache/paimon/casting/CastExecutorTest.java
```diff
@@ -681,6 +690,65 @@ public void testTimeToTimestamp() {
...
+        Map<Object, Object> javaMap = new HashMap<>();
+        javaMap.put(1, BinaryString.fromString("i"));
+        javaMap.put(2, BinaryString.fromString("miss"));
+        javaMap.put(3, BinaryString.fromString("you"));
+        javaMap.put(4, null);
+        GenericMap genericMap = new GenericMap(javaMap);
+        compareCastResult(
+                CastExecutors.resolve(mapType, DataTypes.STRING()),
+                genericMap,
+                BinaryString.fromString("{1 -> i, 2 -> miss, 3 -> you, 4 -> null}"));
```
JingsongLi, line 715: Is this from Spark SQL?

paimon-common/src/test/java/org/apache/paimon/casting/CastExecutorTest.java
```diff
@@ -681,6 +690,65 @@ public void testTimeToTimestamp() {
...
+                        1,
+                        GenericRow.of(
+                                DateTimeUtils.parseDate("2025-01-06"),
+                                new GenericMap(javaMap),
+                                null));
+
+        compareCastResult(
+                CastExecutors.resolve(rowType, DataTypes.STRING()),
+                row,
+                BinaryString.fromString("{1, {2025-01-06, {1 -> [1, null, 2]}, null}}"));
```
JingsongLi, line 749: Ditto, (why not using `(abc, def)`)

paimon-common/src/test/java/org/apache/paimon/casting/CastExecutorTest.java
Zouxxyy: yes, copy from spark

JingsongLi: No for show, you should take a look to Spark SQL.

Zouxxyy: This is the behavior of Spark SQL's cast.

```
select cast (struct(1, 'a') as string), cast (map(1, 'a') as string), cast (array(1,2,3) as string)
+----------------------------+-------------------------+------------------------------+
|CAST(struct(1, a) AS STRING)|CAST(map(1, a) AS STRING)|CAST(array(1, 2, 3) AS STRING)|
+----------------------------+-------------------------+------------------------------+
|                      {1, a}|                 {1 -> a}|                     [1, 2, 3]|
+----------------------------+-------------------------+------------------------------+
```

JingsongLi: Can you take a look to other engines? For example, trino and snowflake.

Zouxxyy: Here are some examples from other engines, I did not find examples of Snowflake.

Flink
```
select cast (row(1, 'a') as string), cast(map[1, 'a'] as string), cast (array[1,2,3] as string);

-- | -- | --
(1,a) | {1=a} | [1, 2, 3]
```

Spark
```
spark-sql> select cast (struct(1, 'a') as string), cast (map(1, 'a') as string), cast (array(1,2,3) as string);

+----------------------------+-------------------------+------------------------------+
|CAST(struct(1, a) AS STRING)|CAST(map(1, a) AS STRING)|CAST(array(1, 2, 3) AS STRING)|
+----------------------------+-------------------------+------------------------------+
|                      {1, a}|                 {1 -> a}|                     [1, 2, 3]|
+----------------------------+-------------------------+------------------------------+
```

Trino (can't cast, just show)
```
trino> select row(1, 'a'), map(ARRAY[1], ARRAY['a']), array[1,2,3];

 _col0  | _col1 |   _col2   
--------+-------+-----------
 {1, a} | {1=a} | [1, 2, 3]
```

StarRocks (can't cast, just show)
```
select row(1, 'a'), map(1, 'a'), [1, 2, 3];

{"col1":1,"col2":"a"} | {1:"a"} | [1,2,3]
```

JingsongLi:
+1

PR #4851, opened 2025-01-07
[flink] Fix the refresh executor not work after reopen

Aitozi:
### Purpose

If we use dynamic partition loader, it may reopen the LookupTable after partition changed. After reopen, the refresh executor is shutdown, we should recreate the executor in the open phase.

image

Linked issue: close #xxx


wwj6591812:
good.

+1

JingsongLi (approved):
+1

PR #4852, opened 2025-01-07
[core] Use cast to string to show partition in system table

Zouxxyy:
### Purpose

```sql
CREATE TABLE T (a INT, p1 DATE, p2 INT) PARTITIONED BY (p1, p2);
INSERT INTO T VALUES(1, cast('2024-10-10' as date), 1);
SELECT partition FROM `T$partitions`;
```

before
```
+----------+
| partition|
+----------+
|[20006, 1]|
+----------+
```

after

```
+---------------+
|      partition|
+---------------+
|{2024-10-10, 1}|
+---------------+
```


JingsongLi:
+1

PR #4856, opened 2025-01-07
[log] logging specific join keys and results  for lookup join

LsomeYeah:
### Purpose

Linked issue: close #xxx

Add logs about join keys and join results when performing lookup join in `DEBUG` mode.


yuzelin (commented):
+1

PR #4860, opened 2025-01-08
[core]Fix drop table in Hive and jdbc catalog when the file path does not exists.

herefree:
### Purpose
Fix drop table in Hive and jdbc catalog when the file path does not exists.

Linked issue: close #4853

PR #4863, opened 2025-01-08
[hive] Fix no longer possible to drop the hms table issue when table …

### LINKED ISSUE #4853: [Bug] When using the Hive catalog, after deleting the table directory, it is no longer possible to drop the table
### Search before asking

### Paimon version

master

### Compute Engine

-

### Minimal reproduce step

1. using hive catalog
2. create table
3. delete table in file system
4. drop table (it is no longer possible to drop the table in hms )

-

### Anything else?

_No response_

XiaoHongbo-Hope:
…not in fs (#4853)

Purpose Fix no longer possible to drop the hms table issue when table not in fs

Linked issue: close #4853

### DIFF
--- a/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
+++ b/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
@@ -694,7 +694,7 @@ public class HiveCatalog extends AbstractCatalog {
             return tableSchemaInFileSystem(
                             getTableLocation(identifier, table),
                             identifier.getBranchNameOrDefault())
-                    .orElseThrow(() -> new TableNotExistException(identifier));
+                    .orElseThrow(() -> new HmsTableNotExistInFsException(identifier));
         }
 
         if (!formatTableDisabled()) {
@@ -705,7 +705,7 @@ public class HiveCatalog extends AbstractCatalog {
             }
         }
 
-        throw new TableNotExistException(identifier);
+        throw new HmsTableNotExistInFsException(identifier);
     }
 
     @Override
@@ -886,6 +886,31 @@ public class HiveCatalog extends AbstractCatalog {
                         tableOptions.get(HIVE_EXTERNAL_TABLE_PROP.toUpperCase()));
         return CatalogTableType.EXTERNAL.equals(tableType)
                 || "TRUE".equalsIgnoreCase(externalPropValue);
+    }
+
+    @Override
+    public void dropTable(Identifier identifier, boolean ignoreIfNotExists)
+            throws TableNotExistException {
+        checkNotBranch(identifier, "dropTable");
+        checkNotSystemTable(identifier, "dropTable");
+
+        try {
+            getTable(identifier);
+        } catch (TableNotExistException e) {
+            if (e instanceof HmsTableNotExistInFsException) {
+                LOG.warn(
+                        "Table {}.{} exists in Hive metastore, but not exist in file system. will try to drop it in Hive metastore only.",
+                        identifier.getDatabaseName(),
+                        identifier.getTableName());
+                dropHmsTableQuietly(identifier);
+            }
+            if (ignoreIfNotExists) {
+                return;
+            }
+            throw new TableNotExistException(identifier);
+        }
+
+        dropTableImpl(identifier);
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


LinMingQiang:
Repeated :  https://github.com/apache/paimon/pull/4860

paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
```diff
@@ -272,10 +272,14 @@ public void dropTable(Identifier identifier, boolean ignoreIfNotExists)
         try {
             getTable(identifier);
         } catch (TableNotExistException e) {
+            if (!tableExistsInFileSystem(getTableLocation(identifier), DEFAULT_MAIN_BRANCH)) {
```
Zouxxyy, line 275: As far as I know, before Spark calls `catalog.dropTable`, it actually invokes `catalog.getTable`. It has already been skipped there.

paimon-spark/paimon-spark-ut/src/test/scala/org/apache/paimon/spark/sql/DDLWithHiveCatalogTestBase.scala
```diff
@@ -621,6 +621,29 @@ abstract class DDLWithHiveCatalogTestBase extends PaimonHiveTestBase {
     }
   }
 
+  test("Paimon DDL with hive catalog: drop table which location has been deleted") {
```
Zouxxyy, line 624: For this case, we need to test both enable and disable the catalog cache

paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
```diff
@@ -272,10 +272,14 @@ public void dropTable(Identifier identifier, boolean ignoreIfNotExists)
...
+                dropTableImpl(identifier);
```
xuzifu666, line 276: Maybe we can make dropTableImpl before getTable in AbstractCatalog? and dropTableImpl in HiveCatalog do the change: 1. client.getTable 2. if 1 not exception out then do client.dropTable.

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
```diff
@@ -668,6 +668,10 @@ public org.apache.paimon.table.Table getDataOrFormatTable(Identifier identifier)
                                     identifier),
                             metastoreClientFactory(identifier).orElse(null)));
         } catch (TableNotExistException ignore) {
+            // do not drop format table, since non paimon table throws TableNotExistException too
+            if (isPaimonTable(table)) {
```
xuzifu666, line 672: Why not put it to finally?

paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
XiaoHongbo-Hope: > Maybe we can move dropTableImpl before getTable in AbstractCatalog? and dropTableImpl in HiveCatalog do the change: 1. client.getTable 2. if 1 not exception out then do client.dropTable.

xxxImpl method does the function itself maybe better?

Zouxxyy:
I'm OK with this modification. The core reason is that before dropping the table, the compute engine will call table exists -> `catalog.getTable()` to check if the table exists. Currently, our `catalog.getTable()` API returns whether the table exists in the file system (even if it exists in HMS). Could @JingsongLi  take another look at this?

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
```diff
@@ -1638,4 +1642,20 @@ public int getBatchGetTableSize() {
...
         }
     }
+
+    private void repairTableQuietly(Identifier identifier) {
+        LOG.warn(
+                "Paimon Table {}.{} exists in HMS but not in the file system during get table; dropping it from HMS.",
+                identifier.getDatabaseName(),
+                identifier.getTableName());
+        try {
+            dropTableImpl(identifier);
```
JingsongLi, line 1652: Drop table in getTable? This behavior is too strange.

JingsongLi:
I am -1 for this. You can enrich exception message. And user can just use drop table again?

adrian-wang:
Or, just recover and return a very basic dummy table when getTable, if we find the table is in hms and not on FileSystem?

XiaoHongbo-Hope:
> I am -1 for this. You can enrich exception message. And user can just use drop table again?
Thanks

> Or, just recover and return a very basic dummy table when getTable, if we find the table is in hms and not on FileSystem?

Thanks

XiaoHongbo-Hope:
> I am -1 for this. You can enrich exception message. And user can just use drop table again?

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
XiaoHongbo-Hope: Ok, thanks

JingsongLi:
No change in this PR, I close this.

XiaoHongbo-Hope:
> No change in this PR, I close this.

Ok

PR #4866, opened 2025-01-08
[core] Introduce batch partition methods in catalog

JingsongLi:
### Purpose

At present, we have scattered many methods in `MetastoreClient`, and we should provide unified support for all of them on the Catalog interface.

1. Only introduce batch methods to reduce complex.
2. Tag to partition table should be handled too.

This is the first PR for removing `MetastoreClient`. This PR just introduce methods in Catalog.


yuzelin (approved):
+1

PR #4869, opened 2025-01-08
[core] Fix that incremental-to-auto-tag return wrong result if snapshots are expired

yuzelin:
### Purpose

Linked issue: close #xxx


JingsongLi (approved):
+1

PR #4871, opened 2025-01-09
[core] Fix remove orphan files with data file path directory

Zouxxyy:
### Purpose

Fix remove orphan files with data file path directory

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/OrphanFilesClean.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/OrphanFilesClean.java
@@ -33,6 +33,7 @@ import org.apache.paimon.schema.SchemaManager;
 import org.apache.paimon.table.FileStoreTable;
 import org.apache.paimon.utils.BranchManager;
 import org.apache.paimon.utils.DateTimeUtils;
+import org.apache.paimon.utils.FileStorePathFactory;
 import org.apache.paimon.utils.Pair;
 import org.apache.paimon.utils.Preconditions;
 import org.apache.paimon.utils.SerializableConsumer;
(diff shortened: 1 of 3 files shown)


paimon-core/src/main/java/org/apache/paimon/utils/FileStorePathFactory.java
```diff
@@ -37,6 +37,17 @@
...
+    public static final String MANIFEST_PREFIX = "manifest-";
+    public static final String MANIFEST_LIST_PREFIX = "manifest-list-";
+    public static final String INDEX_MANIFEST_PREFIX = "index-manifest-";
+
+    public static final String INDEX_PATH = "index";
+    public static final String INDEX_PREFIX = "index-";
+
+    public static final String STATISTICS_PATH = "statistics";
+    public static final String STATISTICS_PREFIX = "stat-";
+
```
wwj6591812, line 50: private

paimon-core/src/main/java/org/apache/paimon/operation/OrphanFilesClean.java
```diff
@@ -252,12 +253,14 @@ protected void collectWithoutDataFileWithManifestFlag(
 
     /** List directories that contains data files and manifest files. */
     protected List<Path> listPaimonFileDirs() {
+        FileStorePathFactory pathFactory = table.store().pathFactory();
```
wwj6591812, line 256: Fix? Could you provide a detailed description of the bug?

paimon-core/src/main/java/org/apache/paimon/utils/FileStorePathFactory.java
Zouxxyy: I think both public and private are fine, and the BUCKET_PATH_PREFIX is also public.

JingsongLi (approved):
+1

wwj6591812:
+1

PR #4874, opened 2025-01-09
[core] Introduce catalog loader in Catalog to replace MetastoreClient

JingsongLi:
### Purpose

Next step of https://github.com/apache/incubator-paimon/pull/4866

At present, we have scattered many methods in `MetastoreClient`, and we should provide unified support for all of them on the Catalog interface.


paimon-core/src/main/java/org/apache/paimon/catalog/CatalogLoader.java
```diff
@@ -18,9 +18,12 @@
 
 package org.apache.paimon.catalog;
 
+import org.apache.paimon.annotation.Public;
+
 import java.io.Serializable;
 
 /** Loader for creating a {@link Catalog}. */
+@Public
```
yuzelin, line 26: Add `@since`

yuzelin (approved):
+1

PR #4875, opened 2025-01-09
[core] Add UT for expire partitions with hms config metastore.partitioned-table 

xuzifu666:
### Purpose

Linked issue: close #xxx
Currently ExpirePartitionProcedure would init MestoreClient whether metastore.partitioned-table true or false, this would cause hms sync error when metastore.partitioned-table=false. This pr is aim to fix it.

### DIFF
--- a/paimon-spark/paimon-spark-ut/src/test/scala/org/apache/paimon/spark/sql/DDLWithHiveCatalogTestBase.scala
+++ b/paimon-spark/paimon-spark-ut/src/test/scala/org/apache/paimon/spark/sql/DDLWithHiveCatalogTestBase.scala
@@ -163,11 +163,11 @@ abstract class DDLWithHiveCatalogTestBase extends PaimonHiveTestBase {
                 spark.sql("insert into paimon_db.paimon_tbl select '2', 'n', '9999-11-01'")
 
                 spark.sql(
-                  "CALL paimon.sys.expire_partitions(table => 'paimon_db.paimon_tbl', expiration_time => '1 d', timestamp_formatter => 'yyyy-MM-dd')")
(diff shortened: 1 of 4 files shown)


LinMingQiang:
+1

JingsongLi:
Thanks @xuzifu666 for the contribution!

Can you rebase the master?

xuzifu666:
@JingsongLi Had test after rebase, the code had fix the bug, and ut can run very well.   so the pr spark/flink procedure not need any more. At the same time I found not contains ut case like the pr, I thought maybe need the ut case in the pr.

paimon-spark/paimon-spark-ut/src/test/scala/org/apache/paimon/spark/sql/DDLWithHiveCatalogTestBase.scala
```diff
@@ -134,6 +134,69 @@ abstract class DDLWithHiveCatalogTestBase extends PaimonHiveTestBase {
     }
   }
 
+  test(
+    "Paimon partition expire test with hive catalog: expire partition for paimon table sparkCatalogName") {
+    Seq(paimonHiveCatalogName).foreach {
+      catalogName =>
+        spark.sql(s"USE $catalogName")
```
Zouxxyy, line 141: We should focus on what the case itself requires, without introducing other elements. For example, if we are testing a catalogName, there is no need for a foreach, nor do we need variables like comment, location, etc.

xuzifu666: Done

JingsongLi:
+1

PR #4876, opened 2025-01-09
[core] Support parsing row type json without field id

Zouxxyy:
### Purpose

As part of #4471, in order to support variant shredding, we need a way to indicate which fields require shredding. To reuse the code, we utilize the existing RowType JSON parser. 

E.g. If we want to shred field v with age int and city string, we can create table like this:

```sql
CREATE TABLE T (v VARIANT)
TBLPROPERTIES (
'fields.v.shredding-schema' = '{"type":"ROW","fields":[{"name":"age","type":"INT"},{"name":"city","type":"STRING"}]}'
)
```

To reduce user conf, support parsing row type json without field id

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/types/DataTypeJsonParser.java
+++ b/paimon-common/src/main/java/org/apache/paimon/types/DataTypeJsonParser.java
@@ -26,8 +26,11 @@ import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+
+import static org.apache.paimon.utils.Preconditions.checkState;
 
 /**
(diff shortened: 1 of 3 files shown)


paimon-common/src/main/java/org/apache/paimon/types/DataTypeJsonParser.java
```diff
@@ -66,8 +73,10 @@ public static DataType parseDataType(JsonNode json) {
                 JsonNode fieldArray = json.get("fields");
                 Iterator<JsonNode> iterator = fieldArray.elements();
                 List<DataField> fields = new ArrayList<>(fieldArray.size());
+                int id = 0;
                 while (iterator.hasNext()) {
-                    fields.add(parseDataField(iterator.next()));
+                    fields.add(parseDataField(iterator.next(), id));
```
JingsongLi, line 78: Not work for nested types?

Zouxxyy: Could you provide an example, not sure which nested types

PR #4880, opened 2025-01-10
[rest] Supports global system tables

JingsongLi:
### Purpose

Refactor `AllTableOptionsTable` to let RESTCatalog supports global system tables.


jerry-024:
LGTM

PR #4881, opened 2025-01-10
[e2e] upgrade the test-flink-sql-connector-kafka version

yangjf2019:
### Purpose

Upgrade the  test-flink-sql-connector-kafka version.


yangjf2019:
Now that flink 1.20 has been released, we should consider upgrading.

PR #4882, opened 2025-01-10
[test] remove meaningless tests in PartialUpdateMergeFunctionTest

LsomeYeah:
### Purpose

Linked issue: close #xxx

As shown below, the same field is projected twice when constructing test projection in `testAdjustProjectionRepeatProject()` and `testMultiSequenceFieldsAdjustProjectionRepeatProject()`.
```
// the field 'f1' is projected twice
int[][] projection = new int[][] {{1}, {1}, {3}, {7}};
```
 Those tests are unreasonable and the reason is as follows:

- Even if selecting two same fields in sql, the pushdown projection will only use one element to describe it, `so int[][] projection = new int[][] {{1}, {1}, {3}, {7}}` is impossible in practice.
- May cause wrong test results.


JingsongLi:
+1

PR #4884, opened 2025-01-11
[core] Supports only execute specified actions during the writing process.

LinMingQiang:
### Purpose

There are several operations during data writing such as `partition-expire`,`snapshot-expire`,`create-tag`...
Currently, these actions are either executed together or not executed at all in the framework.

The purpose of this PR is to supports skipping specified actions during writing.

For example :  set `'write.skip-actions'='partition-expire'` , the partition expiration action will be skipped during data writing, but `snapshot-expire` / `auto create tag` / `full compact` will still execute normally.

Linked issue: close #xxx

### DIFF
--- a/docs/layouts/shortcodes/generated/core_configuration.html
+++ b/docs/layouts/shortcodes/generated/core_configuration.html
@@ -1021,6 +1021,12 @@ If the data size allocated for the sorting task is uneven,which may lead to perf
             <td>Type of the table.<br /><br />Possible values:<ul><li>"table": Normal Paimon table.</li><li>"format-table": A file format table refers to a directory that contains multiple files of the same format.</li><li>"materialized-table": A materialized table combines normal Paimon table and materialized SQL.</li><li>"object-table": A object table combines normal Paimon table and object location.</li></ul></td>
         </tr>
         <tr>
+            <td><h5>write-actions</h5></td>
+            <td style="word-wrap: break-word;">"all"</td>
+            <td>String</td>
+            <td>This parameter is used to specify which actions will be performed during the writing process. This parameter is effective only when write-only is false and has no effect on action or procedure.<br />1. 'all': By default, all actions will be performed.<br />2. 'partition-expire': Perform partition expiration action.<br />3. 'snapshot-expire': Perform snapshot expiration action.<br />4. 'tag-automatic-creation': Perform automatic creation tag action.<br />5. 'full-compact': Perform full compaction action.<br />6. 'minor-compact': Perform minor compaction action.<br />Both can be configured at the same time: 'partition-expire,snapshot-expire,tag-automatic-creation', if you want to skip all actions you can set this to ''.</td>
+        </tr>
+        <tr>
             <td><h5>write-buffer-for-append</h5></td>
             <td style="word-wrap: break-word;">false</td>
             <td>Boolean</td>
--- a/docs/layouts/shortcodes/generated/core_configuration.html
+++ b/docs/layouts/shortcodes/generated/core_configuration.html
@@ -1024,7 +1024,7 @@ If the data size allocated for the sorting task is uneven,which may lead to perf
             <td><h5>write-actions</h5></td>
             <td style="word-wrap: break-word;">"all"</td>
             <td>String</td>
-            <td>This parameter is used to specify which actions will be performed during the writing process. This parameter is effective only when write-only is false and has no effect on action or procedure.<br />1. 'all': By default, all actions will be performed.<br />2. 'partition-expire': Perform partition expiration action.<br />3. 'snapshot-expire': Perform snapshot expiration action.<br />4. 'tag-automatic-creation': Perform automatic creation tag action.<br />5. 'full-compact': Perform full compaction action.<br />6. 'minor-compact': Perform minor compaction action.<br />Both can be configured at the same time: 'partition-expire,snapshot-expire,tag-automatic-creation', if you want to skip all actions you can set this to ''.</td>
+            <td>This parameter is used to specify which actions will be performed during the writing process. This parameter is effective only when write-only is false and has no effect on action or procedure.<br />1. 'all': By default, all actions will be performed.<br />2. 'partition-expire': Perform partition expiration action.<br />3. 'snapshot-expire': Perform snapshot expiration action.<br />4. 'tag-automatic-creation': Perform automatic creation tag action.<br />5. 'full-compact': Perform full compaction action.<br />Both can be configured at the same time: 'partition-expire,snapshot-expire,tag-automatic-creation', if you want to skip all actions you can set this to ''.</td>
         </tr>
         <tr>
             <td><h5>write-buffer-for-append</h5></td>
--- a/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
@@ -465,8 +465,6 @@ public class CoreOptions implements Serializable {
                                     .linebreak()
                                     .text("5. 'full-compact': Perform full compaction action.")
                                     .linebreak()
-                                    .text("6. 'minor-compact': Perform minor compaction action.")
-                                    .linebreak()
                                     .text(
                                             "Both can be configured at the same time: 'partition-expire,"
                                                     + "snapshot-expire,tag-automatic-creation', "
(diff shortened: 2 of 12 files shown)


paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -442,6 +443,25 @@ public class CoreOptions implements Serializable {
                             "If set to true, compactions and snapshot expiration will be skipped. "
                                     + "This option is used along with dedicated compact jobs.");
 
+    public static final ConfigOption<String> WRITE_SKIP_ACTIONS =
+            key("write.skip-actions")
+                    .stringType()
```
wwj6591812, line 448: enumType?

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -2216,6 +2236,27 @@ public boolean writeOnly() {
         return options.get(WRITE_ONLY);
     }
 
+    public HashSet<WriteAction> writeSkippingActions() {
```
wwj6591812, line 2239: public Set<WriteAction> writeSkippingActions(

wwj6591812:
Thanks for @LinMingQiang prepare this pr.
This is a good job, I recently received a similar request from some Paimon user in my company.

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
LinMingQiang: This cannot be an enumType because this is a multi-option.

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -2216,6 +2238,27 @@ public boolean writeOnly() {
         return options.get(WRITE_ONLY);
     }
 
+    public Set<WriteAction> writeSkippingActions() {
+        String str = options.get(WRITE_SKIP_ACTIONS);
```
xuzifu666, line 2242: Here no default value，check isPresent？

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -3163,4 +3206,26 @@ public enum MaterializedTableRefreshStatus {
         ACTIVATED,
         SUSPENDED
     }
+
+    /** Actions performed during table writing. */
+    public enum WriteAction {
+
+        // Actions during commit.
+        PARTITION_EXPIRE("partition-expire"),
```
xuzifu666, line 3214: Maybe we can add a option for all actions?

xuzifu666 (commented):
Good job and left some comments.

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
LinMingQiang: Here is just an enumeration of all action types

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
LinMingQiang: done

xuzifu666:
+1

docs/layouts/shortcodes/generated/core_configuration.html
```diff
@@ -1050,6 +1050,12 @@
             <td>Integer</td>
             <td>Write batch size for any file format if it supports.</td>
         </tr>
+        <tr>
+            <td><h5>write.skip-actions</h5></td>
+            <td style="word-wrap: break-word;">(none)</td>
+            <td>String</td>
+            <td>This parameter only works when write-only is false., You can specify which actions to skip during the write process.<br />1. 'partition-expire': Skip the action of partition expiration.<br />2. 'snapshot-expire': Skip the action of snapshot expiration.<br />3. 'create-tag': Skip automatic tag creation.<br />Both can be configured at the same time: 'partition-expire,snapshot-expire,create-tag'.</td>
```
Zouxxyy, line 1057: I prefer to add a separate enable conf for each such post-commit behavior, with the default set to true.

LinMingQiang: Thanks for review, If the parameters are separated, will it be more cumbersome to use ? For my needs, I need to skip `partition-expire` & `snapshot-expire` & `create tag` but retain `compact`, so I may need to configure three parameters to meet my needs.

JingsongLi: Maybe you can add a `write-actions`? Default is 'all'. The values can be:
`snapshot-expire`, `partition-expire`, `compaction`, `tag-automatic-creation`.

LinMingQiang: Does this `compact` mean `full` or `minor` or both, or should it be separated into two actions `full_compact` and `minor_compact`?

JingsongLi: maybe full or minor should be separate option.

LinMingQiang: Ok, I will fix it.

LinMingQiang:
@JingsongLi PTAL, thanks!

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -3226,4 +3311,31 @@ public String toString() {
...
+        // All write actions will be performed, this is the default behavior.
+        ALL("all"),
+
+        // Actions during commit.
+        PARTITION_EXPIRE("partition-expire"),
+        SNAPSHOT_EXPIRE("snapshot-expire"),
+        TAG_AUTOMATIC_CREATION("tag-automatic-creation"),
+
+        // Actions during writing.
+        MINOR_COMPACT("minor-compact"),
```
JingsongLi, line 3327: It is strange here, in streaming mode, it is hard to say just doing minor compact.

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -2241,6 +2271,61 @@ public boolean writeOnly() {
...
+                .map(action -> WriteAction.valueOf(action.toUpperCase().replace('-', '_')))
+                .collect(Collectors.toCollection(HashSet::new));
+    }
+
+    public boolean doPartitionExpireAction(Set<WriteAction> doWriteActions) {
+        return doAllWriteActions(doWriteActions)
+                || doWriteActions.contains(WriteAction.PARTITION_EXPIRE);
+    }
+
+    public boolean doSnapshotExpireAction(Set<WriteAction> doWriteActions) {
```
JingsongLi, line 2295: All these methods can be in `WriteAction`, they are just static methods.

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
LinMingQiang: We can only keep `full-compact` and `minor-compact` action can be removed.

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -3227,6 +3271,56 @@ public String toString() {
...
+        // All write actions will be performed, this is the default behavior.
+        ALL("all"),
+
+        // Actions during commit.
+        PARTITION_EXPIRE("partition-expire"),
+        SNAPSHOT_EXPIRE("snapshot-expire"),
+        TAG_AUTOMATIC_CREATION("tag-automatic-creation"),
+
+        // Actions during writing.
+        FULL_COMPACT("full-compact");
```
JingsongLi, line 3286: I cannot get why you want to add this. Full compact should be separate option.

LinMingQiang: Should I add parameter `write-action.compact.enable` separately to control whether to execute compaction?
In my original conception, `compact` is one of the `write-actions` types like the others.

baiyangtx:
Do you intend to control the behavior of expire in compact action? 

Based on our experience, the expire behavior is quite heavy and may cause the checkpoint to timeout during compact commit. 

We expect to use separate and independent compact and expire job to handle these things.

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
baiyangtx: Is it more reasonable to directly use "compact"? 

From the implementation code, this parameter only controls the compact behavior, while whether to perform a full compact or a minor compact is controlled by other parameters.

PR #4888, opened 2025-01-13
[core] Introduce conversion from parquet type to paimon type

Zouxxyy:
### Purpose

Subtask of #4471, to support shredding, we need a way to convert from Parquet type to Paimon type.


JingsongLi (approved):
+1

PR #4892, opened 2025-01-13
[flink] Use off heap buffer in memory segment to write data

yangjf2019:
### Purpose

Now, Flink have Added get off heap buffer in memory segment.

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/memory/MemorySegmentAllocator.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/memory/MemorySegmentAllocator.java
@@ -62,9 +62,8 @@ public class MemorySegmentAllocator {
             checkNotNull(segment, "Allocate null segment from memory manager for paimon.");
             checkArgument(segment.isOffHeap(), "Segment is not off heap from memory manager.");
             allocatedSegments.add(segment);
-            // TODO Use getOffHeapBuffer in MemorySegment after
-            // https://issues.apache.org/jira/browse/FLINK-32213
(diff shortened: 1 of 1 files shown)


yangjf2019:
Here is [issues](https://issues.apache.org/jira/browse/FLINK-32213).Hi, @JingsongLi  can we use it to manage the writer buffer better?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/memory/MemorySegmentAllocator.java
```diff
@@ -62,9 +62,8 @@ public MemorySegment allocate() {
             checkNotNull(segment, "Allocate null segment from memory manager for paimon.");
             checkArgument(segment.isOffHeap(), "Segment is not off heap from memory manager.");
             allocatedSegments.add(segment);
-            // TODO Use getOffHeapBuffer in MemorySegment after
-            // https://issues.apache.org/jira/browse/FLINK-32213
-            return MemorySegment.wrapOffHeapMemory((ByteBuffer) offHeapBufferField.get(segment));
+            ByteBuffer offHeapBuffer = segment.getOffHeapBuffer();
```
JingsongLi, line 65: Flink 1.15 and 1.16 and 1.17 will be break by this PR.

JingsongLi:
Please consider Flink old versions.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/memory/MemorySegmentAllocator.java
yangjf2019: Thanks for your remind,got it!

PR #4894, opened 2025-01-13
[rest] Add name to GetTableResponse and remove path

JingsongLi:
### Purpose

1. Table name should be added in `GetTableResponse`, for example, get bundle tables should return `List<GetTableResponse>`.
2. path should be removed, because it should be in options.
3. Add id to table and database.


jerry-024:
LGTM

PR #4897, opened 2025-01-13
[core] Fix that sequence group fields are mistakenly aggregated by default aggregator in partial update

yuzelin:
### Purpose

We forbid to set agg function for sequence group fields, but default agg won't be detected, so the sequence group fields might be aggregated by default agg.


zhuangchong (approved):
+1

wwj6591812:
+1

JingsongLi:
+1

PR #4657, opened 2024-12-07
[core] Unify the order of procedure loading properties

askwang:
### Purpose
If the table property have related properties, we do not need to configure them separately in procedure, unify them for expire_partitions and expire_snapshots.

1. Properties loading priority: procedure param > options param > table properties.
2. Add 'options' parameter for some procedures, we can control some behaviors, such as manifest merge.

PR #4901, opened 2025-01-14
[core] Use table options if we have not specified partition-expire op…

LinMingQiang:
…tions.

### Purpose

Should use table configuration if no partition expiration parameters are specified.

In most cases, these parameters are configured in the table configuration and do not need to be specified manually.

Linked issue: close #xxx


LinMingQiang:
Duplicate submission :   https://github.com/apache/paimon/pull/4657

PR #4380, opened 2024-10-25
[flink] add coordinate and worker operator for small changelog files compaction

LsomeYeah:
### Purpose

Linked issue: close #xxx

Add a Coordinator node to small changelog files compaction pipeline to decide how to concatenate it into a target file size result file, which can be one or multiple files, and  add a worker node to merge those small files.

PR #4902, opened 2025-01-14
[flink] Introduce precommit compact for newly created files in unaware bucket tables

tsreaper:
### Purpose

In #4380 we introduce a pre-commit compaction for changelog files in primary key tables, in order to decrease the number of small files.

In this PR we'll introduce a similar feature for unaware bucket tables, to compact and remove small newly-created files before commit.

### Tests

Unit tests and IT cases.

### API and Format

No format changes.

### Documentation

Document is also updated.


JingsongLi (approved):
+1

PR #4904, opened 2025-01-14
[core] Extract loadTable in CatalogUtils

JingsongLi:
### Purpose

Loading table logicals were in `RESTCatalog` and `AbstractCatalog` and `HiveCatalog`, we can unify them to one class util.


jerry-024:
LGTM

PR #4907, opened 2025-01-14
[flink] Copy bytes with multiple threads when preforming precommit compact for changelogs

tsreaper:
### Purpose

In #4380 we introduce pre-commit compact for changelog files. Multiple changelog files from the same partition will be merged into one big file in one worker parallelism to decrease the number of small files.

However, when the number of changelog files to merge is large (while each file itself is small enough), the copying process will be slow, because opening these many files from the filesystem takes a lot of time.

In this PR, we add a thread pool to the worker operator, so that when performing pre-commit compact for changelogs, we can copy the bytes with multiple threads, thus speeding up the process.

### Tests

Existing IT cases should cover this change. This PR also adds a unit test for the coordinator operator.

### API and Format

No format changes.

### Documentation

Document is also updated.

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/compact/changelog/ChangelogCompactTask.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/compact/changelog/ChangelogCompactTask.java
@@ -23,19 +23,24 @@ import org.apache.paimon.flink.compact.changelog.format.CompactedChangelogReadOn
 import org.apache.paimon.flink.sink.Committable;
 import org.apache.paimon.fs.Path;
 import org.apache.paimon.fs.PositionOutputStream;
-import org.apache.paimon.fs.SeekableInputStream;
 import org.apache.paimon.io.CompactIncrement;
 import org.apache.paimon.io.DataFileMeta;
 import org.apache.paimon.io.DataFilePathFactory;
 import org.apache.paimon.io.DataIncrement;
+import org.apache.paimon.options.MemorySize;
 import org.apache.paimon.table.FileStoreTable;
 import org.apache.paimon.table.sink.CommitMessageImpl;
 import org.apache.paimon.utils.FileStorePathFactory;
 import org.apache.paimon.utils.IOUtils;
 import org.apache.paimon.utils.Preconditions;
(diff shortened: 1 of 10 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/compact/changelog/ChangelogCompactTask.java
```diff
@@ -247,6 +259,39 @@ public String toString() {
...
+            this.path = path;
+            this.bucket = bucket;
+            this.isCompactResult = isCompactResult;
+            this.meta = meta;
+        }
+
+        private void readFully() {
+            try {
+                result = IOUtils.readFully(table.fileIO().newInputStream(path), true);
+                table.fileIO().deleteQuietly(path);
```
wwj6591812, line 288: If job failover after `table.fileIO().deleteQuietly(path);` and before `copy all files into a new big file`. 
Is there a risk of file loss here？

tsreaper: If job fails then no changelog will be committed, thus no risk of file loss.

wwj6591812:
+1

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/compact/changelog/ChangelogCompactTask.java
```diff
@@ -81,69 +91,71 @@ public Map<Integer, List<DataFileMeta>> compactChangelogFiles() {
...
+                    }
+                };
+        addTasks.accept(newFileChangelogFiles, false);
+        addTasks.accept(compactChangelogFiles, true);
+
+        Iterator<ReadTask> it =
+                ThreadPoolUtils.randomlyExecuteSequentialReturn(
+                        executor,
+                        t -> {
+                            // Total lengths of all bytes will not exceed `targetFileSize * 2`,
```
JingsongLi, line 122: I feel it is better to use workers and queue and consumer. Even max `targetFileSize * 2`, if target file size is 1GB, this is too still large.

Workers and queue is safer.

JingsongLi:
+1

PR #4908, opened 2025-01-14
[core] Support view API in Rest catalog

jerry-024:
### Purpose
Support view API in RESTCatalog

Linked issue: #4540

### Tests

- CatalogTestBase.testView


paimon-common/src/main/java/org/apache/paimon/types/RowType.java
```diff
@@ -51,6 +53,7 @@
 public final class RowType extends DataType {
 
     private static final long serialVersionUID = 1L;
+    private static final String FIELD_FIELDS = "fields";
```
JingsongLi, line 56: keep an empty above

JingsongLi (approved):
Looks good to me!

PR #4911, opened 2025-01-14
[core] Introduce SnapshotCommit to abstract atomically commit

JingsongLi:
### Purpose

For `RESTCatalog`, it has no lock implementation, so for multiple jobs committing, we should abstract `SnapshotCommit` interface:

```java
/** Interface to commit snapshot atomically. */
public interface SnapshotCommit extends AutoCloseable {

    boolean commit(Snapshot snapshot, String branch) throws Exception;

    /** Factory to create {@link SnapshotCommit}. */
    interface Factory extends Serializable {
        SnapshotCommit create(Identifier identifier, SnapshotManager snapshotManager);
    }
}
```

- For non-RESTCatalog, just be `RenamingSnapshotCommit` to be compatible to history implementation.
- And for RESTCatalog, it submits `CommitTableRequest` to REST server to do single thread committing.

And for Schema changing, we don't really the lock, all schema changes happened in Catalog, we can lock them in Catalog.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/AbstractFileStore.java
+++ b/paimon-core/src/main/java/org/apache/paimon/AbstractFileStore.java
@@ -19,6 +19,8 @@
 package org.apache.paimon;
 
 import org.apache.paimon.CoreOptions.ExternalPathStrategy;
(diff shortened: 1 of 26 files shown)


wwj6591812:
+1

PR #4919, opened 2025-01-15
[core] TypeUtils: remove quotes of map value

kwontaeheon:
### Purpose

Fix https://github.com/apache/paimon/issues/4920
Sinking cdc with map type, value surrouned with double quotes.
To prevent generating quotes, 
first check if value is text.

### Tests

https://github.com/apache/paimon/issues/4920#issuecomment-2630112634


kwontaeheon:
I haven't tested it properly, but if that part is modified, the https://github.com/apache/paimon/issues/4920 issue will be resolved. 
You can also modify the format or this code.

AshishKhatkar (approved):
looks good to me.

JingsongLi (approved):
+1

PR #4921, opened 2025-01-16
[flink] Add new type_mapping options to ignore some type changes when spcified

### LINKED ISSUE #4783: [Feature] Ignore decimal type evolution in special case

JackeyLee007:
### Purpose

Linked issue: close #4783 

We use ticdc to collect data from tidb to kafka, but it does not provide specific parameters for types. No length for varcahr, and no precision and scale for decimal. The current schema evoluation would improve them to string and decimal(38, 18), which is not expected.
So we want to keep the type specified manually unchanged. This could be controlled by the new type_mapping options. NO_CHNAGE, DECIMAL_NO_CHANGE.

### Tests

Observe the varchar and decimal field definition after new value ingested.

### API and Format

No.

### Documentation

Yes and provided


JackeyLee007:
@yuzelin Please help review the new type_mapping options

PR #4922, opened 2025-01-16
[core] Add min_partition_stats and max_partition_stats columns to manifests system table

tsreaper:
### Purpose

`ManifestFileMeta` contains `partitionStats`, however manifests system table does not contain such information.

This PR adds `min_partition_stats` and `max_partition_stats` columns to the system table.

### Tests

Unit tests.

### API and Format

No format changes.

### Documentation

Document is also updated.


JingsongLi:
+1

PR #4925, opened 2025-01-16
[hotfix] [cdc] Add else conditional statement

yangjf2019:
### Purpose

Add else conditional statement.


yangjf2019:
CC @JingsongLi

LinMingQiang:
What's the purpose of this change?

yangjf2019:
The logic of this code is relatively simple and can be dispensed with.

Zouxxyy:
-1,  the two writing styles are essentially the same; in fact, the old style is even better. Please refer to Early Return programming practices.

JingsongLi:
-1 too, I second @Zouxxyy

yangjf2019:
Thanks, Got it!

PR #4926, opened 2025-01-16
[core] Introduce DeleteTagsByTimestampProcedure to delete tags by timestamp

xuzifu666:
### Purpose

Linked issue: close #xxx
Currently user delete tags can only according to tag names，actually user may need delete their tags according to a time （such as history auto-create tags which not contain retained expire time），need introduce a quick way to delete all tags which earlier than a time，this pr is aimed to support it.

### DIFF
--- a/paimon-flink/paimon-flink-1.18/src/main/java/org/apache/paimon/flink/procedure/DeleteTagsByTimestampProcedure.java
+++ b/paimon-flink/paimon-flink-1.18/src/main/java/org/apache/paimon/flink/procedure/DeleteTagsByTimestampProcedure.java
@@ -1,0 +1,74 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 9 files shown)


LinMingQiang:
Is it possible to just add parameters to the original `delelte_tag`?

xuzifu666:
> Is it possible to just add parameters to the original `delelte_tag`?

Thanks, can only add a new params in delete_tag.

paimon-flink/paimon-flink-1.18/src/main/java/org/apache/paimon/flink/procedure/DeleteTagsByTimestampProcedure.java
```diff
@@ -0,0 +1,74 @@
...
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+
+        if (deleteTags.size() > 0) {
+            table.deleteTags(String.join(",", deleteTags));
+        }
+
+        return new String[] {"Success"};
+    }
```
LinMingQiang, line 68: Returns the all tagnames that have been deleted.

JingsongLi:
So my suggestion is introducing expire_tags

askwang:
The existed expire_tags procedure maybe can meet this demand.

askwang:
> So my suggestion is introducing `expire_tags` procedure, this procedure should be similiar to `expire_snapshots` procedure.

The expire_tags procedure is already supported, maybe can meet this demand.

xuzifu666:
@JingsongLi @askwang  master branch had according to file last modified time to expire tags，so this pr can be close first，Thanks.

PR #4928, opened 2025-01-16
[hotfix] Minor fix the int parsing in PostgresRecordParser

yangjf2019:
### Purpose

Fixed the type conversion method.


yangjf2019:
CC @JingsongLi

JingsongLi:
+1

PR #4930, opened 2025-01-16
[core] Fix race condition for earliest snapshot

### LINKED ISSUE #3479: [Bug]  streaming read  by  `from-timestamp`   may be occur exception（snapshot file not fund）  sometimes
### Search before asking

### Paimon version

0.8

### Compute Engine

Flink

## Analyze source code in  `SnapshotManager.class` 

when read  earliest snapshot file at step2， the snapshot maybe expired and deleted

query job will failed by throw FileNotFoundException

### Anything else?

_No response_

yunfengzhou-hub:
### Purpose

Linked issue: close #3479

This PR fixes a racing bug about earliest snapshot. When thread A is trying to get the earliest snapshot of a table, while the snapshot is being expired by thread B, thread A might throw exception that it cannot get the snapshot information from the earliest snapshot id.

This PR fixes the bug by adding retrying logic to seek the second earliest snapshot if the first one cannot be found.

Note that, theoretically speaking, any snapshot (including the latest one) could encounter this issue, but the probability is quite small, so this PR only fixes the bug for invocations on the earliest snapshot.

### Tests

Tests in SnapshotManagerTest are used to verify against the changes.

### API and Format

### Documentation

This change does not introduce a new feature.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/utils/SnapshotManager.java
+++ b/paimon-core/src/main/java/org/apache/paimon/utils/SnapshotManager.java
@@ -163,24 +163,6 @@ public class SnapshotManager implements Serializable {
         return snapshot;
     }
 
-    private @Nullable Snapshot tryGetEarliestSnapshotLaterThanOrEqualTo(
-            long snapshotId, long stopSnapshotId, boolean includeChangelog) {
-        FunctionWithException<Long, Snapshot, FileNotFoundException> snapshotFunction =
-                includeChangelog ? this::tryGetChangelogOrSnapshot : this::tryGetSnapshot;
-        while (snapshotId <= stopSnapshotId) {
-            try {
-                return snapshotFunction.apply(snapshotId);
-            } catch (FileNotFoundException e) {
-                LOG.warn(
-                        "The earliest snapshot or changelog was once identified but disappeared. "
-                                + "It might have been expired by other jobs operating on this table. "
-                                + "Searching for the second earliest snapshot or changelog instead. ");
-                snapshotId++;
-            }
-        }
-        return null;
-    }
-
     public Changelog changelog(long snapshotId) {
         Path changelogPath = longLivedChangelogPath(snapshotId);
         return Changelog.fromPath(fileIO, changelogPath);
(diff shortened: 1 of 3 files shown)


paimon-core/src/main/java/org/apache/paimon/utils/SnapshotManager.java
```diff
@@ -268,25 +301,43 @@ private Snapshot changelogOrSnapshot(long snapshotId) {
...
      * returned if all snapshots are equal to or later than the timestamp mills.
      */
     public @Nullable Long earlierThanTimeMills(long timestampMills, boolean startFromChangelog) {
-        Long earliestSnapshot = earliestSnapshotId();
+        Long earliestSnapshotId = earliestSnapshotId();
         Long earliest;
         if (startFromChangelog) {
             Long earliestChangelog = earliestLongLivedChangelogId();
-            earliest = earliestChangelog == null ? earliestSnapshot : earliestChangelog;
+            earliest = earliestChangelog == null ? earliestSnapshotId : earliestChangelog;
```
wwj6591812, line 321: earliestChangelogId

paimon-core/src/main/java/org/apache/paimon/utils/SnapshotManager.java
```diff
@@ -209,7 +226,23 @@ public boolean longLivedChangelogExists(long snapshotId) {
 
     public @Nullable Snapshot earliestSnapshot() {
         Long snapshotId = earliestSnapshotId();
-        return snapshotId == null ? null : snapshot(snapshotId);
+        if (snapshotId == null) {
+            return null;
+        }
+
+        Long latest = null;
```
wwj6591812, line 233: latestSnapshotId

paimon-core/src/main/java/org/apache/paimon/utils/SnapshotManager.java
```diff
@@ -268,25 +301,43 @@ private Snapshot changelogOrSnapshot(long snapshotId) {
...
             return null;
         }
 
-        if (changelogOrSnapshot(earliest).timeMillis() >= timestampMills) {
+        Snapshot earliestSnapshot = null;
+        while (earliest <= latest) {
+            try {
+                earliestSnapshot = tryGetChangelogOrSnapshot(earliest);
+                break;
+            } catch (FileNotFoundException e) {
```
wwj6591812, line 335: Add a log here?

paimon-core/src/main/java/org/apache/paimon/utils/SnapshotManager.java
```diff
@@ -209,7 +226,23 @@ public boolean longLivedChangelogExists(long snapshotId) {
...
-        return snapshotId == null ? null : snapshot(snapshotId);
+        if (snapshotId == null) {
+            return null;
+        }
+
+        Long latest = null;
+        do {
+            try {
+                return tryGetSnapshot(snapshotId);
+            } catch (FileNotFoundException e) {
```
wwj6591812, line 237: Add a log here?

paimon-core/src/main/java/org/apache/paimon/utils/SnapshotManager.java
```diff
@@ -163,11 +163,28 @@ public Snapshot tryGetSnapshot(long snapshotId) throws FileNotFoundException {
         return snapshot;
     }
 
+    private @Nullable Snapshot tryGetEarliestSnapshotLaterThanOrEqualTo(
+            long snapshotId, long stopId) {
```
wwj6591812, line 167: stopSnapshotId?

paimon-core/src/main/java/org/apache/paimon/utils/SnapshotManager.java
```diff
@@ -434,9 +487,17 @@ private Snapshot changelogOrSnapshot(long snapshotId) {
...
+
+        if (earliestSnapShot == null) {
+            return null;
+        }
+
         Long earliestWatermark = null;
         // find the first snapshot with watermark
-        if ((earliestWatermark = snapshot(earliest).watermark()) == null) {
+        if ((earliestWatermark = earliestSnapShot.watermark()) == null) {
             while (earliest < latest) {
```
JingsongLi, line 501: I think there are some risks. Earliest should be re-assigned.

Maybe you should just optimize `earliestSnapshot`. And use this method in many places.

yunfengzhou-hub: Given that some earliestSnapshot methods use binary search while others use linear, I extracted and optimized two kinds of paths. Please take a look.

PR #4932, opened 2025-01-16
[test] Fix the unstable testCloneWithSchemaEvolution

tsreaper:
### Purpose

`testCloneWithSchemaEvolution` is an unstable test, because the target table is created in `PickFilesForCloneOperator`, but its latest schema file might be overwritten concurrently in `CopyFileOperator`, and at that time target table can't be retrieved from the catalog.

This PR fixes the issue by copying the latest schema file in `PickFilesForCloneOperator`, so it will never be overwritten.

### Tests

IT cases.

### API and Format

No format changes.

### Documentation

No new feature.


JingsongLi:
+1

PR #4934, opened 2025-01-16
[hotfix] [ci] Keep the version of hugo being used consistent

yangjf2019:
### Purpose

Keep the version of hugo being used consistent.

### DIFF
--- a/tools/ci/docs.sh
+++ b/tools/ci/docs.sh
@@ -17,8 +17,8 @@
 # limitations under the License.
 ################################################################################
 
-HUGO_REPO=https://github.com/gohugoio/hugo/releases/download/v0.80.0/hugo_extended_0.80.0_Linux-64bit.tar.gz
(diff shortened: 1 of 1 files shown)


yangjf2019:
Currently, `.github/workflows/docs-tests.yml` is using version 0.124.1, but `tools/ci/docs.sh` is using version 0.80.0, so we need to think about keeping it consistent.

yangjf2019:
CC @JingsongLi

xuzifu666:
@yangjf2019  Hi，could you elaborate this pr can bring any improvement？seem would bring some other unknown risk.

yangjf2019:
> Currently, `.github/workflows/docs-tests.yml` is using version 0.124.1, but `tools/ci/docs.sh` is using version 0.80.0, so we need to think about keeping it consistent.

Thank you for your patience in replying! I was worried about the impact since I was using two different versions, so that's why I changed it. It did not take into account the risks you mentioned.

yangjf2019:
closed

PR #4936, opened 2025-01-17
[core] Make CatalogContext implements Serializable

JingsongLi:
### Purpose

For #4891, we should make CatalogContext Serializable to create FileIO lazily.


Zouxxyy:
ncountered the same requirement!  And I noticed that we have a duplicate class called `SerializableHadoopConfigWrapper`, and perhaps we can remove it.

PR #4940, opened 2025-01-17
[test] Fix the unstable testNoChangelogProducerStreamingRandom

tsreaper:
### Purpose

In #4933 we fix unstable IT cases in `PrimaryKeyFileStoreTableITCase` by mimicing unbounded streams with lots of records. However `testNoChangelogProducerStreamingRandom` will wait for the job to finish and then start checking result, so it is now prone to time out.

This PR fixes the unstable test by checking the result as soon as enough record is written into Paimon.


JingsongLi (approved):
+1

PR #4941, opened 2025-01-17
[flink] Introduce scan bounded to force bounded in streaming job

JingsongLi:
### Purpose

We need to introduce the pattern of running bounded reads in stream jobs, which is useful in certain application scenarios, such as broadcasting a Paimon table to the stream operator.

### Tests

`BatchFileStoreITCase.testScanBounded`


wwj6591812:
+1

PR #4948, opened 2025-01-20
[refactor] Clean unused codes in Lock

yuzelin:
### Purpose

After #4911 , the `LockFactory` is not used any more.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/Lock.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/Lock.java
@@ -22,7 +22,6 @@ import org.apache.paimon.annotation.Public;
 import org.apache.paimon.catalog.CatalogLock;
 import org.apache.paimon.catalog.Identifier;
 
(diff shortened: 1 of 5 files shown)


paimon-core/src/main/java/org/apache/paimon/operation/Lock.java
```diff
@@ -45,43 +41,10 @@ interface Factory extends Serializable {
...
-    static Factory factory(
-            @Nullable CatalogLockFactory lockFactory,
-            @Nullable CatalogLockContext lockContext,
-            Identifier tablePath) {
-        return lockFactory == null
-                ? new EmptyFactory()
-                : new LockFactory(lockFactory, lockContext, tablePath);
-    }
-
     static Factory emptyFactory() {
```
JingsongLi, line 44: modify this to `Lock Lock.empty()`

wwj6591812:
+1

JingsongLi (approved):
+1

PR #4959, opened 2025-01-20
[rest] Refactor AuthProviders to remove credential concept

JingsongLi:
### Purpose

Now we have too many concepts, token, credential, auth...

We can refactor credential to auth, to be simpler.


jerry-024:
LGTM

PR #4966, opened 2025-01-20
[core] Clear cache when deleting the snapshot

Zouxxyy:
### Purpose

Clear cache when deleting the snapshot

```sql
CREATE TABLE T (id INT);
NSERT INTO T VALUES (1), (2), (3), (4);
DELETE FROM T WHERE id = 1;
CALL sys.rollback(table => 'T', version => '1');
DELETE FROM T WHERE id = 1;
SELECT * FROM T ORDER BY id;
```

```
Job aborted due to stage failure: Task 1 in stage 9.0 failed 1 times, most recent failure: Lost task 1.0 in stage 9.0 (TID 9) (30.221.116.223 executor driver): java.io.FileNotFoundException: File '/private/var/folders/px/y3gybll50ggctcjp2t4r2b500000gp/T/spark-5b259941-137d-4736-968a-5b61e456898a/test.db/T/bucket-0/data-e9507592-0884-457f-86d1-fbfc996225b2-0.parquet' not found, Possible causes: 1.snapshot expires too fast, you can configure 'snapshot.time-retained' option with a larger value. 2.consumption is too slow, you can improve the performance of consumption (For example, increasing parallelism).
	at org.apache.paimon.utils.FileUtils.checkExists(FileUtils.java:113)
	at org.apache.paimon.io.DataFileRecordReader.<init>(DataFileRecordReader.java:55)
	at org.apache.paimon.operation.RawFileSplitRead.createFileReader(RawFileSplitRead.java:242)
	at org.apache.paimon.operation.RawFileSplitRead.lambda$createReader$3(RawFileSplitRead.java:181)
	at org.apache.paimon.mergetree.compact.ConcatRecordReader.create(ConcatRecordReader.java:53)
	at org.apache.paimon.operation.RawFileSplitRead.createReader(RawFileSplitRead.java:189)
	at org.apache.paimon.operation.RawFileSplitRead.createReader(RawFileSplitRead.java:141)
```

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/ExpireSnapshotsImpl.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/ExpireSnapshotsImpl.java
@@ -252,7 +252,7 @@ public class ExpireSnapshotsImpl implements ExpireSnapshots {
             if (expireConfig.isChangelogDecoupled()) {
                 commitChangelog(new Changelog(snapshot));
             }
-            snapshotManager.fileIO().deleteQuietly(snapshotManager.snapshotPath(id));
+            snapshotManager.deleteSnapshot(id);
         }
 
         writeEarliestHint(endExclusiveId);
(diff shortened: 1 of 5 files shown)


JingsongLi (approved):
+1

PR #4972, opened 2025-01-21
[spark] Fix update table with char type

Zouxxyy:
### Purpose

```sql
CREATE TABLE T (id INT, s STRING, c CHAR(1));
INSERT INTO T VALUES (1, 's', 'a');
UPDATE T SET c = 'b' WHERE id = 1;
```

```
UPDATE TABLE is not supported temporarily.
org.apache.spark.SparkUnsupportedOperationException: UPDATE TABLE is not supported temporarily.
	at org.apache.spark.sql.errors.QueryExecutionErrors$.ddlUnsupportedTemporarilyError(QueryExecutionErrors.scala:1089)
	at org.apache.spark.sql.execution.SparkStrategies$BasicOperators$.apply(SparkStrategies.scala:935)
	at org.apache.spark.sql.catalyst.planning.QueryPlanner.$anonfun$plan$1(QueryPlanner.scala:63)
```

Note that this PR actually ignores the `read side padding` in the `Project`, so if the char type is used and `spark.sql.legacy.charVarcharAsString` is set to true (default is false), and the char type is written in a non-standard way, the reading of char will lack padding. However the current Spark's DSV2 `RewriteRowLevelCommand` also does not consider this situation, so we will skip it as well.


JingsongLi:
+1

PR #4974, opened 2025-01-21
[core] Support upper bound in dynamic bucket mode

### LINKED ISSUE #4942: [Feature] Support upper bound in dynamic bucket mode
### Search before asking

### Motivation

In dynamic bucket mode, unlimited buckets lead to an unpredicable number of small files, which lead to stability problems.

### Solution

add option to support the feature

### Anything else?

_No response_

liyubin117:
### Purpose
In dynamic bucket mode, unlimited buckets lead to an unpredicable number of small files, which lead to stability problems. so we should support upper bound in dynamic bucket mode.

Linked issue: close #4942 

### Tests
HashBucketAssignerTest
SimpleHashBucketAssignerTest
- Use new feature, Produce 3 buckets after inserting 12 rows
image
image

- Not use new feature, Produce 6 buckets after inserting 12 rows
image
image

### Documentation
docs/layouts/shortcodes/generated/core_configuration.html

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
@@ -1076,12 +1076,12 @@ public class CoreOptions implements Serializable {
                     .withDescription(
                             "Initial buckets for a partition in assigner operator for dynamic bucket mode.");
 
-    public static final ConfigOption<Integer> DYNAMIC_BUCKET_MAX_BUCKETS_PER_ASSIGNER =
-            key("dynamic-bucket.max-buckets-per-assigner")
+    public static final ConfigOption<Integer> DYNAMIC_BUCKET_MAX_BUCKETS =
+            key("dynamic-bucket.max-buckets")
                     .intType()
                     .defaultValue(-1)
                     .withDescription(
-                            "Max buckets per assigner operator for a partition in dynamic bucket mode, It should "
+                            "Max buckets for a partition in dynamic bucket mode, It should "
                                     + "either be equal to -1 (unlimited), or it must be greater than 0 (fixed upper bound).");
 
     public static final ConfigOption<Integer> DYNAMIC_BUCKET_ASSIGNER_PARALLELISM =
@@ -2227,8 +2227,8 @@ public class CoreOptions implements Serializable {
         return options.get(DYNAMIC_BUCKET_INITIAL_BUCKETS);
     }
 
-    public Integer dynamicBucketMaxBucketsPerAssigner() {
-        return options.get(DYNAMIC_BUCKET_MAX_BUCKETS_PER_ASSIGNER);
+    public Integer dynamicBucketMaxBuckets() {
+        return options.get(DYNAMIC_BUCKET_MAX_BUCKETS);
     }
 
     public Integer dynamicBucketAssignerParallelism() {
--- a/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
@@ -1076,6 +1076,14 @@ public class CoreOptions implements Serializable {
                     .withDescription(
                             "Initial buckets for a partition in assigner operator for dynamic bucket mode.");
 
+    public static final ConfigOption<Integer> DYNAMIC_BUCKET_MAX_BUCKETS_PER_ASSIGNER =
+            key("dynamic-bucket.max-buckets-per-assigner")
+                    .intType()
+                    .defaultValue(-1)
+                    .withDescription(
+                            "Max buckets per assigner operator for a partition in dynamic bucket mode, It should "
+                                    + "either be equal to -1 (unlimited), or it must be greater than 0 (fixed upper bound).");
+
     public static final ConfigOption<Integer> DYNAMIC_BUCKET_ASSIGNER_PARALLELISM =
             key("dynamic-bucket.assigner-parallelism")
                     .intType()
@@ -2217,6 +2225,10 @@ public class CoreOptions implements Serializable {
 
     public Integer dynamicBucketInitialBuckets() {
         return options.get(DYNAMIC_BUCKET_INITIAL_BUCKETS);
+    }
+
+    public Integer dynamicBucketMaxBucketsPerAssigner() {
+        return options.get(DYNAMIC_BUCKET_MAX_BUCKETS_PER_ASSIGNER);
     }
 
     public Integer dynamicBucketAssignerParallelism() {
--- a/paimon-core/src/main/java/org/apache/paimon/index/PartitionIndex.java
+++ b/paimon-core/src/main/java/org/apache/paimon/index/PartitionIndex.java
@@ -24,9 +24,13 @@ import org.apache.paimon.table.sink.KeyAndBucketExtractor;
 import org.apache.paimon.utils.Int2ShortHashMap;
 import org.apache.paimon.utils.IntIterator;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.io.EOFException;
 import java.io.IOException;
 import java.io.UncheckedIOException;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
@@ -39,6 +43,7 @@ import static org.apache.paimon.index.HashIndexFile.HASH_INDEX;
 
 /** Bucket Index Per Partition. */
 public class PartitionIndex {
+    private static final Logger LOG = LoggerFactory.getLogger(PartitionIndex.class);
 
     public final Int2ShortHashMap hash2Bucket;
 
@@ -81,15 +86,9 @@ public class PartitionIndex {
             Long number = entry.getValue();
             if (number < targetBucketRowNumber) {
                 entry.setValue(number + 1);
-                return cacheBucketAndGet(hash, bucket);
+                return cacheBucketAndGet(hash2Bucket, hash, bucket);
             } else {
                 iterator.remove();
-                if (-1 != maxBucketsNum && totalBucket.size() == maxBucketsNum) {
-                    return cacheBucketAndGet(
-                            hash,
-                            KeyAndBucketExtractor.bucketWithUpperBound(
-                                    totalBucket, hash, maxBucketsNum));
-                }
             }
         }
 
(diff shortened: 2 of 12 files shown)


liyubin117:
@JingsongLi CI passed, PTAL, thanks!

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -1076,6 +1076,14 @@ public class CoreOptions implements Serializable {
                     .withDescription(
                             "Initial buckets for a partition in assigner operator for dynamic bucket mode.");
 
+    public static final ConfigOption<Integer> DYNAMIC_BUCKET_MAX_BUCKETS_PER_ASSIGNER =
+            key("dynamic-bucket.max-buckets-per-assigner")
```
JingsongLi, line 1080: I think we should limit total buckets?

liyubin117: This is also my initial purpose, but I found that the bucket generation logic in `BucketAssigner` is a instance behavior and could be parallelized, every assigner handles their own buckets. I have no idea about total buckets limits and maybe we could introudce it later, WDYT?

liyubin117: I found that `dynaimc-bucket.initial-bucket` is also `assigner` granularity behavior, to make a consistence, maybe we should also use `dynamic-bucket.max-buckets` :)

JingsongLi: I still think introducing a max bucket is better.
The option should be very easy to understand.

paimon-core/src/main/java/org/apache/paimon/index/SimpleHashBucketAssigner.java
```diff
@@ -83,7 +88,15 @@ public int assign(int hash) {
 
             Long num = bucketInformation.computeIfAbsent(currentBucket, i -> 0L);
             if (num >= targetBucketRowNumber) {
-                loadNewBucket();
+                if (-1 != maxBucketsNum && bucketInformation.size() >= maxBucketsNum) {
+                    int bucket =
+                            KeyAndBucketExtractor.bucketWithUpperBound(
+                                    bucketInformation.keySet(), hash, maxBucketsNum);
+                    hash2Bucket.put(hash, (short) bucket);
```
wwj6591812, line 95: Add a util function of PartitionIndex#cacheBucketAndGet, then here use?

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
wwj6591812: +1
Add option should be easy to user understand.

liyubin117: reasonable enough, I have found a way to implement total max bucket feature, thanks for reminding that :)

paimon-core/src/main/java/org/apache/paimon/index/SimpleHashBucketAssigner.java
liyubin117: done :)

JingsongLi (changes requested):
I'm not sure if these modifications are effective, so let me give you my suggestion:
We only need to modify PartitionIndex.asign inside:

Only code:
```
// 3. create a new bucket
for (int i = 0; i < Short.MAX_VALUE; i++) {
    if (bucketFilter.test(i) && !totalBucket.contains(i)) {
        hash2Bucket.put(hash, (short) i);
        nonFullBucketInformation.put(i, 1L);
        totalBucket.add(i);
        return i;
    }
}

// 4. too many buckets, throw exception
@SuppressWarnings("OptionalGetWithoutIsPresent")
int maxBucket = totalBucket.stream().mapToInt(Integer::intValue).max().getAsInt();
throw new RuntimeException(
        String.format(
                "Too more bucket %s, you should increase target bucket row number %s.",
                maxBucket, targetBucketRowNumber));
```

New code:
```
// 3. create a new bucket
for (int i = 0; i < max_buckets; i++) {
    if (bucketFilter.test(i) && !totalBucket.contains(i)) {
        hash2Bucket.put(hash, (short) i);
        nonFullBucketInformation.put(i, 1L);
        totalBucket.add(i);
        return i;
    }
}

// 4. exceed max_buckets, just pick a bucket for record.
pick a min bucket (belongs to this task) to the record.
```

liyubin117:
> I'm not sure if these modifications are effective, so let me give you my suggestion: We only need to modify PartitionIndex.asign inside:
> 
> Only code:
> 
> ```
> // 3. create a new bucket
> for (int i = 0; i < Short.MAX_VALUE; i++) {
>     if (bucketFilter.test(i) && !totalBucket.contains(i)) {
>         hash2Bucket.put(hash, (short) i);
>         nonFullBucketInformation.put(i, 1L);
>         totalBucket.add(i);
>         return i;
>     }
> }
> 
> // 4. too many buckets, throw exception
> @SuppressWarnings("OptionalGetWithoutIsPresent")
> int maxBucket = totalBucket.stream().mapToInt(Integer::intValue).max().getAsInt();
> throw new RuntimeException(
>         String.format(
>                 "Too more bucket %s, you should increase target bucket row number %s.",
>                 maxBucket, targetBucketRowNumber));
> ```
> 
> New code:
> 
> ```
> // 3. create a new bucket
> for (int i = 0; i < max_buckets; i++) {
>     if (bucketFilter.test(i) && !totalBucket.contains(i)) {
>         hash2Bucket.put(hash, (short) i);
>         nonFullBucketInformation.put(i, 1L);
>         totalBucket.add(i);
>         return i;
>     }
> }
> 
> // 4. exceed max_buckets, just pick a bucket for record.
> pick a min bucket (belongs to this task) to the record.
> ```

After offline discussion, we have reached a consesus: We can't just update `PartitionIndex` logic because it doesn't handle `SimpleHashBucketAssigner`; When the buckets are full, a random bucket is selected for writing.

paimon-core/src/main/java/org/apache/paimon/index/PartitionIndex.java
```diff
@@ -80,29 +86,35 @@ public int assign(int hash, IntPredicate bucketFilter) {
...
         }
 
-        // 3. create a new bucket
-        for (int i = 0; i < Short.MAX_VALUE; i++) {
-            if (bucketFilter.test(i) && !totalBucket.contains(i)) {
-                hash2Bucket.put(hash, (short) i);
-                nonFullBucketInformation.put(i, 1L);
-                totalBucket.add(i);
-                return i;
+        if (-1 == maxBucketsNum || totalBucket.size() < maxBucketsNum) {
```
JingsongLi, line 95: I think here should be: max bucket should be smaller than maxBucketsNum.

If it is a job restarted from scratch, each task is increasing, and the previous judgment may be problematic.

paimon-core/src/main/java/org/apache/paimon/index/PartitionIndex.java
```diff
@@ -137,4 +149,47 @@ public static PartitionIndex loadIndex(
         }
         return new PartitionIndex(mapBuilder.build(), buckets, targetBucketRowNumber);
     }
+
+    public static int cacheBucketAndGet(Int2ShortHashMap hash2Bucket, int hash, int bucket) {
```
JingsongLi, line 153: Don't need to extract this method, it is too simple.

liyubin117: done :)

paimon-core/src/main/java/org/apache/paimon/index/PartitionIndex.java
liyubin117: Wise consideration! done :)

paimon-core/src/main/java/org/apache/paimon/index/PartitionIndex.java
```diff
@@ -137,4 +153,42 @@ public static PartitionIndex loadIndex(
         }
         return new PartitionIndex(mapBuilder.build(), buckets, targetBucketRowNumber);
     }
+
+    public static int[] getMaxBucketsPerAssigner(int maxBuckets, int assigners) {
```
JingsongLi, line 157: I cannot understand `getMaxBucketsPerAssigner` and `getSpecifiedMaxBuckets`. Why not just same to `maxBucketId < maxBucketsNum - 1`?

paimon-core/src/main/java/org/apache/paimon/index/PartitionIndex.java
```diff
@@ -80,29 +81,46 @@ public int assign(int hash, IntPredicate bucketFilter) {
...
 
-        @SuppressWarnings("OptionalGetWithoutIsPresent")
-        int maxBucket = totalBucket.stream().mapToInt(Integer::intValue).max().getAsInt();
-        throw new RuntimeException(
-                String.format(
-                        "Too more bucket %s, you should increase target bucket row number %s.",
-                        maxBucket, targetBucketRowNumber));
+        // 4. exceed buckets upper bound
+        int bucket =
+                KeyAndBucketExtractor.bucketWithUpperBound(totalBucket, hash, totalBucket.size());
```
JingsongLi, line 121: You should find a bucket for own... Not randomly from all buckets.

paimon-core/src/main/java/org/apache/paimon/index/PartitionIndex.java
```diff
@@ -80,29 +81,46 @@ public int assign(int hash, IntPredicate bucketFilter) {
...
-        for (int i = 0; i < Short.MAX_VALUE; i++) {
-            if (bucketFilter.test(i) && !totalBucket.contains(i)) {
-                hash2Bucket.put(hash, (short) i);
-                nonFullBucketInformation.put(i, 1L);
-                totalBucket.add(i);
-                return i;
+        int maxBucketId =
+                totalBucket.isEmpty()
+                        ? 0
+                        : totalBucket.stream().mapToInt(Integer::intValue).max().getAsInt();
```
JingsongLi, line 94: Here should not invoke the stream every time, you should cache the maxBucketId.

paimon-core/src/main/java/org/apache/paimon/index/PartitionIndex.java
liyubin117: Thanks for your kind remind.
Only when a new hash arrives and the bucket limit is exceeded, a bucket is randomly selected. Then, this bucket and the corresponding hash are placed in the cache. 
After that, whenever this hash arrives, it will always find its own bucket.

paimon-core/src/main/java/org/apache/paimon/index/PartitionIndex.java
liyubin117: Wise consideration :)

JingsongLi (approved):
+1

PR #4975, opened 2025-01-21
[flink] Infer parallelism only in situation of parallelism is not set.

leaves12138:
### Purpose

default.parallelism is not working because scan.infer-parallelism is default true

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/FlinkTableSource.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/FlinkTableSource.java
@@ -149,9 +149,18 @@ public abstract class FlinkTableSource
                     Boolean.parseBoolean(envConfig.toMap().get(FLINK_INFER_SCAN_PARALLELISM)));
         }
         Integer parallelism = options.get(FlinkConnectorOptions.SCAN_PARALLELISM);
-        if (parallelism == null && options.get(FlinkConnectorOptions.INFER_SCAN_PARALLELISM)) {
+        if (parallelism == null
+                // Infer parallelism when parallelism is not set and infer scan parallelism is
+                // enabled.
+                && env.getParallelism() == -1
+                && options.get(FlinkConnectorOptions.INFER_SCAN_PARALLELISM)) {
             if (isUnbounded()) {
-                parallelism = Math.max(1, options.get(CoreOptions.BUCKET));
+                // In unaware bucket or dynamic bucket mode, we can't infer parallelism.
+                if (options.get(CoreOptions.BUCKET) == -1) {
+                    return null;
+                } else {
+                    parallelism = Math.max(1, options.get(CoreOptions.BUCKET));
+                }
             } else {
                 scanSplitsForInference();
                 parallelism = splitStatistics.splitNumber();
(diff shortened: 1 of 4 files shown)


paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/source/DataTableSourceTest.java
```diff
@@ -81,6 +81,8 @@ void testInferScanParallelism() throws Exception {
                         null);
         PaimonDataStreamScanProvider runtimeProvider = runtimeProvider(tableSource);
         StreamExecutionEnvironment sEnv1 = StreamExecutionEnvironment.createLocalEnvironment();
+        sEnv1.setParallelism(-1);
+        System.out.println(sEnv1.getParallelism());
```
yunfengzhou-hub, line 85: This print seems unnecessary.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/FlinkTableSource.java
```diff
@@ -149,9 +149,18 @@ protected Integer inferSourceParallelism(StreamExecutionEnvironment env) {
                     Boolean.parseBoolean(envConfig.toMap().get(FLINK_INFER_SCAN_PARALLELISM)));
         }
         Integer parallelism = options.get(FlinkConnectorOptions.SCAN_PARALLELISM);
-        if (parallelism == null && options.get(FlinkConnectorOptions.INFER_SCAN_PARALLELISM)) {
+        if (parallelism == null
```
yunfengzhou-hub, line 152: I'm trying to understand the parallelism inferring logic this PR wants to achieve, please see if this is correct.

1. If Paimon option SCAN_PARALLELISM is configured, then use this value.
2. Else if Flink configuration `parallelism.default` is configured, then use this value.
3. Else if Paimon option INFER_SCAN_PARALLELISM is set to true, and the parallelism of source can be inferred by Paimon (like when fixed bucket + unbounded stream), then Paimon would provide an inference result.
4. Else Paimon would not set the parallelism of the source operator, instead it would be Flink infra that decide the parallelism of this operator

JingsongLi (approved):
+1

yunfengzhou-hub:
@leaves12138 Thanks for the update. According to our offline discussion, I learned some background information of this PR as follows.

In a Flink job where Paimon source is directly followed by a Paimon sink, the Paimon writer operator's parallelism cannot be dynamically inferred by Flink infra. The reason is because the writer operator sets its parallelism according to the upstream(source) operator, and the source operator's parallelism has explicitly decided.

So in order to allow Flink infra to change the parallelism of the writer operator, this PR made some change to the source operator's implementation. This way sounds indirect and incomplete, as Flink infra can still not set the parallelism in other situations. A better solution would be change
```java
writerDataStream.setParalleism(inputDataStream.getParallelism());
```
to
```java
writerDataStream.getTransformation().setParalleism(inputDataStream.getParallelism(), false);
```

Without regard to the background motivation, this PR itself looks good to me so I'm +1 on it. We may open a next PR in future to use the APIs mentioned above to better facilitate writer operators for Flink infra.

PR #4977, opened 2025-01-21
[core] Fix that sequence fields are mistakenly aggregated by default aggregator in AggregateMergeFunction

yuzelin:
### Purpose

Same as #4897


JingsongLi (approved):
+1

PR #4980, opened 2025-01-22
[core] ObjectRefresh with iterative list and batched commit

### LINKED ISSUE #4971: [Feature] Lower memory usage of `ObjectRefresh`
### Search before asking

### Motivation

The current implementation of `ObjectRefresh` first collects a list of all files under `object-location`, then writes them out to the table. This requires the driver node to have memory enough to reside the entire object listing.

Another problem is that the current implementation generates a new commit for each file in the listing. This can result in an enormous amount of snapshots and poor refresh performance.

### Solution

1. Use `FileIO#listFilesIterative` to load file listing into memory in batches.  
    The final effect of memory saving will depend on the actual implementation of the `FileIO`, but the worst case it can fallback to is what we already have now.
3. Periodically issue commits for batches of writes of a certain batch size.

### Anything else?

_No response_

smdsbz:
### Purpose

Linked issue: close #4971 

* Take advantage of iterative list to lower memory requirement.
* Commit object refresh entries in batches to improve performance.

### Tests

Use existing tests `ObjectTableITCase` and `ObjectTableTest`.

### API and Format

N/A

### Documentation

N/A

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/object/ObjectRefresh.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/object/ObjectRefresh.java
@@ -35,7 +35,7 @@ import java.util.Collections;
 /** Util class for refreshing object table. */
 public class ObjectRefresh {
 
-    private static final long COMMIT_BATCH_SIZE = 1000;
+    private static final long COMMIT_BATCH_SIZE = 10_000;
 
     public static long refresh(ObjectTable table) throws Exception {
         long totalObjs = 0;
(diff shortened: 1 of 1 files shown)


paimon-core/src/main/java/org/apache/paimon/table/object/ObjectRefresh.java
```diff
@@ -23,56 +23,40 @@
...
-import java.util.ArrayList;
 import java.util.Collections;
-import java.util.List;
 
 /** Util class for refreshing object table. */
 public class ObjectRefresh {
 
-    public static long refresh(ObjectTable table) throws Exception {
-        String location = table.objectLocation();
+    private static final long COMMIT_BATCH_SIZE = 1000;
```
JingsongLi, line 38: Too small? Maybe `10_000`?

JingsongLi:
+1

PR #4989, opened 2025-01-23
[spark] Fallback to spark except query if increment query with rescale bucket

Zouxxyy:
### Purpose

Linked issue: close #xxx


JingsongLi (approved):
So fast!

JingsongLi:
We need to handle DELETE records for audit_log table too.

PR #4991, opened 2025-01-24
[parquet] Introduce LongIterator to Parquet RowIndexGenerator

JingsongLi:
### Purpose

Improve performance, we should let rowIndexes to lazied.


leaves12138 (approved):
+1

PR #5001, opened 2025-01-27
[Parquet] Fix parquet decimal type match.

leaves12138:
### Purpose

Fix lenth bytes to decimal type should be convert according to paimon decimal type scale.


JingsongLi (commented):
Please add tests.

leaves12138:
> Please add tests.

Added specific decimal type check tests.

JingsongLi (approved):
+1

PR #5010, opened 2025-02-05
[core] Slightly refatory code in LookupChangelogMergeFunctionWrapper

leaves12138:
### Purpose

We don't need two merge functions in one `LookupChangelogMergeFunctionWrapper `

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/KeyValue.java
+++ b/paimon-core/src/main/java/org/apache/paimon/KeyValue.java
@@ -176,6 +176,16 @@ public class KeyValue {
                 .setLevel(level);
     }
 
+    public KeyValue copy(KeyValueCopier keyValueCopier) {
+        return new KeyValue()
+                .replace(
+                        keyValueCopier.copyKey(key),
+                        sequenceNumber,
+                        valueKind,
+                        keyValueCopier.copyValue(value))
+                .setLevel(level);
+    }
+
     @VisibleForTesting
     public String toString(RowType keyType, RowType valueType) {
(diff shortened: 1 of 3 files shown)


paimon-core/src/main/java/org/apache/paimon/KeyValue.java
```diff
@@ -195,4 +205,23 @@ public static String rowDataToString(InternalRow row, RowType type) {
                                                 .getFieldOrNull(row)))
                 .collect(Collectors.joining(", "));
     }
+
+    /** Copier for KeyValue. */
+    public static class KeyValueCopier {
```
JingsongLi, line 210: Don't introduce this.

paimon-core/src/main/java/org/apache/paimon/mergetree/compact/LookupChangelogMergeFunctionWrapper.java
```diff
@@ -68,6 +67,9 @@ public class LookupChangelogMergeFunctionWrapper<T>
     private final @Nullable DeletionVectorsMaintainer deletionVectorsMaintainer;
     private final Comparator<KeyValue> comparator;
 
+    private final LinkedList<KeyValue> candidates = new LinkedList<>();
+    private final KeyValue.KeyValueCopier keyValueCopier;
```
JingsongLi, line 71: Just store keySerializer and valueSerializer here.

wwj6591812:
+1

JingsongLi:
+1

PR #5011, opened 2025-02-05
[bug] Avoid obtaining the wrong checkpoint configuration in the `SynchronizationAction`

### LINKED ISSUE #4998: [Bug] When CDC Action is not actively configured with checkpoint and is allowed to force snapshots to be submitted, Flink's failover does not take effect.
### Search before asking

### Paimon version

0.8+

### Compute Engine

Flink

### Minimal reproduce step

1. Use cdc action and do not configure checkpoint intervals (Paimon will actively configure checkpoint via: #2461)
2. open  `commit.force-create-snapshot` = `true`
3. Actively kill a taskmanager during a task run, triggering a flink failover.

## Exception
Chances are, we'll find that the task keeps restarting.

## How this led to this Exception

- With `commit.force-create-snapshot` turned on, a commit will be triggered to generate a snapshot of paimon even if no data is written.

- flink's failover resumes from checkpoint, based on paimon's two-phase commit, and resuming from checkpoint may trigger a commit of the manifest file

- If there are no new data files in the commit (via the streamingCheckpointEnabled variable in the `CommitterOperator`), the manifest file commit may be filtered out because we didn't actively set the checkpoint. `streamingCheckpointEnabled= false`, then the commit will be ignored (because no new data file has been generated). Then, after the second job restart, the process will be repeated again, and we will be stuck in a dead loop.

### Anything else?

_No response_

huyuanfeng2018:
### Purpose

Linked issue: close #4998

### DIFF
--- a/paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/action/cdc/SynchronizationActionBase.java
+++ b/paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/action/cdc/SynchronizationActionBase.java
@@ -216,10 +216,10 @@ public abstract class SynchronizationActionBase extends ActionBase {
 
     @Override
     public void run() throws Exception {
-        build();
         if (!env.getCheckpointConfig().isCheckpointingEnabled()) {
             env.enableCheckpointing(DEFAULT_CHECKPOINT_INTERVAL);
         }
+        build();
         execute(syncJobHandler.provideDefaultJobName());
     }
 }


huyuanfeng2018:
@yuzelin Can help review this pr? Thank you for your help~

leaves12138:
Thanks for @huyuanfeng2018, it is a hard to detect bug. Can you provide to test to avoid some one else revert it?

yuzelin:
+1. I think setting env options before using it to build job graph is always right.

PR #5013, opened 2025-02-05
[bug] Fix type bug when using Boolean in bitmap for file index

jerry-024:
### Purpose
Fix type bug when using Boolean in bitmap for file index. Current use byte will cause: 
`java.lang.ClassCastException：class java.lang.Boolean cannot be cast to class java.lang.Byte`

### Tests
TestBitmapFileIndex.testBitmapIndexForBooleanType

### API and Format
NO

### Documentation
NO


leaves12138 (approved):
+1

PR #5023, opened 2025-02-06
[Feature] Support default values for specified fields

### LINKED ISSUE #5015: [Feature] Support default values for specified fields.
### Search before asking

### Motivation

Support default values for specified fields with Spark

### Solution

_No response_

### Anything else?

_No response_

davidyuan1223:
… string type)

### Purpose

Linked issue: close #5015 

paimon spark support default value for specified fields.(Current only string type)
(Testing, not need check)

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/commands/PaimonSparkWriter.scala
+++ b/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/commands/PaimonSparkWriter.scala
@@ -27,7 +27,6 @@ import org.apache.paimon.deletionvectors.append.AppendDeletionFileMaintainer
 import org.apache.paimon.index.{BucketAssigner, SimpleHashBucketAssigner}
 import org.apache.paimon.io.{CompactIncrement, DataIncrement, IndexIncrement}
 import org.apache.paimon.manifest.{FileKind, IndexManifestEntry}
-import org.apache.paimon.operation.DefaultValueAssigner
 import org.apache.paimon.spark.{SparkRow, SparkTableWrite, SparkTypeUtils}
 import org.apache.paimon.spark.schema.SparkSystemColumns.{BUCKET_COL, ROW_KIND_COL}
 import org.apache.paimon.spark.util.SparkRowUtils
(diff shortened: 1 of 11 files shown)


Zouxxyy:
Thanks for the implementation, We hope its final form refers to https://issues.apache.org/jira/browse/SPARK-38334:
1. Allow table creation using Spark's default value syntax, while also saving it in paimon metadata format to facilitate recognition by other clients.
2. Refer to Spark's own implementation of default values as much as possible.

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/commands/PaimonSparkWriter.scala
```diff
@@ -66,12 +67,30 @@ case class PaimonSparkWriter(table: FileStoreTable) {
...
+
+    // 1. init defaultValueAssigner
+    val defaultValueAssigner = DefaultValueAssigner.create(tableSchema)
+
+    // 2. judge need assign default value
+    if (defaultValueAssigner.needToAssign()) {
+      // 3. set default value for null
+      defaultValueAssigner.getDefaultValues.forEach(
+        (fieldName, defaultValue) => {
+          // 使用 isNull 来检查是否为 null 值
```
Zouxxyy, line 80: Please do not use Chinese

davidyuan1223:
> Thanks for the implementation, We hope its final form refers to https://issues.apache.org/jira/browse/SPARK-38334:
> 
> 1. Allow table creation using Spark's default value syntax, while also saving it in paimon metadata format to facilitate recognition by other clients.
> 2. Refer to Spark's own implementation of default values as much as possible.

lgtm, i will try to implement this

davidyuan1223:
@Zouxxyy hello, wanna ask a question, if we load create table with defalt value's logical plan, the default filed should put in where? In fields?
```json
{
  "version" : 3,
  "id" : 0,
  "fields" : [ {
    "id" : 0,
    "name" : "id",
    "type" : "INT"
  }, {
    "id" : 1,
    "name" : "t1",
    "type" : "INT"
  }, {
    "id" : 2,
    "name" : "t2",
    "type" : "INT"
  } ],
  "highestFieldId" : 2,
  "partitionKeys" : [ ],
  "primaryKeys" : [ ],
  "options" : {
    "owner" : "fuyuanyuan"
  },
  "timeMillis" : 1739187234928
}
```
like this?
```json
{
  "version" : 3,
  "id" : 0,
  "fields" : [ {
    "id" : 0,
    "name" : "id",
    "type" : "INT",
    "default": "1"
  }, {
    "id" : 1,
    "name" : "t1",
    "type" : "INT",
    "default": null,
  }, {
    "id" : 2,
    "name" : "t2",
    "type" : "INT",
    "default": "2"
  } ],
  "highestFieldId" : 2,
  "partitionKeys" : [ ],
  "primaryKeys" : [ ],
  "options" : {
    "owner" : "fuyuanyuan"
  },
  "timeMillis" : 1739187234928
}
```

davidyuan1223:
> > Thanks for the implementation, We hope its final form refers to https://issues.apache.org/jira/browse/SPARK-38334:
> > 
> > 1. Allow table creation using Spark's default value syntax, while also saving it in paimon metadata format to facilitate recognition by other clients.
> > 2. Refer to Spark's own implementation of default values as much as possible.
> 
> lgtm, i will try to implement this

basic of this, i implement an esay pr, we could discuss how to improve it

Zouxxyy:
> @Zouxxyy hello, wanna ask a question, if we load create table with defalt value's logical plan, the default filed should put in where? In fields?

Since Paimon already has a prop for default values, we can just use that, what do you think @JingsongLi 

So I think we can implement it in the following steps:

1. Support Paimon's built-in default value syntax for spark reading (this is actually your first PR)

```sql
CREATE TABLE T (id INT, name STRING)TBLPROPERTIES ('fields.name.default-value' = 'x');
```

2. Support Spark's DDL syntax (for Spark version 3.4 and above), which will save 'fields.name.default-value' to Paimon's properties.
```sql
CREATE TABLE T (id INT, name STRING DEFAULT 'x');
```

davidyuan1223:
> > @Zouxxyy hello, wanna ask a question, if we load create table with defalt value's logical plan, the default filed should put in where? In fields?
> 
> @davidyuan1223 Since Paimon already has a prop for default values, we can just use that, what do you think @JingsongLi
> 
> So I think we can implement it in the following steps:
> 
> 1. Support Paimon's built-in default value syntax for spark reading (this is actually your first PR)
> 
> ```sql
> CREATE TABLE t (id INT, name STRING)TBLPROPERTIES ('fields.name.default-value' = 'default v');
> 
> -- check
> INSERT INTO t (id) values (1);
> SELECT * FROM t;
> ```
> 
> 2. Support Spark's DDL syntax (for Spark version 3.4 and above), which will save 'fields.name.default-value' to Paimon's properties.
> 
> ```sql
> CREATE TABLE t (id INT, name STRING DEFAULT 'default v');
> ```

because i saw the pr https://issues.apache.org/jira/browse/SPARK-38334 which implementation metadata field, so i implementation a pr with the metadata field in paimon metadata file, shall we use this? or keep the properties?

Zouxxyy:
Agree, we can also consider supporting schema evolution, like changing the default value. These need to be carefully designed.

davidyuan1223:
LGTM, i think add metadata or other in filed detail may could help us compatiable with more components, we could discuss.

PR #5028, opened 2025-02-06
[common] Introduce BitmapFileIndexMetaV2.

hang8929201:
### Purpose

When the bitmap-indexed column cardinality is high, using the first version of the bitmap index format will take a lot of time to read the entire dictionary. But in fact we don't need a full dictionary when dealing with a small number of predicates, the performance of predicate hits on the bitmap can be improved by creating a secondary index on the dictionary.

https://docs.google.com/document/d/11dJlGlSX3dwYKKrPN0DQ2XQTsx6d9wI6DTBIiiBwomM/edit?tab=t.0

**performance**

cardinality 1000:
image

cardinality 10000:
image

cardinality 30000:
image

cardinality 50000:
image

cardinality 80000:
image

cardinality 100000:
image

### Tests

org.apache.paimon.fileindex.bitmapindex.TestBitmapFileIndex
org.apache.paimon.spark.SparkFileIndexITCase
org.apache.paimon.benchmark.bitmap.BitmapIndexBenchmark

### Documentation

docs/content/concepts/spec/fileindex.md


paimon-common/src/main/java/org/apache/paimon/fileindex/bitmap/BitmapFileIndex.java
```diff
@@ -235,8 +271,18 @@ private RoaringBitmap32 readBitmap(Object bitmapId) {
                     if (offset < 0) {
                         return RoaringBitmap32.bitmapOf(-1 - offset);
                     } else {
-                        seekableInputStream.seek(bodyStart + offset);
+                        seekableInputStream.seek(bitmapFileIndexMeta.getBodyStart() + offset);
```
Tan-JiaLiang, line 274: It will call `findBlocks` in three times (`contains`, `getOffset`, `getLength`) if match in V2 format. Maybe we can avoid this.

paimon-common/src/main/java/org/apache/paimon/fileindex/bitmap/BitmapFileIndexMetaV2.java
```diff
@@ -0,0 +1,469 @@
...
+    @Override
+    public int getLength(Object bitmapId) {
+        if (bitmapId == null) {
+            return nullBitmapLength;
+        }
+        BitmapIndexBlock block = findBlock(bitmapId);
+        return block.getLength(bitmapId);
+    }
+
+    private BitmapIndexBlock findBlock(Object bitmapId) {
```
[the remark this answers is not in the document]
Tan-JiaLiang: > Besides, It seems to be have NPE problem if findBlock return null.

Ignore it, but the `findBlock` will be called three times if bitmapId exists.

paimon-common/src/main/java/org/apache/paimon/fileindex/bitmap/BitmapFileIndex.java
Tan-JiaLiang: This problem is also exists in the `BitmapIndexBlock`.

paimon-common/src/main/java/org/apache/paimon/fileindex/bitmap/BitmapFileIndexMetaV2.java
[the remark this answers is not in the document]
hang8929201: Done

paimon-common/src/main/java/org/apache/paimon/fileindex/bitmap/BitmapFileIndex.java
hang8929201: The metadata information of blocks is in memory, and binary search is used, so the performance is not bad.

Tan-JiaLiang: It makes me feels obsessive-compulsive. So I want to suggest one more time.

For V2: find the blocks and then find the entry, only 2 binary search, but now need to binary search six times.

What do you think about this? Using the `findEntry` instead of `contains`/`getOffset`/`getLength`.

```java
private RoaringBitmap32 readBitmap(Object bitmapId) {
    Entry entry = bitmapFileIndexMeta.findEntry(bitmapId);
    if (entry == null) {
        return new RoaringBitmap32();
    }

    int offset = entry.getOffset();
    int length = entry.getLength();
    if (offset < 0) {
        return RoaringBitmap32.bitmapOf(-1 - offset);
    }

    try {
        seekableInputStream.seek(bitmapFileIndexMeta.getBodyStart() + offset);
        RoaringBitmap32 bitmap = new RoaringBitmap32();
        if (enableNextOffsetToSize && length != -1) {
            DataInputStream input = new DataInputStream(seekableInputStream);
            byte[] bytes = new byte[length];
            input.readFully(bytes);
            bitmap.deserialize(ByteBuffer.wrap(bytes));
        } else {
            bitmap.deserialize(new DataInputStream(seekableInputStream));
        }
        return bitmap;
    } catch (Exception e) {
    }
}
```

hang8929201: Accepted, thanks to JiaLiang for checking the code so carefully.

Tan-JiaLiang (approved):
@hang8929201 Thanks for your patience!

+1 cc @JingsongLi

2k124:
+1

hang8929201:
I found another question that needs to be discussed. Should the index version default to v1 or v2? @Tan-JiaLiang @JingsongLi

docs/content/append-table/query-performance.md
```diff
@@ -64,6 +64,7 @@ scenario. Using a bitmap may consume more space but can result in greater accura
 
 `Bitmap`:
 * `file-index.bitmap.columns`: specify the columns that need bitmap index.
+* `file-index.bitmap.<column_name>.index-block-size`: to config secondary index block size, default value is 16kb.
```
JingsongLi, line 67: Maybe here should link to `fileindex.md`, here just say, "hey, we have the bitmap index, see more information in xxx link."

docs/content/concepts/spec/fileindex.md
```diff
@@ -98,9 +98,71 @@ This class use (64-bits) long hash. Store the num hash function (one integer) an
 
 Define `'file-index.bitmap.columns'`.
 
+Bitmap file index format (V2):
```
JingsongLi, line 101: Document `file-index.bitmap.version`, old version readers can not read this v2 bitmap index, so if the user want to use old version, he/she should know how to configure this.

JingsongLi:
> I found another question that needs to be discussed. Should the index version default to v1 or v2? @Tan-JiaLiang @JingsongLi

I don't have a clear idea, although it may cause compatibility issues, bitmap should not be used too much. Maintaining the default version to 1 is a relatively conservative choice.

paimon-common/src/main/java/org/apache/paimon/fileindex/bitmap/BitmapFileIndex.java
```diff
@@ -49,39 +49,51 @@
 public class BitmapFileIndex implements FileIndexer {
 
     public static final int VERSION_1 = 1;
+    public static final int VERSION_2 = 2;
+
+    public static final String VERSION = "version";
+    public static final String INDEX_BLOCK_SIZE = "index-block-size";
+    public static final String ENABLE_BUFFERED_INPUT = "enable-buffered-input";
+    public static final String ENABLE_NEXT_OFFSET_TO_SIZE = "enable-next-offset-to-size";
```
JingsongLi, line 57: What is enable next offset to size? Under what circumstances can it not be size?

paimon-common/src/main/java/org/apache/paimon/fileindex/bitmap/BitmapFileIndexMetaV2.java
```diff
@@ -0,0 +1,415 @@
...
+ * ｜ serialized bitmap 1                            ｜
+ * +-------------------------------------------------+
+ * ｜ serialized bitmap 2                            ｜
+ * +-------------------------------------------------+  BITMAP BLOCKS
+ * ｜ serialized bitmap 3                            ｜
+ * +-------------------------------------------------+
+ * ｜ ...                                            ｜
+ * +-------------------------------------------------+-----------------
+ *
+ * index block format:
```
JingsongLi, line 88: Maybe we should have ability to compress index block?

paimon-common/src/main/java/org/apache/paimon/fileindex/bitmap/BitmapFileIndexMetaV2.java
```diff
@@ -0,0 +1,415 @@
...
+        int offset;
+        int serializedBytes = Integer.BYTES;
+        List<Entry> entryList;
+        Function<Object, Integer> keyBytesMapper;
+        DataType dataType;
+        SeekableInputStream seekableInputStream;
+        Options options;
+
+        void tryDeserialize() {
+            if (entryList == null) {
```
JingsongLi, line 349: Maybe you can take a look to https://cwiki.apache.org/confluence/display/PAIMON/PIP-25%3A+Introduce+a+key-value+file+format+for+paimon+primary+key+table

Here just something like LevelDB sst block, we can store length to avoid deserialize all entries.

docs/content/append-table/query-performance.md
hang8929201: done

PR #5029, opened 2025-02-06
[hive]Fix hive catalog create partitions using table location.

zhuangchong:
### Purpose

Fix hive catalog create partitions using table location.


JingsongLi (commented):
Can you add a test case for this fixing?

zhuangchong:
> Can you add a test case for this fixing?

done.

JingsongLi:
+1

PR #5030, opened 2025-02-06
[doc]Fix the output error of Batch Reading from the Fallback Branch example

chenxinwei:
### Documentation
Simply fixes the output error of Batch Reading from the Fallback Branch example.


yuzelin:
`scan.fallback-branch` means if the partition exists in main branch, read from the main branch. Otherwise, read from the fallback branch. The `20240725` result shouldn't contains `peach` because paimon reader only read main branch for this partition.

chenxinwei:
> `scan.fallback-branch` means if the partition exists in main branch, read from the main branch. Otherwise, read from the fallback branch. The `20240725` result shouldn't contains `peach` because paimon reader only read main branch for this partition.

Got it. Thanks for your explanation.

PR #5034, opened 2025-02-07
[core] Check that all fields with aggregate functions in partial-update should be protected by sequence-group

yuzelin:
### Purpose

For example,
```
CREATE TABLE SG
(
    pk   INT,
    f0   INT,
    g0   INT,
    f1   INT,
    g1   INT,
    PRIMARY KEY (pk) NOT ENFORCED
) WITH (
      'merge-engine' = 'partial-update',
      'fields.g0.sequence-group' = 'f0',
      'fields.f0.aggregate-function' = 'listagg',
      'fields.f1.aggregate-function' = 'listagg'
);
```
The `f1` lack sequence-group but there's no exception.


zhuangchong (approved):
+1

JingsongLi (approved):
+1

PR #4983, opened 2025-01-22
[core] Support DLF auth

jerry-024:
### Purpose
Support DLF auth in RESTCatalog
Linked issue: https://github.com/apache/paimon/issues/4540
### Tests
- AuthSessionTest.testCreateDLFAuthProviderByAk
- AuthSessionTest.testCreateDLFAuthProviderByFileNoDefineRefresh
- AuthSessionTest.testDLFAuthProviderAuthHeader
- AuthSessionTest.testCreateDLFAuthProviderWithoutNeedConf

### API and Format
No

PR #5044, opened 2025-02-10
[core] Fix missing kotlin and okio dependencies when using okhttp3.

LinMingQiang:
### Purpose

Linked issue: close #xxx

### DIFF
--- a/paimon-core/pom.xml
+++ b/paimon-core/pom.xml
@@ -81,7 +81,7 @@ under the License.
 
         <dependency>
             <groupId>com.squareup.okio</groupId>
-            <artifactId>okio-jvm</artifactId>
+            <artifactId>okio</artifactId>
             <version>${okio.version}</version>
         </dependency>
 
@@ -298,7 +298,7 @@ under the License.
                             <artifactSet>
                                 <includes combine.children="append">
                                     <include>com.squareup.okhttp3:okhttp</include>
-                                    <include>com.squareup.okio:okio-jvm</include>
+                                    <include>com.squareup.okio:okio</include>
                                     <include>org.jetbrains.kotlin:kotlin-stdlib</include>
                                 </includes>
                             </artifactSet>
--- a/paimon-core/pom.xml
+++ b/paimon-core/pom.xml
@@ -75,13 +75,13 @@ under the License.
 
         <dependency>
             <groupId>org.jetbrains.kotlin</groupId>
-            <artifactId>kotlin-stdlib</artifactId>
+            <artifactId>kotlin-stdlib-jdk8</artifactId>
             <version>${kotlin.version}</version>
         </dependency>
 
         <dependency>
             <groupId>com.squareup.okio</groupId>
-            <artifactId>okio</artifactId>
+            <artifactId>okio-jvm</artifactId>
             <version>${okio.version}</version>
         </dependency>
 
@@ -298,8 +298,8 @@ under the License.
                             <artifactSet>
                                 <includes combine.children="append">
                                     <include>com.squareup.okhttp3:okhttp</include>
-                                    <include>com.squareup.okio:okio</include>
-                                    <include>org.jetbrains.kotlin:kotlin-stdlib</include>
+                                    <include>com.squareup.okio:okio-jvm</include>
+                                    <include>org.jetbrains.kotlin:kotlin-stdlib-jdk8</include>
                                 </includes>
                             </artifactSet>
                             <relocations>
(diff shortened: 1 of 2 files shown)


paimon-core/pom.xml
```diff
@@ -284,13 +298,23 @@ under the License.
                             <artifactSet>
                                 <includes combine.children="append">
                                     <include>com.squareup.okhttp3:okhttp</include>
+                                    <include>com.squareup.okio:okio-jvm</include>
```
JingsongLi, line 301: Please modify NOTICE file.

jerry-024:
The reason is okio‘s version so just add okio is ok.

paimon-core/pom.xml
LinMingQiang: I can't find the NOTICE file in `okio-jvm`

paimon-core/pom.xml
```diff
@@ -71,6 +73,12 @@ under the License.
             <version>${okhttp.version}</version>
         </dependency>
 
+        <dependency>
+            <groupId>com.squareup.okio</groupId>
+            <artifactId>okio-jvm</artifactId>
+            <version>${okio.version}</version>
+        </dependency>
+
```
jerry-024, line 81:
```
        <dependency>
            <groupId>com.squareup.okio</groupId>
            <artifactId>okio</artifactId>
            <version>${okio.version}</version>
        </dependency>
```
just add this is ok.

paimon-core/pom.xml
JingsongLi: image

paimon-core/pom.xml
LinMingQiang: ok, I'll try it first.

paimon-core/pom.xml
LinMingQiang: ok, I will modify it later.

jerry-024: you need add content in /paimon/paimon-core/src/main/resources/META-INF/NOTICE

paimon-core/pom.xml
LinMingQiang: Still : java.lang.ClassNotFoundException: kotlin.jvm.internal.Intrinsics

Now. I'll try shade okio.

> ```
>         <dependency>
>             <groupId>com.squareup.okio</groupId>
>             <artifactId>okio</artifactId>
>             <version>${okio.version}</version>
>         </dependency>
> ```
> 
> just add this is ok.

jerry-024: I mean add okio, but still need to shade it. Because there is other version okio.

LinMingQiang:
> The reason is okio‘s version so just add okio is ok.

I have tried, both `okio` and `kotlin-stdlib` need to be added.

jerry-024:
> > The reason is okio‘s version so just add okio is ok.
> 
> I have tried, both `okio` and `kotlin-stdlib` need to be added.

Can you write how to trigger this problem?

LinMingQiang:
> > > The reason is okio‘s version so just add okio is ok.
> > 
> > 
> > I have tried, both `okio` and `kotlin-stdlib` need to be added.
> 
> Can you write how to trigger this problem?

I used the partition mark done action of `http-report`.

jerry-024:
try this  [change](https://github.com/apache/paimon/pull/4983/files#diff-e842b4c37014991e5eb088817c3be19ac8ecea40d5813a3475d331261bb6e033)

LinMingQiang:
> try this [change](https://github.com/apache/paimon/pull/4983/files#diff-e842b4c37014991e5eb088817c3be19ac8ecea40d5813a3475d331261bb6e033)

Still : ClassNotFoundException: kotlin.jvm.internal.Intrinsics.

paimon-core/pom.xml
LinMingQiang: Added

jerry-024:
If you find `kotlin-stdlib` must need. According to okhttp's pom maybe we need use kotlin-stdlib-jdk8: 1.8.21 is better.

LinMingQiang:
> If you find `kotlin-stdlib` must need. According to okhttp's pom maybe we need use kotlin-stdlib-jdk8: 1.8.21 is better.

I get `java.lang.ClassNotFoundException: org.apache.paimon.shade.kotlin.jvm.internal.Intrinsics` in the UT , the reason is  `okio-jvm` and `kotlin-stdlib`  will not be shaded into `paimon-core`, so i did not relocate `okio-jvm` and `kotlin-stdlib` in the paimon-core.

PR #5046, opened 2025-02-10
[filesystems] Listing OSS objects with metadata

### LINKED ISSUE #4990: [Feature] OSSFileIO return object metadata
### Search before asking

### Motivation

Fill metadata fields of `FileStatus` returned by `OSSFileIO`.

### Solution

Fill metadata fields of `FileStatus` returned by `OSSFileIO`.

### Anything else?

_No response_

smdsbz:
### Purpose

Linked issue: close #4990 

Expose OSS objects' metadata, so more fields of object tables built on OSS can be filled.

### Tests

N/A

### API and Format

N/A

### Documentation

N/A

### DIFF
--- a/docs/layouts/shortcodes/generated/catalog_configuration.html
+++ b/docs/layouts/shortcodes/generated/catalog_configuration.html
@@ -87,6 +87,12 @@ under the License.
             <td>Whether to allow static cache in file io implementation. If not allowed, this means that there may be a large number of FileIO instances generated, enabling caching can lead to resource leakage.</td>
         </tr>
         <tr>
+            <td><h5>file-io.populate-meta</h5></td>
+            <td style="word-wrap: break-word;">false</td>
+            <td>Boolean</td>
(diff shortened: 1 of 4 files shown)


smdsbz:
Currently the implementation __always__ fetches metadata when retrieving file statuses, this could slow down the operation of normal tables since the metadata is only used by object tables.

I'm thinking adding `FileIO#enableFeature(String)`, `FileIO#disableFeature(String)` to allow on-the-fly tuning of FileIO behaviors. What's your opinions?

JingsongLi:
> Currently the implementation **always** fetches metadata when retrieving file statuses, this could slow down the operation of normal tables while the metadata is only used by object tables.
> 
> I'm thinking adding `FileIO#enableFeature(String)`, `FileIO#disableFeature(String)` to allow on-the-fly tuning of FileIO behaviors. What's your opinions?

Yes, I think we should disable this by default. Option can be in catalog option.

JingsongLi (approved):
+1

PR #5050, opened 2025-02-11
[core] Fix missing kotlin and okio dependencies when using okhttp3. 

LinMingQiang:
### Purpose

Relation pr : https://github.com/apache/paimon/pull/5044  (this pr can't be reopen because force push.)

When I used the partition mark done action of `http-report` : 
`java. lang.NoClassDefFoundError: okio/Options`
`java. lang.NoClassDefFoundError: kotlin/jvm/internal/Intrinsics`

So we need to shade `okio-jvm` and `kotlin-stdlib`  too.

Linked issue: close #xxx

### DIFF
--- a/paimon-core/pom.xml
+++ b/paimon-core/pom.xml
@@ -81,7 +81,7 @@ under the License.
 
         <dependency>
             <groupId>com.squareup.okio</groupId>
-            <artifactId>okio-jvm</artifactId>
+            <artifactId>okio</artifactId>
             <version>${okio.version}</version>
         </dependency>
 
@@ -298,7 +298,7 @@ under the License.
                             <artifactSet>
                                 <includes combine.children="append">
                                     <include>com.squareup.okhttp3:okhttp</include>
-                                    <include>com.squareup.okio:okio-jvm</include>
+                                    <include>com.squareup.okio:okio</include>
                                     <include>org.jetbrains.kotlin:kotlin-stdlib</include>
                                 </includes>
                             </artifactSet>
--- a/paimon-core/pom.xml
+++ b/paimon-core/pom.xml
@@ -33,8 +33,6 @@ under the License.
 
     <properties>
         <frocksdbjni.version>6.20.3-ververica-2.0</frocksdbjni.version>
-        <kotlin.version>1.8.21</kotlin.version>
-        <okio.version>3.6.0</okio.version>
     </properties>
 
     <dependencies>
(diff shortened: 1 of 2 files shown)


LinMingQiang:
@jerry-024 Here,  it cannot be 'okio', it can only be 'okio-jvm', otherwise ut will not pass.

LinMingQiang:
This is the final version, production test or ut has passed.

wwj6591812:
+1

paimon-core/pom.xml
```diff
@@ -284,13 +284,23 @@ under the License.
                             <artifactSet>
                                 <includes combine.children="append">
                                     <include>com.squareup.okhttp3:okhttp</include>
+                                    <include>com.squareup.okio:okio-jvm</include>
+                                    <include>org.jetbrains.kotlin:kotlin-stdlib</include>
```
jerry-024, line 288: Could we use kotlin-stdlib-jdk8?

jerry-024:
And we need't add okio-jvm and kotlin-stdlib-jdk8 as dependency?

paimon-core/pom.xml
LinMingQiang: It cant be `kotlin-stdlib-jdk8`, because its empty.

LinMingQiang:
> Need't add okio-jvm and kotlin-stdlib-jdk8 as dependency?

No need.

paimon-core/src/main/resources/META-INF/NOTICE
```diff
@@ -5,4 +5,6 @@ This product includes software developed at
 The Apache Software Foundation (http://www.apache.org/).
 
 This project bundles the following dependencies under the Apache Software License 2.0 (http://www.apache.org/licenses/LICENSE-2.0.txt)
-- com.squareup.okhttp3:okhttp:4.12.0
\ No newline at end of file
+- com.squareup.okhttp3:okhttp:4.12.0
+- com.squareup.okio:okio-jvm:3.6.0
+- org.jetbrains.kotlin:kotlin-stdlib:1.8.21
```
jerry-024, line 10: If we don't need add okio-jvm and kotlin-stdlib as dependency. Needn't update this file.

LinMingQiang: It still bundles `okio-jvm` and `kotlin-stdlib` dependencies in this project, It's just that these two dependencies are introduced by `okhttp3`.

jerry-024:
LGTM

JingsongLi:
+1

PR #5051, opened 2025-02-11
[iceberg][fix] obtain FileIO from the absolute warehouse path instead of relative table path

LsomeYeah:
### Purpose

Linked issue: close #xxx

```
Path path =
        new Path(
                String.format(
                        "%s/%s/metadata",
                        icebergIdentifier.getDatabaseName(),
                        icebergIdentifier.getTableName()));
try {
    fileIO = FileIO.get(path, CatalogContext.create(icebergOptions));
} catch (IOException e) {
}
```
If obtain FileIO with the code below, it'll always get LocalFileIO because `path.toUri().getScheme() == null` will always be true. So we should use absolute warehouse path to get FileIO.

### Tests

IcebergMigrateHadoopMetadataTest#testGetIcebergMetadataWithCustomFileIO


wwj6591812:
+1

PR #5054, opened 2025-02-11
[fileio] Add file-io.allow-cache for RESTTokenFileIO

JingsongLi:
### Purpose

1. Add file-io.allow-cache for RESTTokenFileIO, file io should be closed in RESTTokenFileIO, so it can not be cached.
2. Improve `HadoopCompliantFileIO`, it should use `Map.computeIfAbsent`.
3. We should use `new Configuration(false)` to not load files in configuration for improving performance.

### DIFF
--- a/docs/layouts/shortcodes/generated/catalog_configuration.html
+++ b/docs/layouts/shortcodes/generated/catalog_configuration.html
@@ -81,6 +81,12 @@ under the License.
             <td>Configure the size of the connection pool.</td>
         </tr>
         <tr>
+            <td><h5>file-io.allow-cache</h5></td>
+            <td style="word-wrap: break-word;">true</td>
+            <td>Boolean</td>
+            <td>Whether to allow static cache in file io implementation. If not allowed, this means that there may be a large number of FileIO instances generated, enabling caching can lead to resource leakage.</td>
+        </tr>
+        <tr>
             <td><h5>format-table.enabled</h5></td>
             <td style="word-wrap: break-word;">true</td>
             <td>Boolean</td>
@@ -117,6 +123,12 @@ under the License.
             <td>Metastore of paimon catalog, supports filesystem, hive and jdbc.</td>
         </tr>
         <tr>
+            <td><h5>resolving-file-io.enabled</h5></td>
+            <td style="word-wrap: break-word;">false</td>
+            <td>Boolean</td>
+            <td>Whether to enable resolving fileio, when this option is enabled, in conjunction with the table's property data-file.external-paths, Paimon can read and write to external storage paths, such as OSS or S3. In order to access these external paths correctly, you also need to configure the corresponding access key and secret key.</td>
+        </tr>
+        <tr>
             <td><h5>sync-all-properties</h5></td>
             <td style="word-wrap: break-word;">false</td>
             <td>Boolean</td>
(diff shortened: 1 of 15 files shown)


docs/layouts/shortcodes/generated/catalog_configuration.html
```diff
@@ -116,6 +122,12 @@
             <td>String</td>
             <td>Metastore of paimon catalog, supports filesystem, hive and jdbc.</td>
         </tr>
+        <tr>
+            <td><h5>resolving-file-io.enabled</h5></td>
+            <td style="word-wrap: break-word;">false</td>
+            <td>Boolean</td>
+            <td>Whether to enable resolving fileio, when this option is enabled, in conjunction with the table's property data-file.external-paths, Paimon can read and write to external storage paths, such as OSS or S3. In order to access these external paths correctly, you also need to configure the corresponding access key and secret key.</td>
```
jerry-024, line 129: fileio -> file io same with line 87?

paimon-filesystems/paimon-oss-impl/src/main/java/org/apache/paimon/oss/OSSFileIO.java
```diff
@@ -124,7 +134,22 @@ protected FileSystem createFileSystem(org.apache.hadoop.fs.Path path) {
...
+            return CACHE.computeIfAbsent(
+                    new CacheKey(hadoopOptions, scheme, authority), key -> supplier.get());
+        } else {
+            return supplier.get();
+        }
+    }
+
+    @Override
+    public void close() {
+        if (!allowCache) {
```
jerry-024, line 149: This means, if allowCache, the fileIo need to close by outside?

docs/layouts/shortcodes/generated/catalog_configuration.html
```diff
@@ -80,6 +80,12 @@
             <td>Integer</td>
             <td>Configure the size of the connection pool.</td>
         </tr>
+        <tr>
+            <td><h5>file-io.allow-cache</h5></td>
+            <td style="word-wrap: break-word;">true</td>
+            <td>Boolean</td>
+            <td>Whether to allow static cache in file io implementation. If not allowed, this means that there may be a large number of FileIO instances generated, enabling caching can lead to resource leakage.</td>
```
wwj6591812, line 87: I have a question here:
If enabling caching, the number of FileIO instances will be reduce. Why enabling caching can lead to resource leakage?

paimon-filesystems/paimon-oss-impl/src/main/java/org/apache/paimon/oss/OSSFileIO.java
JingsongLi: if not allow cache, the fileio should be closed.

If allow cache, the fileio is in the static cache, we cannot close them.

docs/layouts/shortcodes/generated/catalog_configuration.html
JingsongLi: it is ok in description, we can just be careful for option key, it is API part.

docs/layouts/shortcodes/generated/catalog_configuration.html
JingsongLi: Our original design was to place it in the cache of static variables, but for the case of REST Catalog, it may have a lot of FileIO generated, and there may be a FileIO for each table because each table has different file access permissions.

So if there are too many FileIOs in this situation, we cannot cache them in memory casually. If there are too many, it will lead to too many resources, that is, resource leakage.

jerry-024:
LGTM

PR #5058, opened 2025-02-11
[Core][Bug] Fix bug when read changelog's IndexManifest.

wwj6591812:
### Purpose
We encountered an exception in our production job.
image

After investigate this issue, the reason of this exception is :
We open `Changelog` decouple, the `Changelog` has records the `indexManifest`. But the `Snapshot` of the `Changelog` has been expired. So when read this `Changelog` will read the non-existed `indexManifest`.

We have think two functions to solve this problem.
1、 Don't read the `indexManifest` when streaming read the `Changelog`.
2、Override the `indexManifest` method in `Changelog`, which return null forever.

Due to the design of `Changelog` and `Tag` is extends `Snapshot`, they should not modify the field of `Snapshot`. So after discuss we choose the function 1.

Linked issue: close #xxx


JingsongLi:
Can you add a test case?

wwj6591812:
> Can you add a test case?

Done，Please CC, Thx.

JingsongLi (approved):
+1

PR #5060, opened 2025-02-11
[hotfix] Improve the HiveCatalogITCaseBase class to reduce the number of hive queries

zhuangchong:
### Purpose

Improve the HiveCatalogITCaseBase class to reduce the number of hive queries.

image


JingsongLi:
+1 thanks @zhuangchong

PR #5063, opened 2025-02-11
[core] Public RESTTokenFileIO.validToken for native engines to get the token

JingsongLi:
### Purpose

C++ native engines can use this token to use C++ reader/writer to use data token from REST Server.


jerry-024:
LGTM

PR #5065, opened 2025-02-12
[core] Fix that AggregateMergeFunction handles multiple sequence fields mistakenly

yuzelin:
### Purpose

For example, for sequence fields 's0,s1', if input (null, 1), (1, null), the result will be (1, 1) currently which is wrong.


paimon-core/src/main/java/org/apache/paimon/mergetree/compact/aggregate/AggregateMergeFunction.java
```diff
@@ -149,8 +150,8 @@ public MergeFunction<KeyValue> create(@Nullable int[][] projection) {
 
         private String getAggFuncName(String fieldName, List<String> sequenceFields) {
             if (sequenceFields.contains(fieldName)) {
-                // no agg for sequence fields, use last_non_null_value to do cover
-                return FieldLastNonNullValueAggFactory.NAME;
```
JingsongLi, line 153: This is interface...

PR #5069, opened 2025-02-13
[cdc] Support audit time computed clumns during cdc sync database

### LINKED ISSUE #5068: [Feature] Support special audition computed columns, especially for cdc sync database
### Search before asking

### Motivation

When doing cdc sync database with flink action, that would be helpful if the ingest times are recorded. 
Ingest times means when the record are created(+I), and when the record are updated (mainly+U, -D, but +I also does). 

### Solution

To do so, two special computed columns could be empolyeed. One could be `etl_create_time`, only set when rowkind is +I. Another could be `etl_update_time`, set whenever new records come in. 

And to achieve that, a new special expression should be introduced too.  The expression should compute the time according to rowkind.

* If it's used for `etl_create_time`, only returns the current timestamp when rowkind is '+I', otherwise null.
* If it's used for `etl_update_time` returns the current timestamp whenever called. 

### Anything else?

Another solution is recording the ingest times in audit_log. (refer to https://github.com/apache/paimon/issues/5022)
But that would affect paimon-core, and break some assumption.

JackeyLee007:
### Purpose

Linked issue: close #5068

Support special audition computed columns, especially for cdc sync database

### Tests

org.apache.paimon.flink.action.cdc.kafka.KafkaCanalSyncDatabaseActionITCase#testAuditTime


JackeyLee007:
@wwj6591812 Thanks for your review yesterday. I re-compose a new clean PR #5086.
If possible, have a look please.

PR #5076, opened 2025-02-13
[flink] Support parallelismConfigurable

yunfengzhou-hub:
### Purpose

In Flink 1.17+, the parameter `parallelismConfigurable` is introduced to support specifying operator's initial parallelism while dynamically changing parallelism in some situations (like in AQE). This PR adds this parameter to Paimon Flink operators where parallelism does not affect correctness, and can be changed by Flink infra according to detailed needs.

Related PR: #4975

### Tests

Flink batch execution mode enables AQE by default, so existing batch tests like BatchFileStoreITCase can verify the correctness of these changes.

### API and Format

### Documentation

This change does not introduce a new feature that needs documentation.


yuzelin (approved):
+1

JingsongLi (approved):
+1

PR #5088, opened 2025-02-14
[spark] Add max_pt function

ulysses-you:
### Purpose

Adds a new Spark function `max_pt`. It accpets a string type literal and return a max-valid-toplevel partition value.
- `valid` means the partition contains data files
- `toplevel` means only return the first partition value if the table has multi-partition columns

`max_pt` will throw exception when:
- the table is not a partitioned table
- the partitioned table does not have partition
- all of the partitions do not contains data files

This pr adds `max_pt` through spark v2 function catalog using a fake scalar function and then adds a rule to replace `max_pt` to a literal value during analysis.

Example:
```sql
SELECT max_pt('t')
=>
20250101
```

### API and Format

no

### Documentation

add `sql-functions` docs

### DIFF
--- a/docs/content/spark/sql-functions.md
+++ b/docs/content/spark/sql-functions.md
@@ -1,0 +1,55 @@
+---
+title: "SQL Functions"
+weight: 2
+type: docs
+aliases:
+- /spark/sql-functions.html
+---
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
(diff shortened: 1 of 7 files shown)


JingsongLi (commented):
Maybe add a new documentation: `sql-function`?

ulysses-you:
@JingsongLi added docs

JingsongLi (approved):
+1

PR #5090, opened 2025-02-14
[flink] Support specifying what partitions to scan in Flink

tsreaper:
### Purpose

Lookup joins in streaming SQL is the same as normal joins in batch SQL. However, when specifying what partitions to scan in lookup joins, currently user can specify `max_pt()` through SQL hints to read the latest partition, without specifying a fixed partition. Such SQL hint is not supported in batch joins.

Paimon is a streaming-batch unified lake format. To also support streaming-batch unification in SQL, this PR introduces a new option `scan.partitions`, which accepts both `max_pt()` (in lookup joins) and fixed partitions (in all joins). Users only need to change the value of this option to specify different partitions for streaming and batch jobs, and they don't need to change SQL itself.

### Tests

Unit tests and IT cases.

### API and Format

No format changes.

### Documentation

Document is also added.

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/PartitionLoader.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/PartitionLoader.java
@@ -1,0 +1,139 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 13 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/PartitionLoader.java
```diff
@@ -0,0 +1,139 @@
...
+
+/** Specify partitions for lookup tables. */
+public abstract class PartitionLoader implements Serializable {
+
+    private static final long serialVersionUID = 1L;
+
+    private static final String MAX_PT = "max_pt()";
+    private static final String MAX_TWO_PT = "max_two_pt()";
+
+    protected final FileStoreTable table;
```
wwj6591812, line 50: Why use FileStoreTable instead of Table?

tsreaper: We need specific methods like `schema()` in its subclass.

JingsongLi (approved):
+1

PR #5096, opened 2025-02-17
[hotfix] Fix earliestSnapshot stability

yunfengzhou-hub:
### Purpose

This PR fixes earliestSnapshot test case failures.
* Improve earliestSnapshot retry number according to latest snapshot id
* Fix SnapshotManager#earlierThanTimeMills return value when earliest snapshot cannot be found
* Fix test case expected result according to invocation time.

### Tests

SnapshotManagerTest


JingsongLi:
+1

PR #4859, opened 2025-01-08
[Bug] Add the dependency of RoaringBitmap in paimon-micro-benchmarks.

Cribbee:
### Purpose
Fix the compilation error when packaging paimon-1.1 in moudle “paimon-micro-benchmarks”.
```shell
[INFO] --- maven-compiler-plugin:3.8.0:testCompile (default-testCompile) @ paimon-micro-benchmarks ---
[INFO] Compiling 11 source files to /paimon/paimon-benchmark/paimon-micro-benchmarks/target/test-classes
[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /paimon/paimon-benchmark/paimon-micro-benchmarks/src/test/java/org/apache/paimon/benchmark/bitmap/RoaringBitmapBenchmark.java:[26,25] package org.roaringbitmap does not exist
[INFO] 1 error
```

PR #5098, opened 2025-02-17
[pom] Fix RoaringBitmap dependency problem in partial building

JingsongLi:
### Purpose

See https://github.com/apache/paimon/pull/4859


yuzelin:
+1

PR #5101, opened 2025-02-18
[spark] AbstractSparkInternalRow supports fallback equals and hashCode

ulysses-you:
### Purpose

The `AbstractSparkInternalRow` would be passed to Spark, so it is dangerous if it would throw exception when call equals or hashcode. This pr make it fallback to suport if there is an exception.

### Tests

Pass CI

### API and Format

no

### Documentation

no

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/utils/InternalRowUtils.java
+++ b/paimon-common/src/main/java/org/apache/paimon/utils/InternalRowUtils.java
@@ -62,28 +62,26 @@ public class InternalRowUtils {
         }
         if (data1 != null) {
             if (data1 instanceof InternalRow) {
-                int i = 0;
                 RowType rowType = (RowType) dataType;
                 int len = rowType.getFieldCount();
-                while (i < len) {
+                for (int i = 0; i < len; i++) {
                     Object value1 = get((InternalRow) data1, i, rowType.getTypeAt(i));
                     Object value2 = get((InternalRow) data2, i, rowType.getTypeAt(i));
                     if (!equals(value1, value2, rowType.getTypeAt(i))) {
                         return false;
                     }
-                    i += 1;
                 }
             } else if (data1 instanceof InternalArray) {
-                int i = 0;
+                if (((InternalArray) data1).size() != ((InternalArray) data2).size()) {
+                    return false;
+                }
                 ArrayType arrayType = (ArrayType) dataType;
-                int len = ((InternalArray) data1).size();
-                while (i < len) {
+                for (int i = 0; i < ((InternalArray) data1).size(); i++) {
                     Object value1 = get((InternalArray) data1, i, arrayType.getElementType());
                     Object value2 = get((InternalArray) data2, i, arrayType.getElementType());
                     if (!equals(value1, value2, arrayType.getElementType())) {
                         return false;
                     }
-                    i += 1;
                 }
             } else if (data1 instanceof InternalMap) {
                 if (((InternalMap) data1).size() != ((InternalMap) data2).size()) {
@@ -107,7 +105,6 @@ public class InternalRowUtils {
                                     mapType.getKeyType(),
                                     mapType.getValueType());
                 }
-
                 InternalArray keyArray1 = map1.keyArray();
                 for (int i = 0; i < map1.size(); i++) {
                     Object key = get(keyArray1, i, mapType.getKeyType());
@@ -144,21 +141,17 @@ public class InternalRowUtils {
         int result = 0;
         if (data instanceof InternalRow) {
             RowType rowType = (RowType) dataType;
-            int i = 0;
             int len = rowType.getFieldCount();
-            while (i < len) {
+            for (int i = 0; i < len; i++) {
                 Object v = get((InternalRow) data, i, rowType.getTypeAt(i));
                 result = 37 * result + hash(v, rowType.getTypeAt(i));
-                i += 1;
             }
         } else if (data instanceof InternalArray) {
             ArrayType arrayType = (ArrayType) dataType;
-            int i = 0;
             int len = ((InternalArray) data).size();
-            while (i < len) {
+            for (int i = 0; i < len; i++) {
                 Object v = get((InternalArray) data, i, arrayType.getElementType());
                 result = 37 * result + hash(v, arrayType.getElementType());
-                i += 1;
             }
         } else if (data instanceof InternalMap) {
             MapType mapType = (MapType) dataType;
--- a/paimon-common/src/main/java/org/apache/paimon/utils/InternalRowUtils.java
+++ b/paimon-common/src/main/java/org/apache/paimon/utils/InternalRowUtils.java
@@ -48,12 +48,141 @@ import javax.annotation.Nullable;
 import java.math.BigDecimal;
 import java.math.RoundingMode;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
 /** Utils for {@link InternalRow} structures. */
 public class InternalRowUtils {
+
+    public static boolean equals(Object data1, Object data2, DataType dataType) {
+        if ((data1 == null) != (data2 == null)) {
+            return false;
+        }
+        if (data1 != null) {
+            if (data1 instanceof InternalRow) {
+                int i = 0;
+                RowType rowType = (RowType) dataType;
+                int len = rowType.getFieldCount();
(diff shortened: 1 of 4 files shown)


paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/AbstractSparkInternalRow.java
```diff
@@ -251,11 +251,21 @@ public boolean equals(Object o) {
             return false;
         }
         AbstractSparkInternalRow that = (AbstractSparkInternalRow) o;
-        return Objects.equals(rowType, that.rowType) && Objects.equals(row, that.row);
+        try {
+            return Objects.equals(rowType, that.rowType) && Objects.equals(row, that.row);
+        } catch (Exception e) {
+            // The underlying row may not support equals or hashcode, e.g., `ProjectedRow`,
+            // to be safe, we fallback to super.
+            return super.equals(o);
```
JingsongLi, line 259: The behavior after fallback is not correct, it only compares instances.

ulysses-you: This is mainly used to avoid exception. If we want to make behavior correct, how about doing a deep copy and then compare ?

JingsongLi: Maybe we should implement a method to equals with row type (invoke row.getXXX).

JingsongLi: Maybe introduce `InternalRowUtils.equals(InternalRow row1, InternalRow row2, RowType rowType)`

ulysses-you: addressed

paimon-common/src/main/java/org/apache/paimon/utils/InternalRowUtils.java
```diff
@@ -55,6 +55,56 @@
...
+                } else if (data1 instanceof Float && java.lang.Float.isNaN((Float) data1)) {
+                    if (!java.lang.Float.isNaN((Float) data2)) {
+                        return false;
+                    }
+                } else if (data1 instanceof Double && java.lang.Double.isNaN((Double) data1)) {
+                    if (!java.lang.Double.isNaN((Double) data2)) {
+                        return false;
+                    }
+                } else {
+                    if (!data1.equals(data2)) {
```
JingsongLi, line 82: If it is a complex structure, it should use Utils.equals too.

paimon-common/src/main/java/org/apache/paimon/utils/InternalRowUtils.java
```diff
@@ -145,6 +206,10 @@ public static Object copy(Object o, DataType type) {
                 return copyMap(
                         (InternalMap) o, ((MultisetType) type).getElementType(), new IntType());
             }
+        } else if (o instanceof byte[]) {
+            byte[] copy = new byte[((byte[]) o).length];
+            System.arraycopy(((byte[]) o), 0, copy, 0, ((byte[]) o).length);
+            return copy;
```
ulysses-you, line 212: not related to this pr in fact. just a kind of bug..

paimon-common/src/main/java/org/apache/paimon/utils/InternalRowUtils.java
```diff
@@ -48,13 +48,74 @@
...
 
 /** Utils for {@link InternalRow} structures. */
 public class InternalRowUtils {
 
+    public static boolean equals(Object data1, Object data2, DataType dataType) {
+        if ((data1 == null) != (data2 == null)) {
+            return false;
+        }
+        if (data1 != null) {
+            if (data1 instanceof InternalRow) {
```
ulysses-you, line 64: @JingsongLi  both array and map have their own equals and hashcode, so I leave it to use o.equals

JingsongLi: But they are not safe, if there is a row inside array or map, it is not safe too.

So it is better to add implementation for array and map too.

ulysses-you: addressed

paimon-common/src/main/java/org/apache/paimon/utils/InternalRowUtils.java
```diff
@@ -48,13 +48,142 @@
...
+                int i = 0;
+                RowType rowType = (RowType) dataType;
+                int len = rowType.getFieldCount();
+                while (i < len) {
```
JingsongLi, line 68: Modify this loop to `for`.

paimon-common/src/main/java/org/apache/paimon/utils/InternalRowUtils.java
```diff
@@ -48,13 +48,142 @@
...
+                    if (!equals(value1, value2, rowType.getTypeAt(i))) {
+                        return false;
+                    }
+                    i += 1;
+                }
+            } else if (data1 instanceof InternalArray) {
+                int i = 0;
+                ArrayType arrayType = (ArrayType) dataType;
+                int len = ((InternalArray) data1).size();
+                while (i < len) {
```
JingsongLi, line 80: Modify this loop to for.

paimon-common/src/main/java/org/apache/paimon/utils/InternalRowUtils.java
```diff
@@ -48,13 +48,142 @@
...
+                    Object value2 = get((InternalRow) data2, i, rowType.getTypeAt(i));
+                    if (!equals(value1, value2, rowType.getTypeAt(i))) {
+                        return false;
+                    }
+                    i += 1;
+                }
+            } else if (data1 instanceof InternalArray) {
+                int i = 0;
+                ArrayType arrayType = (ArrayType) dataType;
+                int len = ((InternalArray) data1).size();
```
JingsongLi, line 79:
```
if (len not equal) {
    return false;
}
```

Please add tests for this.

paimon-common/src/main/java/org/apache/paimon/utils/InternalRowUtils.java
```diff
@@ -48,13 +48,142 @@
...
+                                    mapType.getKeyType(),
+                                    mapType.getValueType());
+                    map2 =
+                            copyToGenericMap(
+                                    (InternalMap) data2,
+                                    mapType.getKeyType(),
+                                    mapType.getValueType());
+                }
+
+                InternalArray keyArray1 = map1.keyArray();
```
JingsongLi, line 111:
```
if (len not equal) {
    return false;
}
```

Please add tests for this.

paimon-common/src/main/java/org/apache/paimon/utils/InternalRowUtils.java
```diff
@@ -48,13 +48,142 @@
...
+    public static int hash(Object data, DataType dataType) {
+        if (data == null) {
+            return 0;
+        }
+        int result = 0;
+        if (data instanceof InternalRow) {
+            RowType rowType = (RowType) dataType;
+            int i = 0;
+            int len = rowType.getFieldCount();
+            while (i < len) {
```
JingsongLi, line 149: Modify this loop to for.

paimon-common/src/main/java/org/apache/paimon/utils/InternalRowUtils.java
```diff
@@ -48,13 +48,142 @@
...
+                Object v = get((InternalRow) data, i, rowType.getTypeAt(i));
+                result = 37 * result + hash(v, rowType.getTypeAt(i));
+                i += 1;
+            }
+        } else if (data instanceof InternalArray) {
+            ArrayType arrayType = (ArrayType) dataType;
+            int i = 0;
+            int len = ((InternalArray) data).size();
+            while (i < len) {
```
JingsongLi, line 158: Modify this loop to for.

JingsongLi (commented):
Thanks @ulysses-you for your update!

Left comments.

ulysses-you:
thank you @JingsongLi  addressed

JingsongLi (approved):
Thanks @ulysses-you ! Looks good to me!

I believe equals and hash utils are very useful for other places too.

PR #5102, opened 2025-02-18
[core] Introduce Table.fileIO interface to public fileIO to file operation

JingsongLi:
### Purpose

In some cases, the computing engine needs to obtain FileIO to operate some files on its own, including metadata files and data files. Previously, the engine obtained FileIO by converting the table into a DataTable.

We can directly expose the Table FileIO interface.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/system/CatalogOptionsTable.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/system/CatalogOptionsTable.java
@@ -22,6 +22,8 @@ import org.apache.paimon.data.BinaryString;
 import org.apache.paimon.data.GenericRow;
 import org.apache.paimon.data.InternalRow;
 import org.apache.paimon.disk.IOManager;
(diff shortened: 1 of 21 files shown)


paimon-core/src/main/java/org/apache/paimon/table/system/CatalogOptionsTable.java
```diff
@@ -93,6 +95,12 @@ public List<String> primaryKeys() {
         return Collections.singletonList("key");
     }
 
+    @Override
+    public FileIO fileIO() {
+        // pass a useless file io, should never use this.
+        return new LocalFileIO();
+    }
```
jerry-024, line 102: Could throw UnsupportedOperationException here？

JingsongLi: I have considered this solution, but in compute engine side, maybe it will get FileIO and do something.

jerry-024:
LGTM

PR #5109, opened 2025-02-19
[Flink] dropStats in dedicated compactor jobs

wwj6591812:
### Purpose

In my company, user run a dedicated compactor job, but JobManager's memory is heavy.
Now add dropStats in dedicated compactor jobs.
image

Linked issue: close #xxx

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/CompactorSourceBuilder.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/CompactorSourceBuilder.java
@@ -86,24 +86,29 @@ public class CompactorSourceBuilder {
     }
 
     private Source<RowData, ?, ?> buildSource(CompactBucketsTable compactBucketsTable) {
-
         if (isContinuous) {
             compactBucketsTable = compactBucketsTable.copy(streamingCompactOptions());
             return new ContinuousFileStoreSource(
-                    compactBucketsTable.newReadBuilder().withFilter(partitionPredicate),
-                    compactBucketsTable.options(),
-                    null);
+                    getReadBuilder(compactBucketsTable), compactBucketsTable.options(), null);
         } else {
             compactBucketsTable = compactBucketsTable.copy(batchCompactOptions());
-            ReadBuilder readBuilder =
-                    compactBucketsTable.newReadBuilder().withFilter(partitionPredicate);
             Options options = compactBucketsTable.coreOptions().toConfiguration();
+
             return new StaticFileStoreSource(
-                    readBuilder,
+                    getReadBuilder(compactBucketsTable),
                     null,
                     options.get(FlinkConnectorOptions.SCAN_SPLIT_ENUMERATOR_BATCH_SIZE),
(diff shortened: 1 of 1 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/CompactorSourceBuilder.java
```diff
@@ -96,7 +95,7 @@ public CompactorSourceBuilder withPartitionIdleTime(@Nullable Duration partition
         } else {
             compactBucketsTable = compactBucketsTable.copy(batchCompactOptions());
             ReadBuilder readBuilder =
-                    compactBucketsTable.newReadBuilder().withFilter(partitionPredicate);
+                    compactBucketsTable.newReadBuilder().withFilter(partitionPredicate).dropStats();
```
JingsongLi, line 98: If (MANIFEST_DELETE_FILE_DROP_STATS is true) drop stats? Also work for `isContinuous`.

wwj6591812: Thx, done

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/CompactorSourceBuilder.java
```diff
@@ -86,20 +86,22 @@ public CompactorSourceBuilder withPartitionIdleTime(@Nullable Duration partition
     }
 
     private Source<RowData, ?, ?> buildSource(CompactBucketsTable compactBucketsTable) {
+        ReadBuilder readBuilder =
+                compactBucketsTable.newReadBuilder().withFilter(partitionPredicate);
```
JingsongLi, line 91: I mean here should be:
```
if (CoreOptions.fromMap(table.options()).manifestDeleteFileDropStats()) {
      readBuilder.dropStats();
}
```

wwj6591812: thx, done.

wwj6591812:
The failed ut(TimeoutException) seems not caused by my modify.

JingsongLi (changes requested):
What do you think about `manifest.delete-file-drop-stats`? Why do always avoiding let it work for batch compact?

wwj6591812:
The config `manifest.delete-file-drop-stats` used to provide compatibility when downstream reads the `Stats` from `Snapshot` that generated by `Overwrite` or `Compaction`. 
Thanks for your remind, I have update this PR, please CC, thx.

wwj6591812:
> The config `manifest.delete-file-drop-stats` used to provide compatibility when downstream reads the `Stats` from `Snapshot` that generated by `Overwrite` or `Compaction`.
> Thanks for your remind, I have update this PR, please CC, thx.

wwj6591812:
The failed ut(TimeoutException) seems not caused by my modify.

JingsongLi:
+1

PR #5113, opened 2025-02-19
[core] Return empty instead of throwing exception when getting system table's latestSnapshotId

mxdzs0612:
### Purpose

Engines may call latestSnapshotId interface at table level, however when it comes to system table, it will throw an exception. I suppose it better to return empty now that the return value is an Optional.

### API and Format

no

### Documentation

no need


JingsongLi:
+1

PR #5114, opened 2025-02-19
[rest] Add isExternal to GetTableResponse

JingsongLi:
### Purpose

Add isExternal to REST GetTableResponse, isExternal can be used to distinguish data file io and more things. So it is neccessary to add an isExternal to table response.


sundapeng:
+1

PR #5115, opened 2025-02-19
[core] Fix incremental query with delete after minor compact

Zouxxyy:
### Purpose

Due to the possibility that the before files and after files may offset each other during incremental read, while there are also new deletes in after files, we cannot drop deletes when reading after files.

```
//            tag1                          tag2
// l0         f(+I 10001),f(+I 10002)       f(-D 999)
// l1
// l2
// l3
// l4                                       f(+I 10001,10002)
// l5         f(+I 1-10000)                 f(+I 1-10000)

// before files: f(+I 10001), f(+I 10002)
// after files:  f(-D 999),   f(+I 10001,10002)
```


JingsongLi (approved):
+1 Thanks @Zouxxyy

PR #5116, opened 2025-02-19
[core] Avoid the unnecessary key value copy

### LINKED ISSUE #5103: [Feature] Improve the LookupMergeFunction copy performance
### Search before asking

### Motivation

Image

We'd better copy the underlying binary instead of copy by get object, this will lead to much cpu cost

### Solution

_No response_

### Anything else?

_No response_

Aitozi:
### Purpose

Linked issue: close #5103

1. skip the copy process during the compaction, because the KeyValue of the same key are from the different reader, and will not read the next key, before the same key are all processed
2. copy the underlying BinaryRow directly for the sort write buffer

```
Apple M3 Pro
parquet:                                                                                             Best/Avg Time(ms)    Row Rate(K/s)      Per Row(ns)   Relative
OPERATORTEST_parquet_write                                                                              39571 / 39979             75.8          13190.3       1.0X


OpenJDK 64-Bit Server VM 11.0.24+0 on Mac OS X 14.5
Apple M3 Pro
parquet:                                                                                             Best/Avg Time(ms)    Row Rate(K/s)      Per Row(ns)   Relative
OPERATORTEST_parquet_write                                                                              31918 / 32666             94.0          10639.3       1.0X
```

The lookup compaction writer speed gains about 25%

### DIFF
--- a/paimon-benchmark/paimon-micro-benchmarks/src/test/java/org/apache/paimon/benchmark/TableWriterBenchmark.java
+++ b/paimon-benchmark/paimon-micro-benchmarks/src/test/java/org/apache/paimon/benchmark/TableWriterBenchmark.java
@@ -19,6 +19,7 @@
 package org.apache.paimon.benchmark;
 
 import org.apache.paimon.CoreOptions;
+import org.apache.paimon.disk.IOManagerImpl;
 import org.apache.paimon.options.Options;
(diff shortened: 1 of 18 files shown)


JingsongLi (commented):
Cool!

JingsongLi (approved):
+1

PR #5119, opened 2025-02-20
[flink] Remove useless blocking queue in AlignedSourceReader

tsreaper:
### Purpose

This PR removes the useless blocking queue in `AlignedSourceReader` to clean up code.

### Tests

### API and Format

No format changes.

### Documentation

No new feature.


JingsongLi:
+1

liming30:
@tsreaper hi, `elementsQueue.notifyAvailable()` should not be removed here. Consider this case:

1. The `reader` consumes data too quickly and has already consumed all the data, but the checkpoint has not been triggered yet. As a result, `shouldTriggerCheckpoint() `returns `Optional.empty()`.
2. The checkpoint starts triggering and `Enumerator` sends a `CheckpointEvent` to the `reader`, but `shouldTriggerCheckpoint() `will not be called again.
3. The `SourceTask` receives the checkpoint RPC message, but since `shouldTriggerCheckpoint()` does not return a valid `CheckpointID`, the process hangs.

Therefore, we need to call `elementsQueue.notifyAvailable()` when the `CheckpointEvent` is received to ensure that `shouldTriggerCheckpoint()` returns a valid `CheckpointID`.

PR #5121, opened 2025-02-20
[core] Enables buffer spill when targetFileSize is greater than the w…

LinMingQiang:
…rite buffer size.

### Purpose

The purpose of this PR is to reduce small files.

When I set `target-file-size=1GB` and  `write-buffer-for-append=true,  write-buffer-size=256m`,
 I found that my final file size is always smaller than `write-buffer-size=256m` and will never reach the 1gb I expected.
This resulted in a lot of small files being generated.

So when `target-file-size` is greater than `write-buffer-size`, spill should be enabled automatically.

Linked issue: close #xxx

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
@@ -479,7 +479,7 @@ public class CoreOptions implements Serializable {
                     .booleanType()
                     .noDefaultValue()
                     .withDescription(
-                            "Whether the write buffer can be spillable. Enabled by default when using object storage.");
+                            "Whether the write buffer can be spillable. Enabled by default when using object storage or when 'target-file-size' is greater than 'write-buffer-size'.");
 
     public static final ConfigOption<Boolean> WRITE_BUFFER_FOR_APPEND =
             key("write-buffer-for-append")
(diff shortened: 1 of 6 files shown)


paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -1947,9 +1947,14 @@ public long writeBufferSize() {
         return options.get(WRITE_BUFFER_SIZE).getBytes();
     }
 
-    public boolean writeBufferSpillable(boolean usingObjectStore, boolean isStreaming) {
+    public boolean writeBufferSpillable(
+            boolean usingObjectStore, boolean isStreaming, boolean hasPrimaryKey) {
```
JingsongLi, line 1951: Can you create a test for this?

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -1947,9 +1947,14 @@ public long writeBufferSize() {
...
         // if not streaming mode, we turn spillable on by default.
-        return options.getOptional(WRITE_BUFFER_SPILLABLE).orElse(usingObjectStore || !isStreaming);
+        return options.getOptional(WRITE_BUFFER_SPILLABLE)
+                .orElse(
+                        usingObjectStore
+                                || !isStreaming
+                                || targetFileSize(hasPrimaryKey) > writeBufferSize());
```
wwj6591812, line 1957: modify the description of config WRITE_BUFFER_SPILLABLE in CoreOptions.java

LinMingQiang: nice catch.

LinMingQiang:
ut failure is not caused by this pr.

JingsongLi (approved):
+1

PR #5127, opened 2025-02-20
[doc][spark] Add `__paimon_row_index` metadata column

liyubin117:
### Purpose

Add `__paimon_row_index` metadata column


JingsongLi (approved):
+1

PR #5132, opened 2025-02-21
[Doc] Modify ths doc of commit memory.

wwj6591812:
### Purpose

When we use Flink job to run dedicated compaction with config of `cluster.fine-grained-resource-management.enabled: true` and `sink.committer-memory: 32G`, we receive the follow exception. This PR is add the solution of this question.

At the moment, fine-grained resource management requires batch workloads to be executed with types of all edges being BLOCKING. To do that, you need to configure 'fine-grained.shuffle-mode.all-blocking' to 'true'. Notice that this may affect the performance. See FLINK-20865 for more details.

Linked issue: close #xxx


JingsongLi:
+1

PR #5135, opened 2025-02-21
[core] Add dialects to View

JingsongLi:
### Purpose

In many scenarios, the dialects of each engine are different. We can provide View with dialect fields to provide space for future extensions.

However, there is currently no operation to control dialects, and multiple dialects can only be written directly through the API when creating a View.

Perhaps in the future, we can consider adding some procedures to add, modify, or delete dialects.

Some information in Iceberg community: https://lists.apache.org/thread/k6szpr5smyrh37sy563xpgjor4g6pr81

They seem more inclined to use `Substrait` to only save intermediate states, but that solution appears to be very complex and difficult to control.

BTW, this PR also change `RowType rowType` to `List<DataField> fields` in `ViewSchema`, this is good to be consistent with Table `Schema`.


zhuangchong (approved):
LGTM.

PR #5145, opened 2025-02-25
[rest] Make http client reusable to reduce connection cost

JingsongLi:
### Purpose

The general recommendation is to reuse the HttpClient to improve the performance of the connection pool. Our HttpClient can be placed in a static variable for reuse, which can be beneficial for us to create Catalogs more freely to fetch certain data and close the Catalogs。


jerry-024:
As there are currently no concurrent calls to the RESTCatalog server, it is sufficient to use the default settings.
LGTM

PR #5151, opened 2025-02-25
[cdc] Fix canal cdc data field value is null judgment condition.

zhuangchong:
### Purpose

canal json data:
image

exception:
image


JingsongLi (approved):
+1

PR #5152, opened 2025-02-25
[rest] Add commit statistics when commit to REST Server

JingsongLi:
### Purpose

When we commit a table, we can obtain the statistical information of this change. We can aggregate this statistical information by partition and provide it to the REST service. This way, the server can maintain tables and partitions based on this information to provide better services.


jerry-024:
+1

PR #5153, opened 2025-02-25
[Bug] Fix NumberFoematException of Dedicated Compaction

wwj6591812:
### Purpose

In my company, the batch dedicated compaction throw a exception.
After read the code, we found in CompactorSourceBuilder#batchCompactOptions we will do `put(CoreOptions.SCAN_SNAPSHOT_ID.key(), null);`, so the exception throw.
Now fix it.

image

Linked issue: close #xxx


JingsongLi (approved):
+1

PR #5158, opened 2025-02-26
[metrics] Add total file size metrics of active buckets for primary key tables

tsreaper:
### Purpose

This PR adds total file size metrics of active buckets for primary key tables, so that when users experience a slow compaction,  they can quickly check if some buckets are too large.

### Tests

* `CompactionMetricsTest`.

### API and Format

No format changes.

### Documentation

Document is also updated.


wwj6591812:
Why not also support append table?

PR #5162, opened 2025-02-26
[core] Use PriorityQueue intead of TreeSet. TreeSet will deduplicate record if comparator return 0

leaves12138:
### Purpose

TreeMap will remove record if comparator return 0. But we don't want this happen in BucktedAppendTableWrite.


JingsongLi (approved):
+1

PR #5167, opened 2025-02-26
[build] Optimize Kafka container to use KRaft instead of relying on ZooKeeper

liyubin117:
### Purpose

Reduce the complexity of E2E container and improve the stability. 
many projects such as Iceberg have removed relying on zk


zhuangchong (approved):
LGTM

PR #5170, opened 2025-02-27
[cdc] Eager init to force creating snapshot even no data

### LINKED ISSUE #5169: [Feature] Force creating snapshot for those specified tables even no data ingested
### Search before asking

### Motivation

When ingesting data with kafka_sync_database action, if one table has no data coming since job restarted, no snapshot would be created for it even set as **_commit.force-create-snapshot=true_**, or **_tag.automatic-creation_=process-time**.  This is a **_lazy_** initialization. Only those tables passed before would have the chance to froce creating snapshot.
No snapshot means no tag. But the tag is crucial for batchly processing. We need the tag to detect changes incrementally by calling paimon_incremental_to_auto_tag.

### Solution

Initializing eagerly which means _**register**_ the tales at each startup, instead of waiting the first record passing through.

### Anything else?

_No response_

JackeyLee007:
### Purpose

Linked issue: close #5169 

Register the tales at each startup to force creating snapshot, instead of waiting the first record passing through.


JingsongLi (approved):
+1

PR #5173, opened 2025-02-27
[doc] Add mysql cdc synchronization table comment document.

zhuangchong:
### Purpose

Add mysql cdc synchronization table comment document.
image

### Tests

MySqlSyncTableActionITCase#testSchemaEvolutionWithComment


JingsongLi:
+1

PR #5180, opened 2025-02-28
[Log] Add log to help debug

wwj6591812:
### Purpose

In my company, a job throw the follow exception. 
When debug this problem, we found that some log is absent in Paimon, now add it.

image

Linked issue: close #xxx


paimon-format/src/main/java/org/apache/paimon/format/parquet/ParquetReaderFactory.java
```diff
@@ -150,6 +150,13 @@ public FileRecordReader<InternalRow> createReader(FormatReaderFactory.Context co
                         context.selection());
         MessageType fileSchema = reader.getFileMetaData().getSchema();
         MessageType requestedSchema = clipParquetSchema(fileSchema);
+
+        LOG.info(
```
JingsongLi, line 154: LOG.debug? here will be lots of logs.

wwj6591812: done

JingsongLi:
+1

PR #5182, opened 2025-02-28
[core] RESTCatalog: update parse region from dlf uri

jerry-024:
### Purpose
RESTCatalog: update parse region from DLF uri
Linked issue: #4540 

### Tests
DLFAuthProviderFactoryTest.getRegion


paimon-core/src/main/java/org/apache/paimon/rest/auth/DLFAuthProviderFactory.java
```diff
@@ -51,15 +54,18 @@ public AuthProvider create(Options options) {
         throw new IllegalArgumentException("DLF token path or AK must be set for DLF Auth.");
     }
 
-    private static String getRegion(Options options) {
-        String region = "undefined";
+    protected static String getRegion(String uri) {
```
JingsongLi, line 57: Maybe we can introduce a `dlf.region` option in this PR too.

shyjsarah:
+1

JingsongLi (approved):
+1

PR #5183, opened 2025-02-28
[flink] fix parallelismConfigured in cases when sink parallelism is set

yunfengzhou-hub:
### Purpose

This PR fixes the bug that AdaptiveBatchScheduler can still adjust the parallelism of a writer operator when `sink.parallelism` is configured.

### Tests

ReadWriteTableITCase#testSinkParallelism


JingsongLi (approved):
+1

PR #5188, opened 2025-02-28
[core] Extract ChangelogManager from SnapshotManager

JingsongLi:
### Purpose

Keep the logic of changelog only in a separate class to avoid affecting the logic of Snapshot Manager.


yuzelin (approved):
+1

PR #5192, opened 2025-03-03
[flink] Use RuntimeContextUtils to keep compatible

leaves12138:
getNumberOfParallelSubtasks From getRuntimeContext() may be not found in separated flink version. SO replace it by RuntimeContextUtils.


JingsongLi:
+1

PR #5193, opened 2025-03-03
[flink] Add default parallelism for source and sink in rescale procedure, also make them configurable

tsreaper:
### Purpose

Currently user has no way to configure source and sink parallelism in rescale procedure. This PR adds default parallelism and make it configurable.

### Tests

### API and Format

No format changes.

### Documentation

Document is also added.

### DIFF
--- a/docs/content/flink/procedures.md
+++ b/docs/content/flink/procedures.md
@@ -776,15 +776,15 @@ All available procedures are listed below.
    <tr>
       <td>rescale</td>
       <td>
-         CALL [catalog.]sys.rescale(`table` => 'identifier', `bucket_num` => bucket_num, `partition` => 'partition', `source_parallelism` => 'source_parallelism', `sink_parallelism` => 'sink_parallelism')
+         CALL [catalog.]sys.rescale(`table` => 'identifier', `bucket_num` => bucket_num, `partition` => 'partition', `source.parallelism` => 'source.parallelism', `sink.parallelism` => 'sink.parallelism')
       </td>
       <td>
          Rescale one partition of a table. Arguments:
          <li>identifier: The target table identifier. Cannot be empty.</li>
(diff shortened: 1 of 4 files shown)


docs/content/flink/procedures.md
```diff
@@ -757,13 +757,15 @@ All available procedures are listed below.
    <tr>
       <td>rescale</td>
       <td>
-         CALL [catalog.]sys.rescale(`table` => 'identifier', `bucket_num` => bucket_num, `partition` => 'partition')
+         CALL [catalog.]sys.rescale(`table` => 'identifier', `bucket_num` => bucket_num, `partition` => 'partition', `source_parallelism` => 'source_parallelism', `sink_parallelism` => 'sink_parallelism')
```
JingsongLi, line 760: scan.parallelism and sink.parallelism.

Just be consistent with Flink Connector options.

docs/content/flink/procedures.md
```diff
@@ -757,13 +757,15 @@ All available procedures are listed below.
...
       </td>
       <td>
          Rescale one partition of a table. Arguments:
          <li>identifier: The target table identifier. Cannot be empty.</li>
          <li>bucket_num: Resulting bucket number after rescale. The default value of argument bucket_num is the current bucket number of the table. Cannot be empty for postpone bucket tables.</li>
          <li>partition: What partition to rescale. For partitioned table this argument cannot be empty.</li>
+         <li>source_parallelism: Parallelism of source operator. The default value is the current bucket number of the partition.</li>
+         <li>partition: Parallelism of sink operator. The default value is equal to bucket_num.</li>
```
JingsongLi, line 768: fix it

docs/content/flink/procedures.md
tsreaper: No need. This is a procedure argument, not an option.

wwj6591812:
+1

docs/content/flink/procedures.md
JingsongLi: Can we maintain consistency, why not?

JingsongLi:
+1

PR #5197, opened 2025-03-03
[refactor] Move #earlierThanTimeMills to TimeTravelUtil

yuzelin:
### Purpose

earlierThanTimeMills might be used by other class, so it's better to put it in a util class


JingsongLi:
+1

PR #5198, opened 2025-03-03
[core] Support cross-partition for fallback branch feature

liyubin117:
### Purpose
Support cross-partition for fallback branch feature.
Currently, once writing into a cross-partition table which fallback branch feature enabled, 
```
Flink SQL> create table part1(id int primary key not enforced, name string, dt string) partitioned by (dt);
Flink SQL> call sys.create_branch('branch.part1', 'stream');
Flink SQL> alter table part1 set ('scan.fallback-branch'='stream');
Flink SQL> insert into part1 values (1,'anew','20250301'),(20, 'from stream', '20250303');
```
throws exception as follow, it is caused by FallbackReadFileStoreTable#newScan return a DataTableScan instance, we should use `DataTableScan` instead of `AbstractDataTableScan` in IndexBootstrap#bootstrap.
```
Caused by: java.lang.ClassCastException: org.apache.paimon.table.FallbackReadFileStoreTable$Scan cannot be cast to org.apache.paimon.table.source.AbstractDataTableScan
	at org.apache.paimon.crosspartition.IndexBootstrap.bootstrap(IndexBootstrap.java:87)
	at org.apache.paimon.crosspartition.IndexBootstrap.bootstrap(IndexBootstrap.java:69)
	at org.apache.paimon.flink.sink.index.IndexBootstrapOperator.initializeState(IndexBootstrapOperator.java:60)
	at org.apache.flink.streaming.api.operators.StreamOperatorStateHandler.initializeOperatorState(StreamOperatorStateHandler.java:147)
	at org.apache.flink.streaming.api.operators.AbstractStreamOperator.initializeState(AbstractStreamOperator.java:294)
```

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/crosspartition/IndexBootstrap.java
+++ b/paimon-core/src/main/java/org/apache/paimon/crosspartition/IndexBootstrap.java
@@ -27,8 +27,8 @@ import org.apache.paimon.io.DataFileMeta;
 import org.apache.paimon.reader.RecordReader;
 import org.apache.paimon.schema.TableSchema;
 import org.apache.paimon.table.Table;
-import org.apache.paimon.table.source.AbstractDataTableScan;
 import org.apache.paimon.table.source.DataSplit;
+import org.apache.paimon.table.source.DataTableScan;
 import org.apache.paimon.table.source.ReadBuilder;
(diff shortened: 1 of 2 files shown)


JingsongLi (approved):
+1

PR #5202, opened 2025-03-03
[core] Support partition strategy for unaware bucket partitioned Appe…

LinMingQiang:
…nd only table.

In my company , we has an order table, we use the order create-time dt as the partition and extract MySQL data regularly according to the order update-time, each incremental data will contain order data of 600+ partitions (dt,hm). when we put it into the paimon unaware bucket append table, we find that there are many small files.
Flink streaming write:  set sink.parallelism = 40, one batch of data will be written to (600*40) files and 40 files per partition.

### Purpose

The purpose of this PR is to reduce the problem of small files when writing unaware bucket partitioned append tables. 

Currently, The unaware bucket partitioned append table will repartition the data when writing, this causes each task will receive all partitions data. So, each batch of data written will generate `(sink.parallelism)` files under each partition.
 `(partitionNums * parallelism)` files will be written at once.

Example : We write data to 5 partitions, and set `sink.parallelism = 5`. 

```
CREATE  TABLE xx ( 
..
) PARTITIONED BY (dt) WITH ( 
'bucket' = '-1' 
,'sink.parallelism' = '5' 
)

Step 2 : Write data across 5 partitions...

Step 3 : Watch manifest-list file:  (partitionNums * parallelism) 25 files were written at once. 

```

We can optimize this part by supporting the hash strategy to reduce the number of partitions each task receives.
Making the data of the same partition fall into the same task to reduce the number of files under the partition.

Apply this pr later : There is only one file per partition, while reducing the number of files, it also improves the file compression ratio and reduces storage space.

Linked issue: close #xxx

### Tests

UnawareBucketAppendOnlyTableITCase#testPartitionStrategyForPartitionedTable

### DIFF
--- a/docs/layouts/shortcodes/generated/core_configuration.html
+++ b/docs/layouts/shortcodes/generated/core_configuration.html
@@ -708,6 +708,12 @@ This config option does not affect the default filesystem metastore.</td>
             <td>Mark done action will reports the partition to the remote http server, this can only be used by http-report partition mark done action.</td>
         </tr>
         <tr>
+            <td><h5>partition.strategy.for-unaware-append</h5></td>
+            <td style="word-wrap: break-word;">NONE</td>
+            <td><p>Enum</p></td>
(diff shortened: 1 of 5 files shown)


JingsongLi (approved):
+1

PR #5203, opened 2025-03-03
[doc] Add description in merge-enfine for Custom FieldAggregator.

LinMingQiang:
### Purpose

Linked issue: close #xxx


JingsongLi:
We don't want to public Agg Function APIs, it is better to just keep it as internal API.

PR #5205, opened 2025-03-04
[doc] Add lookup compaction description in primary key compaction page

xiangyuf:
### Purpose

Linked issue: close #xxx

Add lookup compaction description in primary key compaction page


JingsongLi:
+1

PR #5207, opened 2025-03-04
[hotfix] ArrowFormatWriter should reset ArrowFieldWriters after flushing

yuzelin:
### Purpose

The`ArrayWriter` and `MapWriter` need to be reset before writing again, so the `ArrowFormatWriter` should expose the reset method.


leaves12138 (approved):
+1

PR #5209, opened 2025-03-04
[iceberg] fix that dropPartition cannot sychronize the metadata to iceberg

LsomeYeah:
### Purpose

Linked issue: close #xxx

Paimon synchronizes metadata to iceberg by invoking IcebergCommitCallback method when committing. DropPartition operation create `commit` by `AbstractFileStore#newCommit(commitUser)` which will pass an empty list of CommitCallback to `commit` and thus metadata cannot be synchronized to iceberg.

This can be fixed by creating corresponding IcebergCommitCallback when invoking `AbstractFileStore#newCommit(commitUser)`.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/AbstractFileStore.java
+++ b/paimon-core/src/main/java/org/apache/paimon/AbstractFileStore.java
@@ -30,7 +30,9 @@ import org.apache.paimon.index.IndexFileHandler;
 import org.apache.paimon.manifest.IndexManifestFile;
 import org.apache.paimon.manifest.ManifestFile;
 import org.apache.paimon.manifest.ManifestList;
+import org.apache.paimon.metastore.AddPartitionCommitCallback;
 import org.apache.paimon.metastore.AddPartitionTagCallback;
+import org.apache.paimon.metastore.TagPreviewCommitCallback;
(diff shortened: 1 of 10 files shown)


paimon-core/src/main/java/org/apache/paimon/AbstractFileStore.java
```diff
@@ -360,6 +360,8 @@ public TagDeletion newTagDeletion() {
 
     public abstract Comparator<InternalRow> newKeyComparator();
 
+    protected abstract List<CommitCallback> createCommitCallbacks(String commitUser);
```
JingsongLi, line 363: Can you move `AbstractFileStoreTable.createCommitCallbacks` to here and remove `newCommit(user, callbacks)` method?

JingsongLi:
+1

paimon-core/src/main/java/org/apache/paimon/AbstractFileStore.java
caozhen1937: > Can you move `AbstractFileStoreTable.createCommitCallbacks` to here and remove `newCommit(user, callbacks)` method?

Hi @JingsongLi ,  now newCommit function can't transport callbacks, how to get some result by callbacks that transported  by method param? (e.g. I store some meta in Callback implement class)

PR #5210, opened 2025-03-05
[hotfix] RestCatalog: fix potential dead loop when listTables/Views/Part…

XiaoHongbo-Hope:
…itions

[hotfix] RestCatalog: fix potential dead loop when listTables/Views/Partitions if restClient return null

#Linked issue: close #xxx


paimon-core/src/main/java/org/apache/paimon/rest/RESTCatalog.java
```diff
@@ -884,6 +886,7 @@ public List<String> listViews(String databaseName) throws DatabaseNotExistExcept
                             "response of listViews for {} is null with params {}",
                             databaseName,
                             queryParams);
+                    pageToken = null;
```
JingsongLi, line 889: just break?

XiaoHongbo-Hope: updated

JingsongLi:
+1

PR #5211, opened 2025-03-05
[cdc] Improve cdc record conversion to row Failure logs add field name

zhuangchong:
### Purpose

When a cdc record conversion exception occurs, the user does not know which field it is.

image


JingsongLi (approved):
+1

PR #5212, opened 2025-03-05
[core] Refactor FileStore.newCommit and avoid using it

JingsongLi:
### Purpose

The fix in https://github.com/apache/paimon/pull/5209 , will create new Table in FileStore, this is very weird.

We should have a more correct dependency relationship, using reference passing instead of creating another Table, so this PR refactored this part:

1. Reduce external use of FileStoreCommit, and external parties should try to use TableCommit as much as possible.
2. When creating FileStoreCommit, pass the FileStoreTable object in to avoid creating new Table objects.


LsomeYeah (approved):
+1

PR #5220, opened 2025-03-05
[test][hive] Upgrade hiverunner usage to support tests in jdk11

liyubin117:
### Purpose

Currently, hive tests on jdk11 always failed.
After digging into the problem, I found that old hiverunner usage was outdated and can be replaced as `io.github.hiverunner:hiverunner` refer to https://github.com/HiveRunner/HiveRunner/blob/main/README.md.
In addition, The dependency tez 0.9.1 in hiverunner has bugs for jdk0.9+ and have been fixed in 0.10.0 refer to https://issues.apache.org/jira/browse/TEZ-3860.

CI on jdk11: https://github.com/liyubin117/paimon/actions/runs/13676925840

### DIFF
--- a/paimon-hive/paimon-hive-connector-2.3/pom.xml
+++ b/paimon-hive/paimon-hive-connector-2.3/pom.xml
@@ -156,7 +156,7 @@ under the License.
         </dependency>
 
         <dependency>
-            <groupId>com.klarna</groupId>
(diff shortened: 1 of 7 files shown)


JingsongLi:
Looks good to me! Thanks @liyubin117

PR #5223, opened 2025-03-06
[core] RESTCatalog:  encode resource name and query value in url

jerry-024:
### Purpose
RESTCatalog:  encode resource name in URL

Linked issue: https://github.com/apache/paimon/issues/4540

### Tests
- HttpClientTest. testUrl
- ResourcePathsTest. testUrlEncode

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/rest/RESTUtil.java
+++ b/paimon-core/src/main/java/org/apache/paimon/rest/RESTUtil.java
@@ -24,6 +24,11 @@ import org.apache.paimon.utils.Preconditions;
 import org.apache.paimon.shade.guava30.com.google.common.collect.ImmutableMap;
 import org.apache.paimon.shade.guava30.com.google.common.collect.Maps;
 
+import java.io.UncheckedIOException;
+import java.io.UnsupportedEncodingException;
+import java.net.URLDecoder;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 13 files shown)


JingsongLi (approved):
+1

paimon-core/src/main/java/org/apache/paimon/rest/RESTUtil.java
```diff
@@ -58,4 +63,28 @@ public static Map<String, String> merge(
...
+        } catch (UnsupportedEncodingException e) {
+            throw new UncheckedIOException(
+                    String.format(
+                            "Failed to URL encode '%s': UTF-8 encoding is not supported", toEncode),
+                    e);
+        }
+    }
+
+    public static String decodeString(String encoded) {
+        Preconditions.checkArgument(encoded != null, "Invalid string to decode: null");
```
plusplusjiajia, line 80: I'm not sure, if it is null or empty, return empty string directly is better.

paimon-core/src/main/java/org/apache/paimon/rest/ResourcePaths.java
```diff
@@ -55,87 +54,166 @@ public String databases() {
     }
 
     public String database(String databaseName) {
-        return SLASH.join(V1, prefix, DATABASES, databaseName);
```
plusplusjiajia, line 58: prefix also can be encoded?

PR #5224, opened 2025-03-06
[core] Avoid getting file size for manifest list and writing file

JingsongLi:
### Purpose

Performance optimization for removing file size getting.


wwj6591812:
+1

PR #5228, opened 2025-03-07
Preserving old Iceberg Metadata files

### LINKED ISSUE #5189: [Bug] Deletion of older Iceberg metadata files on new writes causes Iceberg Readers to fail
### Search before asking

### Paimon version

In Paimon Iceberg compatibility mode, I have observed that Iceberg metadata files are completely replaces as new data is written from either Spark or Flink. This causes existing reader based on old Iceberg metadata files to fail, making the tables non-usable is scenarios where you have multiple readers involved.

I was looking at the code, and it appears we do delete the old metadata files every time a new one is created. There can be 2 different scenarios for that.

1. In the first case, when Iceberg metadata files are created based on older metadata files [it simply deletes the older file ](https://github.com/apache/paimon/blob/732bff826741054c2ff8bb34b267c58a067678ae/paimon-core/src/main/java/org/apache/paimon/iceberg/AbstractIcebergCommitCallback.java#L427)
2. In the other case, when Iceberg metadata files are created based on snapshots, [expireAllBefore is called here](https://github.com/apache/paimon/blob/732bff826741054c2ff8bb34b267c58a067678ae/paimon-core/src/main/java/org/apache/paimon/iceberg/AbstractIcebergCommitCallback.java#L277) that quietly removes all the old metadata files [here](https://github.com/apache/paimon/blob/732bff826741054c2ff8bb34b267c58a067678ae/paimon-core/src/main/java/org/apache/paimon/iceberg/AbstractIcebergCommitCallback.java#L748)

### Compute Engine

I have verified with both Spark & Flink, but it should be applicable for any compute engine.

### Minimal reproduce step

Reproducing steps.

- Create new table  in Iceberg Compatibility mode and insert some data in it via Spark SQL.
```
CREATE TABLE paimon_catalog.`default`.cities(country STRING,name STRING) TBLPROPERTIES ('metadata.iceberg.storage' = 'hadoop-catalog');


INSERT INTO paimon_catalog.`default`.cities VALUES ('usa', 'sanjose'), ('germany', 'berlin');
```

- At this point a Iceberg metadata file should be created

- Insert some more data into the Paimon table
```
INSERT INTO paimon_catalog.`default`.cities VALUES ('usa', 'chicago'), ('germany', 'hamburg');
```

- The old Iceberg metadata file should be replaced with the newer one.

Right now, the code doesn't leverage the Iceberg options `write.metadata.previous-versions-max` and `write.metadata.delete-after-commit.enabled` as mentioned[ here](https://iceberg.apache.org/docs/1.5.1/maintenance/#remove-old-metadata-files). In my view, we can support [these Iceberg Options](https://github.com/apache/paimon/blob/732bff826741054c2ff8bb34b267c58a067678ae/paimon-core/src/main/java/org/apache/paimon/iceberg/IcebergOptions.java) for Iceberg compatible Paimon table and delete old metadata files based on the configurations.

### Anything else?

Let me know about your thoughts here.

junmuz:
### Purpose

Linked issue: close #5189 

This change preserves the old metadata and manifest files. Right now it lets the snapshot cleanup to happen, so some of the manifest files might point to old data files from previous snapshots that have been deleted

### Tests

Todo

### API and Format

A new table configuration option is exposed. For backward compatibility, the default value is chosen as 0.

### Documentation

It's not a feature, but makes the existing metadata file deletion mechanism configurable.


junmuz:
@JingsongLi Does the above changes looks ok to you?

paimon-core/src/main/java/org/apache/paimon/iceberg/IcebergCommitCallback.java
```diff
@@ -752,7 +752,25 @@ private void expireAllBefore(long snapshotId) throws IOException {
                 }
                 table.fileIO().deleteQuietly(listPath);
             }
-            table.fileIO().deleteQuietly(path);
+            deleteApplicableMetadataFiles(snapshotId);
```
LsomeYeah, line 755: In `expireAllBefore` invoked by  `createMetadataWithoutBase`, all the iceberg manifest lists and manifest files for previous snapshots are deleted, why still retain the previous metadata file?(if the relevant options had been set)

junmuz: Looking at the code here, `createMetadataWithoutBase` would be invoked if metadata file doesn't exist for the latest snapshot. This means a whatever older Iceberg metadata file available would be referenced by the Iceberg readers. Creating the new one and deleting old one quickly can cause some race conditions. Also, it would make it less consistent specially if someone disables deletion of metadata file altogether with option `delete-after-commit.enabled`.

LsomeYeah: Thanks for your explanation!

I agree with the second point, but for the first point, all the snapshots in the old metadata file had been expired in `expireAllBefore`, relevant manifest list files and manifest meta files had been deleted too. In this case iceberg readers could not read data files properly even if the old metadata file was retained.

Therefore, could I assume that the key reason for retaining old metadata is for auditing purposes(may be for tracking schema evolution or some other reason) in `expireAllBefore`?

junmuz: Yes, then in that cases it would be retained for auditing purposes only. Once a new metadata is created, presence of old Iceberg metadata with all the expired snapshots shouldn't be an issue for Iceberg Readers as latest metadata file would always be used.

LsomeYeah (approved):
+1

JingsongLi:
+1

PR #5233, opened 2025-03-08
[core] refine the dlf config key

plusplusjiajia:
### Purpose
In Paimon's configuration items, it seems that lowercase letters and - are usually used to define configuration key names.
Another reason is that when options are passed from SparkCatalog, all keys are converted to lowercase.


JingsongLi (approved):
+1

PR #5237, opened 2025-03-09
[core] Fix that snapshot expire might delete files used by tag mistakenly

JingsongLi:
### Purpose

Another solution for #5229

This PR tries to maintain the original logic of Tag deleting files as much as possible to avoid residual files.


yuzelin (approved):
+1

PR #5240, opened 2025-03-10
[procedure] Merge migrate_file to migrate_table

JingsongLi:
### Purpose

Migrate_file should part of migrate_table, it is useless, we can just add a target_table in migrate_table procedure.


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/procedure/MigrateTableProcedure.java
```diff
@@ -30,6 +29,8 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static org.apache.paimon.utils.ParameterUtils.parseCommaSeparatedKeyValues;
+
 /** Migrate procedure to migrate hive table to paimon table. */
 public class MigrateTableProcedure extends ProcedureBase {
```
LsomeYeah, line 35: Why not apply the migration changes to MigrateTableProcedure in paimon-flink-1.18?

paimon-hive/paimon-hive-connector-common/src/test/java/org/apache/paimon/hive/procedure/MigrateFileProcedureITCase.java
```diff
@@ -36,13 +34,11 @@
...
 
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
 import java.util.Random;
 import java.util.stream.Stream;
 
-/** Tests for {@link MigrateFileProcedure}. */
+/** Tests for {@code MigrateFileProcedure}. */
 public class MigrateFileProcedureITCase extends ActionITCaseBase {
```
LsomeYeah, line 42: Since we have already removed MigrateFileProcedure, would it be better to move these tests to MigrateTableProcedureITCase?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/MigrateTableAction.java
```diff
@@ -53,7 +53,9 @@ public void run() throws Exception {
                 new DefaultProcedureContext(env),
                 connector,
                 hiveTableFullName,
+                null,
                 tableProperties,
-                parallelism);
+                parallelism,
+                null);
     }
```
LsomeYeah, line 60: Would it be better to pass the parameters through the factory instead of directly setting it to null?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/procedure/MigrateTableProcedure.java
JingsongLi: To ensure compatibility.

LsomeYeah (approved):

Issue #4816, opened 2025-01-02
[Feature] Paimon Spark 2025 Roadmap


Zouxxyy:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Motivation

| Name                      | Introduction                                                                                                                     | Link  |
|---------------------------|----------------------------------------------------------------------------------------------------------------------------------|-------|
| Variant Type              | [feat] Support Variant Type, unlocking support for semi-structured data.                                                         | #4471 |
| Optimized Write           | [perf] Optimize table writing, including automatic repartitioning and rebalancing data and so on.                                |       |
| Distributed Planning      | [perf] Support distributed planning in the scan phase.                                                                           |       |
| Dataframe V2              | [feat] Integrate Spark's DataFrame V2 API.                                                                                       |       |
| Liquid Clustering         | [perf] Support liquid clustering.                                                                                                | #4815 |
| Isolation Level           | [feat] Transaction isolation that supports more levels, like serializable isolation level.                                       | #4616 |
| Support For Spark Connect | [feat] Support spark connect, calling "Paimon Connect".                                                                          |       |
| Default Value             | [feat] Support default values for specified fields.                                                                              |       |
| Constraints               | [feat] Support adding constraints to fields, such as not null or other custom constraints.                                       |       |
| Partition Stats           | [feat] Support partition stats.                                                                                                  |       |
| Row Lineage               | [feat] Support tracking tracking row lineage.                                                                                    |       |
| Identity Column           | [feat] When no explicit values are provided during writing, generate unique values for identity column                           |       |
| Generated Columns         | [feat] Support generated columns whose values are automatically generated based on a user-specified function over other columns. |       |
| CDC For Non-PK Table      | [feat] Support CDC for non-pk table.                                                                                             |       |

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

YannByron:
Any other features or requirements that you would like to have, please comment here. Then we can discuss and modify this roadmap together. Thanks.

YannByron:
And if someone want to take one or some, take it and let us know.

Aiden-Dong:
Can I take on this task?

Distributed Planning : [perf] Support distributed planning in the scan phase.

Zouxxyy:
@Aiden-Dong Yes, feel free for it,  you can create an issue for it, additionally, this feature actually requires changes in the core, and then each compute engine will need to support it.

[...]

PR #5242, opened 2025-03-10
[spark]: add paimon bucket functions

zhongyujiang:
### Purpose

Linked issue: part of #4816, split out from #5241

Add the Paimon bucket function transform in the Paimon-Spark connector.
 It can be used to request data clustering based on the Paimon bucket transform from DataSource V2 when integrating V2 write: [use case](https://github.com/apache/paimon/pull/5241/files#diff-dd7326460f0b9fa115c8d266fed0895eb4279307d5c78c41611751c608ddd343R44)

### Tests

BucketFunctionTest


paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/catalog/functions/BucketFunction.java
```diff
@@ -0,0 +1,724 @@
...
+        // Joint bucket fields of common types
+        builder.put("BucketIntegerInteger", BucketIntegerInteger.class);
+        builder.put("BucketIntegerLong", BucketIntegerLong.class);
+        builder.put("BucketIntegerString", BucketIntegerString.class);
+        builder.put("BucketLongInteger", BucketLongInteger.class);
+        builder.put("BucketLongLong", BucketLongLong.class);
+        builder.put("BucketLongString", BucketLongString.class);
+        builder.put("BucketStringInteger", BucketStringInteger.class);
+        builder.put("BucketStringLong", BucketStringLong.class);
+        builder.put("BucketStringString", BucketStringString.class);
```
zhongyujiang, line 97: For primitive types and common composite types, I added separate bucket function classs to improve [performance](https://github.com/apache/spark/blob/1a42aa5bd44e7524bb55463bbd85bea782715834/sql/catalyst/src/main/java/org/apache/spark/sql/connector/catalog/functions/ScalarFunction.java#L35), while `BucketGeneric` handles other types.

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/catalog/functions/BucketFunction.java
```diff
@@ -0,0 +1,724 @@
...
+        if (type instanceof ArrayType
+                || type instanceof MapType
+                || type instanceof RowType
+                || type instanceof VariantType) {
+            return false;
+        }
+
+        if (type instanceof org.apache.paimon.types.TimestampType) {
+            return ((org.apache.paimon.types.TimestampType) type).getPrecision()
+                    == SPARK_TIMESTAMP_PRECISION;
```
zhongyujiang, line 121: The Spark Timestamp type is in microseconds and does not allow custom precision. When binding the Paimon bucket function, Paimon’s type information is not available. Therefore, support is only possible when the Paimon timestamp precision is equal to the Spark timestamp precision.

zhongyujiang (commented):
cc @Zouxxyy @YannByron Can you please take a review when you have time? Thanks!

zhongyujiang:
@Zouxxyy @YannByron Hi, can you please help review this when you have time? Thanks.

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/catalog/functions/BucketFunction.java
Zouxxyy: Have you tested the performance of this implementation? I think the current approach is too specialized, and the data conversion involving Spark and Paimon doesn't reuse the code.

zhongyujiang: https://github.com/zhongyujiang/incubator-paimon/tree/gh/spark-v2-write_function-benchmark
```text
Benchmark                                            Mode  Cnt  Score   Error  Units
PaimonBucketFunctionBenchmark.bucketFunction           ss    5  0.416 ± 0.033   s/op
PaimonBucketFunctionBenchmark.genericBucketFunction    ss    5  0.685 ± 0.302   s/op
```
Here is a JMH benchmark result. 
I think this optimization has limited impact on write performance, since a significant amount of time is spent on network and compression during data writing. However, there is indeed a noticeable performance difference between the two implementations.
This part of the code can't be reused since the invoke method takes different input types.

Zouxxyy: It looks like the difference is small. I added a benchmark https://github.com/apache/paimon/pull/5418. 
Will test the performance of the existing bucket function and this PR's, including the function and e2e write cost.

zhongyujiang: If we only compare these two function implementations, you can see from the results above that there's a clear difference 0.685 / 0.416 = 1.64.
In real write scenarios, I agree that the impact of this optimization is very minimal.

zhongyujiang: FYI, I’ve already added a benchmark comparison between the v2 write and the previous v1 write [here](https://github.com/apache/paimon/pull/5241/files#diff-b60ea9f4f3114528a9e9c1c4880f61c6c1d2b3bc7c467b0b22b44abe6ad69b71), the results is attached in the PR too. But it didn’t include the newly added v1 bucket function optimization, I think the benchmark results from JMH would be more reliable.

zhongyujiang: Sure, let me update this.

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/catalog/functions/BucketFunction.java
```diff
@@ -0,0 +1,724 @@
...
+        public String canonicalName() {
+            return String.format(
+                    "paimon.bucket(%s)",
+                    Arrays.stream(bucketKeyTypes)
+                            .map(DataType::catalogString)
+                            .collect(Collectors.joining(", ")));
+        }
+
+        // org.apache.paimon.table.sink.KeyAndBucketExtractor.bucket(int, int)
+        protected static int bucket(BinaryRow bucketKey, int numBuckets) {
```
Aitozi, line 273: Why not direct use `KeyAndBucketExtractor.bucket`

Aitozi:
Hi, I'm curious what's the different between the `FixedBucketExpression` vs `BucketFunction`. I just try to know why we have to use the `BoundFunction` not the `Expression` here ?

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/catalog/functions/BucketFunction.java
Aitozi: I try to introduce the HashType for the paimon table https://github.com/apache/paimon/pull/5445. So I'd like to see the hash calculation are managed in one place

zhongyujiang: I'm trying to avoid the uncessary `checkArgument`  validate, I can use that directly as the overhead is small

zhongyujiang:
> Hi, I'm curious what's the different between the FixedBucketExpression vs BucketFunction. I just try to know why we have to use the BoundFunction not the Expression here ?

@Aitozi Yea, Zouxxyy also mentioned this. I think when I pulled this PR, `FixedBucketExpression` hadn't been merged yet, so I added `BucketFunction`. I'll update this PR today.

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/catalog/functions/BucketFunction.java
Zouxxyy: Oh,, I alse noticed this check and planned to delete it

paimon-spark/paimon-spark-ut/src/test/java/org/apache/paimon/spark/catalog/functions/BucketFunctionTest.java
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    @Test
+    public void testLZTimestampType() {
+        validateSparkBucketFunction(LZ_TIMESTAMP_COL);
+    }
+
+    @Test
+    public void testDecimalType() {
+        validateSparkBucketFunction(DECIMAL_COL);
```
zhongyujiang, line 295: This would fail if we directly use the current `InternalRowSerializer`.

paimon-spark/paimon-spark-ut/src/test/java/org/apache/paimon/spark/catalog/functions/BucketFunctionTest.java
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    private static void validateSparkBucketFunction(String... bucketColumns) {
+        setupTable(bucketColumns);
+        spark.sql(
+                        String.format(
+                                "SELECT id_col, __paimon_bucket as expected_bucket, paimon.bucket(%s, %s) FROM %s",
+                                NUM_BUCKETS, String.join(",", bucketColumns), TABLE_NAME))
+                .collectAsList()
+                .forEach(row -> Assertions.assertThat(row.getInt(2)).isEqualTo(row.get(1)));
```
zhongyujiang, line 235: Added this test to compare the bucket values calculated by directly writing to the Paimon table with those calculated using the bucket function registered in the Spark function catalog, so this proves that the values calculated by the bucket function in the function catalog are correct.

zhongyujiang (commented):
I've updated to resolve the comments above, please take a look when you have time, cc @Aitozi @Zouxxyy Thanks!

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/catalog/functions/PaimonFunctions.scala
```diff
@@ -0,0 +1,138 @@
...
+  @Nullable
+  def load(name: String): UnboundFunction = FUNCTIONS.get(name)
+}
+
+/**
+ * A function returns the bucket ID based on the number of buckets and bucket keys.
+ *
+ * params arg0: bucket number, arg1...argn bucket keys.
+ */
+class BucketFunction extends UnboundFunction {
```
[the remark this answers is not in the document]
zhongyujiang: Hi @Zouxxyy, what I meant is that when creating the `SparkInternalRowWrapper`, we should pass in `inputType` instead of `bucketKeyStructType`, because the input row passed to this wrapper includes the `NUM_BUCKETS` column.
For `InternalRowSerializer`, we still pass in `bucketKeyStructType`, since as you mentioned, we’ve already removed `NUM_BUCKETS` using `ProjectedRow`.
 
image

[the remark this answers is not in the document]
Zouxxyy: Got it, make sence, thanks!

zhongyujiang:
@Zouxxyy rebased, please take another look when you have time, thanks!

Zouxxyy (approved):
+1, thanks

PR #5245, opened 2025-03-10
[build] Support e2e tests on jdk11

liyubin117:
### Purpose

After the Maven compilation on JDK11, the generated jar is also built with JDK11, accordingly, the JDK in the E2E container should also be JDK11, otherwise it will remain abnormal as it was. so we should:
1. Use `apache/flink:${test.flink.version}-${test.java.version}` instead of `apache/flink:${test.flink.version}-java8`,
2. Annotate `SparkE2eTest` and `HiveE2eTest` as @DisabledOnJre(JAVA_11) because jdk11 container not supported.

e2e-tests on jdk8: https://github.com/liyubin117/paimon/actions/runs/13760620698/job/38475787890
e2e-tests on jdk11: https://github.com/liyubin117/paimon/actions/runs/13760620697/job/38475787701


liyubin117:
@JingsongLi CI failed caused by Nullpointer in `RescaleAction`, and it has been fixed in master branch. PTAL, thanks :)

JingsongLi (approved):
Thanks @liyubin117  +1

PR #5249, opened 2025-03-10
[rest] Add TableSnapshot including basic statistics to SupportsSnapshots

JingsongLi:
### Purpose

Commit Snapshot with statistics, get snapshot should also contain basic statistics too.


wwj6591812:
+1

PR #5253, opened 2025-03-11
[core] do not overwrite options when merging options and token

plusplusjiajia:
### Purpose
When merging option and token in RESTTokenFileIO, the value in options should not be overwritten. DLF server will pass a default value, such as fs.oss.endpoint=oss-cn-hangzhou-internal.aliyuncs.com, which is not available in some client environment, and the value passed in options needs to be used.

### Tests
RESTUtilTest

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/rest/RESTTokenFileIO.java
+++ b/paimon-core/src/main/java/org/apache/paimon/rest/RESTTokenFileIO.java
@@ -153,7 +153,8 @@ public class RESTTokenFileIO implements FileIO {
 
             CatalogContext context = catalogLoader.context();
             Options options = context.options();
-            options = new Options(RESTUtil.merge(options.toMap(), token.token()));
+            // the original options are not overwritten
+            options = new Options(RESTUtil.merge(token.token(), options.toMap()));
             options.set(FILE_IO_ALLOW_CACHE, false);
             context = CatalogContext.create(options, context.preferIO(), context.fallbackIO());
             try {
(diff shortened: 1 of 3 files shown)


paimon-core/src/test/java/org/apache/paimon/rest/RESTUtilTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+/** Test for {@link RESTUtil}. */
+public class RESTUtilTest {
+    @Test
+    public void testMerge() {
+        {
+            Map<String, String> targets = new HashMap<>();
+            targets.put("a", "1");
```
sundapeng, line 34: Better to update the unit test more humanreadable, `1` to `default1` or `target1`

plusplusjiajia: > Better to make test more human-readable, for example, `1` to `default1` or `target1`

Good advice, done.

sundapeng:
+1

paimon-core/src/main/java/org/apache/paimon/rest/RESTTokenFileIO.java
```diff
@@ -153,7 +153,8 @@ private FileIO fileIO() throws IOException {
 
             CatalogContext context = catalogLoader.context();
             Options options = context.options();
-            options = new Options(RESTUtil.merge(options.toMap(), token.token()));
+            // the original options are not overwritten
+            options = new Options(RESTUtil.merge(token.token(), options.toMap()));
```
jerry-024, line 157: Is it possible to provide a separate method to get options from options and token, and add a test to protect this logic?

JingsongLi:
+1

paimon-core/src/main/java/org/apache/paimon/rest/RESTTokenFileIO.java
plusplusjiajia: > Is it possible to provide a separate method to get options from options and token, and add a test to protect this logic?

good idea! done.

PR #5254, opened 2025-03-11
[docs] [procedures] Fix the name of parameter for procedures

yangjf2019:
### Purpose

Fix the name of parameter for procedures documents.


JingsongLi (approved):
+1

PR #5255, opened 2025-03-11
[core] ensure that record-expire takes effect when full compaction

LsomeYeah:
### Purpose

Linked issue: close #xxx

Record-Level expiration happens in compaction, but some specific data files will not be rewritten even in full compaction which may leading some expired records can't be expired for a long time. This pr aims to ensure that record-expire takes effect when full compaction. 

When full compaction, the files with the following kinds will not be rewritten:

- large file, file level is maxLevel and no overlap with other file: do nothing for the file
- large file, outputLevel is not maxLevel or has no deleted records: directly upgrade file level to outputLevel without rewriting

For these files,  rewrite them if they contain expired records.

And full compaction will pick no files for compaction in the following cases:

- lsm tree is empty
- only one maxLevel sorted run in lsm tree

For only one maxLevel sorted run in lsm tree, pick the files containing expired records to perform full compaction.

### Tests

RecordLevelExpireTest#testIsExpireFile
RecordLevelExpireTest#testTotallyExpire

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/io/RecordLevelExpire.java
+++ b/paimon-core/src/main/java/org/apache/paimon/io/RecordLevelExpire.java
@@ -22,6 +22,12 @@ import org.apache.paimon.CoreOptions;
 import org.apache.paimon.KeyValue;
 import org.apache.paimon.data.InternalRow;
 import org.apache.paimon.reader.RecordReader;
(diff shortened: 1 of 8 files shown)


wwj6591812:
+1

wwj6591812:
Good work!

JingsongLi (changes requested):

JingsongLi (approved):
+1

PR #5256, opened 2025-03-11
[test] do not overwrite options when test format table

qingwei727:
Purpose
When merging option in test format table, the value in options should not be written by default value. DLF server cannot handle format table without external path, so the format table test options should be able to be overridden .

Tests
CatalogTestBase.testFormatTable


JingsongLi:
+1

PR #5261, opened 2025-03-12
[core] remove additional signed headers

plusplusjiajia:
### Purpose
Because the host header is easily overwritten after several forwarding, so remove it from add additional signed headers.
### Tests
DLFAuthSignatureTest


JingsongLi:
+1

PR #5268, opened 2025-03-12
[core] TagManager#createTag should check tag existence again when failed to create tag

yuzelin:
### Purpose

Linked issue: close #xxx

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/utils/TagManager.java
+++ b/paimon-core/src/main/java/org/apache/paimon/utils/TagManager.java
@@ -100,7 +100,7 @@ public class TagManager {
     public void createTag(
             Snapshot snapshot,
             String tagName,
-            Duration timeRetained,
+            @Nullable Duration timeRetained,
             List<TagCallback> callbacks,
             boolean ignoreIfExists) {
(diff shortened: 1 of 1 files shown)


JingsongLi:
Please add test.

paimon-core/src/test/java/org/apache/paimon/utils/TagManagerTest.java
```diff
@@ -136,6 +138,66 @@ public void testCreateTagWithTimeRetained() throws Exception {
         assertThat(tags.get(0).getValue()).contains("tag");
     }
 
+    @Test
+    public void testCreateTagWithMultiThreads() throws Exception {
```
JingsongLi, line 142: This method can not be failed in the master.

yuzelin: I've run it many times at my machine.

yuzelin: https://github.com/apache/paimon/blob/0bad12d4d04a6fb52bae2203964d7e70365a76fb/paimon-core/src/main/java/org/apache/paimon/utils/TagManager.java#L137

It used `overwriteFileUtf8` which will overwrite the tag file if some other procedure is creating the tag. And it won't throw exceptions. I think it's not safe?

PR #5269, opened 2025-03-12
[Bug] Improve float precision handling in TypeUtils.

### LINKED ISSUE #5263: [Bug] Paimon throws an exception "cannot be cast to float due to precision loss" when the float data has five decimal places or more.
### Search before asking

### Paimon version

paimon-1.0

### Compute Engine

Flink 1.17

### Minimal reproduce step

Insert any floating-point data with five or more decimal places into the Paimon table.

Thers is the code that cast string to float。 In fact, when a number has four decimal places, converting a float type to a String type will result in scientific notation. In this case, directly comparing their string lengths is meaningless. However, when the number has exactly four decimal places, the lengths happen to be the same (e.g. len(0.0001) = len(2.0E-4)). so when there are exactly faive decimal places, which will throw the exception.
```
case FLOAT:
                double d = Double.parseDouble(s);
                if (d == ((float) d)) {
                    return (float) d;
                } else {
                    // Compatible canal-cdc
                    Float f = Float.valueOf(s);
                    if (f.toString().length() != s.length()) {
                    } else {
                        return f;
                    }
                }
```

Image

Image

org.apache.paimon.utils.TypeUtils#castFromStringInternal() 

### Anything else?

_No response_

danzhewuju:
### Purpose

Linked issue: close #5263

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/utils/TypeUtils.java
+++ b/paimon-common/src/main/java/org/apache/paimon/utils/TypeUtils.java
@@ -144,7 +144,12 @@ public class TypeUtils {
                 } else {
                     // Compatible canal-cdc
                     Float f = Float.valueOf(s);
-                    if (f.toString().length() != s.length()) {
+                    String floatStr = f.toString();
+                    if (s.contains(".") && !s.contains("E")) {
+                        int decimal = s.length() - s.indexOf(".") - 1;
+                        floatStr = String.format("%." + decimal + "f", f);
+                    }
+                    if (!floatStr.equals(s)) {
                         throw new NumberFormatException(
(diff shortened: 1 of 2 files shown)


wwj6591812:
+1

danzhewuju:
> +1

Could you kindly review this PR when you have a moment? I’d really appreciate it. Thank you!

jixianyihao:
+1

leaves12138:
Could you please add a test for method you've changed?

danzhewuju:
> Could you please add a test for method you've changed?

ok, I will add some unit tests.

JingsongLi (approved):
+1

PR #5281, opened 2025-03-13
[core] RestCatalog: support ListDatabasesPaged

XiaoHongbo-Hope:
[core] RestCatalog: support ListDatabasesPaged and optimize mock server
### Tests
RESTCatalogTestBase.testListDatabases
RESTCatalogTestBase.testListDatabasesPaged and other original cases

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
@@ -133,6 +133,11 @@ public abstract class AbstractCatalog implements Catalog {
     }
 
     @Override
+    public PagedList<String> listDatabasesPaged(Integer maxResults, String pageToken) {
+        return new PagedList<>(listDatabases(), null);
+    }
+
+    @Override
     public void createDatabase(String name, boolean ignoreIfExists, Map<String, String> properties)
             throws DatabaseAlreadyExistException {
         checkNotSystemDatabase(name);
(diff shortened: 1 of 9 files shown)


paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
```diff
@@ -132,6 +132,11 @@ protected boolean allowCustomTablePath() {
         return false;
     }
 
+    @Override
+    public PagedList<String> listDatabasesPaged(Integer maxResults, String pageToken) {
```
JingsongLi, line 136: Please add Nullable to all places.

paimon-core/src/main/java/org/apache/paimon/rest/RESTCatalog.java
```diff
@@ -186,13 +185,28 @@ public RESTCatalogLoader catalogLoader() {
...
+    public PagedList<String> listDatabasesPaged(Integer maxResults, String pageToken) {
         ListDatabasesResponse response =
                 client.get(
-                        resourcePaths.databases(), ListDatabasesResponse.class, restAuthFunction);
-        if (response.getDatabases() != null) {
-            return response.getDatabases();
-        }
-        return ImmutableList.of();
+                        resourcePaths.databases(),
+                        buildPagedQueryParams(maxResults, pageToken),
```
JingsongLi, line 202: Please also modify open api yaml file and `RESTCatalogController`.

paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
XiaoHongbo-Hope: done

paimon-core/src/main/java/org/apache/paimon/rest/RESTCatalog.java
XiaoHongbo-Hope: done

JingsongLi (approved):
+1

PR #5283, opened 2025-03-13
[rest] Drop partition using filesystem in RESTCatalog

JingsongLi:
### Purpose

Submitting a commit to record the deletion of all files is a heavy operation that should be completed by the client.


wwj6591812:
+1

Issue #5213, opened 2025-03-05
[Feature] Support refresh existing dataFileMetas in CompactManager


xiangyuf:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Motivation

It is very usual user will use a streaming job and a batch job to update different columns of a paimon table. User can choose to configure the batch job as write-only or streaming job as write-only. 

In either way, the write and compact job will not receive the new files generated by write-only job. This will lead to insufficient compaction both for MergeTreeCompactManager and BucketedAppendCompactManager.

This issue could be addressed by supporting periodically refresh dataFileMetas in CompactManager.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

xiedeyantu:
Hi, can I try it?

xiangyuf:
@xiedeyantu thx for volunteering,  I'm already working on this.

xiedeyantu:
> @xiedeyantu thx for volunteering,  I'm already working on this.

OK, thanks for reply, and there are some easy issue to try?

xiangyuf:
> > [@xiedeyantu](https://github.com/xiedeyantu) thx for volunteering,  I'm already working on this.
> 
> OK, thanks for reply, and there are some easy issue to try?

Maybe you can try this. https://github.com/apache/paimon/issues/4244

PR #5284, opened 2025-03-14
[core] Support watermark partition markdone mode

### LINKED ISSUE #5043: [Feature] Use Parquet Merge in compaction for AppendOnlyTable
### Search before asking

### Motivation

See more in: https://issues.apache.org/jira/browse/PARQUET-382

This could save lots of resource usage for append only table's compaction.

### Solution

_No response_

### Anything else?

_No response_

xiangyuf:
### Purpose

Linked issue: close #5043

Use watermark to trigger partition markdone

### Tests

PartitionMarkDoneTest#testWaterMarkPartitionMarkDone


paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/sink/partition/PartitionMarkDoneTest.java
```diff
@@ -66,6 +69,107 @@ public void testNotTriggerByCompaction() throws Exception {
         innerTest(false);
     }
 
+    @Test
```
wwj6591812, line 72: I suggest that you also add a IT case like FlinkBatchJobPartitionMarkdoneITCase.

LinMingQiang: +1

xiangyuf: You were right. I was just kind of tired writing test code. Already write a lot in #5213  -_-!

xiangyuf: I wiil add ITCase later.

xiangyuf:
> You should ignore `partition.time-interval`, otherwise watermark will not be able to properly determine when the partition should be mark done.

@LinMingQiang Thx for advice. I don't think the `partition.time-interval` will affect the watermark partition mark done behavior here. Can you explain more here?

LinMingQiang:
> > You should ignore `partition.time-interval`, otherwise watermark will not be able to properly determine when the partition should be mark done.
> 
> @LinMingQiang Thx for advice. I don't think the `partition.time-interval` will affect the watermark partition mark done behavior here. Can you explain more here?

e.g.  partition value is '2025-03-10', watermark is '2025-03-11'.   Should partition '2025-03-10' be marked as done?
If not, then there is no problem here.

LinMingQiang:
> > > > You should ignore `partition.time-interval`, otherwise watermark will not be able to properly determine when the partition should be mark done.
> > > 
> > > 
> > > @LinMingQiang Thx for advice. I don't think the `partition.time-interval` will affect the watermark partition mark done behavior here. Can you explain more here?
> > 
> > 
> > e.g. `partition.time-interval = 2 d` ，partition value is '2025-03-10', watermark is '2025-03-11'. Should partition '2025-03-10' be marked as done? The current logic is not.
> > Related code :
> > ```
> > Map.Entry<String, Long> entry = iter.next();
> >             String partition = entry.getKey();
> > 
> >             long lastUpdateTime = entry.getValue();
> >             long partitionStartTime =
> >                     extractDateTime(partition)
> >                             .atZone(ZoneId.systemDefault())
> >                             .toInstant()
> >                             .toEpochMilli();
> >             long partitionEndTime = partitionStartTime + timeInterval;
> >             lastUpdateTime = Math.max(lastUpdateTime, partitionEndTime);
> > 
> >             if (currentTimeMillis - lastUpdateTime > idleTime) {
> >                 needDone.add(partition);
> >                 iter.remove();
> >             }
> > ```
> 
> In your case, I think partition '2025-03-10' should be mark done after watermark pass '2025-03-12'.

👌.

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/sink/partition/PartitionMarkDoneTest.java
xiangyuf: Added E2E test in WatermarkPartitionMarkDoneTest#testWaterMarkPartitionMarkDone

xiangyuf:
@JingsongLi Ready for review

JingsongLi:
+1

PR #5287, opened 2025-03-14
[flink] make tableConf valid for database compaction job

LsomeYeah:
### Purpose

Linked issue: close #xxx

Currently, tables are dynamically loaded when performing dedicated database compaction, and table conf set by user will not be passed to these tables in some stages, e.g., source stage and write stage will not pass the user table conf. This pr aims to fix it.


wwj6591812:
+1

JingsongLi (commented):
Can you add a test case? Maybe you can control file format?

yunfengzhou-hub (commented):
+1 except for the tests.

LsomeYeah:
> Can you add a test case? Maybe you can control file format?

test had been added~

JingsongLi:
+1

PR #5292, opened 2025-03-15
[hotfix] incremental-between-timestamp should return empty data instead of throwing exception

yuzelin:
### Purpose

Linked issue: close #xxx


JackeyLee007:
+1

PR #5297, opened 2025-03-17
[hotfix] Fix reduce scan.manifest.parallelism not take effect

askwang:
### Purpose

When reduce `scan.manifest.parallelism` parameter, it does not take effect.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/utils/ManifestReadThreadPool.java
+++ b/paimon-core/src/main/java/org/apache/paimon/utils/ManifestReadThreadPool.java
@@ -36,7 +36,7 @@ public class ManifestReadThreadPool {
             createCachedThreadPool(Runtime.getRuntime().availableProcessors(), THREAD_NAME);
 
     public static synchronized ThreadPoolExecutor getExecutorService(@Nullable Integer threadNum) {
-        if (threadNum == null || threadNum <= executorService.getMaximumPoolSize()) {
+        if (threadNum == null || threadNum >= executorService.getMaximumPoolSize()) {
             return executorService;
         }
         // we don't need to close previous pool


paimon-core/src/main/java/org/apache/paimon/utils/ManifestReadThreadPool.java
```diff
@@ -36,7 +36,7 @@ public class ManifestReadThreadPool {
             createCachedThreadPool(Runtime.getRuntime().availableProcessors(), THREAD_NAME);
 
     public static synchronized ThreadPoolExecutor getExecutorService(@Nullable Integer threadNum) {
-        if (threadNum == null || threadNum <= executorService.getMaximumPoolSize()) {
+        if (threadNum == null || threadNum >= executorService.getMaximumPoolSize()) {
```
JingsongLi, line 39: If you want increase thread Num, this becomes not work now...

askwang: maybe we can remove check `threadNum <= executorService.getMaximumPoolSize()` condition? increase or reduce both ok.

JingsongLi: Why do you need to reduce the number of threads?

askwang: We find there are 64 manifest thread when doing full manifest merge, it takes up a lot of memory, we want to relieve it by reducing the thream num.

JingsongLi: Maybe when the thread number is less than max pool size, you can introduce a `SemaphoredDelegatingExecutor` to wrap shared `Executor`.

JingsongLi:
@askwang Hi, can you take a look to this PR: https://github.com/apache/paimon/pull/5737

PR #5299, opened 2025-03-17
[core][rest] Remove partition modification method in RESTCatalog

JingsongLi:
### Purpose

Partition information can be calculated in commit snapshot interface, so partition does not need to be calculated by catalog interface.

RESTCatalog only provide list partitions and mark done partitions for partition metadata only.


vox-vox-vox:
+1

PR #5301, opened 2025-03-17
[core] Support MULTI_WRITE mode for KeyValueFileStoreWrite when multiple jobs are writing to one table simultaneously…

### LINKED ISSUE #5214: [Feature] Support level refresh in MergeTreeCompactManager
### Search before asking

### Motivation

subtask of #5213 

### Solution

_No response_

### Anything else?

_No response_

xiangyuf:
### Purpose

Linked issue: close #5214 

### Tests

MergeTreeCompactManagerTest#testSyncOrphanFiles()
KeyValueFileStoreWriteTest#testMultiWriteModeEnabled()
KeyValueFileStoreWriteTest#testMultiWriteModeDisabled()


paimon-core/src/test/java/org/apache/paimon/operation/KeyValueFileStoreWriteTest.java
```diff
@@ -102,33 +125,166 @@ public void testGentleLookupCompactStrategy() throws Exception {
...
-                        .build();
+        Map<String, String> options2 = new HashMap<>();
+        options2.put(CoreOptions.WRITE_ONLY.key(), "true");
+
+        TestFileStore store1 = createStoreWithOptions(options1);
+        TestFileStore store2 = createStoreWithOptions(options2);
+
+        KeyValueFileStoreWrite write1 = (KeyValueFileStoreWrite) store1.newWrite(user1);
+        KeyValueFileStoreWrite write2 = (KeyValueFileStoreWrite) store2.newWrite(user2);
+
```
LinMingQiang, line 144: I tried adding a writer-3 in this test, the result is not as expected, the file written by the writer-1 is not be compacted.

xiangyuf: writer-1 is the write and compact job, writer-3 should be write only. The file added by writer-3 should be compacted by writer-1.

LinMingQiang: Sorry, my expression is wrong.  the file written by the writer-2 is not be compacted. write-1 and write-3 will be compacted.

xiangyuf: @LinMingQiang There is a debug in the previous implementable. Pls check the latest version. Also I've added the writer-3 in the test.

xiangyuf:
@LinMingQiang I've discussed with @JingsongLi about this feature. We need to avoid continuously file scan in writer. So this pr will only be a PoC temporarily.

TejMakode1523:
Hello @xiangyuf , just checking if there has been any progress on the multi-write mode POC.

PR #5303, opened 2025-03-18
[fs] Add JindoFileIO access to OSS without dependency

JingsongLi:
### Purpose

Some environments already have dependencies on Jindo, such as EMR environments, and we can directly provide a Jindo FileIO to access OSS.


paimon-filesystems/paimon-jindo/src/main/resources/META-INF.services/org.apache.paimon.fs.FileIOLoader
shyjsarah, line 1: META-INF.services => META-INF/services

shyjsarah:
+1

PR #5305, opened 2025-03-18
[core] OrphanFilesClean: add log change log path when change log is empty

jerry-024:
### Purpose

OrphanFilesClean: add log change log path when change log is empty


JingsongLi:
+1

PR #5309, opened 2025-03-18
[core] Correct bitmap types support in writer

JingsongLi:
### Purpose

1. We should copy bytes for `BinaryString`.
2. We should not support other types in writer.


leaves12138:
Please add a test

hang8929201 (approved):
+1

PR #5314, opened 2025-03-19
[cdc] Provide the document for computed columns with kafka_sync_database

JackeyLee007:
### Purpose

Just add the document for computed columns with kafka_sync_database.
Also provide the document of the new computing expression now, which get the process-time when ingesting the data.


JingsongLi:
+1

PR #5319, opened 2025-03-20
[core][flink] introduce changelog expire procedure and action

LsomeYeah:
### Purpose

Linked issue: close #xxx

Feature https://github.com/apache/paimon/issues/2899 enables to decouple the lifecycle of snapshot and changelog. The reasons of introducing changelog expire procedure are as follows:

1. `ExpireSnapshotsProcedure` will only expire snapshots. If their lifecycle is decoupled, the previous changelog can not be expired with `ExpireSnapshotsProcedure`.
2. If  a table decoupled the lifecycle at initially, but later changed to not decoupled, there would be some changelogs  which can not be expired anymore. Here is a simple example:

```
First:
snapshot.num-retained.min = 4, changelog.num-retained.min = 10 (lifecycle decoupled)
Snapshot: [7, 8, 9, 10]
Changelog: [1, 2, 3, 4, 5, 6]

Later:
snapshot.num-retained.min = 4, changelog.num-retained.min = 4 (lifecycle not decoupled)
insert some data 10 times and generate 10 new snapshots:
Snapshot: [18, 19, 20, 21] (snapshot#7 - #17 expired)
Changelog: [1, 2, 3, 4, 5, 6] (changelog#1 - #6 not expired, as lifecycle is not decoupled and ExpireChangelogImpl.expire won't be invoked)
```

Commonly, when expiring changelogs, the latest changelog will be protected and not be expired. So to handle the case mentioned above( the latest changelog is out-of-date),  a new argument `delete_all` has been introduced. When this argument is set to true, all changelogs including changelog hint files will be deleted.

### Tests

ExpireChangelogsProcedureITCase

### DIFF
--- a/docs/content/flink/procedures.md
+++ b/docs/content/flink/procedures.md
@@ -601,6 +601,45 @@ All available procedures are listed below.
          CALL sys.expire_snapshots(`table` => 'default.T', older_than => '2024-01-01 12:00:00', max_deletes => 10)<br/><br/>
       </td>
    </tr>
+   <tr>
+      <td>expire_changelogs</td>
+      <td>
+         -- Use named argument<br/>
+         CALL [catalog.]sys.expire_changelogs(<br/>
+            `table` => 'identifier', <br/>
+            retain_max => 'retain_max', <br/>
+            retain_min => 'retain_min', <br/>
+            older_than => 'older_than', <br/>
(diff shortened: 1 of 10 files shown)


JingsongLi (approved):
Looks good to me!

PR #5321, opened 2025-03-20
[core] Support diff scan mode in incremental query

yuzelin:
### Purpose

To support Read snapshot diff like tag diff.

TODO:
1. is #5322  bug?


JingsongLi (approved):
+1

PR #5325, opened 2025-03-21
[rest] Add path and other options to Table Options

JingsongLi:
### Purpose

Expose more information to REST client:
1. path information.
2. owner, updateBy, createBy...


vox-vox-vox:
+1 🚀

PR #5326, opened 2025-03-21
[core] Add supportsVersionManagement to Catalog

JingsongLi:
### Purpose

Avoid using catching UnsupportedException in codes.


vox-vox-vox:
+1 🚀

PR #5327, opened 2025-03-21
[spark] Add checker for PaimonSparkSessionExtensions

Zouxxyy:
### Purpose

`PaimonSparkSessionExtensions` has been used in many aspects, such as analyze, row-level commands, procedures, dynamic partition overwrite and so on. In other formats like Delta and Hudi, it has always been a required option and now we are adding validation for it in paimon.

 However, if you really do not want to add it, you can also set `spark.paimon.requiredSparkConfsCheck.enabled` to `false` to bypass the validation, but it is strongly discouraged to do so.


JingsongLi (approved):
+1

PR #5329, opened 2025-03-22
[cdc] Do not check referenced field if no physical fields provided when computing columns

JackeyLee007:
### Purpose

Linked issue: close #xxx

This PR is for computing columns when using kafka_sync_database. There would be multiple different tables to sync, so it's hard to list the exact data fields. No data fields provided would stop the buidling of computing column expressions.

This PR ignores the check when building expressions for kafka_sync_database, since it provides empty_list as the data fields. And it postpones the check to evaluation, which returns null if the referenced field doesn't exist.

### DIFF
--- a/paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/action/cdc/Expression.java
+++ b/paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/action/cdc/Expression.java
@@ -23,6 +23,7 @@ import org.apache.paimon.types.DataType;
 import org.apache.paimon.types.DataTypeFamily;
 import org.apache.paimon.types.DataTypeJsonParser;
 import org.apache.paimon.types.DataTypes;
+import org.apache.paimon.types.VariantType;
 import org.apache.paimon.utils.DateTimeUtils;
 import org.apache.paimon.utils.SerializableSupplier;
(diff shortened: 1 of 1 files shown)


JingsongLi (commented):
Reasonable fix for `kafka_sync_database` with computing columns.

**The problem**: When syncing multiple tables via `kafka_sync_database`, physical fields aren't known upfront (they vary per table). The expression builder checks referenced fields against the provided physical fields — but an empty list causes the check to fail, blocking all computing column expressions.

**The fix**: Skip the referenced-field check when physical fields list is empty, deferring validation to evaluation time (returns null for non-existent fields).

**Review:**

1. **+8/-5** is minimal. The change is in `Expression.java`.

2. **Safety**: Deferring the check to evaluation time means invalid expressions won't fail fast. Users won't know their computing column references a non-existent field until data flows. This is acceptable for the `kafka_sync_database` use case (where fields are dynamic), but should NOT affect single-table sync (where fields are known upfront).

3. **Test coverage**: No tests added. Consider adding a test that verifies:
   - Empty physical fields → expression builds successfully
   - Non-empty physical fields with invalid reference → still fails at build time

4. **This PR is quite old** (opened well before batch 5). Is it still relevant to current master? The Expression class may have changed.

Please verify this still applies cleanly to master.

PR #5332, opened 2025-03-24
[test] Fixed incorrect use of FileIO in CloneActionITCase

GreatEugenius:
### Purpose

Fixes the bug that targetFileIO is used to read sourceTable's content in CloneActionITCase.


JingsongLi:
+1

PR #5338, opened 2025-03-25
[test] Static import of assertion classes

yangjf2019:
### Purpose

Improve the introduction of assertions in test cases,and make code more concise and improve readability.


paimon-common/src/test/java/org/apache/paimon/sort/zorder/TestZOrderByteUtil.java
```diff
@@ -108,7 +109,7 @@ public void testInterleaveRandomExamples() {
 
             String stringResult = interleaveStrings(testStrings);
 
-            Assert.assertEquals(
+            assertEquals(
```
JingsongLi, line 112: Can you modify these junit4 assertion to juni5 assertThat?

yangjf2019: Thanks for your review, Of course.

yangjf2019: @JingsongLi  done.

JingsongLi (approved):
+1

PR #5340, opened 2025-03-25
[rest] Fix content-security-policy violation issue on RestCatalog API

liyubin117:
### Purpose

Apache refuses to load the https://cdn.jsdelivr.net/npm/redoc@next/bundles/redoc.standalone.js script due to Content Security Policy (CSP) restrictions. We could import the file, and it also makes loading faster.
As mentioned in https://github.com/Redocly/redoc/pull/1191, we should disable search to avoid
```
Refused to create a worker from 'blob:http://localhost:1313/5cb97201-bf84-45b4-ae92-3814a5085396' because it violates the following Content Security Policy directive: "script-src 'self' 'unsafe-inline' 'unsafe-eval'". Note that 'worker-src' was not explicitly set, so 'script-src' is used as a fallback.
```

javascript file refer to https://cdn.jsdelivr.net/npm/redoc@2.0.0/bundles/redoc.standalone.js


LICENSE
```diff
@@ -297,6 +297,7 @@ MIT License
 
 docs/assets/github.css (css generated by Hugo) from https://github.com/alecthomas/chroma version 2
 docs/static/js/anchor.min.js from https://github.com/bryanbraun/anchorjs version 5.0
+docs/static/js/redoc.standalone.js from https://github.com/Redocly/redoc version 2.0.0
```
JingsongLi, line 300: We should add its license to `licenses` directory.

docs/content/concepts/rest/rest-api.md
```diff
@@ -23,6 +23,11 @@ under the License.
 -->
 
 <body>
-<redoc spec-url="https://raw.githubusercontent.com/apache/paimon/master/paimon-open-api/rest-catalog-open-api.yaml"></redoc>
-<script src="https://cdn.jsdelivr.net/npm/redoc@next/bundles/redoc.standalone.js"></script>
+    <div id="redoc-container"></div>
+    <script src="/docs/master/js/redoc.standalone.js"></script>
```
JingsongLi, line 27: Refer it just like `"{{.Site.BaseURL}}/js/flink.js"`?

liyubin117: After setting <script src="{{.Site.BaseURL}}/js/redoc.standalone.js"></script>, throw following error:
```
GET http://localhost:1313/docs/master/concepts/rest/rest-api/%7B%7B.Site.BaseURL%7D%7D/js/redoc.standalone.js net::ERR_ABORTED 404 (Not Found)
```

JingsongLi: Maybe add this `<script` to head.html, and this file just refer js function?

licenses/LICENSE.redoc
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
```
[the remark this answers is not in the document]
liyubin117: Got it

JingsongLi:
+1

PR #5341, opened 2025-03-25
[test] Enhance inheritance and visibility for ITCase support

GreatEugenius:
### Purpose

- Changed the access modifiers of certain methods in Action and Procedure from private to protected:
    - This change enables subclasses to inherit and reuse logic for ITCase scenarios.
    - It also supports future extensibility and customization.

-  Updated the following classes to public access:
    - BranchActionITCase
    - This change allows the classes to be accessed across different packages, improving code reusability and modularity.

### DIFF
--- a/paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/CatalogITCaseBase.java
+++ b/paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/CatalogITCaseBase.java
@@ -109,7 +109,7 @@ public abstract class CatalogITCaseBase extends AbstractTestBase {
         return false;
     }
 
-    private void prepareEnv() {
+    protected void prepareEnv() {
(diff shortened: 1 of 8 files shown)


JingsongLi:
Hi @GreatEugenius , the purpose of this PR is for future use? Should we delay the changes until specific future requirements emerge?

GreatEugenius:
> Hi @GreatEugenius , the purpose of this PR is for future use? Should we delay the changes until specific future requirements emerge?
Hi @JingsongLi, the purpose of this PR is not intended for future use but rather to address the needs of an internal project that requires these changes for inheritance purposes. The modifications are necessary to support current requirements within our internal ecosystem, ensuring alignment with existing workflows and project dependencies. Delaying the changes would hinder progress on the internal project, so it is recommended to proceed with the implementation now.

JingsongLi:
+1

PR #5345, opened 2025-03-25
Add bucket statistics in ManifestFileMeta

tsreaper:
### Purpose

This PR adds bucket statistics in `ManifestFileMeta`. This PR also optimizes `withBucket` filter in `FileStoreScan` with this new statistics.

### Tests

* `ManifestFileMetaSerializerTest`.

### API and Format

`ManifestFileMeta` is changed.

### Documentation

No new feature.


JingsongLi:
Maybe we should add level stats too.

paimon-core/src/main/java/org/apache/paimon/operation/ManifestsReader.java
```diff
@@ -123,6 +129,14 @@ private List<ManifestFileMeta> readManifests(Snapshot snapshot, ScanMode scanMod
 
     /** Note: Keep this thread-safe. */
     private boolean filterManifestFileMeta(ManifestFileMeta manifest) {
+        Integer minBucket = manifest.minBucket();
+        Integer maxBucket = manifest.maxBucket();
+        if (specifiedBucket != null && minBucket != null && maxBucket != null) {
```
JingsongLi, line 134: Here need to handle `onlyReadRealBuckets` too.

paimon-core/src/main/java/org/apache/paimon/manifest/ManifestFileMetaSerializer.java
```diff
@@ -35,7 +35,7 @@ public ManifestFileMetaSerializer() {
 
     @Override
     public int getVersion() {
-        return 2;
+        return 3;
```
JingsongLi, line 38: This version cannot be changed, this PR is a compatible changing, we should keep old version for old readers.

Can you add tests for this?

JingsongLi (approved):
+1

PR #5347, opened 2025-03-26
[core] RESTCatalog: Optimize ErrorResponse to support more ResourceType

shyjsarah:
### Purpose

Optimize ErrorResponse to support more ResourceType

Linked issue: close #xxx


paimon-core/src/main/java/org/apache/paimon/rest/responses/ErrorResponse.java
```diff
@@ -75,10 +71,14 @@ public String getMessage() {
     }
 
     @JsonGetter(FIELD_RESOURCE_TYPE)
-    public ErrorResponseResourceType getResourceType() {
+    public String getResourceType() {
         return resourceType;
     }
 
+    public ErrorResponseResourceType getResourceTypeEnum() {
```
JingsongLi, line 78: This looks weird, can we always use String? We can introduce string constants in this class.

paimon-core/src/main/java/org/apache/paimon/rest/DefaultErrorHandler.java
```diff
@@ -51,13 +51,13 @@ public void accept(ErrorResponse error) {
                 throw new ForbiddenException("Forbidden: %s", message);
             case 404:
                 throw new NoSuchResourceException(
-                        error.getResourceType(), error.getResourceName(), "%s", message);
+                        error.getResourceTypeEnum(), error.getResourceName(), "%s", message);
             case 405:
```
jerry-024, line 59: we could use ErrorResponseResourceType.fromString(resourceType) get enum.

JingsongLi:
+1

PR #5353, opened 2025-03-26
[core] Fix one record cannot fit into a single page in ParallelExecution

JingsongLi:
### Purpose

Now, if a record is too big, in `ParallelExecution`, there will be a busy loop.

We should throw exception to alert user to increase `page-size` table option.


yuzelin (approved):
+1

PR #5355, opened 2025-03-27
[rest] Fix content-security-policy violation on RestCatalog API yaml

liyubin117:
### Purpose

Apache webserver specifies a CSP in the HTTP header, making the CSP defined in the `<meta>` tag invalid. Therefore, we should not access rest-catalog-open-api.yaml via an external link.
image


liyubin117:
@JingsongLi I have used csp in my local environment to test the feature and work fine, PTAL, thanks!
```
 <meta
          http-equiv="Content-Security-Policy"
          content="default-src 'self' data: blob: 'unsafe-inline' https://www.apachecon.com/ https://www.communityovercode.org/ https://analytics.apache.org/; script-src 'self' 'unsafe-inline' 'unsafe-eval' https://www.apachecon.com/ https://*.apache.org/ https://apache.org/; style-src 'self' 'unsafe-inline' data: https://*.apache.org/ https://apache.org/; frame-ancestors 'self'; frame-src 'self' data: blob:; img-src 'self' data: https://*.apache.org/ https://apache.org/ https://www.apachecon.com/; worker-src 'self' data: blob:;">
```

JingsongLi:
Looks good to me!

PR #5356, opened 2025-03-27
[rest] Separate rest management api yaml

JingsongLi:
### Purpose

Rest Management API use catalog name, but Rest catalog open api use catalog id, they are different, so introduce a separate page for rest management API.


jerry-024:
+1

PR #5358, opened 2025-03-27
[WIP][test] Remove this `public` modifier for junit5 Test

yangjf2019:
### Purpose

JUnit5 is more tolerant regarding the visibilities of Test classes than JUnit4, which required everything to be public.So, we can consider removing this `public` modifier for all the junit5 Test.
Refer to this [link](https://junit.org/junit5/docs/current/user-guide/#writing-tests-test-interfaces-and-default-methods).


paimon-arrow/src/test/java/org/apache/paimon/arrow/vector/ArrowFormatWriterTest.java
```diff
@@ -204,7 +204,7 @@ public void testCWriter() {
 
     @ParameterizedTest
     @ValueSource(booleans = {false, true})
-    public void testWriteWithExternalAllocator(boolean allocationFailed) {
+    void testWriteWithExternalAllocator(boolean allocationFailed) {
```
yunfengzhou-hub, line 207: I noticed that there are more test cases in this class. Why is only this test case's visibility is changed?

yangjf2019: I may have missed them, I'm checking them one by one.

yangjf2019:
closed

PR #5362, opened 2025-03-27
[core] RestCatalog: add getCatalogById in rest yaml

XiaoHongbo-Hope:
add getCatalogById in rest open api yaml 
### API and Format
rest-management-api.yaml


JingsongLi (approved):
+1

PR #5364, opened 2025-03-27
[core] Reduce get partition requests in migrating hive table

Zouxxyy:
### Purpose

Reduce get partition requests in migrating hive table


JingsongLi (approved):
+1

PR #5369, opened 2025-03-28
[rest] Fix RESTCatalogController ambiguous mapping between listViewDetails and listViews method

hadoopkandy:
Caused by: java.lang.IllegalStateException: Ambiguous mapping. Cannot map 'RESTCatalogController' method 
org.apache.paimon.open.api.RESTCatalogController#listViewDetails(String, String, Integer, String)
to {GET [/v1/{prefix}/databases/{database}/views]}: There is already 'RESTCatalogController' bean method
org.apache.paimon.open.api.RESTCatalogController#listViews(String, String, Integer, String) mapped.

### Purpose
just as rest-catalog-open-api.yaml described,listViewDetails method should be mappinged into '/v1/{prefix}/databases/{database}/view-details' 

Linked issue: close #xxx


JingsongLi:
+1

PR #5375, opened 2025-03-31
[core] Optimize equals and hashCode implementation in Snapshot

Nivane:
### Purpose

Optimize equals and hashCode implementation in Snapshot


JingsongLi (approved):
+1

PR #5377, opened 2025-03-31
[rest] Redoc page uses UTF-8 charset

liyubin117:
### Purpose

There are charset issues in redoc page
image


JingsongLi:
Sorry, I missed your PR, already fixed in https://github.com/apache/paimon/pull/5380

liyubin117:
@JingsongLi Good to hear that more issues fixed in time :)

PR #5378, opened 2025-03-31
Do not throw exception of fall back branch is not exist

leaves12138:
### Purpose

There is no need to force an exception if fall-back branch is not exist.


JingsongLi:
Already merged...

PR #5380, opened 2025-04-01
[core] fix rest api doc charset and tab's icon and title

jerry-024:
### Purpose

fix rest api doc charset and tab's icon and title


JingsongLi (approved):
+1

PR #4780, opened 2024-12-26
[parquet] Support using file index result to filter row ranges

Tan-JiaLiang:
### Purpose

Support using file index result to filter row ranges.

### Tests

org.apache.paimon.table.AppendOnlyFileStoreTableTest#testBitmapIndexResultFilterParquetRowRanges

### API and Format

No.

### Documentation

No.

PR #5385, opened 2025-04-01
[parquet] Fix that ColumnIndexFilter return empty mistakenly when projection doesn't contain filter fields

yuzelin:
### Purpose

Fix bug of #4780 
Why the bug occurs:
ColumnIndexFilter#applyPredicate: 
```
if (!columns.contains(columnPath)) {
    return rangesForMissingColumns;
}
```
It returns empty range, then VectorizedParquetRecordReader:
```
this.totalRowCount = reader.getFilteredRecordCount(); // always 0
```
Finally, VectorizedParquetRecordReader#nextBatch:
```
if (rowsReturned >= totalRowCount) { // 0 >= 0
    return false;
}
```


JingsongLi:
What compute engine?

JingsongLi:
Paimon should only consider the fields after projection when converting to Parquet Filters, and discard other filters.

PR #5390, opened 2025-04-02
[core] Fix Identifier should parse backquote.

LinMingQiang:
When using flink procedure,  table name can not parsed correctly when use `` quote. 

But Spark request  use `` for SYSTEM_TABLE_SPLITTER ($) .  So this is incompatible with Spark.

e.g :
```
Flink is OK But Spark is ERR  :  CALL sys.create_branch('default.T$branch_branch1', ......)

Flink is ERR But Spark is OK :  CALL sys.create_branch('default.`T$branch_branch1`', ......)

Flink will parse `T$branch_branch1` to table is `T  and branch is branch_branch1`.
```
image

Why :

Flink use `org.apache.paimon.catalog.Identifier` to parse the `tableName` and miss processing backquote,
 spark use `spark.sessionState().sqlParser().parseMultipartIdentifier` to parse.

### Purpose

Linked issue: close #xxx

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/catalog/Identifier.java
+++ b/paimon-core/src/main/java/org/apache/paimon/catalog/Identifier.java
@@ -80,7 +80,10 @@ public class Identifier implements Serializable {
             @JsonProperty(FIELD_DATABASE_NAME) String database,
             @JsonProperty(FIELD_OBJECT_NAME) String object) {
         this.database = database;
-        this.object = object;
+        this.object =
+                object.startsWith("`") && object.endsWith("`")
+                        ? object.substring(1, object.length() - 1)
+                        : object;
     }
 
     public Identifier(String database, String table, @Nullable String branch) {
--- a/paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/BranchSqlITCase.java
+++ b/paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/BranchSqlITCase.java
@@ -173,6 +173,34 @@ public class BranchSqlITCase extends CatalogITCaseBase {
     }
 
     @Test
+    public void testCreateBranchWithBackQuote() {
+        sql(
+                "CREATE TABLE T ("
+                        + " pt INT"
+                        + ", k INT"
+                        + ", v STRING"
+                        + ", PRIMARY KEY (pt, k) NOT ENFORCED"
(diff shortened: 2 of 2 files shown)


liyubin117:
I doubt whether the feature is necessary, because Flink regards `example.view` as object name. refer to https://nightlies.apache.org/flink/flink-docs-release-1.20/docs/dev/table/common/#expanding-table-identifiers

liyubin117:
Thanks for your reply, we may need to consider the following scenairos.
image

LinMingQiang:
> Thanks for your reply, we may need to consider the following scenairos. image image

In your example, you create a table called  table\`#branched.  This is no problem in Flink-SQL parsing engine, as the backticks are handled correctly. But it is not processed in procedure, so the correct table name table\`#branched cannot be parsed. This is also the purpose of this PR, here we can't just do a simple replacement, we need a parser to parse the table name like `SparkUtils#catalogAndIdentifier` does .

gmdfalk:
> > Thanks for your reply, we may need to consider the following scenairos. image image
> 
> In your example, you create a table called table`#branched. This is no problem in Flink-SQL parsing engine, as the backticks are handled correctly. But it is not processed in procedure, so the correct table name table`#branched cannot be parsed. This is also the purpose of this PR, here we can't just do a simple replacement, we need a parser to parse the table name like `SparkUtils#catalogAndIdentifier` does .

I believe you also cannot do a simple replacement like currently if you want to support any entity (catalog, database or table) with special characters in it.
For instance if the database name contains a dot:
```
CALL sys.compact(`table` => '`database_name.containing_dot`.baz');
``

LinMingQiang:
> > > Thanks for your reply, we may need to consider the following scenairos. image image
> > 
> > 
> > In your example, you create a table called table`#branched. This is no problem in Flink-SQL parsing engine, as the backticks are handled correctly. But it is not processed in procedure, so the correct table name table`#branched cannot be parsed. This is also the purpose of this PR, here we can't just do a simple replacement, we need a parser to parse the table name like `SparkUtils#catalogAndIdentifier` does .
> 
> I believe you also cannot do a simple replacement like currently if you want to support any entity (catalog, database or table) with special characters in it. For instance if the database name contains a dot:
> 
> ```
> CALL sys.compact(`table` => '`my_catalog`.`my_database.containing_dot`.my_table');
> ```

Sure, @JingsongLi  WDYT. Do we need to implement a parser to parse the table name like spark procedure do?

PR #5391, opened 2025-04-02
[core] Postpone bucket writing job should not expire snapshots

tsreaper:
### Purpose

Postpone bucket writing job should not expire snapshots. Snapshot expiration should be done in the compaction job.

This PR fixes the issue.

### Tests

* `PostponeBucketTableITCase#testPostponeWriteNotExpireSnapshots`.

### API and Format

No format changes.

### Documentation

No new feature.


JingsongLi (approved):
+1

PR #5392, opened 2025-04-02
[spark] Support scan mode when query with incremental_between

JackeyLee007:
### Purpose

Linked issue: close #xxx

When querying with paimon_incremental_between_timestamp, we want to switch among different scan modes.
* delta or changelog, if every single change is needed. 
* diff, if merge is needed.

In our micro/small batch operation, executed in hourly, we need to know when a record INSERTEd, and also when it's DELETEd. This needs deleta or changelog mode. If with diff mode, the +I and -D operation could be merged, then we won't get the -D operation. 

But when quering the main table, not the audit_log table, the merged result is expected, to just get the INSERTEd and UPDATEd records. So we also need the diff mode.

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/catalyst/plans/logical/PaimonTableValuedFunctions.scala
+++ b/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/catalyst/plans/logical/PaimonTableValuedFunctions.scala
@@ -21,8 +21,9 @@ package org.apache.paimon.spark.catalyst.plans.logical
 import org.apache.paimon.CoreOptions
 import org.apache.paimon.spark.SparkTable
 import org.apache.paimon.spark.catalyst.plans.logical.PaimonTableValuedFunctions._
-import org.apache.paimon.table.{DataTable, FileStoreTable}
+import org.apache.paimon.table.DataTable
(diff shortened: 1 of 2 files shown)


Zouxxyy:
Can you try setting spark.paimon.incremental-between-scan-mode = xx

JingsongLi (commented):
Useful feature. The ability to switch between delta/changelog/diff modes when using `paimon_incremental_between_timestamp` enables different consumption patterns:
- `delta`/`changelog`: every change visible (needed for auditing)
- `diff`: merged view (needed for batch processing)

**Review:**

1. **+46/-11** — adds a `mode` parameter to the table-valued function. Clean.

2. **Implementation**: The change is in `PaimonTableValuedFunctions.scala`. How does the mode parameter propagate to the underlying scanner? It should configure the `SnapshotReader` accordingly.

3. **Test**: The test in `TableValuedFunctionsTest.scala` should cover:
   - Default behavior (unchanged from before)
   - Explicit `delta` mode
   - Explicit `diff` mode
   - Invalid mode value → clear error

4. **Documentation**: No docs update in this PR. The Spark procedures/functions documentation should mention the new `mode` parameter.

5. **Naming**: Is `mode` the right parameter name? Consider `scan_mode` or `incremental_mode` to be more specific and align with existing options.

LGTM conceptually. Please add docs and verify the default behavior hasn't changed.

PR #5394, opened 2025-04-03
[core] make rowType not null for KeyValue and AppendOnlyFileStore

LsomeYeah:
### Purpose

Linked issue: close #xxx

rowType in KeyValue and AppendOnlyFileStore is used for read and write, make it not null.


JingsongLi (approved):
+1

PR #5396, opened 2025-04-03
[improve] As for `TimeTravel`, if the expected timestamp is earlier than the earliest snapshot, add one parameter to decide to return null or not

leaves12138:
### Purpose

Return null if time travel meet earlier timestamp than the earliest snapshot timestamp when set flag.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/source/snapshot/TimeTravelUtil.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/source/snapshot/TimeTravelUtil.java
@@ -159,7 +159,8 @@ public class TimeTravelUtil {
         }
 
         if (earliestSnapshot.timeMillis() >= timestampMills) {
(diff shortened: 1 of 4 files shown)


paimon-core/src/main/java/org/apache/paimon/table/source/snapshot/TimeTravelUtil.java
```diff
@@ -146,7 +146,8 @@ private static Snapshot resolveSnapshotByTagName(
             SnapshotManager snapshotManager,
             ChangelogManager changelogManager,
             long timestampMills,
-            boolean startFromChangelog) {
+            boolean startFromChangelog,
+            boolean returnNullIfTooEarly) {
```
yunfengzhou-hub, line 150: The value of this parameter seems always false.

yunfengzhou-hub:
+1

PR #5398, opened 2025-04-03
[test] RestCatalog: fix unit case annotation typo

XiaoHongbo-Hope:
### Purpose
[test] RestCatalog: fix unit case annotation typo


JingsongLi (approved):
+1

PR #5402, opened 2025-04-03
[flink] Flink batch delete supports aggregation.remove-record-on-delete option

yuzelin:
### Purpose

follow up #5360


leaves12138:
+1

PR #5405, opened 2025-04-06
[core] Make the default version of bitmap index to 2

JingsongLi:
### Purpose

Production environment validation shows that version 2 has significant advantages.


LsomeYeah (approved):
+1

PR #4301, opened 2024-10-10
[flink] Fix query service error while query table with deletion vectors option

Shadowell:
### Purpose
Test query service with deletion vectors options, its okay for table with option `'deletion-vectors.enabled' = 'true'`

Linked issue: close #4265

### Tests

Add test `testQueryServiceWithDeletionVectors`

### API and Format

No
### Documentation

No

Issue #4265, opened 2024-09-26
[Bug] when table set deletion-vectors.enabled =true, call query_service error.


herefree:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Paimon version

1.0

### Compute Engine

flink 1.18

### Minimal reproduce step

1.
```
CREATE TABLE query_service_t (id INT, name STRING,PRIMARY KEY (id) NOT ENFORCED) with(
'bucket' = '1',
 'deletion-vectors.enabled' = 'true'
);
```
2.
`INSERT INTO query_service VALUES (1, 'a_new1'), (3, 'c_new1');`
3.
`call sys.query_service('tmp_rt.query_service_t',1);
`
### What doesn't meet your expectations?

2024-09-26 17:04:18
java.lang.IllegalStateException: SortedRun is not sorted and may contain overlapping key intervals. This is a bug.
	at org.apache.paimon.utils.Preconditions.checkState(Preconditions.java:182)
	at org.apache.paimon.mergetree.SortedRun.validate(SortedRun.java:90)
	at org.apache.paimon.mergetree.SortedRun.fromUnsorted(SortedRun.java:67)
	at org.apache.paimon.mergetree.Levels.updateLevel(Levels.java:190)
	at org.apache.paimon.mergetree.Levels.update(Levels.java:161)
	at org.apache.paimon.table.query.LocalTableQuery.refreshFiles(LocalTableQuery.java:134)
	at org.apache.paimon.flink.service.QueryExecutorOperator.processElement(QueryExecutorOperator.java:113)
	at org.apache.flink.streaming.runtime.tasks.OneInputStreamTask$StreamTaskNetworkOutput.emitRecord(OneInputStreamTask.java:237)
	at org.apache.flink.streaming.runtime.io.AbstractStreamTaskNetworkInput.processElement(AbstractStreamTaskNetworkInput.java:146)
	at org.apache.flink.streaming.runtime.io.AbstractStreamTaskNetworkInput.emitNext(AbstractStreamTaskNetworkInput.java:110)
	at org.apache.flink.streaming.runtime.io.StreamOneInputProcessor.processInput(StreamOneInputProcessor.java:65)
	at org.apache.flink.streaming.runtime.tasks.StreamTask.processInput(StreamTask.java:562)
	at org.apache.flink.streaming.runtime.tasks.mailbox.MailboxProcessor.runMailboxLoop(MailboxProcessor.java:231)
[...]

PR #5406, opened 2025-04-06
[core] Fix FileMonitorTable with Deletion Vectors Enabled

JingsongLi:
### Purpose

close https://github.com/apache/paimon/issues/4265

Fix https://github.com/apache/paimon/pull/4301


LsomeYeah (approved):
+1

PR #5077, opened 2025-02-13
[core] [fix] Partial update should not return empty row with non-null column type exist

leaves12138:
### Purpose

Linked issue: close #xxx

PR #5407, opened 2025-04-07
[core] Aggregate merge function should not return empty row with non-null column type exist

Zouxxyy:
### Purpose

like #5077, aggregate merge function should not return empty row with non-null column type exist


JingsongLi:
+1

PR #5411, opened 2025-04-07
[common] Set SYNC_ALL_PROPERTIES default to true, be kind to user.

leaves12138:
### Purpose

Set SYNC_ALL_PROPERTIES default to true, this change has been approved in paimon dev email.


yuzelin (approved):
+1

PR #1594, opened 2023-07-18
[core] Fix Object reusing bug in PartialUpdateMergeFunction

JingsongLi:
### Purpose

- When delete Ignore is enabled
- and the last data of the key is delete
- which occurs in Sink's WriteBuffer

There will be a bug in object reuse.

We should update key reference even the incoming record is a delete record.

### Tests

Introduce classes in `SortBufferWriteBufferTestBase`:
1. WithPartialUpdateMergeFunctionTest
2. WithAggMergeFunctionTest
3. WithLookupFunctionTest

PR #5414, opened 2025-04-07
[core] Fix aggregate delete bug and refactor SortBufferWriteBufferTestBase

yuzelin:
### Purpose

1. The value of ReusingKeyValue and ReusingTestData also keep primay keys now (closer to real-world use cases)
2. partial update test case add ignore-delelte test again for #1594  (mistakenly removed by other PR).
3. Fix aggregate delete bug and add test: -U retracts and -D deletes; -D re-generates reused row.


leaves12138 (approved):
+1

PR #5416, opened 2025-04-08
[Feature] Add filesystem support for Huawei OBS Storage

qining-mj:
### Purpose

Huawei's OBS storage system is divided into object storage and parallel file system, both of which support the S3 protocol by default. However, in production testing, the object storage fully supports the S3 protocol, while the parallel file system encounters errors when using the S3 protocol. The error is as follows:
Caused by: java.lang.IllegalArgumentException: The path 's3://bucketname/warehouse/xxx' should be a directory. Upon investigation, it was found to be caused by the underlying S3 code.  Therefore, support was provided based on Huawei OBS SDK and hadoop-huaweicloud, and it has been running stably in production for half a year.

### Documentation

https://support.huaweicloud.com/usermanual-obs/obs_13_0007.html

### DIFF
--- a/docs/content/maintenance/filesystems.md
+++ b/docs/content/maintenance/filesystems.md
@@ -514,3 +514,85 @@ spark-sql \
 {{< /tab >}}
 
 {{< /tabs >}}
+
+## OBS
+
+{{< stable >}}
+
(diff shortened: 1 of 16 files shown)


JingsongLi:
`Microsoft Huawei OBS Storage`? `Microsoft` or `Huawei`?

PR #5422, opened 2025-04-08
[core] Introduce 'metadata.stats-mode.per.level'

JingsongLi:
### Purpose

Introduce per level stats mode to avoid stats collection for some level.


leaves12138 (approved):
+1 Looks fine to me

PR #5424, opened 2025-04-09
[test] Refactor SchemaManagerTest to use parameterized tests for move operations

joyCurry30:
### Purpose

Refactor SchemaManagerTest to use parameterized tests for move operations


JingsongLi:
re-open to trigger tests

JingsongLi:
+1

PR #5428, opened 2025-04-09
[core] Add option to always select level 0 files when compacting

tsreaper:
### Purpose

When using async compaction, users may want to monitor the process of compaction by checking the number of level 0 files. This PR adds an option to always select level 0 files when compacting, so user can make sure that, if no level 0 file exists, the compaction has finished.

### Tests

* `KeyValueFileStoreWriteTest`

### API and Format

No format changes.

### Documentation

Document is also updated.


JingsongLi:
+1

PR #5433, opened 2025-04-09
Logging Optimization

haohao0103:
Logging Optimization

add DEBUG-level log checks before logging operations;
log statements use parameterized placeholder


JingsongLi:
`if (LOG.isDebugEnabled())` and `LOG.debug` with arguement is same. We don't need to change them.

PR #5435, opened 2025-04-09
[hive] Fix `sync-all-properties` related tests

liyubin117:
### Purpose

`sync-all-properties` is true instead of false, tests should be also changed.


liyubin117:
@JingsongLi CI passed, PTAL, thanks!

JingsongLi:
+1

PR #5437, opened 2025-04-10
[core] Support writing into postpone bucket tables with table API

tsreaper:
### Purpose

Currently users can only write into postpone bucket tables through Flink. This PR supports writing into postpone bucket tables with table API, so more compute engine can support writing into postpone bucket tables later.

### Tests

* `PrimaryKeySimpleTableTest`

### API and Format

No format changes.

### Documentation

No new feature.


JingsongLi (approved):
+1

PR #5438, opened 2025-04-10
[core][test] Test that snapshot deletion stops when throwing exception for skipping set

yuzelin:
### Purpose

Linked issue: close #xxx

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/ExpireSnapshotsImpl.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/ExpireSnapshotsImpl.java
@@ -230,21 +230,31 @@ public class ExpireSnapshotsImpl implements ExpireSnapshots {
             return 0;
         }
 
-        Set<String> skippingSet = new HashSet<>();
+        Set<String> skippingSet = null;
         try {
-            skippingSet.addAll(snapshotDeletion.manifestSkippingSet(skippingSnapshots));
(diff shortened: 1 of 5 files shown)


paimon-core/src/main/java/org/apache/paimon/table/ExpireSnapshotsImpl.java
```diff
@@ -230,29 +230,38 @@ public int expireUntil(long earliestId, long endExclusiveId) {
...
         }
 
-        Set<String> skippingSet = new HashSet<>();
+        Set<String> skippingSet = null;
         try {
-            skippingSet.addAll(snapshotDeletion.manifestSkippingSet(skippingSnapshots));
+            skippingSet = new HashSet<>(snapshotDeletion.manifestSkippingSet(skippingSnapshots));
         } catch (Exception e) {
-            // maybe snapshot been deleted by other jobs.
-            if (e.getCause() == null || !(e.getCause() instanceof FileNotFoundException)) {
```
JingsongLi, line 238: Add tests for this bug.

yuzelin: ExpireSnapshotsTest#testManifestFileSkippingSetFileNotFoundException

It will fail on master.

JingsongLi:
+1

PR #5440, opened 2025-04-10
[core] Add supportsListObjectsPaged to Catalog

yunfengzhou-hub:
### Purpose

In Catalog.listXXXPaged methods' Javadocs, there is a description as follows.

> NOTE: Currently only RestCatalog will return pagedList data, other catalog will return all
> databases names like listDatabases.

So In order to get the correct paged results, invokers have to identify whether a catalog is a RESTCatalog, which breaks the isolation between the Catalog interface and its implementations. This PR proposes to add a supportsListObjectsPaged method to solve this problem. A similar design could be referenced from Catalog#supportsVersionManagement.

### API and Format

This PR adds a public Catalog#supportsListObjectsPaged method.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
@@ -481,6 +481,11 @@ public abstract class AbstractCatalog implements Catalog {
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


XiaoHongbo-Hope:
Thanks Yunfeng, looks good to me cc @jerry-024

JingsongLi:
+1

PR #5442, opened 2025-04-10
[core] Unaware bucket table compaction generates too many tasks

leaves12138:
### Purpose

Unaware bucket table compaction generates too many tasks.


JingsongLi:
+1

PR #5443, opened 2025-04-10
[flink] Set IOManager to TableRead in Flink source

tsreaper:
### Purpose

Currently `sort-spill-threshold` does not affect Paimon Flink sources, because `IOManager` is not set to `TableRead` in Flink sources. This PR fixes the issue.

### Tests

* `FileStoreSourceReaderTest`

### API and Format

No format changes.

### Documentation

No new feature.


JingsongLi:
+1

PR #5445, opened 2025-04-11
[core] Introduce the BucketFunction interface

### LINKED ISSUE #5444: [Feature] Introduce HashFunction interface
### Search before asking

### Motivation

The first step to make the bucket hash function pluggable

### Solution

_No response_

### Anything else?

_No response_

Aitozi:
### Purpose

Linked issue: close #5444

### DIFF
--- a/docs/layouts/shortcodes/generated/core_configuration.html
+++ b/docs/layouts/shortcodes/generated/core_configuration.html
@@ -57,6 +57,12 @@ under the License.
             <td>Bucket number for file store.<br />It should either be equal to -1 (dynamic bucket mode), -2 (postpone bucket mode), or it must be greater than 0 (fixed bucket mode).</td>
         </tr>
         <tr>
+            <td><h5>bucket-function.type</h5></td>
+            <td style="word-wrap: break-word;">PAIMON</td>
+            <td><p>Enum</p></td>
+            <td>The bucket function for paimon bucket<br /><br />Possible values:<ul><li>"PAIMON"</li></ul></td>
+        </tr>
+        <tr>
             <td><h5>bucket-key</h5></td>
             <td style="word-wrap: break-word;">(none)</td>
             <td>String</td>
--- a/docs/layouts/shortcodes/generated/core_configuration.html
+++ b/docs/layouts/shortcodes/generated/core_configuration.html
@@ -58,9 +58,9 @@ under the License.
         </tr>
         <tr>
             <td><h5>bucket-function.type</h5></td>
-            <td style="word-wrap: break-word;">PAIMON</td>
+            <td style="word-wrap: break-word;">DEFAULT</td>
             <td><p>Enum</p></td>
-            <td>The bucket function for paimon bucket<br /><br />Possible values:<ul><li>"PAIMON"</li></ul></td>
+            <td>The bucket function for paimon bucket<br /><br />Possible values:<ul><li>"DEFAULT"</li></ul></td>
         </tr>
         <tr>
             <td><h5>bucket-key</h5></td>
--- a/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
@@ -129,12 +129,12 @@ public class CoreOptions implements Serializable {
     public static final ConfigOption<BucketFunctionType> BUCKET_FUNCTION_TYPE =
             key("bucket-function.type")
                     .enumType(BucketFunctionType.class)
-                    .defaultValue(BucketFunctionType.PAIMON)
+                    .defaultValue(BucketFunctionType.DEFAULT)
                     .withDescription("The bucket function for paimon bucket");
 
     /** Paimon bucket function type. */
     public enum BucketFunctionType {
-        PAIMON,
+        DEFAULT,
     }
 
     public static final ConfigOption<String> DATA_FILE_EXTERNAL_PATHS =
--- a/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
@@ -125,6 +125,18 @@ public class CoreOptions implements Serializable {
                                                     + "if there is no primary key, the full row will be used.")
                                     .build());
 
+    @Immutable
+    public static final ConfigOption<BucketFunctionType> BUCKET_FUNCTION_TYPE =
+            key("bucket-function.type")
+                    .enumType(BucketFunctionType.class)
+                    .defaultValue(BucketFunctionType.PAIMON)
+                    .withDescription("The bucket function for paimon bucket");
+
+    /** Paimon bucket function type. */
+    public enum BucketFunctionType {
+        PAIMON,
+    }
+
     public static final ConfigOption<String> DATA_FILE_EXTERNAL_PATHS =
             key("data-file.external-paths")
                     .stringType()
@@ -1784,6 +1796,10 @@ public class CoreOptions implements Serializable {
 
     public int bucket() {
         return options.get(BUCKET);
+    }
+
+    public BucketFunctionType bucketFunctionType() {
+        return options.get(BUCKET_FUNCTION_TYPE);
     }
 
     public Path path() {
--- a/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonScan.scala
+++ b/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonScan.scala
@@ -58,7 +58,7 @@ case class PaimonScan(
         // todo introduce bucket transform for different bucket function type
         if (
           bucketSpec.getBucketMode != BucketMode.HASH_FIXED || coreOptions
-            .bucketFunctionType() != BucketFunctionType.PAIMON
+            .bucketFunctionType() != BucketFunctionType.DEFAULT
         ) {
           None
         } else if (bucketSpec.getBucketKeys.size() > 1) {
--- a/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonScan.scala
+++ b/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonScan.scala
@@ -18,6 +18,7 @@
 
 package org.apache.paimon.spark
 
+import org.apache.paimon.CoreOptions.BucketFunctionType
 import org.apache.paimon.predicate.Predicate
 import org.apache.paimon.spark.commands.BucketExpression.quote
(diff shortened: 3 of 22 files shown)


Aitozi:
CC @JingsongLi please take a look when you are free.

BTW, I'm confused the CI compile error do not occur in my local 

image

Aitozi:
> CC @JingsongLi please take a look when you are free.
> 
> BTW, I'm confused the CI compile error do not occur in my local
> 
> image

This seems the bug for the maven 3.9.9. I add the paimon-common dependency to work around this.

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -127,6 +127,18 @@ public class CoreOptions implements Serializable {
                                                     + "if there is no primary key, the full row will be used.")
                                     .build());
 
+    @Immutable
+    public static final ConfigOption<HashType> BUCKET_HASH_TYPE =
```
JingsongLi, line 131: It seems we should introduce a BUCKET_FUNCTION first? Have you researched other systems with similar requirements?

Aitozi: It seems the bucket function's implementation is `hash % bucket_number` for example Hive bucket, So the difference in different system is the hash function.

paimon-core/src/main/java/org/apache/paimon/hash/HashFunction.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+package org.apache.paimon.hash;
+
+import org.apache.paimon.CoreOptions;
+import org.apache.paimon.data.BinaryRow;
+import org.apache.paimon.types.RowType;
+
+/** Bucket key hash function. */
+public interface HashFunction {
```
wwj6591812, line 26: In https://github.com/apache/paimon/issues/5444, the Motivation is `make bucket hash function pluggable`.
But in this PR, I can't find where the logic of load my custom class that implement HashFunction.

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/HashTypeITCase.java
```diff
@@ -0,0 +1,52 @@
...
+import org.junit.jupiter.params.provider.EnumSource;
+
+import java.util.List;
+
+/** Hash type ITCase. */
+public class HashTypeITCase extends CatalogITCaseBase {
+
+    @ParameterizedTest
+    @EnumSource(HashType.class)
+    public void testInsertAndSelect(HashType hashType) throws Exception {
```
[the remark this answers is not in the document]
Aitozi: done

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -127,6 +127,18 @@ public class CoreOptions implements Serializable {
...
+            key("bucket.hash-type")
+                    .enumType(HashType.class)
+                    .defaultValue(HashType.PAIMON)
+                    .withDescription("The hash mode for paimon bucket");
+
+    /** Paimon bucket hash type. */
+    public enum HashType {
```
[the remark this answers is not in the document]
Aitozi: done

paimon-core/src/main/java/org/apache/paimon/hash/HashFunction.java
wwj6591812: OK

wwj6591812:
+1

Pandas886:
PR that is also useful for me

Aitozi:
> PR that is also useful for me

May I ask what's your use case?

Pandas886:
> > PR that is also useful for me
> 
> @Pandas886 May I ask what's your use case?

Translation in English: Currently, Paimon is being integrated into our internal data pipeline tool. When writing to Paimon, if we want to support multiple parallel writes to fixed bucket tables, it requires shuffling by bucket key, with each writer writing data for its own bucket. However, within the data pipeline, the transform phase has already converted the data to an internal format, making it impossible to call the Paimon SDK to retrieve the bucket key.

Aitozi:
resolved conflict. Please take another look again. @JingsongLi 

cc @Zouxxyy @YannByron This PR also add a new parameter (hashType) for the spark bucket function

JingsongLi (changes requested):
I prefer to provide a BUCKET FUNCTION instead HASH FUNCTION.
Now compute a bucket is: `Math.abs(hashcode % numBuckets)`.
I'm not sure if it's universal enough, but BucketFunction is definitely universal enough.

JingsongLi:
cc @Zouxxyy to take a look to spark part.

JingsongLi (commented):
Maybe you should change `PaimonScan` too?

Aitozi:
> PaimonScan

@JingsongLi You mean the `outputPartitioning` ?

Aitozi:
> Maybe you should change `PaimonScan` too?

@JingsongLi Yes, I think we should distinguish in different bucket function. So, we have to introduce new bucket transformer for this. I disable the bucket scan for the other bucket function type now.

Aitozi:
Please take a look again CC @JingsongLi @Zouxxyy

Aitozi:
@luoyuxia can you also help take a look ?

luoyuxia (approved):
@Aitozi Thanks for the pr. After a review,  It does meet our Fluss requirement when integrate with Paimon

paimon-spark/paimon-spark-ut/pom.xml
```diff
@@ -34,9 +34,23 @@ under the License.
     <properties>
         <spark.version>${paimon-spark-common.spark.version}</spark.version>
         <jackson.version>2.15.2</jackson.version>
+        <janino.test.version>3.1.9</janino.test.version>
     </properties>
 
     <dependencies>
+
+        <dependency>
```
Zouxxyy, line 42: why we need these two dependencies

paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -125,6 +125,18 @@ public class CoreOptions implements Serializable {
...
+    @Immutable
+    public static final ConfigOption<BucketFunctionType> BUCKET_FUNCTION_TYPE =
+            key("bucket-function.type")
+                    .enumType(BucketFunctionType.class)
+                    .defaultValue(BucketFunctionType.PAIMON)
+                    .withDescription("The bucket function for paimon bucket");
+
+    /** Paimon bucket function type. */
+    public enum BucketFunctionType {
+        PAIMON,
```
Zouxxyy: This name is a bit strange.

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonScan.scala
```diff
@@ -53,7 +54,11 @@ case class PaimonScan(
     table match {
       case fileStoreTable: FileStoreTable =>
         val bucketSpec = fileStoreTable.bucketSpec()
-        if (bucketSpec.getBucketMode != BucketMode.HASH_FIXED) {
+        // todo introduce bucket transform for different bucket function type
+        if (
+          bucketSpec.getBucketMode != BucketMode.HASH_FIXED || coreOptions
+            .bucketFunctionType() != BucketFunctionType.PAIMON
```
Zouxxyy: This condition also needs to be added to the v2 writer enable

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/catalog/functions/PaimonFunctions.scala
```diff
@@ -60,7 +69,8 @@ object PaimonFunctions {
  *
  * params arg0: bucket number, arg1...argn bucket keys.
  */
-class BucketFunction extends UnboundFunction {
+class BucketFunction(funcName: String, bucketFunctionType: BucketFunctionType)
```
Zouxxyy: No need function name, it can only be "bucket"

paimon-spark/paimon-spark-ut/pom.xml
Aitozi: this seems a bug for maven

paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
Aitozi: I want to introduce other functions like: `hive`. Do you have other suggestions? Or use `DEFAULT`?

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/catalog/functions/PaimonFunctions.scala
Aitozi: does the canonicalName have to be different ?

paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
JingsongLi: DEFAULT looks better to me.

docs/layouts/shortcodes/generated/core_configuration.html
```diff
@@ -56,6 +56,12 @@
             <td>Integer</td>
             <td>Bucket number for file store.<br />It should either be equal to -1 (dynamic bucket mode), -2 (postpone bucket mode), or it must be greater than 0 (fixed bucket mode).</td>
         </tr>
+        <tr>
+            <td><h5>bucket-function.type</h5></td>
+            <td style="word-wrap: break-word;">PAIMON</td>
```
JingsongLi: As discussion, DEFAULT

paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -125,6 +125,18 @@ public class CoreOptions implements Serializable {
...
 
+    @Immutable
+    public static final ConfigOption<BucketFunctionType> BUCKET_FUNCTION_TYPE =
+            key("bucket-function.type")
+                    .enumType(BucketFunctionType.class)
+                    .defaultValue(BucketFunctionType.PAIMON)
+                    .withDescription("The bucket function for paimon bucket");
+
+    /** Paimon bucket function type. */
+    public enum BucketFunctionType {
```
JingsongLi, line 136: Maybe it is better to use String?

paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
Aitozi: renamed

Zouxxyy (approved):
+1

PR #5446, opened 2025-04-11
[core] RESTCatalog: Add Cause to NoPermissionException

shyjsarah:
### Purpose
Add Cause to NoPermissionException

Linked issue: close #xxx

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
@@ -736,6 +736,14 @@ public interface Catalog extends AutoCloseable {
             this(database, null);
         }
 
+        private static String formatMessage(String database, Throwable cause) {
+            String message = String.format(MSG, database);
+            if (cause != null) {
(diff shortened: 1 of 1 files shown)


JingsongLi:
Should there be detailed information about the cause in the exception stack?

shyjsarah:
> Should there be detailed information about the cause in the exception stack?

yes, the detailed information will appear in the stack. This is an optimization for the console situation

JingsongLi:
> > Should there be detailed information about the cause in the exception stack?
> 
> yes, the detailed information will appear in the stack. This is an optimization for the console situation

I don't think it's necessary to optimize at the moment. If we need to modify it, we need to modify all the exceptions, and they should have a consistent strategy.

shyjsarah:
> > > Should there be detailed information about the cause in the exception stack?
> > 
> > 
> > yes, the detailed information will appear in the stack. This is an optimization for the console situation
> 
> I don't think it's necessary to optimize at the moment. If we need to modify it, we need to modify all the exceptions, and they should have a consistent strategy.

ok, l will close this pr.

PR #5448, opened 2025-04-11
[core] Purge files use Table API and no dry run

JingsongLi:
### Purpose

- We should rely on the Table API so that the catalog metadata system can do some cleaning work
- The Table API will no longer support `dry run`, so we cannot add `dry run` capability to all deletion operations


yuzelin (approved):
+1

PR #5453, opened 2025-04-11
[test] Fix incorrect test in StoreCompactOperatorTest

tsreaper:
### Purpose

Currently `StoreCompactOperatorTest` is not a correct test. We should use `SerializationUtils.serializeBinaryRow` to serialize partition, just like the compaction source.


JingsongLi:
+1

PR #5455, opened 2025-04-11
[core] Remove useless 'compaction.max.file-num'

JingsongLi:
### Purpose

The 'compaction.max.file-num' is uesless:
- For append bucketed table, this option is equal to `min.file-num`, it is useless.
- For append unaware table, this option has been removed in https://github.com/apache/incubator-paimon/pull/5442


leaves12138:
+1

PR #5458, opened 2025-04-13
[flink] NoopStoreSinkWriteState should do nothing instead of throwing UnsupportedOperationException

tsreaper:
### Purpose

As the class name suggests, `NoopStoreSinkWriteState` should do nothing when `get` and `put` is called, instead of throwing `UnsupportedOperationException`.

### Tests

This is a simple change without tests.

### API and Format

No format changes.

### Documentation

No new feature.


JingsongLi:
+1

PR #5459, opened 2025-04-13
feat: JindoFileIO supports OSS additional meta information

kaixuanzhang123:
Purpose
Expose jindoFileIo objects' metadata, so more fields of object tables built on OSS can be filled.

Tests
N/A

API and Format
N/A

Documentation
N/A


JingsongLi (approved):
+1

PR #5460, opened 2025-04-14
[doc] Exclude deprecated `branch` option

liyubin117:
### Purpose

As discussed in https://github.com/apache/paimon/pull/5430, update docs.

### DIFF
--- a/docs/layouts/shortcodes/generated/core_configuration.html
+++ b/docs/layouts/shortcodes/generated/core_configuration.html
@@ -45,12 +45,6 @@ under the License.
             <td>Whether to create underlying storage when reading and writing the table.</td>
         </tr>
         <tr>
-            <td><h5>branch</h5></td>
-            <td style="word-wrap: break-word;">"main"</td>
-            <td>String</td>
-            <td>Specify branch name.</td>
-        </tr>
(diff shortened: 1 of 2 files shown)


docs/content/concepts/system-tables.md
```diff
@@ -182,6 +182,13 @@ SELECT * FROM my_table$ro;
 ### Files Table
 You can query the files of the table with specific snapshot.
 
+{{< hint warning >}}
+
+For tables whose current branch has been modified by the deprecated statement
```
JingsongLi, line 187: Actually, there is no documentation for 'branch' key. We don't need this sentence.

docs/layouts/shortcodes/generated/core_configuration.html
```diff
@@ -48,7 +48,7 @@
             <td><h5>branch</h5></td>
```
JingsongLi, line 48: Maybe it is better to just add `ExcludeFromDocumentation` to branch option.

liyubin117: Good idea~

JingsongLi (approved):
+1

PR #5462, opened 2025-04-14
[hotfix] fix the currentFetchEventTimeLag metric will not increase when datastream is completely blocked

### LINKED ISSUE #4244: [Bug] Flink currentFetchEventTimeLag metric calculation is wrong when the datastream is completely blocked
### Search before asking

### Paimon version

master(8ac121a)

### Compute Engine

Flink

### Minimal reproduce step

Check the Flink currentFetchEventTimeLag metric for a completely blocked datastream. It will not grow over time. But the currentEmitEventTimeLag will continue to grow.

The currentFetchEventTimeLag metric should grow over time even if the datastream is completely blocked.

### Anything else?

The lastSplitUpdateTime will not update for a completely blocked datastream.
img_v3_02f1_22fc6ab6-2887-4dbd-94a6-fbb04603193g

In flink, the currentEmitEventTimeLag is calculated by idleTime or currentTime, so this metric can keep grow for a completely blocked datastream.
img_v3_02f1_226f7c12-7838-4cac-8453-2dc9e367812g

xiangyuf:
### Purpose

Linked issue: close #4244

fix the currentFetchEventTimeLag metric will not increase when datastream is completely blocked

### Tests

FileStoreSourceReaderMetricsTest#testFetchLagWhenIdling

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/FileStoreSourceReader.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/FileStoreSourceReader.java
@@ -41,6 +41,7 @@ public class FileStoreSourceReader
                 RecordIterator<RowData>, RowData, FileStoreSourceSplit, FileStoreSourceSplitState> {
 
     private final IOManager ioManager;
+    private final FileStoreSourceReaderMetrics metrics;
 
     private long lastConsumeSnapshotId = Long.MIN_VALUE;
 
@@ -64,6 +65,7 @@ public class FileStoreSourceReader
                 readerContext.getConfiguration(),
                 readerContext);
         this.ioManager = ioManager;
+        this.metrics = metrics;
     }
 
     @Override
@@ -72,6 +74,7 @@ public class FileStoreSourceReader
         if (getNumberOfCurrentlyAssignedSplits() == 0) {
             context.sendSplitRequest();
         }
+        metrics.idlingStarted();
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


xiangyuf:
cc: @JingsongLi

xiangyuf:
cc: @schnappi17 @herefree for review

xiangyuf:
cc: @tsreaper

yunfengzhou-hub:
Hi @xiangyuf, I think we should not change the current behavior of currentFetchEventTimeLag, because "currentFetchEventTimeLag increase with time" has its own meanings and the two situations should not be mixed.

1. If currentFetchEventTimeLag remains unchanged,
  a. If busy/backpressure is low, it means that there is no new data from the upstream Paimon table, and Flink is functioning correctly.
  b. If busy/backpressure is high, it means that the source operator is completedly blocked.
2. If currentFetchEventTimeLag increases with time, it means that the throughput of source operator is lower than that of the incoming data. Furthermore,
  a. If busy/backpressure is low, it means that the bottleneck is at the network layer.
  b. If busy/backpressure is high, it means that the bottleneck is at the Flink operator.

However, if this PR make currentFetchEventTimeLag's value increase in the first situation, we will have less information and may not be able to distinguish between 1.a and 2.a, 1.b and 2.b.

xiangyuf:
@yunfengzhou-hub Thx for reply. I see your point now. The lag increased by the blocked downstream operator or network jitter should be calculated in currentEmitEventTimeLag. The currentFetchEventTimeLag only means the time difference between reading the data file and file creation.

yunfengzhou-hub:
@xiangyuf Generally yes. currentFetchEventTimeLag means the time difference between the record's timestamp and the time when it is read into the Flink job.

xiangyuf:
@yunfengzhou-hub ok, I'll close this issue.

PR #5465, opened 2025-04-14
[core][doc] Document options for 'incremental-to-auto-tag'

JingsongLi:
### Purpose

This option 'incremental-to-auto-tag' requires 'tag.creation-period' and 'tag.period-formatter' configured.


yuzelin (approved):
+1

PR #5466, opened 2025-04-14
[core] Should not close channelManager in BinaryExternalSortBuffer

JingsongLi:
### Purpose

In `BinaryExternalSortBuffer`, should not close `channelManager`, because `channelManager` will be used in `BinaryExternalMerger`.

This bug can be produced in the scenario when spill files exceed `maxNumFileHandles`, this kind of scenario is very rare.

### Tests

`BinaryExternalSortBufferTest.testSpillingAndClearWithMaxFanIn`


yuzelin (approved):
+1

PR #5472, opened 2025-04-15
[core] Remove the checker in bucket function

Zouxxyy:
### Purpose

`static int bucket(int hashcode, int numBuckets)` will be called at the row level, remove the checker in it


JingsongLi (approved):
+1

PR #5473, opened 2025-04-15
[build] Speed up tests execution

liyubin117:
### Purpose

1. End to End Tests Flink 2.x on JDK 11: 13min -> 9min https://github.com/liyubin117/paimon/actions/runs/14464290242
- should not compile for `-Pflink1,spark3`
2. UTCase and ITCase Flink 2.x on JDK 11: 46min -> 40min https://github.com/liyubin117/paimon/actions/runs/14464290191
- should not compile for `-Pflink1,spark3`
- should use `timeout-minutes`, https://github.com/apache/paimon/actions/runs/14462435680/job/40557390576 takes 360min and waste a lot of resources.
- test phase only be responsible to `test verify` instead of `install` again

### DIFF
--- a/.github/workflows/e2e-tests-flink-2.x-jdk11.yml
+++ b/.github/workflows/e2e-tests-flink-2.x-jdk11.yml
@@ -49,8 +49,7 @@ jobs:
 
       - name: Build Flink
         run:  |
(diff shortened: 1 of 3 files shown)


liyubin117:
@JingsongLi This is an effecient optimization, we can speed up the execution cycle to 40min, CI passed, PTAL, thanks:)

JingsongLi:
+1

PR #5475, opened 2025-04-15
[core] introduce v2 deletion vector

LsomeYeah:
### Purpose

Linked issue: close #xxx

Deletion vector is a solution to implement the Merge-On-Read (MOR) structure, which has been also adopted by other formats such as Iceberg and Delta. Paimon's deletion vector PIP: https://cwiki.apache.org/confluence/x/Tws4EQ.

This pr introduced a new style of deletion vector which is compatible with Iceberg.


paimon-common/src/main/java/org/apache/paimon/utils/RoaringBitmap32.java
```diff
@@ -44,6 +44,11 @@ private RoaringBitmap32(RoaringBitmap roaringBitmap) {
         this.roaringBitmap = roaringBitmap;
     }
 
+    // return roaringBitmap.clone() instead of roaringBitmap
+    protected RoaringBitmap get() {
```
JingsongLi, line 48: get => clone?

paimon-common/src/main/java/org/apache/paimon/utils/OptimizedRoaringBitmap64.java
```diff
@@ -0,0 +1,317 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.paimon.utils;
+
+/* This file is based on source code from the Iceberg Project (http://iceberg.apache.org/), licensed by the Apache
```
JingsongLi, line 21: Please add this to `LICENSE` file.

paimon-core/src/main/java/org/apache/paimon/deletionvectors/DeletionVectorsIndexFile.java
```diff
@@ -45,14 +45,26 @@
 public class DeletionVectorsIndexFile extends IndexFile {
 
     public static final String DELETION_VECTORS_INDEX = "DELETION_VECTORS";
+    // Current version id is 1
     public static final byte VERSION_ID_V1 = 1;
+    public static final byte VERSION_ID_V2 = 2;
 
+    private final byte versionID;
```
JingsongLi, line 52: versionID => writeVersionID

paimon-core/src/main/java/org/apache/paimon/deletionvectors/DeletionVectorsIndexFile.java
```diff
@@ -158,16 +170,26 @@ public List<IndexFileMeta> write(Map<String, DeletionVector> input) {
 
     private void checkVersion(InputStream in) throws IOException {
         int version = in.read();
-        if (version != VERSION_ID_V1) {
+        if (version != versionID) {
```
JingsongLi, line 173: Do not check it. 
- If it is version v1, use `readV2DeletionVector`.
- If it is version v2, use `readV2DeletionVector`.

Please add tests for this. Table change v1 to v2, and two versions deletion vectors work together.

LsomeYeah: In this pr, v2 deletion vector has not yet been associated with table read/write operations, nor has the integration of the v1 and v2 dv been implemented. Should these aspects be completed in this PR?

JingsongLi: You don't need to do all things in one PR, but you should do correct thing, it is that we should let table has two versions at the same time.

LsomeYeah: I see. DeletionVectorsIndexFile has been allowed to read v1 and v2 together and a test has been added for this.  Associating v2 dv with table read/write will be committed in the next PRs~

JingsongLi (approved):
+1

PR #5478, opened 2025-04-16
[flink] Support custom lookup shuffle

yunfengzhou-hub:
### Purpose

In Flink Paimon lookup process, the source operators need to read and cache Paimon data locally on Flink subtasks to achive high-performance joining. Given that there have been no prior knowleged about the data from the probe side, each Flink subtask needs to cache all data in the Paimon table, wasting storage resources and actually preventing Flink application in large-scale lookup join situations.

With ([FLIP-462](https://cwiki.apache.org/confluence/display/FLINK/FLIP-462+Support+Custom+Data+Distribution+for+Input+Stream+of+Lookup+Join)), Flink now supports customizing the data distribution pattern for Join operators. Using this mechanism, this PR enables sending data of the same Paimon bucket to the same Flink subtask, and having that subtask only cache data of designated buckets. This design makes better use of Flink's parallel execution ability and solves the problem above.

### Tests

Tests are added to org.apache.paimon.flink.lookup package to verify the correctness of this PR.

### Documentation

This PR introduces a new optimization strategy. A section is added to the document to show when and how to use this optimization.


yunfengzhou-hub:
> Please add description for Flink FLIP and documentation. And also add Paimon documentation.

Thanks for the comment. It seems that Flink has not added a document for custom lookup shuffle yet. I'll submit a PR to Flink for this and, after that PR is reflected on the Flink document website, add a link to Flink's document here.

JingsongLi (approved):
Looks good to me!

Issue #5467, opened 2025-04-14
[Bug] Read metadata info error for pk table


askwang:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Paimon version

1.0

### Compute Engine

spark

### Minimal reproduce step

When we write twice and read metadata info throws exception.
```sql
  test("xxx-pk table") {
    sql("""
          |CREATE TABLE test_tb (
          |    id BIGINT,
          |    name STRING
          |) USING paimon
          |TBLPROPERTIES (
          |    'primary-key' = 'id',
          |    'file.format' = 'parquet',
          |    'file.compression' = 'snappy',
          |    'bucket' = '1'
          |);
          |""".stripMargin)

    spark.sql("insert into test_tb values(1,'a');")
    spark.sql("select __paimon_file_path from test_tb;").show(false)

    spark.sql("insert into test_tb values(3,'c');")
    // throw exception
    spark.sql("select __paimon_file_path from test_tb;").show(false)
  }
```

exceptions:
```text
java.lang.RuntimeException: There need be FileRecordIterator when metadata columns are required.
	at org.apache.paimon.spark.PaimonRecordReaderIterator.validateMetadataColumns(PaimonRecordReaderIterator.scala:53)
	at org.apache.paimon.spark.PaimonRecordReaderIterator.advanceIfNeeded(PaimonRecordReaderIterator.scala:114)
	at org.apache.paimon.spark.PaimonRecordReaderIterator.hasNext(PaimonRecordReaderIterator.scala:62)
	at org.apache.paimon.spark.PaimonPartitionReader.advanceIfNeeded(PaimonPartitionReader.scala:70)
	at org.apache.paimon.spark.PaimonPartitionReader.next(PaimonPartitionReader.scala:51)
	at org.apache.spark.sql.execution.datasources.v2.PartitionIterator.hasNext(DataSourceRDD.scala:119)
```

### What doesn't meet your expectations?

exception

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

askwang:
Only append table or deletion vector table support querying metadata columns

PR #5479, opened 2025-04-16
[log] Add detailed failure log and test for query metadata columns

askwang:
### Purpose
https://github.com/apache/paimon/issues/5467, pk table not support querying metadata columns.


JingsongLi (approved):
+1

PR #5480, opened 2025-04-16
[flink] Support clone external Hive sources to Paimon table

yuzelin:
### Purpose

Usage:
```
clone_hive
--database <source-database>
--table <source-table>
--catalog_conf <source-conf>
--target_database <target-database>
--target_table <target-table>
--target_catalog_conf <target-conf>
--parallelism <parallelism>
--where <partition-filter>                     
```
Support whole database (table not set) and whole catalog (database and table not set).
Please set source-conf to HiveCatalog conf (especially metastore=hive).

If clone multiple tables, parsing partition filter may throw exceptions.

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/CloneAndMigrateAction.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/CloneAndMigrateAction.java
@@ -97,7 +97,7 @@ public class CloneAndMigrateAction extends ActionBase {
         DataStream<List<CloneAndMigrateUtils.MigrateFilesInfo>> files =
                 partitionedSource
                         .process(
-                                CloneAndMigrateUtils.CreateTargetTableAndListFilesFunction(
+                                CloneAndMigrateUtils.createTargetTableAndListFilesFunction(
                                         sourceCatalogConfig, targetCatalogConfig))
                         .name("List Files")
                         .setParallelism(parallelism);
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/CloneAndMigrateAction.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/CloneAndMigrateAction.java
@@ -1,0 +1,123 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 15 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/CloneAndMigrateActionFactory.java
```diff
@@ -0,0 +1,72 @@
...
+package org.apache.paimon.flink.action;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+/** Factory to create {@link CloneAndMigrateAction}. */
+public class CloneAndMigrateActionFactory implements ActionFactory {
+
+    private static final String IDENTIFIER = "clone_migrate";
```
JingsongLi, line 28: clone_hive ? in hive metastore, hive tables, hudi tables and etc?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/CloneAndMigrateAction.java
```diff
@@ -0,0 +1,122 @@
...
+    private final int parallelism;
+
+    public CloneAndMigrateAction(
+            String sourceDatabase,
+            String sourceTableName,
+            Map<String, String> sourceCatalogConfig,
+            String targetDatabase,
+            String targetTableName,
+            Map<String, String> targetCatalogConfig,
+            @Nullable Integer parallelism) {
```
JingsongLi, line 59: We should have the ability to specify partitions, and even the ability to specify formats such as `dt>20250303`.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/clone/CloneHiveUtils.java
```diff
@@ -0,0 +1,415 @@
...
+                        groupedFiles
+                                .computeIfAbsent(files.identifier(), k -> new ArrayList<>())
+                                .add(files);
+                    }
+
+                    for (Map.Entry<Identifier, List<CloneFilesInfo>> entry :
+                            groupedFiles.entrySet()) {
+                        FileStoreTable targetTable =
+                                (FileStoreTable) targetCatalog.getTable(entry.getKey());
+                        commit(entry.getValue(), sourceFileIO, targetTable);
```
JingsongLi, line 226: Commit should be done in a committer node for one table.

yuzelin: 1. In `CloneHiveAction`,  `files` are partitioned by table and forward to `copyAndCommitFunction`;
2. In `copyAndCommitFunction`,  `CloneFilesInfo` are grouped by table. 

So I think I have commit metas for one table once?

Zouxxyy:
The difference between `clone_hive` and the existing `migrate hive` is that clone_hive supports multiple catalogs and uses copy instead of move?

JingsongLi (approved):
+1

JingsongLi:
> The difference between `clone_hive` and the existing `migrate hive` is that clone_hive supports multiple catalogs and uses copy instead of move?

Yes, you are right.

PR #5481, opened 2025-04-16
[core] Improve performance of PartialUpdateMergeFunction with sequence group

Sxnan:
### Purpose

Improve the performance of PartialUpdateMergeFunction with sequence group.  

We made the following changes:

* Use a List, instead of a Map, to store the `FieldsComparator` and `FieldAggregator` to improve the performance for accessing the `FieldsComparator` and `FieldAggregator`.
* Get the field from the value row when it is needed to avoid the unnecessary heavy call of `FieldGetter#getFieldOrNull`.
* Use an array of booleans to record if the sequence group is empty to avoid duplicate checks for the same sequence group.

The optimization gives about 51% performance boost by the benchmark `PartialUpdateMergeFunctionBenchmark` with empty sequence groups, and 28% without empty sequence groups. According to our experience, the `PartialUpdateMergeFunction#updateWithSequenceGroup` method is about 60% workload of the end-to-end compaction with partial update merge engine and sequence group, so we can expect about 30% overall performance of such compaction.

The following benchmark results were run on

```
OpenJDK 64-Bit Server VM 1.8.0_412-b08 on Mac OS X 15.4
Apple M3 Pro
```

Before optimization
```
partial-update-benchmark:                                                                            Best/Avg Time(ms)    Row Rate(K/s)      Per Row(ns)   Relative
OPERATORTEST_partial-update-benchmark_updateWithSequenceGroup                                           12079 / 12883           3311.4            302.0       1.0X
OPERATORTEST_partial-update-benchmark_retractWithSequenceGroup                                          10976 / 11129           3644.3            274.4       1.0X
OPERATORTEST_partial-update-benchmark_updateWithEmptySequenceGroup                                      11636 / 11727           3437.6            290.9       1.0X
```

After optimization
```
partial-update-benchmark:                                                                            Best/Avg Time(ms)    Row Rate(K/s)      Per Row(ns)   Relative
OPERATORTEST_partial-update-benchmark_updateWithSequenceGroup                                            9995 / 10029           4002.0            249.9       1.0X
OPERATORTEST_partial-update-benchmark_retractWithSequenceGroup                                            9311 / 9330           4295.9            232.8       1.0X
OPERATORTEST_partial-update-benchmark_updateWithEmptySequenceGroup                                        7701 / 7747           5194.2            192.5       1.0X
```

### Tests

Already covered by the existing tests, such as `PartialUpdateMergeFunctionTest`

### DIFF
--- a/paimon-benchmark/paimon-micro-benchmarks/src/test/java/org/apache/paimon/benchmark/compact/PartialUpdateMergeFunctionBenchmark.java
+++ b/paimon-benchmark/paimon-micro-benchmarks/src/test/java/org/apache/paimon/benchmark/compact/PartialUpdateMergeFunctionBenchmark.java
@@ -1,0 +1,142 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.paimon.benchmark.compact;
+
+import org.apache.paimon.KeyValue;
+import org.apache.paimon.benchmark.Benchmark;
+import org.apache.paimon.data.GenericRow;
+import org.apache.paimon.mergetree.compact.MergeFunction;
+import org.apache.paimon.mergetree.compact.MergeFunctionFactory;
+import org.apache.paimon.mergetree.compact.PartialUpdateMergeFunction;
+import org.apache.paimon.options.Options;
+import org.apache.paimon.types.DataField;
+import org.apache.paimon.types.DataTypes;
+import org.apache.paimon.types.RowKind;
+import org.apache.paimon.types.RowType;
+
+import org.apache.paimon.shade.guava30.com.google.common.collect.ImmutableList;
+
+import org.junit.jupiter.api.Test;
+
+/** Benchmark for measure the performance for {@link PartialUpdateMergeFunction}. */
+public class PartialUpdateMergeFunctionBenchmark {
+
+    private final int rowCount = 40000000;
+
+    @Test
+    public void testUpdateWithSequenceGroup() {
+        Options options = new Options();
+        options.set("fields.f1.sequence-group", "f2,f3,f4,f5");
+
+        MergeFunctionFactory<KeyValue> factory =
+                PartialUpdateMergeFunction.factory(options, getRowType(6), ImmutableList.of("f0"));
+
+        MergeFunction<KeyValue> func = factory.create();
+
+        Benchmark benchmark =
+                new Benchmark("partial-update-benchmark", rowCount)
+                        .setNumWarmupIters(1)
+                        .setOutputPerIteration(true);
+
+        benchmark.addCase(
+                "updateWithSequenceGroup",
+                5,
+                () -> {
+                    func.reset();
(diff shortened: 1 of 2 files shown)


paimon-core/src/main/java/org/apache/paimon/mergetree/compact/PartialUpdateMergeFunction.java
```diff
@@ -100,22 +102,44 @@ protected PartialUpdateMergeFunction(
...
+        fieldSeqComparators.forEach(
+                (index, comparator) -> {
+                    res.add(new WrapperWithFieldIndex<>(comparator, index));
+                });
+        Collections.sort(res);
+        return res;
+    }
+
+    private List<WrapperWithFieldIndex<FieldAggregator>> getFieldAggregators(
+            Map<Integer, FieldAggregator> fieldAggregators) {
```
yunfengzhou-hub, line 125: getFieldSeqComparators and getFieldAggregators can be merged into one method using generics.

paimon-core/src/main/java/org/apache/paimon/mergetree/compact/PartialUpdateMergeFunction.java
```diff
@@ -238,32 +285,54 @@ private boolean isEmptySequenceGroup(KeyValue kv, FieldsComparator comparator) {
...
+        WrapperWithFieldIndex<FieldsComparator> curComparator =
+                comparatorIter.hasNext() ? comparatorIter.next() : null;
+        Iterator<WrapperWithFieldIndex<FieldAggregator>> aggIter = fieldAggregators.iterator();
+        WrapperWithFieldIndex<FieldAggregator> curAgg = aggIter.hasNext() ? aggIter.next() : null;
+
+        GenericRow currentValueRow = new GenericRow(getters.length);
+        currentValueRow.setRowKind(kv.valueKind());
+        for (int i = 0; i < getters.length; i++) {
+            currentValueRow.setField(i, getters[i].getFieldOrNull(kv.value()));
+        }
```
yunfengzhou-hub, line 299: This adds additional gets and sets to a Row object compared with the original implementation, which might be a potential performance regression.

How about adding a CachingRow implementation that caches the field values of a wrapped InternalRow into an array? This could help avoid invoking all getters at the beginning, while also avoiding invoking field getters multiple times on the same field.

paimon-benchmark/paimon-micro-benchmarks/src/test/java/org/apache/paimon/benchmark/compact/PartialUpdateMergeFunctionBenchmark.java
```diff
@@ -0,0 +1,80 @@
...
+                        add(func, i, RowKind.INSERT, 1, i, 1, 1, 1, 1);
+                    }
+                });
+
+        benchmark.run();
+    }
+
+    private void add(
+            MergeFunction<KeyValue> function, int sequence, RowKind rowKind, Integer... f) {
+        function.add(new KeyValue().replace(GenericRow.of(1), sequence, rowKind, GenericRow.of(f)));
```
yunfengzhou-hub, line 78: It might be better to also add a benchmark for retracts.

Sxnan: Benchmark for retract is added

Sxnan:
@yunfengzhou-hub Thanks for the comment! I updated the PR accordingly. PTAL.

paimon-core/src/main/java/org/apache/paimon/mergetree/compact/PartialUpdateMergeFunction.java
```diff
@@ -226,24 +248,50 @@ private void updateWithSequenceGroup(KeyValue kv) {
         }
     }
 
-    private boolean isEmptySequenceGroup(KeyValue kv, FieldsComparator comparator) {
+    private boolean isEmptySequenceGroup(
+            KeyValue kv, FieldsComparator comparator, boolean[] isEmptySequenceGroup) {
         for (int fieldIndex : comparator.compareFields()) {
+            if (isEmptySequenceGroup[fieldIndex]) {
+                return true;
+            }
```
yunfengzhou-hub, line 256: This if condition only needs to be judged once, before the for loop.

yunfengzhou-hub:
+1

leaves12138 (approved):
+1

PR #5483, opened 2025-04-17
[core] Use write null for uncompact decimal and timestamp in InternalRowSerialize

Zouxxyy:
### Purpose

This PR extracts some modifications from https://github.com/apache/paimon/pull/5242 to the PR, thanks to [zhongyujiang](https://github.com/zhongyujiang)

The `InternalRowSerializer` needs to maintain the handling of null value for uncompact Decimal and timestamp types in `org.apache.paimon.codegen.GenerateUtils.binaryWriterWriteNull` to avoid inconsistencies in hashcode, which could lead to the inconsistencies in bucket calculation.

```java
  def binaryWriterWriteNull(indexTerm: String, writerTerm: String, t: DataType): String =
    t.getTypeRoot match {
      // ordered by type root definition
      case DECIMAL if !Decimal.isCompact(getPrecision(t)) =>
        s"$writerTerm.writeDecimal($indexTerm, null, ${getPrecision(t)})"
      case TIMESTAMP_WITHOUT_TIME_ZONE | TIMESTAMP_WITH_LOCAL_TIME_ZONE
          if !Timestamp.isCompact(getPrecision(t)) =>
        s"$writerTerm.writeTimestamp($indexTerm, null, ${getPrecision(t)})"
      case _ =>
        s"$writerTerm.setNullAt($indexTerm)"
    }
```

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/data/serializer/InternalRowSerializer.java
+++ b/paimon-common/src/main/java/org/apache/paimon/data/serializer/InternalRowSerializer.java
@@ -24,14 +24,20 @@ import org.apache.paimon.data.BinaryRow;
 import org.apache.paimon.data.BinaryRowWriter;
 import org.apache.paimon.data.BinaryWriter;
 import org.apache.paimon.data.BinaryWriter.ValueSetter;
+import org.apache.paimon.data.Decimal;
 import org.apache.paimon.data.GenericRow;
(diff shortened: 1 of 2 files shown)


zhongyujiang (approved):
Makes sense to me.

JingsongLi (approved):
+1

PR #5486, opened 2025-04-17
[core] Resolve hms Permission.NotAllow error

### LINKED ISSUE #5485: [Feature] identify the "Permission.NotAllow" error returned by HMS.
### Search before asking

### Motivation

When calling the getTable interface of HMS, if the server-side validation finds that the user does not have permission and returns an exception, Paimon directly throws a RuntimeException with the message 'Cannot determine if table ' + identifier.getFullName() + ' is a Paimon table.' This makes it difficult to intuitively obtain the actual exception information.

### Solution

When identifying the exception information that contains "Permission.NotAllow", throw TableNoPermissionException

### Anything else?

_No response_

gavin9402:
### Purpose

Linked issue: close #5485

Identify the "Permission.NotAllow" error returned by HMS

### Tests

N/A

### API and Format

N/A

### Documentation

N/A

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
@@ -789,12 +789,12 @@ public interface Catalog extends AutoCloseable {
     /** Exception for trying to operate on the table that doesn't have permission. */
     class TableNoPermissionException extends RuntimeException {
 
-        private static final String MSG = "Table %s has no permission.";
+        private static final String MSG = "Table %s has no permission: %s.";
 
         private final Identifier identifier;
 
         public TableNoPermissionException(Identifier identifier, Throwable cause) {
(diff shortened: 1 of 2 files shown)


paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
```diff
@@ -789,12 +789,12 @@ public Identifier identifier() {
     /** Exception for trying to operate on the table that doesn't have permission. */
     class TableNoPermissionException extends RuntimeException {
 
-        private static final String MSG = "Table %s has no permission.";
+        private static final String MSG = "Table %s has no permission: %s.";
```
JingsongLi, line 792: Can you revert this modification? the root cause already contain the message.

gavin9402: done

JingsongLi:
+1

PR #5491, opened 2025-04-18
[flink] Fix IllegalArgumentException when setting scan.partitions to max_pt() in older Flink versions

tsreaper:
### Purpose

In older Flink versions (for example Flink 1.17), when setting `scan.partitions` to `max_pt()` on a lookup table source, `IllegalArgumentException` will be thrown. This PR fixes the issue.

### Tests

* `LookupTableITCase`.

### API and Format

No format changes.

### Documentation

No new feature.


JingsongLi:
+1

PR #5492, opened 2025-04-19
[core] Introduce 'compaction.delete-ratio-threshold' to reduce compaction

JingsongLi:
…

### Purpose

1. Currently, data file with deletion file always be compacted. We should avoid too many compaction.
2. Rename unaware to append to make it as first-citizen feature.


leaves12138 (approved):
+1

PR #5493, opened 2025-04-21
[core] Reduce the file size per append table compaction task

leaves12138:
### Purpose

TargetFileSize * 50 is too greater for append only table compaction. If 1000 small files, open cost is 4M, then one task will contain 256 * 50 / 4 = 3200 files to compact. It is a great burden for compaction task.


JingsongLi:
+1

PR #5494, opened 2025-04-21
[doc] fix clone hive database doc

plusplusjiajia:
### Purpose
Fix clone hive database doc in clone-from-hive.md


JingsongLi:
+1

PR #5499, opened 2025-04-21
[doc] alter streaming write to paimon table grammar

qingwei727:
### Purpose
alter grammar of streaming write to paimon table in spark sql-write doc

### DIFF
--- a/docs/content/spark/sql-write.md
+++ b/docs/content/spark/sql-write.md
@@ -248,7 +248,7 @@ val stream = df
   .outputMode("append")
   .option("checkpointLocation", "/path/to/checkpoint")
   .format("paimon")
-  .start("/path/to/paimon/sink/table")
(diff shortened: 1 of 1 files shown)


docs/content/spark/sql-write.md
```diff
@@ -248,7 +248,7 @@ val stream = df
   .outputMode("append")
   .option("checkpointLocation", "/path/to/checkpoint")
   .format("paimon")
-  .start("/path/to/paimon/sink/table")
+  .toTable("${database}.${table}")
```
JingsongLi, line 251: modify schema evolution too.

And maybe we can add a line to old path style, just like:
```
.toTable("${database}.${table}")
// .start("/path/to/paimon/sink/table")
```

JingsongLi:
+1

PR #5504, opened 2025-04-22
[core] fix refresh token to earlier refresh and log refresh

jerry-024:
### Purpose

fix refresh token: add log and expire token when in 1 hour.


paimon-core/src/main/java/org/apache/paimon/rest/RESTTokenFileIO.java
```diff
@@ -186,10 +191,11 @@ private void tryToRefreshToken() {
     }
 
     private boolean shouldRefresh() {
-        return token == null || System.currentTimeMillis() > token.expireAtMillis();
+        return token == null || token.expireAtMillis() - System.currentTimeMillis() < 3600_000L;
     }
 
     private void refreshToken() {
+        LOG.info("begin refresh token for identifier [{}]", identifier);
```
JingsongLi, line 198: log expire time after refreshing

JingsongLi (approved):
+1

PR #5505, opened 2025-04-22
chore: bump Scala version to 2.13.16 & 2.12.20

He-Pin:
### Purpose

Bump the Scala version to the latest one


pom.xml
```diff
@@ -86,8 +86,8 @@ under the License.
         <paimon-flinkx-common>paimon-flink1-common</paimon-flinkx-common>
         <paimon-flink-common.flink.version>1.20.1</paimon-flink-common.flink.version>
         <flink.scala.binary.version>2.12</flink.scala.binary.version>
-        <scala212.version>2.12.15</scala212.version>
```
YannByron, line 89: I prefer to use `2.12.18` which is same with the latest scala version that Spark 3.5 use.

PR #5506, opened 2025-04-22
[hive][clone] check compatibility if paimon table already exists

LsomeYeah:
### Purpose

Linked issue: close #xxx

When clone hive table to paimon, if target paimon table already exists, check its compatibility with hive table.


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/clone/hive/ListHiveFilesFunction.java
```diff
@@ -95,6 +118,71 @@ public void processElement(
...
+                        existedTable
+                                .schema()
+                                .projectedLogicalRowType(existedTable.partitionKeys())
+                                .getFields());
+
+        if (sourcePartitionFields.size() != existedPartitionFields.size()) {
+            throw new IllegalStateException(
+                    "size of source table partition keys not equal existed paimon table partition keys.");
+        }
+        checkCompatible(sourcePartitionFields, existedPartitionFields);
```
JingsongLi, line 156: For partitions, we can just check `List<String>` names.

JingsongLi (approved):
+1

PR #5507, opened 2025-04-22
Revert "[doc] alter streaming write to paimon table grammar"

Zouxxyy:
Reverts apache/paimon#5499 because paimon spark currently does not support this syntax


JingsongLi:
+1

PR #5508, opened 2025-04-22
[doc] Add missing `options` for related procedures

liyubin117:
### Purpose

`options` is a very useful feature, introduced by https://github.com/apache/paimon/pull/4657, we should expose it for users.


JingsongLi:
+1

PR #5513, opened 2025-04-22
[flink] Fix file name and overwrite semantic for Clone Hive

JingsongLi:
### Purpose

1. File name should be generated by UUID.
2. Clone is OVERWRITE semantic that will overwrite the partitions of the target table according to the data.


wwj6591812:
+1

PR #5514, opened 2025-04-22
[core] Disable PARTITION_IDLE_TIME_TO_REPORT_STATISTIC by default

JingsongLi:
### Purpose

Since the introduction of stats reporting to batch jobs, the stability of many online tasks has been affected, and the batch task reporting time is unavailable, seriously affecting stability. We should default to disabling reporting.


Zouxxyy (approved):
+1

PR #5515, opened 2025-04-22
Feature compatible sr bitmap

AOKI123:
[core] Add the `bitmap_agg` aggregation function to be compatible with StarRocks' bitmap format.
### Purpose

Linked issue: close #xxx

### DIFF
--- a/paimon-ext/lib/org.apache.paimon.factories.Factory
+++ b/paimon-ext/lib/org.apache.paimon.factories.Factory
@@ -1,0 +1,42 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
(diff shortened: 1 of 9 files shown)


JingsongLi:
Why we need to have a starrocks bitmap?

AOKI123:
> Why we need to have a starrocks bitmap?
 @JingsongLi 
When I use the bitmap data type of the Paimon table and query Paimon data using StarRocks, and use the StarRocks function `bitmap_from_binary`, an error occurs stating that it cannot be serialized into the Bitmap type of StarRocks. However, it is quite strange that both use RoaringBitmap as the implementation. Why does a serialization conflict occur?

AOKI123:
> Why we need to have a starrocks bitmap?

@JingsongLi 
Hi. When I use the bitmap data type of the Paimon table and query Paimon data with StarRocks, and use the StarRocks function bitmap_from_binary, an error occurs stating that it cannot be serialized into the Bitmap type of StarRocks. So I extended it to be compatible with the bitmap type of StarRocks, enabling me to use various bitmap functions on the Paimon table in StarRocks.

JingsongLi:
@AOKI123 I suggest you use the custom jar to customize the merge function, and do not introduce SR dependencies in Paimon.

PR #5518, opened 2025-04-23
[flink] Deprecate Copy Files procedure

JingsongLi:
### Purpose

1. Rename clone to copy_files.
2. Deprecate copy_files, the normal process should commit a snapshot to the catalog, but this action does not do so. Currently, this action can only be applied to the FileSystemCatalog.
3. In future, we should develop `clone_hive` to cover paimon to paimon table clone.


LsomeYeah (approved):
+1

PR #5519, opened 2025-04-23
[flink] Rename clone_hive to clone and add procedure

JingsongLi:
### Purpose

1. Rename clone_hive to clone
2. add clone procedure

And document clone:
```
Currently, clone supports:

1. Clone Hive tables in Hive Catalog to Paimon Catalog, supports Parquet, ORC, Avro formats, target table will
   be append table.

The source table below is currently under development:
1. Clone Hudi tables in Hive Catalog to Paimon Catalog, target table will be append table.
2. Clone Paimon tables to Paimon tables, target table can be primary table or append table.
```

### DIFF
--- a/docs/content/migration/clone-to-paimon.md
+++ b/docs/content/migration/clone-to-paimon.md
@@ -1,9 +1,9 @@
 ---
-title: "Clone From Hive"
+title: "Clone To Paimon"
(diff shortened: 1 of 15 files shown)


LsomeYeah (approved):
+1

PR #5521, opened 2025-04-23
[flink] Clean empty directory in FlinkOrphanFilesClean

JingsongLi:
### Purpose

We should also clean empty directories in OrphanFilesClean, it already list directories.


LsomeYeah (approved):
+1

PR #5522, opened 2025-04-23
[WIP] [Arrow] Adjust ArrowFieldWriterFactory to support nullable with correct parameters

lan3394:
### Purpose

ArrowFieldWriter does not check whether it could write null yet, if the record is null and the field is not nullable, it will meet problem. 
For example, if the target field is `VarChar(length=Integer.MAX_VALUE, nullable=false)`, then
we should throw an exception when we call writer.write(null). Because it could not be null.

### DIFF
--- a/paimon-arrow/src/main/java/org/apache/paimon/arrow/writer/ArrowFieldWriter.java
+++ b/paimon-arrow/src/main/java/org/apache/paimon/arrow/writer/ArrowFieldWriter.java
@@ -25,14 +25,18 @@ import org.apache.arrow.vector.FieldVector;
 
 import javax.annotation.Nullable;
 
+import static java.lang.String.format;
+
 /** A reusable writer to convert a field into Arrow {@link FieldVector}. */
 public abstract class ArrowFieldWriter {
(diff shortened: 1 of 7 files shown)


paimon-arrow/src/main/java/org/apache/paimon/arrow/writer/ArrowFieldWriter.java
```diff
@@ -67,6 +69,9 @@ protected abstract void doWrite(
     /** Get the value from the row at the given position and write to specified row index. */
     public void write(int rowIndex, DataGetters getters, int pos) {
         if (getters.isNullAt(pos)) {
+            if (!isNullable) {
+                throw new RuntimeException();
```
leaves12138, line 73: Please specific exception message

leaves12138: format("Parquet does not support null values in non-nullable fields. Field name : %s expected not null but found null", fieldName)

PR #5523, opened 2025-04-24
[checkstyle] Upgrade checkstyle

yuzelin:
### Purpose

We follow up flink's IDEA settings:
https://nightlies.apache.org/flink/flink-docs-release-2.0/docs/flinkdev/ide_setup/

The checkstyle plugin has been upgraded, and the xml file should be changed by:
https://github.com/apache/flink/pull/24540/files#diff-45e2166a534b07a500480f01991ad43b0a77454c03158633b4d7a8d0d32532c0

I only changed the file minimal to make the plugin work.


yuzelin:
We should also upgrade maven-checkstyle-plugin in POM. Closed temporarily.

PR #5525, opened 2025-04-24
[core] RESTCatalog: fix npe when merge conf

jerry-024:
### Purpose

[core] RESTCatalog: fix npe when merge conf
### Tests
- RESTUtilTest.testMerge

### API and Format
NO


JingsongLi:
+1

PR #5535, opened 2025-04-25
[flink] Fix lookup failed to join keys when using max_pt() and lookup table is updated by insert overwrite

tsreaper:
### Purpose

If Flink lookup source table is updated by `INSERT OVERWRITE` and `max_pt()` is used to specify partition, lookup join operator might fail to join keys. This PR fixes the issue.

### Tests

* `LookupJoinITCase`

### API and Format

No format changes.

### Documentation

No new feature.


JingsongLi:
+1

PR #5537, opened 2025-04-25
[flink] Fix Flink Lookup Join for Postpone bucket table

JingsongLi:
### Purpose

Num bucket should be extracted from data splits.


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/PrimaryKeyPartialLookupTable.java
```diff
@@ -241,12 +264,14 @@ public void refresh() {
...
-                    List<DataFileMeta> before = ((DataSplit) split).beforeFiles();
-                    List<DataFileMeta> after = ((DataSplit) split).dataFiles();
+                    DataSplit dataSplit = (DataSplit) split;
+                    BinaryRow partition = dataSplit.partition();
+                    int bucket = dataSplit.bucket();
+                    List<DataFileMeta> before = dataSplit.beforeFiles();
+                    List<DataFileMeta> after = dataSplit.dataFiles();
 
                     tableQuery.refreshFiles(partition, bucket, before, after);
+                    totalBuckets.put(partition, dataSplit.totalBuckets());
```
tsreaper, line 274: `totalBuckets` might be empty because the table is created by older versions of Paimon. In this case, use `bucket` from `table` instead. Make sure that `bucket > 0`. Add a test case for this.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/FixedBucketFromPkExtractor.java
```diff
@@ -62,49 +49,16 @@ public FixedBucketFromPkExtractor(TableSchema schema) {
...
-    public BinaryRow partition() {
-        return partitionProjection.apply(primaryKey);
-    }
-
-    private BinaryRow bucketKey() {
+    public int bucket(int numBuckets, InternalRow primaryKey) {
+        BinaryRow bucketKey;
         if (sameBucketKeyAndTrimmedPrimaryKey) {
-            return trimmedPrimaryKey();
+            bucketKey = trimmedPrimaryKeyProjection.apply(primaryKey);
```
tsreaper, line 57: No need for this check. Unlike `KeyAndBucketExtractor`, trimmed primary key is not needed in this class.

Actually, why do you need two classes to implement this? Just put all these code into `PrimaryKeyPartialLookupTable`.

tsreaper (approved):

yunfengzhou-hub:
+1

PR #5538, opened 2025-04-25
[arrow] Fix bug: field writer contruct in ArrowFormatWriter should use nested field nullable to create sub writers

leaves12138:
### Purpose

Linked issue: close #xxx


yuzelin (approved):
+1

PR #5539, opened 2025-04-26
change method name style

kaixuanzhang123:
### Purpose
style: change dataFiles method to dataFileMetas 。

DataFiles makes it easy for us to confuse data files with metadata files。method and field DataFiles not a good name

### Tests


JingsongLi (changes requested):
-1 Please do not change the method name casually, as it will seriously affect compatibility

PR #5542, opened 2025-04-27
[spark] Close ioManager in PaimonPartitionReader

Zouxxyy:
### Purpose

Linked issue: close #xxx


paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonPartitionReader.scala
```diff
@@ -117,6 +127,7 @@ case class PaimonPartitionReader(
       if (currentRecordReader != null) {
         currentRecordReader.close()
       }
+      ioManager.close()
```
wwj6591812, line 130: Why not add this in finally?

JingsongLi:
+1

PR #5544, opened 2025-04-27
[core] ro table support fallback branch

jerry-024:
### Purpose

ReadOptimizedTable support fallback branch

### Tests
CatalogTableITCase.testReadOptimizedTableFallBack

### API and Format
NO

### Documentation
NO


JingsongLi:
We should get wrapped table and fallback table from FallbackReadFileStoreTable, and create mainScan and fallbackScan to create a FallbackReadFileStoreTable.Scan.

`FallbackReadFileStoreTable.Scan` can be a static class.

JingsongLi:
+1

PR #5546, opened 2025-04-28
[core] Integrate v2 dv with read/write

LsomeYeah:
### Purpose

Linked issue: close #xxx

In pr https://github.com/apache/paimon/pull/5475, we had introduced a new style deletion vector(v2 dv) which is compatible with dv in iceberg. This pr aims to integrate v2 dv with read/write operations by introducing 'deletion-vector.version' argument. when 'deletion-vector.version' = '2', paimon will use v2 dv as deletion vectors.

note:  

- ~~Default value of 'deletion-vector.version' should be 1,  here will set the default value to 2 temporarily, and run all tests to verify if there are any potential issues~~. All CI tests had passed, default value has been set to 1.
- As discussed in pr https://github.com/apache/paimon/pull/5475, table should be allowed to have two versions at the same time, the compatibility of v1 and v2 will be accomplished in next pr.

### Tests
All tests for validating v1 have been extended to include validations for v2.


paimon-core/src/main/java/org/apache/paimon/deletionvectors/DeletionVector.java
```diff
@@ -85,6 +84,9 @@ default boolean checkedDelete(long position) {
     /** @return the number of distinct integers added to the DeletionVector. */
     long getCardinality();
 
+    /** @return the version of the deletion vector. */
+    int dvVersion();
```
JingsongLi, line 88: version

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -1510,6 +1510,13 @@ public class CoreOptions implements Serializable {
                     .defaultValue(MemorySize.ofMebiBytes(2))
                     .withDescription("The target size of deletion vector index file.");
 
+    public static final ConfigOption<Integer> DELETION_VECTOR_VERSION =
+            key("deletion-vector.version")
+                    .intType()
+                    .defaultValue(2)
```
JingsongLi, line 1516: default is 1

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -1510,6 +1510,13 @@ public class CoreOptions implements Serializable {
                     .defaultValue(MemorySize.ofMebiBytes(2))
                     .withDescription("The target size of deletion vector index file.");
 
+    public static final ConfigOption<Integer> DELETION_VECTOR_VERSION =
+            key("deletion-vector.version")
```
JingsongLi, line 1514: deletion-vectors.version

paimon-core/src/main/java/org/apache/paimon/deletionvectors/BitmapDeletionVector.java
```diff
@@ -31,6 +31,7 @@
  * not exceeding {@link RoaringBitmap32#MAX_VALUE}.
  */
 public class BitmapDeletionVector implements DeletionVector {
+    public static final int VERSION = 1;
```
JingsongLi, line 34: keep an empty line up

paimon-core/src/main/java/org/apache/paimon/deletionvectors/DeletionVector.java
```diff
@@ -98,40 +100,55 @@ default boolean checkedDelete(long position) {
...
+        } else {
+            throw new RuntimeException("Invalid deletion vector version: " + version);
         }
     }
 
     static DeletionVector read(FileIO fileIO, DeletionFile deletionFile) throws IOException {
         Path path = new Path(deletionFile.path());
         try (SeekableInputStream input = fileIO.newInputStream(path)) {
+            // read dv version
+            int version = input.read();
```
JingsongLi, line 117: Can we avoid this reading? Can we just use magic number to decide which version?

paimon-core/src/main/java/org/apache/paimon/deletionvectors/DeletionVector.java
```diff
@@ -98,40 +100,55 @@ default boolean checkedDelete(long position) {
...
+
+                byte[] bytes = new byte[actualSize];
+                dis.readFully(bytes);
+                return BitmapDeletionVector.deserializeFromBytes(bytes);
+            } else if (version == Bitmap64DeletionVector.VERSION) {
+                // read v2 deletion vector
+                byte[] bytes = new byte[(int) deletionFile.length()];
+                dis.readFully(bytes);
+
+                return Bitmap64DeletionVector.deserializeFromBytes(bytes);
```
JingsongLi, line 142: Why v2 doesn't need to skip 4 bytes?

paimon-core/src/main/java/org/apache/paimon/deletionvectors/DeletionVectorIndexFileWriter.java
```diff
@@ -127,9 +125,25 @@ private long writtenSizeInBytes() {
...
+                        String.format(
+                                "write version id is %s, but deletionVector is not an instance of %s, actual class is %s.",
+                                writeVersionId,
+                                BitmapDeletionVector.class.getName(),
+                                deletionVector.getClass().getName()));
+
                 writeV1(key, deletionVector);
-            } else if (versionId == VERSION_ID_V2) {
+            } else if (writeVersionId == Bitmap64DeletionVector.VERSION) {
+                Preconditions.checkArgument(
```
JingsongLi, line 139: Please use checkArgument(bool, string, args...)

paimon-core/src/main/java/org/apache/paimon/deletionvectors/DeletionVectorIndexFileWriter.java
```diff
@@ -127,9 +125,25 @@ private long writtenSizeInBytes() {
 
         private void write(String key, DeletionVector deletionVector) throws IOException {
             Preconditions.checkNotNull(dataOutputStream);
-            if (versionId == VERSION_ID_V1) {
+            if (writeVersionId == BitmapDeletionVector.VERSION) {
+                Preconditions.checkArgument(
+                        deletionVector instanceof BitmapDeletionVector,
+                        String.format(
```
JingsongLi, line 131: Please use checkArgument(bool, string, args...)

paimon-common/src/main/java/org/apache/paimon/CoreOptions.java
LsomeYeah: The default value is set to 2 temporarily, and run all CI tests to verify if there are any potential issues. It will be set to 1 after all tests passed~

paimon-core/src/main/java/org/apache/paimon/deletionvectors/DeletionVector.java
LsomeYeah: Serialized bytes of v2 dv include the 4 bytes for length, so it don't require to read the 4 bytes separately. Instead pass the entire byte array for deserialization.

JingsongLi:
+1

PR #5547, opened 2025-04-28
[orc] Default value of 'orc.timestamp-ltz.legacy.type' should be true

JingsongLi:
### Purpose

See https://lists.apache.org/thread/gr2k5lnvobhjxvvmgjwto9ojv027ym49


leaves12138:
+1

PR #5549, opened 2025-04-28
[hotfix] avoid to use `false` as default value of legacyTimestampLtzType in orc format

xiangyuf:
### Purpose

Linked issue: close #xxx

avoid to use `false` as default value of legacyTimestampLtzType in orc format


JingsongLi:
Duplicate in https://github.com/apache/paimon/pull/5547

PR #5551, opened 2025-04-28
[hive][clone] Output the table name if the format is unsupported.

sundapeng:
### Purpose

CloneHiveTable would fail when hive table format is unsupported, better to print the table name if the format is unsupported.

### Tests

N/A

### API and Format

N/A

### Documentation

N/A


JingsongLi (approved):
+1

PR #5552, opened 2025-04-28
[core] Rename 'deletion-vectors.version' to 'deletion-vectors.bitmap64'

JingsongLi:
### Purpose

Linked issue: close #xxx

PR #5553, opened 2025-04-29
[core] add compatible test for reading old dv index file

LsomeYeah:
### Purpose

Linked issue: close #xxx

Since some refactoring was performed in https://github.com/apache/paimon/pull/5552, this pr tests if the new version can handle the dv index file written in old version, and the answer is true.


JingsongLi:
+1

PR #5556, opened 2025-04-29
[spark] Avoid using bucket function for timestamp with spark unsupported precision

Zouxxyy:
### Purpose

The reason of this is that Spark's timestamp precision is fixed to 6, and in [[BucketFunction.bind]], we use `InternalRowSerializer(bucketKeyRowType)` to convert paimon rows, but the `bucketKeyRowType` is derived from Spark's StructType which will lose the true precision of timestamp, leading to anomalies in bucket calculations.


JingsongLi (approved):
+1

PR #5562, opened 2025-04-30
[rest] Simplify auth provider to synchronously refresh token

JingsongLi:
### Purpose

1. Simpler.
2. Fix bug: read file may fail and never refresh again.


jerry-024 (approved):

jerry-024:
+1

PR #5565, opened 2025-05-06
[rest] ignore empty securityToken and expiration in DLFToken.

plusplusjiajia:
### Purpose
ignore empty securityToken and expiration in DLFToken.

### Tests
AuthProviderTest#testCreateDLFAuthProviderByEmptyParams


JingsongLi:
Can you explain when the securityToken and expiration will be empty?

plusplusjiajia:
> Can you explain when the securityToken and expiration will be empty?
For example, when getting a DLFToken from a local file, there is a SecurityToken or Expiration property key, but the value is empty, or user configures the catalog options keys from a template, which include the dlf.security-token key, but actually the value is empty, so the option will be like dlf.security-token=''

JingsongLi:
> > Can you explain when the securityToken and expiration will be empty?
> 
> For example, when getting a DLFToken from a local file, there is a SecurityToken or Expiration property key, but the value is empty, or user configures the catalog options keys from a template, which include the dlf.security-token key, but actually the value is empty, so the option will be like dlf.security-token=''

Thanks for the explanation, I cannot see the specific cases, let's improve this in future.

PR #5566, opened 2025-05-06
[core] Fix computed column and projection of BinlogTable

yuzelin:
### Purpose

This PR fixes two problems. 
1. Projection was not be handled.
2. Computed column is not supported.
For example, if you create a able like 
```
CREATE TABLE test_table (a int, b int, c AS a + b)
```
You will get an exception while querying binlog:
`Cannot apply '+' to arguments of type '<INTEGER ARRAY> + <INTEGER ARRAY>'.`. It is complicated to support it because we should handle flink computed column expressions in this occasion.


JingsongLi:
Please enrich description.

JingsongLi:
+1

PR #5573, opened 2025-05-07
[core] Introduce Catalog.authTableQuery to auth query select and filter

JingsongLi:
### Purpose

This PR introduces an interface to give Catalog the opportunity to perform column level and row level permission validation on queries.

Introduce Catalog method:
```
    /**
     * Auth table query select and filter.
     *
     * @param identifier path of the table to alter partitions
     * @param select projection fields
     * @param filter query filter
     * @throws TableNotExistException if the table does not exist
     */
    void authTableQuery(Identifier identifier, List<String> select, List<String> filter)
            throws TableNotExistException;
```

Introduce rest api:
```
  /v1/{prefix}/databases/{database}/tables/{table}/auth:
    post:
      tags:
        - table
      summary: Auth table query
      operationId: authTableQuery
      parameters:
        - name: prefix
          in: path
          required: true
          schema:
            type: string
        - name: database
          in: path
          required: true
          schema:
            type: string
        - name: table
          in: path
          required: true
          schema:
            type: string
      requestBody:
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/AuthTableQueryRequest'
      responses:
        "200":
          description: Success, no content
        "401":
          $ref: '#/components/responses/UnauthorizedErrorResponse'
        404:
          description:
            Not Found
            - TableNotExistException, table does not exist
          content:
            application/json:
              schema:
                $ref: '#/components/responses/ResourceNotExistErrorResponse'
              examples:
                TableNotExist:
                  $ref: '#/components/examples/TableNotExistError'
        "500":
          $ref: '#/components/responses/ServerErrorResponse'
```

NOTE: This PR does not cover all cases, and many scans may need to adjust the code to set the required projection, otherwise it will result in verification permission failure.

### DIFF
--- a/docs/static/rest-catalog-open-api.yaml
+++ b/docs/static/rest-catalog-open-api.yaml
@@ -590,6 +590,51 @@ paths:
           $ref: '#/components/responses/TableNotExistErrorResponse'
         "500":
           $ref: '#/components/responses/ServerErrorResponse'
+  /v1/{prefix}/databases/{database}/tables/{table}/auth:
+    post:
+      tags:
+        - table
+      summary: Auth table query
+      operationId: authTableQuery
+      parameters:
+        - name: prefix
+          in: path
+          required: true
+          schema:
+            type: string
+        - name: database
+          in: path
+          required: true
+          schema:
+            type: string
+        - name: table
+          in: path
+          required: true
+          schema:
+            type: string
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/AuthTableQueryRequest'
+      responses:
+        "200":
+          description: Success, no content
+        "401":
+          $ref: '#/components/responses/UnauthorizedErrorResponse'
+        404:
+          description:
+            Not Found
+            - TableNotExistException, table does not exist
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/responses/ResourceNotExistErrorResponse'
+              examples:
+                TableNotExist:
+                  $ref: '#/components/examples/TableNotExistError'
+        "500":
+          $ref: '#/components/responses/ServerErrorResponse'
   /v1/{prefix}/databases/{database}/tables/{table}/snapshot:
     get:
       tags:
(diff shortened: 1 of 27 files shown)


paimon-core/src/test/java/org/apache/paimon/rest/RESTCatalogServer.java
```diff
@@ -606,6 +616,29 @@ private Optional<MockResponse> checkTablePartitioned(Identifier identifier) {
...
+        List<String> columnAuth = columnAuthHandler.get(identifier.getFullName());
+        if (columnAuth != null) {
+            requestBody
+                    .select()
+                    .forEach(
+                            column -> {
+                                if (!columnAuth.contains(column)) {
+                                    throw new Catalog.TableNoPermissionException(identifier);
+                                }
+                            });
```
plusplusjiajia, line 637: I'm not sure requestBody.filter() how to check.

JingsongLi: Deserialize filters to `List<Predicate>` and do check.

For example, f1>0, If we configure row level permissions on the server, such as only supporting user A to obtain data with f1>0, then the filter here must contain f1>0 or f1>5 in order to pass the verification.

plusplusjiajia: Looks reasonable.

docs/static/rest-catalog-open-api.yaml
```diff
@@ -590,6 +590,51 @@ paths:
...
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/AuthTableQueryRequest'
+      responses:
+        "200":
+          description: Success, no content
+        "401":
+          $ref: '#/components/responses/UnauthorizedErrorResponse'
```
plusplusjiajia, line 624: I think the 403 ForbiddenErrorResponse should be added.

plusplusjiajia:
LGTM，+1

PR #5574, opened 2025-05-07
[core] Introduce 'cache.expire-after-write' to expire a max interval

JingsongLi:
### Purpose

Our previous cache expiration never expired under frequent access, which can cause some problems, such as schema updates that prevent users from seeing the new schema of the table.

So in this PR:

1. Introduce 'cache.expire-after-write'.
2. Rename 'cache.expiration-interval' to 'cache.expire-after-access'.


wwj6591812:
+1

PR #5575, opened 2025-05-07
[hive][clone] support clone file format including options

shyjsarah:
### Purpose
support clone file format including options

Currently, the file format and related configurations of the source hive table are not cloned. The cloned target paimon table uses the default option 'file.format' =  'parquet', which will affect the subsequent writing of the paimon table. For example, after a hive table using orc is cloned to the paimon table and written again, the new file will be in parquet format, which is unexpected. This PR is to fix this problem.

Linked issue: close #xxx


JingsongLi:
Please enrich description.

shyjsarah:
> Please enrich description.

done

JingsongLi:
+1

PR #5576, opened 2025-05-08
[flink] Fix deletion vector support for postpone bucket tables

tsreaper:
### Purpose

Currently, deletion vectors are not working in postpone bucket tables, because compaction job does not handle index files correctly. This PR fixes the problem.

### Tests

* `PostponeBucketTableITCase#testDeletionVector`

### API and Format

No format changes.

### Documentation

No new feature.

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/CompactAction.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/CompactAction.java
@@ -277,41 +277,31 @@ public class CompactAction extends TableActionBase {
 
             bucketOptions = new HashMap<>(table.options());
             bucketOptions.put(CoreOptions.BUCKET.key(), String.valueOf(bucketNum));
-            // bucket-postpone files won't be read if deletion vectors are set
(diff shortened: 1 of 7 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/CompactAction.java
```diff
@@ -277,35 +277,41 @@ private boolean buildForPostponeBucketCompaction(
 
             bucketOptions = new HashMap<>(table.options());
             bucketOptions.put(CoreOptions.BUCKET.key(), String.valueOf(bucketNum));
-            FileStoreTable realTable = table.copy(table.schema().copy(bucketOptions));
+            // bucket-postpone files won't be read if deletion vectors are set
```
JingsongLi, line 280: Maybe here should use `SnapshotReader`, needLookup will skip all level -2 and 0 files. There are two many magic in `TableScan`, we should use underly reader.

Please also add cases for other lookup cases.

tsreaper: We use `Table` here, because currently all source operators rely on `ReadBuilder`, which is created from a `Table`.

PR #5579, opened 2025-05-08
[parquet] Remove old parquet reader

JingsongLi:
### Purpose

After two versions, the new Parquet reader is now very stable, and we can delete the old reader.


leaves12138 (approved):
+1

PR #5580, opened 2025-05-08
[core] Support: scan by creation timestamp

jerry-024:
### Purpose

Support scan by creation timestamp:
- If the timestamp in (earliest, latest] snapshot, will choose the equal or later snapshot
- Or filter by time in manifest

### Tests

- CreationTimestampStartingScannerTest


paimon-core/src/main/java/org/apache/paimon/table/source/snapshot/CreationTimestampStartingScanner.java
```diff
@@ -0,0 +1,92 @@
...
+        super(snapshotManager);
+        Long startingSnapshotPrevId =
+                TimeTravelUtil.earlierThanTimeMills(
+                        snapshotManager,
+                        changelogManager,
+                        creationMillis,
+                        changelogDecoupled,
+                        true);
+        Optional.ofNullable(startingSnapshotPrevId)
+                .map(id -> id + 1)
```
yuzelin, line 53: Why not just use `TimeTravelUtil.laterOrEaqual` so that we don't need to map `id + 1`.

paimon-core/src/main/java/org/apache/paimon/table/source/AbstractDataTableScan.java
```diff
@@ -244,6 +245,15 @@ protected StartingScanner createStartingScanner(boolean isStreaming) {
             case FROM_FILE_CREATION_TIME:
                 Long fileCreationTimeMills = options.scanFileCreationTimeMills();
                 return new FileCreationTimeStartingScanner(snapshotManager, fileCreationTimeMills);
+            case FROM_CREATION_TIMESTAMP:
+                Long creationTimeMills = options.scanCreationTimeMills();
+                return new CreationTimestampStartingScanner(
```
yuzelin, line 250: I think ContinuousFromSnapshotStartingScanner is not needed. Just use a method to create StartingScanner.
And just test the method that created StartingScanner is right.

paimon-core/src/main/java/org/apache/paimon/table/source/snapshot/CreationTimestampStartingScanner.java
jerry-024: Here isn't later or equal, if the ts is earlier than the earliest snapshot or changelog, we need to get null.

yuzelin (approved):
+1

PR #5581, opened 2025-05-08
[Bug] Fix bug of redundant mark partition done

wwj6591812:
### Purpose

In our production environment, we use Flink bounded stream job write data to Paimon table. We use the 'partition mark done' feature to make downstream knows that our Flink bounded stream job has finished. But we found multi partition mark down will be trigger when :
1、Some writer task has finished.
2、Enable checkpoint.
3、Failover.
image
image
image
image
image

After troubleshooting, we found the reason is when some wirte task finish, follow code will be called:
image

Then when failover, the committer operator will mark partition done:
image

Finally, Our downstream started consuming before the Flink bounded stream job was completely finished, resulting in data loss online.

So, after discuss offline, we think when failover, no need to call PartitionListeners#notifyCommittable.

Linked issue: close #xxx

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/CommitterOperator.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/CommitterOperator.java
@@ -210,7 +210,7 @@ public class CommitterOperator<CommitT, GlobalCommitT> extends AbstractStreamOpe
             // So when `endInput` is called, we must check if the corresponding snapshot exists.
             // However, if the snapshot does not exist, then append files must be new files. So
             // there is no need to check for duplicated append files.
-            committer.filterAndCommit(committables, false);
+            committer.filterAndCommit(committables, false, false);
         } else {
             committer.commit(committables);
         }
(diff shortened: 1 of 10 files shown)


wwj6591812:
I think no need UT, because CommitterOperatorTest#testFailIntentionallyAfterRestore  and PartitionMarkDoneTest already cover my modify.

wwj6591812:
@JingsongLi Hi，Please CC，thx.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/CommitterOperator.java
```diff
@@ -210,7 +210,7 @@ private void commitUpToCheckpoint(long checkpointId) throws Exception {
             // So when `endInput` is called, we must check if the corresponding snapshot exists.
             // However, if the snapshot does not exist, then append files must be new files. So
             // there is no need to check for duplicated append files.
-            committer.filterAndCommit(committables, false);
+            committer.filterAndCommit(committables, false, true);
```
leaves12138, line 213: We can't set this to true always.

Add a parameter in `PartitionListener` to tag whether it is recover from failover.

wwj6591812: Thx, done.

leaves12138:
+1

JingsongLi:
@wwj6591812 @leaves12138 
During the normal process, PartitionMarkDone has not been completed yet. Even after restarting and restoring from state without notifying, it may result in some partitions being missed.

I think adding option is the appropriate solution. Should we go to MarkPartitionDown when restoring the state.

wwj6591812:
> @wwj6591812 @leaves12138 During the normal process, PartitionMarkDone has not been completed yet. Even after restarting and restoring from state without notifying, it may result in some partitions being missed.
> 
> I think adding option is the appropriate solution. Should we go to MarkPartitionDown when restoring the state.

@JingsongLi 
Thanks, See https://github.com/apache/paimon/pull/5648

PR #5582, opened 2025-05-08
[parquet] Fix timestamp type and decimal type, if the file schema is not correctly match the schema in metadata

leaves12138:
### Purpose

If the timestamp field, in paimon, precision is 3. But in file, it is Binary. We will get an exception.

Fix this issue, do the converter for timestamp type and decimal type.


JingsongLi:
+1

PR #5586, opened 2025-05-09
[flink] Clone action support Hudi table

yuzelin:
### Purpose

Currently, we only support hudi table in Hive metastore, and it will be convert to paimon unaware bucket table.


paimon-hive/paimon-hive-catalog/pom.xml
```diff
@@ -112,6 +112,35 @@ under the License.
...
+            <artifactId>hudi-flink</artifactId>
+            <version>${hudi.version}</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-core</artifactId>
+            <version>${paimon-flink-common.flink.version}</version>
+            <scope>provided</scope>
```
JingsongLi, line 134: there deps should be test?

paimon-hive/paimon-hive-catalog/pom.xml
```diff
@@ -112,6 +112,35 @@ under the License.
             </exclusions>
         </dependency>
 
+        <!-- for hudi clone -->
+        <dependency>
+            <groupId>org.apache.hudi</groupId>
+            <artifactId>hudi-common</artifactId>
```
JingsongLi, line 118: Can you just create a model for hudi? for example, paimon-hudi? And in hive catalog we can use java spi to load hudi files factory.

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hudi/FileIndex.java
```diff
@@ -0,0 +1,253 @@
...
+
+import static org.apache.hudi.configuration.FlinkOptions.METADATA_ENABLED;
+import static org.apache.paimon.utils.Preconditions.checkState;
+
+/**
+ * A file index which supports listing files efficiently through metadata table.
+ *
+ * @see org.apache.hudi.source.FileIndex
+ */
+public class FileIndex {
```
JingsongLi, line 66: HudiFileIndex

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/clone/HiveCloneUtils.java
```diff
@@ -158,6 +161,23 @@ public static Schema hiveTableToPaimonSchema(HiveCatalog hiveCatalog, Identifier
         return schemaBuilder.build();
     }
 
+    private static List<FieldSchema> getHiveSchema(
+            IMetaStoreClient client, Table hiveTable, String database, String table)
+            throws Exception {
+        List<FieldSchema> fields = client.getSchema(database, table);
+        if (!HudiCloneUtils.isHoodieTable(hiveTable)) {
```
JingsongLi, line 168: Maybe we can have an interface `HiveCloneExtractor`:
```java
interface HiveCloneExtractor {
      boolean matches();
    
      List<FieldSchema> extractSchema(IMetaStoreClient client, Table hiveTable, String database, String table);

     List<HivePartitionFiles> extractFiles(HiveCatalog hiveCatalog,
            Identifier identifier,
            RowType partitionRowType,
            String defaultPartitionName,
            @Nullable PartitionPredicate predicate) throws Exception;
}
```

JingsongLi:
And please add documentation too.

JingsongLi:
And please test with Hudi 0.8 tables, this should work for most of Hudi versions.

JingsongLi:
Fixed comments in f4b8e3dd96b49304f1c576ac7e4549efa7a3de25

PR #5589, opened 2025-05-12
[core][doc] Support creating multiple indexes on one column and correct parameter name for specifying bitmap index version.

hang8929201:
### Purpose

The paimon documentation does not clearly state that only one index is allowed to be created on a column. If multiple indexes are created on a column during use, no error will be reported but the index will be lost. Therefore, we have fixed the relevant code logic to allow multiple indexes to be created on a column.

We also found an error in the parameter name used to specify the bitmap version in the documentation and corrected it.

### Tests

org.apache.paimon.io.DataFileIndexWriterTest::testCreatingMultipleIndexesOnOneColumn

### DIFF
--- a/docs/content/concepts/spec/fileindex.md
+++ b/docs/content/concepts/spec/fileindex.md
@@ -170,7 +170,7 @@ length:                        4 bytes int
 
 (Legacy) Bitmap file index format (V1):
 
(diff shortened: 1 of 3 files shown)


JingsongLi:
+1

PR #5590, opened 2025-05-12
[spark], [infra] run spark integration tests in CI.

zhongyujiang:
### Purpose

The current github CI for Spark module is missing integration tests, and some of Spark's integration tests are actually failing, they've just been consistently ignored by the CI.

### DIFF
--- a/.github/workflows/utitcase-spark-4.x.yml
+++ b/.github/workflows/utitcase-spark-4.x.yml
@@ -58,6 +58,6 @@ jobs:
           test_modules+="org.apache.paimon:paimon-spark-${suffix},"
           done
           test_modules="${test_modules%,}"
-          mvn -T 2C -B test -pl "${test_modules}" -Duser.timezone=$jvm_timezone -Pspark4,flink1
+          mvn -T 2C -B verify -pl "${test_modules}" -Duser.timezone=$jvm_timezone -Pspark4,flink1
         env:
           MAVEN_OPTS: -Xmx4096m
--- a/.github/workflows/utitcase-spark-3.x.yml
+++ b/.github/workflows/utitcase-spark-3.x.yml
@@ -58,6 +58,6 @@ jobs:
           test_modules+="org.apache.paimon:paimon-spark-${suffix},"
           done
           test_modules="${test_modules%,}"
-          mvn -T 2C -B test -pl "${test_modules}" -Duser.timezone=$jvm_timezone
+          mvn -T 2C -B verify -pl "${test_modules}" -Duser.timezone=$jvm_timezone
         env:
           MAVEN_OPTS: -Xmx4096m
--- a/paimon-core/src/main/java/org/apache/paimon/table/AbstractFileStoreTable.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/AbstractFileStoreTable.java
@@ -479,18 +479,13 @@ abstract class AbstractFileStoreTable implements FileStoreTable {
     }
 
     private Optional<TableSchema> tryTimeTravel(Options options) {
-        Snapshot snapshot;
-        try {
-            snapshot =
-                    TimeTravelUtil.tryTravelToSnapshot(options, snapshotManager(), tagManager())
-                            .orElse(null);
-        } catch (Exception e) {
-            return Optional.empty();
-        }
-        if (snapshot == null) {
-            return Optional.empty();
-        }
-        return Optional.of(schemaManager().schema(snapshot.schemaId()).copy(options.toMap()));
+        Optional<Snapshot> snapshot =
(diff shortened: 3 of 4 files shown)


.github/workflows/utitcase-spark-4.x.yml
```diff
@@ -58,6 +58,6 @@ jobs:
           test_modules+="org.apache.paimon:paimon-spark-${suffix},"
           done
           test_modules="${test_modules%,}"
-          mvn -T 2C -B test -pl "${test_modules}" -Duser.timezone=$jvm_timezone -Pspark4,flink1
+          mvn -T 2C -B verify -pl "${test_modules}" -Duser.timezone=$jvm_timezone -Pspark4,flink1
```
zhongyujiang, line 61: `test` phase only cover unit tests, not integration tests.

zhongyujiang (commented):
cc @Zouxxyy @YannByron Can you please take a look when you  have time? Thanks!

zhongyujiang:
Some unit tests are failing, it looks like it's releated to the changes in the time travel part. Let me take a look.

.github/workflows/utitcase-spark-4.x.yml
Zouxxyy: Oh, under `test`, won't `SparkWriteITCase` be tested?

zhongyujiang:
@Zouxxyy Hi, sorry, I missed your message earlier. Previously, mvn test didn't run the tests for these ITCases, so I pulled this PR. Let me fix test failures and resolve these conflicts.

Zouxxyy:
Can you replace the comparison test logic in `testUpdateNestedColumnTypeInArray` with `checkAnswer(df, Seq())`

```
Error:    SparkSchemaEvolutionITCase.testUpdateNestedColumnTypeInArray:1019 
Expecting actual:
  ["[1,ArraySeq([apple,100], [banana,101])]",
    "[2,ArraySeq([cat,200], [dog,201])]"]
to contain exactly in any order:
  ["[1,WrappedArray([apple,100], [banana,101])]",
    "[2,WrappedArray([cat,200], [dog,201])]"]
elements not found:
  ["[1,WrappedArray([apple,100], [banana,101])]",
    "[2,WrappedArray([cat,200], [dog,201])]"]
and elements not expected:
  ["[1,ArraySeq([apple,100], [banana,101])]",
    "[2,ArraySeq([cat,200], [dog,201])]"]
```

zhongyujiang:
@Zouxxyy Hi, thank you for reminding me about the `checkAnswer`. I was planning to write a custom comparison method myself. However, this method is a protected method in a Scala class, and to use it, I would need to convert the entire test `SparkSchemaEvolutionITCase` class to Scala. Do you think that's okay?

Zouxxyy:
> @Zouxxyy Hi, thank you for reminding me about the `checkAnswer`. I was planning to write a custom comparison method myself. However, this method is a protected method in a Scala class, and to use it, I would need to convert the entire test `SparkSchemaEvolutionITCase` class to Scala. Do you think that's okay?

Thanks, either is fine, you can choose the one with less work

paimon-spark/paimon-spark-ut/src/test/scala/org/apache/paimon/spark/RowTestHelper.scala
```diff
@@ -0,0 +1,53 @@
...
+import org.apache.spark.sql.Row
+import org.apache.spark.sql.SparkSession
+
+import scala.collection.JavaConverters._
+
+/**
+ * A helper class for facilitating the comparison of Spark Row objects in Java unit tests, which
+ * leverages QueryTest.checkAnswer for the comparison.
+ */
+class RowTestHelper extends QueryTest {
```
zhongyujiang, line 33: Added this helper class for comparing rows in Java.

paimon-core/src/main/java/org/apache/paimon/table/source/snapshot/TimeTravelUtil.java
```diff
@@ -147,6 +147,8 @@ private static void adaptScanVersion(Options options, TagManager tagManager) {
         } else if (version.chars().allMatch(Character::isDigit)) {
             options.set(SCAN_SNAPSHOT_ID.key(), version);
         } else {
+            // by here, the scan version should be a tag.
+            options.set(SCAN_TAG_NAME.key(), version);
```
zhongyujiang, line 151: Previously, when querying a tag using the `VERSION AS OF` syntax, if a tag did not exist, the query would not throw an error but instead return the result of the latest snapshot, which is wrong. This is  because the scan version was removed from the options during time travel.

JingsongLi (approved):
+1

PR #5592, opened 2025-05-12
[core] Postpone bucket should introduce a new BucketMode

JingsongLi:
### Purpose

The previous `HASH_FIXED` pattern already assumed that the `bucket` was a positive value, which is already assumed in many engines. We need to introduce a new mode to Postpone Bucket to avoid bugs.

For example, Spark bucketed Join.

### Tests

We should add tests for Spark postpone bucket reading, but currently, we lack the Table API for compact postpone bucket table. We should add these table api later.


wwj6591812:
+1

Zouxxyy (approved):
+1

PR #5595, opened 2025-05-13
[parquet] Support the nonstandard list and map parquet message type for hive migration

leaves12138:
### Purpose

The map and list type migrated from hive, may be non-standard. So the middle message type:
repeated group <name> may not be "list" (which for list type) and "key_value" (which for map type).

Support this situation for reading data from hive migration.


JingsongLi (approved):
+1

PR #5597, opened 2025-05-13
[hotfix] add test for ro fallback

jerry-024:
### Purpose
Add test for to table fallback when: main branch is append table and fallback branch is pk table

### Tests
- CatalogTableITCase.testReadOptimizedTableFallBack


JingsongLi:
+1

PR #5599, opened 2025-05-13
[hotfix] Fixed error words from unware to unaware.

zhuangchong:
### Purpose

Fixed error words from unware to unaware.


JingsongLi:
+1

PR #5602, opened 2025-05-14
[parquet] Support parquet two level list type protobuf representation

leaves12138:
### Purpose

Parquet list may have two level representation protos.

For example, this is a standard two level list type protobuf schema in parquet:
```
ParquetSchemaExampleB {
  required group list_of_ints (LIST) {
    repeated int32 list_of_ints_tuple;
  }
}
```

While this is the three level representation:
```
ParquetSchemaExampleA {
  optional group list_of_ints (LIST) {
    repeated group list {
      required int32 element;
    }
  }
}
```

see link: https://impala.apache.org/docs/build/html/topics/impala_parquet_array_resolution.html


JingsongLi:
+1

PR #5603, opened 2025-05-14
[core] Remove paimon-open-api module

JingsongLi:
### Purpose

1. Automatic generation tools are currently not useful because many objects cannot be recognized.
2. The Yaml definition is already in the document, we don't need to repeat the definition.

So remove whole paimon-open-api module.


jerry-024:
+1

PR #5605, opened 2025-05-14
[core] Introduce paimon-api module for light rest api, part 1

JingsongLi:
### Purpose

This module is for light API, aims to simplify dependencies as much as possible and avoid
introducing dependencies such as Hadoop, including:

1. Including types, table, view, function and etc.
2. Including http client REST API.


jerry-024:
+1

PR #5606, opened 2025-05-15
[core] Use LocalDate#parse to replace LocalDateTime#parse

wwj6591812:
### Purpose

In our company, we found the expire_partition action runs very slow.
We try to jstack and found:
image
and the params of we call PartitionTimeExtractor#toLocalDateTime is "20250510" and "yyyyMMdd".
After ut, we found the exception is : "ava.time.format.DateTimeParseException: Text '20250510' could not be parsed: Unable to obtain LocalDateTime from TemporalAccessor: {},ISO resolved to 2025-05-10 of type java.time.format.Parsed".

So we want optimize this.

Linked issue: close #xxx

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/partition/PartitionTimeExtractor.java
+++ b/paimon-core/src/main/java/org/apache/paimon/partition/PartitionTimeExtractor.java
@@ -112,7 +112,8 @@ public class PartitionTimeExtractor {
         DateTimeFormatter dateTimeFormatter =
                 DateTimeFormatter.ofPattern(Objects.requireNonNull(formatterPattern), Locale.ROOT);
         try {
-            return LocalDateTime.parse(timestampString, Objects.requireNonNull(dateTimeFormatter));
+            return LocalDate.parse(timestampString, Objects.requireNonNull(dateTimeFormatter))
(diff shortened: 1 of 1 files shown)


paimon-core/src/main/java/org/apache/paimon/partition/PartitionTimeExtractor.java
```diff
@@ -112,7 +112,8 @@ private static LocalDateTime toLocalDateTime(
         DateTimeFormatter dateTimeFormatter =
                 DateTimeFormatter.ofPattern(Objects.requireNonNull(formatterPattern), Locale.ROOT);
         try {
-            return LocalDateTime.parse(timestampString, Objects.requireNonNull(dateTimeFormatter));
+            return LocalDate.parse(timestampString, Objects.requireNonNull(dateTimeFormatter))
+                    .atStartOfDay();
```
JingsongLi, line 116: What about hour partition?

wwj6591812: If timestampString is "2025051008" and formatterPattern is "yyyyMMddHH".
The exception will not happen.

PR #5608, opened 2025-05-15
[hotfix] Fix tag expiration is unexpect for batch write

askwang:
### Purpose

Table set `tag.num-retained-max=1`, batch write twice, the total tag is empty after the sencond write.

```sql
set spark.paimon.tag.batch.customized-name=tag-1;
insert into tb (1, 'a');  // generate tag-1

set spark.paimon.tag.batch.customized-name=tag-2;
insert into tb (2, 'b');   // both tag-1 and tag-2 expired
```


JingsongLi:
CC @Aitozi

PR #5609, opened 2025-05-15
[core]Improve partial update merge engine configuration ignore-delete and remove-record-on-sequence-group cannot be enabled at the same time.

zhuangchong:
### Purpose

1. check ignore-delete and remove-record-on-sequence-group cannot be enabled at the same time
2. optimize sequenceFields from storage field names to field indexes to reduce duplicate operations
image

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/PartialUpdateMergeFunction.java
+++ b/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/PartialUpdateMergeFunction.java
@@ -37,8 +37,6 @@ import org.apache.paimon.utils.Preconditions;
 import org.apache.paimon.utils.Projection;
 import org.apache.paimon.utils.UserDefinedSeqComparator;
 
-import org.apache.paimon.shade.guava30.com.google.common.primitives.Ints;
-
 import javax.annotation.Nullable;
 
 import java.util.ArrayList;
@@ -422,7 +420,6 @@ public class PartialUpdateMergeFunction implements MergeFunction<KeyValue> {
                                                     .split(FIELDS_SEPARATOR))
                                     .mapToInt(fieldName -> requireField(fieldName, fieldNames))
                                     .toArray();
-                    allSequenceFields = Ints.asList(sequenceFields);
 
                     Supplier<FieldsComparator> userDefinedSeqComparator =
                             () -> UserDefinedSeqComparator.create(rowType, sequenceFields, true);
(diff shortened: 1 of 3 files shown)


paimon-core/src/main/java/org/apache/paimon/mergetree/compact/PartialUpdateMergeFunction.java
```diff
@@ -391,45 +393,41 @@ private static class Factory implements MergeFunctionFactory<KeyValue> {
...
                                                             k.length()
                                                                     - SEQUENCE_GROUP.length()
                                                                     - 1)
                                                     .split(FIELDS_SEPARATOR))
-                                    .map(fieldName -> validateFieldName(fieldName, fieldNames))
-                                    .collect(Collectors.toList());
-                    allSequenceFields.addAll(sequenceFields);
+                                    .mapToInt(fieldName -> requireField(fieldName, fieldNames))
+                                    .toArray();
+                    allSequenceFields = Ints.asList(sequenceFields);
```
yuzelin, line 425: Should be `allSequenceFields.addAll()` because here is in a for loop.

paimon-core/src/main/java/org/apache/paimon/mergetree/compact/PartialUpdateMergeFunction.java
```diff
@@ -442,9 +440,9 @@ private Factory(Options options, RowType rowType, List<String> primaryKeys) {
                                     });
 
                     // add self
-                    sequenceFields.forEach(
-                            fieldName -> {
-                                int index = fieldNames.indexOf(fieldName);
+                    allSequenceFields.forEach(
```
yuzelin, line 443: Here should use `sequenceFields` (the same reason that here is in a for loop).

paimon-core/src/main/java/org/apache/paimon/mergetree/compact/PartialUpdateMergeFunction.java
zhuangchong: done.

paimon-core/src/main/java/org/apache/paimon/mergetree/compact/PartialUpdateMergeFunction.java
zhuangchong: done.

zhuangchong:
thanks. @yuzelin

yuzelin (approved):

PR #5611, opened 2025-05-16
[flink] by default, use paimon catalog to create views in flink generic catalog.

### LINKED ISSUE #5489: [Feature] How to create view in FlinkGenericCatalog ?
### Search before asking

### Motivation

I have a Flink Generic Catalog which Hive Metastore. 

But i want to create a view. It seem prompt me 

```sql
create view paimon_gen_catalog.mydb.`my_overall_share` as 
	select * from paimon_gen_catalog.mydb.`share` where id = 1313;

```

Caused by: java.lang.RuntimeException: FlinkGenericCatalog can not create table without 'connector' key.
	at org.apache.paimon.flink.FlinkGenericCatalog.createTable(FlinkGenericCatalog.java:184) ~[paimon-flink-1.20-1.0.1.jar:1.0.1]

### Solution

_No response_

### Anything else?

_No response_

zhuangchong:
### Purpose

Linked issue: close #5489

By default, use the Paimon catalog to create the view. Create views with options is not supported.


JingsongLi:
+1

Issue #5420, opened 2025-04-08
[Feature] Support function in paimon


jerry-024:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Motivation

The defining function includes the parameter, output, and function body. The output result is a single value or a table. It is modeled after unity catalog，and it could support the following scenarios:
1. Support filtering or encrypting column in row: select decode_email(email) from user; 
2. Support parameterized view: select filter_fruit_by_color('red');
3. Support use the function in AI.

### Solution

Provide the ability to define function.
[PIP-31: Introduce Function](https://cwiki.apache.org/confluence/display/PAIMON/PIP-31%3A+Introduce+Function)

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #5613, opened 2025-05-16
[core] fix: function need belong to database

jerry-024:
### Purpose

Linked issue: #5420

update function api
- function belong to database
- inputParams and returnParams is optional


JingsongLi (approved):
+1

PR #5621, opened 2025-05-19
[core] (work in progress) Iceberg REST

nickdelnano:
This is a work in progress and not ready for merging. I am opening it as a draft to share my work and to post my notes. Feedback is welcomed and appreciated.

I am exploring two design decisions:

1.  Use [Paimon RESTCatalog](https://github.com/apache/paimon/blob/master/paimon-core/src/main/java/org/apache/paimon/rest/RESTCatalog.java#L88) or [Iceberg RESTCatalog](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java#L47)
2. If using Iceberg RESTCatalog, whether to use it as an http client directly and call functions like [`create`](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java#L754) or through a Spark/Flink/Trino catalog instance, like [this](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java#L47)) and use DDL for create database, create table, alter table.

There are pros and cons to each. Some factors are:
- The Paimon and Iceberg REST APIs are not the same. They would need to be adapted. This encourages the use of Iceberg RESTCatalog.
- Paimon has 2 REST auth providers [(bear, dlf)](https://paimon.apache.org/docs/master/concepts/rest/overview/) and these rely on using classes in the Paimon codebase to function. If using Iceberg RESTCatalog they would require code changes to work.
- Using a Spark/Flink/Trino catalog and DDL would require some engine specific code to execute statements to create database, create table, alter table. Currently `IcebergCommitCallback` and related classes are not specific to any engine.

My personal motivation is to get Iceberg REST working with the REST catalog servers provided by Apache Polaris and AWS Glue but I also want to support the community's interests.

### DIFF
--- a/paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/RESTCatalogITCase.java
+++ b/paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/RESTCatalogITCase.java
@@ -19,13 +19,26 @@
 package org.apache.paimon.flink;
 
 import org.apache.paimon.catalog.Identifier;
+import org.apache.paimon.options.CatalogOptions;
+import org.apache.paimon.rest.RESTCatalogInternalOptions;
+import org.apache.paimon.rest.RESTCatalogServer;
 import org.apache.paimon.rest.RESTToken;
+import org.apache.paimon.rest.RESTTokenFileIO;
+import org.apache.paimon.rest.auth.AuthProvider;
+import org.apache.paimon.rest.auth.BearTokenAuthProvider;
+import org.apache.paimon.rest.responses.ConfigResponse;
 
 import org.apache.paimon.shade.guava30.com.google.common.collect.ImmutableMap;
 
+import org.apache.flink.table.api.TableEnvironment;
+import org.apache.flink.table.api.TableResult;
 import org.apache.flink.types.Row;
+import org.apache.flink.util.CloseableIterator;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.UUID;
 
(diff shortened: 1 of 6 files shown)


paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/RESTCatalogITCase.java
```diff
@@ -102,4 +115,95 @@ public void testExpiredDataToken() {
         assertThat(batchSql(String.format("SELECT * FROM %s.%s", DATABASE_NAME, TABLE_NAME)))
                 .containsExactlyInAnyOrder(Row.of("1", 11.0D), Row.of("2", 22.0D));
     }
+
+    @ParameterizedTest
+    // TODO add other file types back
+    @ValueSource(strings = {"avro"})
+    public void testIcebergRESTCatalog(String format) throws Exception {
```
nickdelnano, line 122: This test is an example. It does not pass.

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/RESTCatalogITCase.java
```diff
@@ -102,4 +115,95 @@ public void testExpiredDataToken() {
...
+        tEnv.executeSql(
+                        "INSERT INTO paimon.`default`.T VALUES "
+                                + "(1, MAP[10, ARRAY[ROW('apple', 100), ROW('banana', 101)], 20, ARRAY[ROW('cat', 102), ROW('dog', 103)]], 1000), "
+                                + "(2, MAP[10, ARRAY[ROW('cherry', 200), ROW('pear', 201)], 20, ARRAY[ROW('tiger', 202), ROW('wolf', 203)]], 2000)")
+                .await();
+
+        tEnv.executeSql(
+                "CREATE CATALOG iceberg WITH (\n"
+                        + "  'type' = 'iceberg',\n"
+                        + "  'catalog-type' = 'rest',\n"
```
nickdelnano, line 177: Iceberg rest catalog

JingsongLi:
Thanks @nickdelnano for the proposal! +1 to your analysis, implement an `IcebergRESTMetadataCommitter` looks good to me!

PR #5623, opened 2025-05-19
[core] Start channel should not be negative when partition.hashCode() equals to Integer.MIN_VALUE

tsreaper:
### Purpose

Currently in `ChannelComputer`, if `partition.hashCode()` equals to `Integer.MIN_VALUE`, the resulting channel will also be negative. This PR fixes the issue.

### Tests

None.

### API and Format

No format changes.

### Documentation

No new feature.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/sink/ChannelComputer.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/sink/ChannelComputer.java
@@ -35,12 +35,25 @@ public interface ChannelComputer<T> extends Serializable {
     int channel(T record);
 
     static int select(BinaryRow partition, int bucket, int numChannels) {
-        int startChannel = Math.abs(partition.hashCode()) % numChannels;
-        return (startChannel + bucket) % numChannels;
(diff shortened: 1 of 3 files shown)


paimon-core/src/main/java/org/apache/paimon/table/sink/ChannelComputer.java
```diff
@@ -35,14 +35,27 @@ public interface ChannelComputer<T> extends Serializable {
...
+        if (hashCode == Integer.MIN_VALUE) {
+            hashCode = Integer.MAX_VALUE;
+        }
+        // Due to backward compatibility (Flink users may recover from state),
+        // we need to use this formula.
+        // However, if hashCode equals Integer.MIN_VALUE,
+        // Math.abs will still return Integer.MIN_VALUE,
+        // and this formula will produce a negative integer.
+        // So we specially handle this case above.
+        return Math.abs(hashCode) % numChannels;
```
JingsongLi, line 56: Add a test for this.

JingsongLi:
+1

PR #5629, opened 2025-05-19
[WIP][rest] support RepairTable on RestCatalog

sundapeng:
### Purpose

Add RepairTable API for RestCatalog

### Tests

TODO

### API and Format

Add `RepairTableWithOptions`

### Documentation

TODO

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
@@ -562,7 +562,7 @@ public interface Catalog extends AutoCloseable {
         throw new UnsupportedOperationException();
     }
 
-    /** Repair the table with options */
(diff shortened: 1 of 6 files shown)


paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
```diff
@@ -562,6 +562,11 @@ default void repairTable(Identifier identifier) throws TableNotExistException {
         throw new UnsupportedOperationException();
     }
 
+    /** Repair the table with options. */
+    default void repairTable(Identifier identifier, Map<String, String> repairOptions) {
```
JingsongLi, line 566: What is `repairOptions`?

JingsongLi:
We have introduced register table for REST Api.

PR #5630, opened 2025-05-19
[flink-common] Don't allow converting nullable columns and fields to non nullable

AshishKhatkar:
### Purpose
This PR disallows converting nullable column types to non nullable during schema evolution.

Linked issue: close #xxx

As of today, adding new column enforces that it has to be nullable but anyone can circumvent that condition by adding a null column and then changing the column type to be non nullable. Consider the case as follows:

```

- USE CATALOG paimon;

- CREATE TABLE evolution_test(field1 INT, field2 INT, PRIMARY KEY (field1) NOT ENFORCED);

- DESCRIBE evolution_test;

+--------+------+-------+-------------+--------+-----------+
|   name | type |  null |         key | extras | watermark |
+--------+------+-------+-------------+--------+-----------+
| field1 |  INT | FALSE | PRI(field1) |        |           |
| field2 |  INT |  TRUE |             |        |           |
+--------+------+-------+-------------+--------+-----------+

- INSERT INTO evolution_test VALUES (1, 1);
- INSERT INTO evolution_test VALUES (2, CAST (NULL AS INT));
- SELECT * FROM evolution_test;
- # this returns the rows correctly

- ALTER TABLE evolution_test MODIFY field2 INT NOT NULL;
- SELECT * FROM evolution_test;
# this fails as field2 has null value written into it but schema specifies that field2 can't be null

- ALTER TABLE evolution_test ADD (field3 INT NOT NULL);
# this fails as it expects the newly added fields as nullable.

# but if I first add new column as nullable and then convert it to non null it works
- ALTER TABLE evolution_test ADD (field3 INT);
- ALTER TABLE evolution_test MODIFY field3 INT NOT NULL;
# this succeeds
```

### Tests

Added new tests to verify the change for complex types (Row, Array and Map) as well.

### DIFF
--- a/paimon-api/src/main/java/org/apache/paimon/options/CatalogOptions.java
+++ b/paimon-api/src/main/java/org/apache/paimon/options/CatalogOptions.java
@@ -184,12 +184,4 @@ public class CatalogOptions {
                     .defaultValue(false)
                     .withDescription(
                             "Whether to populate file metadata while listing or getting file status.");
-
-    public static final ConfigOption<Boolean> DISABLE_ALTER_COLUMN_NULL_TO_NOT_NULL =
-            ConfigOptions.key("alter-column-null-to-not-null.disabled")
-                    .booleanType()
-                    .defaultValue(true)
-                    .withDescription(
-                            "If true, it disables altering column type from null to not null. Default is true. "
-                                    + "Users can disable this option to explicitly convert null column type to not null.");
 }
--- a/paimon-api/src/main/java/org/apache/paimon/options/CatalogOptions.java
+++ b/paimon-api/src/main/java/org/apache/paimon/options/CatalogOptions.java
@@ -184,4 +184,12 @@ public class CatalogOptions {
                     .defaultValue(false)
                     .withDescription(
                             "Whether to populate file metadata while listing or getting file status.");
+
+    public static final ConfigOption<Boolean> DISABLE_ALTER_COLUMN_NULL_TO_NOT_NULL =
+            ConfigOptions.key("alter-column-null-to-not-null.disabled")
+                    .booleanType()
+                    .defaultValue(true)
+                    .withDescription(
+                            "If true, it disables altering column type from null to not null. Default is true. "
+                                    + "Users can disable this option to explicitly convert null column type to not null.");
 }
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkCatalog.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkCatalog.java
@@ -783,10 +783,6 @@ public class FlinkCatalog extends AbstractCatalog {
         }
 
         if (oldType.isNullable() != newType.isNullable()) {
-            Preconditions.checkArgument(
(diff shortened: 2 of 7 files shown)


JingsongLi (commented):
Thanks @AshishKhatkar for the contribution.

I remember that in some scenarios, it is necessary to change from 'null' to 'not null', as long as the business ensures that its data is not null.

There are two ways to consider:
1. Add an option to determine if it is possible.
2. When making changes, check the statistical information of the file and see if there is any null data in this table.

AshishKhatkar:
>  I remember that in some scenarios, it is necessary to change from 'null' to 'not null', as long as the business ensures that its data is not null.

If that is the case shouldn't business start with not null columns. If they do have this scenario they can run a migration from one table to another and then using that new table with updated schema. In the current scenario paimon specifically disallows adding not null columns which is redundant if we allow converting null to non null. In general I don't see how going from null to not null makes sense.

> Add an option to determine if it is possible.

What does this mean? Is it a user config that you are talking about using which user can choose to allow null to not null evolution which solves for the business case you mentioned?

> When making changes, check the statistical information of the file and see if there is any null data in this table.

Is there any kind of statistics maintained for a table as of today to determine this information?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkCatalog.java
```diff
@@ -783,6 +783,10 @@ private void generateNestedColumnUpdates(
         }
 
         if (oldType.isNullable() != newType.isNullable()) {
+            Preconditions.checkArgument(
```
JingsongLi, line 786: We should modify `SchemaManager.generateTableSchema` instead of here.

AshishKhatkar:
> Hi @AshishKhatkar
> 
> > What does this mean? Is it a user config that you are talking about using which user can choose to allow null to not null evolution which solves for the business case you mentioned?
> 
> Yes, we can introduce a table option: `alter-column-not-null.disabled`, its default value is true, we should add its key in the exception message.

Cool, I will update the PR soon.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkCatalog.java
AshishKhatkar: This is done.

AshishKhatkar:
@JingsongLi can you check the PR now.

paimon-api/src/main/java/org/apache/paimon/options/CatalogOptions.java
```diff
@@ -184,4 +184,12 @@ public class CatalogOptions {
                     .defaultValue(false)
                     .withDescription(
                             "Whether to populate file metadata while listing or getting file status.");
+
+    public static final ConfigOption<Boolean> DISABLE_ALTER_COLUMN_NULL_TO_NOT_NULL =
```
JingsongLi, line 188: Add this to CoreOptions.

AshishKhatkar: Done

JingsongLi (approved):
Looks good to me!

PR #5632, opened 2025-05-20
[core] keep lowest high-level kv for lookup compaction.

### LINKED ISSUE #5631: [Bug] lookup compaction may lead to wrong results.
### Search before asking

### Paimon version

1.0

### Compute Engine

flink 1.18

### Minimal reproduce step

`LookupChangelogMergeFunctionWrapper function =
                new LookupChangelogMergeFunctionWrapper(
                        LookupMergeFunction.wrap(DeduplicateMergeFunction.factory()),
                        highLevel::get,
                        null,
                        LookupStrategy.from(false, true, false, false),
                        null,
                        null);

        function.reset();
        function.add(new KeyValue().replace(row(1), 1, INSERT, row(2)).setLevel(1));
        function.add(new KeyValue().replace(row(1), 1, INSERT, row(1)).setLevel(2));
        ChangelogResult result = function.getResult();
        assertThat(result).isNotNull();
        assertThat(result.changelogs()).isEmpty();
        KeyValue kv = result.result();
        assertThat(kv).isNotNull();
        assertThat(kv.value().getInt(0)).isEqualTo(2);`

Image

A key may exist in multi high-levels and have same sequence number，when merge the same key in multi high-levels with lookup compaction, we should keep the record with lowest level number. The current implement cannot guarantee this.

### Anything else?

_No response_

zhoulii:
### Purpose

Linked issue: close #5631

### Tests

`org.apache.paimon.mergetree.compact.LookupChangelogMergeFunctionWrapperTest#testKeepLowestHighLevel`

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/LookupChangelogMergeFunctionWrapper.java
+++ b/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/LookupChangelogMergeFunctionWrapper.java
@@ -114,7 +114,7 @@ public class LookupChangelogMergeFunctionWrapper<T>
             KeyValue kv = descending.next();
             if (kv.level() > 0) {
                 descending.remove();
-                if (highLevel == null) {
+                if (highLevel == null || kv.level() < highLevel.level()) {
                     highLevel = kv;
                 }
             } else {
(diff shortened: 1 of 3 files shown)


JingsongLi:
Is it a binding relationship with lookup? Is there a problem with defining sequence numbers?

JingsongLi:
Looks good to me!

PR #5633, opened 2025-05-20
[api] fix: 404 exception comment for snapshot api

jerry-024:
### Purpose

fix: 404 exception comment for snapshot api


zhuangchong (approved):
+1

PR #5634, opened 2025-05-20
[rest] Adjust auth table for row level access control

JingsongLi:
### Purpose

For row level access control, the general design is to add additional filtering conditions to make users see less data, so we need to return additional filtering instead of passing the filtering conditions to the server.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
@@ -791,14 +791,14 @@ public interface Catalog extends AutoCloseable {
     // ==================== Table Auth ==========================
 
     /**
-     * Auth table query select and filter.
+     * Auth table query select and get the filter for row level access control.
      *
      * @param identifier path of the table to alter partitions
-     * @param select selected fields
-     * @param filter query filters
+     * @param select selected fields, null if select all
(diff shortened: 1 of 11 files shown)


paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
```diff
@@ -791,14 +791,14 @@ void alterFunction(
...
      *
      * @param identifier path of the table to alter partitions
-     * @param select selected fields
-     * @param filter query filters
+     * @param select selected fields, null if select all
+     * @return additional filter for row level access control
      * @throws TableNotExistException if the table does not exist
      */
-    void authTableQuery(Identifier identifier, List<String> select, List<String> filter)
+    List<String> authTableQuery(Identifier identifier, @Nullable List<String> select)
```
shyjsarah, line 801: When will select be null ? For example, in the query "select *", is the select null or all fields?

JingsongLi: See comments, null if select all

shyjsarah: ok, I get it

plusplusjiajia:
LGTM, +1.

PR #5636, opened 2025-05-20
[ci] Build first for 'UTCase and ITCase Others'

JingsongLi:
### Purpose

In order to make the test logs smaller when failure


jerry-024:
+1

PR #5637, opened 2025-05-20
[refactor] Clarify option commit.force-create-snapshot

yuzelin:
### Purpose

The description of `commit.force-create-snapshot` is confusing.


JingsongLi:
+1

PR #5638, opened 2025-05-20
[core] pick large files which have dv index for full compaction

LsomeYeah:
### Purpose

Linked issue: close #xxx

In deletion-vector mode, all deletion vector index should be removed and the corresponding data files should be rewrite after full compaction. But for the following case below:

Only one file in maxLevel, and using `DELETE` sql to delete some records, it'll generate a dv for the file in maxLevel. When performing full compaction, it' possible that the file won't be picked for compaction, causing that the dv index can't be totally removed after full compaction which is unexpected.

### Tests

DeletionVectorITCase#testRemoveDvsAfterFullCompaction: test whether we can select the correct data after disabling dv.


JingsongLi (approved):
+1

PR #5644, opened 2025-05-21
[rest] Add list partitions tests for non-metastore part table

JingsongLi:
### Purpose

1. Add list partitions tests for non-metastore part table
2. List partitions should return list desc to show latest partition


qingwei727:
+1

PR #5648, opened 2025-05-22
[Flink] Hot fix of partition mark down bug.

wwj6591812:
### Purpose

After https://github.com/apache/paimon/pull/5581, we found a potential bug:

When a checkpoint finished, it should notify PartitionMarkDone in ComminterOperator#notifyCheckpointComplete. But failover before call ComminterOperator#notifyCheckpointComplete. 
Then after restarted, RestoreAndFailedCommittableStateManager#recover will be called, the PartitionMarkDone will not notify, which miss the PartitionMarkDone.

So we add an option here which means whether trigger partition mark done when recover from state, the default value is true.

Linked issue: close #xxx


JingsongLi (approved):
+1

PR #5651, opened 2025-05-22
[hudi][clone] Fix Hudi clone to pass configuration for Hudi extractFiles

timmyyao:
CloneHive tool meets problem when clone Hudi table to paimon, which is caused by some necessary FileSystem configuration not passed to Hudi connector.


JingsongLi (approved):
+1

PR #5652, opened 2025-05-22
[core] Fix npe problem in bitmap file index v2

hang8929201:
### Purpose

When all values ​​are empty, a null pointer exception will be generated during the serialization of bitmap index v2.

### Tests

org.apache.paimon.fileindex.bitmapindex.BitmapFileIndexTest::testAllNull


JingsongLi (approved):
+1

PR #5662, opened 2025-05-25
[spark] Avoid partition to be negative when partition.hashCode() equals to Integer.MIN_VALUE

Zouxxyy:
### Purpose

like https://github.com/apache/paimon/pull/5623


JingsongLi:
+1

PR #5666, opened 2025-05-26
[core] Call Iceberg callback on replace_tag procedure

nickdelnano:
Issue:  https://github.com/apache/paimon/pull/5570

Relates to https://github.com/apache/paimon/pull/5570

I will use Iceberg tags in production soon and want replace_tag to update Iceberg metadata. Otherwise the Iceberg tags can become inconsistent with Paimon tags.

In https://github.com/apache/paimon/pull/4346#discussion_r1807119066 there was discussion of not adding callbacks on tag replace. For Iceberg it makes sense. The other callbacks - success file and add partition - may not be idempotent so I only added the Iceberg callback.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/utils/TagManager.java
+++ b/paimon-core/src/main/java/org/apache/paimon/utils/TagManager.java
@@ -122,7 +122,7 @@ public class TagManager {
                 tagName,
                 timeRetained,
                 callbacks.stream()
-                        .filter(callback -> IcebergCommitCallback.class.isInstance(callback))
+                        .filter(callback -> callback instanceof IcebergCommitCallback)
                         .collect(Collectors.toList()));
     }
 
--- a/paimon-core/src/main/java/org/apache/paimon/utils/TagManager.java
+++ b/paimon-core/src/main/java/org/apache/paimon/utils/TagManager.java
@@ -23,6 +23,7 @@ import org.apache.paimon.annotation.VisibleForTesting;
 import org.apache.paimon.fs.FileIO;
 import org.apache.paimon.fs.FileStatus;
 import org.apache.paimon.fs.Path;
+import org.apache.paimon.iceberg.IcebergCommitCallback;
 import org.apache.paimon.manifest.ExpireFileEntry;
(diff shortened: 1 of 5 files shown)


paimon-flink/paimon-flink-1.16/src/test/java/org/apache/paimon/flink/iceberg/Flink116IcebergITCase.java
```diff
@@ -36,4 +36,9 @@ public void testCreateTags(String format) throws Exception {
     public void testDeleteTags(String format) throws Exception {
         // Flink 1.16 does not support delete_tag procedure so we skip this test.
     }
+
+    @Override
+    public void testReplaceTags(String format) throws Exception {
+        // Flink 1.17 does not support replace_tag procedure so we skip this test.
```
nickdelnano, line 42:
```suggestion
        // Flink 1.16 does not support replace_tag procedure so we skip this test.
```

nickdelnano: fix flink version number in comment

paimon-core/src/main/java/org/apache/paimon/utils/TagManager.java
```diff
@@ -112,10 +113,17 @@ public void createTag(
...
+            Snapshot snapshot, String tagName, Duration timeRetained, List<TagCallback> callbacks) {
         checkArgument(!StringUtils.isNullOrWhitespaceOnly(tagName), "Tag name shouldn't be blank.");
         checkArgument(tagExists(tagName), "Tag '%s' doesn't exist.", tagName);
-        createOrReplaceTag(snapshot, tagName, timeRetained, null);
+        createOrReplaceTag(
+                snapshot,
+                tagName,
+                timeRetained,
+                callbacks.stream()
+                        .filter(callback -> IcebergCommitCallback.class.isInstance(callback))
```
LsomeYeah: agree with the Copilot~, `callback instanceof IcebergCommitCallback` or `IcebergCommitCallback.class::isInstance` might be better.

PR #5667, opened 2025-05-27
[hotfix]fix: record expire get second bug

jerry-024:
### Purpose

fix: record expire get second bug,   int is used as the unit of expire second type, which will overflow in 2038.
### Tests

RecordLevelExpireTest.testConvertFieldToSecond


JingsongLi (approved):
+1

PR #5668, opened 2025-05-27
Hive catalog configuration to get owner for table creation

cxzl25:


paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
```diff
@@ -1382,12 +1385,25 @@ private Table newHmsTable(
             @Nullable FormatTable.Format provider,
             boolean externalTable) {
         long currentTimeMillis = System.currentTimeMillis();
+
+        CatalogTableOwnerType ownerType = options.get(TABLE_OWNER_TYPE);
```
JingsongLi: We don't need introduce an option, just:
```
public static String currentUser() {
    String username = null;
    try {
      username = UserGroupInformation.getCurrentUser().getShortUserName();
    } catch (IOException e) {
      LOG.warn("Failed to get Hadoop user", e);
    }

    if (username != null) {
      return username;
    } else {
      LOG.warn("Hadoop user is null, defaulting to user.name");
      return System.getProperty("user.name");
    }
  }
```

PR #5669, opened 2025-05-27
[Bug][spark] Fix the bucket join may produce wrong result after bucke…

### LINKED ISSUE #5654: [Bug] The bucket join may not right after bucket rescaled
### Search before asking

### Paimon version

1.2

### Compute Engine

Spark

### Minimal reproduce step

Currently, Spark bucket join use the table properties bucket number to report the output partition. But, this information can be not right for the partitioned table. 
Because, different partition can have different bucket number, this could lead to the wrong results for the bucket join.

.

### Anything else?

_No response_

Aitozi:
…t rescaled

### Purpose

Linked issue: close #5654

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonScan.scala
+++ b/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonScan.scala
@@ -79,6 +79,9 @@ case class PaimonScan(
     }
   }
 
+  /**
+   * Extract the bucket number from the splits only if all splits have the same totalBuckets number.
+   */
   private def extractBucketNumber(): Option[Int] = {
     val splits = getOriginSplits
     if (splits.exists(!_.isInstanceOf[DataSplit])) {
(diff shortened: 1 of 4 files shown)


paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonScan.scala
```diff
@@ -62,18 +62,38 @@ case class PaimonScan(
           // so we only support one bucket key case.
           assert(bucketSpec.getNumBuckets > 0)
           assert(bucketSpec.getBucketKeys.size() == 1)
-          val bucketKey = bucketSpec.getBucketKeys.get(0)
-          if (requiredSchema.exists(f => conf.resolver(f.name, bucketKey))) {
-            Some(Expressions.bucket(bucketSpec.getNumBuckets, bucketKey))
-          } else {
-            None
+          extractBucketNumber() match {
```
JingsongLi, line 65: Add some comments to explain this extraction.

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonScan.scala
```diff
@@ -62,18 +62,38 @@ case class PaimonScan(
...
+            case _ => None
           }
         }
 
       case _ => None
     }
   }
 
+  private def extractBucketNumber(): Option[Int] = {
+    val splits = getOriginSplits
```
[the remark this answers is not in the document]
Aitozi: the lazyInputPartitions depends on the `bucket` info, so if we use the `lazyInputPartitions`, we will meet StackOverflow here.

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonScan.scala
Aitozi: fixed

PR #5670, opened 2025-05-28
[iceberg] support deletion vector in Iceberg compatibility

LsomeYeah:
### Purpose

Linked issue: close #xxx

Iceberg has supported deletion vector in format-version V3, so we can enable Iceberg to recognize Paimon's deletion vectors.

The implement includes two cases:

1. **create iceberg metadata without base.** In this case, we will get the information of deletion vectors from `deletionFiles` in `DataSplit`.  The advantage is that the corresponding deletion vector can be obtained simultaneously with the data file, without having to read the DV index file again. The downside is that the DeletionFile in Paimon does not record the size of the original DV index file, so the file size in the constructed Iceberg DeletionFile can only be set to the default value of -1.
2. **create iceberg metadata with base.** In this case, we add the newly created indexFiles of the new snapshot in the CommitCallBack, if the indexFiles is not empty, we'll read the index manifest file of the new snapshot to get all deletion vectors, and reconstruct Iceberg DeletionFile.

There are some limitations if you want to recognize Paimon's deletion vectors in Iceberg:

- 'deletion-vectors.enabled' and 'deletion-vectors.bitmap64' must be true, because only 64-bit bitmap implementation of deletion vector in Paimon is compatible with Iceberg.
- 'metadata.iceberg.format-version' must be set to 3, because Iceberg only supports deletion vector in V3.

The format version of iceberg can be upgraded from 2 to 3, and if the version has changed, it will recreate the metadata without using the base metadata. 

Since deletion vector in iceberg is supported in iceberg-1.8.0 and iceberg has stopped supporting JDK 8 since version 1.7.0, we create a new module paimon-iceberg. This module only supports being compiled and tested in JDK 11.

### Tests

org.apache.paimon.core.IcebergDVCompatibilityTest
org.apache.paimon.flink.FlinkIcebergITCase


LsomeYeah:
I want to trigger jdk11 tests. /jdk11

JingsongLi (commented):
Please document this feature.

LsomeYeah:
> Please document this feature.

The document has been added~

paimon-core/src/main/java/org/apache/paimon/iceberg/IcebergCommitCallback.java
```diff
@@ -217,13 +245,19 @@ public void retry(ManifestCommittable committable) {
                                                         + " and identifier "
                                                         + committable.identifier()
                                                         + ". This is unexpected."));
+        Snapshot snapshot = snapshotManager.snapshot(snapshotId);
```
JingsongLi, line 248: You can just get `Snapshot` from `SnapshotManager` above.

paimon-core/src/main/java/org/apache/paimon/iceberg/IcebergCommitCallback.java
```diff
@@ -532,6 +677,9 @@ private boolean shouldAddFileToIceberg(DataFileMeta meta) {
         if (table.primaryKeys().isEmpty()) {
             return true;
         } else {
+            if (needAddDvToIceberg()) {
+                return meta.level() != 0;
```
JingsongLi, line 681: `meta.level() > 0`

paimon-core/src/main/java/org/apache/paimon/iceberg/IcebergCommitCallback.java
```diff
@@ -935,10 +1085,99 @@ public void notifyDeletion(String tagName) {
         }
     }
 
+    // -------------------------------------------------------------------------------------
+    // Deletion vectors
+    // -------------------------------------------------------------------------------------
+
+    private boolean needAddDvToIceberg() {
```
JingsongLi, line 1092: Can you extract a field: `boolean needAddDvToIceberg`?

paimon-core/src/main/java/org/apache/paimon/iceberg/IcebergCommitCallback.java
```diff
@@ -388,28 +513,47 @@ private void createMetadataWithBase(
...
             // new metadata files.
-            newManifestFileMetas = new ArrayList<>(baseManifestFileMetas);
-            newManifestFileMetas.addAll(createNewlyAddedManifestFileMetas(addedFiles, snapshotId));
+            newDataManifestFileMetas = new ArrayList<>(baseDataManifestFileMetas);
+            newDataManifestFileMetas.addAll(
+                    createNewlyAddedManifestFileMetas(addedFiles, snapshotId));
             snapshotSummary = IcebergSnapshotSummary.APPEND;
         } else {
             Pair<List<IcebergManifestFileMeta>, IcebergSnapshotSummary> result =
                     createWithDeleteManifestFileMetas(
```
JingsongLi, line 527: Will the old DELETE entries manifests will be deleted here?

LsomeYeah: For data files, delete entries will be merged in compacting phase.

JingsongLi:
Looks good to me!

PR #5673, opened 2025-05-28
[paimon-core] Add support for map value type and array element type evolution

AshishKhatkar:
### Purpose
This PR adds support for map value type and array element type evolution. For ex: ARRAY<INT> -> ARRAY<BIGINT>

Linked issue: close #xxx

As of today there is no support for evolution map value type and array element type. This change adds support for the same. Consider the case as follows:
```

- USE CATALOG paimon;

- CREATE TABLE evolution_test(field1 INT, field2 ARRAY<INT>, PRIMARY KEY (field1) NOT ENFORCED);

- ALTER TABLE evolution_test MODIFY field2 ARRAY<BIGINT>;
# this throws exception : java.lang.IllegalArgumentException: field2 is not a structured type.

- CREATE TABLE evolution_test1(field1 INT, field2 ARRAY<ROW(f1 INT, f2 INT) NOT NULL>);

- ALTER TABLE evolution_test1 MODIFY field2 ARRAY<ROW(f1 INT, f2 INT)>;
# this throws exception : java.lang.ArrayIndexOutOfBoundsException: 2


With the added support both the above commands evolve the schema as expected
```

### Tests
Added new tests in SchemaChangeITCase to verify the change.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/schema/SchemaManager.java
+++ b/paimon-core/src/main/java/org/apache/paimon/schema/SchemaManager.java
@@ -417,25 +417,36 @@ public class SchemaManager implements Serializable {
                         newFields,
                         update.fieldNames(),
                         (field) -> {
-                            DataType targetType = update.newDataType();
+                            // find the dataType at depth and update the type for it
+                            DataType sourceRootType =
+                                    getRootType(field.type(), 1, update.fieldNames().length);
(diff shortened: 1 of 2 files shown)


JingsongLi (approved):
Very thanks to @AshishKhatkar 

Looks good to me!

PR #5676, opened 2025-05-29
[hive] HiveCatalog should not delete dir for drop partition

JingsongLi:
### Purpose

Old snapshots dependent on files.


yuzelin (approved):
+1

PR #5679, opened 2025-05-30
[core][rest] Add tableId to commit snapshot to avoid wrong commit

JingsongLi:
### Purpose

Consider situation:
1. Create a table named "a".
2. Run streaming job to write "a".
3. Drop table.
4. Create a new table named "a".

The streaming job may write wrong snapshot to new table, causing abnormal table status that cannot be restored.

We can add tableId to commit snapshot method, server can check the id should be the correct id.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/CatalogEnvironment.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/CatalogEnvironment.java
@@ -95,7 +95,7 @@ public class CatalogEnvironment implements Serializable {
     public SnapshotCommit snapshotCommit(SnapshotManager snapshotManager) {
         SnapshotCommit.Factory factory;
         if (catalogLoader != null && supportsVersionManagement) {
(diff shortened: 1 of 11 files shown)


paimon-core/src/main/java/org/apache/paimon/table/CatalogEnvironment.java
```diff
@@ -95,7 +95,7 @@ public boolean supportsVersionManagement() {
     public SnapshotCommit snapshotCommit(SnapshotManager snapshotManager) {
         SnapshotCommit.Factory factory;
         if (catalogLoader != null && supportsVersionManagement) {
-            factory = new CatalogSnapshotCommit.Factory(catalogLoader);
+            factory = new CatalogSnapshotCommit.Factory(catalogLoader, uuid);
```
jerry-024, line 98: Add uuid when factory.create?

jerry-024:
+1

PR #5680, opened 2025-05-30
[core] Don't use snapshotManager.latestSnapshot in AbstractFileStoreWrite to prevent flooding catalog

tsreaper:
### Purpose

This PR creates a `latestSnapshotFromFileSystem` method in `SnapshotManager` and use it in `AbstractFileStoreWrite`. Because `latestSnapshot` will query from REST catalog (if the table is created from REST catalog) and we don't want to flood the catalog with high concurrency.

### Tests

### API and Format

No format changes.

### Documentation

No new feature.


JingsongLi:
+1

PR #5684, opened 2025-06-02
[spark] Throw exception when creating partitioned table via CTAS

Zouxxyy:
### Purpose

Currently, because Spark DSv2 does not support writing partitioned tables, the partitioned format table created by ctas is not written to the partition normally, so throw exception for now


JingsongLi (approved):
+1

PR #5686, opened 2025-06-03
[core] refactor: delete snapshot commit factory

jerry-024:
### Purpose

[core] refactor: delete snapshot commit factory


JingsongLi (approved):
+！

Issue #5687, opened 2025-06-03
[Bug] external file can not be all delete when drop table


kaixuanzhang123:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Paimon version

paimon master branch, commit id: 9aafa2a689554645baf6abb

### Compute Engine

flink 1.17

### Minimal reproduce step

1. set external data-file.external-paths: "/tmp/a"
2. write data
3. set external data-file.external-paths: "/tmp/b"
4. write data
5. delete table 
The files in the tmp/a path have not been deleted
 ` public void dropTable(Identifier identifier, boolean ignoreIfNotExists)
            throws TableNotExistException {
        checkNotBranch(identifier, "dropTable");
        checkNotSystemTable(identifier, "dropTable");

        List<Path> externalPaths = new ArrayList<>();
        try {
            Table table = getTable(identifier);
            if (table instanceof FileStoreTable) {
                FileStoreTable fileStoreTable = (FileStoreTable) table;
                externalPaths = fileStoreTable.store().pathFactory().getExternalPaths();
            }
        } catch (TableNotExistException e) {
            if (ignoreIfNotExists) {
                return;
            }
            throw new TableNotExistException(identifier);
        }

        dropTableImpl(identifier, externalPaths);
    }`
externalPaths = fileStoreTable.store().pathFactory().getExternalPaths();  can not get all external path

### What doesn't meet your expectations?

delete all file 

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #5688, opened 2025-06-04
when drop table delete all external file 

kaixuanzhang123:
### Purpose
bug fix : when drop table delete all external file 
Linked issue: close https://github.com/apache/paimon/issues/5687


paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
```diff
@@ -326,7 +328,18 @@ public void dropTable(Identifier identifier, boolean ignoreIfNotExists)
             Table table = getTable(identifier);
             if (table instanceof FileStoreTable) {
                 FileStoreTable fileStoreTable = (FileStoreTable) table;
-                externalPaths = fileStoreTable.store().pathFactory().getExternalPaths();
+                externalPaths =
+                        fileStoreTable.schemaManager().listAll().stream()
```
JingsongLi, line 332: This is very strange... I don't think it is a good way to clean all files.

PR #5691, opened 2025-06-04
[hive] Update columns when committing Iceberg metadata

### LINKED ISSUE #5690: [Bug] Hive columns are not updated
### Search before asking

### Paimon version

paimon-1.1

### Compute Engine

flink-1.20

### Minimal reproduce step

Update the table and add a new nullable column.

We use AWS Glue and Athena to query the data in a table with Iceberg-compatibility layer enabled. After adding a nullable column, the column is added to the Paimon and Iceberg metadata, and the updated metadata file is referenced by AWS Glue, however, the Glue columns are not updated.

### Anything else?

_No response_

siadat:
### Purpose

Linked issue: close #5690

Previously, updating columns did not mirror the logic applied in createTable. This could lead to inconsistent table schemas. This PR refactors the update process to use the [same mechanism as table creation](https://github.com/apache/paimon/blob/e6c913216ce19a42f030bf463a9010454435df05/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/iceberg/IcebergHiveMetadataCommitter.java#L210-L213).

### Tests

Verified that the new column appears in AWS Glue.

### API and Format

N/A

### Documentation

This PR matches the expected/assumed behavior. If explicit documentation is needed for this update, please advise and I can add/expand docs accordingly.

### DIFF
--- a/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/iceberg/IcebergHiveMetadataCommitter.java
+++ b/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/iceberg/IcebergHiveMetadataCommitter.java
@@ -147,6 +147,12 @@ public class IcebergHiveMetadataCommitter implements IcebergMetadataCommitter {
                     .put("previous_metadata_location", baseMetadataPath.toString());
         }
 
+        StorageDescriptor sd = hiveTable.getSd();
+        sd.setCols(
+                table.schema().fields().stream()
+                        .map(this::convertToFieldSchema)
+                        .collect(Collectors.toList()));
+
         Options options = new Options(table.options());
         boolean skipAWSGlueArchive = options.get(IcebergOptions.GLUE_SKIP_ARCHIVE);
         EnvironmentContext environmentContext = new EnvironmentContext();


paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/iceberg/IcebergHiveMetadataCommitter.java
```diff
@@ -147,6 +147,12 @@ private void commitMetadataImpl(Path newMetadataPath, @Nullable Path baseMetadat
                     .put("previous_metadata_location", baseMetadataPath.toString());
         }
 
+        StorageDescriptor sd = hiveTable.getSd();
+        sd.setCols(
+                table.schema().fields().stream()
```
siadat, line 152: We think we might have to get the latest refreshed schema to avoid overwriting Glue with stale schema fields, using [latest()](https://github.com/apache/paimon/blob/a1df942b454871d18a3836e3eb0dc038bbcc6cfb/paimon-core/src/main/java/org/apache/paimon/schema/SchemaManager.java#L112) instead of FileStoreTable `table.schema()`

siadat: The new column was added to the table when I verified with AWS Glue, so the current implementation seems to be ok.

JingsongLi (approved):
+1

PR #5692, opened 2025-06-04
[core, api] Add option to disable explicit type casting

AshishKhatkar:
### Purpose
This PR disables explicit type casting by default.

Linked issue: close #xxx

As of today, for example BIGINT type can be converted to INT which is not desirable as this conversion leads to overflow. Consider the case as follows:

```

- USE CATALOG paimon;

- CREATE TABLE type_cast_test(field1 INT, field2 INT, PRIMARY KEY (field1) NOT ENFORCED);
- ALTER TABLE type_cast_test MODIFY field1 TINYINT;
- DESCRIBE type_cast_test;
+--------+---------+------+-----+--------+-----------+
|   name |    type | null | key | extras | watermark |
+--------+---------+------+-----+--------+-----------+
| field1 | TINYINT | TRUE |     |        |           |
| field2 |     INT | TRUE |     |        |           |
+--------+---------+------+-----+--------+-----------+

After the change I did

- ALTER TABLE type_cast_test MODIFY field1 TINYINT;
This throws exception :
java.lang.IllegalStateException: Column type field1[INT] cannot be converted to TINYINT without loosing information.
```

### Tests

Added new tests to verify the change for complex types (Row, Array and Map) as well.

### DIFF
--- a/docs/content/primary-key-table/merge-engine/partial-update.md
+++ b/docs/content/primary-key-table/merge-engine/partial-update.md
@@ -130,7 +130,7 @@ CREATE TABLE SG
 INSERT INTO SG
 VALUES (1, 1, 1, 1, 1, 1, 1, 1);
 
--- g_2, g_3 should not be updated
+-- g_3 is null, g_2, g_3 are not bigger, c, d should not be updated
 INSERT INTO SG
 VALUES (1, 2, 2, 2, 2, 2, 1, CAST(NULL AS INT));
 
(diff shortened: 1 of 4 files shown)


AshishKhatkar:
The failing test passes for me locally. It seems to be transient failure.

JingsongLi:
Hi @AshishKhatkar , previously, for the sake of flexibility, all casts were opened up.

Maybe we can just introduce an option to limit explicit casting?

AshishKhatkar:
@JingsongLi I have updated the default to false.

JingsongLi (approved):
+1

PR #5695, opened 2025-06-05
[core]fix: rest catalog options uri schema could be empty

jerry-024:
### Purpose

Fix: when users use the  rest catalog URI, it must be with schema

### Tests
- HttpClientTest.testServerUriSchema

### DIFF
--- a/paimon-api/src/main/java/org/apache/paimon/rest/HttpClient.java
+++ b/paimon-api/src/main/java/org/apache/paimon/rest/HttpClient.java
@@ -195,6 +195,11 @@ public class HttpClient implements RESTClient {
         return fullPath;
     }
 
+    @VisibleForTesting
+    public String uri() {
+        return uri;
+    }
+
     private <T extends RESTResponse> T exec(Request request, Class<T> responseType) {
         try (Response response = HTTP_CLIENT.newCall(request).execute()) {
             String responseBodyStr = response.body() != null ? response.body().string() : null;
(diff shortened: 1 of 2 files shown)


paimon-api/src/main/java/org/apache/paimon/rest/HttpClient.java
```diff
@@ -173,6 +182,9 @@ public <T extends RESTResponse> T delete(
     @VisibleForTesting
     protected static String getRequestUrl(
             String uri, String path, Map<String, String> queryParams) {
+        if (!uri.startsWith("http://") && !uri.startsWith("https://")) {
```
JingsongLi: We don't need this check.

paimon-api/src/main/java/org/apache/paimon/rest/HttpClient.java
```diff
@@ -68,11 +68,20 @@ public class HttpClient implements RESTClient {
...
+                serverUri = uri.substring(0, uri.length() - 1);
+            } else {
+                serverUri = uri;
+            }
+            if (!uri.startsWith("http://") && !uri.startsWith("https://")) {
+                serverUri = String.format("http://%s", serverUri);
+            }
         } else {
-            this.uri = uri;
+            throw new IllegalArgumentException("uri is empty which must be define");
```
JingsongLi: be defined?

paimon-api/src/main/java/org/apache/paimon/rest/HttpClient.java
```diff
@@ -171,8 +180,7 @@ public <T extends RESTResponse> T delete(
     }
 
     @VisibleForTesting
-    protected static String getRequestUrl(
-            String uri, String path, Map<String, String> queryParams) {
+    protected String getRequestUrl(String path, Map<String, String> queryParams) {
```
jerry-024, line 183: As we will check url  when init client. So here just use client's uri.

JingsongLi (approved):
+1

PR #5698, opened 2025-06-05
[flink] Writer operator support SlotSharingGroup settings

yuzelin:
### Purpose

This is a convenient way for users to set writer cpu and memory.


JingsongLi (approved):
+1

PR #5702, opened 2025-06-06
[core] fix: add function name validate

jerry-024:
### Purpose

Function name is allowed to contain characters such as letters (a-zA-Z), numbers (0-9), dots (.), hyphens (-), underscores (_), and must contain letters.
### Tests
- RESTCatalogTest.testValidateFunctionName
- RESTCatalogTest.testFunction

### DIFF
--- a/paimon-api/src/main/java/org/apache/paimon/rest/RESTApi.java
+++ b/paimon-api/src/main/java/org/apache/paimon/rest/RESTApi.java
@@ -24,6 +24,7 @@ import org.apache.paimon.annotation.Public;
 import org.apache.paimon.annotation.VisibleForTesting;
 import org.apache.paimon.catalog.Identifier;
 import org.apache.paimon.function.FunctionChange;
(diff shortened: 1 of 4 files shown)


paimon-api/src/main/java/org/apache/paimon/rest/RESTApi.java
```diff
@@ -851,6 +852,7 @@ public List<String> listFunctions(String databaseName) {
      * @throws ForbiddenException if the user lacks permission to access the function
      */
     public GetFunctionResponse getFunction(Identifier identifier) {
+        FunctionNameValidator.check(identifier.getObjectName());
```
JingsongLi: if (!isValidName(name)) {
   throw new NoSuchResourceException
}

jerry-024: If using NoSuchResourceException, this exception is not reasonable when creating a function. So here, IllegalArgumentException may be better. We could give different in function's api in catalog.

PR #5703, opened 2025-06-06
[core] Introduce properties field in Snapshot

### LINKED ISSUE #5649: [Feature] Paimon's snapshot should support store summary property
### Search before asking

### Motivation

While integrating Paimon with Fluss, I want to store some summary property to record the current Fluss tiering status when Fluss commit a snapshot . But Paimon haven't supported it, I think we can introduce a summary property into snapshot just like iceberg does.

### Solution

_No response_

### Anything else?

_No response_

luoyuxia:
### Purpose

Linked issue: close #5649 

Introduce properties field in Snapshot

### Tests
testCommitManifestWithProperties


paimon-api/src/main/java/org/apache/paimon/Snapshot.java
```diff
@@ -214,7 +222,8 @@ public Snapshot(
             @Nullable Long deltaRecordCount,
             @Nullable Long changelogRecordCount,
             @Nullable Long watermark,
-            @Nullable String statistics) {
+            @Nullable String statistics,
+            Map<String, String> properties) {
```
yuzelin: @Nullable

paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
```diff
@@ -1175,7 +1199,8 @@ private ManifestCompactResult compactManifest(@Nullable ManifestCompactResult la
                         0L,
                         0L,
                         latestSnapshot.watermark(),
-                        latestSnapshot.statistics());
+                        latestSnapshot.statistics(),
+                        Collections.emptyMap());
```
yuzelin: Here should be latestSnapshot.properties()

paimon-api/src/main/java/org/apache/paimon/Snapshot.java
```diff
@@ -195,6 +196,13 @@ public class Snapshot implements Serializable {
     @Nullable
     protected final String statistics;
 
+    // properties
+    // null for paimon <= 1.1
```
JingsongLi: What about empty? We don't want to include empty map.

paimon-core/src/main/java/org/apache/paimon/manifest/ManifestCommittable.java
```diff
@@ -72,6 +84,10 @@ public void addLogOffset(int bucket, long offset, boolean allowDuplicate) {
         logOffsets.put(bucket, newOffset);
     }
 
+    public void addProperty(String key, String value) {
```
JingsongLi, line 87: Can you show the example? How to add property? Add what property.

paimon-api/src/main/java/org/apache/paimon/Snapshot.java
luoyuxia: Now, I modified it to be null when empty map

paimon-core/src/main/java/org/apache/paimon/manifest/ManifestCommittable.java
luoyuxia: Maybe something like:
```
ManifestCommittable committable = new ManifestCommittable(COMMIT_IDENTIFIER);
committable. addProperty("fluss_offset_partition_1_bucket1", "14")
committable. addProperty("fluss_offset_partition_2_bucket2", "15")
```

paimon-core/src/test/java/org/apache/paimon/manifest/ManifestCommittableSerializerCompatibilityTest.java
```diff
@@ -104,7 +104,7 @@ public void testCompatibilityToV3CommitV7() throws IOException {
 
         ManifestCommittableSerializer serializer = new ManifestCommittableSerializer();
```
JingsongLi, line 179: Add a new test for current version.

paimon-core/src/test/java/org/apache/paimon/manifest/ManifestCommittableSerializerCompatibilityTest.java
```diff
@@ -463,7 +463,7 @@ public void testCompatibilityToV2CommitV2() throws IOException {
...
         byte[] bytes = serializer.serialize(manifestCommittable);
-        ManifestCommittable deserialized = serializer.deserialize(3, bytes);
+        ManifestCommittable deserialized = serializer.deserialize(serializer.getVersion(), bytes);
```
JingsongLi, line 540: Please don't change this.

paimon-core/src/test/java/org/apache/paimon/manifest/ManifestCommittableSerializerCompatibilityTest.java
luoyuxia: done

paimon-core/src/test/java/org/apache/paimon/manifest/ManifestCommittableSerializerCompatibilityTest.java
luoyuxia: After look in deep again, maybe we still need to change to use `serializer.getVersion()` which is v4 to deserialize it. 

According to java doc, the version is for in which the data was serialized in here, the data is serialized with v4.
And in v4, we will serialize `properties` into the bytes, it deserialize with v3, it'll skip the `properties` bytes and try to deserialize it as `commitMessages` and throw `BufferUnderflowException`.

JingsongLi (approved):
+1

PR #5705, opened 2025-06-06
[spark] Extract postCommit for V2 write

Zouxxyy:
### Purpose

extract postCommit(messages: Seq[CommitMessage]) for v1 or v2 write


paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/write/WriteHelper.scala
```diff
@@ -0,0 +1,114 @@
...
+  val table: FileStoreTable
+
+  lazy val coreOptions: CoreOptions = table.coreOptions()
+
+  def postCommit(messages: Seq[CommitMessage]): Unit = {
+    if (messages.isEmpty) {
+      return
+    }
+
+    reportToHms(messages)
```
YannByron, line 44: why have to do these in paimon-spark, rather than uniformly in paimon-core ?

YannByron:
Can you explain what this pr focuses on? @Zouxxyy

Issue #5697, opened 2025-06-05
[Feature] [spark] Bump Spark 4.0.0


YannByron:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Motivation

upgrade spark 4.x version from preview to release version

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #5711, opened 2025-06-06
[spark] Upgrade spark version to 4.0.0

YannByron:
### Purpose

https://github.com/apache/paimon/issues/5697


Zouxxyy:
Unfortunately, it seems that under the current architecture, we can no longer use `org.apache.spark.sql.classic.ClassicConversions`

YannByron:
> Unfortunately, it seems that under the current architecture, we can no longer use `org.apache.spark.sql.classic.ClassicConversions`

I can still have the new class `Classic4Api` inherit `org.apache.spark.sql.classic.ClassicConversions` in spark4-common. The advantage is that all the classic conversion-related can be encapsulated here, which is clear and definite.

paimon-spark/paimon-spark-4.0/pom.xml
```diff
@@ -81,6 +81,12 @@ under the License.
             <version>${spark.version}</version>
             <classifier>tests</classifier>
             <scope>test</scope>
+            <exclusions>
```
Zouxxyy, line 84: Why should we exclude this here

YannByron: to resolve conflicts. This is similar to the behavior of spark 4.x.

PR #5712, opened 2025-06-07
[core] Simplify rollback to only clean snapshots

JingsongLi:
### Purpose

At present, rollback is a relatively dangerous operation. It not only deletes unnecessary snapshots, but also cleans up all unnecessary data files, which makes it very prone to problems. Once there is a bug, it will cause the table to malfunction.

Actually, rollback is a very low-frequency operation, and we don't need to clean up all the unnecessary data files. Just leave it to orphan file clean.


leaves12138 (approved):
+1

PR #5714, opened 2025-06-08
[hotfix] Fix fallback branch cannot query data in rest catalog

qingwei727:
Purpose
Fix the bug that we query data nothing after we insert data into fallback branch using rest catalog. 

Tests
CatalogTableITCase.testReadOptimizedTableFallBack using rest catalog.


JingsongLi (approved):
+1

PR #5715, opened 2025-06-08
[Flink] Add batch read operator source.

wwj6591812:
### Purpose

In https://github.com/apache/paimon/pull/5650 we resolve the JM memory problem by support withShard read of Flink.
But that method has a drawback that each task will read a lot of the same Meta.
So, this pr resolve the problem by move the plan from JM to a single task, then this task assign the splits to the downstream readers.

Linked issue: close #xxx

### DIFF
--- a/docs/content/append-table/query-performance.md
+++ b/docs/content/append-table/query-performance.md
@@ -75,3 +75,14 @@ we use the procedure, you should config appropriate configurations in target tab
 `file-index.<filter-type>.columns` to the table.
 
 How to invoke: see [flink procedures]({{< ref "flink/procedures#procedures" >}}) 
+
+## Dedicated Split Generation
+When Paimon table snapshots contain large amount of source splits, Flink jobs reading from this table might endure long initialization time or even OOM in JobManagers. In this case, you can configure `'scan.dedicated-split-generation' = 'true'` to avoid such problem. This option would enable executing the source split generation process in a dedicated subtask that runs on TaskManager, instead of in the source coordinator on the JobManager.
+
+Note that this feature could have some side effects on your Flink jobs. For example:
+
+1. It will change the DAG of the flink job, thus breaking checkpoint compatibility if enabled on an existing job.
+2. It may lead to the Flink AdaptiveBatchScheduler inferring a small parallelism for the source reader operator. you can configure `scan.infer-parallelism` to avoid this possible drawback.
+3. The failover strategy of the Flink job would be forced into global failover instead of regional failover, given that the dedicated source split generation task would be connected to all downstream subtasks.
+
+So please make sure these side effects are acceptable to you before enabling it.
--- a/docs/content/primary-key-table/query-performance.md
+++ b/docs/content/primary-key-table/query-performance.md
@@ -73,3 +73,15 @@ we use the procedure, you should config appropriate configurations in target tab
 `file-index.<filter-type>.columns` to the table.
 
 How to invoke: see [flink procedures]({{< ref "flink/procedures#procedures" >}}) 
+
+## Dedicated Split Generation
+When Paimon table snapshots contain large amount of source splits, Flink jobs reading from this table might endure long initialization time or even OOM in JobManagers. In this case, you can configure `'scan.dedicated-split-generation' = 'true'` to avoid such problem. This option would enable executing the source split generation process in a dedicated subtask that runs on TaskManager, instead of in the source coordinator on the JobManager.
+
+Note that this feature could have some side effects on your Flink jobs. For example:
+
+1. It will change the DAG of the flink job, thus breaking checkpoint compatibility if enabled on an existing job.
+2. It may lead to the Flink AdaptiveBatchScheduler inferring a small parallelism for the source reader operator. you can configure `scan.infer-parallelism` to avoid this possible drawback.
+3. The failover strategy of the Flink job would be forced into global failover instead of regional failover, given that the dedicated source split generation task would be connected to all downstream subtasks.
+
+So please make sure these side effects are acceptable to you before enabling it.
+
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/FlinkSourceBuilder.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/FlinkSourceBuilder.java
@@ -294,6 +294,9 @@ public class FlinkSourceBuilder {
         }
 
         if (sourceBounded) {
+            if (conf.get(FlinkConnectorOptions.SCAN_DEDICATED_SPLIT_GENERATION)) {
+                return buildContinuousStreamOperator(true);
+            }
             return buildStaticFileSource();
         }
         TableScanUtils.streamingReadingValidate(table);
@@ -325,16 +328,16 @@ public class FlinkSourceBuilder {
             } else if (conf.contains(CoreOptions.CONSUMER_ID)
                     && conf.get(CoreOptions.CONSUMER_CONSISTENCY_MODE)
                             == CoreOptions.ConsumerMode.EXACTLY_ONCE) {
-                return buildContinuousStreamOperator();
+                return buildContinuousStreamOperator(false);
             } else {
                 return buildContinuousFileSource();
             }
         }
     }
 
-    private DataStream<RowData> buildContinuousStreamOperator() {
+    private DataStream<RowData> buildContinuousStreamOperator(boolean isBounded) {
         DataStream<RowData> dataStream;
-        if (limit != null) {
+        if (limit != null && !isBounded) {
             throw new IllegalArgumentException(
                     "Cannot limit streaming source, please use batch execution mode.");
         }
@@ -346,10 +349,11 @@ public class FlinkSourceBuilder {
                         createReadBuilder(projectedRowType()),
                         conf.get(CoreOptions.CONTINUOUS_DISCOVERY_INTERVAL).toMillis(),
                         watermarkStrategy == null,
-                        conf.get(
-                                FlinkConnectorOptions.STREAMING_READ_SHUFFLE_BUCKET_WITH_PARTITION),
+                        conf.get(FlinkConnectorOptions.READ_SHUFFLE_BUCKET_WITH_PARTITION),
                         unawareBucket,
-                        outerProject());
+                        outerProject(),
+                        isBounded,
+                        limit);
         if (parallelism != null) {
             dataStream.getTransformation().setParallelism(parallelism);
         }
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/operator/ReadOperator.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/operator/ReadOperator.java
@@ -36,6 +36,8 @@ import org.apache.flink.streaming.api.operators.AbstractStreamOperator;
 import org.apache.flink.streaming.api.operators.OneInputStreamOperator;
 import org.apache.flink.streaming.runtime.streamrecord.StreamRecord;
 import org.apache.flink.table.data.RowData;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nullable;
 
@@ -46,6 +48,8 @@ import javax.annotation.Nullable;
  */
 public class ReadOperator extends AbstractStreamOperator<RowData>
         implements OneInputStreamOperator<Split, RowData> {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ReadOperator.class);
 
     private static final long serialVersionUID = 2L;
 
@@ -64,12 +68,15 @@ public class ReadOperator extends AbstractStreamOperator<RowData>
     private transient long emitEventTimeLag = FileStoreSourceReaderMetrics.UNDEFINED;
     private transient long idleStartTime = FileStoreSourceReaderMetrics.ACTIVE;
     private transient Counter numRecordsIn;
+    @Nullable private final Long limit;
 
     public ReadOperator(
             SerializableSupplier<TableRead> readSupplier,
-            @Nullable NestedProjectedRowData nestedProjectedRowData) {
+            @Nullable NestedProjectedRowData nestedProjectedRowData,
+            @Nullable Long limit) {
         this.readSupplier = readSupplier;
         this.nestedProjectedRowData = nestedProjectedRowData;
+        this.limit = limit;
     }
 
     @Override
@@ -98,6 +105,10 @@ public class ReadOperator extends AbstractStreamOperator<RowData>
 
     @Override
     public void processElement(StreamRecord<Split> record) throws Exception {
+        if (reachLimit()) {
+            return;
+        }
+
         Split split = record.getValue();
         // update metric when reading a new split
         long eventTime =
@@ -122,6 +133,10 @@ public class ReadOperator extends AbstractStreamOperator<RowData>
                     numRecordsIn.inc();
                 }
 
+                if (reachLimit()) {
+                    return;
+                }
+
                 reuseRow.replace(iterator.next());
                 if (nestedProjectedRowData == null) {
                     reuseRecord.replace(reuseRow);
@@ -142,6 +157,14 @@ public class ReadOperator extends AbstractStreamOperator<RowData>
         if (ioManager != null) {
             ioManager.close();
         }
+    }
+
+    private boolean reachLimit() {
+        if (limit != null && numRecordsIn.getCount() > limit) {
+            LOG.info("Reader {} reach the limit record {}.", this, limit);
+            return true;
+        }
+        return false;
     }
 
     private void idlingStarted() {
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/utils/TableScanUtils.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/utils/TableScanUtils.java
@@ -30,12 +30,14 @@ import java.util.HashSet;
 import java.util.Optional;
 import java.util.Set;
 
+import static org.apache.paimon.flink.FlinkConnectorOptions.SCAN_DEDICATED_SPLIT_GENERATION;
+
 /** Utility methods for {@link TableScan}, such as validating. */
 public class TableScanUtils {
 
     public static void streamingReadingValidate(Table table) {
-        CoreOptions options = CoreOptions.fromMap(table.options());
-        CoreOptions.MergeEngine mergeEngine = options.mergeEngine();
+        CoreOptions coreOptions = CoreOptions.fromMap(table.options());
+        CoreOptions.MergeEngine mergeEngine = coreOptions.mergeEngine();
         HashMap<CoreOptions.MergeEngine, String> mergeEngineDesc =
                 new HashMap<CoreOptions.MergeEngine, String>() {
                     {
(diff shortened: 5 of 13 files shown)


docs/layouts/shortcodes/generated/core_configuration.html
```diff
@@ -228,7 +228,7 @@
...
             <td><p>Enum</p></td>
-            <td>Specify the consumer consistency mode for table.<br /><br />Possible values:<ul><li>"exactly-once": Readers consume data at snapshot granularity, and strictly ensure that the snapshot-id recorded in the consumer is the snapshot-id + 1 that all readers have exactly consumed.</li><li>"at-least-once": Each reader consumes snapshots at a different rate, and the snapshot with the slowest consumption progress among all readers will be recorded in the consumer.</li></ul></td>
+            <td>Specify the consumer consistency mode for table.<br /><br />Possible values:<ul><li>"exactly-once": Readers consume data at snapshot granularity, and strictly ensure that the snapshot-id recorded in the consumer is the snapshot-id + 1 that all readers have exactly consumed.</li><li>"exactly-once-for-batch": Readers consume the latest snapshot data with batch mood, and the plan will run on TM.</li><li>"at-least-once": Each reader consumes snapshots at a different rate, and the snapshot with the slowest consumption progress among all readers will be recorded in the consumer.</li></ul></td>
```
yunfengzhou-hub: consumer.mode is supposed to be used together with consumer.id, and it seems that exactly-once-for-batch does not mean so. Thus how about introducing an individual option that have the following information for users?

1. The option could notify users that this feature is not bound with the consumer mechanism. They can enable this feature so long as they are facing a Flink Paimon source coordinator bottleneck in batch job.
2. The option changes the DAG of the Flink job, thus breaking checkpoint compatibility. So this option is not enabled by default and users have to consider this side effect before enabling it.

wwj6591812:
The failed ut is not caused by my modification. 
Please CC, Thx.
@JingsongLi

docs/layouts/shortcodes/generated/core_configuration.html
wwj6591812: Very thanks, done.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/operator/ReadOperator.java
```diff
@@ -130,6 +137,14 @@ public void processElement(StreamRecord<Split> record) throws Exception {
                     reuseRecord.replace(nestedProjectedRowData);
                 }
                 output.collect(reuseRecord);
+
+                if (limit != null && numRecordsIn.getCount() == limit) {
+                    LOG.info(
+                            "When limit is not null, reader {} reach the limit record {}.",
+                            this,
+                            limit);
+                    return;
```
yunfengzhou-hub: When will this condition be met? limit should have been applied to the ReadBuilder in FlinkSourceBuilder#createReadBuilder, so the limit should not come into effect here.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/operator/BatchReadSource.java
```diff
@@ -0,0 +1,140 @@
...
+                DataSplit dataSplit = (DataSplit) splits.get(0);
+                LOG.info(
+                        "Read the snapshot-{} of table {} in BatchSource.",
+                        dataSplit.snapshotId(),
+                        tableName);
+            }
+
+            splits.forEach(readerOutput::collect);
+
+            return InputStatus.END_OF_INPUT;
```
yunfengzhou-hub: How about reusing MonitorSource instead of introducing a new one? For example,
```java
if (isBounded) {
    return InputStatus.END_OF_INPUT;
} else {
    return InputStatus.MORE_AVAILABLE;
}
```

paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -1744,6 +1744,13 @@ public class CoreOptions implements Serializable {
                             "If true, it disables explicit type casting. For ex: it disables converting LONG type to INT type. "
                                     + "Users can enable this option to disable explicit type casting");
 
+    public static final ConfigOption<Boolean> SCAN_BATCH_PLAN_MOVE_FROM_JM_TO_TM_ENABLED =
+            key("scan.batch.plan.move-from-jm-to-tm.enabled")
+                    .booleanType()
+                    .defaultValue(false)
+                    .withDescription(
+                            "If true, scan plan will run on a single TM, you can enable it when Paimon source coordinator encounter with memory bottleneck in Flink batch job. This feature is not bounded with the consumer mechanism and it will change the DAG of the flink job, thus breaking checkpoint compatibility. So this option is not enabled by default and users have to consider this side effect before enabling it.");
```
yunfengzhou-hub: The current configuration name and description still seems complicated to me. In order to use this configuration from the current description, users have to know Flink and Paimon implementation details such as plan, jm, tm, which influences the readability of this configuration. Is there some better naming and description that could resolve this issue?

Besides, you can use Description.builder() to avoid a long line of description.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/FlinkSourceBuilder.java
```diff
@@ -294,6 +295,9 @@ public DataStream<RowData> build() {
         }
 
         if (sourceBounded) {
+            if (conf.get(CoreOptions.SCAN_BATCH_PLAN_MOVE_FROM_JM_TO_TM_ENABLED)) {
```
yunfengzhou-hub: If users configured this option and their job is in streaming mode, can this optimization still take effect? Would MonitorSource achieve a similar optimization in streaming mode?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/operator/BatchReadSource.java
```diff
@@ -0,0 +1,140 @@
...
+        DataStream<Split> sourceDataStream =
+                unawareBucket
+                        ? shuffleUnawareBucket(singleOutputStreamOperator)
+                        : shuffleNonUnawareBucket(
+                                singleOutputStreamOperator, shuffleBucketWithPartition);
+
+        return sourceDataStream.transform(
+                name + "-Reader",
+                typeInfo,
+                new ReadOperator(readBuilder::newRead, nestedProjectedRowData, limit));
```
yunfengzhou-hub: Unlike streaming mode, in batch mode Flink would enable auto parallelism inference by default. Given that the parallelism of downstream operators would be inferred according to `execution.batch.adaptive.auto-parallelism.avg-data-volume-per-task`, and the plan splits transmitted here are usually of very small size, I'm a little concerned that the parallelism of readOperator will be inferred into a small value, and users have to manually set the parallelism of this operator to the correct range, which obeys the convenient brought by auto parallelism inference. What do you think?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/operator/ReadOperator.java
wwj6591812: FlinkSourceBuilder#buildBatchReadOperator -> BatchReadSource#buildSource -> ReadOperator#ReadOperator

I have test whether add this condition. The result is :

一、with this condition（select * from `paimon-1.2-ali-165091-0608-001`.`tddl_sc_p4p_new_app_na175`.`p4p_cust_property` /*+ OPTIONS('scan.parallelism'='30', 'consumer.mode'='exactly-once-for-batch') */limit 10;）
image

二、without this condition（select * from `paimon-1.2-ali-165091-0608-002`.`tddl_sc_p4p_new_app_na175`.`p4p_cust_property` /*+ OPTIONS('scan.parallelism'='30', 'consumer.mode'='exactly-once-for-batch') */limit 10;
）
image

So, I think the limit should come into effect here.

paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
wwj6591812: Thanks, done.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/FlinkSourceBuilder.java
wwj6591812: Thanks very much, you are right.
1、Add some description in the option's description.
2、Add parameter check in TableScanUtils#streamingReadingValidate.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/operator/BatchReadSource.java
wwj6591812: Thanks very much, you are right.
I have added some description in the option's description.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/operator/BatchReadSource.java
wwj6591812: I don't want to reusing MonitorSource instead of BatchReadSource, because:
1、MonitorSource is a periodic check for new snapshots, which semantically means streaming processing, not batch processing.
2、There is lots of checkpoint states logic in MonitorSource, thus not used in batch mode. And the BatchReadSource's logic is simple. 
3、It will add some if...else... in MonitorSource, for example in MonitorSource#getBoundedness、MonitorSource#Reader#pollNext、MonitorSource#addSplits.

So I don't think reusing MonitorSource is a better choose.

wwj6591812:
@yunfengzhou-hub 
hi, I have addressd, thanks.
please CC

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/operator/ReadOperator.java
yunfengzhou-hub: In this example, the limit works as expected because there is only one source split, and only one subtask among the ReadOperators is working. In cases where the amount of data grows bigger, the actual records sent in this way would be `limit * numOfSourceSplits`. This is unexpected.

Besides, I tried running testBatchReadSourceWithSnapshot after removing this logic from ReadOperator, and the `SELECT ... LIMIT 2` syntax still functioned correctly. This demonstrates that there is another place for limit to work, which is at FlinkSourceBuilder#createReadBuilder.

wwj6591812: @yunfengzhou-hub 
Hi, thanks for your reply.

1、The reason why I add this condition here is :
The reader should end as soon as possible after reading the limit data, it should not read unnecessary data.

2、The limit of FlinkSourceBuilder#createReadBuilder is used in DataTableBatchScan#applyPushDownLimit which try it best to cut splits.

3、In testBatchReadSourceWithSnapshot, remove this logic from ReadOperator, the `SELECT ... LIMIT 2` syntax still functioned correctly, the reason is:
Flink plan will generate a global limit operator, it will limit 2.
image

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/FlinkSourceBuilder.java
```diff
@@ -294,6 +295,9 @@ public DataStream<RowData> build() {
         }
 
         if (sourceBounded) {
+            if (conf.get(CoreOptions.SCAN_GET_SPLITS_ON_SINGLE_TASK)) {
```
yunfengzhou-hub: Can we reuse CoreOptions#scanGetSplitsOnSingleTask here?

paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -1753,6 +1753,27 @@ public class CoreOptions implements Serializable {
                                     + "starting from the snapshot after this one. If found, commit will be aborted. "
                                     + "If the value of this option is -1, committer will not check for its first commit.");
 
+    public static final ConfigOption<Boolean> SCAN_GET_SPLITS_ON_SINGLE_TASK =
```
yunfengzhou-hub: This option should be moved to FlinkConnectorOptions, instead of at CoreOptions.

paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -1753,6 +1753,27 @@ public class CoreOptions implements Serializable {
...
+                                    .text("This feature have some side effects, for example : ")
+                                    .linebreak()
+                                    .text(
+                                            "1. It will change the DAG of the flink job, thus breaking checkpoint compatibility.")
+                                    .linebreak()
+                                    .text(
+                                            "2. It may infer a small parallelism of reader if you enable Flink AdaptiveBatchScheduler. You can set Flink CoreOptions.DEFAULT_PARALLELISM to avoid.")
+                                    .linebreak()
+                                    .text(
+                                            "So it's default value is false and you have to consider the side effects before enable it.")
```
yunfengzhou-hub: How about the following?
* key: scan.dedicated-split-generation
* description: If true, the split generation process would be performed during runtime on a Flink task, instead of on the JobManager during initialization phase.

As for the concerns users need to consider, we can put them in Paimon's document like the Query Performance document. In this way we can simplify the description here to avoid including too much non-critical information.

Besides, it comes to me that there is another concern for users: when this feature is enabled, the original possible regional failover would be changed into global failover now, as we have a BatchSource subtask that connects to all downstream ReadOperators. We can also add this bullet to the list in document.

yunfengzhou-hub (commented):
Thanks for the update! Some comments are as follows.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/FlinkSourceBuilder.java
wwj6591812: moved to FlinkConnectorOptions

paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
wwj6591812: thanks, done

paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
wwj6591812: good idea! done

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/operator/ReadOperator.java
```diff
@@ -130,6 +137,11 @@ public void processElement(StreamRecord<Split> record) throws Exception {
                     reuseRecord.replace(nestedProjectedRowData);
                 }
                 output.collect(reuseRecord);
+
+                if (limit != null && numRecordsIn.getCount() == limit) {
```
yunfengzhou-hub: When the records in the first split have reached this limit, the read operator would still consume at least one record from the next received split, causing `numRecordsIn.getCount() == (limit + 1) != limit`. Thus it is better to move this logic to before iterator.next, and change such condition into `numRecordsIn.getCount() >= limit`.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/utils/TableScanUtils.java
```diff
@@ -45,14 +50,22 @@ public static void streamingReadingValidate(Table table) {
...
                 throw new RuntimeException(
                         mergeEngineDesc.get(mergeEngine)
                                 + " streaming reading is not supported. You can use "
                                 + "'lookup' or 'full-compaction' changelog producer to support streaming reading. "
                                 + "('input' changelog producer is also supported, but only returns input records.)");
             }
         }
+
+        if (options.containsKey(SCAN_DEDICATED_SPLIT_GENERATION.key())
+                && parseBoolean(options.get(SCAN_DEDICATED_SPLIT_GENERATION.key()))) {
```
yunfengzhou-hub: nit: change `CoreOptions coreOptions = CoreOptions.fromMap(table.options());` to `Options options = Options.fromMap(table.options());`, and then we won't need parseBoolean here.

docs/content/primary-key-table/query-performance.md
```diff
@@ -73,3 +73,14 @@ we use the procedure, you should config appropriate configurations in target tab
...
+
+## Optimize Memory By Dedicated Split Generation
+If you facing a Flink Paimon source coordinator memory bottleneck in batch job. You can set this option to true.
+* 'scan.dedicated-split-generation' = 'true'
+
+This feature have some side effects, for example :
+1. It will change the DAG of the flink job, thus breaking checkpoint compatibility.
+2. It may infer a small parallelism of reader if you enable Flink AdaptiveBatchScheduler. You can set Flink CoreOptions.DEFAULT_PARALLELISM to avoid.
+3. It will change the failover strategy from the original possible region failover to global failover. Because a Monitor subtask that connects to all downstream ReadOperators.
+So it's default value is false and you have to consider the side effects before enable it.
```
yunfengzhou-hub: How about the following description?

```markdown
## Dedicated Split Generation

When Paimon table snapshots contain large amount of source splits, Flink jobs reading from this table might endure long initialization time or even OOM in JobManagers. In this case, you can configure `'scan.dedicated-split-generation' = 'true'` to avoid such problem. This option would enable executing the source split generation process in a dedicated subtask that runs on TaskManager, instead of in the source coordinator on the JobManager.

Note that this feature could have some side effects on your Flink jobs. For example:

1. It will change the DAG of the flink job, thus breaking checkpoint compatibility if enabled on an existing job.
2. It may lead to the Flink AdaptiveBatchScheduler inferring a small parallelism for the source reader operator. you can configure `scan.infer-parallelism` to avoid this possible drawback. 
3. The failover strategy of the Flink job would be forced into global failover instead of regional failover, given that the dedicated source split generation task would be connected to all downstream subtasks.

So please make sure these side effects are acceptable to you before enabling it.

```

docs/content/append-table/query-performance.md
```diff
@@ -75,3 +75,6 @@ we use the procedure, you should config appropriate configurations in target tab
 `file-index.<filter-type>.columns` to the table.
 
 How to invoke: see [flink procedures]({{< ref "flink/procedures#procedures" >}}) 
+
+## Optimize Memory By Dedicated Split Generation
+see [query performance]({{< ref "primary-key-table/query-performance#Optimize Memory By Dedicated Split Generation" >}})
```
yunfengzhou-hub: It might be better to directly copy the content from the other document here instead of adding a link.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/utils/TableScanUtils.java
wwj6591812: Thanks.

Change `Map<String, String> options = table.options();` to `Options options = Options.fromMap(table.options());`

Then cut the parseBoolean statement.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/operator/ReadOperator.java
wwj6591812: done, thx.
And add this condition check at the beginning of processElement. This can pass the follow splits if already reach the limit.

docs/content/primary-key-table/query-performance.md
wwj6591812: Very thanks. done

docs/content/append-table/query-performance.md
wwj6591812: thanks
done

wwj6591812:
@yunfengzhou-hub 
Hi, I have addressd, thanks.
please CC

yunfengzhou-hub:
Thanks for the update. LGTM

JingsongLi (approved):
+1 Thanks @wwj6591812 and @yunfengzhou-hub

PR #5716, opened 2025-06-08
[core] Add truncate bucket strategy

### LINKED ISSUE #3538: [Feature] Introduce mod bucket generator to paimon
### Search before asking

### Motivation

Currently, when we define the `bucket-key` in the Paimon table, the bucket is generated using `hash(bucket-key) % bucket_number`. However, there are cases where we have already determined the target bucket at the computing edge.

For instance, let's consider a Paimon table that stores user ID bitmaps. We aim to distribute users into 32 buckets evenly. Therefore, we calculate a `bucket_no` key using `hash(user_id) % 32`. Yet, under the current mechanism, it will be recalculated as `hash(bucket_no) % bucket_number`, potentially causing data skew within buckets due to hash collisions.

So, what about allowing the definition of the bucket generator eg:

```
HASH_MOD
MOD
```

By this, we can directly use `MOD` to distribute the data.

### Solution

_No response_

### Anything else?

_No response_

luoyuxia:
### Purpose

Linked issue: close #3538 

Introduce a [`truncate`](https://iceberg.apache.org/spec/#partition-transforms) strategy to distribute data into bucket instead of always using hash.

### Tests
Add the test in bucket releated code for `truncate` strategy to verify it works.

### DIFF
--- a/docs/layouts/shortcodes/generated/core_configuration.html
+++ b/docs/layouts/shortcodes/generated/core_configuration.html
@@ -60,7 +60,13 @@ under the License.
             <td><h5>bucket-key</h5></td>
             <td style="word-wrap: break-word;">(none)</td>
             <td>String</td>
-            <td>Specify the paimon distribution policy. Data is assigned to each bucket according to the hash value of bucket-key.<br />If you specify multiple fields, delimiter is ','.<br />If not specified, the primary key will be used; if there is no primary key, the full row will be used.</td>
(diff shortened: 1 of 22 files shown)


JingsongLi:
We should do an abstraction between this and #5445.

luoyuxia:
Thanks for that. Haven't been awared of this pr. Will have a syn with the author

luoyuxia:
After  discussion with @Aitozi , I'll rebase after #5445  is merged. Thanks @Aitozi for the work.

PR #5718, opened 2025-06-09
[core] Optimize DataFileMeta.keyStats to EMPTY_STATS to reduce memory usage

JingsongLi:
### Purpose

Reduce 300 bytes in memory per `DataFileMeta`.


LsomeYeah (approved):
+1

PR #5725, opened 2025-06-10
[core] Total buckets should remain the same within the same partition when commit

tsreaper:
### Purpose

This PR adds a check before commit, that total buckets should remain the same within the same partition.

This check is to avoid the following problem: If the user runs two compact actions with different bucket numbers on a postpone bucket table, there will be files with different total buckets in the same partition, and the same primary key may appear in two buckets.

### Tests

* `TableCommitTest#testGiveUpCommitWhenTotalBucketsChanged`

### API and Format

No format changes.

### Documentation

No new feature.


paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
```diff
@@ -1252,10 +1257,47 @@ private List<SimpleFileEntry> readAllEntriesFromChangedPartitions(
     private void noConflictsOrFail(
             String baseCommitUser,
             List<SimpleFileEntry> baseEntries,
-            List<SimpleFileEntry> changes) {
+            List<SimpleFileEntry> changes,
+            Snapshot.CommitKind commitKind) {
         List<SimpleFileEntry> allEntries = new ArrayList<>(baseEntries);
         allEntries.addAll(changes);
 
+        if (commitKind != Snapshot.CommitKind.OVERWRITE) {
```
JingsongLi, line 1265: Why we need to exclude overwrite? Overwrite has no old files.

PR #5727, opened 2025-06-10
[core] Introduce commit strict mode to avoid data loss in rescale job

tsreaper:
### Purpose

Consider the following scenario: For a postpone bucket table, when a rescale job is running, another compact job commits. However, as the rescale job is started without knowing the COMMIT snapshot, the resulting snapshot will not include its changes and the data is lost.

This PR adds a "strict mode" for commit, where commit is aborted if committer found another user's COMPACT / OVERWRITE snapshot.

### Tests

* `TableCommitTest#testStrictMode`.

### API and Format

No format changes.

### Documentation

Document is updated.

### DIFF
--- a/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
@@ -1744,6 +1744,24 @@ public class CoreOptions implements Serializable {
                             "If true, it disables explicit type casting. For ex: it disables converting LONG type to INT type. "
                                     + "Users can enable this option to disable explicit type casting");
 
+    public static final ConfigOption<Boolean> COMMIT_STRICT_MODE =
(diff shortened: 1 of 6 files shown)


paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -1744,6 +1744,24 @@ public class CoreOptions implements Serializable {
                             "If true, it disables explicit type casting. For ex: it disables converting LONG type to INT type. "
                                     + "Users can enable this option to disable explicit type casting");
 
+    public static final ConfigOption<Boolean> COMMIT_STRICT_MODE =
```
JingsongLi: We don't need this option. Just use `COMMIT_STRICT_MODE_LAST_SAFE_SNAPSHOT`, with this option, we should do strict check.

PR #5729, opened 2025-06-10
[hotfix] TagPreview utils support DataTable

yuzelin:
### Purpose

AuditLogTable (DataTable) also supports reading tags.


jerry-024:
+1

PR #5734, opened 2025-06-11
Support user agent for JindoFileIO

timmyyao:
In RESTCatalog, header.user-agent is used to configure user agent for meta reat api. JindoFileIO also supports setting user agent for sending OSS requests. So we can use header.user-agent to configure both meta rest api and OSS requests so that access log in both sides can be easily distinguished by the same user agent tag.


JingsongLi:
I think we should just use `fs.oss.user.agent.extended`. No need to modify jindo now.

PR #5736, opened 2025-06-12
[core] Also remove metadata.stats-mode=none in postpone bucket writers when not using avro

tsreaper:
### Purpose

Like #5733, this PR also remove `metadata.stats-mode=none` in postpone bucket writers when not using avro. Because collecting stats is costly only avro, but simple for other formats.

### Tests

This is a simple change without additional tests.

### API and Format

No format changes.

### Documentation

No new feature.


JingsongLi:
+1

PR #5737, opened 2025-06-12
[core] Limit max task number in ManifestReadThreadPool

JingsongLi:
### Purpose

Introduce `SemaphoredDelegatingExecutor` to limit max task number in `ManifestReadThreadPool`.

This solves #5297


askwang:
+1

PR #5739, opened 2025-06-12
[flink] Fix paimon-flink-2.0 releasing process

yunfengzhou-hub:
### Purpose

This PR fixes the bug that `update_branch_version.sh` does not work for paimon-flink1-common and paimon-flink2-common.

* Extracts paimon-flinkx-common from paimon-flink-common's jar

### Tests

Manually verified that `update_branch_version.sh` could now work on all pom.xml.


yuzelin (approved):
+1

PR #5740, opened 2025-06-12
[core] ParquetInputStream support vectored read

zhuanshenbsj1:
### Purpose

Linked issue: close #xxx


zhuanshenbsj1:
https://github.com/apache/paimon/blob/335599e4b549b7476db65214b5d3d007a47fc420/paimon-format/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java#L650-L673

But ParquetFileReader.readVectored still need to call method ParquetInputStream#readVectored ?

> We have override `ParquetFileReader` to support vector io, not through Parquet InputStream. This PR is useless for vector io.

PR #5744, opened 2025-06-13
[core] Introduce auto-buffer-spill for postpone bucket write

JingsongLi:
### Purpose

To prevent the situation of OOM caused by users writing too many partitions.


leaves12138 (approved):
+1

PR #5745, opened 2025-06-13
add test case for ListPartitionsPaged WithMultiLevel

XiaoHongbo-Hope:
### Purpose
add test case for ListPartitionsPaged WithMultiLevel

### Tests
RESTCatalogTest.testListPartitionsPagedWithMultiLevel

### DIFF
--- a/paimon-api/src/main/java/org/apache/paimon/rest/RESTUtil.java
+++ b/paimon-api/src/main/java/org/apache/paimon/rest/RESTUtil.java
@@ -24,9 +24,6 @@ import org.apache.paimon.utils.Preconditions;
 import org.apache.paimon.shade.guava30.com.google.common.collect.ImmutableMap;
 import org.apache.paimon.shade.guava30.com.google.common.collect.Maps;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import java.io.UncheckedIOException;
 import java.io.UnsupportedEncodingException;
 import java.net.URLDecoder;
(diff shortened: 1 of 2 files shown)


paimon-api/src/main/java/org/apache/paimon/rest/RESTUtil.java
```diff
@@ -34,6 +37,8 @@
 /** Util for REST. */
 public class RESTUtil {
 
+    private static final Logger LOG = LoggerFactory.getLogger(RESTUtil.class);
```
JingsongLi: Not used?

paimon-core/src/main/java/org/apache/paimon/partition/PartitionUtils.java
```diff
@@ -69,4 +73,15 @@ public static Pair<Pair<int[], RowType>, List<DataField>> constructPartitionMapp
     public static PartitionInfo create(@Nullable Pair<int[], RowType> pair, BinaryRow binaryRow) {
         return pair == null ? null : new PartitionInfo(pair.getLeft(), pair.getRight(), binaryRow);
     }
+
+    public static String buildPartitionName(Map<String, String> partitionSpec) {
+        if (partitionSpec.isEmpty()) {
+            return EMPTY_PARTITION_NAME;
```
JingsongLi: just use ""

paimon-api/src/main/java/org/apache/paimon/rest/RESTUtil.java
XiaoHongbo-Hope: fixed

XiaoHongbo-Hope: > Not used?

fixed

paimon-core/src/main/java/org/apache/paimon/partition/PartitionUtils.java
XiaoHongbo-Hope: > just use ""

fixed

JingsongLi:
+1

PR #5748, opened 2025-06-14
[flink] Introduce 'memory' type for 'lookup.cache' to in-memory cache

JingsongLi:
### Purpose

This PR aims to introduce a new in-memory mode for Flink Lookup Join. To use in-memory cache to replace RocksDB.
1. If the dim table is large, RocksDB bootstrap is too slow.
2. RocksDB get performance is not so good.

Note, use this feature has OOM risk.


JingsongLi:
> Good work! One question: If OOM, user only can add the TM's Heap memory?Are they have any other ways?
> 
> Can we add a threshold, such as specifying that only 100MB can be used to store in-memory-indexes, and using Rocksdb for the excess portion?

Good point, We can proceed to complete it next, but it is a difficult decision because we cannot determine how much memory can be used.

wwj6591812:
+1

PR #5751, opened 2025-06-15
[core] Simplify force rewrite files in Compact Task

JingsongLi:
### Purpose

At present, the code assumes that the subsequent compaction will be upgraded file by file, which will not cause the order to be disrupted. However, there are some code risks that are very subtle.

This PR refactored here and directly generated a task that must be rewriting the file.


JingsongLi:
cc @mxdzs0612

JingsongLi:
cc @LsomeYeah

wwj6591812:
The code LGTM.
But a question, of `However, there are some code risks that are very subtle.`
What the code risks that are very subtle.?

mxdzs0612 (approved):
LGTM

PR #5753, opened 2025-06-16
[core][test] add test case for picking some small files in top level when full compaction

LsomeYeah:
### Purpose

Linked issue: close #xxx

As mentioned in https://github.com/apache/paimon/pull/5751, if only pick some small files in top level and rewriting them to a big file when full compaction, it might lead to the generated big file overlapped with other big files. https://github.com/apache/paimon/pull/5751 fixed this bug, and this pr added a test for this case.


JingsongLi:
+1 Thanks @LsomeYeah

PR #5754, opened 2025-06-16
[core][spark] Introduce default value when writing from Spark DDL

JingsongLi:
### Purpose

Paimon allows specifying default values for columns. When users write to these tables without explicitly providing
values for certain columns, Paimon automatically generates default values for these columns.

#### Create Table

You can create a table with columns with default values using the following SQL:

```sql
CREATE TABLE my_table (
    a BIGINT,
    b STRING DEFAULT 'my_value',
    c INT DEFAULT 5 
);
```

#### Insert Table

For SQL commands that execute table writes, such as the `INSERT`, `UPDATE`, and `MERGE` commands, the `DEFAULT` keyword
or `NULL` value is parsed into the default value specified for the corresponding column.

#### Limitation

Currently, only specifying default values when creating tables is supported, and the following usage is not supported:

1. Not support alter table add column with default value, for example: `ALTER TABLE T ADD COLUMN d INT DEFAULT 5;`.
2. Not support alter table alter column set default value, for example: `ALTER TABLE T ALTER COLUMN d SET DEFAULT 5;`.

### DIFF
--- a/paimon-spark/paimon-spark-3.2/src/main/scala/org/apache/spark/sql/connector/catalog/TableCatalogCapability.java
+++ b/paimon-spark/paimon-spark-3.2/src/main/scala/org/apache/spark/sql/connector/catalog/TableCatalogCapability.java
@@ -1,0 +1,52 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.spark.sql.connector.catalog;
+
+/**
+ * Capabilities that can be provided by a {@link TableCatalog} implementation.
+ */
+public enum TableCatalogCapability {
+
+  /**
+   * Signals that the TableCatalog supports defining generated columns upon table creation in SQL.
+   * <p>
+   * Without this capability, any create/replace table statements with a generated column defined
+   * in the table schema will throw an exception during analysis.
+   * <p>
+   * A generated column is defined with syntax: {@code colName colType GENERATED ALWAYS AS (expr)}
+   * <p>
+   * Generation expression are included in the column definition for APIs like
+   * {@link TableCatalog#createTable}.
+   */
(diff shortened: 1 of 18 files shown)


paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/SparkCatalog.java
```diff
@@ -103,6 +111,10 @@ public class SparkCatalog extends SparkBaseCatalog
 
     private String defaultDatabase;
 
+    public Set<TableCatalogCapability> capabilities() {
```
Zouxxyy, line 114: Can SparkGenericCatalog support this capability?

yuzelin (approved):
+1

paimon-spark/paimon-spark-3.2/src/main/scala/org/apache/spark/sql/connector/catalog/TableCatalogCapability.java
```diff
@@ -0,0 +1,52 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.spark.sql.connector.catalog;
+
+/**
+ * Capabilities that can be provided by a {@link TableCatalog} implementation.
+ */
+public enum TableCatalogCapability {
```
Zouxxyy, line 24: The `TableCatalogCapability` interface is introduced from Spark 3.4, will Spark 3.2 and 3.3 also have this capability after this, or just for compilation compatibility?

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/SparkCatalog.java
JingsongLi: No, not in this PR.

paimon-spark/paimon-spark-3.2/src/main/scala/org/apache/spark/sql/connector/catalog/TableCatalogCapability.java
JingsongLi: Just for compatibility. Copy file to old versions.

paimon-spark/paimon-spark-ut/src/test/java/org/apache/paimon/spark/SparkWriteITCase.java
```diff
@@ -78,7 +79,49 @@ public void stopMetastoreAndSpark() {
...
     @AfterEach
     public void afterEach() {
-        spark.sql("DROP TABLE T");
+        spark.sql("DROP TABLE IF EXISTS T");
+    }
+
+    @Test
+    public void testWriteWithDefaultValue() {
+        spark.sql(
+                "CREATE TABLE T (a INT, b INT DEFAULT 2, c STRING DEFAULT 'my_value') TBLPROPERTIES"
```
xumanbu, line 88: @JingsongLi I have test with `'primary-key'  = a` , the default value mechanism not work, is not support with PK table?

JingsongLi: Thanks @xumanbu , fix this in https://github.com/apache/paimon/pull/6050

PR #5760, opened 2025-06-17
[core] Support format reader and writer use File IO directly

leaves12138:
### Purpose

To support lance file format:
we need to
* Allow file format class to create its own file io
* Pass the storage options to file format class

The is a pre-pull request for lance file format support.

### DIFF
--- a/paimon-api/src/main/java/org/apache/paimon/options/Options.java
+++ b/paimon-api/src/main/java/org/apache/paimon/options/Options.java
@@ -23,6 +23,7 @@ import org.apache.paimon.annotation.Public;
 import javax.annotation.concurrent.ThreadSafe;
 
 import java.io.Serializable;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
@@ -45,6 +46,8 @@ import static org.apache.paimon.options.OptionsUtils.removePrefixMap;
 @Public
 @ThreadSafe
 public class Options implements Serializable {
+
+    public static final Options EMPTY_OPTIONS = new Options(Collections.emptyMap());
 
     private static final long serialVersionUID = 1L;
 
--- a/paimon-api/src/main/java/org/apache/paimon/options/Options.java
+++ b/paimon-api/src/main/java/org/apache/paimon/options/Options.java
@@ -23,7 +23,6 @@ import org.apache.paimon.annotation.Public;
 import javax.annotation.concurrent.ThreadSafe;
 
 import java.io.Serializable;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
@@ -46,8 +45,6 @@ import static org.apache.paimon.options.OptionsUtils.removePrefixMap;
 @Public
 @ThreadSafe
 public class Options implements Serializable {
-
-    public static final Options EMPTY_OPTIONS = new Options(Collections.emptyMap());
 
     private static final long serialVersionUID = 1L;
 
--- a/paimon-common/src/main/java/org/apache/paimon/format/FileFormat.java
+++ b/paimon-common/src/main/java/org/apache/paimon/format/FileFormat.java
@@ -21,7 +21,6 @@ package org.apache.paimon.format;
 import org.apache.paimon.CoreOptions;
 import org.apache.paimon.factories.FormatFactoryUtil;
 import org.apache.paimon.format.FileFormatFactory.FormatContext;
-import org.apache.paimon.fs.FileIO;
 import org.apache.paimon.options.Options;
 import org.apache.paimon.predicate.Predicate;
 import org.apache.paimon.statistics.SimpleColStatsCollector;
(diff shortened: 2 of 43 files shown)


TaoZex (approved):
+1

paimon-api/src/main/java/org/apache/paimon/options/Options.java
```diff
@@ -46,6 +47,8 @@
 @ThreadSafe
 public class Options implements Serializable {
 
+    public static final Options EMPTY_OPTIONS = new Options(Collections.emptyMap());
```
JingsongLi: Do not introduce a mutable final object.

paimon-common/src/main/java/org/apache/paimon/fs/FileIO.java
```diff
@@ -380,6 +382,12 @@ default void copyFiles(Path sourceDirectory, Path targetDirectory, boolean overw
         }
     }
 
+    default Options storageOptions() {
```
JingsongLi: I prefer not introducing this method. Just introduce inner methods for fileio implementations. For example, introduce some to `OSSFileIO`.

Sometimes, FileIO needs whole `CatalogContext` to init it.

paimon-common/src/main/java/org/apache/paimon/format/FileFormat.java
```diff
@@ -77,10 +77,12 @@ public Optional<SimpleStatsExtractor> createStatsExtractor(
         return Optional.empty();
     }
 
-    public static FileFormat fromIdentifier(String identifier, Options options) {
+    public static FileFormat fromIdentifier(
+            String identifier, Options storageOptions, Options options) {
```
JingsongLi: Can we just give FileIO to the file format?

leaves12138: OK

paimon-common/src/main/java/org/apache/paimon/fs/FileIO.java
leaves12138: OK

paimon-api/src/main/java/org/apache/paimon/options/Options.java
leaves12138: OK

leaves12138:
> Can we introduce an interface:
> 
> ```java
> public interface SupportsDirectWrite extends FileFormat {
>       FormatWriter create(FileIO fileIO, Path path, String compression) throws IOException;
> }
> ```
> 
> In this way, we don't need introduce field to factory.

OK

paimon-filesystems/paimon-jindo/src/main/java/org/apache/paimon/jindo/JindoFileIO.java
```diff
@@ -125,6 +125,10 @@ public void configure(CatalogContext context) {
         }
     }
 
+    public Options storageOptions() {
```
JingsongLi: hadoopOptions

paimon-filesystems/paimon-oss-impl/src/main/java/org/apache/paimon/oss/OSSFileIO.java
```diff
@@ -115,6 +115,10 @@ public void configure(CatalogContext context) {
         }
     }
 
+    public Options storageOptions() {
```
JingsongLi: hadoopOptions

paimon-core/src/main/java/org/apache/paimon/io/SingleFileWriter.java
```diff
@@ -67,20 +68,31 @@ public SingleFileWriter(
...
 
-        try {
-            out = fileIO.newOutputStream(path, false);
-            if (asyncWrite) {
-                out = new AsyncPositionOutputStream(out);
+        if (factory instanceof SupportsDirectWrite) {
+            try {
+                writer = ((SupportsDirectWrite) factory).create(fileIO, path, compression);
+            } catch (IOException e) {
+                LOG.warn(
```
JingsongLi: Why copying this log and copy try catch?

paimon-filesystems/paimon-jindo/src/main/java/org/apache/paimon/jindo/JindoFileIO.java
leaves12138: OK

paimon-filesystems/paimon-oss-impl/src/main/java/org/apache/paimon/oss/OSSFileIO.java
leaves12138: OK

paimon-core/src/main/java/org/apache/paimon/io/SingleFileWriter.java
leaves12138: Removed duplicated try catch

JingsongLi (approved):
+1

leaves12138:
> How to use lance format in PAIMON？

It's not totally ready yet. But if you want to test it, you can contact me.

PR #5763, opened 2025-06-17
[flink] Support loading custom CommitListener at runtime

yuzelin:
### Purpose

Enhance #4398

### DIFF
--- a/docs/layouts/shortcodes/generated/flink_connector_configuration.html
+++ b/docs/layouts/shortcodes/generated/flink_connector_configuration.html
@@ -87,6 +87,18 @@ under the License.
             <td>The blacklist contains several time periods. During these time periods, the lookup table's cache refreshing is forbidden. Blacklist format is start1-&gt;end1,start2-&gt;end2,... , and the time format is yyyy-MM-dd HH:mm. Only used when lookup table is FULL cache mode.</td>
         </tr>
         <tr>
+            <td><h5>partition.custom-listener.#.param</h5></td>
+            <td style="word-wrap: break-word;">(none)</td>
+            <td>String</td>
+            <td>Parameter string for initializing the listener class #. Listener class should parse the parameter by itself.</td>
+        </tr>
+        <tr>
+            <td><h5>partition.custom-listeners</h5></td>
+            <td style="word-wrap: break-word;">(none)</td>
+            <td>String</td>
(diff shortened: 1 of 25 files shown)


docs/layouts/shortcodes/generated/flink_connector_configuration.html
```diff
@@ -86,6 +86,18 @@
             <td>String</td>
             <td>The blacklist contains several time periods. During these time periods, the lookup table's cache refreshing is forbidden. Blacklist format is start1-&gt;end1,start2-&gt;end2,... , and the time format is yyyy-MM-dd HH:mm. Only used when lookup table is FULL cache mode.</td>
         </tr>
+        <tr>
+            <td><h5>partition.custom-listener.#.param</h5></td>
```
JingsongLi: Do we really need this?

docs/layouts/shortcodes/generated/flink_connector_configuration.html
```diff
@@ -86,6 +86,18 @@
...
+            <td style="word-wrap: break-word;">(none)</td>
+            <td>String</td>
+            <td>Parameter string for initializing the listener class #. Listener class should parse the parameter by itself.</td>
+        </tr>
+        <tr>
+            <td><h5>partition.custom-listeners</h5></td>
```
JingsongLi: just listener instead listeners

docs/layouts/shortcodes/generated/flink_connector_configuration.html
yuzelin: We can pass parameters by table dynamic options.

docs/layouts/shortcodes/generated/flink_connector_configuration.html
yuzelin: Should we follow the name style of option `commit.callbacks` in CoreOptions?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/listener/ReportPartStatsListener.java
```diff
@@ -52,7 +53,6 @@
  */
 public class ReportPartStatsListener implements CommitListener {
```
JingsongLi, line 53: Do not create factory for inner listeners.

PR #5765, opened 2025-06-18
[core] Rename PartitionListener to CommitListener

JingsongLi:
### Purpose

1. It should be CommitListener
2. Do not add `partitionMarkDoneRecoverFromState` to CommitListener


yuzelin (approved):
+1

PR #5769, opened 2025-06-18
[flink] fall back to full cache mode when partial-lookup is unavailable.

### LINKED ISSUE #5768: [Feature] fall back to full cache when partial-lookup is unavailable
### Search before asking

### Motivation

if dim table uses deduplicate merge engine with sequence fields and the table primary keys equal join keys, partial-lookup mode will be matched. Then the following exception happens:

Image

Most users do not know these limits well, it would be better fall back to full cache mode when partial-lookup is unavailable instead of throwing a exception.

### Solution

_No response_

### Anything else?

_No response_

zhoulii:
### Purpose

Linked issue: close #5768


JingsongLi (approved):
+1 Thanks @zhoulii

Issue #5770, opened 2025-06-18
[Feature] Hope to retrieve FileStoreTable information via HMS in SparkSource#loadTable.


Cribbee:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Motivation

In the current implementation, when writing to an existing Paimon table via Spark DataFrame (e.g., df.write.format("paimon").mode("append").option("path",xxx).save()), the SparkTable initialization can only create a FileStoreTable instance through CoreOptions.path(options), even if a Hive Catalog has already been configured.

While using table-path for initialization is ok, there exists a scenario where the path of an existing Paimon table does not conform to the standard path format. This results in validation failure and write errors, with the following message:

"Path 'xxx' is not a valid path. Please use a catalog table path instead: 'warehouse_path/your_database.db/your_table'."

### Solution

Add a method in CoreOptions to retrieve TABLE_ID (the full name of the table), then instantiate a FileStoreTable via HiveCatalog to resolve the aforementioned issues.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #5772, opened 2025-06-18
[spark] Instantiate FileStoreTable by accessing HMS (Hive Metastore) via tableId in SparkSource#loadTable.

Cribbee:
### Purpose

This PR aims to retrieve FileStoreTable information via HMS in SparkSource#loadTable.
1.  DataFrame can be written using df.write.format("paimon").mode("append").option("table",  "default.T").save()"

2. The operation remains functional even if the table path in MetaStore deviates from the strict format like "warehouse_path/your_database.db/your_table" .

Linked issue:  #5770

### DIFF
--- a/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
@@ -165,6 +165,13 @@ public class CoreOptions implements Serializable {
                     .noDefaultValue()
                     .withDescription("The file path of this table in the filesystem.");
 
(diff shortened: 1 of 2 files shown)


Zouxxyy:
Can you try dataframe `saveAsTable ` in https://paimon.apache.org/docs/master/spark/quick-start/#insert-table

Zouxxyy:
> @Zouxxyy Oh, it works Thanx! And is it necessary to extend df.save() to support writing via table-name specification as well, if not i’ll close it.

I think using the provided saveAsTable API directly would be sufficient for now, and there's no need to add new options at this time

PR #5774, opened 2025-06-19
[core] rest: support function paged api

jerry-024:
### Purpose

support function list page api
- listFunctionsPaged
- listFunctionDetailsPaged
- listFunctionsPagedGlobally
### Tests
- org.apache.paimon.rest.RESTCatalogTest#testListFunctions


JingsongLi (approved):
Looks good to me!

PR #5571, opened 2025-05-07
[Bug] The Hive connector encounters time offset issues when reading and writing data of the timestamp with local time zone field type.

Jack1007:
### Purpose

Linked issue: open #5568
This PR fixes time zone offset issues in the Hive connector when handling the `timestamp with local time zone` data type. To explain the problem, we first introduce two semantic interpretations of timestamps: **Instant semantics** and **Local semantics**.

### Instant Semantics

Instant semantics corresponds to timestamp with local time zone (equivalent to Flink's TIMESTAMP_LTZ type). It stores a fixed UTC timestamp, which is displayed according to the user's current process or session time zone. Users in different time zones see different local times, but they all represent the same absolute moment. The following Java code demonstrates Instant semantics:
```java
// Instant semantics
long nowMs = System.currentTimeMillis();
System.out.println(Instant.ofEpochMilli(nowMs).atZone(TimeZone.getTimeZone("Asia/Shanghai").toZoneId()));
System.out.println(Instant.ofEpochMilli(nowMs).atZone(TimeZone.getTimeZone("Asia/Tokyo").toZoneId()));

```
**Output (same moment, different time zones):**
```
2025-05-07T10:40:32.107+08:00[Asia/Shanghai]
2025-05-07T11:40:32.107+09:00[Asia/Tokyo]
```
### Local Semantics
Local semantics corresponds to a timezone-agnostic `timestamp` type. It displays local time without adjusting for time zones. Users in different time zones see the same displayed time, but these times do not represent the same absolute moment. The following Java code demonstrates Local semantics:
``` java
// Local semantics
LocalDateTime localDateTime = LocalDateTime.now();
System.out.println(localDateTime.atZone(ZoneId.of("Asia/Shanghai")));
System.out.println(localDateTime.atZone(ZoneId.of("Asia/Tokyo")));
```
**Output (same displayed time, different moments):**
```
2025-05-07T10:40:32.222+08:00[Asia/Shanghai]
2025-05-07T10:40:32.222+09:00[Asia/Tokyo]
```
### Root Cause Analysis
The Hive connector incorrectly treated timestamp with local time zone data using Local semantics, while engines like Spark and Flink correctly followed Instant semantics. This discrepancy caused time zone offsets in GMT+8:

- **Hive writes + Spark/Flink reads:** Timestamps appeared **8 hours ahead**

- **Spark writes + Hive reads:** Timestamps appeared **8 hours behind**

### Solution

[...]

PR #5776, opened 2025-06-19
[core] Optimize commit retry on exception

JingsongLi:
### Purpose

We can retry commit in `FileStoreCommitImpl` when exception. This can significantly improve the robustness of the code.
1. Remove file reusing in `RetryResult`.
2. Add retry when commit exception.

### Tests

In `MockRESTCatalogTest`.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
@@ -778,12 +778,6 @@ public class FileStoreCommitImpl implements FileStoreCommit {
         long startMillis = System.currentTimeMillis();
         while (true) {
             Snapshot latestSnapshot = snapshotManager.latestSnapshot();
-
-            if (alreadyCommitted(identifier, retryResult, latestSnapshot)) {
-                // already success in table
-                break;
-            }
-
             CommitResult result =
                     tryCommitOnce(
                             retryResult,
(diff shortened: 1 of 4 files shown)


paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
```diff
@@ -801,11 +801,15 @@ private int tryCommit(
...
-                throw new RuntimeException(
+                String message =
                         String.format(
                                 "Commit failed after %s millis with %s retries, there maybe exist commit conflicts between multiple jobs.",
-                                commitTimeout, retryCount));
+                                commitTimeout, retryCount);
+                if (retryResult.exception != null) {
+                    throw new RuntimeException(message, retryResult.exception);
+                } else {
+                    throw new RuntimeException(message);
```
tsreaper:
```suggestion
                throw new RuntimeException(message, retryResult.exception);
```

paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
```diff
@@ -1110,54 +1088,57 @@ CommitResult tryCommitOnce(
...
-            LOG.info(
-                    "Successfully commit snapshot {} to table {} by user {} "
-                            + "with identifier {} and kind {}.",
+        boolean success;
+        try {
+            success = commitSnapshotImpl(newSnapshot, deltaStatistics);
+        } catch (Exception e) {
+            // commit exception, not sure about the situation and should not clean up the files
+            LOG.warn("Retry commit for exception.", e);
+            return new RetryResult(latestSnapshot, baseDataFiles, e);
```
tsreaper, line 1094: When committing `APPEND` snapshot, `conflictCheck` will be a `noConflictCheck`. Let's say `commitSnapshotImpl` throws an exception, but the commit actually succeeds. If we retry from `tryCommit`, new files will not be checked for conflicts, and the same new file might be committed twice.

Any exception must be thrown from this method. Retry for exception should only be done outside this class.

tsreaper (changes requested):

paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
tsreaper: Alright, after #5771 this is not a problem.

tsreaper: #5571 should also check type of snapshots must be the same.

PR #5780, opened 2025-06-19
[doc] remove spaces after backslash which will cause bash command execution fail

Shawn-Hx:
### Purpose

The space after \ will cause failure when executing command.

image


JingsongLi:
+1

PR #5784, opened 2025-06-20
[cdc] Support computed columns when sync_database

JackeyLee007:
### Purpose

To use computed columns when sync_database. The computed columns could refer to existed fields or other computed columns.

### Tests

1. Test computed columns when sync_database, with e2e case 
org.apache.paimon.flink.action.cdc.kafka.KafkaCanalSyncDatabaseActionITCase#testExpressionNow
, by adding a new filed: pt=date_format(etl_update_time,yyyy-MM-dd).

2. Test the sortting of computed columns, with case: 
org.apache.paimon.flink.action.cdc.ComputedColumnUtilsTest


paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/action/cdc/Expression.java
```diff
@@ -50,12 +51,18 @@ public interface Expression extends Serializable {
...
+
     /** Return {@link DataType} of computed value. */
     DataType outputType();
 
     /** Compute value from given input. Input and output are serialized to string. */
     String eval(String input);
 
+    /** Compute value from given input. Input and output are serialized to string. */
+    String eval(String input, DataType inputType);
+
```
yuzelin: I think we can just leave the `eval(String input, DataType inputType)` method.

PR #5790, opened 2025-06-24
[core] fix that batch CompactAction makes no sense with all level-0 files

LsomeYeah:
### Purpose

Linked issue: close #xxx

If the table is dv enabled or using first-row as merge engine, and its writing job is write-only, compact action in batch mode for this table will generate 0 splits. This is because `DataTableBatchScan` will skip all level-0 files when dv enabled or with first-row merge engine while files written with write-only set to true are all in level-0. 

This pr introduced a new option "batch-scan-mode" which should only be used internally, if `batch-scan-mode = compact`, the `scan` won't apply the level-0 filter.

### Tests

CompactActionITCase#testCompactWhenSkipLevel0


JingsongLi (approved):
+1

PR #5793, opened 2025-06-25
Provide option to disable use of source primary keys if primary keys in action command are not specified for CDC ingestion.

AshishKhatkar:
### Purpose
To add an option to not use source primary keys when doing CDC ingestion. 

Linked issue: close #xxx

We have a use case where we want to do ingestion of keyed topics from kafka in append only mode. In the existing case the table is created with primary keys even if we don't specify primary keys using `--primary_keys`. This change will help us to automatically create tables without having primary keys even for keyed topics (cdc topics) using action jar. The default is kept as true as I didn't want to change the existing behaviour.

### DIFF
--- a/docs/layouts/shortcodes/generated/kafka_sync_database.html
+++ b/docs/layouts/shortcodes/generated/kafka_sync_database.html
@@ -115,8 +115,14 @@ under the License.
     <tr>
         <td><h5>--primary_keys</h5></td>
         <td>The primary keys for Paimon table. If there are multiple primary keys, connect them with comma, for example "buyer_id,seller_id".
-            If the keys are not in source table, but the source table has primary keys, the sink table will use source table's primary keys.
-            Otherwise, the sink table won't set primary keys.</td>
+            If the keys are not provided, but the source has primary keys, the sink table will use source's primary keys.
+            Otherwise, the sink table won't set primary keys.
+            If the keys are not provided, but the source has primary keys, and you don't want to use source's primary keys,
+            use --use_pkeys_from_source_for_paimon_schema.</td>
+    </tr>
+    <tr>
+        <td><h5>--use_pkeys_from_source_for_paimon_schema</h5></td>
+        <td>This is used to specify if primary keys from source should be used in paimon schema if primary keys using --primary_keys are not specified. The default is true.</td>
     </tr>
     <tr>
     <tr>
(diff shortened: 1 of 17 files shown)


docs/layouts/shortcodes/generated/kafka_sync_database.html
```diff
@@ -115,8 +115,14 @@
...
         <td>The primary keys for Paimon table. If there are multiple primary keys, connect them with comma, for example "buyer_id,seller_id".
-            If the keys are not in source table, but the source table has primary keys, the sink table will use source table's primary keys.
-            Otherwise, the sink table won't set primary keys.</td>
+            If the keys are not provided, but the source has primary keys, the sink table will use source's primary keys.
+            Otherwise, the sink table won't set primary keys.
+            If the keys are not provided, but the source has primary keys, and you don't want to use source's primary keys,
+            use --use_pkeys_from_source_for_paimon_schema.</td>
+    </tr>
+    <tr>
+        <td><h5>--use_pkeys_from_source_for_paimon_schema</h5></td>
```
JingsongLi: `--sync_primary_keys_from_source_schema`?

paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/action/cdc/CdcActionCommonUtils.java
```diff
@@ -205,7 +210,7 @@ private static void setPrimaryKeys(
             }
         }
 
-        if (!sourceSchemaPrimaryKeys.isEmpty()) {
+        if (usePKeysFromSourceForPaimonSchema && !sourceSchemaPrimaryKeys.isEmpty()) {
```
JingsongLi: Can you add a test case for this PR?

AshishKhatkar: The only tests that I find for this are IT tests. I can add them. Do you want me to add them for all CDC sources or just adding one and verifying it is enough?

JingsongLi: Just add one is OK.

AshishKhatkar:
@JingsongLi can you review the PR now.

AshishKhatkar:
@JingsongLi can you review the PR? We need this internally for our use cases.

PR #5795, opened 2025-06-26
[core] Introduce Row Id for append table

leaves12138:
### Purpose

This pull introduce global row id for paimon table.

A ROW ID is defined as a globally unique, auto-incrementing sequence, serving as the "identity card" for each piece of data. Every single data record has one, and only one, globally unique ROW ID.

ROW IDs possess the following characteristics:

* Immutability: Once a ROW ID is generated, it cannot be changed. It serves as the sole identifier for a data record. When performing COMPACTION (data merging), the corresponding ROW IDs must be merged, but they remain unchangeable themselves. When UPDATING data, a new ROW ID should be generated for the updated record, treating the updated data as a distinct record from the original.

* Auto-incrementing: ROW IDs are allocated based on the largest ROW ID from the previous SNAPSHOT. The maximum allocated ROW ID for the current SNAPSHOT is then recorded, facilitating allocation for the next SNAPSHOT.

* Hidden: The ROW ID functions as an internal sequence number. It is not recorded in the table schema and is not exposed or perceptible externally.

image

This pull request is the base of `Global Index` and `Quick Addition Column`

* System Table
Introduced system table for row id.
e.g.  
```
SELECT * FROM <table>$with_row_id
```
Results in:
_row_id, columns0, column1
0              a                 a
1              b                 b
2              c                 c

### DIFF
--- a/paimon-api/src/main/java/org/apache/paimon/Snapshot.java
+++ b/paimon-api/src/main/java/org/apache/paimon/Snapshot.java
@@ -88,6 +88,7 @@ public class Snapshot implements Serializable {
     protected static final String FIELD_WATERMARK = "watermark";
     protected static final String FIELD_STATISTICS = "statistics";
     protected static final String FIELD_PROPERTIES = "properties";
+    protected static final String FIELD_NEXT_ROW_ID = "nextRowId";
 
     // version of snapshot
     // null for paimon <= 0.2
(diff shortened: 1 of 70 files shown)


JingsongLi (changes requested):
Why not using `_MIN_SEQUENCE_NUMBER` and `_MAX_SEQUENCE_NUMBER`?

wwj6591812:
What problem is this feature introduced to solve?

leaves12138:
> What problem is this feature introduced to solve?

To address the rapid addition of column values in machine learning structured training scenarios and the fast point query problem in multi-modal scenarios.

leaves12138:
This pull request should be closed because strategy adjustment， see https://github.com/apache/paimon/pull/5935

PR #5796, opened 2025-06-26
[flink] All lookup writers should store active buckets in state to make sure changelog can be produced

tsreaper:
### Purpose

This PR fixes the same bug in #3512, but in the scenario when `lookup-wait` is `true`. Consider the following case:

A lookup writer has produced both new files and compacted changelogs during one checkpoint, however only the APPEND snapshot is committed and the COMPACT snapshot fails to commit due to exception.

When the job restarts, the COMPACT snapshot will not be committed again to avoid potential manifest conflicts. If records of this bucket never comes after restart, the changelogs will be lost.

### Tests

* `WriterOperatorTest#testLookupWithFailure`.

### API and Format

No format changes.

### Documentation

No new feature.

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkSink.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkSink.java
@@ -143,10 +143,10 @@ public abstract class FlinkSink<T> implements Serializable {
             }
         }
 
-        if (coreOptions.laziedLookup()) {
+        if (coreOptions.needLookup()) {
             return (table, commitUser, state, ioManager, memoryPool, metricGroup) -> {
                 assertNoSinkMaterializer.run();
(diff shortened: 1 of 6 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkSink.java
```diff
@@ -143,10 +143,10 @@ private StoreSinkWrite.Provider createWriteProvider(
             }
         }
 
-        if (coreOptions.laziedLookup()) {
+        if (coreOptions.needLookup()) {
             return (table, commitUser, state, ioManager, memoryPool, metricGroup) -> {
                 assertNoSinkMaterializer.run();
-                return new AsyncLookupSinkWrite(
+                return new LookupSinkWrite(
```
JingsongLi: We should create this only write only is false.

PR #5798, opened 2025-06-26
[spark] Format table support show partitions

Zouxxyy:
### Purpose

Support show partitions for Format table
```sql
SHOW PARTITIONS format_tbl
SHOW PARTITIONS format_tbl PARTITION (p1=1)
```


JingsongLi (approved):
Thanks @Zouxxyy !

PR #5807, opened 2025-06-27
[flink] Fix ClassNotFoundException when deserializing writer coordination response

tsreaper:
### Purpose

Currently, Flink's RPC gateway can only deserialize classes accessible from Flink's class loader (see [apache/flink](https://github.com/apache/flink/blob/ce03ea11d10df28affa040847dbaf02680eb785d/flink-rpc/flink-rpc-akka/src/main/java/org/apache/flink/runtime/rpc/pekko/PekkoRpcService.java#L543), so we have to wrap our own response with a class (`CollectCoordinationResponse`) in Flink's project.

### Tests

E2e tests.

### API and Format

No format changes.

### Documentation

No new feature.


JingsongLi:
+1

PR #5808, opened 2025-06-27
[flink] Fix the document of Fixed Bucket Large Scale Lookup optimization

yunfengzhou-hub:
### Purpose

This PR fixes the document for Fixed Bucket Large Scale Lookup optimization. This optimization takes effect only when the join keys contain all the bucket keys (See BaseDataTableSource#supportBucketShufflePartitioner). The current document missed this point.


JingsongLi:
+1

PR #5810, opened 2025-06-27
[rest] fix base header options in RESTApi

shyjsarah:
### Purpose
Fixed base header not being set when 'configRequired' is false

Linked issue: close #xxx

### Tests

org.apache.paimon.rest.MockRESTCatalogTest#testHeaderOptions


paimon-api/src/main/java/org/apache/paimon/rest/RESTApi.java
```diff
@@ -170,13 +170,13 @@ public RESTApi(Options options, boolean configRequired) {
         this.client = new HttpClient(options.get(RESTCatalogOptions.URI));
         AuthProvider authProvider = createAuthProvider(options);
         Map<String, String> baseHeaders = Collections.emptyMap();
```
JingsongLi: Map<String, String> baseHeaders = extractPrefixMap(options, HEADER_PREFIX);

JingsongLi (approved):
+1

PR #5812, opened 2025-06-27
[expire] Optimize expired partition strategy

zhuanshenbsj1:
### Purpose

The current parsing of partition time uses the earliest time of the partition, which is unreasonable for partition expiration checks.

Example:

Set partition.expiration-time = '1 d' .

Current time: 2025-06-27 09:00:00, we need to verify if data in the 25th partition can be deleted.
The default parsed time for the 26th partition is 2025-06-26 00:00:00.
2025-06-27 09:00:00 - 2025-06-26 00:00:00 = 33 hours > 24 hours, triggering incorrect deletion eligibility.

Proposed solution:

Default to parsing the latest time of the day for partitions, and support user-defined configurations .
Examples:
Hourly partitions: $day $hour:59:59
Minute partitions: $day $hour:$minute:59


zhuanshenbsj1:
> values-time

Even if we use values-time, this is not reasonable.  For example, if a user sets a 1D partition expiration time for a Paimon table, and  when they query yesterday's partition data on the morning of the next day,  they find that the data has already been cleaned up.

JingsongLi:
I don't think this is optimization, nor should it be directly modified. This is a semantic issue, and we can add documentation to explain this behavior. Hi @zhuanshenbsj1 , If you wish to continue contributing, you can propose a new PR.

PR #5814, opened 2025-06-29
Fix: Exlude parquet 1.13.1 from spark-sql in paimon-spark-ut to avoid overwriting parquet 1.15.1.

HannanKan:
### Purpose
Fix: Exlude parquet 1.13.1 from spark-sql in paimon-spark-ut to avoid overwriting parquet 1.15.1.

bugfix: when running org.apache.paimon.spark.sql.BucketedTableQueryTest, got following error message:
image

root cause: Importing spark-sql in test scope will improt parquet 1.13.1 too. However, Paimon uses parquet 1.15.1 and there are some new functions  in 1.15.1 called by Paimon.

After exluding, tests case run successfully.

Linked issue: close #xxx

### DIFF
--- a/paimon-spark/paimon-spark-ut/pom.xml
+++ b/paimon-spark/paimon-spark-ut/pom.xml
@@ -106,6 +106,10 @@ under the License.
                     <groupId>com.fasterxml.jackson.core</groupId>
                     <artifactId>*</artifactId>
                 </exclusion>
+                <exclusion>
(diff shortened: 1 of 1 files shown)


JingsongLi:
Hi @HannanKan , The test was unsuccessful and Spark POM has been refactored. If there are still any issues, please submit a new PR.

PR #5821, opened 2025-07-01
[hotfix] add detail information for NoPermissionException

qingwei727:
### Purpose
We cannot judge where or why we have no permission for the table/database, so we need include more detailed descriptions of lacking permissions.

### Tests
MockRESTCatalogTest.testDatabaseApiWhenNoPermission
MockRESTCatalogTest.testApiWhenTableNoPermission


JingsongLi:
+1

PR #5822, opened 2025-07-01
[core] update test as rest catalog support alter format table

jerry-024:
### Purpose
Update test as rest catalog support alter format table


paimon-core/src/test/java/org/apache/paimon/catalog/CatalogTestBase.java
```diff
@@ -1329,8 +1329,13 @@ public void testFormatTable() throws Exception {
 
         // alter table
         SchemaChange schemaChange = SchemaChange.addColumn("new_col", DataTypes.STRING());
-        assertThatThrownBy(() -> catalog.alterTable(identifier, schemaChange, false))
-                .hasMessageContaining("Only data table support alter table.");
+        if (supportsAlterFormatTable()) {
+            assertDoesNotThrow(() -> catalog.alterTable(identifier, schemaChange, false));
```
JingsongLi: Please test column, options, really work

PR #5825, opened 2025-07-01
[core] Fix stats filter by pushdown after schema evolution

yuzelin:
### Purpose

Reproduce steps:
```
CREATE TABLE T (a STRING);
INSERT INTO T VALUES ('9'), ('10'), ('11');
ALTER TABLE T MODIFY (a INT);
SELECT * FROM T WHERE a > 9; -- return empty but expected (10), (11)
```

The reason is stats type has been modified (similar to #4705 ).

### Tests

Note:
SchemaEvolutionTest#testAddField for `nullSafe` arg.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/KeyValueFileStoreScan.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/KeyValueFileStoreScan.java
@@ -125,7 +125,8 @@ public class KeyValueFileStoreScan extends AbstractFileStoreScan {
         }
 
         Predicate safeKeyFilter =
-                fieldKeyStatsConverters.toEvolutionSafeFilter(entry.file().schemaId(), keyFilter);
+                fieldKeyStatsConverters.toEvolutionSafeStatsFilter(
+                        entry.file().schemaId(), keyFilter);
         if (safeKeyFilter == null) {
             return true;
         }
@@ -159,7 +160,7 @@ public class KeyValueFileStoreScan extends AbstractFileStoreScan {
                     schemaId2DataFilter.computeIfAbsent(
                             entry.file().schemaId(),
                             id ->
-                                    fieldValueStatsConverters.toEvolutionSafeFilter(
+                                    fieldValueStatsConverters.toEvolutionSafeStatsFilter(
                                             entry.file().schemaId(), valueFilter));
             return predicate.evaluate(dataPredicate).remain();
         } catch (IOException e) {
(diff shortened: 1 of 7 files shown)


yuzelin:
SchemaEvolutionTableTestBase and related tests are removed.

yuzelin:
> > SchemaEvolutionTableTestBase and related tests are removed.
> 
> now which tests can cover the cases here?

I will check all the tests and find tests that can be kept.

paimon-core/src/main/java/org/apache/paimon/stats/SimpleStatsEvolutions.java
```diff
@@ -78,19 +86,27 @@ public SimpleStatsEvolution getOrCreate(long dataSchemaId) {
...
                 Objects.requireNonNull(
                         SchemaEvolutionUtil.devolveDataFilters(
-                                schemaFields.apply(tableSchemaId),
+                                tableDataFields,
                                 schemaFields.apply(dataSchemaId),
-                                Collections.singletonList(filter)));
+                                Collections.singletonList(filter),
+                                isKeyStats,
+                                nullSafe));
         return devolved.isEmpty() ? null : devolved.get(0);
```
JingsongLi: Here should consider and filter, should deal with multiple predicates situation.

paimon-core/src/main/java/org/apache/paimon/stats/SimpleStatsEvolutions.java
```diff
@@ -78,19 +86,27 @@ public SimpleStatsEvolution getOrCreate(long dataSchemaId) {
...
     }
 
+    // Stats filter is unsafe if it should be devolved and the devolving is unsafe
+    // null safe: it's safe if the predicate field is added later and filter it on the old schema
+    public boolean statsFilterUnsafe(ManifestEntry entry, Predicate statsFilter) {
```
JingsongLi: Should not return bool, we should return predicate, here should consider and filter, should deal with multiple predicates situation.

paimon-core/src/main/java/org/apache/paimon/operation/KeyValueFileStoreScan.java
```diff
@@ -88,7 +88,8 @@ public KeyValueFileStoreScan(
         this.fieldKeyStatsConverters =
                 new SimpleStatsEvolutions(
                         sid -> keyValueFieldsExtractor.keyFields(scanTableSchema(sid)),
```
JingsongLi: We should not introduce isKeyFilter. We should remove `_KEY_` for keyFields here.

PR #4503, opened 2024-11-12
[core] small fix of FilesTable system table

wwj6591812:
### Purpose

Now search FilesTable System table, the file_path column only shows file_name, not file_path.
This PR fix it.

Linked issue: close #xxx

PR #5827, opened 2025-07-01
[test]  fix failed test in SparkWriteITCase

askwang:
### Purpose

In https://github.com/apache/paimon/pull/4503, files system table changes file_name to file_path, fix failed test.


JingsongLi:
+1

PR #5830, opened 2025-07-02
[doc] add manually expire snapshot in Manage Snapshots

jerry-024:
### Purpose
add manually expire snapshot in Manage Snapshots


yuzelin (approved):
+1

PR #5831, opened 2025-07-02
[flink] support specify max_pt for partition level

LsomeYeah:
### Purpose

Linked issue: close #xxx

Sometimes users want to specify max_pt() for a fixed-level partition and load all subpartitions for lookup. For example,  set 'scan.partitions' = 'ds=max_pt()', then all subpartitions in max 'ds' partition will be loaded.

### DIFF
--- a/docs/content/flink/sql-lookup.md
+++ b/docs/content/flink/sql-lookup.md
@@ -173,6 +173,12 @@ The option `scan.partitions` can also specify fixed partitions in the form of `k
 Multiple partitions should be separated by semicolon (`;`).
 When specifying fixed partitions, this option can also be used in batch joins.
 
+The option `scan.partitions` can also specify max_pt() for parent partition in the form of `key1=max_pt(),key2=max_pt()`.
+All subpartitions for the latest parent partition will be loaded. For example, if partition keys is `'year', 'day', 'hh'`, 
+you can specify `year=max_pt()`, it will find the latest partition for `year` and load all its subpartitions for lookup.
+Only supports partitions to be specified hierarchically. For example, if setting `year=max_pt(),hh=max_pt()`, `hh=max_pt()`
+makes no sense.
+
 ## Query Service
 
 You can run a Flink Streaming Job to start query service for the table. When QueryService exists, Flink Lookup Join
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/DynamicPartitionLevelLoader.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/DynamicPartitionLevelLoader.java
@@ -1,0 +1,192 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.paimon.flink.lookup;
+
+import org.apache.paimon.data.BinaryRow;
+import org.apache.paimon.data.InternalRow;
+import org.apache.paimon.table.FileStoreTable;
+import org.apache.paimon.types.RowType;
+import org.apache.paimon.utils.Preconditions;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.time.Duration;
+import java.time.LocalDateTime;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Collectors;
+
+/** Dynamic partition loader which can specify the partition level to load for lookup. */
+public class DynamicPartitionLevelLoader extends DynamicPartitionLoader {
+
+    private static final Logger LOG = LoggerFactory.getLogger(DynamicPartitionLevelLoader.class);
+
+    private static final long serialVersionUID = 2L;
+
+    private final int maxPartitionLoadLevel;
+    private final List<InternalRow.FieldGetter> fieldGetters;
+
+    DynamicPartitionLevelLoader(
+            FileStoreTable table,
+            Duration refreshInterval,
+            Map<String, String> partitionLoadConfig) {
+        super(table, refreshInterval);
+        maxPartitionLoadLevel =
+                getMaxPartitionLoadLevel(partitionLoadConfig, table.partitionKeys());
+        fieldGetters = createPartitionFieldGetters();
+    }
+
+    @Override
+    public boolean checkRefresh() {
+        if (lastRefresh != null
+                && !lastRefresh.plus(refreshInterval).isBefore(LocalDateTime.now())) {
+            return false;
+        }
+
+        LOG.info(
+                "DynamicPartitionLevelLoader(maxPartitionLoadLevel={},table={}) refreshed after {} second(s), refreshing",
+                maxPartitionLoadLevel,
+                table.name(),
+                refreshInterval.toMillis() / 1000);
+
+        List<BinaryRow> newPartitions = getMaxPartitions();
+        lastRefresh = LocalDateTime.now();
+
+        if (newPartitions.size() != partitions.size()) {
+            partitions = newPartitions;
+            logNewPartitions();
+            return true;
+        } else {
+            for (int i = 0; i < newPartitions.size(); i++) {
+                if (comparator.compare(newPartitions.get(i), partitions.get(i)) != 0) {
+                    partitions = newPartitions;
+                    logNewPartitions();
+                    return true;
+                }
+            }
+            LOG.info(
+                    "DynamicPartitionLevelLoader(maxPartitionLoadLevel={},table={}) didn't find new partitions.",
+                    maxPartitionLoadLevel,
+                    table.name());
+            return false;
+        }
+    }
+
+    @Override
+    protected List<BinaryRow> getMaxPartitions() {
+        List<BinaryRow> newPartitions =
+                table.newReadBuilder().newScan().listPartitions().stream()
+                        .sorted(comparator.reversed())
+                        .collect(Collectors.toList());
+
+        if (maxPartitionLoadLevel == table.partitionKeys().size() - 1) {
+            // if maxPartitionLoadLevel is the max partition level, we only need to load the max
+            // partition
+            if (newPartitions.size() <= 1) {
+                return newPartitions;
+            } else {
+                return newPartitions.subList(0, 1);
+            }
+        }
+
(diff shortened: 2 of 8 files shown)


docs/content/flink/sql-lookup.md
```diff
@@ -173,6 +173,12 @@ The option `scan.partitions` can also specify fixed partitions in the form of `k
 Multiple partitions should be separated by semicolon (`;`).
 When specifying fixed partitions, this option can also be used in batch joins.
 
+The option `scan.partitions` can also specify max_pt() for parent partition in the form of `key1=max_pt(),key2=max_pt()`.
```
yunfengzhou-hub, line 176: In the paragraph above, it said that multiple partitions should be separated by semicolon (`;`). However here comma(`,`) is used. This practice is inconsistent and better unified.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/DynamicPartitionLevelLoader.java
```diff
@@ -0,0 +1,195 @@
...
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Collectors;
+
+/** Dynamic partition loader which can specify the partition level to load for lookup. */
+public class DynamicPartitionLevelLoader extends DynamicPartitionLoader {
+
+    private static final Logger LOG = LoggerFactory.getLogger(DynamicPartitionLevelLoader.class);
+
+    private static final long serialVersionUID = 2L;
```
yunfengzhou-hub: Newly introduced class could have its serialVersionUID starts from 1L.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/DynamicPartitionLevelLoader.java
```diff
@@ -0,0 +1,195 @@
...
+            Duration refreshInterval,
+            Map<String, String> partitionLoadConfig) {
+        super(table, refreshInterval);
+        maxPartitionLoadLevel =
+                getMaxPartitionLoadLevel(partitionLoadConfig, table.partitionKeys());
+        fieldGetters = createPartitionFieldGetters();
+    }
+
+    @Override
+    public boolean checkRefresh() {
```
yunfengzhou-hub: Given that abstract `DynamicPartitionLoader` has been introduced to hold common logics of `DynamicPartitionLevelLoader` and `DynamicPartitionNumberLoader`, it might be better to also move most `checkRefresh` and `logNewPartitions` implementation to that abstract class.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/DynamicPartitionLoader.java
```diff
@@ -25,32 +25,25 @@
...
 import java.util.List;
 import java.util.stream.Collectors;
 
 /** Dynamic partition for lookup. */
-public class DynamicPartitionLoader extends PartitionLoader {
-
-    private static final Logger LOG = LoggerFactory.getLogger(DynamicPartitionLoader.class);
+public abstract class DynamicPartitionLoader extends PartitionLoader {
 
     private static final long serialVersionUID = 2L;
```
yunfengzhou-hub: This serialVersionUID needs to be updated.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/DynamicPartitionLevelLoader.java
```diff
@@ -0,0 +1,195 @@
...
+                        "unsupported load config '{}' for partition field '{}', currently only support '{}'.",
+                        config,
+                        partitionFields.get(i),
+                        MAX_PT);
+                maxLoadLevel = Math.max(maxLoadLevel, i);
+
+            } else {
+                // if level-i partition don't set config, we won't consider the partition level
+                // which is greater than level-i
+                return maxLoadLevel;
```
yunfengzhou-hub: How about throwing exception if `partitionLoadConfig` contains partition levels greater than level-i? This can make Flink jobs fail fast and avoid having users misbelieving that they are using the dynamic partition loader correctly.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/DynamicPartitionLevelLoader.java
```diff
@@ -0,0 +1,195 @@
...
+
+        for (int i = 0; i < maxPartitionLoadLevel + 1; i++) {
+            fieldGetters.add(InternalRow.createFieldGetter(partitionType.getTypeAt(i), i));
+        }
+        return fieldGetters;
+    }
+
+    private List<BinaryRow> extractMaxPartitionsForFixedLevel(
+            List<BinaryRow> partitions, int level) {
+        AtomicInteger currentDistinct = new AtomicInteger(0);
```
yunfengzhou-hub: Why do we use AtomicInteger here? Seems that there is no parallel execution issue here.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/DynamicPartitionLevelLoader.java
```diff
@@ -0,0 +1,195 @@
...
+        Object lastField = null;
+        for (int i = 0; i < partitions.size(); i++) {
+            BinaryRow partition = partitions.get(i);
+            Object newField = fieldGetters.get(level).getFieldOrNull(partition);
+            // if newField is null, it's the default partition
+            if (newField == null) {
+                newField = "__DEFAULT_PARTITION__";
```
yunfengzhou-hub: Might be better to introduce a constant for `"__DEFAULT_PARTITION__"`. Or use `CoreOptions.PARTITION_DEFAULT_NAME.defaultValue()` here.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/DynamicPartitionLevelLoader.java
```diff
@@ -0,0 +1,195 @@
...
+            }
+            if (lastField == null || !lastField.equals(newField)) {
```
yunfengzhou-hub: nit: `!newField.equals(lastField)` is simpler. It avoids NPE.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/PartitionLoader.java
```diff
@@ -128,12 +128,22 @@ public static PartitionLoader of(FileStoreTable table) {
...
         if (maxPartitionNum == -1) {
+            if (scanPartitions.contains(MAX_PT)) {
+                Duration refresh =
+                        options.get(
+                                FlinkConnectorOptions.LOOKUP_DYNAMIC_PARTITION_REFRESH_INTERVAL);
+                return new DynamicPartitionLevelLoader(
+                        table,
+                        refresh,
+                        ParameterUtils.parseCommaSeparatedKeyValues(scanPartitions));
+            }
```
yunfengzhou-hub, line 139: Would it be better to somehow reorganize and improve the readability of this code block? The current organization is seemingly showing the following points:
1. `DynamicPartitionLevelLoader` is closer to `StaticPartitionLoader` than `DynamicPartitionNumberLoader`.
2. When `DynamicPartitionLevelLoader` is used, `maxPartitionNum == -1`. i.e. no max_pt() is used.

Both these points seems weird.

yunfengzhou-hub (commented):

LsomeYeah:
> Thanks for the PR. Left some comments as below.

@yunfengzhou-hub Thank you very much for your detailed review. I have updated the code accordingly. Please review it at your convenience.

docs/content/flink/sql-lookup.md
LsomeYeah: ";" is used to separate multiple partitions while ","is used to separate multi-level partition, i think "," would be better to specify multi-level partition for mat_pt.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/DynamicPartitionLevelLoader.java
```diff
@@ -0,0 +1,166 @@
...
+                            partitionFields.get(i),
+                            j,
+                            partitionFields.get(j));
+                }
+
+                return maxLoadLevel;
+            }
+        }
+        return maxLoadLevel;
+    }
```
tsreaper:
```suggestion
private int getMaxPartitionLoadLevel(Map<String, String> toLoad, List<String> fields) {
        Preconditions.checkArgument(toLoad.size() <= fields.size());

        int maxLoadLevel = fields.size() - 1;
        for (int i = 0; i < fields.size(); i++) {
            if (!toLoad.containsKey(fields.get(i))) {
                maxLoadLevel = i - 1;
                break;
            }
        }

        Preconditions.checkArgument(
                maxLoadLevel >= 0, "the top level partition must set load config.");
        for (int i = maxLoadLevel + 1; i < fields.size(); i++) {
            Preconditions.checkArgument(
                    !toLoad.containsKey(fields.get(i)),
                    "Partition field %s with level %s doesn't set MAX_PT, "
                            + "while partition field %s with a higher level %s sets MAX_PT. This is unsupported.",
                    fields.get(i),
                    i,
                    fields.get(j),
                    j);
        }

        return maxLoadLevel;
    }
```

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/DynamicPartitionLevelLoader.java
```diff
@@ -0,0 +1,166 @@
...
+            newPartitions = extractMaxPartitionsForFixedLevel(newPartitions, level);
+            if (LOG.isDebugEnabled()) {
+                LOG.debug(
+                        "DynamicPartitionLevelLoader(currentPartitionLoadLevel={},table={}) finds new partitions: {}.",
+                        level,
+                        table.name(),
+                        partitionsToString(newPartitions));
+            }
+        }
+        return newPartitions;
```
tsreaper:
```suggestion
        newPartitions = extractMaxPartitionsForFixedLevel(newPartitions, maxPartitionLoadLevel);
        if (LOG.isDebugEnabled()) {
            LOG.debug(
                    "DynamicPartitionLevelLoader(currentPartitionLoadLevel={},table={}) finds new partitions: {}.",
                    level,
                    table.name(),
                    partitionsToString(newPartitions));
        }
        return newPartitions;
```

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/LookupJoinITCase.java
```diff
@@ -1013,6 +1013,47 @@ public void testOverwriteDimTable(boolean isPkTable) throws Exception {
         iterator.close();
     }
 
+    @ParameterizedTest
+    @EnumSource(LookupCacheMode.class)
+    public void testLookupPartitionLevelMaxPt(LookupCacheMode mode) throws Exception {
+        sql(
+                "CREATE TABLE PARTITIONED_DIM (pt1 STRING, pt2 INT, i INT, v INT)"
+                        + "PARTITIONED BY (`pt1`, `pt2`) WITH ("
+                        + "'scan.partitions' = 'pt1=max_pt()', "
```
tsreaper, line 1022: Add another IT case where `scan.partitions` contains multiple partition keys.

tsreaper (approved):

PR #5832, opened 2025-07-02
[core]  One snapshot should have only one auto tag

### LINKED ISSUE #5800: [Bug] Automatic tag could be creaetd on the same snapshot, but reporting error while insertting new data
### Search before asking

### Paimon version

1.1

### Compute Engine

spark

### Minimal reproduce step

```sql
-- create a table with tag.automatic-creation
CREATE TABLE test.test_try_tag (
  id INT NOT NULL,
  val INT)
TBLPROPERTIES (
  'bucket' = '1',
  'primary-key' = 'id',
  'tag.automatic-creation' = 'process-time',
  'tag.creation-delay' = '10m',
  'tag.creation-period' = 'daily',
  'tag.num-retained-max' = '90'
);

-- insert value normally
insert into test.test_try_tag_951 values(1,1);

-- There should be a tag created automatically.
select * from test.`test_try_tag$tags`;

-- create another tag on the same snapshot
alter table test.test_try_tag create tag `2025-06-01`;

-- There should be 2 tags on the same snapshot: 1.
select * from test.`test_try_tag$tags`;

-- try to insert new value, exception reported 
insert into test_try_tag values(2, 1); 
```

Image

The 2nd tag should not be created since the same snapshot has already been tagged.

### Anything else?

_No response_

JackeyLee007:
### Purpose

Linked issue: close #5800

Prevent different auto-tags created on the same snapshot.
If not, inserting data would incur errors as the issue #5800 shows.

### Tests

org.apache.paimon.tag.TagAutoManagerTest#testMultiAutoTagsOnOneSnapshot

### Documentation

No

### DIFF
--- a/docs/content/spark/sql-ddl.md
+++ b/docs/content/spark/sql-ddl.md
@@ -340,7 +340,7 @@ ALTER TABLE T REPLACE TAG `TAG-4` AS OF VERSION 2 RETAIN 24 HOURS;
 -- create or replace a tag, create tag if it not exist, replace tag if it exists.
 ALTER TABLE T CREATE OR REPLACE TAG `TAG-5` AS OF VERSION 2 RETAIN 24 HOURS;
 ```
-> NOTE: If the table's property tag.automatic-creation is not none, which is default, only one auto-tag could be created for one snapshot. 
+NOTE: If tag.automatic-creation is set, only one auto-tag could be created for one snapshot.
 
 ### Delete Tag
 Delete a tag or multiple tags of a table.
--- a/docs/content/spark/sql-ddl.md
+++ b/docs/content/spark/sql-ddl.md
@@ -340,6 +340,7 @@ ALTER TABLE T REPLACE TAG `TAG-4` AS OF VERSION 2 RETAIN 24 HOURS;
 -- create or replace a tag, create tag if it not exist, replace tag if it exists.
 ALTER TABLE T CREATE OR REPLACE TAG `TAG-5` AS OF VERSION 2 RETAIN 24 HOURS;
 ```
+> NOTE: If the table's property tag.automatic-creation is not none, which is default, only one auto-tag could be created for one snapshot. 
 
 ### Delete Tag
 Delete a tag or multiple tags of a table.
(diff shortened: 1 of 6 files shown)


docs/content/spark/sql-ddl.md
```diff
@@ -340,6 +340,7 @@ ALTER TABLE T REPLACE TAG `TAG-4` AS OF VERSION 2 RETAIN 24 HOURS;
 -- create or replace a tag, create tag if it not exist, replace tag if it exists.
 ALTER TABLE T CREATE OR REPLACE TAG `TAG-5` AS OF VERSION 2 RETAIN 24 HOURS;
 ```
+> NOTE: If the table's property tag.automatic-creation is not none, which is default, only one auto-tag could be created for one snapshot.
```
yuzelin:
```suggestion
NOTE: If tag auto-creation is set, only one auto-tag could be created for one snapshot. 
```

paimon-core/src/main/java/org/apache/paimon/utils/TagManager.java
```diff
@@ -64,20 +67,26 @@ public class TagManager {
...
-        this(fileIO, tablePath, DEFAULT_MAIN_BRANCH);
+        this(fileIO, tablePath, DEFAULT_MAIN_BRANCH, CoreOptions.fromMap(Collections.emptyMap()));
     }
 
-    /** Specify the default branch for data writing. */
     public TagManager(FileIO fileIO, Path tablePath, String branch) {
+        this(fileIO, tablePath, branch, CoreOptions.fromMap(Collections.emptyMap()));
+    }
+
+    /** Specify the default branch for data writing. */
```
yuzelin, line 81: No need to move this comment.

paimon-core/src/main/java/org/apache/paimon/utils/TagManager.java
```diff
@@ -437,4 +457,24 @@ private int findIndex(Snapshot taggedSnapshot, List<Snapshot> taggedSnapshots) {
                         "Didn't find tag with snapshot id '%s'. This is unexpected.",
                         taggedSnapshot.id()));
     }
+
+    private boolean isAutoTag(String tagName) {
+        TagPeriodHandler periodHandler = TagPeriodHandler.create(coreOptions);
```
yuzelin, line 471: We should test if `TagAutoCreation` is set, then we can test isAutoTag.

PR #5836, opened 2025-07-03
[flink][core] Flink supports setting job level context to FileIO 

yuzelin:
### Purpose

Many cloud service companies provide 'Fully Managed Service' for flink. Users can deploy flink jobs and start them with custom flink conf. 

However, our FileIO is initialized from catalog options and won't be changed when job starting and running. But sometimes users may put different filesystem options on different jobs, so I suggest let FileIO refresh with these options.

Linked issue: close #xxx

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/fs/FileIO.java
+++ b/paimon-common/src/main/java/org/apache/paimon/fs/FileIO.java
@@ -73,13 +73,8 @@ public interface FileIO extends Serializable, Closeable {
     /** Configure by {@link CatalogContext}. */
     void configure(CatalogContext context);
 
-    /**
-     * Set dynamic filesystem options and construct a new FileIO. Usually used for job-level
-     * settings.
-     */
-    default FileIO copy(Map<String, String> dynamicOptions) {
-        return this;
-    }
+    /** Set filesystem options at runtime. Usually used for job-level settings. */
+    default void setRuntimeContext(Map<String, String> options) {}
 
     /**
      * Opens an SeekableInputStream at the indicated Path.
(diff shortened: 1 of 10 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/AbstractFlinkTableFactory.java
```diff
@@ -225,7 +225,12 @@ Table buildPaimonTable(DynamicTableFactory.Context context) {
                 throw new RuntimeException(e);
             }
         }
-        table = fileStoreTable.copyWithoutTimeTravel(newOptions);
+
+        FileStoreTable table = fileStoreTable.copyWithoutTimeTravel(newOptions);
+
+        if (Options.fromMap(table.options()).get(FILESYSTEM_JOB_LEVEL_SETTINGS_ENABLED)) {
+            table = table.copyFileIODynamicOptions(getAllOptions(context));
```
JingsongLi: `table.fileIO().setRuntimeContext(Map<String, String> options);`

paimon-common/src/main/java/org/apache/paimon/fs/FileIO.java
```diff
@@ -73,6 +73,14 @@ public interface FileIO extends Serializable, Closeable {
     /** Configure by {@link CatalogContext}. */
     void configure(CatalogContext context);
 
+    /**
+     * Set dynamic filesystem options and construct a new FileIO. Usually used for job-level
+     * settings.
+     */
+    default FileIO copy(Map<String, String> dynamicOptions) {
```
JingsongLi: `default void setRuntimeContext(Map<String, String> options) {}`

JingsongLi (approved):
+1

PR #5838, opened 2025-07-03
[flink] Allow ignoring bucket configuration during read.

zhuanshenbsj1:
### Purpose

For append tables, even if  fixed-bucket index is used,  there is no need to read data according to the fixed bucket during reading. Otherwise,  the performance will be affected due to the too few number of upstream table buckets.

Add an optional configuration to allow ignoring bucket settings when reading data.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
+++ b/paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
@@ -537,7 +537,7 @@ public class SchemaValidation {
         if (bucket == -1) {
             if (options.toMap().get(BUCKET_KEY.key()) != null) {
                 throw new RuntimeException(
-                        "Cannot define 'bucket-key' with bucket -1, please specify a bucket number.");
+                        "Cannot define 'bucket-key' with bucket = -1, please remove the 'bucket-key' setting or specify a bucket number.");
             }
 
             if (schema.primaryKeys().isEmpty()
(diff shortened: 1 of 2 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/FlinkSourceBuilder.java
```diff
@@ -97,11 +96,9 @@ public class FlinkSourceBuilder {
 
     public FlinkSourceBuilder(Table table) {
         this.table = table;
-        this.unawareBucket =
-                table instanceof FileStoreTable
-                        && ((FileStoreTable) table).bucketMode() == BucketMode.BUCKET_UNAWARE;
         this.sourceName = table.name();
         this.conf = Options.fromMap(table.options());
+        this.unawareBucket = !this.conf.get(CoreOptions.BUCKET_APPEND_ORDERD);
```
wwj6591812: Why remove the original judgment logic？

JingsongLi: +1, here should be bug...

zhuanshenbsj1: > Why remove the original judgment logic？

As the original logic， append tables that use fixed-buckets still distribute splits according to the fixed buckets during downstream streaming reads . This is unnecessary because append  tables do not require deduplication，and this  will causes the number of buckets to limit the downstream read concurrency. Even if the number of readers is increased, there will be no performance improvement.

Therefore, I added a configuration option to allow downstream streaming reads to ignore the upstream table bucket settings.

PR #5843, opened 2025-07-04
[core] Paimon core should import the correct package

leaves12138:
### Purpose

Some package import paths are wrong. Paimon core could be compiled due to maven bugs, it should be be compiled. Because those imports (which I repair in this pr) are not existing in its dependencies.


JingsongLi:
Can we add check in tools/maven/checkstyle.xml just like guava?

JingsongLi:
+1

PR #5845, opened 2025-07-07
[spark] Fix reading CSV Format Table with field-delimiter

JingsongLi:
### Purpose

This PR only fix reading, cannot fix writing.

For `PartitionedCSVTable`, Spark uses Fallback format to write csv files. So it is very hard to fix writing in Paimon side.


Zouxxyy (approved):
+1

PR #5846, opened 2025-07-07
[core] Support mod bucket function

### LINKED ISSUE #3538: [Feature] Introduce mod bucket generator to paimon

luoyuxia:
### Purpose

Linked issue: close #3538 

Add the support for `mod` bucket function
Introduce a new `ModBucketFunction`

### Tests

`ModBucketFunctionTest`

### DIFF
--- a/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
@@ -132,7 +132,9 @@ public class CoreOptions implements Serializable {
 
     /** Paimon bucket function type. */
     public enum BucketFunctionType implements DescribedEnum {
-        DEFAULT("default", "The default bucket function."),
+        DEFAULT(
+                "default",
+                "The default bucket function which will use arithmetic: bucket_id = Math.abs(hash_bucket_binary_row % numBuckets) to get bucket."),
         MOD(
                 "mod",
                 "The modulus bucket function which will use modulus arithmetic: bucket_id = Math.floorMod(bucket_key_value, numBuckets) to get bucket. "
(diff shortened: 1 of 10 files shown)


paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -128,11 +128,43 @@ public class CoreOptions implements Serializable {
...
                     .enumType(BucketFunctionType.class)
                     .defaultValue(BucketFunctionType.DEFAULT)
-                    .withDescription("The bucket function for paimon bucket");
+                    .withDescription("The bucket function for paimon bucket.");
 
     /** Paimon bucket function type. */
-    public enum BucketFunctionType {
-        DEFAULT,
+    public enum BucketFunctionType implements DescribedEnum {
+        DEFAULT("default", "The default bucket function."),
```
JingsongLi: Document: bucket = Math.abs(hash_bucket_binary_row % numBuckets) to get bucket.

paimon-core/src/main/java/org/apache/paimon/bucket/BucketFunction.java
```diff
@@ -39,6 +39,8 @@ static BucketFunction create(
         switch (bucketFunctionType) {
             case DEFAULT:
                 return new PaimonBucketFunction();
```
JingsongLi: Rename to DefaultBucketFunction, and add serialVersionUID

JingsongLi:
+1

PR #5848, opened 2025-07-08
[core] Redesign Object Table to a Virtual table

JingsongLi:
### Purpose

After the object table was developed, almost no one used it. Object Table is a real Paimon table, it requires a refresh service to refresh the metadata of files, which is difficult to complete in operation.

In fact, we don't need to refresh it, and accessing the file system directly during reading is not a bad thing.

This discussion discusses whether it is possible to transform an object table into a virtual existence. Users can:

1. Use the virtual file system to read and write files.
2. Or use the SQL computing engine to read it as a structured file list.

On this basis, we can develop virtual file systems, such as through similar approaches vfs://catalog_name/database_name/table_name/ Use the path to access files in the specific object table.

By using this method, we can integrate file operations on object tables with the permission system of data catalogs.

### DIFF
--- a/docs/content/concepts/table-types.md
+++ b/docs/content/concepts/table-types.md
@@ -230,7 +230,8 @@ Object Table is a virtual table for unstructured data objects in the specified o
 1. Use the virtual file system (Under development) to read and write files.
 2. Or use the SQL computing engine to read it as a structured file list.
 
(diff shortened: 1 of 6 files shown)


timmyyao:
+1.

PR #5849, opened 2025-07-08
[core] Copy max key in KeyValueDataFileWriter

JingsongLi:
### Purpose

We should copy the maxKey, otherwise if the reader is closed first, the maxKey inside may be reset, such as the dictionary in Parquet Reader.

### DIFF
--- a/paimon-arrow/src/main/java/org/apache/paimon/arrow/reader/ArrowBatchReader.java
+++ b/paimon-arrow/src/main/java/org/apache/paimon/arrow/reader/ArrowBatchReader.java
@@ -23,7 +23,6 @@ import org.apache.paimon.data.InternalRow;
 import org.apache.paimon.data.columnar.ColumnVector;
 import org.apache.paimon.data.columnar.ColumnarRow;
 import org.apache.paimon.data.columnar.VectorizedColumnBatch;
-import org.apache.paimon.data.serializer.InternalRowSerializer;
 import org.apache.paimon.types.DataField;
 import org.apache.paimon.types.RowType;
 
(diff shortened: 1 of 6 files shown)


leaves12138:
We can remove TODO in:
image

paimon-arrow/src/main/java/org/apache/paimon/arrow/reader/ArrowBatchReader.java
```diff
@@ -93,11 +93,7 @@ public boolean hasNext() {
                     public InternalRow next() {
                         columnarRow.setRowId(position);
                         position++;
-                        // when pk merge, the last value will be referenced by maxKey. If reader
-                        // close, the maxKey will be useless. We must avoid this situation
-                        return position == rowCount
-                                ? internalRowSerializer.toBinaryRow(columnarRow)
```
yuzelin, line 99: Remove class member `internalRowSerializer`.

paimon-common/src/main/java/org/apache/paimon/data/RowHelper.java
```diff
@@ -0,0 +1,100 @@
...
+import org.apache.paimon.types.DataTypeChecks;
+import org.apache.paimon.types.DecimalType;
+import org.apache.paimon.types.LocalZonedTimestampType;
+import org.apache.paimon.types.TimestampType;
+
+import java.io.Serializable;
+import java.util.List;
+
+/** A keeper to keep {@link BinaryRow} from {@link InternalRow}. */
+public class RowHelper implements Serializable {
```
yuzelin, line 35: BinaryRowHelper/Reuser ?

JingsongLi: It also provide setter and getter and serializer... So not just for `BinaryRow`.

PR #5854, opened 2025-07-08
Add git-commit-id-plugin

cxzl25:
### Purpose

Linked issue: close #xxx

https://github.com/apache/paimon/commit/025c032441873df1e20dc1896769f325c8703095#commitcomment-161263081

### Tests

Master
image

PR
image

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -479,6 +479,7 @@ under the License.
                 <configuration>
                     <archive>
                         <manifestEntries>
+                            <!-- The SCM revision is set by the git-commit-id-plugin -->
                             <SCM-Revision>${git.commit.id}</SCM-Revision>
                         </manifestEntries>
                         <manifest>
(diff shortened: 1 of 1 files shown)


cxzl25:
> but many developers have reported that compilation will be much slower

Tested, it was a little different.  M2 Max

```bash
time mvn clean package -DskipTests
```

master
```bash
506.56s user 62.62s system 178% cpu 5:18.58 total
```
PR
```bash
513.05s user 76.75s system 97% cpu 10:06.45 total
```

---

> removed commit id

So can we remove these lines? It seems that in the new version of release there is no commit id

https://github.com/apache/paimon/blob/7f6e5747f2a2bdcb10d58f84684875b8129169f9/pom.xml#L481-L483

JingsongLi:
> So can we remove these lines? It seems that in the new version of release there is no commit id

Yes, we should remove these lines.

PR #5857, opened 2025-07-09
[core] Enable Auto Spill in Streaming mode

JingsongLi:
### Purpose

In order to reduce memory consumption or frequent interaction with DFS in streaming mode, we also apply Spill (And Auto Spill) to streaming mode.


leaves12138 (approved):
+1

PR #5859, opened 2025-07-09
[rest] Introduce register api definition in rest

qingwei727:
### Purpose
In paimon catalog scene, sometimes we need to register metadata of a data table into the Paimon Catalog for centralized management and access.


docs/static/rest-catalog-open-api.yaml
```diff
@@ -206,6 +206,45 @@ paths:
           $ref: '#/components/responses/DatabaseNotExistErrorResponse'
         "500":
           $ref: '#/components/responses/ServerErrorResponse'
+  /v1/{prefix}/databases/{database}/register:
```
JingsongLi, line 209: /v1/{prefix}/databases/{database}/tables/{table}/register?

paimon-api/src/main/java/org/apache/paimon/rest/requests/RegisterTableRequest.java
```diff
@@ -0,0 +1,59 @@
...
+public class RegisterTableRequest implements RESTRequest {
+
+    private static final String FIELD_IDENTIFIER = "identifier";
+    private static final String FIELD_LOCATION = "location";
+
+    @JsonProperty(FIELD_IDENTIFIER)
+    private final Identifier identifier;
+
+    @JsonProperty(FIELD_LOCATION)
+    private final String location;
```
JingsongLi: path

paimon-api/src/main/java/org/apache/paimon/rest/requests/RegisterTableRequest.java
```diff
@@ -0,0 +1,59 @@
...
+
+/** Request for register table. */
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class RegisterTableRequest implements RESTRequest {
+
+    private static final String FIELD_IDENTIFIER = "identifier";
+    private static final String FIELD_LOCATION = "location";
+
+    @JsonProperty(FIELD_IDENTIFIER)
+    private final Identifier identifier;
```
JingsongLi, line 38: we don't need identifier in this class.

paimon-api/src/main/java/org/apache/paimon/rest/responses/RegisterTableResponse.java
```diff
@@ -0,0 +1,48 @@
...
+
+import org.apache.paimon.shade.jackson2.com.fasterxml.jackson.annotation.JsonCreator;
+import org.apache.paimon.shade.jackson2.com.fasterxml.jackson.annotation.JsonGetter;
+import org.apache.paimon.shade.jackson2.com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import org.apache.paimon.shade.jackson2.com.fasterxml.jackson.annotation.JsonInclude;
+import org.apache.paimon.shade.jackson2.com.fasterxml.jackson.annotation.JsonProperty;
+
+/** Response for register table. */
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class RegisterTableResponse implements RESTResponse {
```
JingsongLi: Just like create table, no response?

docs/static/rest-catalog-open-api.yaml
qingwei727: Registering a table is inherently a database-level action rather than table-level, so maybe it is more suitable to locate it at the database level？

paimon-api/src/main/java/org/apache/paimon/rest/requests/RegisterTableRequest.java
qingwei727: May be simplify to tableName?

paimon-api/src/main/java/org/apache/paimon/rest/responses/RegisterTableResponse.java
qingwei727: sure

docs/static/rest-catalog-open-api.yaml
JingsongLi: ok

paimon-api/src/main/java/org/apache/paimon/rest/requests/RegisterTableRequest.java
JingsongLi: Just keep it as it is identifier

JingsongLi (approved):
+1

PR #5866, opened 2025-07-10
[core][flink] Streaming first plan refactor level0 filtering to avoid that full mode wrong

yuzelin:
### Purpose

Reproduce steps：
```
CREATE TABLE test (a INT, b INT) WITH ('deletion-vectors.enabled' = 'true');
INSERT INTO test VALUES (1, 1);
INSERT INTO test VALUES (2, 2);
-- then do compact --

SELECT * FROM test /*+ OPTIONS('scan.mode' = 'latest-full') */
```
Cannot get data because level0 file is skipped and append table only has level0 files. Then it step into nextPlan, but compact snapshot won't be read.

Then if:
```
INSERT INTO test VALUES (3, 3);
```
SELECT result is (3, 3) without full data because StreamScan move to nextPlan to read DELTA data.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/source/DataTableStreamScan.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/source/DataTableStreamScan.java
@@ -22,7 +22,6 @@ import org.apache.paimon.CoreOptions;
 import org.apache.paimon.CoreOptions.StreamScanMode;
 import org.apache.paimon.Snapshot;
 import org.apache.paimon.consumer.Consumer;
-import org.apache.paimon.lookup.LookupStrategy;
 import org.apache.paimon.manifest.PartitionEntry;
 import org.apache.paimon.predicate.Predicate;
 import org.apache.paimon.schema.TableSchema;


paimon-core/src/main/java/org/apache/paimon/table/source/DataTableStreamScan.java
```diff
@@ -204,6 +195,15 @@ private Plan tryFirstPlan() {
         return SnapshotNotExistPlan.INSTANCE;
     }
 
+    private boolean skipLevel0() {
+        LookupStrategy lookupStrategy = options.lookupStrategy();
+        if (!lookupStrategy.produceChangelog && lookupStrategy.deletionVector) {
+            // Read level0 data for DELETION VECTOR ONLY mode
```
JingsongLi: Why we need to change behavior of pk table with dv?

yuzelin: Because there is the same problem for pk table with `'deletion-vectors.enabled' = 'true', 'changelog-producer' = 'none'`. Test is added to `DeletionVectorITCase`.

paimon-core/src/main/java/org/apache/paimon/table/source/DataTableStreamScan.java
```diff
@@ -157,7 +157,7 @@ private Plan tryFirstPlan() {
         StartingScanner.Result result;
         if (scanMode == FILE_MONITOR) {
             result = startingScanner.scan(snapshotReader);
-        } else if (options.needLookup()) {
+        } else if (skipLevel0()) {
```
JingsongLi: options.changelogProducer().equals(LOOKUP)

JingsongLi (approved):
+1

PR #5867, opened 2025-07-10
[flink] Partition predicate builder push down to core, avoid stack overflow in building partition predicates.

leaves12138:
### Purpose

Avoid List<Map<String, String>> partitions convert to class Predicate.


JingsongLi:
Use `PartitionPredicate` instead of `Either`.

leaves12138:
see https://github.com/apache/paimon/pull/5880

PR #5881, opened 2025-07-11
[spark] SparkGenericCatalog support default value

Zouxxyy:
### Purpose

SparkGenericCatalog can enable default value like SparkCatalog


JingsongLi:
+1

PR #5882, opened 2025-07-11
[core] python: add test for rest api

jerry-024:
### Purpose

python: add test for rest api

### Tests
- tests.api_test.ApiTestCase.test


JingsongLi:
Looks good to me!

PR #5892, opened 2025-07-14
[core] Validate retract records in PostponeBucketWriter

JingsongLi:
### Purpose

Throw an exception earlier.


paimon-core/src/main/java/org/apache/paimon/postpone/PostponeBucketWriter.java
```diff
@@ -112,6 +117,14 @@ public void write(KeyValue record) throws Exception {
         }
     }
 
+    private void validateRetract(KeyValue kv) {
+        if (kv.valueKind().isRetract()) {
+            mergeFunction.reset();
+            mergeFunction.add(kv);
+            mergeFunction.getResult();
+        }
```
tsreaper: Add a class member `validated` so we only need to validate once.

PR #5893, opened 2025-07-14
[core] formatTable supports defaultValue

XiaoHongbo-Hope:
Purpose
formatTable supports defaultValue

### Tests
DDLWithHiveCatalogTestBase.test("Paimon DDL with hive catalog: default value")

### DIFF
--- a/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
+++ b/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
@@ -56,6 +56,7 @@ import org.apache.paimon.utils.PartitionPathUtils;
 import org.apache.paimon.view.View;
 import org.apache.paimon.view.ViewImpl;
 
+import org.apache.paimon.shade.guava30.com.google.common.base.Joiner;
 import org.apache.paimon.shade.guava30.com.google.common.collect.Lists;
 
 import org.apache.flink.table.hive.LegacyHiveClasses;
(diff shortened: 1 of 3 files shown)


paimon-spark/paimon-spark-ut/src/test/scala/org/apache/paimon/spark/sql/DDLWithHiveCatalogTestBase.scala
```diff
@@ -747,11 +747,9 @@ abstract class DDLWithHiveCatalogTestBase extends PaimonHiveTestBase {
 
               // test non paimon table
               // todo: support default value for paimon format table
```
Zouxxyy, line 749: We can delete this line

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
```diff
@@ -1071,6 +1073,13 @@ private Table createHiveFormatTable(
         FormatTable.Format provider = FormatTable.parseFormat(coreOptions.formatType());
 
         Map<String, String> tblProperties = new HashMap<>();
+        tableSchema.fields().stream()
```
Zouxxyy, line 1076: This method maybe can be placed in more general place, such as alter table, which may also require this logic

XiaoHongbo-Hope: > This method maybe can be placed in more general place, such as alter table, which may also require this logic

Currently, formatTable does not support alter table, maybe we can support it in next PR.

JingsongLi:
Currently, we don't have requirement to support default value for format table.

XiaoHongbo-Hope:
> Currently, we don't have requirement to support default value for format table.

ok, thanks

PR #5894, opened 2025-07-15
[bug] Log the right name of the deleting tag.

JackeyLee007:
### Purpose

Just to correct the right information when deleting outdated snapshot.


JingsongLi (approved):
+1

PR #5897, opened 2025-07-15
[flink] fix that AppendBypassCoordinateOperator does not generate compaction tasks.

### LINKED ISSUE #5896: [Bug] AppendBypassCoordinateOperator does not generate compaction tasks.
### Search before asking

### Paimon version

1.0

### Compute Engine

flink 1.18

### Minimal reproduce step

Once an exception occurs in method `AppendBypassCoordinateOperator#asyncPlan`, it will not be scheduled for execution again.

The Flink job should fail in this scenario. Otherwise, the issue will go unnoticed by users, preventing small files from being compacted and ultimately resulting in an excessive number of small files on the storage system.

### Anything else?

_No response_

zhoulii:
### Purpose

Linked issue: close #5896

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/AppendBypassCoordinateOperator.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/AppendBypassCoordinateOperator.java
@@ -23,6 +23,7 @@ import org.apache.paimon.append.AppendCompactTask;
 import org.apache.paimon.flink.utils.RuntimeContextUtils;
 import org.apache.paimon.table.FileStoreTable;
 import org.apache.paimon.utils.ExecutorUtils;
+import org.apache.paimon.utils.Preconditions;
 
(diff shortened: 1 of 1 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/AppendBypassCoordinateOperator.java
```diff
@@ -85,10 +87,15 @@ public void open() throws Exception {
...
-                break;
+            try {
+                List<AppendCompactTask> tasks = coordinator.run();
+                compactTasks.addAll(tasks);
+                if (tasks.isEmpty()) {
+                    break;
+                }
+            } catch (Throwable t) {
+                LOG.error("Fatal exception happened when generating compaction tasks.", t);
+                this.throwable = t;
```
JingsongLi, line 98: We should return here?

zhoulii: > We should return here?

fixed.

JingsongLi (approved):
+1

PR #5900, opened 2025-07-15
add test for data type

jerry-024:
### Purpose
[python] api: add test for data type and refactor json parse

### Tests
- ApiTestCase.test_parse_data


JingsongLi (approved):
+1

PR #5905, opened 2025-07-15
[variant][spark] Add support for read and write shredding

Zouxxyy:
### Purpose

Subtask of https://github.com/apache/paimon/issues/4471

usage:

```sql
CREATE TABLE T (id INT, v VARIANT)
TBLPROPERTIES
('parquet.variant.shreddingSchema' =
'{"type":"ROW","fields":[{"name":"v","type":{"type":"ROW","fields":[{"name":"age","type":"INT"},{"name":"city","type":"STRING"}]}}]}'
);

INSERT INTO T VALUES parse_json('{"age":27,"city":"Beijing"};

SELECT variant_get(v, '$.age', 'int') FROM T ORDER BY id;
```

### DIFF
--- a/paimon-format/src/main/java/org/apache/paimon/format/parquet/reader/ParquetColumnVector.java
+++ b/paimon-format/src/main/java/org/apache/paimon/format/parquet/reader/ParquetColumnVector.java
@@ -78,8 +78,8 @@ public class ParquetColumnVector {
         if (column.variantFileType().isPresent()) {
             ParquetField fileContentCol = column.variantFileType().get();
             WritableColumnVector fileContent =
-                    ParquetSplitReaderUtil.createWritableColumnVector(
-                            capacity, fileContentCol.getType(), fileContentCol.getParquetType(), 0);
+                    ParquetReaderUtil.createWritableColumnVector(
+                            capacity, fileContentCol.getType());
             ParquetColumnVector contentVector =
                     new ParquetColumnVector(
                             fileContentCol, fileContent, capacity, missingColumns, false);
--- a/paimon-format/src/main/java/org/apache/paimon/format/parquet/ParquetReaderFactory.java
+++ b/paimon-format/src/main/java/org/apache/paimon/format/parquet/ParquetReaderFactory.java
@@ -56,8 +56,8 @@ import java.util.List;
 import static org.apache.paimon.format.parquet.ParquetSchemaConverter.PAIMON_SCHEMA;
 import static org.apache.paimon.format.parquet.ParquetSchemaConverter.parquetListElementType;
 import static org.apache.paimon.format.parquet.ParquetSchemaConverter.parquetMapKeyValueType;
-import static org.apache.paimon.format.parquet.reader.ParquetSplitReaderUtil.buildFieldsList;
-import static org.apache.paimon.format.parquet.reader.ParquetSplitReaderUtil.createWritableColumnVector;
+import static org.apache.paimon.format.parquet.reader.ParquetReaderUtil.buildFieldsList;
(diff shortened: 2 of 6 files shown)


JingsongLi:
See https://github.com/delta-io/delta/blob/master/protocol_rfcs/variant-shredding.md

JingsongLi:
Can we just also introduce a `variant-shredding.enabled` and just support reading shredding data?

paimon-api/src/main/java/org/apache/paimon/types/VariantType.java
```diff
@@ -32,6 +34,8 @@ public class VariantType extends DataType {
 
     private static final String FORMAT = "VARIANT";
 
+    @Nullable private RowType shreddingSchema;
```
JingsongLi: Don't add this into type definition.

JingsongLi:
cc @leaves12138 to take a look to parquet part.

leaves12138:
As for read, can we follow spark?

paimon-common/src/main/java/org/apache/paimon/data/variant/GenericVariantBuilder.java
```diff
@@ -274,7 +274,7 @@ public void appendTimestampNtz(long microsSinceEpoch) {
     public void appendFloat(float f) {
         checkCapacity(1 + 4);
         writeBuffer[writePos++] = primitiveHeader(FLOAT);
-        writeLong(writeBuffer, writePos, Float.floatToIntBits(f), 8);
+        writeLong(writeBuffer, writePos, Float.floatToIntBits(f), 4);
```
Zouxxyy, line 277: Port [SPARK-52833](https://issues.apache.org/jira/browse/SPARK-52833)

paimon-format/src/main/java/org/apache/paimon/format/parquet/reader/ParquetColumnVector.java
```diff
@@ -67,7 +75,20 @@ public class ParquetColumnVector {
             return;
         }
 
-        if (isPrimitive) {
+        if (column.variantFileType().isPresent()) {
+            ParquetField fileContentCol = column.variantFileType().get();
+            WritableColumnVector fileContent =
+                    ParquetSplitReaderUtil.createWritableColumnVector(
```
Zouxxyy: Here I need a clean `createWritableColumnVector` method. like Spark's: `public OffHeapColumnVector(int capacity, DataType type)` 

But this method in paimon seems to require parquetType, therefore, I have no choice but to introduce the `parquetType` type in `ParquetField`.

paimon-format/src/main/java/org/apache/paimon/format/parquet/reader/ParquetSplitReaderUtil.java
```diff
@@ -51,98 +52,50 @@
...
 import static org.apache.parquet.schema.Type.Repetition.REQUIRED;
 
 /** Util for generating parquet readers. */
 public class ParquetSplitReaderUtil {
 
     public static WritableColumnVector createWritableColumnVector(
-            int batchSize,
-            DataType fieldType,
-            Type type,
-            List<ColumnDescriptor> columnDescriptors,
```
Zouxxyy, line 81: `columnDescriptors` looks confused, and only used in validate so just remove it

paimon-format/src/main/java/org/apache/paimon/format/parquet/reader/ParquetColumnVector.java
Zouxxyy: Addressed, let's follow Spark, slim down the `createWritableColumnVector`

leaves12138 (approved):
+1

PR #5906, opened 2025-07-15
[core]  fix read duplicate records where table is PRIMARY table

### LINKED ISSUE #5874: [Bug] Enable deletion vector may lead to the wrong result
### Search before asking

### Paimon version

1.2

### Compute Engine

all

### Minimal reproduce step

```
    @Test
    public void testEnableDV() {
        sql(
                "CREATE TABLE T (a INT, b INT, PRIMARY KEY (a) NOT ENFORCED) WITH ("
                        + "'changelog-producer' = 'lookup', "
                        + "'bucket' = '1')");
        sql("INSERT INTO T VALUES (1, 2), (2, 2), (3, 3)");
        sql("INSERT INTO T VALUES (1, 4)");
        sql("INSERT INTO T VALUES (1, 5)");
        sql("INSERT INTO T VALUES (1, 6)");
        sql("INSERT INTO T VALUES (1, 7)");
        sql("ALTER TABLE T SET ('deletion-vectors.enabled' = 'true')");
        List<Row> rows = sql("SELECT * FROM T");
        System.out.println(rows);
    }
```

The result is 

```
[+I[1, 6], +I[2, 2], +I[3, 3], +I[1, 7]]
```

.

### Anything else?

_No response_

WenDing-Y:
### Purpose

Linked issue: close #xxx

 fix  #5874

Actually, it has nothing to do with the deletion vector parameter.

The root cause is that the generated Parquet files contain records with the same ID, but the generated split information is incorrect. It selected new SplitGroup(files, true), which eventually led to using RawFileSplitReadProvider to read the data. This reader does not merge duplicate records.
```
@Test
    public void testEnableDV() {
        sql(
                "CREATE TABLE T (a INT, b INT, PRIMARY KEY (a) NOT ENFORCED) WITH ("
                        + "'changelog-producer' = 'lookup', "
                        + "'bucket' = '1')");
        sql("INSERT INTO T VALUES (1, 2), (2, 2), (3, 3)");
        sql("INSERT INTO T VALUES (1, 4)");
        sql("INSERT INTO T VALUES (1, 5)");
        sql("INSERT INTO T VALUES (1, 6)");
        sql("INSERT INTO T VALUES (1, 7)");
        sql("ALTER TABLE T SET ('deletion-vectors.enabled' = 'true')");
        List<Row> rows = sql("SELECT * FROM T");
        System.out.println(rows);
    }
```

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/source/MergeTreeSplitGenerator.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/source/MergeTreeSplitGenerator.java
@@ -72,7 +72,11 @@ public class MergeTreeSplitGenerator implements SplitGenerator {
         boolean oneLevel =
                 files.stream().map(DataFileMeta::level).collect(Collectors.toSet()).size() == 1;
 
-        if (rawConvertible && (deletionVectorsEnabled || mergeEngine == FIRST_ROW || oneLevel)) {
+        boolean noKeyOverlap = !hasIntersectedKeyRange(files);
+
+        if (rawConvertible
+                && noKeyOverlap
+                && (deletionVectorsEnabled || mergeEngine == FIRST_ROW || oneLevel)) {
             Function<DataFileMeta, Long> weightFunc =
                     file -> Math.max(file.fileSize(), openFileCost);
             return BinPacking.packForOrdered(files, weightFunc, targetSplitSize).stream()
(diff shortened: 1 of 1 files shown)


WenDing-Y:
ould like to verify if this behavior constitutes a bug. Specifically, when records with the same ID are present in different Parquet files, duplicate results appear during querying. Is the distribution of identical IDs across multiple Parquet files a random occurrence? 

@shyjsarah

JingsongLi:
Currently, we don't support changing to dv mode without full compaction.

PR #5909, opened 2025-07-16
[spark] Support create auto tag syntax

JackeyLee007:
### Purpose

Support the spark sql syntax as "alter table t create auto tag".
1. No need providing a tag name.
2. An auto tag would be created on the latest snapshot  if the table is an auto-tag table and the target tag doesn't exist.  
3. Nothing would happen if the target tag alredy existed.
4. Also nothing would happen if this table is not an auto-tag table.

### Tests

org.apache.paimon.spark.PaimonSparkTestBase#test("Tag ddl: alter table t create auto tag syntax")
1. First create a non auto-tag table
2. Insert data and no tag was created
3. Alter table to auto-tag with setting tblproperties 
4. Alter table T create auto tag, then an auto-tag would be created

### Documentation

Yes.


JingsongLi (commented):
It seems this should be a procedure: `trigger_tag_automatic_creation`?

JackeyLee007:
> It seems this should be a procedure: `trigger_tag_automatic_creation`?

How to tell the difference? Why is the procedure preferred?

yuzelin (approved):
+1

PR #5911, opened 2025-07-16
[flink] Add clone from Paimon tables in clone

JingsongLi:
### Purpose

Previously, clones only supported Hive tables, now support for Paimon sources has been added.

### Tests

`CloneActionITCase.testCloneFromPaimon`


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/clone/spits/ListCloneSplitsFunction.java
```diff
@@ -0,0 +1,137 @@
...
+                                return;
+                            }
+                            builder.option(k, v);
+                        });
+
+        if (bucket != null) {
+            builder.option(BUCKET.key(), bucket.toString());
+        }
+
+        targetCatalog.createTable(tuple.f1, builder.build(), true);
```
shyjsarah, line 125: If target table already exists, source table schema changes will not be cloned. Is it necessary to support repeated migration?

JingsongLi: I think we can leave it for later. Currently, there is no sync on the Hive side too, only one check.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/clone/spits/ListCloneSplitsFunction.java
```diff
@@ -0,0 +1,132 @@
...
+                                    || k.equalsIgnoreCase(PATH.key())) {
+                                return;
+                            }
+                            builder.option(k, v);
+                        });
+
+        if (!sourceTable.primaryKeys().isEmpty()) {
+            // for primary key table, only postpone bucket supports clone
+            builder.option(BUCKET.key(), "-2");
+        }
```
shyjsarah, line 123: If append table specifies bucket-key, the target table needs to specify a bucket number, like 
```
else if (sourceTable.options().containsKey(BUCKET_KEY.key())) {
    builder.option(BUCKET.key(), sourceTable.options().get(BUCKET.key()));
}
```

PR #5913, opened 2025-07-16
[python] api:  support ecs role auth

jerry-024:
### Purpose
python api: support ecs role auth

### Tests
test_ecs_loader_token


JingsongLi (approved):
Looks good to me!

PR #5918, opened 2025-07-17
[spark] Introduce upsert-key for insert into non-pk table 

Zouxxyy:
### Purpose

```sql
CREATE TABLE T (k INT, ts INT, c STRING)
TBLPROPERTIES ('upsert-key' = 'k', 'sequence.field' = 'ts')

INSERT INTO T values (1, 1, 'c1'), (null, 2, 'c2'), (3, 3, 'c3')
INSERT INTO T values (null, 22, 'c22'), (1, 11, 'c11'), (4, 4, 'c4')
```

select  result
```sql
Row(null, 22, "c22"), Row(1, 11, "c11"), Row(3, 3, "c3"), Row(4, 4, "c4")
```

### DIFF
--- a/docs/content/spark/dataframe.md
+++ b/docs/content/spark/dataframe.md
@@ -1,5 +1,5 @@
 ---
-title: "Dataframe"
+title: "DataFrame"
 weight: 9
 type: docs
 aliases:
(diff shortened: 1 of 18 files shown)


JingsongLi:
Can you create `spark/sql-upsert` document?

PR #5922, opened 2025-07-18
[arrow] Fix map type conversion from paimon type to arrow type

leaves12138:
### Purpose

Map value type should be nullable if paimon map type value is nullable.
Do not force non-null for arrow map value type.


yuzelin (approved):
+1

PR #5924, opened 2025-07-18
[flink] Enable adaptive parallelism for FlinkStreamPartitioner

yunfengzhou-hub:
### Purpose

This PR allows configuring the `parallelismConfigured` property of `FlinkStreamPartitioner` to false. 

With this PR and previous commits, Paimon can now have most of its Flink sink operators dynamically changing their parallelism in AdaptiveBatch situations. Exceptional operators are as follows:
* Operators that are forced not to be parallel because of their syntax: Global Committers, Compact Coordinators and discarding sinks
* Operators in dynamic bucket mode and batch execution mode: dynamic bucket assigner and the writer operator.

### Tests

`ParallelismConfiguredITCase` is added to verify the `parallelismConfigured` property of these operators and the end-to-end correctness.


yunfengzhou-hub:
Hi @Sxnan could you please take a look at this PR?

Sxnan:
@yunfengzhou-hub Thanks for the PR! LGTM

JingsongLi:
+1

PR #5926, opened 2025-07-19
[doc] Keep old javadoc plugin for paimon website building

yuzelin:
### Purpose

New javadoc settings break the website building


JingsongLi:
+1

PR #5928, opened 2025-07-19
[core] Support off-peak hours

xiaochen-zhou:
### Purpose

Implementing a feature similar to HBase's off-peak:

Allows set a different (by default, more aggressive) percentage ratio for determining whether larger sorted run sizes are included in compactions during off-peak hours. Works in the same way as compaction.size-ratio. Only applies if offpeak.start.hour and offpeak.end.hour are also enabled.

### Tests

Add UT cases: OffPeakHoursTest、UniversalCompactionTest#testOffPeakRatioThreshold

### DIFF
--- a/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
@@ -680,6 +680,32 @@ public class CoreOptions implements Serializable {
                                     + "size is 1% smaller than the next sorted run's size, then include next sorted run "
                                     + "into this candidate set.");
 
+    public static final ConfigOption<Integer> OFFPEAK_START_HOUR =
+            key("offpeak.start.hour")
+                    .intType()
+                    .defaultValue(-1)
+                    .withDescription(
+                            "The start of off-peak hours, expressed as an integer between 0 and 23, inclusive"
+                                    + " Set to -1 to disable off-peak");
+
+    public static final ConfigOption<Integer> OFFPEAK_END_HOUR =
+            key("offpeak.end.hour")
+                    .intType()
+                    .defaultValue(-1)
+                    .withDescription(
+                            "The end of off-peak hours, expressed as an integer between 0 and 23, inclusive. Set"
+                                    + " to -1 to disable off-peak.");
+
+    public static final ConfigOption<Integer> COMPACTION_OFFPEAK_RATIO =
+            key("compaction.offpeak-ratio")
+                    .intType()
+                    .defaultValue(0)
+                    .withDescription(
+                            "Allows you to set a different (by default, more aggressive) percentage ratio for determining "
+                                    + " whether larger sorted run's size are included in compactions during off-peak hours. Works in the "
+                                    + " same way as compaction.size-ratio. Only applies if offpeak.start.hour and "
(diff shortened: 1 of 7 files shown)


paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -680,6 +680,32 @@ public InlineElement getDescription() {
...
+    public static final ConfigOption<Integer> OFFPEAK_START_HOUR =
+            key("offpeak.start.hour")
+                    .intType()
+                    .defaultValue(-1)
+                    .withDescription(
+                            "The start of off-peak hours, expressed as an integer between 0 and 23, inclusive"
+                                    + " Set to -1 to disable off-peak");
+
+    public static final ConfigOption<Integer> OFFPEAK_END_HOUR =
+            key("offpeak.end.hour")
```
JingsongLi: Add `compaction.`

paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -680,6 +680,32 @@ public InlineElement getDescription() {
                                     + "size is 1% smaller than the next sorted run's size, then include next sorted run "
                                     + "into this candidate set.");
 
+    public static final ConfigOption<Integer> OFFPEAK_START_HOUR =
+            key("offpeak.start.hour")
```
JingsongLi: Add `compaction.`

paimon-api/src/main/java/org/apache/paimon/OffPeakHours.java
```diff
@@ -0,0 +1,108 @@
...
+         * @param endHour exclusive
+         */
+        OffPeakHoursImpl(int startHour, int endHour) {
+            this.startHour = startHour;
+            this.endHour = endHour;
+        }
+
+        @Override
+        public boolean isOffPeak() {
+            return isOffPeak(ZonedDateTime.now(ZoneOffset.UTC).getHour());
```
JingsongLi: Why using UTC instead local?

xiaochen-zhou: > Why using UTC instead local?

My mistake, I modified the code. Please review it again.

paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -680,6 +680,32 @@ public InlineElement getDescription() {
...
+                    .withDescription(
+                            "The end of off-peak hours, expressed as an integer between 0 and 23, inclusive. Set"
+                                    + " to -1 to disable off-peak.");
+
+    public static final ConfigOption<Integer> COMPACTION_OFFPEAK_RATIO =
+            key("compaction.offpeak-ratio")
+                    .intType()
+                    .defaultValue(0)
+                    .withDescription(
+                            "Allows you to set a different (by default, more aggressive) percentage ratio for determining "
```
JingsongLi: Can you add a use case in description?

xiaochen-zhou: > Can you add a use case in description?

Done.

JingsongLi (approved):
+1

PR #5929, opened 2025-07-19
[doc][spark] Add dataframe page

Zouxxyy:
### Purpose

There are still many users who use dataframe api, add a document to facilitate users


JingsongLi (approved):
+1

Issue #5932, opened 2025-07-22
[Bug] Spark 3.3 throws write exceptions during partial-update engine operations


Aiden-Dong:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Paimon version

master

### Compute Engine

spark 3.3.1

### Minimal reproduce step

create table 
```
CREATE TABLE T( 
    f1 int, 
    f2 string, 
    f3 string, 
    f4 string) 
 TBLPROPERTIES (    
  'bucket'='1', 
  'primary-key'='f1', 
  'write-only'='true', 
  'merge-engine'='partial-update',   
  'file.format'='parquet', 
  'file.compression' = 'zstd'  
 )
```
insert 
```
INSERT INTO T(f1, f4) VALUES(1, 'test')
```

### What doesn't meet your expectations?

```
Cannot write incompatible data for the table `test.T`, the number of data columns don't match with the table schema's.
java.lang.RuntimeException: Cannot write incompatible data for the table `test.T`, the number of data columns don't match with the table schema's.
	at org.apache.paimon.spark.catalyst.analysis.PaimonAnalysis.resolveQueryColumnsByPosition(PaimonAnalysis.scala:135)
	at org.apache.paimon.spark.catalyst.analysis.PaimonAnalysis.org$apache$paimon$spark$catalyst$analysis$PaimonAnalysis$$resolveQueryColumns(PaimonAnalysis.scala:78)
	at org.apache.paimon.spark.catalyst.analysis.PaimonAnalysis$$anonfun$apply$1.applyOrElse(PaimonAnalysis.scala:43)
	at org.apache.paimon.spark.catalyst.analysis.PaimonAnalysis$$anonfun$apply$1.applyOrElse(PaimonAnalysis.scala:41)
	at org.apache.spark.sql.catalyst.plans.logical.AnalysisHelper.$anonfun$resolveOperatorsDownWithPruning$2(AnalysisHelper.scala:170)
	at org.apache.spark.sql.catalyst.trees.CurrentOrigin$.withOrigin(TreeNode.scala:176)
	at org.apache.spark.sql.catalyst.plans.logical.AnalysisHelper.$anonfun$resolveOperatorsDownWithPruning$1(AnalysisHelper.scala:170)
	at org.apache.spark.sql.catalyst.plans.logical.AnalysisHelper$.allowInvokingTransformsInAnalyzer(AnalysisHelper.scala:323)
[...]

PR #5933, opened 2025-07-22
[spark] Spark 3.3 throws write exceptions during partial-update engine operations

Aiden-Dong:
### Purpose

Linked issue: https://github.com/apache/paimon/issues/5932

fix bug with Spark 3.3 throws write exceptions during partial-update engine operations

```
CREATE TABLE T( 
    f1 int, 
    f2 string, 
    f3 string, 
    f4 string) 
 TBLPROPERTIES (    
  'bucket'='1', 
  'primary-key'='f1', 
  'write-only'='true', 
  'merge-engine'='partial-update',   
  'file.format'='parquet', 
  'file.compression' = 'zstd'  
 )
```

```
INSERT INTO T(f1, f4) VALUES(1, 'test')
```

### Tests

```
spark.sql(
      "CREATE TABLE T( f1 int, f2 string, f3 string, f4 string)\n" +
        "TBLPROPERTIES (\n" + "  'bucket'='1',\n" +
        "  'primary-key'='f1',\n" +
        "  'write-only'='true',\n" +
        "  'merge-engine'='partial-update',\n" + "" +
        "  'file.format'='parquet',\n" +
        "  'file.compression' = 'zstd'\n" + ")")

    spark.sql("INSERT INTO T(f1, f4) VALUES(1, 'test')")

    checkAnswer(
      spark.sql("SELECT * FROM T"),
      Row(1, null, null, "test")
    )
```

### DIFF
--- a/paimon-spark/paimon-spark-3.3/src/main/scala/org/apache/paimon/spark/catalyst/analysis/Spark3ResolutionRules.scala
+++ b/paimon-spark/paimon-spark-3.3/src/main/scala/org/apache/paimon/spark/catalyst/analysis/Spark3ResolutionRules.scala
@@ -35,8 +35,8 @@ import org.apache.spark.sql.connector.catalog.Identifier
 import org.apache.spark.sql.execution.datasources.v2.DataSourceV2Relation
 import org.apache.spark.sql.types.{ArrayType, DataType, IntegerType, Metadata, StructField, StructType}
 
-import scala.collection.mutable
 import scala.collection.JavaConverters._
+import scala.collection.mutable
 
 case class Spark3ResolutionRules(session: SparkSession)
   extends Rule[LogicalPlan]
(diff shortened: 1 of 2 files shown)


paimon-spark/paimon-spark-3.3/src/main/scala/org/apache/paimon/spark/catalyst/analysis/Spark3ResolutionRules.scala
```diff
@@ -0,0 +1,245 @@
...
+            val table = Identifier.of(ns.toArray, pattern)
+            val resolvedSpec =
+              PaimonResolvePartitionSpec.resolve(catalog.asTableCatalog, table, spec)
+            PaimonShowTablePartitionCommand(output, catalog.asTableCatalog, table, resolvedSpec)
+        }
+        .getOrElse {
+          PaimonShowTablesExtendedCommand(catalog.asTableCatalog, ns, pattern, output)
+        }
+
+    case insertPlan @ InsertIntoStatement(
```
cxzl25, line 65: Interesting function, so that we don't need a higher version of Spark and we can only specify some columns to update.

---

[SPARK-38795][SQL] Support INSERT INTO user specified column lists with DEFAULT values

https://issues.apache.org/jira/browse/SPARK-38795

Aiden-Dong:
> Hi @Aiden-Dong , I ran this test in current master, works fine. Can you just add this test?

Hi, to confirm - this partial column update test works fine for you on master/spark-3.3? I encountered issues with it on master recently, so just double-checking before I add the test.

PR #5936, opened 2025-07-22
[rest] Use apache httpclient instead of okhttp

qingwei727:
### Purpose
In high-concurrency scenarios, OkHttpClient may occasionally encounter RST (TCP connection reset) issues, while introducing Apache HttpClient can resolve these types of problems.

### Tests
TestExponentialHttpRequestRetryStrategy
HttpClientTest


paimon-api/pom.xml
```diff
@@ -56,6 +56,17 @@ under the License.
             <artifactId>okhttp</artifactId>
```
sundapeng: okhttp could be removed?

qingwei727: DLFECSTokenLoader already used

JingsongLi: DLFECSTokenLoader should use apache http too.

paimon-api/src/main/java/org/apache/paimon/rest/RESTCatalogOptions.java
```diff
@@ -90,4 +90,10 @@ public class RESTCatalogOptions {
                     .stringType()
                     .noDefaultValue()
                     .withDescription("REST Catalog auth DLF token ecs role name.");
+
+    public static final ConfigOption<Integer> REST_CLIENT_MAX_CONNECTIONS =
+            ConfigOptions.key("rest.client.max-connections")
```
JingsongLi: We don't need to introduce this option, just 100, I think there is almost no need to modify it

paimon-api/src/main/java/org/apache/paimon/rest/HttpClient.java
```diff
@@ -19,89 +19,87 @@
...
-                    .connectionSpecs(Arrays.asList(MODERN_TLS, COMPATIBLE_TLS, CLEARTEXT))
-                    .addInterceptor(new ExponentialHttpRetryInterceptor(5))
-                    .addInterceptor(new LoggingInterceptor())
-                    .connectTimeout(Duration.ofMinutes(3))
-                    .readTimeout(Duration.ofMinutes(3))
-                    .writeTimeout(Duration.ofMinutes(3))
-                    .build();
-
-    private static final MediaType MEDIA_TYPE = MediaType.parse("application/json");
+    private static volatile CloseableHttpClient httpClient;
```
JingsongLi: Why creating httpClient lazily? Will creating a client start a thread, or is it just lazy

paimon-api/pom.xml
qingwei727: done

paimon-api/src/main/java/org/apache/paimon/rest/RESTCatalogOptions.java
qingwei727: ok, already rm it

paimon-api/src/main/java/org/apache/paimon/rest/HttpClient.java
qingwei727: use static client instead

paimon-api/pom.xml
jerry-024, line 1: Should we remove the okhttp dependency?

qingwei727: OK, I will remove to core test for okhttp has been used a lot in test

paimon-api/src/main/java/org/apache/paimon/rest/SimpleHttpClient.java
jerry-024, line 1: Could move code to HttpClient, here maybe we need not a new class?

paimon-api/pom.xml
jerry-024: And check the NOTICE

JingsongLi: Just remove it in this PR.

paimon-api/src/main/java/org/apache/paimon/rest/SimpleHttpClient.java
qingwei727: We have discussed it, and the benefits of migrating to httpClient are minimal. We will not proceed with the migration for now.

paimon-api/pom.xml
```diff
@@ -109,6 +109,10 @@ under the License.
                                     <pattern>okio</pattern>
                                     <shadedPattern>org.apache.paimon.shade.okio</shadedPattern>
```
JingsongLi: Remove this too.

qingwei727: done

paimon-api/pom.xml
```diff
@@ -90,24 +94,20 @@ under the License.
...
+                                    <include>org.apache.httpcomponents.client5:httpclient5</include>
                                 </includes>
                             </artifactSet>
                             <relocations>
-                                <relocation>
-                                    <pattern>okhttp3</pattern>
-                                    <shadedPattern>org.apache.paimon.shade.okhttp3</shadedPattern>
-                                </relocation>
                                 <relocation>
                                     <pattern>kotlin</pattern>
```
jerry-024: remove

paimon-api/src/main/resources/META-INF/NOTICE
```diff
@@ -5,6 +5,6 @@ This product includes software developed at
 The Apache Software Foundation (http://www.apache.org/).
 
 This project bundles the following dependencies under the Apache Software License 2.0 (http://www.apache.org/licenses/LICENSE-2.0.txt)
-- com.squareup.okhttp3:okhttp:4.12.0
-- com.squareup.okio:okio-jvm:3.6.0
 - org.jetbrains.kotlin:kotlin-stdlib:1.8.21
```
jerry-024: remove

paimon-core/src/main/resources/META-INF/NOTICE
```diff
@@ -5,6 +5,7 @@ This product includes software developed at
 The Apache Software Foundation (http://www.apache.org/).
 
 This project bundles the following dependencies under the Apache Software License 2.0 (http://www.apache.org/licenses/LICENSE-2.0.txt)
-- com.squareup.okhttp3:okhttp:4.12.0
 - com.squareup.okio:okio-jvm:3.6.0
```
jerry-024: remove

jerry-024:
+1

JingsongLi:
+1 Thanks @qingwei727 @jerry-024

PR #5937, opened 2025-07-22
[core] Fix scan metric report for extra file-index files

wkang13579:
### Purpose

Reading table files contains scan phase and partition read phase. When we use bloom-filter or other file indexs, we found the scan metrics are the total data files of table, the file index seem to be not effective.

After analysis, in scan phase, the file index evaluation works  just effectively  for embedded file index, it does not work for extra file index. But extra file index actually works in partition read phase, this results inaccurate reporting metrics.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/AppendOnlyFileStore.java
+++ b/paimon-core/src/main/java/org/apache/paimon/AppendOnlyFileStore.java
@@ -155,6 +155,8 @@ public class AppendOnlyFileStore extends AbstractFileStore<InternalRow> {
                 schemaManager,
                 schema,
                 manifestFileFactory(),
+                fileIO,
+                pathFactory(),
                 options.scanManifestParallelism(),
                 options.fileIndexReadEnabled());
     }
--- a/paimon-core/src/main/java/org/apache/paimon/io/FileIndexEvaluator.java
+++ b/paimon-core/src/main/java/org/apache/paimon/io/FileIndexEvaluator.java
@@ -22,7 +22,6 @@ import org.apache.paimon.fileindex.FileIndexPredicate;
 import org.apache.paimon.fileindex.FileIndexResult;
 import org.apache.paimon.fs.FileIO;
 import org.apache.paimon.predicate.Predicate;
-import org.apache.paimon.predicate.PredicateBuilder;
 import org.apache.paimon.schema.TableSchema;
 
 import java.io.IOException;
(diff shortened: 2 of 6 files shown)


Tan-JiaLiang:
@Akwangg Thank you for your contribution! Did you mean evaluating the extra file index during the scanning phase?

Tan-JiaLiang:
> > @Akwangg Thank you for your contribution! Did you mean evaluating the extra file index during the scanning phase?感谢您的贡献！您的意思是在扫描阶段评估额外的文件索引吗？
> 
> yes, scan phase only evaluating the embedded file index, the scan result is incorrect for extra file index.是的，扫描阶段仅评估嵌入式文件索引，额外文件索引的扫描结果不正确。

I think it was done on purpose. During the scanning phase, we aim to evaluate as quickly as possible, so we only evaluate the embedding file index, which is stored in the manifest. The extra file index, however, is stored independently and requires additional I/O to load, it may slow down the scanning process.

Akwangg:
> > > @Akwangg Thank you for your contribution! Did you mean evaluating the extra file index during the scanning phase?
> > 
> > 
> > yes, scan phase only evaluating the embedded file index, the scan result is incorrect for extra file index.
> 
> I think it was done on purpose. During the scanning phase, we aim to evaluate as quickly as possible, so we only evaluate the embedding file index, which is stored in the manifest. The extra file index, however, is stored independently and requires additional I/O to load, it may slow down the scanning process.

If so, the reported scan metrics are the total table files, but some files are filtered when the data is actually read, which can easily lead to misunderstand that the index is not effective. Is there a better solution?

Tan-JiaLiang:
> If so, the reported scan metrics are the total table files, but some files are filtered when the data is actually read, which can easily lead to misunderstand that the index is not effective. Is there a better solution?

I think we can introduce some metrics for the file index evaluation phase.

JingsongLi (commented):
Good observability fix. The metrics gap between embedded file index (works in scan phase) and extra file index (works in read phase) causes confusing metrics — users think the file index isn't working.

**The problem**: Scan metrics report total data files because extra file index filtering happens later in the read phase. Users monitoring scan metrics see no benefit from their file index.

**Fix direction**: Either:
- Move extra file index evaluation into the scan phase (if feasible), OR
- Report additional metrics from the read phase that show post-filtering file counts

**Questions:**

1. Which approach does this PR take? The file list includes `FileIndexEvaluator.java` and `AppendOnlyFileStoreScan.java` — suggesting the evaluation is moved to scan phase?

2. **+127/-30** with test changes in both Java and Spark. The scope suggests a meaningful refactoring of where file index evaluation happens in the pipeline.

3. **Performance**: Moving extra file index evaluation to scan phase means reading more file index metadata during planning. Is this acceptable? For tables with many files, scan phase latency could increase even as total read time decreases.

4. **Append-only specific**: Changes are in `AppendOnlyFileStore` and `AppendOnlyFileStoreScan`. Does this also affect primary key tables with file indexes?

Please document the tradeoff (scan-time increase vs. accurate metrics) in the PR.

PR #5941, opened 2025-07-23
[core] make ignoreEmptyCommit configurable

wkang13579:
### Purpose

In our scenarios, we want to know the delete commit result explicitly even if it is empty, so make ignoreEmptyCommit configurable.

### DIFF
--- a/paimon-spark/paimon-spark-ut/src/test/java/org/apache/paimon/spark/SparkWriteITCase.java
+++ b/paimon-spark/paimon-spark-ut/src/test/java/org/apache/paimon/spark/SparkWriteITCase.java
@@ -516,13 +516,13 @@ public class SparkWriteITCase {
                 -1, Objects.requireNonNull(snapshotManager.latestSnapshot()).deltaRecordCount());
 
         // set false to allow commit empty snapshot
-        spark.conf().set("spark.paimon.ignore.empty.commit", "false");
+        spark.conf().set("spark.paimon.snapshot.ignore-empty-commit", "false");
         spark.sql("delete from T where id = 1");
         Assertions.assertEquals(3, snapshotManager.latestSnapshotId());
         Assertions.assertEquals(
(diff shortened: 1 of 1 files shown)


paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -1912,6 +1912,12 @@ public InlineElement getDescription() {
                                     + "in 'sink.clustering.by-columns'. 'order' is used for 1 column, 'zorder' for less than 5 columns, "
                                     + "and 'hilbert' for 5 or more columns.");
 
+    public static final ConfigOption<Boolean> IGNORE_EMPTY_COMMIT =
+            key("ignore.empty.commit")
```
JingsongLi: snapshot.ignore-empty-commit

wkangzz: done

paimon-spark/paimon-spark-ut/src/test/java/org/apache/paimon/spark/SparkWriteITCase.java
```diff
@@ -487,6 +489,42 @@ public void testChangelogFileSuffixName() throws Exception {
...
+                -1, Objects.requireNonNull(snapshotManager.latestSnapshot()).deltaRecordCount());
+
+        // in batch write, ignore.empty.commit default is true
+        spark.sql("delete from T where id = 1");
+        Assertions.assertEquals(2, snapshotManager.latestSnapshotId());
+        Assertions.assertEquals(
+                -1, Objects.requireNonNull(snapshotManager.latestSnapshot()).deltaRecordCount());
+
+        // set false to allow commit empty snapshot
+        spark.conf().set("spark.paimon.ignore.empty.commit", "false");
```
JingsongLi: test should be failed.

JingsongLi:
+1

PR #5942, opened 2025-07-23
[vfs] Support Mkdir in non-existing table

JingsongLi:
### Purpose

We should also support recursive mkdir, if the table does not exist, it should be automatically created.


timmyyao:
+1

PR #5944, opened 2025-07-23
[core] Automatically detect changes to the expiration policy.

### LINKED ISSUE #5943: [Feature] Automatically detect changes to the expiration policy.
### Search before asking

### Motivation

After modifying the snapshot expiration policy, a restart of the writer or compactor job is required for the new configuration to take effect. This process is inconvenient, and it would be better if the system could automatically detect configuration changes.

### Solution

detect configuration changes when expire snapshot.

### Anything else?

_No response_

zhoulii:
### Purpose

Linked issue: close #5943

### DIFF
--- a/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
@@ -444,7 +444,7 @@ public class CoreOptions implements Serializable {
                             "The maximum number of snapshots allowed to expire at a time.");
 
     public static final ConfigOption<Boolean> DETECT_EXPIRATION_SETTING_ENABLED =
-            key("detect-expiration-setting.enabled")
+            key("snapshot.expire.detect-config.enabled")
                     .booleanType()
                     .defaultValue(false)
                     .withDescription("Whether detect changes to expiration settings.");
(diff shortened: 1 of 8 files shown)


paimon-core/src/main/java/org/apache/paimon/table/ExpireChangelogImpl.java
```diff
@@ -79,6 +87,12 @@ public ExpireSnapshots config(ExpireConfig expireConfig) {
 
     @Override
     public int expire() {
+        TableSchema latestTableSchema = this.schemaManager.latest().get();
```
JingsongLi: This will bring additional IO access. Can you introduce a new option to describe whether a refresh is needed?

paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -443,6 +443,12 @@ public InlineElement getDescription() {
                     .withDescription(
                             "The maximum number of snapshots allowed to expire at a time.");
 
+    public static final ConfigOption<Boolean> DETECT_EXPIRATION_SETTING_ENABLED =
+            key("detect-expiration-setting.enabled")
```
JingsongLi: `snapshot.expire.detect-config.enabled`?

zhoulii:
Thanks for the review @JingsongLi . All comments have been addressed, Please take another look when you have a moment.

PR #5948, opened 2025-07-23
[hotfix] Fix npe when init TagManager with empty branch

qingwei727:
### Purpose
Fix npe when init TagManager with empty branch, org.apache.paimon.utils.BranchManager#isMainBranch may reduce npe


yuzelin (approved):
+1

PR #5952, opened 2025-07-24
[core] fix that branch can't be streamingly read if setting fallback branch

LsomeYeah:
### Purpose

Linked issue: close #xxx

```
Caused by: java.lang.ClassCastException: org.apache.paimon.table.source.DataSplit cannot be cast to org.apache.paimon.table.FallbackReadFileStoreTable$FallbackDataSplit
```
An exception would be thrown when attempting to read a branch configured with the `scan.fallback-branch` property. This occurs because the `FallbackReadFileStoreTable` class does not implement the `newStreamScan()` method. As a result, the data splits generated in streaming mode are instances of `DataSplit` rather than `FallbackDataSplit`, leading to a ClassCastException.


JingsongLi (approved):
+1

PR #5927, opened 2025-07-19
[core] process-time TagTimeExtractor supports custom zone

yuzelin:
### Purpose

In case that some full managed service machines cannot change timezone setting.

PR #5954, opened 2025-07-24
[core] Batch auto tag supports custom time zone

yuzelin:
### Purpose

Follow https://github.com/apache/paimon/pull/5927

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/tag/TagBatchCreation.java
+++ b/paimon-core/src/main/java/org/apache/paimon/tag/TagBatchCreation.java
@@ -45,6 +45,7 @@ public class TagBatchCreation {
     private final TagManager tagManager;
     private final SnapshotManager snapshotManager;
     private final TagDeletion tagDeletion;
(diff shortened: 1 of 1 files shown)


paimon-core/src/main/java/org/apache/paimon/tag/TagBatchCreation.java
```diff
@@ -45,22 +45,24 @@ public class TagBatchCreation {
...
 
     public void createTag() {
         Snapshot snapshot = snapshotManager.latestSnapshot();
         if (snapshot == null) {
             return;
         }
-        Instant instant = Instant.ofEpochMilli(snapshot.timeMillis());
-        LocalDateTime localDateTime = LocalDateTime.ofInstant(instant, ZoneId.systemDefault());
+        LocalDateTime localDateTime =
+                Instant.ofEpochMilli(snapshot.timeMillis()).atZone(zoneId).toLocalDateTime();
```
JingsongLi, line 65: Add a test for this.

Issue #5955, opened 2025-07-24
[Feature] Optimize CDC sync database action to avoid blocking on listTables operation with large number of tables


huyuanfeng2018:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Motivation

When using CDC synchronous database operations on a database containing a large number of tables (hundreds of thousands), the "catalog. listTables (database)" operation in "SyncDataActionBase" may take a long time to complete, causing the entire synchronization job to start blocking for a long time. This will significantly affect the duration of CDC synchronization task initiation.

### Current Behavior
The current implementation calls `catalog.listTables(database)` during initialization and maintains a `createdTables` set to track table creation status. This approach:
1. Blocks the entire sync process while listing all tables
2. Consumes unnecessary memory to maintain the `createdTables` set
3. Performs redundant operations when tables are created lazily

### Expected Behavior
The sync process should:
1. Avoid blocking on `listTables` operation during initialization
2. Create tables lazily when needed without maintaining a global `createdTables` set
3. Improve overall performance for databases with large numbers of tables

### Solution
Optimize the table creation logic by:
1. Removing the upfront `listTables` call in `SyncDatabaseActionBase`
2. Eliminating the `createdTables` set from `RichCdcMultiplexRecordEventParser`
3. Implementing lazy table creation in `CdcDynamicTableParsingProcessFunction` with existence checks

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

water-flower30:
Hello, I noticed that this issue has been paused for some time. Now we are also being affected by this issue. Can I proceed with the merging of this pr @huyuanfeng2018

PR #5956, opened 2025-07-24
[cdc] Optimize SyncDatabaseAction performance by removing listTables calls

huyuanfeng2018:
## Purpose
from #5955 

### What is the purpose of the change
Optimize SyncDatabaseAction performance by removing expensive listTables operations during initialization, improving scalability for databases with many tables.

### Brief change log
- Remove `listTables()` call from `RichCdcMultiplexRecordEventParser`
- Implement lazy table creation in `CdcDynamicTableParsingProcessFunction#processElement`
- Remove `createdTables` Set to reduce memory usage

### Verifying this change
- Verified existing functionality remains intact

### Testing
This optimization does not require additional test cases as the existing functionality is already covered by:
- `SyncDatabaseActionBaseTest.testSyncTablesWithoutDbLists()` - validates table filtering logic
- `SyncDatabaseActionBaseTest.testSyncTablesWithDbList()` - validates database filtering logic  
- `SyncDatabaseActionBaseTest.testSycTablesCrossDB()` - validates cross-database filtering scenarios

All these tests create and use RichCdcMultiplexRecordEventParser, ensuring the optimization doesn't break existing functionality.

### DIFF
--- a/paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/action/cdc/SyncDatabaseActionBase.java
+++ b/paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/action/cdc/SyncDatabaseActionBase.java
@@ -38,10 +38,8 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
 import java.util.regex.Pattern;
 
 import static org.apache.paimon.flink.action.MultiTablesSinkMode.COMBINED;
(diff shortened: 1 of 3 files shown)


JingsongLi (commented):
Good performance optimization for CDC database sync with many tables.

**The problem**: `listTables()` is called during initialization, which is expensive for databases with hundreds/thousands of tables. For each newly discovered CDC record, the system needs to know if the table exists — but calling `listTables()` eagerly loads all metadata upfront.

**The fix**: Lazy table creation in `processElement` — tables are created on-demand when CDC records arrive, rather than pre-listing all existing tables.

**Review:**

1. **+18/-23** — net reduction in code. Removing `createdTables` Set reduces memory. Good.

2. **Behavior change**: Previously, the system knew all existing tables at startup. Now it discovers them lazily. Does this affect:
   - Table filtering logic? (e.g., `--including-tables` pattern matching)
   - Startup validation? (e.g., verifying the source database matches the target)

3. **Race condition**: Multiple records for the same new table could arrive concurrently in `processElement`. Is the table creation idempotent / thread-safe?

4. **Existing tests cover this** per the description. Good.

LGTM if the above questions are addressed by the existing implementation.

JingsongLi:
+1

Issue #5851, opened 2025-07-08
[Feature] Introduce virtual filesystem for Object Table to provide file-level operations


timmyyao:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Motivation

Object Table provides us with a unified metadata view for raw files in the same catalog. We can introduce a virtual filesystem to visit the data in an Object Table.
1.  Implement HadoopFileSystem interface to use in most Hadoop ecosystem
2. Implement Object interface, like OSS SDK, to operate files in Object Table in the form of object operation
3. Unified table-level permission system, which is the same as normal Paimon tables.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #5958, opened 2025-07-25
[VFS] Enable table cache in PVFS

timmyyao:
### Purpose
Enable table cache in virtual filesystem to reduce rest requests to rest server.

Linked issue: https://github.com/apache/paimon/issues/5851

### DIFF
--- a/paimon-vfs/paimon-vfs-common/src/main/java/org/apache/paimon/vfs/VFSOperations.java
+++ b/paimon-vfs/paimon-vfs-common/src/main/java/org/apache/paimon/vfs/VFSOperations.java
@@ -18,6 +18,7 @@
 
 package org.apache.paimon.vfs;
 
+import org.apache.paimon.annotation.VisibleForTesting;
 import org.apache.paimon.catalog.CatalogContext;
 import org.apache.paimon.catalog.Identifier;
 import org.apache.paimon.fs.FileIO;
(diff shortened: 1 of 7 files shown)


paimon-vfs/paimon-vfs-common/src/main/java/org/apache/paimon/vfs/VFSOperations.java
```diff
@@ -238,4 +288,35 @@ private GetTableResponse loadTableMetadata(Identifier identifier) throws IOExcep
...
+    private VFSTableInfo loadTableInfo(Identifier identifier) throws IOException {
+        // Get table from REST server
+        GetTableResponse table;
+        table = loadTableMetadata(identifier);
+
+        if (table.isExternal()) {
+            throw new IOException("Do not support visiting external table " + identifier);
+        }
+        Path tablePath = new Path(table.getPath());
+        FileIO fileIO = new RESTTokenFileIO(context, api, identifier, tablePath);
```
JingsongLi, line 301: We should cache `FileIO` too.

paimon-vfs/pom.xml
```diff
@@ -49,43 +49,43 @@ under the License.
...
-            <scope>test</scope>
+            <scope>provided</scope>
         </dependency>
         <dependency>
             <groupId>org.apache.paimon</groupId>
             <artifactId>paimon-core</artifactId>
             <classifier>tests</classifier>
             <version>${project.version}</version>
-            <scope>test</scope>
+            <scope>provided</scope>
```
JingsongLi: Why change to provided?

paimon-vfs/paimon-vfs-common/src/main/java/org/apache/paimon/vfs/VFSOperations.java
```diff
@@ -55,8 +65,40 @@ public class VFSOperations {
     private final RESTApi api;
     private final CatalogContext context;
 
+    private Cache<Identifier, VFSTableInfo> tableCache;
```
JingsongLi: `@Nullable`.

paimon-vfs/paimon-vfs-common/src/main/java/org/apache/paimon/vfs/VFSOperations.java
```diff
@@ -55,8 +65,40 @@ public class VFSOperations {
...
+    private final boolean cacheEnabled;
```
JingsongLi: Just use tableCache is null or not null?

paimon-vfs/pom.xml
timmyyao: It's a misoperation, I will revert this.

JingsongLi (approved):
+1

PR #5961, opened 2025-07-25
[python] Support database query function of python rest catalog

discivigour:
### Purpose

Support database query function of python rest catalog. The implemnetion is in RestCatalog class.

### Tests

ApiTestCase.test_rest_catalog


JingsongLi (approved):
+1

PR #5962, opened 2025-07-25
[Python] Support get_table function of python RestCatalog

discivigour:
### Purpose

Support get_table function of python RestCatalog. All the changes are around get_table() in RestCatalog class

### Tests

api_test.ApiTestCase.test_rest_catalog


JingsongLi:
+1

PR #5963, opened 2025-07-26
[core] Add total compaction count metric

xiaochen-zhou:
### Purpose

Add total compaction count metric:  compactionTotalCount

### Tests

update CompactionMetricsTest#testReportMetrics()


JingsongLi (approved):
+1

PR #5964, opened 2025-07-28
[core]: fix column rename when columns referenced by table options.

zhongyujiang:
### Purpose

Currently, column renames don't automatically update table properties. For example, if a table has a column `foo` configured with `fields.foo.aggregate-function=collect`, rename `foo`  to `bar` won't update table options synchronously. This can lead to change errors:
```
java.lang.IllegalArgumentException: Field foo can not be found in table schema.
```

This PR apply column rename changes to table options to fix this.
### Tests

org.apache.paimon.table.SchemaEvolutionTest#testRenameFieldReferencedByOptions

### API and Format

No

### Documentation

No

Hi @JingsongLi @Zouxxyy, can you take a look at this when you have time? Thanks!


JingsongLi (approved):
+1

PR #5965, opened 2025-07-28
[core]add cache for table in pvfs

jerry-024:
### Purpose

Linked issue: close #xxx

### DIFF
--- a/paimon-python/pypaimon/api/rest_json.py
+++ b/paimon-python/pypaimon/api/rest_json.py
@@ -17,7 +17,7 @@
 
 import json
 from dataclasses import field, fields, is_dataclass
-from typing import Any, Type, Dict, TypeVar
+from typing import Any, Type, Dict, TypeVar, get_origin, get_args, Union
 
 T = TypeVar("T")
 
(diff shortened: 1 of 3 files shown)


paimon-python/pypaimon/api/config.py
```diff
@@ -40,5 +40,10 @@ class RESTCatalogOptions:
 
 
 class PVFSOptions:
+    TABLE_CACHE_SIZE = "table-cache-size"
```
JingsongLi: Please use option keys in Java `CatalogOptions`.

paimon-api/src/main/java/org/apache/paimon/options/CatalogOptions.java
```diff
@@ -111,6 +111,13 @@ public class CatalogOptions {
                     .withDescription(
                             "Controls the max number for which partitions in the catalog are cached.");
 
+    public static final ConfigOption<Long> CACHE_TABLE_MAX_NUM =
```
JingsongLi: We don't need this option. Just Int.max

JingsongLi (approved):
+1

PR #5970, opened 2025-07-29
[doc][hotfix] Correct Doris's URL link

yangjf2019:
### Purpose

Correct the doris url link in the _docs/content/ecosystem/overview.md_ document.


JingsongLi:
+1

PR #5971, opened 2025-07-29
[flink] Change filesystem.job-level-settings.enabled default value to true and add doc

yuzelin:
### Purpose

FileIO runtime context should be implemented by user, so if user have implemented it, of course he wants this feature. So adjust the default value for convenience.

### DIFF
--- a/docs/content/flink/quick-start.md
+++ b/docs/content/flink/quick-start.md
@@ -270,11 +270,7 @@ SELECT * FROM ....;
 
 ## Setting dynamic options
 
-When interacting with the Paimon table, table options can be tuned without changing the options in the table or catalog. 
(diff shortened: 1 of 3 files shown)


docs/content/flink/quick-start.md
```diff
@@ -295,3 +299,9 @@ SET 'paimon.mycatalog.default.T1.scan.timestamp-millis' = '1697018249000';
 SET 'scan.timestamp-millis' = '1697018249001';
 SELECT * FROM T1 JOIN T2 ON xxxx;
 ```
+
+### Set catalog filesystem options
```
JingsongLi: Don't need to add doc.

JingsongLi (approved):

PR #5972, opened 2025-07-29
[cdc] Support computed column referring to each other while sync_table

JackeyLee007:
### Purpose

Linked issue: close #xxx

1. As the title suggestd to support computed columns referring to each other while sync_table by sortting first according to dependency.
2. Also refactor the computed  columns sortting method to template manner.

### DIFF
--- a/paimon-api/src/main/java/org/apache/paimon/utils/DfsSort.java
+++ b/paimon-api/src/main/java/org/apache/paimon/utils/DfsSort.java
@@ -1,0 +1,103 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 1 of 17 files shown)


paimon-api/src/main/java/org/apache/paimon/utils/DfsSort.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/** DFS sort algorithm for topological sorting of a DAG (Directed Acyclic Graph). */
+public class DfsSort {
+    public static <K> LinkedHashMap<K, K> sort(LinkedHashMap<K, K> depMap) {
```
yuzelin: I think this shouldn't be in paimon-api.

PR #5973, opened 2025-07-29
[core] Introduce 'compaction.total-size-threshold' to do full compaction

JingsongLi:
### Purpose

Introduce 'compaction.total-size-threshold', when total size is smaller than this threshold, force a full compaction.


yuzelin (approved):
+1

PR #5975, opened 2025-07-29
[VFS] Fix PVFS trash behavior for Hadoop TrashPolicy

timmyyao:
### Purpose

Hadoop TrashPolicy will mkdir /user/root/.Trash, and we should reject this operation and return false, which indicates trash is not supported for TrashPolicy.


JingsongLi (approved):
+1

PR #5977, opened 2025-07-29
[core] Add lance table type for rest catalog

leaves12138:
### Purpose

Linked issue: close #xxx

### DIFF
--- a/paimon-api/src/main/java/org/apache/paimon/TableType.java
+++ b/paimon-api/src/main/java/org/apache/paimon/TableType.java
@@ -33,7 +33,8 @@ public enum TableType implements DescribedEnum {
             "materialized-table",
             "A materialized table combines normal Paimon table and materialized SQL."),
     OBJECT_TABLE(
-            "object-table", "An object table combines normal Paimon table and object location.");
+            "object-table", "An object table combines normal Paimon table and object location."),
+    LANCE_TABLE("lance-table", "Lance format table.");
 
     private final String value;
     private final String description;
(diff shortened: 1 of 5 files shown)


paimon-api/src/main/java/org/apache/paimon/TableType.java
```diff
@@ -33,7 +33,8 @@ public enum TableType implements DescribedEnum {
             "materialized-table",
             "A materialized table combines normal Paimon table and materialized SQL."),
     OBJECT_TABLE(
-            "object-table", "An object table combines normal Paimon table and object location.");
+            "object-table", "An object table combines normal Paimon table and object location."),
+    LANCE_TABLE("lance-table", "Lance format table.");
```
JingsongLi: "A lance table, see 'https://lancedb.github.io/lance/'."

paimon-core/src/main/java/org/apache/paimon/table/lance/LanceTable.java
```diff
@@ -0,0 +1,90 @@
...
+
+import org.apache.paimon.catalog.Identifier;
+import org.apache.paimon.fs.FileIO;
+import org.apache.paimon.table.Table;
+import org.apache.paimon.types.RowType;
+
+import java.util.Map;
+
+/**
+ * An object table refers to a directory that contains multiple objects (files). Object table
```
JingsongLi: Rewrite comments.

paimon-api/src/main/java/org/apache/paimon/TableType.java
leaves12138: OK

paimon-core/src/main/java/org/apache/paimon/table/lance/LanceTable.java
leaves12138: OK

PR #5978, opened 2025-07-29
[core] Log a warning for invalid partition values instead of throwing an exception when enable partition mark done. 

### LINKED ISSUE #5976: [Bug] Log a warning for invalid partition values instead of throwing an exception when enable partition mark done.
### Search before asking

### Paimon version

1.0

### Compute Engine

flink 1.18

### Minimal reproduce step

1.enable partition mark done;
2.write into `__DEFAULT_PARTITION__` 
3.when mark done `__DEFAULT_PARTITION__` , exception happens and job would fail over:

Image

It is more reasonable to log a warning rather than throwing an exception directly, similar to how partition expiration is handled. This approach prevents the job from getting stuck in a continuous failover loop when encountering invalid partition values.

### Anything else?

_No response_

zhoulii:
### Purpose

Linked issue: close #5976

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/listener/PartitionMarkDoneTrigger.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/listener/PartitionMarkDoneTrigger.java
@@ -45,6 +45,7 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 
 import static org.apache.paimon.CoreOptions.PARTITION_MARK_DONE_WHEN_END_INPUT;
 import static org.apache.paimon.flink.FlinkConnectorOptions.PARTITION_IDLE_TIME_TO_DONE;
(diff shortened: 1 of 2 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/listener/PartitionMarkDoneTrigger.java
```diff
@@ -167,7 +176,10 @@ LocalDateTime extractDateTime(String partition) {
         try {
             return timeExtractor.extract(extractPartitionSpecFromPath(new Path(partition)));
         } catch (DateTimeParseException e) {
-            throw new RuntimeException("Can't extract datetime from partition " + partition, e);
+            LOG.warn(
+                    "Can't extract datetime from partition {}, please check configuration items 'partition.timestamp-formatter' and 'partition.timestamp-pattern'.",
+                    partition);
+            return LocalDateTime.MIN;
```
JingsongLi: Can we just return Optional.empty?

JingsongLi:
+1

PR #5980, opened 2025-07-29
[common] Correct the range-bitmap's dictionary return results

Tan-JiaLiang:
### Purpose

The `Dictionary#find(java.lang.Object)` result is same as the `Collections.binarySearch`. 
Return the code by the search key if it is contained in the dictionary; otherwise, `(-(insertion point) - 1)` instead.


JingsongLi (approved):
+1

PR #5981, opened 2025-07-29
[core] Add required Iceberg Field IDs to support ID-based column pruning

### LINKED ISSUE #5875: [Bug] field-id missing for Iceberg-compatible manifest Avro schema
### Search before asking

### Paimon version

1.2.0

### Compute Engine

Flink 1.20.1

### Minimal reproduce step

I am inspecting the metadata and do not have the ability to reproduce the issue myself. I still hope this report is still useful to the community.

When writing Iceberg-compatible manifest files, the generated Avro schema currently omits field-id annotations in nested fields such as `null_value_counts`. This omission leads to compatibility issues with query engines and tools that expect fully compliant Iceberg metadata, including such as BigQuery and PyIceberg 

For example, the current schema for null_value_counts appears as from the Paimon compat metafile:

```json
{
  "name": "null_value_counts",
  "type": [
    "null",
    {
      "type": "array",
      "items": {
        "type": "record",
        "name": "r2_null_value_counts",
        "fields": [
          { "name": "key", "type": "int" },
          { "name": "value", "type": "long" }
        ]
      },
      "logicalType": "map"
    }
  ],
  "default": null
}
```

However, the expected schema (as produced by PyIceberg) includes explicit field-id annotations:

```json
{
  "name": "null_value_counts",
  "type": [
    "null",
    {
      "type": "array",
      "items": {
        "type": "record",
        "name": "k121_v122",
        "fields": [
          { "name": "key", "type": "int", "field-id": 121 },
          { "name": "value", "type": "long", "field-id": 122 }
        ]
      },
      "logicalType": "map"
    }
  ],
  "doc": "Map of column id to null value count",
  "default": null,
  "field-id": 110
}
```

It would be great if Paimon could update its manifest generation logic to include field-id metadata where required by the Iceberg spec. This small change would improve out-of-the-box interoperability with downstream engines.

For more see section https://iceberg.apache.org/spec/#avro, 
> Iceberg struct, list, and map types identify nested types by ID. When writing data to Avro files, these IDs must be stored in the Avro schema to support ID-based column pruning.

### Anything else?

_No response_

muttcg:
### Purpose

Linked issue: close #5875 

Apache Iceberg specification for Avro requires to have field-ids in order [to support ID-based column pruning.](https://iceberg.apache.org/spec/#avro)

For example Google Big Query engine as well as PyIceberg don't work at all without field-ids causing critical issues:

```
Error while reading data, error message: The Apache Avro library failed to read data with the following error: Cannot resolve: { "type": "record", "name": "r2_null_value_counts", "fields": [ { "name": "key", "type": "int" }, { "name": "value", "type": "long" } ] } with { "type": "record", "name": "k121_v122", "fields": [ { "name": "key", "type": "int", "field-id": 121 }, { "name": "value", "type": "long", "field-id": 122 } ] }; Failed to dispatch pruner query for ***.example.flair_paimon.meta original query id: *** with new query_id: *** File: bigstore/*****-test/table_v2/iceberg/example/flair_paimon/metadata/7d8b21d4-e536-49cd-bc1c-ee68b154c178-m8.avro
```

The change affects only Iceberg Avro schema creation, bringing new Iceberg custom properties following the specification, as well as fixing the wrong ID for `partitions` values.

### Tests
All existing tests must work. Added org.apache.paimon.iceberg.IcebergCompatibilityTest.testIcebergAvroFieldIds test to cover all required ids.

### API and Format

No


LsomeYeah (commented):
Thanks a lot for your contribution! Looks good to me!

Additionally, I was wondering if you could try reading the data using PyIceberg after this change? I ran some tests locally and found the following:
1. compress strategy
  - `snappy`: When the manifest's compression strategy is set to `snappy`, I encountered `ValueError: Checksum failure` when reading manifests.
  - `null`: When the manifest's compression strategy is `null`, the manifests can be read correctly without any issues.
2. file format

  - `avro`: I ran into errors when trying to read Avro-formatted data files. This seems to be due to PyArrow not supporting the reading of Avro files.
  - `parquet`: when the manifest's compression strategy is `null` and the data file's format is Parquet, I was able to successfully read the data using PyIceberg.

LsomeYeah:
> @LsomeYeah Thank you for checking We found and fixed another bug in PyIceberg [apache/iceberg-python#2252](https://github.com/apache/iceberg-python/pull/2252) Initially we tested with 'deflate', as I understand by default nobody uses avro compression for Iceberg manifests or so

Thanks for your reply.
The default compression for Iceberg manifest files in Paimon is `Snappy`. Perhaps we should consider changing the default compression for iceberg manifest file later. CC @JingsongLi

LsomeYeah (approved):
+1

muttcg:
@LsomeYeah 
I have re-checked Google BQ and Polars(uses PyIceberg and PyArrow under the hood) using fix and deflate, and everything works as I expect. 
Examples I used:
BQ query:
```
SELECT * 
FROM `***.example.flair_paimon_iceberg_v7`
WHERE partition_date = "2024-07-01"
AND partition_hh = "17"
AND app_version = '10.335.0'
ORDER BY country DESC
LIMIT 1000
```
Polars query:
import polars as pl
```

table_path = "gs://***/table_v2/iceberg/example/flair_paimon_iceberg_v7/metadata/v3.metadata.json"

df = (
  pl.scan_iceberg(table_path)
  .filter(
    (pl.col("partition_date") == "2024-07-01") &
    (pl.col("partition_hh") == "17") &
    (pl.col("app_version") == "10.335.0")
  )
  .sort("country", descending=True)
  .limit(1000)
  .collect()
)

print(df)
```

JingsongLi (approved):
Thanks @muttcg and @LsomeYeah , +1

PR #5983, opened 2025-07-29
[core] Fix incomplete large file filtering in BucketedAppendCompactManager

WenDing-Y:
### Purpose

This PR fixes an issue where the compaction logic in `BucketedAppendCompactManager` only filtered the first large file from the queue, leaving subsequent large files in the compaction set

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/append/BucketedAppendCompactManager.java
+++ b/paimon-core/src/main/java/org/apache/paimon/append/BucketedAppendCompactManager.java
@@ -39,6 +39,7 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Comparator;
+import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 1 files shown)


JingsongLi (commented):
Good bug fix. The old code had a subtle logic error:

**Before (broken):**
```java
while (!toCompact.isEmpty()) {
    DataFileMeta file = toCompact.peekFirst();
    if (file.fileSize() >= targetFileSize && !hasDeletionFile(file)) {
        toCompact.poll();
        continue;
    }
    break;  // ← stops at first non-large file!
}
```

This only removes large files from the *front* of the queue. If the queue is `[large, small, large]`, it removes the first large file, then hits the small file and `break`s — leaving the second large file in the compaction set.

**After (correct):**
Uses an `Iterator` to scan the entire list and remove all qualifying large files regardless of position.

**Concern**: The original code preserved order (only removing from the front). The new code removes from anywhere. Does the ordering of `toCompact` matter for compaction correctness? If files are supposed to be compacted in a specific order, removing from the middle could change semantics. But for full compaction tasks, the order likely doesn't matter — all remaining files are compacted together.

+9/-8 — minimal, focused fix. LGTM.

PR #5985, opened 2025-07-30
[core][spark] Support reading row lineage metadata column

Zouxxyy:
### Purpose

In this PR:

1. ~~add `metadataRowType()` and `fullRowType()` in `interface InnerTable`~~ remove the schema checker in `withReadType(RowType readType)`
3. make spark scan accpet `InnerTable` instead of `Table`
4. add `_ROW_ID` and `_SEQUENCE_NUMBER` metadata column in spark scan

```sql
SELECT *, _ROW_ID, _SEQUENCE_NUMBER FROM t
```

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/ColumnPruningAndPushDown.scala
+++ b/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/ColumnPruningAndPushDown.scala
@@ -18,24 +18,38 @@
 
 package org.apache.paimon.spark
 
+import org.apache.paimon.CoreOptions
 import org.apache.paimon.predicate.{Predicate, PredicateBuilder}
 import org.apache.paimon.spark.schema.PaimonMetadataColumn
-import org.apache.paimon.table.Table
+import org.apache.paimon.table.{InnerTable, SpecialFields}
 import org.apache.paimon.table.source.ReadBuilder
-import org.apache.paimon.types.RowType
+import org.apache.paimon.types.{DataField, RowType}
 import org.apache.paimon.utils.Preconditions.checkState
 
 import org.apache.spark.internal.Logging
 import org.apache.spark.sql.connector.read.Scan
 import org.apache.spark.sql.types.StructType
 
+import java.util.{ArrayList => JArrayList}
+
 trait ColumnPruningAndPushDown extends Scan with Logging {
-  def table: Table
+  def table: InnerTable
   def requiredSchema: StructType
   def filters: Seq[Predicate]
   def pushDownLimit: Option[Int] = None
 
-  lazy val tableRowType: RowType = table.rowType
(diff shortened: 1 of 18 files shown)


paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/ColumnPruningAndPushDown.scala
```diff
@@ -18,24 +18,38 @@
...
   def pushDownLimit: Option[Int] = None
 
-  lazy val tableRowType: RowType = table.rowType
+  lazy val tableRowType: RowType = {
+    val coreOptions: CoreOptions = CoreOptions.fromMap(table.options())
+    if (coreOptions.rowTrackingEnabled()) {
+      val fields = new JArrayList[DataField](table.rowType().getFields)
+      fields.add(SpecialFields.ROW_ID)
+      fields.add(SpecialFields.SEQUENCE_NUMBER)
+      new RowType(fields)
```
JingsongLi: Use `SpecialFields.rowTypeWithRowLineage`

leaves12138 (approved):
+1

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/ColumnPruningAndPushDown.scala
```diff
@@ -30,12 +31,20 @@ import org.apache.spark.sql.connector.read.Scan
...
-  def table: Table
+  def table: InnerTable
   def requiredSchema: StructType
   def filters: Seq[Predicate]
   def pushDownLimit: Option[Int] = None
 
-  lazy val tableRowType: RowType = table.rowType
+  lazy val tableRowType: RowType = {
+    val coreOptions: CoreOptions = CoreOptions.fromMap(table.options())
+    if (coreOptions.rowTrackingEnabled()) {
```
JingsongLi, line 41: Will add this for all scan to row tracking table?

JingsongLi: Maybe we should add a `isTableRewrite` to let scan add row lineage fields?

Zouxxyy: Actually  only when there are row lineage fields in `requiredSchema` (line 56) , then they will enter to `withReadType` to scan. 

`tableRowType` here just like full rowtype can be scanned

Zouxxyy: And `requiredSchema` is pushed by spark scan

JingsongLi:
+1

PR #5987, opened 2025-07-30
[Python] Support snapshot and manifest for PyPaimon

chenghuichen:
### Purpose
Support snapshot and manifest for PyPaimon


JingsongLi:
+1

PR #5988, opened 2025-07-30
[format] Introduce 'write.batch-memory' to control memory in arrow

leaves12138:
### Purpose

If every record is too big to put into one batch, then write.batch-size may be useless.
Introduce a new config write.batch-memory for this situation.

### DIFF
--- a/paimon-arrow/src/main/java/org/apache/paimon/arrow/vector/ArrowFormatWriter.java
+++ b/paimon-arrow/src/main/java/org/apache/paimon/arrow/vector/ArrowFormatWriter.java
@@ -27,7 +27,7 @@ import org.apache.paimon.types.RowType;
 
 import org.apache.arrow.memory.BufferAllocator;
 import org.apache.arrow.memory.RootAllocator;
-import org.apache.arrow.vector.ValueVector;
+import org.apache.arrow.vector.FieldVector;
 import org.apache.arrow.vector.VectorSchemaRoot;
 import org.apache.arrow.vector.util.OversizedAllocationException;
 import org.slf4j.Logger;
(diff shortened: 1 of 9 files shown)


paimon-arrow/src/main/java/org/apache/paimon/arrow/vector/ArrowFormatWriter.java
```diff
@@ -89,6 +115,13 @@ public boolean write(InternalRow currentRow) {
         return true;
     }
 
+    public long memoryUsed() {
+        vectorSchemaRoot.setRowCount(rowId);
+        return vectorSchemaRoot.getFieldVectors().stream()
```
JingsongLi: Please use for each instead of java stream, java stream has poor performance.

paimon-arrow/src/main/java/org/apache/paimon/arrow/vector/ArrowFormatWriter.java
```diff
@@ -75,6 +90,17 @@ public boolean write(InternalRow currentRow) {
         if (rowId >= batchSize) {
             return false;
         }
+        if (memoryUsedMaxInBytes != null && rowId % 20 == 0) {
```
JingsongLi: Do not use 20, maybe 32? 32 has very good performance.

leaves12138: OK

paimon-arrow/src/main/java/org/apache/paimon/arrow/vector/ArrowFormatWriter.java
leaves12138: OK

JingsongLi:
+1

PR #5990, opened 2025-07-30
[python] Change Schema to TableSchema in Class GetTableResponse. 

discivigour:
### Purpose

- Change Schema to TableSchema in Class GetTableResponse. 
- fix JSON.__from_dict() and some other specifications

### Tests

rest_catalog_test.RESTCatalogTestCase.test_rest_catalog()


JingsongLi:
+1

PR #5991, opened 2025-07-30
[core] Row-tracking row should keep their row_id and sequence_number in compaction

leaves12138:
### Purpose

Row id and Sequence number should be kept their origin value during compaction.

### DIFF
--- a/paimon-api/src/main/java/org/apache/paimon/table/SpecialFields.java
+++ b/paimon-api/src/main/java/org/apache/paimon/table/SpecialFields.java
@@ -20,8 +20,11 @@ package org.apache.paimon.table;
 
 import org.apache.paimon.types.DataField;
 import org.apache.paimon.types.DataTypes;
+import org.apache.paimon.types.RowType;
 import org.apache.paimon.types.VarCharType;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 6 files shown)


paimon-api/src/main/java/org/apache/paimon/table/SpecialFields.java
```diff
@@ -135,4 +141,20 @@ public static int getMapValueFieldId(int mapFieldId, int depth) {
                 + mapFieldId * STRUCTURED_TYPE_FIELD_DEPTH_LIMIT
                 + depth;
     }
+
+    public static RowType fieldsWithRowLineage(RowType rowType) {
```
JingsongLi: rowTypeWithRowLineage

paimon-api/src/main/java/org/apache/paimon/table/SpecialFields.java
```diff
@@ -99,6 +102,9 @@ public class SpecialFields {
                             ROW_ID.name())
                     .collect(Collectors.toSet());
 
+    public static final Set<String> ROW_LINEAGE_FIELD_NAMES =
```
JingsongLi: remove this? no use.

JingsongLi (approved):
+1

PR #5992, opened 2025-07-30
[wip]Optimize the operations for object storage

sundapeng:
### Purpose

Reduce the number of io calls for object storage. 

### Tests

N/A

### API and Format

N/A

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/fs/FileIO.java
+++ b/paimon-common/src/main/java/org/apache/paimon/fs/FileIO.java
@@ -252,7 +252,7 @@ public interface FileIO extends Serializable, Closeable {
             return true;
         }
 
-        if (dst.getParent().equals(src.getParent()) && dst.equals(src.getParent())) {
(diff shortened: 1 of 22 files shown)


JingsongLi (commented):
This PR is marked as [WIP]. The goal of reducing IO calls for object storage is valuable.

**Questions before this can be reviewed in depth:**

1. **Which specific IO calls are being reduced?** The file list touches `FileIO`, `SchemaManager`, `TableCommitImpl`, and `FileSystemBranchManager` — these are core paths. What's the strategy? Batching? Caching? Lazy evaluation?

2. **+72/-29** is modest, but touching `FileIO.java` (the base interface) means all implementations must be compatible with the new contract.

3. **No tests** listed. For an optimization that changes IO patterns, benchmark data (before/after IO call counts) would be very helpful.

4. **WIP status**: Please fill in the PR description with the specific optimizations and mark as ready when complete. As-is, the changes are difficult to evaluate without understanding the intended optimization pattern.

Happy to review once the design is documented and the WIP tag is removed.

PR #6000, opened 2025-07-31
[python] Support mutual conversion between paimon DataField and pyarrow Field

discivigour:
### Purpose

- Support partial mutual conversion between nested paimon DataField and pyarrow Field, including list and map.
- Support create Schema from pyarrow Schema

### Tests

schema_test.SchemaTestCase.test_types()


JingsongLi:
+1

PR #6001, opened 2025-07-31
[Python] Support table write for PyPaimon

chenghuichen:
### Purpose
Support table write for PyPaimon

### Tests

filesystem_catalog_test.py
writer_test.py


JingsongLi (approved):
+1

PR #6003, opened 2025-07-31
[VFS] Support inline endpoint in pvfs uri, like pvfs://catalog.endpoint/

timmyyao:
### Purpose
To align with S3 and OSS FileSystem, PVFS shall also support use case like pvfs://catalog.endpoint/, thus do not need to configure endpoint in configuration.


JingsongLi (approved):
+1

PR #6004, opened 2025-07-31
[Core] Custom commit callback support to be initialized with table

yuzelin:
### Purpose

In some cases, the callback should get some message from Table.


JingsongLi:
+1

PR #6005, opened 2025-07-31
[FLINK] Predicate add is true and is false

WenDing-Y:
### Purpose

flink PredicateConverter  add is true and is false


JingsongLi:
+1

PR #6006, opened 2025-07-31
Site/nit: Replace PPMC with PMC

snazy:
Paimon is a top-level-project, PPMC is probably a left-over from graduation.


JingsongLi:
Good catch! Thank you @snazy !

JingsongLi (approved):

PR #6013, opened 2025-08-01
[core] Introduce expireForEmptyCommit to InnerTableCommit

JingsongLi:
### Purpose

Some cases we can do nothing for empty commit, no expire.


yuzelin (commented):
+1

PR #6014, opened 2025-08-01
[core][spark] Support UPDATE for row lineage

Zouxxyy:
### Purpose

In thir pr:
- Add `withWriteType` in  interface `TableWrite`
- Support spark update with row lineage table

e.g.

```sql
CREATE TABLE t (id INT, data INT) TBLPROPERTIES ('row-tracking.enabled' = 'true');
INSERT INTO t SELECT /*+ REPARTITION(1) */ id, id AS data FROM range(1, 4);
SELECT *, _ROW_ID, _SEQUENCE_NUMBER FROM t ORDER BY id;

+---+----+-------+----------------+
| id|data|_ROW_ID|_SEQUENCE_NUMBER|
+---+----+-------+----------------+
|  1|   1|      0|               1|
|  2|   2|      1|               1|
|  3|   3|      2|               1|
+---+----+-------+----------------+

in parquet file:
D select * from 'data-c7882bf0-110f-4e30-a054-03171c53cbe3-0.parquet';
┌───────┬───────┐
│  id   │ data  │
│ int32 │ int32 │
├───────┼───────┤
│     1 │     1 │
│     2 │     2 │
│     3 │     3 │
└───────┴───────┘
```

```sql
UPDATE t SET data = 22 WHERE id = 2;
SELECT *, _ROW_ID, _SEQUENCE_NUMBER FROM t ORDER BY id;

+---+----+-------+----------------+
| id|data|_ROW_ID|_SEQUENCE_NUMBER|
+---+----+-------+----------------+
|  1|   1|      0|               1|
|  2|  22|      1|               2|
|  3|   3|      2|               1|
+---+----+-------+----------------+

in parquet file:
D select * from 'data-717cf519-f630-4543-8498-718a4842adf9-0.parquet';
┌───────┬───────┬─────────┬──────────────────┐
│  id   │ data  │ _ROW_ID │ _SEQUENCE_NUMBER │
│ int32 │ int32 │  int64  │      int64       │
├───────┼───────┼─────────┼──────────────────┤
│     1 │     1 │       0 │                1 │
│     2 │    22 │       1 │             NULL │
│     3 │     3 │       2 │                1 │
└───────┴───────┴─────────┴──────────────────┘
```

### DIFF
--- a/paimon-api/src/main/java/org/apache/paimon/table/SpecialFields.java
+++ b/paimon-api/src/main/java/org/apache/paimon/table/SpecialFields.java
@@ -77,7 +77,7 @@ public class SpecialFields {
     public static final int KEY_FIELD_ID_START = SYSTEM_FIELD_ID_START;
 
     public static final DataField SEQUENCE_NUMBER =
-            new DataField(Integer.MAX_VALUE - 1, "_SEQUENCE_NUMBER", DataTypes.BIGINT());
+            new DataField(Integer.MAX_VALUE - 1, "_SEQUENCE_NUMBER", DataTypes.BIGINT().notNull());
 
     public static final DataField VALUE_KIND =
             new DataField(Integer.MAX_VALUE - 2, "_VALUE_KIND", DataTypes.TINYINT().notNull());
@@ -152,7 +152,7 @@ public class SpecialFields {
                     }
                 });
         fieldsWithRowLineage.add(SpecialFields.ROW_ID);
-        fieldsWithRowLineage.add(SpecialFields.SEQUENCE_NUMBER);
+        fieldsWithRowLineage.add(SpecialFields.SEQUENCE_NUMBER.copy(true));
         return new RowType(fieldsWithRowLineage);
     }
 }
--- a/paimon-api/src/main/java/org/apache/paimon/table/SpecialFields.java
+++ b/paimon-api/src/main/java/org/apache/paimon/table/SpecialFields.java
@@ -143,6 +143,12 @@ public class SpecialFields {
         return rowTypeWithRowLineage(rowType, false);
     }
 
+    /**
+     * Add row lineage fields to rowType.
+     *
+     * @param sequenceNumberNullable sequence number is not null for user, but is nullable when read
+     *     and write
+     */
     public static RowType rowTypeWithRowLineage(RowType rowType, boolean sequenceNumberNullable) {
         List<DataField> fieldsWithRowLineage = new ArrayList<>(rowType.getFields());
 
--- a/paimon-api/src/main/java/org/apache/paimon/table/SpecialFields.java
+++ b/paimon-api/src/main/java/org/apache/paimon/table/SpecialFields.java
@@ -140,6 +140,10 @@ public class SpecialFields {
     }
 
     public static RowType rowTypeWithRowLineage(RowType rowType) {
+        return rowTypeWithRowLineage(rowType, false);
+    }
+
+    public static RowType rowTypeWithRowLineage(RowType rowType, boolean sequenceNumberNullable) {
         List<DataField> fieldsWithRowLineage = new ArrayList<>(rowType.getFields());
 
         fieldsWithRowLineage.forEach(
@@ -152,7 +156,10 @@ public class SpecialFields {
                     }
                 });
         fieldsWithRowLineage.add(SpecialFields.ROW_ID);
-        fieldsWithRowLineage.add(SpecialFields.SEQUENCE_NUMBER.copy(true));
+        fieldsWithRowLineage.add(
+                sequenceNumberNullable
+                        ? SpecialFields.SEQUENCE_NUMBER.copy(true)
+                        : SpecialFields.SEQUENCE_NUMBER);
         return new RowType(fieldsWithRowLineage);
     }
 }
--- a/paimon-api/src/main/java/org/apache/paimon/table/SpecialFields.java
+++ b/paimon-api/src/main/java/org/apache/paimon/table/SpecialFields.java
@@ -77,7 +77,7 @@ public class SpecialFields {
     public static final int KEY_FIELD_ID_START = SYSTEM_FIELD_ID_START;
 
     public static final DataField SEQUENCE_NUMBER =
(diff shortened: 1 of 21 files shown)


paimon-api/src/main/java/org/apache/paimon/table/SpecialFields.java
```diff
@@ -77,7 +77,7 @@ public class SpecialFields {
     public static final int KEY_FIELD_ID_START = SYSTEM_FIELD_ID_START;
 
     public static final DataField SEQUENCE_NUMBER =
-            new DataField(Integer.MAX_VALUE - 1, "_SEQUENCE_NUMBER", DataTypes.BIGINT().notNull());
```
JingsongLi: Be careful, here will make old avro data file in primary key table not compatible.

paimon-core/src/main/java/org/apache/paimon/utils/FormatReaderMapping.java
```diff
@@ -174,7 +174,7 @@ public FormatReaderMapping build(
                 String formatIdentifier, TableSchema tableSchema, TableSchema dataSchema) {
 
             // extract the whole data fields in logic.
-            List<DataField> allDataFields = fieldsExtractor.apply(dataSchema);
+            List<DataField> allDataFields = new ArrayList<>(fieldsExtractor.apply(dataSchema));
```
Zouxxyy, line 177: The fieldsExtractor here directly retrieves the fields list from the `dataSchema`. Adding fields directly can introduce unknown bugs.

So I changed the fields in the `TableSchema` to be unmodifiable.
`this.fields = Collections.unmodifiableList(new ArrayList<>(fields));`

paimon-api/src/main/java/org/apache/paimon/table/SpecialFields.java
Zouxxyy: Change `SpecialFields.SEQUENCE_NUMBER.copy(true)` in `rowTypeWithRowLineage`

leaves12138: Maybe, we should just copy nullable when read from file.
Because, sequence number actually is not null for user.
User select _SEQUENCE_NUMBER will never get a NULL

paimon-api/src/main/java/org/apache/paimon/table/SpecialFields.java
```diff
@@ -140,6 +140,10 @@ public static int getMapValueFieldId(int mapFieldId, int depth) {
     }
 
     public static RowType rowTypeWithRowLineage(RowType rowType) {
+        return rowTypeWithRowLineage(rowType, false);
+    }
+
+    public static RowType rowTypeWithRowLineage(RowType rowType, boolean sequenceNumberNullable) {
```
leaves12138, line 152: Comment "sequence number is not null for user, but is nullable when read and write"

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/compact/AppendTableCompactor.java
```diff
@@ -69,7 +71,11 @@ public AppendTableCompactor(
             @Nullable MetricGroup metricGroup) {
         this.table = table;
         this.commitUser = commitUser;
+        CoreOptions coreOptions = new CoreOptions(table.options());
```
leaves12138: Use table.coreOptions()

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/procedure/CompactProcedure.java
```diff
@@ -421,6 +422,13 @@ private void compactUnAwareBucketTable(
                                             BaseAppendFileStoreWrite write =
                                                     (BaseAppendFileStoreWrite)
                                                             table.store().newWrite(commitUser);
+                                            CoreOptions coreOptions =
+                                                    new CoreOptions(table.options());
```
leaves12138: table.coreOptions()

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/SparkTableWrite.scala
```diff
@@ -29,12 +29,27 @@ import org.apache.spark.sql.{PaimonUtils, Row}
 import scala.collection.JavaConverters._
 import scala.collection.mutable.ListBuffer
 
-class SparkTableWrite(writeBuilder: BatchWriteBuilder, rowType: RowType, rowKindColIdx: Int)
+case class SparkTableWrite(
```
leaves12138, line 32: Why case class? Is this class used for case match?

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/commands/PaimonSparkWriter.scala
```diff
@@ -51,22 +51,37 @@ import java.util.Collections.singletonMap
 
 import scala.collection.JavaConverters._
 
-case class PaimonSparkWriter(table: FileStoreTable) extends WriteHelper {
+case class PaimonSparkWriter(table: FileStoreTable, writeRowLineage: Boolean = false)
```
leaves12138, line 54: Why case class?

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/SparkTableWrite.scala
Zouxxyy: to indicate that the members in it are immutable,  can be added or not.

leaves12138:
Can we add a test that:
Update one record on table, then we search it's lineage.

For, example,
Before: name: zhang.san  id: 1  score: 100  (Snapshot 0)
After: name:zhang.san id:1 score: 120  (Snapshot 1)

then, we search for it
select *, _SEQUENCE_NUMBER FROM  paimon_incremental_query('tableName', 0, 2) WHERE name = 'zhang.san'; 

We get:
name: zhang.san  id: 1  score: 100 _SEQUENCE_NUMBER:0
name:zhang.san id:1 score: 120 _SEQUENCE_NUMBER:1

leaves12138 (approved):
+1

PR #6017, opened 2025-08-03
[test] Fix FlinkRestCatalogITCase ut failed in idea

qingwei727:
### Purpose
Fix the failure of FlinkRestCatalogITCase in IDEA, as httpcomponents is relocated in the paimon-api module, and IDEA cannot recognize this scenario. The dependency should be explicitly defined.

### Documentation


JingsongLi:
+1

PR #6025, opened 2025-08-04
[core] Fix the commit kind when performing row-level changes on non-pk table

Zouxxyy:
### Purpose

Currently, there are some issues with the commit kind for non-PK tables under row-level commands:

- UPDATE, DELETE, and MERGE operations that remove old files should be classified as OVERWRITE, not APPEND.
- For non-PK DV (Deletion Vector) tables, only a single snapshot commit is needed (no need two), and it should also be of type OVERWRITE.

Here is a brief summary of commit kinds after modifcation:

- APPEND: 
  - insert into table
  - place data to PK table's level 0 (e.g. the first step of PK tables' DV generation).
- COMPACT:
  - Compact non-PK tables (e.g. merge small files)
  - Compact PK tables (e.g. triggered by LSM tree compaction)
  - Compact PK DV tables (e.g. triggered by DV generation)
- OVERWRITE:
  - insert overwrite (partition) table
  - drop partition
  - delete, update, and merge on non-PK tables (Note: this commit kind applies to both DV and non-DV tables) (Note: if this operation does not delete any old files, it will be classified as an APPEND. e.g. a merge can be considered as APPEND when it only insert condition matched)
- ANALYZE: 
  - analyze table

### DIFF
--- a/paimon-api/src/main/java/org/apache/paimon/Snapshot.java
+++ b/paimon-api/src/main/java/org/apache/paimon/Snapshot.java
@@ -496,13 +496,16 @@ public class Snapshot implements Serializable {
     /** Type of changes in this snapshot. */
     public enum CommitKind {
 
-        /** Changes flushed from the mem table. */
+        /** New data files are appended to the table and no data file is deleted. */
         APPEND,
 
         /** Changes by compacting existing data files. */
         COMPACT,
 
-        /** Changes that clear up the whole partition and then add new records. */
+        /**
+         * New data files are added to overwrite existing data files or just delete existing data
+         * files.
+         */
         OVERWRITE,
 
         /** Collect statistics. */
--- a/paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
@@ -747,7 +747,7 @@ public class FileStoreCommitImpl implements FileStoreCommit {
                     .newIndexFiles()
                     .forEach(
                             m ->
-                                    appendIndexFiles.add(
+                                    compactIndexFiles.add(
                                             new IndexManifestEntry(
                                                     FileKind.ADD,
                                                     commitMessage.partition(),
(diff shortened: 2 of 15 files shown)


paimon-api/src/main/java/org/apache/paimon/Snapshot.java
```diff
@@ -496,13 +496,13 @@ public boolean equals(Object o) {
     /** Type of changes in this snapshot. */
     public enum CommitKind {
 
-        /** Changes flushed from the mem table. */
+        /** New data is appended to the table and no data is deleted. */
```
JingsongLi: `New data files are appended to the table and no data file is deleted.`

paimon-api/src/main/java/org/apache/paimon/Snapshot.java
```diff
@@ -496,13 +496,13 @@ public boolean equals(Object o) {
...
         APPEND,
 
         /** Changes by compacting existing data files. */
         COMPACT,
 
-        /** Changes that clear up the whole partition and then add new records. */
+        /** New data is added to overwrite existing data or just delete existing data. */
```
JingsongLi: `New data files are added to overwrite existing data files or just delete existing data files.`

JingsongLi (commented):
Can you just create a separate `overwrite` method in `FileStoreCommitImpl`?

JingsongLi (changes requested):
Thanks @Zouxxyy !

Left two comments:
1. In `CommitMessageImpl`, `indexIncrement` should be separated into `dataIncrement` and `compactIncrement`, append index and compact index.
2. `FileStoreCommitImpl.noConflictsOrFail` should contains check for deletion vectors.

paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
```diff
@@ -426,8 +428,23 @@ private void reportCommit(
         commitMetrics.reportCommit(commitStats);
     }
 
+    private boolean hasDelete(
```
JingsongLi, line 431: `containsFileDeletionOrDeletionVectors`

JingsongLi:
+1

> FileStoreCommitImpl.noConflictsOrFail should contains check for deletion vectors.

We can finish this in next PR.

PR #6027, opened 2025-08-05
[flink] Bump Flink version to 2.1

Sxnan:
### Purpose

Bump Flink version to 2.1

### DIFF
--- a/paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/SchemaChangeITCase.java
+++ b/paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/SchemaChangeITCase.java
@@ -21,7 +21,6 @@ package org.apache.paimon.flink;
 import org.apache.paimon.data.BinaryString;
 import org.apache.paimon.utils.DateTimeUtils;
 
-import org.apache.flink.table.api.TableException;
 import org.apache.flink.table.api.config.ExecutionConfigOptions;
 import org.apache.flink.types.Row;
 import org.junit.jupiter.api.Test;
@@ -685,7 +684,6 @@ public class SchemaChangeITCase extends CatalogITCaseBase {
                                         "INSERT INTO T VALUES('aaa', 'bbb', 'ccc', 1, CAST(NULL AS FLOAT))"))
                 .satisfies(
                         anyCauseMatches(
-                                TableException.class,
                                 "Column 'e' is NOT NULL, however, a null value is being written into it."));
 
         // Not null -> nullable
@@ -718,7 +716,6 @@ public class SchemaChangeITCase extends CatalogITCaseBase {
                                         "INSERT INTO T VALUES('aaa', 'bbb', CAST(NULL AS STRING), 1, CAST(NULL AS FLOAT))"))
                 .satisfies(
                         anyCauseMatches(
-                                TableException.class,
                                 "Column 'c' is NOT NULL, however, a null value is being written into it."));
 
         // Insert a null value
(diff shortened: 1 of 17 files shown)


Sxnan:
@yunfengzhou-hub, can you review this PR?

paimon-flink/paimon-flink1-common/src/main/java/org/apache/flink/table/data/RowData.java
```diff
@@ -0,0 +1,314 @@
...
+
+    /**
+     * Returns the row value at the given position.
+     *
+     * <p>The number of fields is required to correctly extract the row.
+     */
+    RowData getRow(int pos, int numFields);
+
+    /** Returns the variant value at the given position. */
+    Variant getVariant(int pos);
```
yunfengzhou-hub, line 206: Do we need to add a RowData interface with getVariant method to paimon-flink-2.0 as well?

pom.xml
```diff
@@ -466,13 +466,15 @@ under the License.
             <id>flink2</id>
             <properties>
                 <paimon-flinkx-common>paimon-flink2-common</paimon-flinkx-common>
-                <paimon-flink-common.flink.version>2.0.0</paimon-flink-common.flink.version>
+                <paimon-flink-common.flink.version>2.1.0</paimon-flink-common.flink.version>
                 <test.flink.main.version>2.0</test.flink.main.version>
```
yunfengzhou-hub: Should this property also be changed to 2.1?

pom.xml
yunfengzhou-hub, line 1: Please check the pom of paimon-e2e-tests. Maybe we need to add a new profile for flink-2.0 now.

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/SchemaChangeITCase.java
```diff
@@ -685,7 +684,6 @@ public void testModifyNullability() {
                                         "INSERT INTO T VALUES('aaa', 'bbb', 'ccc', 1, CAST(NULL AS FLOAT))"))
                 .satisfies(
                         anyCauseMatches(
-                                TableException.class,
```
yunfengzhou-hub, line 688: Will the introduction of Flink 2.1 break this exception class?

yunfengzhou-hub (commented):
Thanks for the PR! Left some comments as below.

paimon-flink/paimon-flink1-common/src/main/java/org/apache/flink/table/data/RowData.java
Sxnan: I think it should, but it feels off that this is not caught by the CI. I will have a look at the CI.

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/SchemaChangeITCase.java
Sxnan: Yes, Flink 2.1 changes the exception type from `TableException` to `EnforcerException`. Instead of making the test to adapt to both cases, I think it is easier to just ignore the type and just make sure the exception message is correct.

see https://github.com/apache/flink/pull/26540

paimon-flink/paimon-flink1-common/src/main/java/org/apache/flink/table/data/RowData.java
Sxnan: It is added. And the e2e test for flink 2.0 is added to cover the situation.

PR #6033, opened 2025-08-06
[common] Minor fix for range-bitmap

Tan-JiaLiang:
### Purpose

Fix the following two issues:
1. An unexpected error will be thrown when building the range-bitmap if all the values are NULL.
2. If the cardinality of a value is 1, an unexpected exception will be thrown when reading the range-bitmap's bsi.


JingsongLi:
+1

PR #6037, opened 2025-08-07
[spark] Unify config extraction for maxPartitionBytes and openCostInBytes between paimon and spark

Zouxxyy:
### Purpose

For maxPartitionBytes, paimon has `source.split.target-size`, spark has `spark.sql.files.maxPartitionBytes`

For file openCostInBytes, paimon has `source.split.open-file-cost`, spark has `spark.sql.files.openCostInBytes`

This pr combine them, and make Paimon's conf takes precedence over Spark's if we set them both


JingsongLi:
+1

PR #6038, opened 2025-08-07
[core] Enabled field merge read in row-tracking table

leaves12138:
### Purpose

Enable merge field read if DataFileMetas (APPEND kind) have the same firstRowId.

### Tests

paimon-core/src/test/java/org/apache/paimon/table/AppendMergeFieldTest.java


paimon-core/src/main/java/org/apache/paimon/io/RowDataFileWriter.java
```diff
@@ -117,6 +120,7 @@ public DataFileMeta result() throws IOException {
                 fileSource,
                 statsPair.getKey(),
                 externalPath,
-                null);
+                null,
+                fieldIds);
```
JingsongLi: Null if writing all fields.
Be careful for the table without this feature enabled.

paimon-core/src/main/java/org/apache/paimon/io/DataFileMeta.java
```diff
@@ -128,6 +130,8 @@ public class DataFileMeta {
 
     private final @Nullable Long firstRowId;
 
+    private final @Nullable int[] fieldIds;
```
JingsongLi: `String[] _WRITE_COLS`?

paimon-core/src/main/java/org/apache/paimon/io/RowDataFileWriter.java
leaves12138: OK

paimon-core/src/main/java/org/apache/paimon/io/DataFileMeta.java
leaves12138: OK

paimon-core/src/main/java/org/apache/paimon/table/source/RowLineageSplitGenerator.java
```diff
@@ -0,0 +1,111 @@
...
+                                (f1, f2) -> {
+                                    // If firstRowId is the same, we should read the file with
+                                    // larger sequence number first. Because larger sequence number
+                                    // file is more fresh
+                                    return Long.compare(
+                                            f2.maxSequenceNumber(), f1.maxSequenceNumber());
+                                }));
+
+        // Split files by firstRowId
+        long lastRowId = -1;
```
JingsongLi: Should throw exception for exception case.

leaves12138: OK

JingsongLi (commented):
Maybe we should introduce a new option: `data-evolution.enabled`.

PR #6043, opened 2025-08-08
[python]Support RESTCatalog create, read and write append table

discivigour:
### Purpose

- Support RESTCatalog create, read and write append table
- fix s3filesystem remote write bug

### Tests
- rest_catalog_test: RESTCatalogTest


paimon-python/pypaimon/common/file_io.py
```diff
@@ -62,6 +62,7 @@ def _initialize_oss_fs(self) -> FileSystem:
 
         client_kwargs = {
             "endpoint_override": self.properties.get(OssOptions.OSS_ENDPOINT),
+            # "endpoint_override": "oss-cn-hangzhou.aliyuncs.com",
```
JingsongLi: remove this.

discivigour: ok

PR #6049, opened 2025-08-10
[spark] Optimize the config reading for load table in SparkSource

Zouxxyy:
### Purpose

- If only path conf is provided, the catalog configs should not be read.
- Provides catalog, database, and table options, to allow the user to specify the catalog name.
- Fix other options reading, like time travel

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/SparkSource.scala
+++ b/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/SparkSource.scala
@@ -19,18 +19,18 @@
 package org.apache.paimon.spark
 
 import org.apache.paimon.CoreOptions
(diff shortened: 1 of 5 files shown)


paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/SparkSource.scala
```diff
@@ -82,24 +82,6 @@ class SparkSource
     SparkSource.toBaseRelation(table, sqlContext)
   }
 
-  private def loadTable(options: JMap[String, String]): DataTable = {
-    val path = CoreOptions.path(options)
-    val catalogContext = CatalogContext.create(
-      Options.fromMap(
-        mergeSQLConfWithIdentifier(
-          options,
-          extractCatalogName().getOrElse(NAME),
```
Zouxxyy, line 91: The `extractCatalogName()` and `NAME` here actually has no meaning and has nothing to do with the catalog name

PR #6052, opened 2025-08-11
[bug] Fix bytes bug and isort

discivigour:
### Purpose

- fix bytes bug.
- change 'data.token.enabled' to 'data-token.enabled'.
- add isort command in the script.


JingsongLi (approved):
+1

PR #6060, opened 2025-08-12
[parquet] Support generate the RowRanges from the selection

Tan-JiaLiang:
### Purpose

Currently, file index results can be used to filter Apache Parquet pages, but they must rely on the ColumnIndexFilter.

This PR aims to generate the Parquet Rowranges directly from the file index results without relying on the ColumnIndexFilter. This has the following advantages:
1. TopN results can be pushed down to Parquet to filter pages.
2. Limit results can be pushed down to Parquet to filter pages.
3. It is also suitable for supporting file index predicates fully pushdown from the engine in the future.


paimon-format/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -792,23 +796,67 @@ public ColumnIndexStore getColumnIndexStore(int blockIndex) {
...
+            offsets =
+                    paths.keySet().stream().map(store::getOffsetIndex).collect(Collectors.toList());
+        }
+
+        long rowCount = block.getRowCount();
+        long rowIndexOffset = block.getRowIndexOffset();
+        RowRanges rowRanges = RowRanges.createSingle(rowCount);
+        for (OffsetIndex offset : offsets) {
+            if (offset != null) {
+                RowRanges result = RowRanges.create(rowCount, rowIndexOffset, offset, selection);
```
JingsongLi, line 852: Why we need to handle `OffsetIndex`? Base on `selection` we can just create `RowRanges`?

JingsongLi: Because we need to avoid creating too more ranges? Based on Pages?

paimon-format/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -792,23 +796,67 @@ public ColumnIndexStore getColumnIndexStore(int blockIndex) {
...
+            }
             blockRowRanges.set(blockIndex, rowRanges);
         }
         return rowRanges;
     }
 
+    private RowRanges calculateRowRanges(int blockIndex, RoaringBitmap32 selection) {
+        List<OffsetIndex> offsets;
+        BlockMetaData block = blocks.get(blockIndex);
+        if (paths.isEmpty()) {
```
JingsongLi, line 830: Can we just skip for empty?

paimon-format/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
Tan-JiaLiang: Yes, If using the selection directly, too much rowranges fragment.
e.g. selection: [1,3,5,7,9] ==> rowranges: [[1,1], [3,3], [5,5], [7,7], [9,9]]

paimon-format/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
JingsongLi: For example, return `Optional<RowRanges>`.

Tan-JiaLiang: I want to using the selection to let the limit pushdown to filter the pages in the another PR.
If skip here, selection can not works in the following example, and it will be fallback to read all pages.
```sql
SELECT COUNT(*) FROM T LIMIT 10;
```

JingsongLi: `SELECT COUNT(*) FROM T LIMIT 10;` will just use pushAggregation and just read manifests?

JingsongLi: Let's finish this in another PR.

PR #6061, opened 2025-08-12
[core] optimize validate fields functions

qingwei727:
### Purpose
Optimize the time-consuming process of creating large lists; the validateFields function currently takes O(n²) time due to checking for duplicate fields, so optimize the validateFields function to run in O(n) time.


sundapeng:
+1

JingsongLi:
+1

PR #6064, opened 2025-08-12
[fs] Set credentials provider in JindoFileIO

plusplusjiajia:
### Purpose
Set SimpleCredentialsProvider as default credentials provider for JindoFileIO to use right token to access OSS.


sundapeng:
+1

PR #6066, opened 2025-08-12
[common]Replace illegal imports with shaded versions

xiaochen-zhou:
### Purpose

Currently, `mvn spotless:apply` does not support replacing shaded packages, which may result in code style checks failing after execution.

image

### Tests

Add SpotlessRegexReplacementTest.java


JingsongLi:
+1

PR #6024, opened 2025-08-04
[flink] Automatically detect changes of external-paths strategy

LsomeYeah:
### Purpose

Linked issue: close #xxx

The Flink job needs to be restarted for changes to take effect when changing the configration of external-paths. With rapid data lake growth and frequent storage expansions, external-paths changes occur often,  therefore some users require updates without restarting flink streaming job.

This pr checks latest schema every checkpoint, and if the schema has changed, it'll try update the `write` with the latest external-paths config. You can configure 'external-paths.detect-config.enabled' = 'true' to enable this feature. Cause it tries to read latest schema every chekpoint, it'll bring extra IO cost.

Currently, this feature takes affects for flink streaming writing  job and table-level dedicated compaction job.

### Tests

Streaming write:

1. AppendOnlyTableITCase#testTableReadWriteWithChangedExternalPath
2. PrimaryKeyFileStoreTableITCase#testTableReadWriteWithChangedExternalPath

table-level dedicated compaction:

1. CompactActionITCase#testStreamingCompactWithChangedExternalPath
2. CompactActionITCase#testUnawareBucketStreamingCompactWithChangedExternalPath

PR #6070, opened 2025-08-13
[flink] Refactor WriterRefresher to better abstraction

JingsongLi:
### Purpose

After https://github.com/apache/paimon/pull/6024


LsomeYeah (approved):
+1, thanks for this better abstraction!

PR #6073, opened 2025-08-13
[core] Continue improving parameter naming in HadoopSecuredFileSystem

plusplusjiajia:
### Purpose
 Continue improving parameter naming in HadoopSecuredFileSystem


JingsongLi:
+1

PR #6079, opened 2025-08-15
[core] python: support rest commit

jerry-024:
### Purpose

Linked issue: close #xxx


paimon-python/pypaimon/common/lock.py
```diff
@@ -0,0 +1,70 @@
...
+# limitations under the License.
+################################################################################
+
+from abc import ABC, abstractmethod
+from typing import Callable, TypeVar
+
+T = TypeVar('T')
+
+
+class Lock(ABC):
```
JingsongLi: Remove Lock in this PR. We don't need it now.

PR #6083, opened 2025-08-17
[spark] Only read necessary meta columns when building new scan plan

Zouxxyy:
### Purpose

Only read necessary meta columns when building new scan plan


paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/util/ScanPlanHelper.scala
```diff
@@ -0,0 +1,89 @@
...
+    val outputNames = relation.outputSet.map(_.name)
+
+    def isOutputColumn(colName: String) = {
+      val resolve = conf.resolver
+      outputNames.exists(resolve(colName, _))
+    }
+
+    val appendMetaColumns = sparkTable.metadataColumns
+      .map(_.asInstanceOf[PaimonMetadataColumn].toAttribute)
+      .filter(col => isOutputColumn(col.name))
```
Zouxxyy, line 80: the key modification is here, we should use `isOutputColumn` instead of `!isOutputColumn`

Issue #6022, opened 2025-08-04
[Feature] Support row lineage.


leaves12138:
<h1>Motivation</h1><p>In batch mode, "MERGE INTO" or "UPDATE" operations frequently occur. Rows may variety from snapshots. We have no idea how to track the row variation.</p><p>It's useful to find the whole row lineage for user if they must track the source for one abnormal record. For example, if we meet one abnormal record, we may wanna</p><p>how it produce and how it changes to today we see. To better support these, we would like to add a new table function to tag the datas' lineage.</p><p><br /></p>

See pip: https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=373886618

leaves12138:
Core pull request: 
https://github.com/apache/paimon/pull/5935

Spark related: 
https://github.com/apache/paimon/pull/5985
https://github.com/apache/paimon/pull/6014

PR #6086, opened 2025-08-18
[spark] Support DELETE for row tracking table

Zouxxyy:
### Purpose

Subtask of https://github.com/apache/paimon/issues/6022, support spark DELETE for row tracking table

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/commands/DeleteFromPaimonTableCommand.scala
+++ b/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/commands/DeleteFromPaimonTableCommand.scala
@@ -147,10 +147,13 @@ case class DeleteFromPaimonTableCommand(
 
       // Step4: build a dataframe that contains the unchanged data, and write out them.
       val toRewriteScanRelation = Filter(Not(condition), newRelation)
-      val data = createDataset(sparkSession, toRewriteScanRelation)
+      var data = createDataset(sparkSession, toRewriteScanRelation)
+      if (coreOptions.rowTrackingEnabled()) {
+        data = selectWithRowLineage(data)
+      }
 
       // only write new files, should have no compaction
-      val addCommitMessage = dvSafeWriter.writeOnly().write(data)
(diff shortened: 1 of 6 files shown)


paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/commands/DeleteFromPaimonTableCommand.scala
```diff
@@ -147,10 +147,13 @@ case class DeleteFromPaimonTableCommand(
...
       val toRewriteScanRelation = Filter(Not(condition), newRelation)
-      val data = createDataset(sparkSession, toRewriteScanRelation)
+      var data = createDataset(sparkSession, toRewriteScanRelation)
+      if (coreOptions.rowTrackingEnabled()) {
+        data = selectWithRowLineageMetaCols(data)
+      }
 
       // only write new files, should have no compaction
-      val addCommitMessage = dvSafeWriter.writeOnly().write(data)
+      val addCommitMessage = dvSafeWriter.withRowLineage().write(data)
```
JingsongLi: Should keep writeOnly?

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/commands/PaimonCommand.scala
```diff
@@ -54,7 +54,10 @@ trait PaimonCommand
   with SQLConfHelper {
 
   lazy val dvSafeWriter: PaimonSparkWriter = {
-    if (table.primaryKeys().isEmpty && table.bucketMode() == BucketMode.HASH_FIXED) {
+    if (
+      table.coreOptions().deletionVectorsEnabled() && table.primaryKeys().isEmpty && table
```
Zouxxyy: Currently, we always use dvSafeWriter regardless of whether DV is enabled or not. 

Here I add a new judgment.

JingsongLi: Why add dv? Do compact for bucketed Append table is dangerous.

PR #6093, opened 2025-08-19
[core]format table: support csv format

jerry-024:
### Purpose
This PR enhances the CSV format support in Paimon.

### Tests

- CsvFileFormatTest

PR #6095, opened 2025-08-19
[core]support format table read

jerry-024:
### Purpose

This PR introduces support for reading format tables in Paimon core. Building on the recent CSV (#6093) and JSON (#6096) format implementations, this enhancement enables comprehensive read operations for text-based format tables.

Key features:
- Unified format table reading infrastructure
- Support for multiple text-based formats (CSV, JSON)
- Integration with existing file store read operations
- Enhanced table query capabilities for format tables

This change extends Paimon's capability to handle diverse data formats while maintaining performance and consistency with the existing table reading mechanisms.

### Tests
- CatalogTestBase.testFormatTableRead


paimon-core/src/main/java/org/apache/paimon/table/format/FormatReadBuilder.java
```diff
@@ -0,0 +1,203 @@
...
+                        null,
+                        0,
+                        Collections.emptyMap());
+        return fileRecordReader;
+    }
+
+    // ===================== Unsupported ===============================
+    @Override
+    public ReadBuilder withLimit(int limit) {
+        throw new UnsupportedOperationException("limit is not supported for FormatTable.");
```
JingsongLi: `return this;` is ok.

paimon-core/src/main/java/org/apache/paimon/table/format/FormatReadBuilder.java
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    @Override
+    public ReadBuilder withTopN(TopN topN) {
+        throw new UnsupportedOperationException("TopN is not supported for FormatTable.");
```
JingsongLi, line 178: `return this;` is ok.

paimon-core/src/main/java/org/apache/paimon/table/format/FormatReadBuilder.java
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    @Override
+    public ReadBuilder dropStats() {
+        throw new UnsupportedOperationException("Format Table does not support dropStats.");
```
JingsongLi, line 183: `return this;` is ok.

JingsongLi (approved):
+1 Thanks @jerry-024

PR #6097, opened 2025-08-19
[core] Support TopN pushdown in deletion-vector mode

Tan-JiaLiang:
### Purpose

For now, we have been support the TopN evaluation in range-bitmap file index. The deletion-vector bitmap and the range-bitmap index are both base row position, we can merge the bitmap result and push it down to the DataFile.

So the Append/KeyValue table in deletion-vector mode can also support TopN pushdown.

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/fileindex/FileIndexPredicate.java
+++ b/paimon-common/src/main/java/org/apache/paimon/fileindex/FileIndexPredicate.java
@@ -71,8 +71,7 @@ public class FileIndexPredicate implements Closeable {
         this.reader = FileIndexFormat.createReader(inputStream, fileRowType);
     }
 
-    public FileIndexResult evaluate(
-            @Nullable Predicate predicate, @Nullable FileIndexResult input) {
+    public FileIndexResult evaluate(@Nullable Predicate predicate) {
         if (predicate == null) {
             return REMAIN;
         }
@@ -84,9 +83,6 @@ public class FileIndexPredicate implements Closeable {
             LOG.debug(
                     "One file has been filtered: "
                             + (path == null ? "in scan stage" : path.toString()));
-        }
-        if (input != null) {
-            result = result.and(input);
         }
         return result;
     }
--- a/paimon-common/src/main/java/org/apache/paimon/fileindex/FileIndexPredicate.java
+++ b/paimon-common/src/main/java/org/apache/paimon/fileindex/FileIndexPredicate.java
@@ -71,7 +71,8 @@ public class FileIndexPredicate implements Closeable {
         this.reader = FileIndexFormat.createReader(inputStream, fileRowType);
     }
 
-    public FileIndexResult evaluate(@Nullable Predicate predicate) {
+    public FileIndexResult evaluate(
+            @Nullable Predicate predicate, @Nullable FileIndexResult input) {
         if (predicate == null) {
             return REMAIN;
         }
@@ -83,6 +84,9 @@ public class FileIndexPredicate implements Closeable {
             LOG.debug(
                     "One file has been filtered: "
                             + (path == null ? "in scan stage" : path.toString()));
+        }
+        if (input != null) {
+            result = result.and(input);
         }
         return result;
     }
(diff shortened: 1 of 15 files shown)


paimon-core/src/main/java/org/apache/paimon/operation/RawFileSplitRead.java
```diff
@@ -247,21 +243,6 @@ private FileRecordReader<InternalRow> createFileReader(
...
 
-        RoaringBitmap32 deletion = null;
-        DeletionVector deletionVector = dvFactory == null ? null : dvFactory.get();
-        if (deletionVector instanceof BitmapDeletionVector) {
-            deletion = ((BitmapDeletionVector) deletionVector).get();
-        }
-
-        if (selection != null) {
-            if (deletion != null) {
-                selection = RoaringBitmap32.andNot(selection, deletion);
```
JingsongLi, line 258: Why delete this?

Tan-JiaLiang: It has been moved to the FileIndexEvaluator#evaluate.
```java
FileIndexResult result =
        new BitmapIndexResult(() -> RoaringBitmap32.bitmapOfRange(0, file.rowCount()));
if (deletionVector instanceof BitmapDeletionVector) {
    RoaringBitmap32 deletion = ((BitmapDeletionVector) deletionVector).get();
    result = ((BitmapIndexResult) result).andNot(deletion);
}
```

JingsongLi: This selection is for format pushdown? Not for file index filtering.

Tan-JiaLiang: In my opinion, it can be also for file index filtering. e.g.
row count is 6 ==> selection is [0,1,2,3,4,5], if deletion is [1,3,5]. After selection AND NOT deletion, the result is [0,2,4], so after merge, the selection is [0,2,4].

We can base the [0,2,4] position to pick the TopN and the Predicate Filter.

Tan-JiaLiang:
@JingsongLi Wait a minute, if deletion-vector only, should not pushdown.

paimon-core/src/main/java/org/apache/paimon/operation/RawFileSplitRead.java
JingsongLi: I got your point, already merged in `FileIndexEvaluator#evaluate`, return `selection` already contains dv.

paimon-common/src/main/java/org/apache/paimon/fileindex/FileIndexPredicate.java
```diff
@@ -71,7 +71,8 @@ public FileIndexPredicate(SeekableInputStream inputStream, RowType fileRowType)
         this.reader = FileIndexFormat.createReader(inputStream, fileRowType);
     }
 
-    public FileIndexResult evaluate(@Nullable Predicate predicate) {
+    public FileIndexResult evaluate(
+            @Nullable Predicate predicate, @Nullable FileIndexResult input) {
```
JingsongLi: Why add input here? You can just use `FileIndexPredicate.evaluate(predicate)` and use result to add input.

JingsongLi: Also modify `visitTopN`, it seems only need `SortValue value, int limit`. We don't need to pass `FileIndexResult` to the method.

Tan-JiaLiang: > Why add input here? You can just use FileIndexPredicate.evaluate(predicate) and use result to add input.

It ok to rollback, just want to same as the evaluateTopN.

> Also modify visitTopN, it seems only need SortValue value, int limit. We don't need to pass FileIndexResult to the method.

We need it. We need to pick TopN base the last step FileIndexResult(e.g. Selection AND NOT deletion).

Also, maybe in the future, if we support file index fully pushdown from the engine. The FileIndexResult is from the Predicate Filter. e.g. the column `type` EQ predicate result evaluate first, then the result pass to the TopN evaluation.
```sql
SELECT * FROM T WHERE type='click' ORDER BY time DESC LIMIT 10;
```

JingsongLi (approved):
+1

Issue #6002, opened 2025-07-31
[Bug] flakey test CloneActionITCase.testMigrateOneNonPartitionedTable


zhongyujiang:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Paimon version

master

### Compute Engine

-

### Minimal reproduce step

CloneActionITCase.testMigrateOneNonPartitionedTable occasionally fails.
e.g. : https://github.com/apache/paimon/actions/runs/16566765649/job/46848649523?pr=5968#step:5:11428

### What doesn't meet your expectations?

-

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #6100, opened 2025-08-19
[test] Fix flaky test for CloneActionITCase

ArnavBalyan:
- testMigrateOneNonPartitionedTable occasionally fails with "Broken pipe" HMS connectivity errors
 - Network/timing issues in CI cause transient Thrift connection failures
 - Simple retry with backoff gives HMS time to recover from temporary issues
 - Fixes: #6002


JingsongLi:
+1

PR #6101, opened 2025-08-19
[spark] Add temporary v1 function and UDAF support to SparkCatalog

Zouxxyy:
### Purpose

- Support temporary v1 function
- Suppot UDAF


JingsongLi (commented):
Also document an example to `spark/sql-functions/`.

PR #6103, opened 2025-08-20
[core] text support compress

jerry-024:
### Purpose

Linked issue: close #xxx


JingsongLi (commented):
We should introduce an unified compression for both csv and json.

paimon-core/src/main/java/org/apache/paimon/io/DataFilePathFactory.java
```diff
@@ -89,7 +90,13 @@ public Path newPath(String prefix) {
     private String newFileName(String prefix) {
         String extension;
         if (fileSuffixIncludeCompression) {
-            extension = "." + fileCompression + "." + formatIdentifier;
+            String compressionExtension =
+                    CompressionType.fromValue(fileCompression).fileExtension();
+            if (isTextFormat(formatIdentifier)) {
```
JingsongLi: is Text, must add `compressionExtension`.

PR #6104, opened 2025-08-20
[core] update null-literal default value for csv format

jerry-024:
### Purpose
1. update null-literal default value `"null"` to `""` for csv format
2.  add test for diff conf 
3. support BINARY and VARBINARY

### Tests
- CsvFileFormatTest


JingsongLi (approved):
+1

PR #6105, opened 2025-08-20
[core] Support limit pushdown to the DataFile

Tan-JiaLiang:
### Purpose

For now, `selection` can push down to filter the DataFile's rowgroups and pages, and the `limit` can trim the `selection`, then more rowgroups and pages will be filtered out.

1. support append-only table, append-only/primary-key table in deletion-vector mode.
2. limit can not using with `TopN` and `Predicate Filter`.

Benchmarks:
Before
```text
limit:                                                                                               Best/Avg Time(ms)    Row Rate(K/s)      Per Row(ns)   Relative
OPERATORTEST_limit_300000-1000                                                                             498 /  498           2007.2            498.2       1.0X
```

After
```text
limit:                                                                                               Best/Avg Time(ms)    Row Rate(K/s)      Per Row(ns)   Relative
OPERATORTEST_limit_300000-1000                                                                              27 /   27          37500.0             26.7       1.0X
```


Tan-JiaLiang (commented):
Go on with the https://github.com/apache/paimon/pull/6060 discussion.

cc @JingsongLi

PR #6108, opened 2025-08-21
[flink] Support read and write variant with Flink

Sxnan:
### Purpose

Support read and write variant with Flink


Sxnan:
@yunfengzhou-hub Can you have a look at this PR?

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/CatalogTableITCase.java
```diff
@@ -1219,6 +1222,24 @@ public void testIndexesTable() {
         assertThat(row.getField(6)).isNotNull();
     }
 
+    @Test
+    @EnabledIf("isFlink2_0")
```
yunfengzhou-hub: Variant and PARSE_JSON are introduced in Flink 2.1. Do we need to run this test case in Flink 2.0?

Sxnan: You are right that we should only run this case in Flink 2.1 and above

yunfengzhou-hub:
+1

PR #6109, opened 2025-08-21
[core] FileEntry supports file creation time

yuzelin:
### Purpose

Linked issue: close #xxx

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommit.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommit.java
@@ -36,6 +36,8 @@ public interface FileStoreCommit extends AutoCloseable {
 
     FileStoreCommit withPartitionExpire(PartitionExpire partitionExpire);
 
+    FileStoreCommit baseEntryContainsTime(boolean ignoreEmptyCommit);
+
     /** Find out which committables need to be retried when recovering from the failure. */
     List<ManifestCommittable> filterCommitted(List<ManifestCommittable> committables);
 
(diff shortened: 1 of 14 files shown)


JingsongLi:
Can you add a new FileEntry implementation class instead of modifying FileEntry?

paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommit.java
```diff
@@ -36,6 +36,8 @@ public interface FileStoreCommit extends AutoCloseable {
 
     FileStoreCommit withPartitionExpire(PartitionExpire partitionExpire);
 
+    FileStoreCommit baseEntryContainsTime(boolean ignoreEmptyCommit);
```
JingsongLi, line 39: ignoreEmptyCommit?

paimon-core/src/main/java/org/apache/paimon/operation/FileStoreScan.java
```diff
@@ -106,11 +106,15 @@ default Long totalRecordCount(Snapshot snapshot) {
...
+    default List<SimpleFileEntry> readSimpleEntries() {
+        return readSimpleEntries(false);
+    }
+
     /**
      * Read {@link SimpleFileEntry}s, SimpleFileEntry only retains some critical information, so it
      * cannot perform filtering based on statistical information.
      */
-    List<SimpleFileEntry> readSimpleEntries();
+    List<SimpleFileEntry> readSimpleEntries(boolean containsTime);
```
JingsongLi, line 117: Can you introduce another method `readSimpleEntries(boolean containsTime)`? Keep old `readSimpleEntries()`.

Issue #5872, opened 2025-07-10
[Bug]


wuwenchi:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Paimon version

1.2.0

### Compute Engine

flink 1.20.1

### Minimal reproduce step

very simple with 3 steps.

1. create a table with primary key：
```sql
create table tb98 (c1 int primary key not enforced, c2 int);
```

2. Two writers are attempting to update the data with primary key 1 simultaneously.
One is updating it to 8, and the other is updating it to 9.
 ：
```sql
insert into tb98 values (1, 8);
```
```sql
insert into tb98 values (1, 9);
```

3. call `compaction` procedures.

<img width="1594" height="1396" alt="Image" src="https://github.com/user-attachments/assets/80ee71bc-c701-4072-88a6-37eedfdf3222" />

### What doesn't meet your expectations?

The value of primary key 1 queried from snapshot-2 and snapshot-3 is 9, but it changes to 8 after compaction. Compaction is only supposed to compress data and should not modify the data, right?

Therefore, if the data was 9 before compaction, it should remain 9 after compaction.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #6111, opened 2025-08-21
[core] minSequenceNumber for deterministic file ordering in compaction

ArnavBalyan:
- Prevents data corruption during concurrent writes when compaction is done.
 - Ensures deterministic compaction behavior by using minSequenceNumber and create time
 - Fixes: #5872


JingsongLi:
Good solution! +1

Issue #6107, opened 2025-08-20
[Feature] [core] Iceberg:  createMetadataWithoutBase should include full history of Paimon snapshots


nickdelnano:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Motivation

Code path:  https://github.com/apache/paimon/blob/ff02f6bf3ceccf8dcac38bc58cf6db390509bd46/paimon-core/src/main/java/org/apache/paimon/iceberg/IcebergCommitCallback.java#L305C49-L305C59

createMetadataWithoutBase is called under certain conditions like
- enabling Iceberg compatibility for the first time
- committing Iceberg metadata after a previous commit failed in the Iceberg layer (for many possible reasons)

In some cases the Paimon table will have many snapshots already (e.g. snapshots 1 to 1000) and after calling createMetadataWithoutBase, only the latest Paimon commit will be synced to Iceberg.

By syncing the whole Paimon history to Iceberg, the Iceberg compatibility feature becomes suitable for production use cases that require Iceberg time travel.

My use case is this:
- Sync MySQL tables to Paimon using Flink CDC
- Tag daily Paimon snapshots [automatically](https://paimon.apache.org/docs/master/maintenance/manage-tags/)
- Iceberg readers read daily snapshots

When a failure happens in the Iceberg committer the metadata needs to be recreated on the next commit. Only the latest Paimon snapshot is included. The daily snapshot component of this pipeline has broken and cannot be recovered.

### Solution

A simple solution may exist. Assume snapshot range [x, y]:

If creating metadata without base:
- createMetadataWithoutBase for the earliest snapshot, x
- (for i=x, i <= y, i++) call createMetadataWithBase(i)

### Anything else?

Consider making this feature opt-in with configuration as it may be a costly operation to sync many Paimon snapshots to Iceberg and therefore reach the Flink checkpoint timeout.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

nickdelnano:
@LsomeYeah what do you think about this issue and the suggested solution?

LsomeYeah:
> [@LsomeYeah](https://github.com/LsomeYeah) what do you think about this issue and the suggested solution?

@nickdelnano I think this case is reasonable, and I also prefer adding a new option.

PR #6121, opened 2025-08-22
[core] Iceberg:  Only add tags that reference existing Iceberg snapshots

nickdelnano:
### Purpose

Only add Iceberg tags when the referenced snapshot exists in Iceberg. Otherwise all reads on the Iceberg table fail which is a big problem for Iceberg users.

This change makes the Iceberg layer more reliable as condition (1) described in the issue will happen due to unavoidable failures in a distributed system.

Linked issue: https://github.com/apache/paimon/issues/6107

### Tests
UT tests, existing IT tests cover basic tag behavior

### Documentation
Updated


LsomeYeah:
@nickdelnano ok, I will review it as soon.

paimon-core/src/main/java/org/apache/paimon/iceberg/IcebergCommitCallback.java
```diff
@@ -1030,6 +1043,24 @@ public void notifyCreation(String tagName, long snapshotId) {
...
+                    break;
+                }
+            }
+
+            if (!tagSnapshotInIceberg) {
+                LOG.info(
+                        "Snapshot {} does not exist in Iceberg metadata. Unable to create tag {}.",
+                        snapshotId,
+                        tagName);
+                return;
```
LsomeYeah: NIT: LOG.warn or LOG.error

LsomeYeah (approved):
Looks good to me, just leave one nit.

nickdelnano:
thank you @LsomeYeah !

nickdelnano:
@JingsongLi I have a fix for iceberg compatibility here, could you help to review/merge it?

PR #6124, opened 2025-08-22
[flink] Restart dedicated compact job if schema change is detected

tsreaper:
### Purpose

Currently, streaming compact job does not deal with schema change. If the user adds a column, it will be ignored by the compact job and this column will be lost.

This PR restarts the compact job if schema change is detected, so new columns will not be lost.

### Tests

* `CompactProcedureITCase#testFailWhenSchemaChanged`

### API and Format

No format changes.

### Documentation

No new feature.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/system/CompactBucketsTable.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/system/CompactBucketsTable.java
@@ -80,8 +80,8 @@ public class CompactBucketsTable implements DataTable, ReadonlyTable {
 
     private final FileStoreTable wrapped;
     private final boolean isContinuous;
-
     @Nullable private final String databaseName;
+    private final long baseSchemaId;
 
     private static final RowType ROW_TYPE =
             RowType.of(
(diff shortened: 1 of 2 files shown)


JingsongLi (commented):
Also modify `AppendTableCompactSource`?

paimon-core/src/main/java/org/apache/paimon/table/system/CompactBucketsTable.java
```diff
@@ -260,6 +261,19 @@ public RecordReader<InternalRow> createReader(Split split) throws IOException {
             }
 
             DataSplit dataSplit = (DataSplit) split;
+            // in case of schema evolution
+            for (DataFileMeta file : dataSplit.dataFiles()) {
+                System.out.println(
+                        "File schema id " + file.schemaId() + ", base schema id " + baseSchemaId);
+                if (file.schemaId() > baseSchemaId) {
+                    throw new RuntimeException(
```
JingsongLi, line 269: Just failover? Will writer use new schema to write?

JingsongLi:
+1

PR #6132, opened 2025-08-24
[core] Introduce 'index-file-in-data-file-dir' to put index files in data dir

JingsongLi:
### Purpose

In a particularly large-scale table, if the deletion vector is enabled, such as with thousands of buckets, saving a large number of snapshots will generate millions of index files.

Millions of index files in one directory can put significant pressure on the file system, and we also support external paths. We need to support 'index-file-in-data-file-dir' to alleviate this pressure.

### Tests

In `DeletionVectorITCase`
- `testIndexFileInDataFileDir`
- `testIndexFileInIndexDir`
- `testIndexFileInDataFileDirWithExternalPath`


leaves12138 (approved):
+1

PR #6134, opened 2025-08-24
[core] Make ManifestEntry and DataFileMeta to interfaces

JingsongLi:
### Purpose

Afterwards, we can consider introducing row-based `ManifestEntry` to significantly reduce its serialization and deserialization overhead.


leaves12138 (approved):
+1

PR #6144, opened 2025-08-25
[parquet] Add encryption for parquet

Zouxxyy:
### Purpose

Add encryption for parquet


JingsongLi:
It seems that neither Iceberg nor Delta supports direct Parquet encryption...

PR #6145, opened 2025-08-26
[core] fix: could not find compression when write throw exception

jerry-024:
### Purpose
If it could not find compression when writing, it throws an exception, as the  file name has a compression extension

### Tests
-

### DIFF
--- a/paimon-format/src/main/java/org/apache/paimon/format/text/HadoopCompressionUtils.java
+++ b/paimon-format/src/main/java/org/apache/paimon/format/text/HadoopCompressionUtils.java
@@ -30,7 +30,6 @@ import org.apache.hadoop.io.compress.CompressionCodecFactory;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
-import java.util.Optional;
 
 /** Utility class for handling file compression and decompression using Hadoop codecs. */
 public class HadoopCompressionUtils {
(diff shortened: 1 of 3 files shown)


paimon-format/src/test/java/org/apache/paimon/format/TextCompressionTest.java
```diff
@@ -65,6 +66,41 @@ public abstract class TextCompressionTest {
...
+        Path filePath = new Path(tempDir.resolve(fileName).toString());
+        FileIO fileIO = new LocalFileIO();
+
+        // Attempt to write with unsupported compression should handle gracefully
+        FormatWriterFactory writerFactory = format.createWriterFactory(rowType);
+
+        // This should either:
+        // 1. Fall back to no compression and succeed, or
+        // 2. Throw an appropriate exception
+        try (FormatWriter writer =
```
JingsongLi: Use `assertThrows`.

paimon-format/src/main/java/org/apache/paimon/format/text/HadoopCompressionUtils.java
```diff
@@ -71,34 +86,8 @@ public static InputStream createDecompressedInputStream(
...
-        }
-    }
-
-    /**
-     * Gets a compression codec by compression type.
-     *
-     * @param compression The compression type
-     * @return Optional CompressionCodec instance
-     */
-    public static Optional<CompressionCodec> getCompressionCodecByCompression(String compression) {
```
JingsongLi: Why to modify this?

paimon-format/src/test/java/org/apache/paimon/format/text/HadoopCompressionUtilsTest.java
```diff
@@ -0,0 +1,368 @@
...
+
+            assertThat(compressedStream).isNotNull();
+
+            // Write and close to ensure compression happens
+            compressedStream.write(TEST_DATA.getBytes(StandardCharsets.UTF_8));
+            compressedStream.close();
+
+            // Verify that some data was written
+            assertThat(byteArrayOutputStream.toByteArray()).isNotEmpty();
+        } catch (IOException e) {
```
JingsongLi: Do not use try catch to validate exception.

PR #6146, opened 2025-08-26
[flink] Add log for partition mark done trigger

leaves12138:
### Purpose

Asking why my partition does not expire among too many users. So add these logs for debug

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/listener/PartitionMarkDoneTrigger.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/listener/PartitionMarkDoneTrigger.java
@@ -132,19 +132,22 @@ public class PartitionMarkDoneTrigger {
         if (timeInterval == null || idleTime == null) {
             return Collections.emptyList();
         }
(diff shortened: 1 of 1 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/listener/PartitionMarkDoneTrigger.java
```diff
@@ -165,14 +169,29 @@ List<String> donePartitions(
                                 .toInstant()
                                 .toEpochMilli();
             }
+            LOG.debug("Partition {} start time: {}", partition, partitionStartTime);
             long partitionEndTime = partitionStartTime + timeInterval;
+            LOG.debug("Partition {} end time: {}", partition, partitionEndTime);
             lastUpdateTime = Math.max(lastUpdateTime, partitionEndTime);
+            LOG.debug("Partition {} last update time after compare: {}", partition, lastUpdateTime);
```
JingsongLi: Can you merge these logs? Just one line.

leaves12138: OK

PR #6149, opened 2025-08-26
[spark] Support partial insert for data evolution merge into

leaves12138:
### Purpose

Support data evolution mode merge into partial columns exist.


JingsongLi (approved):
+1

PR #6151, opened 2025-08-26
[format] update: format createReaderFactory add dataSchemaRowType

jerry-024:
### Purpose
In CSV format, we need dataSchemaRowType when projecting


JingsongLi (approved):
+1

PR #6152, opened 2025-08-26
[flink] Local TZ convert parity for Paimon / Hive

### LINKED ISSUE #5722: [Bug] timestamp reading from paimon is different from hive after hive-cloning
### Search before asking

### Paimon version

Paimon's version: 1.2-snapshot(2025/06/09)

### Compute Engine

Hive's version: hive 2.3.10

Flink's version: 1.20.1

### Minimal reproduce step

When selecting the paimon table  which was cloned from hive through hive-clone action, the selected result for timestamp is different between paimon and hive. 

**Note: file format is parquet.**

Minimal reproduce step:
1. create a table in hive and insert some data:
```sql
-- create a table
CREATE TABLE test.hive_test_table(
    `a` int COMMENT 'The a field', 
    `ts` timestamp COMMENT 'The ts field'
) STORED AS PARQUET;

-- insert a value
INSERT INTO test.hive_test_table VALUES (1, '2025-06-03 16:00:00');

-- select
SELECT * FROM test.hive_test_table;
-- result: 1, 2025-06-03T16:00
```

2. run hive-clone action

3. select the cloned paimon table in flink
```sql
SELECT * FROM paimon_hive_test_table;
-- result: 1, 2025-06-03T08:00
```

The reaseon may be:
Hive will transform the timestamp to UTC for storage, and transform the stored timestamp to `writer.time.zone` or system default time zone when reading. While paimon will directly read the original timestamp stored in parquet file without any transforms, this leads to difference.

The selected result for timestamp has changed after cloning to paimon.

### Anything else?

_No response_

ArnavBalyan:
- When reading parquet, hive converts the TZ to local timezone while querying which causes result mismatch against paimon
 - Add support to do the same when querying through paimon behind a feature flag. Behaviour will be enabled if user enables the config.
 - Fixes: #5722

### DIFF
--- a/paimon-format/src/main/java/org/apache/paimon/format/parquet/ParquetReaderFactory.java
+++ b/paimon-format/src/main/java/org/apache/paimon/format/parquet/ParquetReaderFactory.java
@@ -96,6 +96,7 @@ public class ParquetReaderFactory implements FormatReaderFactory {
                                 context.fileIO(), context.filePath(), context.fileSize()),
                         builder.build(),
                         context.selection());
+
         MessageType fileSchema = reader.getFileMetaData().getSchema();
         MessageType requestedSchema = clipParquetSchema(fileSchema);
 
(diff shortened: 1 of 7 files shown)


paimon-common/src/main/java/org/apache/paimon/utils/TimezoneOptionUtils.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package org.apache.paimon.utils;
+
+import org.apache.paimon.options.Options;
+
+/** Helper for timestamp timezone conversion option. */
+public final class TimezoneOptionUtils {
+
+    public static final String CONVERSION_ENABLED = "timestamp.timezone.conversion.enabled";
```
JingsongLi: Maybe introduce an option to `ParquetOptions`, named `parquet.timestamp-int96-adjust-zone`.

paimon-format/src/main/java/org/apache/paimon/format/parquet/ParquetReaderFactory.java
```diff
@@ -115,8 +117,20 @@ public FileRecordReader<InternalRow> createReader(FormatReaderFactory.Context co
...
+
         return new VectorizedParquetRecordReader(
-                context.filePath(), reader, fileSchema, fields, writableVectors, batchSize);
+                context.filePath(),
+                reader,
+                fileSchema,
+                fields,
+                writableVectors,
+                batchSize,
+                timezoneConversionEnabled,
```
JingsongLi: Just pass a arguement: `adjustInt96Timestamp`.

PR #6153, opened 2025-08-26
[flink] Fail fast validation for delete with write-only mode enabled

ArnavBalyan:
- DeleteAction now checks CoreOptions.writeOnly() at the start of run.
 - If 'write-only' = 'true', it throws a clear UnsupportedOperationException.
 - avoids deep FileStoreCommitImpl manifest conflict errors.
 - Added testDeleteActionWriteOnlyRejected to assert the new guard.
 - Related to #6137

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/DeleteAction.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/DeleteAction.java
@@ -54,7 +54,12 @@ public class DeleteAction extends TableActionBase {
 
     @Override
     public void run() throws Exception {
-        CoreOptions.MergeEngine mergeEngine = CoreOptions.fromMap(table.options()).mergeEngine();
+        CoreOptions coreOptions = CoreOptions.fromMap(table.options());
+        if (coreOptions.writeOnly()) {
+            throw new UnsupportedOperationException(
+                    "DELETE is not supported when 'write-only'='true'. Remove the hint or set it to false for the target table.");
+        }
+        CoreOptions.MergeEngine mergeEngine = coreOptions.mergeEngine();
         if (mergeEngine != DEDUPLICATE) {
             throw new UnsupportedOperationException(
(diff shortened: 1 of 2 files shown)


ASiegeLion:
f I add ‘write-only = true‘’ to the with attribute of the paimon table,
example :
`CREATE TABLE paimon.dw_liupeiyue.yydw_dm_entity_person_paimon_mini(
`hdid` string,
`attribute` string,
`last_update_dt` string,
`proctime` AS PROCTIME(),
PRIMARY KEY(`hdid`) NOT ENFORCED
) COMMENT ''
WITH (
'bucket' = '64',
'snapshot.time-retained' = '24h',
'write-only' = 'true',
'metastore.partitioned-table' = 'true',
'consumer.expiration-time' = '3d'
)`

executor flink batch sql ：
`delete from paimon.dw_liupeiyue.yydw_dm_entity_person_paimon_mini  where  hdid = '10074379'`

flink web ui  show ：
image

so, delete from should be supported write-only=true mode

JingsongLi (commented):
Clean improvement - fail-fast validation for delete operations in write-only mode.

1. **Problem**: When `write-only=true` is set, DeleteAction would proceed deep into the commit path before failing with confusing manifest conflict errors.

2. **Fix**: Check `CoreOptions.writeOnly()` at the start of `DeleteAction.run()` and throw a clear `UnsupportedOperationException` immediately.

3. **Implementation**: +43/-1, well-scoped. The test `testDeleteActionWriteOnlyRejected` validates the new guard.

4. **UX improvement**: Users now get a clear error message instead of a cryptic internal failure.

LGTM - simple, well-motivated, and properly tested.

JingsongLi:
Reverted in db6e644627b32a1af7f5b4aa166778ce20d1d482

We should not break behavior in write-only. This is incompatible.

ArnavBalyan:
Thanks for taking a look, I'll follow up with having an optional config flag and not reject the whole operation. Will open a follow up shortly. Apologies for the churn.

PR #6158, opened 2025-08-27
[iceberg] Add comprehensive tests for IcebergDataField and fix nested type nullability handling

plusplusjiajia:
### Purpose
Fix IcebergDataField nested type nullability handling.
### Tests
paimon-core/src/test/java/org/apache/paimon/iceberg/metadata/IcebergDataFieldTest.java


JingsongLi:
+1

PR #6164, opened 2025-08-27
[core] Add validation for default value

Zouxxyy:
### Purpose

- ~Add current_timestamp and current_date default value~
- Add default value validate


JingsongLi (approved):
+1

PR #6170, opened 2025-08-28
[Python] Avoid committing JARs by cloning them during setup

chenghuichen:
### Purpose

This PR removes the two JAR files that were previously committed directly for testing.


JingsongLi:
We should not rely on old legacy pypaimon py4j implementation, and for this maintain 1000+ lines is not good.

PR #6171, opened 2025-08-28
[python] support python 3.6

jerry-024:
### Purpose

Linked issue: close #xxx


paimon-python/setup.py
```diff
@@ -23,13 +23,15 @@
...
 install_requires = [
     'readerwriterlock==1.0.9',
-    'fsspec==2024.3.1',
-    'cachetools==5.3.3',
-    'ossfs==2023.12.0',
-    'pyarrow==16.0.0',
-    'polars==1.32.0',
-    'fastavro==1.11.1',
-    'zstandard==0.24.0'
+    'fsspec==2021.10.1',
```
JingsongLi, line 26: If for high python version, can we use dependencies with high version?

jerry-024: I will check it.

PR #6173, opened 2025-08-29
[core] Fix checkpoint recovery failure for compacted changelog files

uicosp:
[core] Fix checkpoint recovery failure for compacted changelog files

  ### Purpose

  Fixes checkpoint recovery failures when using precommit-compact functionality introduced in commit [flink] add coordinate and worker operator for small changelog files compaction (#4380).

  **Root Cause:**
  Compacted changelog files have two types of file names:
  1. Real files: `compacted-changelog-xxx$bid-len.cc-format`
  2. Fake files: `compacted-changelog-xxx$bid-len-off-len2.cc-format`

  Fake file names point to segments of real files but don't exist in the filesystem. The `checkFilesExistence` method was directly checking these fake file paths, causing recovery failures.

  **Solution:**
  - Created `CompactedChangelogPathResolver` utility class to resolve fake file paths to real file paths
  - Modified `TableCommitImpl.checkFilesExistence()` to resolve all paths before checking existence
  - Added deduplication logic since multiple fake files may resolve to the same real file
  - Path resolution rules:
    - Real files (`xxx$bid-len.cc-format`): return original path
    - Fake files (`xxx$bid-len-off-len2.cc-format`): resolve to `bucket-bid/xxx$bid-len.cc-format`

  ### Tests

  - Added unit tests in `CompactedChangelogPathResolverTest` to verify path resolution logic
  - Existing checkpoint recovery tests should now pass with compacted changelog files

  ### API and Format

  No changes to public API or storage format. This is an internal fix for file path resolution.

  ### Documentation

  No new features introduced. This is a bug fix for existing precommit-compact functionality.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/utils/CompactedChangelogPathResolver.java
+++ b/paimon-core/src/main/java/org/apache/paimon/utils/CompactedChangelogPathResolver.java
@@ -1,0 +1,67 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.paimon.utils;
+
+import org.apache.paimon.fs.Path;
+
+/**
+ * Utility class for resolving compacted changelog file paths.
+ *
+ * <p>This class provides functionality to resolve fake compacted changelog file paths to their real
+ * file paths, following the same protocol as CompactedChangelogFormatReaderFactory.
+ */
+public class CompactedChangelogPathResolver {
+
+    /**
+     * Resolves compacted changelog file path to its real file path. Handles both real and fake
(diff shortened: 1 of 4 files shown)


paimon-core/src/main/java/org/apache/paimon/utils/CompactedChangelogPathResolver.java
```diff
@@ -0,0 +1,67 @@
...
+
+package org.apache.paimon.utils;
+
+import org.apache.paimon.fs.Path;
+
+/**
+ * Utility class for resolving compacted changelog file paths.
+ *
+ * <p>This class provides functionality to resolve fake compacted changelog file paths to their real
+ * file paths, following the same protocol as CompactedChangelogFormatReaderFactory.
```
JingsongLi: Can you also move logic codes into this class from `CompactedChangelogFormatReaderFactory`? It is better to keep protocol in one single class.

paimon-core/src/main/java/org/apache/paimon/utils/CompactedChangelogPathResolver.java
```diff
@@ -0,0 +1,67 @@
...
+    /**
+     * Resolves compacted changelog file path to its real file path. Handles both real and fake
+     * compacted changelog file names as described in CompactedChangelogFormatReaderFactory.
+     *
+     * @param path the file path to resolve
+     * @return the resolved real file path
+     */
+    public static Path resolveCompactedChangelogPath(Path path) {
+        // Check if this is a compacted changelog file by its .cc- extension
+        String[] nameAndFormat = path.getName().split("\\.");
```
JingsongLi: Maybe just create a method `isCompactedChangelogPath`. And just check it is starting with `compacted-changelog-`?

JingsongLi:
Thanks @uicosp for the contribution! Left comments.

uicosp:
Refactor: 
1. Extract decodePath logic from CompactedChangelogFormatReaderFactory into CompactedChangelogPathResolver.  
2. Introduce isCompactedChangelogPath to check whether a file is a compacted changelog (by prefix "compacted-changelog-").

tsreaper (approved):

Issue #6183, opened 2025-09-01
[Bug] full-compaction.delta-commits


junmuz:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Paimon version

77f130bde03381e3b26b86e4f0e95cff5c1dc622

### Compute Engine

Flink

### Minimal reproduce step

I created a pkey Paimon table with the following table properties with Iceberg compatibility enabled.

```
changelog-producer: lookup
full-compaction.delta-commits: 1
metadata.iceberg.storage-location: table-location
metadata.iceberg.manifest-legacy-version: true
metadata.iceberg.manifest-compression: snappy
metadata.iceberg.previous-versions-max: 24
metadata.iceberg.storage: hive-catalog
metadata.iceberg.hive-client-class: com.amazonaws.glue.catalog.metastore.AWSCatalogMetastoreClient
metadata.iceberg.glue.skip-archive: true
metadata.iceberg.uri: ""
metadata.iceberg.hive-conf-dir: ""
metadata.iceberg.hadoop-conf-dir: ""
```

I performed the inserts and update operations on the above created table.

### What doesn't meet your expectations?

After performing the updates, I don't see the updates getting reflected from Iceberg compatibility layer. I was following the source code, and it appears that the delta-commits trigger only happens when the changelog-producer is set to full-compaction. The delta-commits is handled only inside the [GlobalFullCompactionSinkWrite](https://github.com/apache/paimon/blob/77f130bde03381e3b26b86e4f0e95cff5c1dc622/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/GlobalFullCompactionSinkWrite.java#L172-L176), which only gets invoked when the [changelog-producer is full-compaction](https://github.com/apache/paimon/blob/77f130bde03381e3b26b86e4f0e95cff5c1dc622/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/StoreSinkWrite.java#L146-L146). Inside the docs as well, the full-compaction.delta-commits is only mentioned for [Full Compaction changelog-producer](https://paimon.apache.org/docs/1.2/primary-key-table/changelog-producer/#full-compaction).

[...]

PR #6184, opened 2025-09-01
[core] Don't allow full compaction with lookup changelog producer

ArnavBalyan:
- Currently there is no check against full compaction on lookup changelog producer. 
 - User reported unexpected behaviour in #6183
 - Disable setting both the properties together instead of the silent behaviour.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
+++ b/paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
@@ -139,6 +139,14 @@ public class SchemaValidation {
                             ChangelogProducer.LOOKUP));
         }
 
+        if (options.toMap().get(FULL_COMPACTION_DELTA_COMMITS.key()) != null
+                && changelogProducer == ChangelogProducer.LOOKUP) {
+            throw new UnsupportedOperationException(
+                    String.format(
+                            "'%s' property is not supported for '%s' changelog producer type.",
+                            FULL_COMPACTION_DELTA_COMMITS.key(), ChangelogProducer.LOOKUP));
+        }
+
         checkArgument(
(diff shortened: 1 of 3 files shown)


paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
```diff
@@ -139,6 +139,14 @@ public static void validateTableSchema(TableSchema schema) {
                             ChangelogProducer.LOOKUP));
         }
 
+        if (options.toMap().get(FULL_COMPACTION_DELTA_COMMITS.key()) != null
+                && changelogProducer == ChangelogProducer.LOOKUP) {
```
junmuz, line 143: I believe the condition should be `changelogProducer != ChangelogProducer.FULL_COMPACTION` as this property only works with full-compaction changelog producer.

JingsongLi (commented):
Good validation improvement. Preventing incompatible configuration combinations at schema creation time is better than silent incorrect behavior at runtime.

1. **Problem**: Setting both `changelog-producer=lookup` and `full-compaction.delta-commits` leads to unexpected behavior (issue #6183). These are incompatible options.

2. **Fix**: Added validation in `SchemaValidation` to reject this combination with a clear error message. Also updated documentation.

3. **Scope**: +36/-1 across 3 files (validation, test, docs) - minimal and well-structured.

4. **Testing**: `SchemaValidationTest` added to verify the new check.

5. **Doc update**: Good that the changelog-producer docs are updated to mention this restriction.

LGTM - fail-fast with a clear error message is always better than silent misbehavior.

JingsongLi:
Please rebase master

ArnavBalyan:
Rebased and raised through new changes here #7999 thanks!

PR #6191, opened 2025-09-03
[core] Introduce manifest entry cache grouped by partition and bucket

JingsongLi:
### Purpose

Significantly accelerate the performance of scan query cache with partition filter and bucket filter.

This PR has also optimized the situation for small files. Previously, small files needed to occupy a Page no matter what. This PR has optimized the size of the Page, allowing it to be reduced from 32KB or 64KB to a suitable size.


leaves12138 (approved):
+1

PR #6197, opened 2025-09-04
[core]fix: filter contains partition filter

jerry-024:
### Purpose

For the format table: a user may define partition filters in filters, so we need to get partition filters from filters when a user defines filters for reading

### Tests
- CatalogTestBase#testFormatTableRead

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/predicate/PredicateBuilder.java
+++ b/paimon-common/src/main/java/org/apache/paimon/predicate/PredicateBuilder.java
@@ -26,6 +26,7 @@ import org.apache.paimon.types.DataField;
 import org.apache.paimon.types.DataType;
 import org.apache.paimon.types.DecimalType;
 import org.apache.paimon.types.RowType;
+import org.apache.paimon.utils.Pair;
 import org.apache.paimon.utils.Preconditions;
 
 import javax.annotation.Nullable;
@@ -205,6 +206,21 @@ public class PredicateBuilder {
         List<Predicate> result = new ArrayList<>();
         splitCompound(And.INSTANCE, predicate, result);
         return result;
+    }
+
+    public static Pair<List<Predicate>, List<Predicate>> splitAndForPartitionAndNonPartitionFilter(
+            Predicate predicate, int[] fieldIdxToPartitionIdx) {
+        List<Predicate> partitionFilters = new ArrayList<>();
+        List<Predicate> nonPartitionFilters = new ArrayList<>();
+        for (Predicate p : PredicateBuilder.splitAnd(predicate)) {
+            Optional<Predicate> mapped = transformFieldMapping(p, fieldIdxToPartitionIdx);
+            if (mapped.isPresent()) {
+                partitionFilters.add(mapped.get());
+            } else {
+                nonPartitionFilters.add(p);
+            }
+        }
+        return Pair.of(partitionFilters, nonPartitionFilters);
     }
 
     public static List<Predicate> splitOr(@Nullable Predicate predicate) {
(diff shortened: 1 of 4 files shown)


paimon-common/src/main/java/org/apache/paimon/predicate/PredicateBuilder.java
```diff
@@ -207,6 +208,21 @@ public static List<Predicate> splitAnd(@Nullable Predicate predicate) {
         return result;
     }
 
+    public static Pair<List<Predicate>, List<Predicate>> splitAndForPartitionAndNonPartitionFilter(
```
leaves12138: Name too long, can you make it shorter? Just splitFilter with comment is fine.

paimon-core/src/main/java/org/apache/paimon/table/format/FormatReadBuilder.java
```diff
@@ -79,10 +81,27 @@ public RowType readType() {
...
+        List<Predicate> unPartitionFilters = partitionAndNonPartitionFilter.getRight();
+        if (partitionFilters.size() > 0 && this.partitionFilter == null) {
+            RowType partitionType = table.rowType().project(table.partitionKeys());
+            PartitionPredicate partitionPredicate =
+                    PartitionPredicate.fromPredicate(
+                            partitionType, PredicateBuilder.and(partitionFilters));
+            withPartitionFilter(partitionPredicate);
+        }
+
+        if (unPartitionFilters.size() > 0) {
```
leaves12138: nonPartitionFilters?

paimon-core/src/main/java/org/apache/paimon/table/source/snapshot/SnapshotReaderImpl.java
```diff
@@ -222,24 +221,16 @@ public SnapshotReader withFilter(Predicate predicate) {
...
+                splitAndForPartitionAndNonPartitionFilter(predicate, fieldIdxToPartitionIdx);
+        List<Predicate> partitionFilters = partitionAndNonPartitionFilter.getLeft();
+        List<Predicate> unPartitionFilters = partitionAndNonPartitionFilter.getRight();
         if (partitionFilters.size() > 0) {
             scan.withPartitionFilter(PredicateBuilder.and(partitionFilters));
         }
 
-        if (nonPartitionFilters.size() > 0) {
-            nonPartitionFilterConsumer.accept(scan, PredicateBuilder.and(nonPartitionFilters));
+        if (unPartitionFilters.size() > 0) {
```
leaves12138: nonPartitionFilters

paimon-core/src/main/java/org/apache/paimon/table/format/FormatReadBuilder.java
```diff
@@ -79,10 +81,27 @@ public RowType readType() {
...
-            this.filter = predicate;
-        } else {
-            this.filter = PredicateBuilder.and(this.filter, predicate);
+        int[] fieldIdxToPartitionIdx =
+                PredicateBuilder.fieldIdxToPartitionIdx(table.rowType(), table.partitionKeys());
+        Pair<List<Predicate>, List<Predicate>> partitionAndNonPartitionFilter =
+                splitAndForPartitionAndNonPartitionFilter(predicate, fieldIdxToPartitionIdx);
+        List<Predicate> partitionFilters = partitionAndNonPartitionFilter.getLeft();
+        List<Predicate> unPartitionFilters = partitionAndNonPartitionFilter.getRight();
+        if (partitionFilters.size() > 0 && this.partitionFilter == null) {
```
leaves12138: Can you check this.partitionFilter first?

I think we can do this splitting in scan method, just like ReadBuilderImpl. In case multiple times withFilter

PR #6198, opened 2025-09-04
[parquet] Catch MissingOffsetIndexException in ParquetFileReader

JingsongLi:
### Purpose

For newer Parquet writers, fields with high base digit strings may not generate row level offset indexes, which leads to errors here.

We need to deal with this situation.

### Tests

Cannot re-produce this by Java Parquet.

### DIFF
--- a/paimon-format/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
+++ b/paimon-format/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
@@ -812,36 +812,40 @@ public class ParquetFileReader implements Closeable {
     private RowRanges calculateRowRanges(int blockIndex) {
         BlockMetaData block = blocks.get(blockIndex);
         RowRanges rowRanges = RowRanges.createSingle(block.getRowCount());
-        if (selection != null) {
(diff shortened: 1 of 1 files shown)


paimon-format/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -813,11 +814,17 @@ private RowRanges calculateRowRanges(int blockIndex) {
         RowRanges rowRanges = RowRanges.createSingle(block.getRowCount());
         if (selection != null) {
             ColumnIndexStore store = getColumnIndexStore(blockIndex);
-            List<OffsetIndex> offsets =
-                    paths.keySet().stream()
-                            .map(store::getOffsetIndex)
```
Zouxxyy, line 818: Seemed internalReadFilteredRowGroup also use this

JingsongLi: When there is no column index in it, the partition will invalidate all column indices (see getColumnIndexStore), so the safety here is that only after filtering will it enter this method.

Zouxxyy (approved):
+1

PR #6202, opened 2025-09-05
[core]support format table write

jerry-024:
### Purpose

This PR introduces **Format Table Write** functionality to Paimon. This feature provides:

1. **ACID Guarantees**: Implements two-phase commit mechanism using `CommittablePositionOutputStream` abstraction
2. **Multi-Storage Support**: Robust support for cloud storage (S3, OSS) via multi-part uploads and HDFS via atomic rename operations

**Key Components**:
- `CommittablePositionOutputStream` - Core abstraction for two-phase commit writes
- `MultiPartUploadStore` - Interface for cloud storage multi-part uploads  
- `FormatTableFileWrite` - Implementation for format table writing
- Storage-specific implementations for S3, OSS, and HDFS

### Tests

**Unit Tests**:
- `CommittablePositionOutputStreamTest` - Core abstraction functionality
- `MultiPartUploadCommittablePositionOutputStreamTest` - Multi-part upload logic
- `TempFileCommittablePositionOutputStreamTest` - HDFS rename commit mechanism
- `FormatTableFileWriteTest` - Format table write operations

**Integration Tests**:
- `CatalogTestBase#testFormatTableReadAndWrite`


paimon-common/src/main/java/org/apache/paimon/fs/local/LocalFileIO.java
```diff
@@ -92,6 +94,12 @@ public PositionOutputStream newOutputStream(Path path, boolean overwrite) throws
         return new LocalPositionOutputStream(toFile(path));
     }
 
+    @Override
+    public CommittablePositionOutputStream newCommittableOutputStream(Path path, boolean overwrite)
```
JingsongLi, line 98: Remove this? Just use default implementation.

paimon-common/src/main/java/org/apache/paimon/fs/CommittablePositionOutputStream.java
```diff
@@ -0,0 +1,57 @@
...
+
+package org.apache.paimon.fs;
+
+import java.io.IOException;
+
+/**
+ * CommittablePositionOutputStream provides a way to write to a file and get a committer that can
+ * commit.
+ */
+public abstract class CommittablePositionOutputStream extends PositionOutputStream {
```
JingsongLi, line 27: Or name this to `TwoPhaseOutputStream`?

paimon-common/src/main/java/org/apache/paimon/fs/RenameTempFileCommittablePositionOutputStream.java
```diff
@@ -0,0 +1,145 @@
...
+
+import java.io.IOException;
+import java.util.UUID;
+
+/**
+ * A {@link CommittablePositionOutputStream} implementation that writes to a temporary file and
+ * commits by renaming to the target path. This follows HDFS-style commit semantics.
+ */
+@Public
+public class RenameTempFileCommittablePositionOutputStream extends CommittablePositionOutputStream {
```
JingsongLi, line 31: `RenamingTwoPhaseOutputStream`

paimon-core/src/main/java/org/apache/paimon/table/format/FormatTableFileWrite.java
```diff
@@ -0,0 +1,132 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ExecutorService;
+import java.util.function.Function;
+
+import static org.apache.paimon.format.FileFormat.fileFormat;
+
+/** File write for format table. */
+public class FormatTableFileWrite extends MemoryFileStoreWrite<InternalRow> {
```
JingsongLi, line 48: Can we not use `MemoryFileStoreWrite`? Just create a new class for it.

paimon-core/src/main/java/org/apache/paimon/table/sink/BatchTableWrite.java
```diff
@@ -29,14 +29,14 @@
  * @since 0.4.0
  */
 @Public
-public interface BatchTableWrite extends TableWrite {
+public interface BatchTableWrite<T> extends TableWrite {
```
JingsongLi, line 32: Never modify this, this is public API.

JingsongLi (changes requested):

JingsongLi:
Can you extract `FileIO` codes to a separate PR?

paimon-core/src/main/java/org/apache/paimon/table/format/FormatTableFileWrite.java
JingsongLi: If you haven't used most of the related items, don't reuse them

PR #6207, opened 2025-09-05
[python] Fix failing to access DLF under python 3.6

discivigour:
### Purpose

- DLF access demo runs successfully.
- Add some tests
- Specify package versions in setup.py under python 3.6

### Tests
pypaimon.tests.py36.ao_read_write_test


JingsongLi (approved):
+1

PR #6178, opened 2025-08-31
[core] Support multi ordering TopN with allow duplicate in bitmap

ArnavBalyan:
- Add support for multi-column topN with allowing duplicates to topK filtering in roaring bitmap.
 - Idea is to allow duplicates in the primary column to allow the engine to perform the global sort after collection.
 - Since all rows belonging to the top n category are selected, the engine has enough context to break the tie and perform secondary sort on the other columns (which is typically much smaller than the whole result set).
 - Added UTs

PR #6209, opened 2025-09-05
[core] The TopN predicate for range-bitmap supports multiple sort keys

Tan-JiaLiang:
### Purpose

Wrap up #6178

Suppose the `id` has created with `range-bitmap`.
position | id | price
--|--|--|
0|10 | 100
1|20 | 100
2|20 | 200
3|20 | 300
4|30 | 100

`SELECT * FROM id DESC LIMIT 2` will return
position | id | price
--|--|--|
3|20 | 300
4|30 | 100

`SELECT * FROM id DESC, price LIMIT 2` will return this. And the compute engine will rerank the result.
position | id | price
--|--|--|
1|20 | 100
2|20 | 200
3|20 | 300
4|30 | 100


JingsongLi (approved):
+1

PR #6210, opened 2025-09-06
[flink] Fix file not found when using changelog compact with multiple extra paths

tsreaper:
### Purpose

Changelog compact tasks currently does not set the correct external path for fake file metas. This will cause `FileNotFoundException` when reading changelogs. This PR fixes the issue.

### Tests

* `ChangelogCompactTaskTest`

### API and Format

No format changes.

### Documentation

No new feature.


JingsongLi (approved):
+1

PR #6213, opened 2025-09-08
[format] Format writer check null value for non-null field and throw useful message

yuzelin:
### Purpose

A brief case is:
```
CREATE TABLE test (a INT PRIMARY KEY NOT ENFORCED, b INT NOT NULL) WITH ('merge-engine' = 'aggregation', 'changelog-producer' = 'lookup');

INSERT INTO test VALUES (1, 1);
-- Then the upstream input -D(1, 1)
```

The problem is that default agg for b is last-non-null-value which produces null when retracting.

### DIFF
--- a/paimon-format/src/main/java/org/apache/paimon/format/avro/FieldWriterFactory.java
+++ b/paimon-format/src/main/java/org/apache/paimon/format/avro/FieldWriterFactory.java
@@ -18,6 +18,7 @@
 
 package org.apache.paimon.format.avro;
 
+import org.apache.paimon.CoreOptions;
 import org.apache.paimon.data.DataGetters;
 import org.apache.paimon.data.Decimal;
 import org.apache.paimon.data.GenericRow;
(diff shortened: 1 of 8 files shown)


paimon-format/src/main/java/org/apache/paimon/format/avro/FieldWriterFactory.java
```diff
@@ -259,7 +268,18 @@ public void write(DataGetters container, int index, Encoder encoder) throws IOEx
 
         public void writeRow(InternalRow row, Encoder encoder) throws IOException {
             for (int i = 0; i < fieldWriters.length; i += 1) {
-                fieldWriters[i].write(row, i, encoder);
+                try {
```
JingsongLi: try wrap `for` loop. better performance.

paimon-format/src/main/java/org/apache/paimon/format/orc/writer/RowDataVectorizer.java
```diff
@@ -50,10 +58,18 @@ public void vectorize(InternalRow row, VectorizedRowBatch batch) {
         for (int i = 0; i < row.getFieldCount(); ++i) {
             ColumnVector fieldColumn = batch.cols[i];
             if (row.isNullAt(i)) {
+                checkArgument(
```
JingsongLi: use if instead of checkArgument.

JingsongLi:
+1

PR #6214, opened 2025-09-08
[core] Add lazied fields in RowType to improve performance

JingsongLi:


leaves12138:
Should we make RowType thread-safe?

leaves12138 (approved):
+1

PR #6216, opened 2025-09-08
fix inconsistent attribute names

freesinger:
Inconsistent attribute names lead to ambiguity and it is impossible to determine which one to use

### Purpose

Linked issue: close #xxx


JingsongLi:
+1

PR #6218, opened 2025-09-08
[core] Refactor AggregateMergeFunction and PartialUpdateMergeFunction for better maintenance and add test for retracting

yuzelin:
### Purpose

Use with #6213 for finding which agg cuases NPE.


JingsongLi (approved):
+1

PR #6219, opened 2025-09-08
[core] Introduce external buffer in lookup merging

JingsongLi:
### Purpose

- Introduce `lookup.merge-buffer-size` and `lookup.merge-records-threshold`.
- When this threshold is exceeded, a spillable external buffer will be used to avoid OOM exceptions caused by too much records.


leaves12138 (approved):
+1

PR #6222, opened 2025-09-09
[flink] Supports read-optimized via hints

### LINKED ISSUE #6221: [Feature] Supports read-optimized via hints
### Search before asking

### Motivation

When need to improve query performance, one of the methods is to switch to read the `RO` table. However, when integrating other systems internally, it is found that the downstream system does not support switching tables, so hope that the query behavior can be changed through hint

### Solution

Add table option

### Anything else?

None

zhuyufeng0809:
### Purpose

Linked issue: close #6221 

read-optimized via hints

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/AbstractFlinkTableFactory.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/AbstractFlinkTableFactory.java
@@ -36,6 +36,8 @@ import org.apache.paimon.schema.SchemaManager;
 import org.apache.paimon.table.FileStoreTable;
 import org.apache.paimon.table.FileStoreTableFactory;
 import org.apache.paimon.table.Table;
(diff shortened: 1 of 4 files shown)


zhuyufeng0809:
Hi @Aitozi , could you help me to review this PR ?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/AbstractFlinkTableFactory.java
```diff
@@ -110,6 +113,11 @@ && parseBoolean(options.get(SCAN_BOUNDED.key()))) {
         }
         if (origin instanceof SystemCatalogTable) {
             return new SystemTableSource(table, unbounded, context.getObjectIdentifier());
+        } else if (!unbounded && parseBoolean(options.get(READ_OPTIMIZED.key()))) {
```
JingsongLi, line 116: It seems `READ_OPTIMIZED` is not a generic option, only work for Flink SQL.

PR #6225, opened 2025-09-09
[python] Fix string size being an integer multiple of bytes

discivigour:
### Tests

- RESTTableReadWritePy36Test.test_to_bytes_with_long_string()
- ReaderBasicTest.test_to_bytes_with_long_string()


discivigour:
added

discivigour:
> Please add a unit test.

added

PR #6231, opened 2025-09-10
[core] Support push down branchesTable by branchName

huyuanfeng2018:
### Purpose

Linked issue: close #xxx

In Paimon's system table $branches, when a user needs to query specific branch information, the previous implementation did not support pushing down the filtering criteria of branch_name.

This means that even if we only want to query one or a few specific branches, Paimon will read all branch information and then filter it at the computation engine layer

This change aims to optimize the query performance of the $branches table by implementing the pushdown of filtering conditions for the branch_name field, allowing queries to be filtered directly at the storage layer based on the branch name, thereby significantly improving query efficiency.

### Tests

org.apache.paimon.flink.BranchSqlITCase##testBranchesTableFilter with adding more case for the method.


JingsongLi:
> @JingsongLi Can take some time to help me review this PR? Thank you so much~

Thanks @huyuanfeng2018 for the contribution. Do your scenes have many branches? I rarely see situations where there are many branches, so I haven't optimized it here either.

PR #6232, opened 2025-09-10
[core] RESTCatalog: add DLF OSS endpoint support and improve configuration merge

plusplusjiajia:
### Purpose
When using Paimon with DLF (Data Lake Formation) server, the server passes default OSS endpoint values (e.g., fs.oss.endpoint=oss-cn-hangzhou-internal.aliyuncs.com) through tokens. However, these default internal endpoints are not accessible in all client environments, causing connectivity issues. The client-provided options should take precedence over server defaults, but the previous merge logic didn't handle this correctly. This PR  introduced DLF_OSS_ENDPOINT configuration option in RESTCatalogOptions to allow clients to specify their preferred OSS endpoint.

### Tests
org.apache.paimon.utils.RESTUtilTest


paimon-api/src/main/java/org/apache/paimon/rest/RESTUtil.java
```diff
@@ -72,19 +74,30 @@ public static Map<String, String> merge(
...
+        for (Map.Entry<String, String> entry : targets.entrySet()) {
             if (entry.getValue() != null) {
-                builder.put(entry.getKey(), entry.getValue());
+                result.put(entry.getKey(), entry.getValue());
             }
         }
 
-        return builder.build();
+        // Handle special case: dlf.oss-endpoint should override fs.oss.endpoint
+        String dlfOssEndpoint = result.get(DLF_OSS_ENDPOINT.key());
```
JingsongLi: Can you just modify `RESTTokenFileIO`? `RESTUtil.merge` should be a very simple method, just `updates` merge into `targets`.

plusplusjiajia: > Can you just modify `RESTTokenFileIO`? `RESTUtil.merge` should be a very simple method, just `updates` merge into `targets`.
Sure, updated.

PR #6246, opened 2025-09-12
[core]format table: use format table read for flink

jerry-024:
### Purpose

Use Paimon format table read to read in Flink

### Tests
-  FormatTableReadITCaseBase


JingsongLi (approved):
+1

PR #6247, opened 2025-09-12
[core] Ensure system tables use the correct identifier for loadTableToken in RESTTokenFileIO

plusplusjiajia:
### Purpose
Fix REST token handling for system tables


JingsongLi (approved):
+1

PR #6256, opened 2025-09-15
[avro] Optimize map<string, string> in avro

tsreaper:
### Purpose

`map<string, string>` is a useful data type in many tables. However, current `map<string, string>` in Paimon avro is slow, because Paimon reader decodes it into `HashMap`s.

This PR optimizes `map<string, string>` by storing the original bytes from avro, so in scenarios like compacting, where we don't care about the exact values of a row, we don't need to pay the decoding and encoding costs. When the exact value is needed, these bytes will be decoded into `BinaryArray`s.

### Tests

* `FormatReadWriteTest`

### API and Format

No format changes.

### Documentation

No new feature.


paimon-common/src/main/java/org/apache/paimon/data/BinaryWriter.java
```diff
@@ -62,7 +62,11 @@ public interface BinaryWriter {
 
     void writeString(int pos, BinaryString value);
 
-    void writeBinary(int pos, byte[] bytes);
+    default void writeBinary(int pos, byte[] bytes) {
```
JingsongLi: Do not introduce to performance critical part.

paimon-format/src/main/java/org/apache/paimon/format/avro/AvroBytesStringMap.java
```diff
@@ -0,0 +1,164 @@
...
+            BinaryArray.calculateFixLengthPartSize(DataTypes.STRING());
+
+    // map size
+    private int size;
+    // stores original bytes from avro files
+    private byte[] bytes;
+    // total bytes length
+    private int lengthInBytes;
+    // offset of each string
+    private final List<Integer> off;
```
JingsongLi: use `IntArrayList`?

paimon-format/src/main/java/org/apache/paimon/format/avro/AvroBytesStringMap.java
```diff
@@ -0,0 +1,164 @@
...
+    // length of each string
+    private final List<Integer> len;
```
JingsongLi: Use `IntArrayList`?

paimon-format/src/main/java/org/apache/paimon/format/avro/AvroBytesStringMap.java
```diff
@@ -0,0 +1,164 @@
...
+
+import org.apache.avro.io.BinaryData;
+import org.apache.avro.io.BinaryDecoder;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+/** An {@link InternalMap} stored in avro format bytes. */
+public class AvroBytesStringMap implements InternalMap {
```
JingsongLi, line 34: Please add tests for `AvroBytesStringMap`.

JingsongLi (commented):
Left comments.

paimon-format/src/main/java/org/apache/paimon/format/avro/AvroBytesStringMap.java
```diff
@@ -0,0 +1,163 @@
...
+    private int lengthInBytes;
+    // offset of each string
+    private final IntArrayList off;
+    // length of each string
+    private final IntArrayList len;
+
+    private BinaryArray keyArray;
+    private BinaryArray valueArray;
+
+    public AvroBytesStringMap(BinaryDecoder decoder, boolean valueNullable) throws IOException {
```
JingsongLi: Can you create a static method for this? And just pass all fields to this constructor.

In this way, `AvroBytesStringMap` can be an immutable class.

PR #6257, opened 2025-09-16
[core] Fix manifest file not cleaned if compact_manifest failed

wkang13579:
### Purpose

- In paimon 1.0, if compact_manifest failed, legacy manifest files will be deleted, pr https://github.com/apache/paimon/pull/5776 remove this deletion.
- Add default commit user-prefix for spark procedure.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
@@ -1283,7 +1283,13 @@ public class FileStoreCommitImpl implements FileStoreCommit {
                         latestSnapshot.properties(),
                         latestSnapshot.nextRowId());
 
-        return commitSnapshotImpl(newSnapshot, emptyList());
+        boolean success = commitSnapshotImpl(newSnapshot, emptyList());
+        if (!success) {
+            LOG.info("Commit failed for compact manifest, will be retried again.");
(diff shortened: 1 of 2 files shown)


JingsongLi (commented):
Important fix for manifest file leak when compact_manifest fails.

**Problem**: In Paimon 1.0, when manifest compaction fails mid-way, the newly created manifest files are not cleaned up, leading to orphan files accumulating over time.

**Fix**:
1. After `commitSnapshotImpl` returns false (CAS failure), clean up the delta manifest list and unused tmp manifests.
2. Added a default commit user prefix for the Spark CompactManifestProcedure to avoid empty commit user.

**Review**:
1. The cleanup logic (`manifestList.delete` + `cleanUpNoReuseTmpManifests`) is correct - these files were created for the failed commit attempt and are now orphans.
2. The LOG.info before cleanup is helpful for debugging.
3. Since this is in a retry path (`compactManifestOnce` returns false, caller retries), cleaning up before retry prevents accumulation.

**Question**: Is there a race condition where another committer might reference these manifests? Since CAS failed, it means another snapshot was committed concurrently - but these specific manifests were freshly created and not referenced by any snapshot yet, so deletion should be safe.

LGTM - fixes a real resource leak issue.

JingsongLi:
Reverted in f6ccc0902da55ebf1e763cd60ff25f3e05372563

This is very dangerous. We are not sure if we have committed it

PR #6261, opened 2025-09-16
[spark] Compact field files for data evolution table.

leaves12138:
### Purpose

Compact field files for data evolution table (Which is not compact table just compact single fields in small files)


JingsongLi (changes requested):
I think this logical should be included in `compact` procedure.

If we feel that the cost is too high, we can consider introducing a minor compact strategy for `compact` procedure in the future.

PR #6267, opened 2025-09-16
[test] Fix method visibility in CatalogTestBase

plusplusjiajia:
### Purpose
Fix CatalogTestBase#testFormatTableRead visibility


JingsongLi:
+1

PR #6268, opened 2025-09-16
[core] Introduce blob type and blob data

JingsongLi:
### Purpose

See https://cwiki.apache.org/confluence/display/PAIMON/PIP-35%3A+Introduce+Blob+to+store+multimodal+data


leaves12138 (approved):
+1

PR #6271, opened 2025-09-17
[hive] Fix sync-all-properties cannot configurable

wkang13579:
### Purpose

We set `sync-all-properties=false` and find all properties also sync to hms, we should check it by using paimon table options.


JingsongLi (commented):
Good bug fix. The issue is clear: `sync-all-properties` was being read only from catalog options, ignoring the table-level override.

**Fix**: Changed `syncAllProperties()` to `syncAllProperties(Map<String, String> tableOptions)` which merges table options on top of catalog options before checking the flag. This allows per-table configuration to override the catalog default.

**Review**:
1. The merge order is correct: table options override catalog options (`merged.putAll(tableOptions)`).
2. Applied consistently in both `createHiveTable` and `updateHmsTablePars`.
3. Small and focused change (+9/-7).

**Minor concern**: Creating a new `HashMap` and `Options` object on every call to `syncAllProperties` is a minor allocation cost. For frequently called paths this could be optimized, but for HMS operations (network-bound) this is negligible.

LGTM.

JingsongLi:
Please rebase master

PR #6277, opened 2025-09-17
[core] Add name for catalog

Zouxxyy:
### Purpose

1. Add `name` in interface `Catalog`

2. Make fileStoreTable's `fullName` return `catalog.db.table` if catalogName if present. 
  - This is beneficial for systems like Kyuubi to obtain the catalogName for authorization purposes.
  - Enables more comprehensive representation in the UI.

image

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/catalog/CatalogContext.java
+++ b/paimon-common/src/main/java/org/apache/paimon/catalog/CatalogContext.java
@@ -48,18 +48,21 @@ public class CatalogContext implements Serializable {
     private final SerializableConfiguration hadoopConf;
     @Nullable private final FileIOLoader preferIOLoader;
     @Nullable private final FileIOLoader fallbackIOLoader;
+    @Nullable final String catalogName;
 
     private CatalogContext(
             Options options,
             @Nullable Configuration hadoopConf,
             @Nullable FileIOLoader preferIOLoader,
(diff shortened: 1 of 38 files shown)


JingsongLi (commented):
Adding a `name` property to the Catalog interface. Review:

1. **Purpose**: Enabling `catalog.db.table` as the full name is useful for authorization (Kyuubi) and UI representation. Makes sense.

2. **Scope**: +164/-55 across 38 files is a lot of churn for what is conceptually a simple addition. Most changes are passing the catalog name through constructors and test updates.

3. **API change**: Adding `name()` to the `Catalog` interface is a public API change. Consider:
   - Default implementation returning empty/null for backward compatibility with custom catalog implementations
   - Documentation of the expected format and lifecycle

4. **Implementation**: The name flows from `CatalogContext` → `CatalogFactory` → `AbstractCatalog` → `CatalogEnvironment` → `Table.fullName()`. The chain is long but logical.

5. **Cross-engine impact**: Changes span Flink, Spark, and Hive catalogs. All need to provide the name consistently.

6. **Testing**: Test updates look like they're adapting to the new constructor signatures. Dedicated tests for the fullName format would be valuable.

The feature is reasonable but the PR touches many files. Consider whether there's a way to make the name optional at the interface level to reduce blast radius for external catalog implementations.

PR #6285, opened 2025-09-19
[core] Separate  indexIncrement into dataIncrement and compactIncrement

Zouxxyy:
### Purpose

Separate indexIncrement into dataIncrement and compactIncrement, to distinguish between compact and non-compact commits for #6025

```java
public class DataIncrement {
    ...
    private final List<IndexFileMeta> newIndexFiles;
    private final List<IndexFileMeta> deletedIndexFiles;

}

public class CompactIncrement {
    ...
    private final List<IndexFileMeta> newIndexFiles;
    private final List<IndexFileMeta> deletedIndexFiles;
}
```


JingsongLi:
+1

PR #6291, opened 2025-09-19
[cdc] Fix cannot be cast to float due to precision loss

### LINKED ISSUE #6289: [Bug] Paimon cdc throws exception "175.26562 cannot be cast to float due to precision loss"
### Search before asking

### Paimon version

1.2.0

### Compute Engine

flink 1.20

### Minimal reproduce step

1. insert float 175.26562 value to postgresql table
2. start paimon flink cdc table sync

I have seen that someone has fixed the float type conversion problem (#5263), but this method is not universal and still results in errors for some special values. For example: float 175.26562 will be converted to 175.26563 according to the following code, so it will report an error. The expected value is the same.
```java
double d = Double.parseDouble(s);
if (d == ((float) d)) {
    return (float) d;
} else {
    // Compatible canal-cdc
    Float f = Float.valueOf(s);
    String floatStr = f.toString();
    if (s.contains(".") && !s.contains("E")) {
        int decimal = s.length() - s.indexOf(".") - 1;
        floatStr = String.format("%." + decimal + "f", f);
    }
    if (!floatStr.equals(s)) {
    } else {
        return f;
    }
}
```

change to:
```java
if (!String.valueOf(ff).equals(String.valueOf(d))) {
} else {
    return f;
}
```

Image

### Anything else?

_No response_

lizc9:
### Purpose

Linked issue: close #6289

### Tests

- org.apache.paimon.utils.TypeUtilsTest#testFloatCastFromString

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/utils/TypeUtils.java
+++ b/paimon-common/src/main/java/org/apache/paimon/utils/TypeUtils.java
@@ -144,12 +144,7 @@ public class TypeUtils {
                 } else {
                     // Compatible canal-cdc
                     Float f = Float.valueOf(s);
-                    String floatStr = f.toString();
-                    if (s.contains(".") && !s.contains("E")) {
-                        int decimal = s.length() - s.indexOf(".") - 1;
-                        floatStr = String.format("%." + decimal + "f", f);
-                    }
-                    if (!floatStr.equals(s)) {
+                    if (!f.toString().equals(Double.toString(d))) {
                         throw new NumberFormatException(
                                 s + " cannot be cast to float due to precision loss");
                     } else {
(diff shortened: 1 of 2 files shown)


paimon-common/src/main/java/org/apache/paimon/utils/TypeUtils.java
```diff
@@ -144,12 +144,7 @@ public static Object castFromStringInternal(String s, DataType type, boolean isC
                 } else {
                     // Compatible canal-cdc
                     Float f = Float.valueOf(s);
-                    String floatStr = f.toString();
-                    if (s.contains(".") && !s.contains("E")) {
```
JingsongLi, line 148: Can you add a test for `E`?

lizc9: > Can you add a test for `E`?

Of course, an example has been added to the unit test below: 6.1042607E-4

JingsongLi:
+1

PR #6292, opened 2025-09-19
[spark] Resolve function after all args have been resolved

Zouxxyy:
### Purpose

Fix this

```
CREATE VIEW v AS SELECT udf_add2(a, b) AS c1 FROM t;
SELECT * FROM v;
```

```
No handler for UDF/UDAF/UDTF 'org.apache.hadoop.hive.contrib.udf.example.UDFExampleAdd2': org.apache.spark.sql.catalyst.analysis.UnresolvedException: Invalid call to dataType on unresolved object
org.apache.spark.sql.AnalysisException: No handler for UDF/UDAF/UDTF 'org.apache.hadoop.hive.contrib.udf.example.UDFExampleAdd2': org.apache.spark.sql.catalyst.analysis.UnresolvedException: Invalid call to dataType on unresolved object
	at org.apache.spark.sql.catalyst.analysis.UnresolvedAttribute.dataType(unresolved.scala:264)
	at org.apache.spark.sql.hive.HiveSimpleUDFEvaluator.$anonfun$method$1(hiveUDFEvaluators.scala:74)
	at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:286)
```


JingsongLi:
+1

Issue #6282, opened 2025-09-18
[Bug] NullPointException when readHadoopXml


linqichen:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Paimon version
paimon 1.1
spark 3.2

### Compute Engine

 Spark 、 Hive 

### Minimal reproduce step

It always happened when my hdfs-site.xml like this:
```
<property>
  <name>xxxx</name>
</property>
```

### What doesn't meet your expectations?

 NullPointException  when readHadoopXml。 

error info:
java.lang.NullPointerException
at org.apache.paimon.utils .HadoopUtils.readHadoopXml(HadoopUtils.java:228)
at org.apache.paimon.utils .HadoopUtils.addHadoopconfIfFound(HadoopUtils .java:175)
....

becasuse  hdfs-site.xml like this:
```
<property>
  <name>xxxx</name>
</property>
```

### Anything else?

i have already got down it by modifing  org.apache.paimon.utils.readHadoopXml()  .
it is very easy.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

linqichen:
please assign it to me . 
My solution is to add some non-empty checks. I have verified it in my company's internal version.

PR #6294, opened 2025-09-20
[fix Bug] NullPointException when readHadoopXml apache/paimon#6282

linqichen:
### Purpose

fix bug: NullPointException when readHadoopXml 
https://github.com/apache/paimon/issues/6282

### Tests
org.apache.paimon.utils.HadoopUtilsITCase#testReadHadoopXml  passed.

image

### API and Format
not involve

### Documentation
not involve

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/utils/HadoopUtils.java
+++ b/paimon-common/src/main/java/org/apache/paimon/utils/HadoopUtils.java
@@ -221,13 +221,19 @@ public class HadoopUtils {
         }
         for (int i = 0; i < propertyNodes.getLength(); i++) {
             Node propertyNode = propertyNodes.item(i);
-            if (propertyNode.getNodeType() == 1) {
+            if (1 == propertyNode.getNodeType()) {
                 Element propertyElement = (Element) propertyNode;
-                String key = propertyElement.getElementsByTagName("name").item(0).getTextContent();
-                String value =
-                        propertyElement.getElementsByTagName("value").item(0).getTextContent();
-                if (!StringUtils.isNullOrWhitespaceOnly(value)) {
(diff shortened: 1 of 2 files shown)


paimon-common/src/main/java/org/apache/paimon/utils/HadoopUtils.java
```diff
@@ -221,13 +221,19 @@ public static void readHadoopXml(String xml, Configuration conf) {
...
             Node propertyNode = propertyNodes.item(i);
-            if (propertyNode.getNodeType() == 1) {
+            if (1 == propertyNode.getNodeType()) {
                 Element propertyElement = (Element) propertyNode;
-                String key = propertyElement.getElementsByTagName("name").item(0).getTextContent();
-                String value =
-                        propertyElement.getElementsByTagName("value").item(0).getTextContent();
-                if (!StringUtils.isNullOrWhitespaceOnly(value)) {
-                    conf.set(key, value);
+                if (null != propertyElement.getElementsByTagName("name")
```
JingsongLi, line 226: Can you extract local field for `propertyElement.getElementsByTagName("name")`?

linqichen: sorry， i donot understand your question.   what do "extract local field"  mean ?

JingsongLi:
```
String name = propertyElement.getElementsByTagName("name");
```

And use name.

linqichen: OK, i will close this pr  and create a new one.

PR #6295, opened 2025-09-20
[rest] Add tableType parameter support for listing tables

plusplusjiajia:
### Purpose
This PR enhances the REST Catalog API by introducing a new optional tableType parameter to the list tables endpoints. This allows users to filter tables based on their type, improving query efficiency and providing more granular control over table discovery.

### DIFF
--- a/docs/static/rest-catalog-open-api.yaml
+++ b/docs/static/rest-catalog-open-api.yaml
@@ -350,6 +350,12 @@ paths:
           in: query
           schema:
             type: string
+        - name: tableType
+          description: Filter tables by table type. All table types will be returned if not set or empty.
+          in: query
+          schema:
+            type: string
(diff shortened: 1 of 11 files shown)


docs/static/rest-catalog-open-api.yaml
```diff
@@ -350,6 +350,12 @@ paths:
           in: query
           schema:
             type: string
+        - name: tableType
+          description: Filter tables by table type. All table types will be returned if not set or empty.
+          in: query
+          schema:
+            type: string
+            enum: ["table", "format-table", "materialized-table", "object-table", "lance-table", "iceberg-table"]
```
JingsongLi: remove this enum, we don't need maintain enum here.

plusplusjiajia: > remove this enum, we don't need maintain enum here.

removed.

JingsongLi:
+1

PR #6300, opened 2025-09-22
[core]support  whether only partition value in path for format table

jerry-024:
### Purpose
This PR adds support for a format table that only contains the partition value in the file path.

user need add option: `'format-table.partition-path-only-value'='true'` to enable this feature 

### Tests

- FormatTableScanTest#testGeneratePartitions
- CatalogTestBase#testFormatTableOnlyPartitionValueRead


paimon-api/src/main/java/org/apache/paimon/options/FormatTableOptions.java
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.paimon.options;
+
+/** Options for format table. */
+public class FormatTableOptions {
+
+    public static final ConfigOption<Boolean> READ_ENABLE_PARTITION_ONLY_VALUE_IN_PATH =
+            ConfigOptions.key("format-table.path.only-partition-value")
```
JingsongLi: Just put this into CoreOptions.

JingsongLi: format-table.partition-path-only-value

paimon-core/src/main/java/org/apache/paimon/utils/PartitionPathUtils.java
```diff
@@ -75,6 +77,16 @@ private static boolean needsEscaping(char c) {
      * @return An escaped, valid partition name.
      */
     public static String generatePartitionPath(LinkedHashMap<String, String> partitionSpec) {
+        return generatePartitionPathUtil(partitionSpec, false);
+    }
+
+    public static String generatePartitionPathOnlyValue(
```
JingsongLi: Remove this method, just invoke `generatePartitionPathUtil(partitionSpec, true)`.

paimon-core/src/test/java/org/apache/paimon/table/format/FormatTableScanTest.java
```diff
@@ -93,7 +93,7 @@ void testComputeScanPathAndLevelNoPartitionKeys() {
 
         Pair<Path, Integer> result =
                 FormatTableScan.computeScanPathAndLevel(
-                        tableLocation, partitionKeys, partitionFilter, partitionType);
+                        tableLocation, partitionKeys, partitionFilter, partitionType, false);
 
         assertThat(result.getLeft()).isEqualTo(tableLocation);
```
JingsongLi: For this test, we should add only value to all tests. You can use `@ExtendWith(ParameterizedTestExtension.class)`.

PR #6301, opened 2025-09-22
[lance] Fix read lance with range position

Zouxxyy:
### Purpose

Fix read lance with range position


pom.xml
```diff
@@ -428,6 +428,7 @@ under the License.
                 <scala.version>${scala213.version}</scala.version>
                 <paimon-spark-common.spark.version>4.0.1</paimon-spark-common.spark.version>
                 <paimon-sparkx-common>paimon-spark4-common_2.13</paimon-sparkx-common>
+                <arrow.version>18.1.0</arrow.version>
```
Zouxxyy, line 431: For spark4 we need use high version arrow

PR #6303, opened 2025-09-22
[core] Add dv conflict detection during commit

Zouxxyy:
### Transaction Isolation Level

1. Snapshot Isolation: For writes to the same file (such as COW or DV writes), the first write to complete wins.
2. Serializability: In addition to level 1, it also checks whether new files generated from reading the snapshot ID to writing the snapshot ID are also applied to the processing logic (for example, newly added files also need to be updated).

Serializability relies on reading the snapshot ID, which is currently not supported by Paimon. Paimon currently achieves snapshot isolation by comparing two consecutive snapshots to see if they conflict.

### This PR 

Enhances snapshot isolation for DV tables. Compared with non-DV table conflict detection, DV table needs to compare <file, db> pair.

non dv table
snapshot t-1 -> base files -> ADD f
snapshot t -> delete files + add files -> ADD f or DELETE f

dv table
snapshot t-1 -> base files + dv files -> ADD <f, dv>
snapshot t -> delete files + new files + delete dv files + new dv files -> ADD <f, dv> or DELETE <f, dv>

Other fix:

Use scan snapshot to persistDeletionVectors in spark row level command with dv table. This can ensure that the snapshot used for DV generation and reading is the same as much as possible. (Although I feel it is not so accurate. Perhaps this can only be achieved if we implement read snapshot id at the table interface level)

### Todo

Since PK table's compact dv index and Non-PK HASH table's dv index entry only contains ADD type, we can't perform conflict detection on them.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/AbstractFileStore.java
+++ b/paimon-core/src/main/java/org/apache/paimon/AbstractFileStore.java
@@ -298,7 +298,10 @@ abstract class AbstractFileStore<T> implements FileStore<T> {
                 options.commitMinRetryWait(),
                 options.commitMaxRetryWait(),
                 options.commitStrictModeLastSafeSnapshot().orElse(null),
-                options.rowTrackingEnabled());
+                options.rowTrackingEnabled(),
+                !schema.primaryKeys().isEmpty(),
+                options.deletionVectorsEnabled(),
+                newIndexFileHandler());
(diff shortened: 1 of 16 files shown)


JingsongLi:
Thanks @Zouxxyy for the contribution! We will soon be safe in the scenario of concurrent DV writing!

PR #6308, opened 2025-09-23
[spark] Add latestSnapshotId to the description in paimon scan

kerwin-zk:
### Purpose

[spark] Add latestSnapshotId to the description in paimon scan

image

### Tests

CI

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonBaseScan.scala
+++ b/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonBaseScan.scala
@@ -132,7 +132,12 @@ abstract class PaimonBaseScan(
     } else {
       ""
     }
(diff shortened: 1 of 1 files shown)


paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonBaseScan.scala
```diff
@@ -132,7 +132,12 @@ abstract class PaimonBaseScan(
     } else {
       ""
     }
-    s"PaimonScan: [${table.name}]" + pushedFiltersStr + pushedTopNFilterStr +
+    val latestSnapshotIdStr = if (table.latestSnapshot().isPresent) {
```
Zouxxyy, line 135: you should use currentSnapshot

kerwin-zk: @Zouxxyy This PR is to obtain the latest snapshot ID, while the current snapshot ID will be obtained in another PR.

PR #6320, opened 2025-09-24
[core] Introduce creation methods to Blob

JingsongLi:
### Purpose

Introduce util methods:
```java
    static Blob fromData(byte[] data) {
        return new BlobData(data);
    }

    static Blob fromLocal(String file) {
        return fromFile(LocalFileIO.create(), new BlobDescriptor(file, 0, -1));
    }

    static Blob fromHttp(String uri) {
        return fromDescriptor(UriReader.fromHttp(), new BlobDescriptor(uri, 0, -1));
    }

    static Blob fromFile(FileIO fileIO, BlobDescriptor descriptor) {
        return fromDescriptor(UriReader.fromFile(fileIO), descriptor);
    }

    static Blob fromDescriptor(UriReader reader, BlobDescriptor descriptor) {
        return new BlobRef(reader, descriptor);
    }
```


lucasfang:
+1

PR #6324, opened 2025-09-24
[Core] Support read increment changelog and delta between two tag.

wwj6591812:
### Purpose

In our company, we have a business requirement to compute the primary keys of new (+I) and deleted (-D) records between today's and yesterday's partitions of an ODPS (Alibaba Cloud MaxCompute) table using Paimon.

We have conducted a POC that a daily Flink batch job reads a partition from the ODPS table and writes it to a Paimon primary key table. This job is configured with the following parameters:
`'full-compaction.delta-commits'='1', 'changelog-producer'='full-compaction', 'tag.automatic-creation' = 'batch', 'tag.batch.customized-name'='ds=20250922'`
This gives us two snapshots and one tag daily.

We found that we can read the changelog by providing snapshot IDs, like in the SQL below, but we can't do the same using tags.
`SELECT  rowkind, item_id, sku_id, ds
FROM    `alake`.`omega_alake`.`kk_invalid_sku_prediction_v4$audit_log` 
/*+ OPTIONS('scan.parallelism'='128', 'incremental-between-scan-mode'='changelog', 'incremental-between'='2,4')*/
where rowkind = '+U' OR rowkind = '-U'
limit 100;`

So, this PR adds support for reading the incremental changelog and delta between two specified tags.

Linked issue: close #xxx

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/source/AbstractDataTableScan.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/source/AbstractDataTableScan.java
@@ -69,6 +69,8 @@ import java.util.Optional;
 import java.util.TimeZone;
 
 import static org.apache.paimon.CoreOptions.FULL_COMPACTION_DELTA_COMMITS;
+import static org.apache.paimon.CoreOptions.IncrementalBetweenScanMode.CHANGELOG;
+import static org.apache.paimon.CoreOptions.IncrementalBetweenScanMode.DELTA;
 import static org.apache.paimon.CoreOptions.IncrementalBetweenScanMode.DIFF;
 import static org.apache.paimon.utils.Preconditions.checkArgument;
 import static org.apache.paimon.utils.Preconditions.checkNotNull;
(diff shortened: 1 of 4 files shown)


paimon-core/src/main/java/org/apache/paimon/table/source/AbstractDataTableScan.java
```diff
@@ -338,8 +340,18 @@ private StartingScanner createIncrementalStartingScanner(SnapshotManager snapsho
             Optional<Tag> endTag = tagManager.get(incrementalBetween.getRight());
 
             if (startTag.isPresent() && endTag.isPresent()) {
-                return IncrementalDiffStartingScanner.betweenTags(
-                        startTag.get(), endTag.get(), snapshotManager, incrementalBetween);
+                CoreOptions.IncrementalBetweenScanMode scanMode =
+                        options.incrementalBetweenScanMode();
+                if (scanMode == CHANGELOG || scanMode == DELTA) {
+                    return IncrementalDeltaStartingScanner.betweenSnapshotIds(
```
JingsongLi, line 346: What if the snapshot corresponding to the tag does not exist?

wwj6591812: The exception will be throw in IncrementalDeltaStartingScanner#betweenSnapshotIds, In this case, we don't support read delta / changelog between tags.

PR #6329, opened 2025-09-24
[spark] Disable v2 write in v1 append location unit tests

kerwin-zk:
### Purpose

Disable v2 write in v1 append location unit tests

The operation ```batch.write.format("paimon").mode("append").save(location)``` throws an exception when v2 write is enabled. Therefore, it is necessary to explicitly disable v2 write in the relevant unit tests.

### Tests

CI

### DIFF
--- a/paimon-spark/paimon-spark-3.4/src/test/scala/org/apache/paimon/spark/sql/CreateAndDeleteTagProcedureTest.scala
+++ b/paimon-spark/paimon-spark-3.4/src/test/scala/org/apache/paimon/spark/sql/CreateAndDeleteTagProcedureTest.scala
@@ -29,107 +29,109 @@ class CreateAndDeleteTagProcedureTest extends PaimonSparkTestBase with StreamTes
   import testImplicits._
 
   test("Paimon Procedure: create and delete tag") {
-    failAfter(streamingTimeout) {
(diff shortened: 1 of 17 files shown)


paimon-spark/paimon-spark-3.4/src/test/scala/org/apache/paimon/spark/sql/CreateAndDeleteTagProcedureTest.scala
```diff
@@ -29,107 +29,109 @@ class CreateAndDeleteTagProcedureTest extends PaimonSparkTestBase with StreamTes
...
-          val inputData = MemoryStream[(Int, String)]
-          val stream = inputData
-            .toDS()
-            .toDF("a", "b")
-            .writeStream
-            .option("checkpointLocation", checkpointDir.getCanonicalPath)
-            .foreachBatch {
-              (batch: Dataset[Row], _: Long) =>
-                batch.write.format("paimon").mode("append").save(location)
+    withSparkSQLConf(("spark.paimon.write.use-v2-write", "false")) {
```
Zouxxyy, line 32: When V2 write is enabled by default, it should bypass V2 write instead of throwing an exception in scenarios that do not support it.

PR #6333, opened 2025-09-25
[core] Add version in BlobDescriptor

lucasfang:
### Purpose

Introduces a version field into class BlobDescriptor to establish a clear binary format for its serialized data, ensuring better compatibility in the future.


JingsongLi (approved):
+1

PR #6336, opened 2025-09-25
[python] Support getting catalog option from FileStoreTable

discivigour:
### Purpose

- Get catalog option from FileStoreTable when reading and writing table.

### Tests

- test_catalog_environment()

### DIFF
--- a/paimon-python/pypaimon/table/file_store_table.py
+++ b/paimon-python/pypaimon/table/file_store_table.py
@@ -19,6 +19,7 @@
 from pathlib import Path
 from typing import Optional
 
+from pypaimon.api.options import Options
 from pypaimon.catalog.catalog_environment import CatalogEnvironment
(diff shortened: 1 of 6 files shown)


paimon-python/pypaimon/table/file_store_table.py
```diff
@@ -104,3 +105,6 @@ def create_row_key_extractor(self) -> RowKeyExtractor:
             return DynamicBucketRowKeyExtractor(self.table_schema)
         else:
             raise ValueError(f"Unsupported bucket mode: {bucket_mode}")
+
+    def get_catalog_options(self) -> Options:
```
JingsongLi, line 109: What is this method for?

discivigour: Zhan yu mentioned he wants to get some options that do not exist when creating table.

PR #6343, opened 2025-09-26
[spark] Improve error msg for creating a function on an existing tmp function

Zouxxyy:
### Purpose

Improve error msg for creating a function on an existing tmp function


JingsongLi:
+1

PR #6344, opened 2025-09-28
[core] Support blob read and write

leaves12138:
### Purpose

Support read and write blob file in data evolution table.
Related PIP: https://cwiki.apache.org/confluence/display/PAIMON/PIP-35:+Introduce+Blob+to+store+multimodal+data


JingsongLi (approved):
+1

PR #6349, opened 2025-09-29
[core] Add deleted index generation to bucketed dv and hash maintainer

Zouxxyy:
### Purpose

Add deleted index generation to bucketed dv and hash maintainer.

Therefore, we can fully include the deleted index and added index in the commit message for subsequent conflict detection.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/compact/CompactDeletionFile.java
+++ b/paimon-core/src/main/java/org/apache/paimon/compact/CompactDeletionFile.java
@@ -57,7 +57,7 @@ public interface CompactDeletionFile {
     /** A generated files implementation of {@link CompactDeletionFile}. */
     class GeneratedDeletionFile implements CompactDeletionFile {
 
-        @Nullable private final IndexFileMeta deleteDeletionFile;
+        @Nullable private IndexFileMeta deleteDeletionFile;
         @Nullable private final IndexFileMeta newDeletionFile;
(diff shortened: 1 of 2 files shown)


paimon-core/src/main/java/org/apache/paimon/compact/CompactDeletionFile.java
```diff
@@ -81,18 +89,20 @@ public CompactDeletionFile mergeOldFile(CompactDeletionFile old) {
                 throw new IllegalStateException("old should not be get, this is a bug.");
             }
 
-            if (deletionFile == null) {
+            if (newDeletionFile == null) {
                 return old;
             }
 
             old.clean();
+            // Keep the old deletion file.
```
Zouxxyy, line 97: Here, we should keep the old deletion file

paimon-core/src/main/java/org/apache/paimon/index/DynamicBucketIndexMaintainer.java
```diff
@@ -87,9 +89,11 @@ public List<IndexFileMeta> prepareCommit() {
                 throw new RuntimeException(e);
             }
             modified = false;
-            return Collections.singletonList(entry);
+            IndexFileMeta toRemove = beforeFile;
```
Zouxxyy, line 92: In the streaming scenario, the index maintainer will always hold it and not create it repeatedly, so this logic needs to be added

PR #6359, opened 2025-10-01
[spark] Paimon Spark support describe table partition

liujiayi771:
### Purpose

Spark does not support DESCRIBE TABLE PARTITION for v2 tables. 

```
Error running query: [_LEGACY_ERROR_TEMP_1111] org.apache.spark.sql.AnalysisException: DESCRIBE does not support partition for v2 tables.
```

This PR supports this usage via a strategy rule.

### Tests

Add new test case in `DescribeTableTestBase`.

### API and Format

N/A

### Documentation

N/A

### DIFF
--- a/paimon-spark/paimon-spark-ut/src/test/java/org/apache/paimon/spark/SparkReadITCase.java
+++ b/paimon-spark/paimon-spark-ut/src/test/java/org/apache/paimon/spark/SparkReadITCase.java
@@ -359,7 +359,7 @@ public class SparkReadITCase extends SparkReadTestBase {
                         + "PARTITIONED BY (a)\n");
         assertThat(spark.sql("DESCRIBE PartitionedTable").collectAsList().toString())
                 .isEqualTo(
-                        "[[a,bigint,], [b,string,], [# Partition Information,,], [# col_name,data_type,comment], [a,bigint,]]");
+                        "[[a,bigint,null], [b,string,null], [# Partition Information,,], [# col_name,data_type,comment], [a,bigint,null]]");
     }
 
     @Test


liujiayi771:
cc @Zouxxyy @YannByron

paimon-spark/paimon-spark-ut/src/test/scala/org/apache/paimon/spark/sql/DescribeTableTestBase.scala
```diff
@@ -156,4 +157,71 @@ abstract class DescribeTableTestBase extends PaimonSparkTestBase {
     // check comment in schema
     Assertions.assertTrue(Objects.equals(comment, loadTable(tableName).schema().comment()))
   }
+
+  protected def describeTablePartitionCreateTable(): Unit = {
```
YannByron, line 161: This method don't have to be overwritten by sub-classes, right ?

paimon-spark/paimon-spark-ut/src/test/scala/org/apache/paimon/spark/sql/DescribeTableTestBase.scala
```diff
@@ -156,4 +157,71 @@ abstract class DescribeTableTestBase extends PaimonSparkTestBase {
...
+    checkAnswer(
+      res
+        .filter("col_name = 'name' AND comment != ''")
+        .select("col_name", "data_type", "comment"),
+      Row("name", "string", "'Bob'") :: Nil
+    )
+  }
+
+  test("Paimon describe: describe table partition") {
+    describeTablePartitionCreateTable()
```
YannByron, line 181: This is the only place which call `describeTablePartitionCreateTable`, so shall we create table here directly without any function call.

paimon-spark/paimon-spark-common/src/main/scala/org/apache/spark/sql/execution/PaimonDescribeTableExec.scala
```diff
@@ -0,0 +1,248 @@
...
+case class PaimonDescribeTableExec(
+    output: Seq[Attribute],
+    catalog: SparkBaseCatalog,
+    identifier: Identifier,
+    table: SparkTable,
+    partitionSpec: TablePartitionSpec,
+    isExtended: Boolean)
+  extends PaimonLeafV2CommandExec {
+
+  override protected def run(): Seq[InternalRow] = {
```
YannByron, line 47: Can we call `DescribeTableExec(output, table, isExtended).run()`  which has already implement in Spark at first, and then append the partition info if `partitionSpec` is not empty?

paimon-spark/paimon-spark-ut/src/test/scala/org/apache/paimon/spark/sql/DescribeTableTestBase.scala
liujiayi771: spark 3.2/3 not support default keyword, so they override this methods and create table without default keyword.

paimon-spark/paimon-spark-common/src/main/scala/org/apache/spark/sql/execution/PaimonDescribeTableExec.scala
liujiayi771: `DescribeTableExec.run` is protected.

paimon-spark/paimon-spark-ut/src/test/scala/org/apache/paimon/spark/sql/DescribeTableTestBase.scala
YannByron: ok

PR #6302, opened 2025-09-22
[core] fix Bug-NullPointException when readHadoopXml for issue#6282

linqichen:
### Purpose
fix bug: NullPointException when readHadoopXml

Linked issue:  (https://github.com/apache/paimon/issues/6282)

### Tests
image

### API and Format
not involve

### Documentation
not involve

PR #6361, opened 2025-10-02
[core] fix Bug-NullPointException when readHadoopXml,  for issue#6282 #6302

linqichen:
Purpose
fix bug: NullPointException when readHadoopXml

Linked issue: (https://github.com/apache/paimon/issues/6282)

Tests
image

API and Format
not involve

Documentation
not involve


paimon-common/src/main/java/org/apache/paimon/utils/HadoopUtils.java
```diff
@@ -221,13 +221,19 @@ public static void readHadoopXml(String xml, Configuration conf) {
         }
         for (int i = 0; i < propertyNodes.getLength(); i++) {
             Node propertyNode = propertyNodes.item(i);
-            if (propertyNode.getNodeType() == 1) {
+            if (1 == propertyNode.getNodeType()) {
```
wombatu-kun: it's better not to make unnecessary changes

paimon-common/src/main/java/org/apache/paimon/utils/HadoopUtils.java
```diff
@@ -221,13 +221,19 @@ public static void readHadoopXml(String xml, Configuration conf) {
...
+                        && null != propertyElement.getElementsByTagName("name").item(0)
+                        && null != propertyElement.getElementsByTagName("value")
+                        && null != propertyElement.getElementsByTagName("value").item(0)) {
+                    String name =
+                            propertyElement.getElementsByTagName("name").item(0).getTextContent();
+                    String value =
+                            propertyElement.getElementsByTagName("value").item(0).getTextContent();
+                    if (!StringUtils.isNullOrWhitespaceOnly(value)) {
+                        conf.set(name, value);
+                    }
```
wombatu-kun:
```
                NodeList keyElements = propertyElement.getElementsByTagName("name");
                NodeList valueElements = propertyElement.getElementsByTagName("value");
                if (keyElements.getLength() > 0
                        && keyElements.item(0) != null
                        && valueElements.getLength() > 0
                        && valueElements.item(0) != null) {
                    String value = valueElements.item(0).getTextContent();
                    if (!StringUtils.isNullOrWhitespaceOnly(value)) {
                        conf.set(keyElements.item(0).getTextContent(), value);
                    }
                }
```
Don't you think that this would look better?

wombatu-kun:
hi @linqichen ! It's all just my suggestions to make this PR better. I'm not a Paimon committer.

JingsongLi (commented):
NPE fix in `HadoopUtils.readHadoopXml`. The linked issue #6282 describes the crash scenario.

**Review:**

1. **+32/-6** is small. The fix likely adds null checks before accessing XML elements or attributes from the Hadoop configuration file.

2. **Test**: The screenshot shows a test case but it would be better to include the test in the PR diff. Is there a `HadoopUtilsITCase` test added?

3. **Root cause**: Which specific null reference causes the NPE? Is it:
   - A missing XML element in the config file?
   - An element with no text content?
   - A malformed XML file?

4. **Fix approach**: Null-defensive reads are appropriate for external config parsing. But ensure the fix either:
   - Skips the malformed entry with a warning (best for user experience)
   - Throws a descriptive error (if the entry is required)
   - Doesn't silently return empty/default values that could cause subtle issues later

The fix direction is right. LGTM assuming the test validates the NPE scenario.

JingsongLi:
+1

PR #6364, opened 2025-10-08
[spark] Spark write supports 'full-compaction.delta-commits'

Zouxxyy:
### Purpose

Spark write supports 'full-compaction.delta-commits'


paimon-core/src/main/java/org/apache/paimon/table/sink/TableWrite.java
```diff
@@ -57,10 +59,12 @@ default TableWrite withMemoryPool(MemorySegmentPool memoryPool) {
     int getBucket(InternalRow row);
 
     /** Write a row to the writer. */
-    void write(InternalRow row) throws Exception;
+    @Nullable
+    SinkRecord write(InternalRow row) throws Exception;
```
JingsongLi: SinkRecord is not public API, should not return it here.

PR #6365, opened 2025-10-09
[core] spark: support format table write

jerry-024:
### Purpose
spark: support format table write

### Tests
- PaimonFormatTableTest


JingsongLi (approved):
+1

PR #6391, opened 2025-10-13
[Python] Introduce incremental-between read by timestamp

discivigour:
### Purpose

Read incremental changes between start timestamp (exclusive) and end timestamp, for example, 't1,t2' means changes between timestamp t1 and timestamp t2.

### Tests

- test_incremental_timestamp()

### DIFF
--- a/paimon-python/pypaimon/read/table_scan.py
+++ b/paimon-python/pypaimon/read/table_scan.py
@@ -20,16 +20,8 @@ from typing import List, Optional
 
 from pypaimon.common.core_options import CoreOptions
 from pypaimon.common.predicate import Predicate
-from pypaimon.common.predicate_builder import PredicateBuilder
-from pypaimon.manifest.manifest_file_manager import ManifestFileManager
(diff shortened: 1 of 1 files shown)


paimon-python/pypaimon/read/scanner/incremental_starting_scanner.py
```diff
@@ -0,0 +1,77 @@
...
+        super().__init__(table, predicate, limit, read_type)
+        self.startingSnapshotId = start
+        self.endingSnapshotId = end
+
+    def plan_files(self) -> List[ManifestEntry]:
+        snapshots_in_range = []
+        for snapshot_id in range(self.startingSnapshotId + 1, self.endingSnapshotId + 1):
+            snapshot = self.snapshot_manager.get_snapshot_by_id(snapshot_id)
+            if snapshot:
+                snapshots_in_range.append(snapshot)
```
leaves12138, line 39: If delta scan, we should skip APPEND && OVERWRITE snapshots

leaves12138 (approved):
+1

PR #6392, opened 2025-10-13
[core] Blob support generate from input stream

leaves12138:
### Purpose

Related to PIP: https://cwiki.apache.org/confluence/display/PAIMON/PIP-35:+Introduce+Blob+to+store+multimodal+data

Blob data could be loaded from input stream.


paimon-common/src/main/java/org/apache/paimon/data/Blob.java
```diff
@@ -63,4 +64,12 @@ static Blob fromFile(FileIO fileIO, String file, long offset, long length) {
     static Blob fromDescriptor(UriReader reader, BlobDescriptor descriptor) {
         return new BlobRef(reader, descriptor);
     }
+
+    static Blob fromInputStream(InputStream inputStream) {
```
JingsongLi: Maybe the API should be `fromInputStream(Supplier<SeekableInputStream> supplier)`.

paimon-common/src/main/java/org/apache/paimon/fs/SeekableInputStream.java
```diff
@@ -64,4 +64,38 @@ public abstract class SeekableInputStream extends InputStream {
      * @exception IOException if an I/O error occurs.
      */
     public abstract void close() throws IOException;
+
+    /**
+     * Wrap a normal {@link InputStream} to a {@link SeekableInputStream}. The returned {@link
+     * SeekableInputStream} does not support seek operations.
+     */
+    public static SeekableInputStream wrap(InputStream inputStream) {
```
JingsongLi, line 72: You can remove `LimitedSeekableInputStream` class.

paimon-common/src/main/java/org/apache/paimon/data/Blob.java
leaves12138: OK

paimon-common/src/main/java/org/apache/paimon/fs/SeekableInputStream.java
leaves12138: OK

PR #6403, opened 2025-10-15
[python] Support blob write

leaves12138:
### Purpose

Related PIP: https://cwiki.apache.org/confluence/display/PAIMON/PIP-35:+Introduce+Blob+to+store+multimodal+data
Python write support blob type


discivigour:
+1

PR #6404, opened 2025-10-15
[core]Python: fix blob write when blob_as_descriptor is true

jerry-024:
### Purpose
Python: fix blob write when blob_as_descriptor is true

### Tests
- blob_test.BlobEndToEndTest.test_blob_end_to_end_with_descriptor


leaves12138:
Please add blob-as-descriptor for read as well.

PR #6407, opened 2025-10-16
[core] Introduce deletion vector meta cache at bucket level

bryndenZh:
### Purpose

In high-concurrency point query scenarios on the primary key table, we observed high CPU usage mainly caused by deserialization overhead of DV metadata. Currently, reading deletion vector metadata for a single bucket requires reading and deserializing a large number of entries from the index manifest, if the table has many partition and buckets.
image

This PR introduces a bucket-level dv meta cache which  reduces CPU load and significantly improves QPS for single-bucket query scenarios on primary key tables.

### DIFF
--- a/docs/layouts/shortcodes/generated/catalog_configuration.html
+++ b/docs/layouts/shortcodes/generated/catalog_configuration.html
@@ -33,6 +33,12 @@ under the License.
             <td>Controls whether the catalog will cache databases, tables, manifests and partitions.</td>
         </tr>
         <tr>
+            <td><h5>cache.deletion-vectors.max-bucket-num</h5></td>
+            <td style="word-wrap: break-word;">20000</td>
+            <td>Integer</td>
+            <td>Controls the maximum number of bucket-level deletion vector meta that can be cached.</td>
+        </tr>
+        <tr>
             <td><h5>cache.expire-after-access</h5></td>
             <td style="word-wrap: break-word;">10 min</td>
             <td>Duration</td>


paimon-api/src/main/java/org/apache/paimon/options/CatalogOptions.java
```diff
@@ -136,6 +136,19 @@ public class CatalogOptions {
                     .withDescription(
                             "Controls the max number for snapshots per table in the catalog are cached.");
 
+    public static final ConfigOption<Boolean> CACHE_DV_ENABLED =
+            key("cache.dv.enabled")
+                    .booleanType()
+                    .defaultValue(false)
```
JingsongLi: Maybe we should enable this by default.

paimon-api/src/main/java/org/apache/paimon/options/CatalogOptions.java
```diff
@@ -136,6 +136,19 @@ public class CatalogOptions {
                     .withDescription(
                             "Controls the max number for snapshots per table in the catalog are cached.");
 
+    public static final ConfigOption<Boolean> CACHE_DV_ENABLED =
+            key("cache.dv.enabled")
```
JingsongLi: We don't need this one, just use max num to decide whether we should enable dv cache. (Max num is zero, disable)

paimon-api/src/main/java/org/apache/paimon/options/CatalogOptions.java
```diff
@@ -136,6 +136,19 @@ public class CatalogOptions {
...
+                    .booleanType()
+                    .defaultValue(false)
+                    .withDescription("Whether to enable deletion vector meta cache.");
+
+    public static final ConfigOption<Integer> CACHE_DV_MAX_NUM =
+            key("cache.dv.max-num")
```
JingsongLi: cache.deletion-vectors.max-num

paimon-core/src/main/java/org/apache/paimon/index/IndexFileHandler.java
```diff
@@ -87,6 +101,123 @@ public Optional<IndexFileMeta> scanHashIndex(
...
+    // Construct DataFile -> DeletionFile based on IndexFileMeta
+    public Map<String, DeletionFile> extractDeletionFileByMeta(
+            BinaryRow partition, Integer bucket, IndexFileMeta fileMeta) {
+        if (fileMeta.dvRanges() != null && fileMeta.dvRanges().size() > 0) {
+            Map<String, DeletionFile> result = new HashMap<>();
+            for (DeletionVectorMeta dvMeta : fileMeta.dvRanges().values()) {
+                result.put(
+                        dvMeta.dataFileName(),
+                        new DeletionFile(
+                                dvIndex(partition, bucket).path(fileMeta).toString(),
```
JingsongLi: DeletionVectorsIndexFile dvIndex = dvIndex(partition, bucket); to avoid create everytime.

paimon-core/src/main/java/org/apache/paimon/index/IndexFileHandler.java
```diff
@@ -87,6 +101,123 @@ public Optional<IndexFileMeta> scanHashIndex(
...
 
+    public void withCacheMetrics(@Nullable CacheMetrics cacheMetrics) {
+        this.cacheMetrics = cacheMetrics;
+    }
+
+    @Nullable
+    // Construct DataFile -> DeletionFile based on IndexFileMeta
+    public Map<String, DeletionFile> extractDeletionFileByMeta(
+            BinaryRow partition, Integer bucket, IndexFileMeta fileMeta) {
+        if (fileMeta.dvRanges() != null && fileMeta.dvRanges().size() > 0) {
```
JingsongLi: extract a local field for `fileMeta.dvRanges()`.

paimon-core/src/main/java/org/apache/paimon/index/IndexFileHandler.java
```diff
@@ -87,6 +101,123 @@ public Optional<IndexFileMeta> scanHashIndex(
         return result.isEmpty() ? Optional.empty() : Optional.of(result.get(0));
     }
 
+    public void withCacheMetrics(@Nullable CacheMetrics cacheMetrics) {
+        this.cacheMetrics = cacheMetrics;
+    }
+
+    @Nullable
+    // Construct DataFile -> DeletionFile based on IndexFileMeta
+    public Map<String, DeletionFile> extractDeletionFileByMeta(
```
JingsongLi, line 110: Remove public and add `@VisibleForTesting`.

paimon-core/src/main/java/org/apache/paimon/index/IndexFileHandler.java
```diff
@@ -87,6 +101,123 @@ public Optional<IndexFileMeta> scanHashIndex(
...
+    // returns <DataFile: DeletionFile> map grouped by partition and bucket
+    public Map<Pair<BinaryRow, Integer>, Map<String, DeletionFile>> scanDVIndex(
+            Snapshot snapshot, Set<Pair<BinaryRow, Integer>> partitionBuckets) {
+        Map<Pair<BinaryRow, Integer>, List<IndexFileMeta>> partitionFileMetas =
+                scan(
+                        snapshot,
+                        DELETION_VECTORS_INDEX,
+                        partitionBuckets.stream().map(Pair::getLeft).collect(Collectors.toSet()));
+        Map<Pair<BinaryRow, Integer>, Map<String, DeletionFile>> result = new HashMap<>();
+        partitionBuckets.forEach(
```
JingsongLi: Just use ` partitionFileMetas.forEach`?

paimon-core/src/main/java/org/apache/paimon/index/IndexFileHandler.java
```diff
@@ -52,20 +58,28 @@ public class IndexFileHandler {
...
 
     public IndexFileHandler(
             FileIO fileIO,
             SnapshotManager snapshotManager,
             IndexManifestFile indexManifestFile,
             IndexFilePathFactories pathFactories,
             MemorySize dvTargetFileSize,
-            boolean dvBitmap64) {
+            boolean dvBitmap64,
+            boolean enableDVMetaCache) {
```
JingsongLi: Why not just pass `DVMetaCache` here? You don't need use `IndexManifestFile` for cache at all.

paimon-core/src/main/java/org/apache/paimon/index/IndexFileHandler.java
```diff
@@ -87,6 +101,123 @@ public Optional<IndexFileMeta> scanHashIndex(
...
+                                    }
+                                });
+                    }
+                });
+        return result;
+    }
+
+    // Scan DV Meta Cache first, if not exist, scan DV index file, returns the exact deletion file
+    // of the specified partition/buckets
+    public Map<String, DeletionFile> scanDVIndexWithCache(
```
JingsongLi: We can just have one `Map<Pair<BinaryRow, Integer>, Map<String, DeletionFile>> scanDVIndex(Snapshot snapshot, Set<Pair<BinaryRow, Integer>> partitionBuckets)`, we can deal with cache in it.

paimon-core/src/main/java/org/apache/paimon/index/IndexFileHandler.java
```diff
@@ -87,6 +101,123 @@ public Optional<IndexFileMeta> scanHashIndex(
...
+            Snapshot snapshot, BinaryRow partition, Integer bucket) {
+        if (snapshot == null || snapshot.indexManifest() == null) {
+            return Collections.emptyMap();
+        }
+        // read from cache
+        String indexManifestName = snapshot.indexManifest();
+        Map<String, DeletionFile> result =
+                indexManifestFile.readFromDVMetaCache(indexManifestName, partition, bucket);
```
JingsongLi: Inline `indexManifestFile.readFromDVMetaCache`.

paimon-core/src/main/java/org/apache/paimon/index/IndexFileHandler.java
```diff
@@ -87,6 +101,123 @@ public Optional<IndexFileMeta> scanHashIndex(
...
+                                    extractDeletionFileByMeta(
+                                            partitionBucket.getLeft(),
+                                            partitionBucket.getRight(),
+                                            meta);
+                            if (bucketDVMetas != null) {
+                                bucketDeletionFiles.putAll(bucketDVMetas);
+                            }
+                        });
+                // bucketDeletionFiles can be empty
+                indexManifestFile.fillDVMetaCache(
```
JingsongLi: Inline `indexManifestFile.fillDVMetaCache`.

paimon-core/src/main/java/org/apache/paimon/utils/DVMetaCache.java
```diff
@@ -0,0 +1,154 @@
...
+                                    file.path(),
+                                    (int) file.offset(),
+                                    (int) file.length(),
+                                    file.cardinality());
+                    cacheValue.add(dvMetaCacheValue);
+                });
+        this.cache.put(key, cacheValue);
+    }
+
+    private static class DVMetaCacheValue extends DeletionVectorMeta {
```
JingsongLi: Don't `extends DeletionVectorMeta`.

paimon-core/src/main/java/org/apache/paimon/utils/DVMetaCache.java
```diff
@@ -0,0 +1,154 @@
...
+        private final String fileName;
+
+        public DVMetaCacheValue(
+                String fileName,
+                String dataFileName,
```
JingsongLi, line 96: It is `deletionFilePath`.

paimon-core/src/main/java/org/apache/paimon/utils/DVMetaCache.java
```diff
@@ -0,0 +1,154 @@
...
+                    cacheValue.add(dvMetaCacheValue);
+                });
+        this.cache.put(key, cacheValue);
+    }
+
+    private static class DVMetaCacheValue extends DeletionVectorMeta {
+        private final String fileName;
+
+        public DVMetaCacheValue(
+                String fileName,
```
JingsongLi: It is dataFileName.

paimon-core/src/main/java/org/apache/paimon/utils/DVMetaCache.java
```diff
@@ -0,0 +1,154 @@
...
+import java.util.Map;
+import java.util.Objects;
+
+/** Cache for deletion vector meta. */
+public class DVMetaCache {
+    private final Cache<DVMetaCacheKey, List<DVMetaCacheValue>> cache;
+
+    public DVMetaCache(long maxElementSize) {
+        this.cache =
+                Caffeine.newBuilder().maximumSize(maxElementSize).executor(Runnable::run).build();
```
JingsongLi: Should it be set to max number of `DVMetaCacheValue`? And use `softValues`?

bryndenZh: I've added softValues. Do you mean to use `List<DVMetaCacheValue>` as the weight? I think the List<DVMetaCacheValue> might be empty, and those cache items won't be restricted because of the 0 weight

paimon-core/src/main/java/org/apache/paimon/index/IndexFileHandler.java
bryndenZh: remove public will lead to compilation problems

paimon-core/src/main/java/org/apache/paimon/index/IndexFileHandler.java
bryndenZh: I think unnecessary buckets need to be filtered out?

PR #6413, opened 2025-10-16
[filesystems] add fs.oss.sld.enabled to support oss private link

shyjsarah:
### Purpose

Linked issue: close #xxx

In cross-cloud scenarios, OSS needs to establish a private connection channel through PrivateLink. 
This requires the OSSClient to enable secondary domain name access, but AliyunOSSFileSystem is not configured to support this. This PR introduces `fs.oss.sld.enabled` to support.


JingsongLi:
+1

PR #6419, opened 2025-10-16
[python] Filter manifest files by partition predicate in scan

JingsongLi:
### Purpose

Improve performance. We should only read manifest files belong to partition which we need to read.

### DIFF
--- a/paimon-python/pypaimon/read/push_down_utils.py
+++ b/paimon-python/pypaimon/read/push_down_utils.py
@@ -42,6 +42,7 @@ def to_partition_predicate(input_predicate: 'Predicate', all_fields: List[str], 
 
     return _change_index(new_predicate, mapping)
 
+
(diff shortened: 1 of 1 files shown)


paimon-python/pypaimon/read/push_down_utils.py
```diff
@@ -21,6 +21,50 @@
...
+    part_to_index = {element: idx for idx, element in enumerate(partition_keys)}
+    mapping: Dict[int, int] = {
+        i: part_to_index.get(all_fields[i], -1)
+        for i in range(len(all_fields))
+    }
+
+    return change_index(new_predicate, mapping)
+
+
+def split_and(input_predicate: 'Predicate'):
```
chenghuichen: Consider making `split_and` and `change_index` private — they’re not part of push_down_utils’ public API.

PR #6421, opened 2025-10-17
[spark] Introduce a base class for SparkTable to support multi-version APIs

kerwin-zk:
### Purpose

The single SparkTable implementation in the common module makes it difficult to support different Spark API versions, as newer versions like Spark 3.5 introduce interfaces that are absent in older ones.

This commit refactors the structure by:
- Extracting common, version-agnostic logic from `SparkTable` into a new `PaimonSparkTableBase` class.
- Allowing version-specific modules (e.g., for Spark 3.3, 3.4, 3.5) to provide their own `SparkTable` implementations that extend this base class.

This new design isolates version-specific code, making it much cleaner to add support for new Spark versions and maintain existing ones.

### Tests

CI


YannByron:
+1

PR #6427, opened 2025-10-18
[core] Disable blob type force spill yet

leaves12138:
### Purpose

Blob type now in data-evolution mode is not support read it back in writer and it should not call toBufferWriter method (Cause blob is big, read it back cost too much)（Blob does not support spill yet)


paimon-core/src/main/java/org/apache/paimon/operation/BaseAppendFileStoreWrite.java
```diff
@@ -235,6 +236,9 @@ protected void forceBufferSpill() throws Exception {
         if (ioManager == null) {
             return;
         }
+        if (writeType.getFieldTypes().stream().anyMatch(t -> t.is(BLOB))) {
```
JingsongLi: Add a class field `boolean withBlob`?

leaves12138: OK

PR #6432, opened 2025-10-20
[flink] disable clustering during writing if incremental clustering enabled

LsomeYeah:
### Purpose

Linked issue: close #xxx

Disable clustering operation during writing if incremental clustering is enabled.


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkTableSinkBase.java
```diff
@@ -136,6 +137,7 @@ public SinkRuntimeProvider getSinkRuntimeProvider(Context context) {
                                             dataStream.getExecutionEnvironment(),
                                             dataStream.getTransformation()))
                             .clusteringIfPossible(
+                                    conf.get(CLUSTERING_INCREMENTAL),
```
JingsongLi: Do not pass this bool, Just avoid invoking this method.

PR #6433, opened 2025-10-20
[Python] clean code for pypaimon

chenghuichen:
### Purpose
- Use dict to dispatch method for `predicate.py`.
- Remove function `predicate.test_by_value`, as it is no longer in use.
- Simplify the code by leveraging new utility functions such as `get_trimmed_primary_keys()` and `field_names`.
- Move `push_down_predicate` from `table_read` to `split_read `to eliminate unnecessary parameter passing.


paimon-python/pypaimon/common/predicate.py
```diff
@@ -50,93 +50,50 @@ def new_literals(self, literals: List[Any]):
             literals=literals)
 
     def test(self, record: InternalRow) -> bool:
+        if self.method == 'and':
```
JingsongLi, line 53: Can we create a `dict[str, function]` for this?

chenghuichen: Good idea.

PR #6435, opened 2025-10-20
[spark] Refine paimon_sparkx_common dependency

jackylee-ch:
### Purpose

While executing `mvn test -Pspark4 -pl "org.apache.paimon:paimon-spark-4.0_2.13"`, I encountered a failure in the unit tests because `paimon-spark3-common` was being loaded. 
image
Further investigation revealed that when running `mvn dependency:tree -pl "org.apache.paimon:paimon-spark-4.0_2.13"`, I could see the `paimon-spark3-common` dependency brought in via `paimon-spark-ut_2.13`. The root cause is that the profile defined in `paimon-spark-ut_2.13-1.4.0-SNAPSHOT.jar` was not being propagated, which caused the default `paimon-sparkx-common` configuration to be applied.
image

This PR resolves the issue by explicitly specifying the profile, ensuring that the profile metadata is passed along, and thus enabling `mvn test -Pspark4 -pl "org.apache.paimon:paimon-spark-4.0_2.13"` to run successfully.

### Tests

GA

### API and Format

No.

### Documentation

No.


Zouxxyy:
Hi, I think I'm encountering the same issue. I tried your PR, but it didn't work for me. So I've submitted a new one—could you please give it a try? Thanks! https://github.com/apache/paimon/pull/7065

Zouxxyy:
#7065 This was only avoided by using the `exclude` method. If you have a better way, please feel free to share it

JingsongLi (commented):
Good Maven dependency fix. The problem is clear: `paimon-spark-ut` depends on `paimon-sparkx-common` via profile, but when running with `-Pspark4`, the profile isn't propagated from the dependency's POM, causing the wrong Spark version to be loaded.

+2/-30 suggests this simplifies the POM files by making the dependency explicit per Spark version rather than relying on profile propagation.

The fix is verified by GA (GitHub Actions). LGTM.

jackylee-ch:
@JingsongLi Hello, I see that the current PR is currently closed without merged. If it look good, would you consider merging it in?

PR #6437, opened 2025-10-20
[fix] Trim branch name when deleting branch

xiaochen-zhou:
### Purpose

Trim branch name when batch deleting branchs

### Tests

Exist tests.


JingsongLi (commented):
Simple defensive fix. Adding `.trim()` when processing branch names for deletion handles trailing whitespace from user input (e.g., copy-paste from docs or CLI).

+4/-2 is minimal. The change is in `Table.java` — presumably in the `deleteBranch` or batch-delete method.

LGTM. Consider whether similar trimming should be applied to other name-based operations (create branch, create tag, rename tag) for consistency.

PR #6438, opened 2025-10-20
[test][rest] add test case for two sessions with cache for rest commitTable

XiaoHongbo-Hope:
### Purpose

add test case for two sessions with cache for rest commitTable

### Tests
PaimonSparkTwoSessionCacheTest


JingsongLi:
+1

PR #6441, opened 2025-10-21
[flink] Produce real random id in SourceSplitGenerator

JingsongLi:
### Purpose

Flink's Source requires us to generate completely different IDs. Otherwise, if the same ID is generated after a failed restart, it may cause issues with the Source's state, resulting in Split being cleared and causing exceptions.


leonardBang (commented):
Thanks @JingsongLi for the fix, the changes looks good to me, Could you override FileStoreSourceSplit#toString method so that we can see splitId from log as well?

leonardBang (approved):
Thanks for the update，LGTM

PR #6443, opened 2025-10-21
[core] Bug: disable data-evolution manifest filter for now

leaves12138:
### Purpose

Data evolution file manifest filter is not easy, so disable it for now.


paimon-core/src/main/java/org/apache/paimon/AppendOnlyFileStore.java
```diff
@@ -168,6 +169,17 @@ public AppendOnlyFileStoreScan newScan() {
                     return Optional.empty();
                 };
 
+        if (options().dataEvolutionEnabled()) {
```
JingsongLi, line 172: No partition filter push down?

leaves12138: Partition predicate is not here

JingsongLi (approved):
+1

PR #6444, opened 2025-10-21
[Python] SimpleStats supports BinaryRow

discivigour:
### Purpose

When scanning the manifest, in order to speed up the reading, value_stats and key_stats are not parsed in advance. The specified columns are selected for parsing during filtering by stats.

### DIFF
--- a/paimon-python/pypaimon/table/row/generic_row.py
+++ b/paimon-python/pypaimon/table/row/generic_row.py
@@ -28,6 +28,7 @@ from pypaimon.table.row.binary_row import BinaryRow
 from pypaimon.table.row.internal_row import InternalRow, RowKind
 from pypaimon.table.row.blob import BlobData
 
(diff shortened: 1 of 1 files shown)


paimon-python/pypaimon/table/row/binary_row.py
```diff
@@ -0,0 +1,45 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+from typing import Any
+from pypaimon.schema.data_types import DataType
+
+
+class BinaryRow:
+
+    def __init__(self, data: bytes):
```
JingsongLi: Can we pass `fields: List[DataField]` to `BinaryRow` and make this `BinaryRow` extends from `InternalRow`?

JingsongLi (approved):
+1

PR #6447, opened 2025-10-22
[flink] Revert method of FileStoreSourceSplitGenerator

yuzelin:
### Purpose

The method `createSplits(List<Split> splits)` was used by other system.


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/FileStoreSourceSplitGenerator.java
```diff
@@ -35,7 +36,11 @@ public class FileStoreSourceSplitGenerator {
     private final AtomicInteger idCounter = new AtomicInteger(1);
 
     public List<FileStoreSourceSplit> createSplits(TableScan.Plan plan) {
```
JingsongLi, line 38: Can you remove this method?

yuzelin:
I think it's OK to just use `createSplits(TableScan.Plan plan)`.

PR #6448, opened 2025-10-22
[hotfix] Fix bug of comparing string and int in row_key_extractor

universe-hcy:
### Purpose

fix bug of comparing string and int in row_key_extractor.py

### Tests

add ut test_with_shard_ao_unaware_bucket_manual: test with setting bucket -1 manually


JingsongLi:
+1

PR #6455, opened 2025-10-22
[core] Enable manifest filter in data evolution table

leaves12138:
### Purpose

It is not easy to filter manifest entry for data evolution table

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/reader/DataEvolutionArray.java
+++ b/paimon-common/src/main/java/org/apache/paimon/reader/DataEvolutionArray.java
@@ -40,10 +40,6 @@ public class DataEvolutionArray implements InternalArray {
         this.fieldOffsets = fieldOffsets;
     }
 
-    public int rowNumber() {
-        return rows.length;
-    }
-
     public void setRow(int pos, InternalArray row) {
         if (pos >= rows.length) {
             throw new IndexOutOfBoundsException(


paimon-core/src/main/java/org/apache/paimon/operation/DataEvolutionFileStoreScan.java
```diff
@@ -47,13 +69,281 @@ public DataEvolutionFileStoreScan(
...
+                    SpecialFields.rowTypeWithRowTracking(dataFileSchema.logicalRowType())
+                            .getFields().stream()
+                            .mapToInt(DataField::id)
+                            .toArray();
+
+            int count = 0;
+            for (int j = 0; j < fieldsCount; j++) {
+                for (int fieldId : fieldIds) {
+                    if (allFields[j] == fieldId) {
+                        // TODO: If type not match (e.g. int -> string), we need to skip this, set
```
JingsongLi, line 146: What this TODO mean? What case it could improve?

paimon-core/src/main/java/org/apache/paimon/operation/DataEvolutionFileStoreScan.java
```diff
@@ -47,13 +69,281 @@ public DataEvolutionFileStoreScan(
...
+        return new SimpleStatsEvolution.Result(finalMin, finalMax, finalNullCounts);
+    }
+
     /** Note: Keep this thread-safe. */
     @Override
     protected boolean filterByStats(ManifestEntry entry) {
         return true;
     }
+
+    private static class FakeDataFileMeta implements DataFileMeta {
```
JingsongLi: I don't get why we need to have this fake.

paimon-core/src/main/java/org/apache/paimon/operation/DataEvolutionFileStoreScan.java
leaves12138: If schema evolution, changes string to int, predicate is a(which is now int, previous is string) > 1. Then I should ignore this predicate if the type in file is string. Otherwise, may cause ClassCastException(String could not be cast to int)

PR #6457, opened 2025-10-22
[core] check exists before open file

cxzl25:
### Purpose

Linked issue: close #xxx

When writing an empty table, in the commit stage, some exceptions of FileNotFoundException are always thrown, although it does not affect the final execution result.

```
Caused by: org.apache.hadoop.ipc.RemoteException(java.io.FileNotFoundException): File does not exist: /user/hive/warehouse/xx.db/xxx/snapshot/EARLIEST
```

```
Caused by: org.apache.hadoop.ipc.RemoteException(java.io.FileNotFoundException): File does not exist: /user/hive/warehouse/xx.db/xxx/snapshot/LATEST
```


JingsongLi (changes requested):
-1. This will bring another file io.

PR #6461, opened 2025-10-23
[core] disable dv mode for incremental clustering table

LsomeYeah:
### Purpose

Linked issue: close #xxx

Currently, incremental clustering does not take the DV mode into account. We will temporarily disable DV and add support for it in the future.


JingsongLi:
+1

PR #6465, opened 2025-10-24
[Python] Blob read supports with_shard

discivigour:
### Purpose

[Python] Blob read supports with_shard

### Tests

- DataEvolutionTest
- DataBlobWriterTest


JingsongLi:
+1

PR #6466, opened 2025-10-24
[Doc] Supplementary the document of python REST API

discivigour:
### Purpose

Add example codes to supplementary the document of python REST API.


JingsongLi (approved):
+1

PR #6470, opened 2025-10-24
[core] format table: jindo io use upload part to support two-phase commit

jerry-024:
### Purpose
format table: jindo io use upload part to support two-phase commit


paimon-common/src/main/java/org/apache/paimon/fs/MultiPartUploadTwoPhaseOutputStream.java
```diff
@@ -89,10 +89,34 @@ public void write(byte[] b, int off, int len) throws IOException {
         if (closed) {
             throw new IOException("Stream is closed");
         }
-        buffer.write(b, off, len);
-        position += len;
-        if (buffer.size() >= partSizeThreshold()) {
-            uploadPart();
+        int remaining = len;
```
jerry-024, line 97: If write data is larger than the threshold, we need to split it.

JingsongLi (approved):
+1

PR #6472, opened 2025-10-26
[core] support automatically clustering historical partition

LsomeYeah:
### Purpose

Linked issue: close #xxx

Introduce three options to support automatically performing full clustering for historical partition.

1. `clustering.history-partition.auto.enabled`. Whether to allow  performing full clustering for historical partition.
2. `clustering.partition.idle-time`. The duration after which a partition without new updates is considered a historical partition.
3. `clustering.history-partition.limit`. The limit of history partition number for automatically performing full clustering.

When clustering.history-partition.auto.enabled is set to true, submitting an incremental clustering job with specified partitions will not only trigger incremental clustering on those partitions, but also automatically detect historical partitions that still contain low-level files and perform full clustering on them.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/append/cluster/IncrementalClusterManager.java
+++ b/paimon-core/src/main/java/org/apache/paimon/append/cluster/IncrementalClusterManager.java
@@ -19,11 +19,9 @@
 package org.apache.paimon.append.cluster;
 
 import org.apache.paimon.CoreOptions;
-import org.apache.paimon.annotation.VisibleForTesting;
 import org.apache.paimon.compact.CompactUnit;
 import org.apache.paimon.data.BinaryRow;
 import org.apache.paimon.io.DataFileMeta;
-import org.apache.paimon.manifest.PartitionEntry;
 import org.apache.paimon.mergetree.LevelSortedRun;
 import org.apache.paimon.mergetree.SortedRun;
 import org.apache.paimon.partition.PartitionPredicate;
@@ -32,7 +30,6 @@ import org.apache.paimon.table.FileStoreTable;
 import org.apache.paimon.table.source.DataSplit;
 import org.apache.paimon.table.source.SplitGenerator;
 import org.apache.paimon.table.source.snapshot.SnapshotReader;
-import org.apache.paimon.utils.BiFilter;
 import org.apache.paimon.utils.InternalRowPartitionComputer;
 
 import org.slf4j.Logger;
@@ -40,18 +37,12 @@ import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nullable;
 
-import java.time.Duration;
-import java.time.LocalDateTime;
-import java.time.ZoneId;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.Set;
 import java.util.stream.Collectors;
 
 import static org.apache.paimon.CoreOptions.CLUSTERING_INCREMENTAL;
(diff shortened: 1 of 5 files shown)


paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -1957,6 +1957,29 @@ public InlineElement getDescription() {
                     .defaultValue(false)
                     .withDescription("Whether enable incremental clustering.");
 
+    public static final ConfigOption<Boolean> CLUSTERING_HISTORY_PARTITION_AUTO =
+            key("clustering.history-partition.auto.enabled")
```
JingsongLi: just introduce two options:
1. clustering.history-partition.idle-to-full-sort: Duration.
2. clustering.history-partition.limit: Integer.

paimon-core/src/main/java/org/apache/paimon/append/cluster/IncrementalClusterManager.java
```diff
@@ -51,13 +60,19 @@
...
+    private final InternalRowPartitionComputer partitionComputer;
 
     private final SnapshotReader snapshotReader;
 
     private final IncrementalClusterStrategy incrementalClusterStrategy;
     private final CoreOptions.OrderType clusterCurve;
     private final List<String> clusterKeys;
 
+    private final boolean historyPartitionAutoCluster;
+    @Nullable private final Duration partitionIdleTime;
```
JingsongLi: historyPartitionIdleTime

paimon-core/src/main/java/org/apache/paimon/append/cluster/IncrementalClusterManager.java
```diff
@@ -69,14 +84,35 @@ public IncrementalClusterManager(
...
+                        table.coreOptions().legacyPartitionName());
+
+        this.specifiedPartitions = specifiedPartitions;
+
+        // config for history partition auto clustering
+        this.historyPartitionAutoCluster = options.clusteringHistoryPartitionAutoEnabled();
+        this.partitionIdleTime = options.clusteringPartitionIdleTime();
+        this.historyPartitionLimit = options.clusteringHistoryPartitionLimit();
+
+        if (historyPartitionAutoCluster) {
```
JingsongLi: Maybe we can do a dedicated snapshot read for history partition clustering, with level filter.

LsomeYeah: I added a `manifestLevelFilter` in `ManifestReader` to filter the manifest file according to the minLevel value. 
For `readPartitionEntries` in history clustering, if the minLevel > maxLevel/2, we can filter this manifest file.

paimon-core/src/main/java/org/apache/paimon/append/cluster/IncrementalClusterManager.java
```diff
@@ -169,11 +219,30 @@ public Map<BinaryRow, List<LevelSortedRun>> constructLevels() {
...
-                    .computeIfAbsent(dataSplit.partition(), k -> new ArrayList<>())
-                    .addAll(dataSplit.dataFiles());
+        Map<BinaryRow, List<DataFileMeta>> partitionFiles = getPartitionFiles(dataSplits);
+
+        if (specifiedPartitions != null
+                && historyPartitionIdleTime != null
+                && historyPartitionLimit > 0) {
+            List<DataSplit> splitsForHistory =
+                    table.newSnapshotReader()
+                            .withManifestLevelFilter(partitionLevelFilter)
```
JingsongLi: Filter is just for finding history partitions, you can not use them to do full cluster.

paimon-core/src/main/java/org/apache/paimon/append/cluster/IncrementalClusterManager.java
```diff
@@ -118,11 +153,26 @@ public Map<BinaryRow, CompactUnit> prepareForCluster(boolean fullCompaction) {
...
+                                        entry -> {
+                                            if (fullCompaction) {
+                                                return incrementalClusterStrategy.pick(
+                                                        maxLevel, entry.getValue(), true);
+                                            } else {
+                                                // if clustering is not in full mode, then specified
+                                                // partitions should perform incremental clustering
+                                                // and history partitions should perform full
+                                                // clustering
+                                                if (specifiedPartitions != null
```
JingsongLi: Don't need to modify here?

paimon-core/src/main/java/org/apache/paimon/append/cluster/IncrementalClusterManager.java
```diff
@@ -51,12 +61,19 @@
...
 
+    private final FileStoreTable table;
     private final SnapshotReader snapshotReader;
 
     private final IncrementalClusterStrategy incrementalClusterStrategy;
     private final CoreOptions.OrderType clusterCurve;
     private final List<String> clusterKeys;
+    @Nullable private final PartitionPredicate specifiedPartitions;
+
+    @Nullable private final Duration historyPartitionIdleTime;
```
JingsongLi: Create a dedicated class: HistoryPartitionCluster.

PR #6474, opened 2025-10-27
[core] format table: update format-table.implementation default value from engine to paimon

jerry-024:
### Purpose

format table: update `format-table.implementation` default value from `engine` to `paimon`

### DIFF
--- a/paimon-spark/paimon-spark-ut/src/test/java/org/apache/paimon/spark/SparkCatalogWithHiveTest.java
+++ b/paimon-spark/paimon-spark-ut/src/test/java/org/apache/paimon/spark/SparkCatalogWithHiveTest.java
@@ -65,9 +65,7 @@ public class SparkCatalogWithHiveTest {
 
             // test orc table
 
-            spark.sql(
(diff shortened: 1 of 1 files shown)


paimon-spark/paimon-spark-ut/src/test/scala/org/apache/paimon/spark/sql/FormatTableTestBase.scala
```diff
@@ -132,7 +137,8 @@ abstract class FormatTableTestBase extends PaimonHiveTestBase {
 
   test("Format table: field delimiter in HMS") {
     withTable("t1") {
-      sql("CREATE TABLE t1 (id INT, p1 INT, p2 INT) USING csv OPTIONS ('csv.field-delimiter' ';')")
+      sql(
+        "CREATE TABLE t1 (id INT, p1 INT, p2 INT) USING csv OPTIONS ('csv.field-delimiter' ';') TBLPROPERTIES ('format-table.implementation'='engine')")
```
JingsongLi: Why change to engine? You can change compression to let test passing.

PR #6483, opened 2025-10-28
[core] Support row id push down

leaves12138:
### Purpose

This is an important pull request for global index building.

The global index (secondary index) is built on row id (primary index) system.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/io/DataFileMeta.java
+++ b/paimon-core/src/main/java/org/apache/paimon/io/DataFileMeta.java
@@ -327,25 +327,12 @@ public interface DataFileMeta {
 
     DataFileMeta copy(byte[] newEmbeddedIndex);
 
+    RoaringBitmap32 toFileSelection(List<Long> indices);
+
     static long getMaxSequenceNumber(List<DataFileMeta> fileMetas) {
         return fileMetas.stream()
                 .map(DataFileMeta::maxSequenceNumber)
                 .max(Long::compare)
                 .orElse(-1L);
-    }
-
-    static RoaringBitmap32 readIndices(List<Long> indices, DataFileMeta file) {
-        RoaringBitmap32 selection = null;
-        if (indices != null && file.firstRowId() != null) {
-            selection = new RoaringBitmap32();
-            long start = file.firstRowId();
-            long end = start + file.rowCount();
-            for (long rowId : indices) {
-                if (rowId >= start && rowId < end) {
-                    selection.add((int) (rowId - start));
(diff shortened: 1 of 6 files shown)


paimon-core/src/main/java/org/apache/paimon/table/source/ReadBuilder.java
```diff
@@ -149,6 +149,8 @@ default ReadBuilder withFilter(List<Predicate> predicates) {
      */
     ReadBuilder withShard(int indexOfThisSubtask, int numberOfParallelSubtasks);
 
+    ReadBuilder withRowIds(List<Long> indices);
```
JingsongLi, line 158: Add some comments.

paimon-common/src/main/java/org/apache/paimon/reader/FileRecordIterator.java
```diff
@@ -107,4 +109,47 @@ public void releaseBatch() {
             }
         };
     }
+
+    default FileRecordIterator<T> selection(RoaringBitmap32 selection) {
```
JingsongLi, line 113: Add a ut for this.

paimon-core/src/main/java/org/apache/paimon/io/DataFileMeta.java
```diff
@@ -332,4 +333,19 @@ static long getMaxSequenceNumber(List<DataFileMeta> fileMetas) {
                 .max(Long::compare)
                 .orElse(-1L);
     }
+
+    static RoaringBitmap32 readIndices(List<Long> indices, DataFileMeta file) {
```
JingsongLi: add a non-static method to `DataFileMeta`: `RoaringBitmap32 toFileSelection(List<Long> indices)`.

Should throw exception if `firstRowId` is null.

paimon-core/src/main/java/org/apache/paimon/table/source/ReadBuilder.java
leaves12138: OK

paimon-common/src/main/java/org/apache/paimon/reader/FileRecordIterator.java
leaves12138: Done

paimon-core/src/main/java/org/apache/paimon/io/DataFileMeta.java
leaves12138: Sure~

PR #6489, opened 2025-10-29
[core] Fix data-evolution row count should be more accurate.

leaves12138:
### Purpose

Data evolution table total records in snapshot should be more accurate


paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
```diff
@@ -1120,8 +1123,16 @@ CommitResult tryCommitOnce(
...
-            long deltaRecordCount = recordCountAdd(deltaFiles) - recordCountDelete(deltaFiles);
-            long totalRecordCount = previousTotalRecordCount + deltaRecordCount;
+            long deltaRecordCount;
+            long totalRecordCount;
+
+            if (dataEvolutionEnabled) {
+                deltaRecordCount = nextRowIdStart - firstRowIdStart;
+                totalRecordCount = nextRowIdStart;
+            } else {
+                deltaRecordCount = recordCountAdd(deltaFiles) - recordCountDelete(deltaFiles);
```
JingsongLi, line 1133: I think we don't need to modify here, we should also modify split row count just like python.

leaves12138: Maybe we can modify here too, it depends on how it defines

leaves12138:
Close this for now, maybe open it in the future

PR #6497, opened 2025-10-30
[core]format table: refactor overwrite

jerry-024:
### Purpose

Introduced FormatTableCommit to support commit/abort and handle overwrite.

### Tests
- org.apache.paimon.catalog.CatalogTestBase#testFormatTableOverwrite
- org.apache.paimon.flink.source.FormatTableITCase#testPartitionedTableInsertOverwrite
- org.apache.paimon.flink.source.FormatTableITCase#testUnPartitionedTableInsertOverwrite


JingsongLi (approved):
+1

PR #6498, opened 2025-10-30
[core] Introduce transform predicate for complex predicate

JingsongLi:
### Purpose

For predicate just like `concat(year, month, day) = 20251010`, we should support this predicate pushdown to Paimon core. Especially important for partition conditions.


Zouxxyy (approved):
+1

PR #6500, opened 2025-10-31
[core] format table: support file compression and custom default for diff format

jerry-024:
### Purpose

format table: support file compression and custom default for diff format
- parque： snappy
- orc: zstd
- csv and json: none

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/rest/RESTCatalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/rest/RESTCatalog.java
@@ -80,7 +80,6 @@ import java.util.Set;
 import java.util.stream.Collectors;
 
 import static org.apache.paimon.CoreOptions.BRANCH;
-import static org.apache.paimon.CoreOptions.FORMAT_TABLE_FILE_COMPRESSION;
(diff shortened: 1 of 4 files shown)


paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -2294,6 +2301,15 @@ public String fileCompression() {
         return options.get(FILE_COMPRESSION);
     }
 
+    public String formatTableFileImplementation() {
```
JingsongLi: merge `getFormatTableFileCompression`.

paimon-core/src/main/java/org/apache/paimon/rest/RESTCatalog.java
```diff
@@ -448,6 +449,11 @@ public void createTable(Identifier identifier, Schema schema, boolean ignoreIfEx
             checkNotSystemTable(identifier, "createTable");
             validateCreateTable(schema);
             createExternalTablePathIfNotExist(schema);
+            if (Options.fromMap(schema.options()).get(TYPE) == TableType.FORMAT_TABLE) {
```
JingsongLi: Revert this

PR #6506, opened 2025-11-01
[core][spark] Support push down transform predicate

Zouxxyy:
### Purpose

- Add spark push down transform predicate support (spark version >= 3.3)
- Enable concat pushdown (spark version >= 3.4)

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/predicate/ConcatTransform.java
+++ b/paimon-common/src/main/java/org/apache/paimon/predicate/ConcatTransform.java
@@ -35,4 +35,9 @@ public class ConcatTransform extends StringTransform {
     public BinaryString transform(List<BinaryString> inputs) {
         return BinaryString.concat(inputs);
     }
+
+    @Override
+    public Transform withNewInputs(List<Object> inputs) {
+        return new ConcatTransform(inputs);
+    }
 }
--- a/paimon-common/src/main/java/org/apache/paimon/predicate/FieldTransform.java
+++ b/paimon-common/src/main/java/org/apache/paimon/predicate/FieldTransform.java
@@ -1,0 +1,77 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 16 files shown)


paimon-common/src/main/java/org/apache/paimon/predicate/Transform.java
Zouxxyy, line 30: I believe we can introduce a `Literal` object in the future, so that our inputs will only be of either `Literal` or `FieldRef` type now.

paimon-common/src/main/java/org/apache/paimon/predicate/LeafPredicate.java
```diff
@@ -30,76 +30,73 @@
...
+import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 
 import static org.apache.paimon.utils.InternalRowUtils.get;
 
 /** Leaf node of a {@link Predicate} tree. Compares a field in the row with literals. */
-public class LeafPredicate implements Predicate {
+public class LeafPredicate extends TransformPredicate {
```
Zouxxyy, line 41: Since LeafPredicate extends TransformPredicate, and we have
```
T visit(LeafPredicate predicate);
T visit(TransformPredicate predicate);
```
Maybe we can combine them in the future

paimon-common/src/main/java/org/apache/paimon/predicate/ConcatTransform.java
```diff
@@ -35,4 +35,9 @@ public ConcatTransform(List<Object> inputs) {
     public BinaryString transform(List<BinaryString> inputs) {
         return BinaryString.concat(inputs);
     }
+
+    @Override
+    public Transform withNewInputs(List<Object> inputs) {
```
JingsongLi, line 40: copy

paimon-common/src/main/java/org/apache/paimon/predicate/FieldTransform.java
```diff
@@ -0,0 +1,77 @@
...
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+
+import static org.apache.paimon.utils.InternalRowUtils.get;
+
+/** Transform that extracts a field from a row. */
+public class FieldTransform implements Transform {
+
```
JingsongLi, line 32: Add ser id.

paimon-common/src/main/java/org/apache/paimon/predicate/LeafPredicate.java
```diff
@@ -30,76 +30,70 @@
...
 
     private static final long serialVersionUID = 1L;
```
JingsongLi, line 43: change ser id.

JingsongLi:
+1

PR #6507, opened 2025-11-01
[doc] Update spark procedure doc

Zouxxyy:
### Purpose

- sink.parallelism is flink options, change it
- The procedure doc table should have only three columns; delete the extra columns


JingsongLi:
+1

PR #6508, opened 2025-11-02
[csv] Csv Parser should not depends on jackson

JingsongLi:
### Purpose

For some illegal csv lines, jackson will throw exceptions. We should accept all lines for Hadoop ecosystem.


jerry-024:
+1

PR #6510, opened 2025-11-03
[core] format table: support write file in _temporary at first

jerry-024:
### Purpose

Format table: support write file in _temporary at first

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/fs/RenamingTwoPhaseOutputStream.java
+++ b/paimon-common/src/main/java/org/apache/paimon/fs/RenamingTwoPhaseOutputStream.java
@@ -29,7 +29,7 @@ import java.util.UUID;
  */
 @Public
 public class RenamingTwoPhaseOutputStream extends TwoPhaseOutputStream {
-    public static final String TEMP_DIR_NAME = "_temporary";
+    private static final String TEMP_DIR_NAME = "_temporary";
 
     private final Path targetPath;
     private final Path tempPath;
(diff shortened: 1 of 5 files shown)


paimon-core/src/main/java/org/apache/paimon/table/format/FormatTableCommit.java
```diff
@@ -100,6 +102,21 @@ public void commit(List<CommitMessage> commitMessages) {
             for (TwoPhaseOutputStream.Committer committer : committers) {
                 committer.commit(this.fileIO);
             }
+            if (committers.stream()
+                    .filter(c -> c instanceof RenamingTwoPhaseOutputStream.TempFileCommitter)
+                    .findAny()
+                    .isPresent()) {
+                if (partitionPaths.size() > 1) {
```
JingsongLi: You should introduce a clean method to `TwoPhaseOutputStream`.

paimon-core/src/main/java/org/apache/paimon/table/format/FormatTableCommit.java
```diff
@@ -80,16 +80,17 @@ public void commit(List<CommitMessage> commitMessages) {
                                     + commitMessage.getClass().getName());
                 }
             }
+            Set<Path> partitionPaths = new HashSet<>();
```
JingsongLi: Why extract this field?

paimon-common/src/main/java/org/apache/paimon/fs/RenamingTwoPhaseOutputStream.java
```diff
@@ -132,5 +133,13 @@ public void discard(FileIO fileIO) throws IOException {
         public Path targetFilePath() {
             return targetPath;
         }
+
+        @Override
+        public void clean(FileIO fileIO) throws IOException {
+            Path path = tempPath.getParent();
+            if (fileIO.exists(path)) {
+                fileIO.deleteQuietly(path);
```
JingsongLi, line 141: Maybe you should use `deleteDirectoryQuietly`.

PR #6512, opened 2025-11-03
[improvement] When dropping table, data should be moved to trash instead of being deleted directly

### LINKED ISSUE #6511: [Improvement] When dropping table, data should be moved to trash instead of being deleted directly
### Search before asking

### Motivation

In Paimon, when dropping table, the data is directly removed. We can implement a feature where data is first moved to trash, making it easier to recover accidentally deleted data.

### Solution

Provides a method called `moveToTrash` that is called when data is cleaned up to move data to the trash.

### Anything else?

_No response_

tomscut:
### Purpose

Linked issue: close #6511

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/fs/FileIO.java
+++ b/paimon-common/src/main/java/org/apache/paimon/fs/FileIO.java
@@ -227,6 +227,13 @@ public interface FileIO extends Serializable, Closeable {
     boolean delete(Path path, boolean recursive) throws IOException;
 
     /**
+     * Move the path to trash.
+     *
+     * @param path the path to move to trash.
+     */
+    boolean moveToTrash(Path path) throws IOException;
+
+    /**
      * Make the given file and all non-existent parents into directories. Has the semantics of Unix
      * 'mkdir -p'. Existence of the directory hierarchy is not an error.
      *
--- a/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
+++ b/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
@@ -961,7 +961,7 @@ public class HiveCatalog extends AbstractCatalog {
                                     identifier.getTableName(),
                                     !externalTable,
                                     false,
-                                    true));
+                                    false));
 
             // When drop a Hive external table, only the hive metadata is deleted and the data files
             // are not deleted.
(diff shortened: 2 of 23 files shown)


paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
```diff
@@ -961,7 +961,7 @@ protected void dropTableImpl(Identifier identifier, List<Path> externalPaths) {
                                     identifier.getTableName(),
                                     !externalTable,
                                     false,
-                                    true));
+                                    false));
```
cxzl25, line 964: This change is quite critical and prevents some physical deletions.

Iceberg does not have ifPurge when dropping table.

https://github.com/apache/iceberg/blob/5dae5fc856e228cd8e1b69ba14909d0067006650/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java#L247-L251

paimon-common/src/main/java/org/apache/paimon/fs/FileIO.java
```diff
@@ -226,6 +226,13 @@ default FileStatus[] listDirectories(Path path) throws IOException {
      */
     boolean delete(Path path, boolean recursive) throws IOException;
 
+    /**
+     * Move the path to trash.
+     *
+     * @param path the path to move to trash.
+     */
+    boolean moveToTrash(Path path) throws IOException;
```
cxzl25, line 234:
```suggestion
    default boolean moveToTrash(Path path) throws IOException {
        throw new UnsupportedOperationException();
    }
```

JingsongLi (commented):
Moving table data to trash instead of direct deletion is a good safety mechanism.

**Review:**

1. **API design**: Adding a `moveToTrash(Path)` or similar method to `FileIO` is the right abstraction level. But the change touches many FileIO implementations (Hadoop, Local, Azure, COSN, REST, etc.) — each must implement the trash behavior.

2. **Questions:**
   - What's the trash path structure? Is it `{warehouse}/.trash/{table}/{timestamp}/`?
   - How is trash cleaned up? Is there a TTL or explicit purge operation?
   - What happens on object storage where "move" is copy+delete (expensive for large tables)?
   - Is there a way to restore from trash?

3. **Scope**: 141 additions across 10+ files. This is cross-cutting — ensure all FileIO implementations handle the trash case, including edge cases like:
   - Trash directory already exists
   - Insufficient permissions to create trash directory
   - Cross-filesystem moves (not supported → fallback to delete?)

4. **Configuration**: Should there be a `table.drop-to-trash=true/false` option so users can opt out?

5. **PR description is minimal** — only links the issue. Please document the design decisions (trash location, cleanup, restore).

Good direction but needs more design clarity.

PR #6513, opened 2025-11-03
[fix] [core] Non null column with write type

leaves12138:
### Purpose

Write with write type which contains non-null column


JingsongLi (approved):
+1

PR #6518, opened 2025-11-03
[core] [fix] Fix with blob rolling parquet files.

leaves12138:
### Purpose

Fix blob writing failed with parquet file rolling.


JingsongLi:
+1

PR #6519, opened 2025-11-03
[core] Remove hash lookup to simplify interface

JingsongLi:
### Purpose

Sort lookup is better than hash lookup always. We can remove hash lookup to simplify interface, like write interface without returning value.

Next, we can store lookup file in remote storage to reduce init cost.

### DIFF
--- a/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
@@ -3805,32 +3805,6 @@ public class CoreOptions implements Serializable {
         }
     }
 
-    /** Specifies the local file type for lookup. */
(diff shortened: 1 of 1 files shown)


paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -1121,12 +1121,6 @@ public InlineElement getDescription() {
                     .withDescription(
                             "Define partition by table options, cannot define partition on DDL and table options at the same time.");
 
-    public static final ConfigOption<LookupLocalFileType> LOOKUP_LOCAL_FILE_TYPE =
-            key("lookup.local-file-type")
-                    .enumType(LookupLocalFileType.class)
```
tsreaper, line 1126: Also remove this enum class from `CoreOptions`.

PR #6520, opened 2025-11-03
[Python] Keep the variable names of Identifier consistent with Java

universe-hcy:
### Purpose

Keep the variable names of Identifier consistent with Java


paimon-python/pypaimon/schema/data_types.py
discivigour, line 1: Why change this here?

discivigour:
+1

PR #6524, opened 2025-11-04
[core] skip validate main branch before orphan files cleaning

XiaoHongbo-Hope:
### Purpose
skip validate main branch before doOrphanClean, since getTable is invoked before validate branches.


jerry-024:
+1

PR #6525, opened 2025-11-04
[core] format table: fix two-phase commit in FileIO for which use multi-part upload

jerry-024:
### Purpose

Fix two-phase commit in FileIO for which use multi-part upload

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/fs/FileIO.java
+++ b/paimon-common/src/main/java/org/apache/paimon/fs/FileIO.java
@@ -114,21 +114,6 @@ public interface FileIO extends Serializable, Closeable {
     }
 
     /**
-     * Create a MultiPartUploadStore for multipart upload.
-     *
-     * @param path the file target path
-     * @return a MultiPartUploadStore
-     * @throws IOException Thrown, if the store could not be created because of an I/O.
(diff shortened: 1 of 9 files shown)


paimon-common/src/main/java/org/apache/paimon/fs/FileIO.java
```diff
@@ -113,6 +113,21 @@ default TwoPhaseOutputStream newTwoPhaseOutputStream(Path path, boolean overwrit
...
 
+    /**
+     * Create a MultiPartUploadStore for multipart upload.
+     *
+     * @param path the file target path
+     * @return a MultiPartUploadStore
+     * @throws IOException Thrown, if the store could not be created because of an I/O.
+     * @throws UnsupportedOperationException if the filesystem does not support multipart upload
+     */
+    default MultiPartUploadStore newMultiPartUploadStore(Path path) throws IOException {
```
JingsongLi: This method should not in here. Why need to invoke `FileIO`? `TwoPhaseOutputStream` is created by `FileIO`.

jerry-024: When committing, we need to use FileIO to get MultiPartUploadStore.

PR #6526, opened 2025-11-04
[Python] Suppport multi prepare commit in the same TableWrite 

discivigour:
### Purpose

- Suppport multi prepare commit in the same TableWrite.

### Tests

- TableWriteTest

### DIFF
--- a/paimon-python/pypaimon/table/file_store_table.py
+++ b/paimon-python/pypaimon/table/file_store_table.py
@@ -28,7 +28,7 @@ from pypaimon.schema.schema_manager import SchemaManager
 from pypaimon.schema.table_schema import TableSchema
 from pypaimon.table.bucket_mode import BucketMode
 from pypaimon.table.table import Table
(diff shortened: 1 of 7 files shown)


paimon-python/pypaimon/write/batch_table_commit.py
```diff
@@ -41,15 +41,13 @@ def __init__(self, table, commit_user: str, static_partition: Optional[dict]):
         self.file_store_commit = FileStoreCommit(snapshot_commit, table, commit_user)
         self.batch_committed = False
 
-    def commit(self, commit_messages: List[CommitMessage]):
+    def commit(self, commit_messages: List[CommitMessage], commit_identifier: int = COMMIT_IDENTIFIER):
```
JingsongLi: commit_identifier must be Long.MAX, do not provide parameter.

JingsongLi:
It is better to introduce a new stream write just like Java.

PR #6527, opened 2025-11-04
[python] support custom source split target size and split open file cost

XiaoHongbo-Hope:
### Purpose
Support `source.split.target-size` and `source.split.open-file-cost`table options for  split generation in pypaimon instead of using hard code value.

### Tests
- `reader_base_test.ReaderBasicTest.test_split_target_size`
- `reader_base_test.ReaderBasicTest.test_split_open_file_cost`


XiaoHongbo-Hope:
@chenghuichen @discivigour Could you please help take a look at this PR.

JingsongLi:
+1

PR #6528, opened 2025-11-04
[core] Add exception for create partition

Zouxxyy:
### Purpose

User may be confused, when perform `alter table add partition` without any exceptions,

but run `show partitions` return nothing

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
@@ -548,7 +548,10 @@ public abstract class AbstractCatalog implements Catalog {
 
     @Override
     public void createPartitions(Identifier identifier, List<Map<String, String>> partitions)
-            throws TableNotExistException {}
+            throws TableNotExistException {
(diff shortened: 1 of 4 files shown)


JingsongLi (commented):
Good UX improvement. Silent success followed by empty `show partitions` is confusing.

**The problem**: `ALTER TABLE ADD PARTITION` succeeds without error but the partition doesn't appear because the table type doesn't support manual partition creation (e.g., non-partitioned tables or tables where partitions are created implicitly by data writes).

**Questions:**

1. What specific conditions trigger the exception? Is it:
   - Adding partitions to non-partitioned tables?
   - Adding partitions to tables with `metastore.partitioned-table=false`?
   - Something else?

2. What's the error message? Users need a clear explanation of *why* the operation is unsupported and what to do instead (e.g., "write data to create partitions").

3. The PR touches `AbstractCatalog`, `RESTCatalog`, and `HiveCatalog` — does each catalog type have different partition creation semantics that need different handling?

4. The Spark test suggests this is validated at the REST catalog layer too. Good.

Please clarify the exact conditions in the PR description.

PR #6533, opened 2025-11-05
[core] Introduce 'lookup.remote-file.enabled' to optimize lookup

JingsongLi:
### Purpose

For lookup table (changelog-producer=lookup, or deletion-vectors enabled, or first-row):

If the job restarts, it will convert almost all the required data files locally to a lookup file, which consumes a lot of energy and may also cause the first checkpoint to fail.

We can store the converted lookup file remotely to avoid conversion overhead during restart.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/RemoteLookupFileManager.java
+++ b/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/RemoteLookupFileManager.java
@@ -59,9 +59,14 @@ public class RemoteLookupFileManager<T> implements RemoteFileDownloader {
     }
 
     public DataFileMeta genRemoteLookupFile(DataFileMeta file) throws IOException {
-        LookupFile lookupFile = lookupLevels.createLookupFile(file);
         String remoteSstName = lookupLevels.remoteSstName(file.fileName());
+        if (file.extraFiles().contains(remoteSstName)) {
+            // ignore existed
+            return file;
+        }
+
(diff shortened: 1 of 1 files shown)


paimon-core/src/main/java/org/apache/paimon/mergetree/compact/MergeTreeCompactRewriter.java
```diff
@@ -104,7 +104,9 @@ protected CompactResult rewriteCompaction(
 
         List<DataFileMeta> before = extractFilesFromSections(sections);
         notifyRewriteCompactBefore(before);
-        return new CompactResult(before, writer.result());
+        List<DataFileMeta> after = writer.result();
+        after = notifyRewriteCompactAfter(after);
+        return new CompactResult(before, after);
```
tsreaper, line 109: 1. For files in `before`, you also need to delete their lookup files. Add tests on snapshot expire to check this.
2. No need to create another lookup file, if the original file is upgraded, but not compacted.

JingsongLi: > For files in before, you also need to delete their lookup files. Add tests on snapshot expire to check this.

Extra files have been used in file index, it will be deleted by snapshot expiration.

> No need to create another lookup file, if the original file is upgraded, but not compacted.

We should avoid creating same lookup file twice, but it is hard to create test for upgrading.

tsreaper (approved):

PR #6541, opened 2025-11-06
[core] File size should be extract from path if it is zero

leaves12138:
### Purpose

Linked issue: close #xxx


paimon-core/src/main/java/org/apache/paimon/io/RowDataFileWriter.java
```diff
@@ -98,6 +98,9 @@ public void close() throws IOException {
     @Override
     public DataFileMeta result() throws IOException {
         long fileSize = outputBytes;
```
JingsongLi:
```
@Nullable
Long outputBytes
```

paimon-core/src/main/java/org/apache/paimon/io/SingleFileWriter.java
```diff
@@ -53,7 +55,7 @@ public abstract class SingleFileWriter<T, R> implements FileWriter<T, R> {
     private FormatWriter writer;
     private PositionOutputStream out;
 
-    protected long outputBytes;
+    @Nullable protected Long outputBytes;
```
JingsongLi: private

leaves12138: OK

JingsongLi (commented):
Please add test for this, for example in `LanceFormatTest`.

JingsongLi:
+1

PR #6543, opened 2025-11-06
[core]update SingleFileWriter.outputBytes during close for writers with SupportsDirectWrite

ColdL:
### Purpose

I found a suspected bug in `SingleFileWriter`.

If it is a writer that implements `SupportsDirectWrite` (currently only `LanceWriter`), then `outputBytes` will not be updated, ultimately causing the `file_size` in the manifest entry to always be `zero`.

I discovered this while testing PK Table & LanceFile & deletion-vectors. I found that no matter how I executed `DELETE`, the dv index was never generated as expected. Subsequently, I found that the `$files` system table showed `file_size_in_bytes` as `zero`, indicating that the `file_size` in the manifest is zero, but the actual data file ending with .lance is 1.5MB in size on the storage side.

After applying the fix in the PR, I found that the `$files` system table could display the actual file size normally, and the dv index could also be generated as expected.

### Tests

Not yet

### API and Format

No

### Documentation

No

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/io/SingleFileWriter.java
+++ b/paimon-core/src/main/java/org/apache/paimon/io/SingleFileWriter.java
@@ -32,6 +32,7 @@ import org.apache.paimon.utils.IOUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.FileNotFoundException;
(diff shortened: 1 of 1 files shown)


JingsongLi:
Already fixed in https://github.com/apache/paimon/pull/6541

PR #6549, opened 2025-11-06
[Python] Remove the use of reference types in DATA_FILE_META_SCHEMA

zjw1111:
### Purpose

Writing manifest schema using Avro's reference types will cause failure when reading with the native reader.

```python
DATA_FILE_META_SCHEMA = {
    "type": "record",
    "name": "DataFileMeta",
    "fields": [
        {"name": "_KEY_STATS", "type": SIMPLE_STATS_SCHEMA},
        {"name": "_VALUE_STATS", "type": "SimpleStats"},  # reference type
    ]
}
```

The purpose of this PR is to change it to use origin types.


JingsongLi:
Please add test for this.

JingsongLi:
+1

PR #6550, opened 2025-11-06
[core] Fix row id push down: blob file may not continuous caused by row id filter

leaves12138:
### Purpose

Fix test when blob file is not continuous


JingsongLi:
+1

PR #6552, opened 2025-11-07
[core] format table: support init jindo by hadoop conf

jerry-024:
### Purpose

[core] format table: support init jindo by hadoop conf


paimon-filesystems/paimon-jindo/src/main/java/org/apache/paimon/jindo/JindoLoader.java
```diff
@@ -30,11 +30,7 @@ public class JindoLoader implements FileIOLoader {
 
     @Override
     public List<String[]> requiredOptions() {
-        List<String[]> options = new ArrayList<>();
-        options.add(new String[] {"fs.oss.endpoint"});
-        options.add(new String[] {"fs.oss.accessKeyId"});
-        options.add(new String[] {"fs.oss.accessKeySecret"});
-        return options;
+        return new ArrayList<>();
```
Zouxxyy, line 33: Is there a corresponding fallback mechanism? If the configuration does not meet the requirements?

jerry-024: If the scheme is oss which condition we need fallback to hadoop file io?

PR #6559, opened 2025-11-07
[core] support incremental clustering in dv mode

LsomeYeah:
### Purpose

Linked issue: close #xxx

Deletion vector is also supported for append table. This pr will support incremental clustering for append table in dv mode.


JingsongLi (approved):
+1

PR #6566, opened 2025-11-08
[core] Add distinct parameter support to FieldListaggAgg operator

759434091:
### Purpose

Add distinct support to FileListagg aggregate function.

- Reuses `fields.{fieldName}.distinct` option
- May significantly reduce computation overhead vs `array.collect.distinct`, `map.merge_map`

### Tests

org.apache.paimon.mergetree.compact.aggregate.FieldAggregatorTest#testFieldListAggWithDefaultDelimiterAndDistinct

### API and Format

- `fields.{fieldName}.distinct` option


JingsongLi (approved):
+1

PR #6578, opened 2025-11-10
[python] support rolling for blob write when blob-as-descriptor

XiaoHongbo-Hope:
[python] Support blob.target-file-size (target-file-size) rolling for blob tables when blob-as-descriptor

## Purpose

Currently, Python Paimon blob tables did not support file rolling based on `target-file-size` in `blob-as-descriptor=true` mode. 

## This PR

- **Supports `target-file-size` rolling in `blob-as-descriptor=true` mode**: Implements row-by-row writing with actual file size checking, aligned with Java's `RollingBlobFileWriter` architecture
- **Fixes `blob.target-file-size` configuration**: Ensures `blob.target-file-size` is respected in both descriptor and non-descriptor modes instead of being ignored

## Tests

- `blob_table_test.py`
-  `RollingBlobFileWriterTest`, used to check blob rolling in java

### DIFF
--- a/paimon-python/pypaimon/write/writer/blob_writer.py
+++ b/paimon-python/pypaimon/write/writer/blob_writer.py
@@ -172,17 +172,11 @@ class BlobWriter(AppendOnlyDataWriter):
         else:
             # row_count only (from BlobFileWriter)
             row_count = data_or_row_count
-            # For blob files, we don't have stats
-            # Use the actual blob column name instead of hardcoded 'blob_data'
-            # This ensures the stats schema matches the actual table schema
             data_fields = [PyarrowFieldParser.to_paimon_schema(pa.schema([(self.blob_column, pa.large_binary())]))[0]]
             min_value_stats = [None]
             max_value_stats = [None]
             value_null_counts = [0]
 
-        # min_seq = seqNumCounter.getValue() - super.recordCount()
-        # max_seq = seqNumCounter.getValue() - 1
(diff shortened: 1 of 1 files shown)


paimon-python/pypaimon/common/core_options.py
```diff
@@ -76,3 +78,19 @@ def get_split_open_file_cost(options: dict) -> int:
...
+            return MemorySize.parse(size_str).get_bytes()
+        return MemorySize.of_mebi_bytes(128 if has_primary_key else 256).get_bytes()
+
+    @staticmethod
+    def get_blob_target_file_size(options: dict) -> int:
+        """Get blob target file size from options, default to target-file-size (256MB for append-only table)."""
+        if CoreOptions.BLOB_TARGET_FILE_SIZE in options:
+            size_str = options[CoreOptions.BLOB_TARGET_FILE_SIZE]
+            return MemorySize.parse(size_str).get_bytes()
+        return CoreOptions.get_target_file_size(options, has_primary_key=False)
```
jerry-024, line 96: Is `has_primary_key` always false here?

XiaoHongbo-Hope: > Is `has_primary_key` always false here?

Should be yes, found schema check in schema.py. And following java conf cc @leaves12138 
image
image

leaves12138 (approved):
+1

PR #6579, opened 2025-11-10
[Python] Add Mixed read and write test between Java and Python.

discivigour:
### Purpose

Add Mixed read and write test between Java and Python:
- Java write and python read.
- Python write and Java read.

### Tests

- run_mixed_tests.sh
- JavaPyReadWriteTest
- JavaPyE2ETest


JingsongLi (approved):
+1

PR #6596, opened 2025-11-13
[flink] Don't reopen the dim table if the scanned partitions are not overwritten

yuzelin:
### Purpose

For example, lookup with scan MAX_PT(), and the current MAX_PT is 20251111. If overwrite pt 20241110, we don't need to reopen the dim table.

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/FullCacheLookupTable.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/FullCacheLookupTable.java
@@ -55,7 +55,6 @@ import javax.annotation.Nullable;
 import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/LookupDataTableScan.java
```diff
@@ -49,29 +53,28 @@ public class LookupDataTableScan extends DataTableStreamScan {
 
     private static final Logger LOG = LoggerFactory.getLogger(LookupDataTableScan.class);
 
+    private final FileStoreTable table;
     private final StartupMode startupMode;
     private final LookupStreamScanMode lookupScanMode;
 
+    private List<BinaryRow> scanPartitions = Collections.emptyList();
```
JingsongLi: default null?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/LookupDataTableScan.java
```diff
@@ -81,15 +84,45 @@ public LookupDataTableScan(
...
+        if (shouldReopen(snapshot)) {
+            LOG.info("Dim table found OVERWRITE snapshot {}, reopen.", snapshot.id());
+            throw new ReopenException();
+        } else {
+            return null;
+        }
+    }
+
+    private boolean shouldReopen(Snapshot snapshot) {
+        if (scanPartitions.isEmpty()) {
```
JingsongLi: if scan partitions is null

JingsongLi (approved):
+1

PR #6597, opened 2025-11-13
Also check row type if schema id not match in RemoteLookupFileManager

tsreaper:
### Purpose

This PR also reuse remote lookup file when schema id does not match, but row type matches. This can avoid invalidating remote lookup files when the user only changes some table options.

### Tests

<TODO>

### API and Format

No format changes.

### Documentation

No new feature.


JingsongLi:
+1

PR #6598, opened 2025-11-13
[core] rest catalog: support default create table options

jerry-024:
### Purpose

rest catalog: support default create table options

### API and Format

- MockRESTCatalogTest#testCreateTableDefaultOptions


JingsongLi (approved):
+1

PR #6604, opened 2025-11-14
[python] use str to replace pathlib.Path to avoid scheme missing in pypaimon 

XiaoHongbo-Hope:
## Purpose

Fixes URI scheme loss issue in blob-as-descriptor mode by replacing `pathlib.Path` with `str` throughout the codebase. This ensures URI schemes (e.g., `oss://`, `s3://`, `file://`) are preserved when handling blob descriptors and file paths.

## Key Changes

**Replaced `pathlib.Path` with `str`**: Unified path handling to preserve URI schemes across all filesystem operations.

## Impact

- **I/O & Catalog**: `FileIO`, `FileSystemCatalog`, `RESTCatalog` - all path operations use `str`
- **Manifest & Metadata**: `DataFileMeta`, `SchemaManager`, manifest managers - -  path storage use `str`
- **Read/Write Operations**: All readers and writers (`FormatBlobReader`, `DataWriter`, `BlobWriter`, etc.) - consistent `str` path handling
- **Table Management**: `FileStoreTable`, `SnapshotManager` --table path use `str`
- **Blob Support**: `Blob`, `BlobDescriptor` - URI scheme preservation for external blob storage

## Tests

- `test_blob_write_read_end_to_end_with_descriptor`: Verifies URI scheme preservation in blob descriptors
- Sample script `oss_blob_as_descriptor.py`: Demonstrates OSS blob-as-descriptor functionality with scheme preservation
- `test_file_io`
- `rest_token_file_io`

### DIFF
--- a/paimon-python/pypaimon/sample/rest_catalog_read_write_sample.py
+++ b/paimon-python/pypaimon/sample/rest_catalog_read_write_sample.py
@@ -106,4 +106,3 @@ def main():
 
 if __name__ == '__main__':
     main()
-


paimon-python/pypaimon/catalog/filesystem_catalog.py
```diff
@@ -108,18 +106,13 @@ def get_table_schema(self, identifier: Identifier):
             raise TableNotExistException(identifier)
         return table_schema
 
-    def get_database_path(self, name) -> Path:
-        return self._trim_schema(self.warehouse) / f"{name}{Catalog.DB_SUFFIX}"
+    def get_database_path(self, name) -> str:
+        warehouse = self.warehouse.rstrip('/')
+        return f"{warehouse}/{name}{Catalog.DB_SUFFIX}"
```
lucasfang, line 111: how about use os.path.join to construct the path? then you don’t need to use rstrip to remove trailing slashes from the warehouse path

XiaoHongbo-Hope: > > how about use os.path.join to construct the path? then you don’t need to use rstrip to remove trailing slashes from the warehouse path
> 
> Thanks for your kind suggestion. I tried use os.path but found that `os.path.join(self.warehouse, "db.db")` produces `oss://bucket/path\db.db` on Windows instead of `oss://bucket/path/db.db`. So I choose str concat finally

JingsongLi:
+1

PR #6605, opened 2025-11-14
[flink] support rowId push down for RowTrackingTable

steFaiz:
### Purpose
This PR is about to support rowid pushdown for flink, following https://github.com/apache/paimon/pull/6483

Linked issue: None

### Tests

Please see the test cases in org.apache.paimon.flink.source.FlinkTableSourceTest

### API and Format

This PR does not change any user API


paimon-common/src/main/java/org/apache/paimon/predicate/RowIdPredicateVisitor.java
```diff
@@ -0,0 +1,104 @@
...
+ * </ul>
+ */
+public class RowIdPredicateVisitor implements PredicateVisitor<Set<Long>> {
+
+    @Override
+    public Set<Long> visit(LeafPredicate predicate) {
+        if (ROW_ID.name().equals(predicate.fieldName())) {
+            LeafFunction function = predicate.function();
+            // _ROW_ID is always not null, so we can push down IsNull as an empty set.
+            if (function instanceof Equal || function instanceof In || function instanceof IsNull) {
```
JingsongLi: Remove `IS_NULL`? Is null optimization is for all non-null columns, not just for row id.

steFaiz: @JingsongLi  Thanks for your review! I'm not very familiar about it, sorry for that. I've removed `IS_NULL` condition and added some ITCases. PTAL if you have some time!

PR #6606, opened 2025-11-14
[flink] Support cdc source

yunfengzhou-hub:
### Purpose

This PR adds support to use Paimon as a Flink CDC Pipeline Source.

### Tests

Tests are added to the `org.apache.paimon.flink.pipeline.cdc.source` package of `paimon-flink-cdc` module to verify its correctness

### API and Format

It changes the API by adding a new Flink source implementation.

### Documentation

It introduces new features that are documented in docs/content/cdc-ingestion/flink-cdc-paimon-pipeline-connector.md.

### DIFF
--- a/paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/pipeline/cdc/schema/CDCMetadataAccessor.java
+++ b/paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/pipeline/cdc/schema/CDCMetadataAccessor.java
@@ -78,6 +78,24 @@ public class CDCMetadataAccessor implements MetadataAccessor {
         for (String databaseName : databaseNames) {
             try {
                 for (String tableName : catalog.listTables(databaseName)) {
+                    Identifier identifier = Identifier.create(databaseName, tableName);
+                    try {
+                        Table table = catalog.getTable(identifier);
+                        if (!(table instanceof FileStoreTable)) {
+                            LOG.info(
+                                    "listTables found table {}, but it is a {} instead of a FileStoreTable. Skipping this table.",
+                                    identifier,
+                                    table.getClass().getSimpleName());
+                            continue;
+                        }
+                    } catch (Catalog.TableNotExistException e) {
+                        LOG.warn(
+                                "Table {} does not exist. Perhaps it is dropped in the middle of this method. Skipping this table.",
+                                identifier,
+                                e);
+                        continue;
+                    }
+
                     tableIds.add(TableId.tableId(databaseName, tableName));
                 }
             } catch (Catalog.DatabaseNotExistException e) {
--- a/paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/pipeline/cdc/source/TableAwareFileStoreSourceSplit.java
+++ b/paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/pipeline/cdc/source/TableAwareFileStoreSourceSplit.java
@@ -20,7 +20,6 @@ package org.apache.paimon.flink.pipeline.cdc.source;
 
 import org.apache.paimon.catalog.Identifier;
 import org.apache.paimon.flink.source.FileStoreSourceSplit;
-import org.apache.paimon.schema.TableSchema;
 import org.apache.paimon.table.source.Split;
 import org.apache.paimon.utils.InstantiationUtil;
 import org.apache.paimon.utils.JsonSerdeUtil;
@@ -42,32 +41,32 @@ import java.util.Objects;
  */
 public class TableAwareFileStoreSourceSplit extends FileStoreSourceSplit {
     private final Identifier identifier;
-    @Nullable private final TableSchema lastSchema;
-    private final TableSchema schema;
+    private final @Nullable Long lastSchemaId;
+    private final long schemaId;
 
     public TableAwareFileStoreSourceSplit(
             String id,
             Split split,
             long recordsToSkip,
             Identifier identifier,
-            @Nullable TableSchema lastSchema,
-            TableSchema schema) {
+            @Nullable Long lastSchemaId,
+            long schemaId) {
         super(id, split, recordsToSkip);
         this.identifier = identifier;
-        this.lastSchema = lastSchema;
-        this.schema = schema;
+        this.lastSchemaId = lastSchemaId;
+        this.schemaId = schemaId;
     }
 
     public Identifier getIdentifier() {
         return identifier;
     }
 
-    public @Nullable TableSchema getLastSchema() {
-        return lastSchema;
+    public @Nullable Long getLastSchemaId() {
+        return lastSchemaId;
     }
 
-    public TableSchema getSchema() {
-        return schema;
+    public long getSchemaId() {
+        return schemaId;
     }
 
     @Override
@@ -81,13 +80,14 @@ public class TableAwareFileStoreSourceSplit extends FileStoreSourceSplit {
                 && Objects.equals(split(), other.split())
                 && recordsToSkip() == other.recordsToSkip()
                 && identifier.equals(other.identifier)
-                && Objects.equals(lastSchema, other.lastSchema)
-                && schema.equals(other.schema);
+                && Objects.equals(lastSchemaId, other.lastSchemaId)
+                && schemaId == other.schemaId;
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(splitId(), split(), recordsToSkip(), identifier, lastSchema, schema);
+        return Objects.hash(
+                splitId(), split(), recordsToSkip(), identifier, lastSchemaId, schemaId);
     }
 
     @Override
@@ -102,16 +102,17 @@ public class TableAwareFileStoreSourceSplit extends FileStoreSourceSplit {
                 + recordsToSkip()
                 + ", identifier="
                 + identifier
-                + ", lastSchema="
-                + lastSchema
-                + ", schema="
-                + schema
+                + ", lastSchemaId="
+                + lastSchemaId
+                + ", schemaId="
+                + schemaId
                 + '}';
     }
 
     /** The serializer for {@link TableAwareFileStoreSourceSplit}. */
     public static class Serializer
             implements SimpleVersionedSerializer<TableAwareFileStoreSourceSplit> {
+        private static final Long NULL_SCHEMA_ID = -1L;
 
         @Override
         public int getVersion() {
@@ -126,8 +127,9 @@ public class TableAwareFileStoreSourceSplit extends FileStoreSourceSplit {
             InstantiationUtil.serializeObject(view, split.split());
             view.writeLong(split.recordsToSkip());
             view.writeUTF(JsonSerdeUtil.toJson(split.getIdentifier()));
-            view.writeUTF(JsonSerdeUtil.toJson(split.getLastSchema()));
-            view.writeUTF(JsonSerdeUtil.toJson(split.getSchema()));
+            view.writeLong(
+                    split.getLastSchemaId() == null ? NULL_SCHEMA_ID : split.getLastSchemaId());
+            view.writeLong(split.getSchemaId());
             return out.toByteArray();
         }
 
@@ -145,10 +147,13 @@ public class TableAwareFileStoreSourceSplit extends FileStoreSourceSplit {
             }
             long recordsToSkip = view.readLong();
             Identifier identifier = JsonSerdeUtil.fromJson(view.readUTF(), Identifier.class);
-            TableSchema lastSchema = TableSchema.fromJson(view.readUTF());
-            TableSchema schema = TableSchema.fromJson(view.readUTF());
+            Long lastSchemaId = view.readLong();
+            if (lastSchemaId.equals(NULL_SCHEMA_ID)) {
+                lastSchemaId = null;
+            }
+            long schemaId = view.readLong();
             return new TableAwareFileStoreSourceSplit(
-                    splitId, split, recordsToSkip, identifier, lastSchema, schema);
+                    splitId, split, recordsToSkip, identifier, lastSchemaId, schemaId);
         }
     }
 }
--- a/paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/pipeline/cdc/source/enumerator/CDCSourceEnumerator.java
+++ b/paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/pipeline/cdc/source/enumerator/CDCSourceEnumerator.java
@@ -25,7 +25,6 @@ import org.apache.paimon.catalog.CatalogContext;
 import org.apache.paimon.catalog.Identifier;
 import org.apache.paimon.flink.metrics.FlinkMetricRegistry;
 import org.apache.paimon.flink.pipeline.cdc.source.TableAwareFileStoreSourceSplit;
-import org.apache.paimon.schema.TableSchema;
 import org.apache.paimon.table.FileStoreTable;
 import org.apache.paimon.table.Table;
 import org.apache.paimon.table.source.DataSplit;
@@ -272,16 +271,16 @@ public class CDCSourceEnumerator
         FileStoreTable table = tableStatusMap.get(tableAwarePlan.identifier).table;
         List<TableAwareFileStoreSourceSplit> splits = new ArrayList<>();
         for (Split split : plan.splits()) {
-            TableSchema lastSchema = tableStatusMap.get(tableAwarePlan.identifier).schema;
+            Long lastSchemaId = tableStatusMap.get(tableAwarePlan.identifier).schemaId;
             TableAwareFileStoreSourceSplit tableAwareFileStoreSourceSplit =
                     toTableAwareSplit(
                             splitIdGenerator.getNextId(),
                             split,
                             table,
                             tableAwarePlan.identifier,
-                            lastSchema);
-            tableStatusMap.get(tableAwarePlan.identifier).schema =
-                    tableAwareFileStoreSourceSplit.getSchema();
+                            lastSchemaId);
+            tableStatusMap.get(tableAwarePlan.identifier).schemaId =
+                    tableAwareFileStoreSourceSplit.getSchemaId();
             splits.add(tableAwareFileStoreSourceSplit);
         }
 
@@ -295,13 +294,12 @@ public class CDCSourceEnumerator
             Split split,
             FileStoreTable table,
             Identifier identifier,
-            @Nullable TableSchema lastSchema) {
+            @Nullable Long lastSchemaId) {
         Preconditions.checkState(split instanceof DataSplit);
         long snapshotId = ((DataSplit) split).snapshotId();
         long schemaId = table.snapshot(snapshotId).schemaId();
-        TableSchema schema = table.schemaManager().schema(schemaId);
         return new TableAwareFileStoreSourceSplit(
-                splitId, split, 0, identifier, lastSchema, schema);
+                splitId, split, 0, identifier, lastSchemaId, schemaId);
     }
 
     /**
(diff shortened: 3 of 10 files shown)


docs/layouts/shortcodes/generated/cdc_configuration.html
```diff
@@ -0,0 +1,48 @@
...
+            <td style="word-wrap: break-word;">(none)</td>
+            <td>String</td>
+            <td>Name of the table to be scanned. By default, all tables will be scanned.</td>
+        </tr>
+        <tr>
+            <td><h5>table.discovery-interval</h5></td>
+            <td style="word-wrap: break-word;">1 min</td>
+            <td>Duration</td>
+            <td>The discovery interval of new tables. Only effective when database or table is not set.</td>
+        </tr>
```
lvyanquan, line 46: Missing documentation for parameters with the prefix `catalog.properties`.

paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/pipeline/cdc/source/enumerator/CDCSourceEnumerator.java
```diff
@@ -0,0 +1,453 @@
...
+        this.numTablesWithSubtaskAssigned = 0;
+
+        int splitMaxPerTask = catalogContext.options().get(CoreOptions.SCAN_MAX_SPLITS_PER_TASK);
+        this.splitMaxNum = context.currentParallelism() * splitMaxPerTask;
+
+        this.splitAssigner = new CDCSplitAssigner(splitMaxPerTask, context.currentParallelism());
+        this.readersAwaitingSplit = new LinkedHashSet<>();
+        this.splitIdGenerator = new SplitIdGenerator();
+
+        if (checkpoint != null) {
```
lvyanquan: The processing here can be merged into the previous `if (checkpoint != null)` condition.

paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/pipeline/cdc/source/enumerator/CDCSourceEnumerator.java
```diff
@@ -0,0 +1,453 @@
...
+        if (lastTableDiscoveryTime < System.currentTimeMillis() - tableDiscoveryInterval) {
+            return;
+        }
+        lastTableDiscoveryTime = System.currentTimeMillis();
+
+        List<String> databaseNames;
+        if (database == null || database.isEmpty()) {
+            Preconditions.checkArgument(
+                    table == null || table.isEmpty(),
+                    "Tables should not be specified when databases is not. But tables is specified as "
```
lvyanquan: Tables should not be specified when databases is not  =>
Tables should not be specified when databases is null.

paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/pipeline/cdc/source/enumerator/CDCSourceEnumerator.java
```diff
@@ -0,0 +1,453 @@
...
+            List<String> tableNames;
+            if (table == null || table.isEmpty()) {
+                // If tables parameter is not specified but database is specified, dynamically scan
+                // all tables in the database
+                tableNames = catalog.listTables(dbName);
+            } else {
+                tableNames = Collections.singletonList(table);
+            }
+
+            for (String tableName : tableNames) {
```
lvyanquan, line 370: It seems that regular expression matching is not supported here. I'm not sure whether Paimon Source has a requirement to process only certain tables, as sharding by database/table is unlikely in Paimon tables. 

Therefore, you may consider whether to allow configuring database/table names as regular expressions (or adding a new compatible parameter in the future, which would also be acceptable).

paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/pipeline/cdc/source/enumerator/CDCSourceEnumerator.java
```diff
@@ -0,0 +1,453 @@
...
+                    LOG.warn(
+                            "Table {} used to be a FileStoreTable in the last checkpoint, but after recovery it turned out to be a {}. Skipping this table.",
+                            identifier,
+                            table.getClass().getSimpleName());
+                    continue;
+                }
+
+                TableStatus tableStatus = new TableStatus(context, (FileStoreTable) table);
+                tableStatus.restore(checkpoint.getCurrentSnapshotIdMap().get(identifier));
+                tableStatusMap.put(identifier, tableStatus);
```
lvyanquan, line 144: It's better to add more logs here for easier troubleshooting.

paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/pipeline/cdc/CDCOptions.java
```diff
@@ -0,0 +1,105 @@
...
+import org.apache.paimon.options.FallbackKey;
+
+import org.apache.flink.util.IterableUtils;
+
+import java.time.Duration;
+
+/** Options for the cdc source. */
+public class CDCOptions {
+    /** prefix for passing properties for catalog creation. */
+    public static final String PREFIX_CATALOG_PROPERTIES = "catalog.properties.";
```
lvyanquan, line 32: Should we provide the ability to specify parameters like `scan.mode` and `consumer-id` for the source table?

paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/pipeline/cdc/source/reader/CDCSourceReader.java
```diff
@@ -0,0 +1,94 @@
...
+        // it is possible that one response from the coordinator contains multiple splits
+        // we should only require for more splits after we've consumed all given splits
+        if (getNumberOfCurrentlyAssignedSplits() == 0) {
+            context.sendSplitRequest();
+        }
+    }
+
+    @Override
+    protected FileStoreSourceSplitState initializedState(TableAwareFileStoreSourceSplit split) {
+        return new FileStoreSourceSplitState(split);
```
lvyanquan, line 84: We can print the split information here to monitor the data reading progress during job startup.

We can also add logs in `toSplitType` to track progress during checkpointing (this depends on your preference, as `toSplitType` is not only called during checkpointing).

paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/pipeline/cdc/source/reader/CDCRecordsWithSplitIds.java
```diff
@@ -0,0 +1,140 @@
...
+        long timestamp = TimestampAssigner.NO_TIMESTAMP;
+        if (metrics.getLatestFileCreationTime() != FileStoreSourceReaderMetrics.UNDEFINED) {
+            timestamp = metrics.getLatestFileCreationTime();
+        }
+
+        // This metric only counts the number of RecordIterator<Event> emitted,
+        // however what we really want is to count the number of Events emitted,
+        // so we replenish the missing record count here.
+        org.apache.flink.metrics.Counter numRecordsIn =
+                context.metricGroup().getIOMetricGroup().getNumRecordsInCounter();
```
lvyanquan, line 123: The Counter here should be at the Source level. 
Should we provide metrics for individual tables?

docs/layouts/shortcodes/generated/cdc_configuration.html
yunfengzhou-hub: This file is automatically generated through `ConfigOptionsDocGenerator` and cannot be manually modified. I added description for such parameters directly to `flink-cdc-paimon-pipeline-connector.md`.

paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/pipeline/cdc/source/enumerator/CDCSourceEnumerator.java
yunfengzhou-hub: Since this PR has been so big for reviewers and regex is an extensive function, I prefer to add such support in future in a separated PR.

paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/pipeline/cdc/CDCOptions.java
yunfengzhou-hub: `scan.mode` should have been supported, as `AbstractDataTableScan#createStartingScanner` has been using this parameter.

For `consumer-id` and other potential parameters, I prefer to have them checked and supported in separated PRs in future. This is because this PR has been big enough for reviewers, and such parameters are extensive functionalities built upon the basic read/write correctness, and thus can be separated into future PRs.

paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/pipeline/cdc/source/reader/CDCRecordsWithSplitIds.java
yunfengzhou-hub: Same as above, this PR has been big enough for reviewers and I prefer to have detailed metrics supported in future PRs.

paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/pipeline/cdc/source/enumerator/CDCSourceEnumerator.java
```diff
@@ -0,0 +1,457 @@
...
+
+        int subtaskId = numTablesWithSubtaskAssigned % context.currentParallelism();
+        this.tableStatusMap.get(split.getIdentifier()).subtaskId = subtaskId;
+        this.numTablesWithSubtaskAssigned++;
+        LOG.info("Assigning table {} to subtask {}", split.getIdentifier(), subtaskId);
+        return subtaskId;
+    }
+
+    private void discoverTables() throws Exception {
+        if (lastTableDiscoveryTime < System.currentTimeMillis() - tableDiscoveryInterval) {
```
Sxnan: Should it be `>` instead of `<`?

paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/pipeline/cdc/schema/CDCMetadataAccessor.java
```diff
@@ -0,0 +1,108 @@
...
+    public Schema getTableSchema(TableId tableId) {
+        if (tableId.getNamespace() != null) {
+            throw new UnsupportedOperationException("Paimon does not support namespaces");
+        }
+
+        try {
+            Identifier identifier =
+                    Identifier.create(tableId.getSchemaName(), tableId.getTableName());
+            Table table = catalog.getTable(identifier);
+            return convertPaimonSchemaToFlinkCDCSchema(((FileStoreTable) table).schema());
```
Sxnan, line 128: Should we check if the `Table` is `FileStoreTable`? If not, should we just ignore or throw an exception?

paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/pipeline/cdc/source/enumerator/CDCSourceEnumerator.java
```diff
@@ -0,0 +1,457 @@
...
+        private final StreamTableScan scan;
+        private TableSchema schema;
+        private Integer subtaskId;
+        private Long nextSnapshotId;
+
+        private TableStatus(SplitEnumeratorContext<?> context, FileStoreTable table) {
+            this.table = table;
+            this.scan = table.newStreamScan();
+            if (metricGroup(context) != null) {
+                ((StreamDataTableScan) this.scan)
```
Sxnan: I think the `org.apache.paimon.table.InnerTable#newStreamScan` method returns `StreamDataTableScan` already.  Why do we need the cast here? Can we make the field `scan` as `StreamDataTableScan` type?

paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/pipeline/cdc/source/enumerator/CDCSourceEnumerator.java
```diff
@@ -0,0 +1,457 @@
...
+
+    private final int splitMaxNum;
+    private final CDCSplitAssigner splitAssigner;
+    private final Set<Integer> readersAwaitingSplit;
+    private final SplitIdGenerator splitIdGenerator;
+    private final Map<Identifier, TableStatus> tableStatusMap = new HashMap<>();
+    private final Set<Identifier> discoveredNonFileStoreTables = new HashSet<>();
+    private int numTablesWithSubtaskAssigned;
+
+    private boolean stopTriggerScan = false;
```
Sxnan, line 93: If I understand correctly, this is used to avoid scanNextSnapshot for every `handleSplitRequest` if the SNAPSHOT does not exist at the moment. In this case, we rely on the background discovery that is triggered at `discoveryInterval`.

We may want to document such behavior

paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/pipeline/cdc/source/enumerator/CDCSourceEnumerator.java
```diff
@@ -0,0 +1,457 @@
...
+
+/** {@link SplitEnumerator} for CDC source. */
+public class CDCSourceEnumerator
+        implements SplitEnumerator<TableAwareFileStoreSourceSplit, CDCCheckpoint> {
+    private static final Logger LOG = LoggerFactory.getLogger(CDCSourceEnumerator.class);
+
+    private final SplitEnumeratorContext<TableAwareFileStoreSourceSplit> context;
+    private final long discoveryInterval;
+    private final long tableDiscoveryInterval;
+    private long lastTableDiscoveryTime = Long.MAX_VALUE;
```
Sxnan: Should this be initialized as `0`?

paimon-flink/paimon-flink-cdc/pom.xml
```diff
@@ -92,6 +92,31 @@ under the License.
...
 
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-cdc-common</artifactId>
+            <version>${flink.cdc.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-cdc-runtime</artifactId>
```
Sxnan, line 103: Why do we need to include `flink-cdc-runtime`? If I understand correctly, we are only implementing some `flink-cdc` interface. And should not depend on the runtime module.

paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/pipeline/cdc/source/reader/CDCSourceSplitReader.java
```diff
@@ -0,0 +1,390 @@
...
+    }
+
+    private void checkSplitOrStartNext() throws IOException {
+        if (currentReader != null) {
+            return;
+        }
+
+        final TableAwareFileStoreSourceSplit nextSplit = splits.poll();
+        if (nextSplit == null) {
+            throw new IOException("Cannot fetch from another split - no split remaining");
```
Sxnan: Is it an unexpected state to throw an exception? I think it is possible that the reader consumes all the splits at the moment, but the split enumerator may assign more splits to the reader

paimon-flink/paimon-flink-cdc/pom.xml
yunfengzhou-hub: This PR needs to use `BinaryRecordDataGenerator` to convert Paimon records to CDC Events, which is introduced through `flink-cdc-runtime`. I've discusses with @lvyanquan to see if we can provide such a generator tool class in non-runtime Flink cdc modules.

paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/pipeline/cdc/source/enumerator/CDCSourceEnumerator.java
yunfengzhou-hub: Thanks for the suggestion. I added description of this behavior to the "Usage Notes" section of `flink-cdc-paimon-pipeline-connector.md`.

paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/pipeline/cdc/source/TableAwareFileStoreSourceSplit.java
```diff
@@ -0,0 +1,154 @@
...
+    @Nullable private final TableSchema lastSchema;
+    private final TableSchema schema;
+
+    public TableAwareFileStoreSourceSplit(
+            String id,
+            Split split,
+            long recordsToSkip,
+            Identifier identifier,
+            @Nullable TableSchema lastSchema,
+            TableSchema schema) {
```
lvyanquan: Could we store the id of `schema` and get `TableSchema` using `SchemaManager#schema` to reduce the size of state?

paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/pipeline/cdc/source/TableAwareFileStoreSourceSplit.java
```diff
@@ -0,0 +1,154 @@
...
+        public int getVersion() {
+            return 1;
+        }
+
+        @Override
+        public byte[] serialize(TableAwareFileStoreSourceSplit split) throws IOException {
+            ByteArrayOutputStream out = new ByteArrayOutputStream();
+            DataOutputViewStreamWrapper view = new DataOutputViewStreamWrapper(out);
+            view.writeUTF(split.splitId());
+            InstantiationUtil.serializeObject(view, split.split());
```
lvyanquan, line 127: I am concerned that this part may lead to state incompatibility due to modifications in Split. However, since `FileStoreSourceSplitSerializer` also implements this approach, I believe such incompatibility issues should be avoided through adjustments in Split.

paimon-flink/paimon-flink-cdc/pom.xml
lvyanquan: Yes, this is a legacy issue from historical development. 
We will discuss in the community to move the dependent code in this part to the common module.

paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/pipeline/cdc/source/TableAwareFileStoreSourceSplit.java
lvyanquan: Since our Split is `DataSplit`(Based on the CDCSourceEnumerator#toTableAwareSplit method), we do not need to store historical SchemaIds.

lvyanquan: Although this reduces the size of the state, it results in additional file read operations on the subtask, refer to https://github.com/apache/paimon/blob/87ee257c2d267093cec41c1f95dd58211f2f3347/paimon-core/src/main/java/org/apache/paimon/schema/SchemaManager.java#L1034. 
Therefore, it is acceptable to store the schema in the state.

Sxnan (approved):
Thanks for the update! LGTM

PR #6607, opened 2025-11-14
[spark] Wrap the corresponding table to ensure it guarantees its capabilities

Zouxxyy:
### Purpose

 Iceberg, lance and object table should not support read and write yet, add wrap for them


JingsongLi:
+1

PR #6608, opened 2025-11-15
[common] Bump janino version to 3.1.12

YannByron:
### Purpose

Bump janino version to 3.1.12


JingsongLi:
Why do you want to upgrade janino? It was shaded and relocated.

PR #6609, opened 2025-11-15
[WIP] [core] Global index support code in paimon-core part.

leaves12138:
### Purpose

Related to pip: https://cwiki.apache.org/confluence/display/PAIMON/PIP-38%3A+Introduce+Global+Index+for+Paimon+Table

This pull request is the POC code in core part.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/index/IndexFileMeta.java
+++ b/paimon-core/src/main/java/org/apache/paimon/index/IndexFileMeta.java
@@ -60,10 +60,16 @@ public class IndexFileMeta {
                                     6,
                                     "_GLOBAL_INDEX",
                                     new RowType(
-                                            false,
+                                            true,
                                             Arrays.asList(
-                                                    new DataField(0, "_ROW_RANGE_START", new BigIntType(true)),
-                                                    new DataField(1, "_ROW_RANGE_END", new BigIntType(true)),
+                                                    new DataField(
+                                                            0,
+                                                            "_ROW_RANGE_START",
+                                                            new BigIntType(true)),
+                                                    new DataField(
+                                                            1,
+                                                            "_ROW_RANGE_END",
+                                                            new BigIntType(true)),
                                                     new DataField(
                                                             2,
                                                             "_INDEX_FIELD_ID",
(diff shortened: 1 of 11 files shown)


paimon-core/src/main/java/org/apache/paimon/index/IndexFileMeta.java
```diff
@@ -53,13 +55,20 @@ public class IndexFileMeta {
...
+                            new DataField(6, "_SHARD", new IntType(true)),
+                            new DataField(7, "_INDEX_FIELD_ID", new IntType(true)),
+                            new DataField(8, "_INDEX_META", new BinaryType())));
 
     private final String indexType;
     private final String fileName;
     private final long fileSize;
     private final long rowCount;
 
+    @Nullable private final Integer shard;
```
JingsongLi: Record row id range instead of shard id.

JingsongLi: Introduce a new structure: `GlobalIndexMeta`.

paimon-common/src/main/java/org/apache/paimon/fileindex/bitmap/BitmapReaderHelper.java
```diff
@@ -0,0 +1,221 @@
...
+import java.util.Map;
+import java.util.function.Function;
+
+/**
+ * Helper class for bitmap readers that provides common functionality for reading and querying
+ * bitmap indexes using composition pattern.
+ *
+ * @param <BITMAP> The bitmap type (e.g., RoaringBitmap32 or RoaringBitmap64)
+ */
+public class BitmapReaderHelper<BITMAP> {
```
JingsongLi, line 38: Should not extract helpers for bitmap, you can just use `BitmapFileIndex` to build global index.

paimon-core/src/main/java/org/apache/paimon/index/globalindex/GlobalIndexResult.java
```diff
@@ -0,0 +1,165 @@
...
+        Iterator<Range> it2 = fileIndexResult.results();
+
+        Set<Range> intersection = RangeUtils.or(it1, it2);
+
+        return wrap(intersection);
+    }
+
+    boolean empty();
+
+    Iterator<Range> results();
```
JingsongLi, line 114: You can just introduce results to an unified `IndexResult`.

paimon-common/src/main/java/org/apache/paimon/fileindex/bitmap/BitmapReaderHelper.java
JingsongLi, line 38: And you can introduce 64 support for bitmap first.

PR #6612, opened 2025-11-16
[spark] support rescale procedure in spark

XiaoHongbo-Hope:
### Tests
`RescaleProcedureTest`

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/procedure/RescaleProcedure.java
+++ b/paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/procedure/RescaleProcedure.java
@@ -117,7 +117,7 @@ public class RescaleProcedure extends BaseProcedure {
                         throw new IllegalArgumentException(
                                 "Table "
                                         + table.fullName()
-                                        + " has no snapshot. No need to rescale.");
+                                        + " has no snapshot, no need to rescale.");
                     }
                     Snapshot snapshot = optionalSnapshot.get();
 
--- a/paimon-spark/paimon-spark-ut/src/test/scala/org/apache/paimon/spark/procedure/RescaleProcedureTest.scala
+++ b/paimon-spark/paimon-spark-ut/src/test/scala/org/apache/paimon/spark/procedure/RescaleProcedureTest.scala
@@ -26,12 +26,11 @@ import org.apache.paimon.table.FileStoreTable
 import org.apache.spark.sql.Row
 import org.assertj.core.api.Assertions
 
-import java.util
 import java.util.{Arrays, Collections, Map => JMap}
 
 import scala.collection.JavaConverters._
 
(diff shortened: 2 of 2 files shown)


paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/procedure/RescaleProcedure.java
```diff
@@ -0,0 +1,245 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(RescaleProcedure.class);
+
+    private static final ProcedureParameter[] PARAMETERS =
+            new ProcedureParameter[] {
+                ProcedureParameter.required("table", StringType),
+                ProcedureParameter.optional("bucket_num", IntegerType),
+                ProcedureParameter.optional("partitions", StringType),
+                ProcedureParameter.optional("where", StringType),
+                ProcedureParameter.optional("scan_parallelism", IntegerType),
```
Zouxxyy: Remove the scan_parallelism and sink_parallelism, Spark's read parallelism is automatically calculated based on the size of splits. and write parallelism is also automatically calculated through shuffle.

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/procedure/RescaleProcedure.java
```diff
@@ -0,0 +1,245 @@
...
+        Dataset<Row> datasetForRead =
+                PaimonUtils.createDataset(
+                        spark(),
+                        ScanPlanHelper$.MODULE$.createNewScanPlan(
+                                dataSplits.toArray(new DataSplit[0]), relation));
+
+        Dataset<Row> datasetForWrite = datasetForRead.repartition(scanParallelism);
+
+        Map<String, String> bucketOptions = new HashMap<>(table.options());
+        bucketOptions.put(CoreOptions.BUCKET.key(), String.valueOf(bucketNum));
```
Zouxxyy: Do we need to configure this when using postpone bucket

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/procedure/RescaleProcedure.java
XiaoHongbo-Hope: > Remove the scan_parallelism and sink_parallelism, Spark's read parallelism is automatically calculated based on the size of splits. and write parallelism is also automatically calculated through shuffle.

removed

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/procedure/RescaleProcedure.java
XiaoHongbo-Hope: > Do we need to configure this when using postpone bucket

updated

PR #6618, opened 2025-11-17
[core] Lookup supports Variant type

JingsongLi:
### Purpose

Lookup (changelog-producer=lookup or dv enabled with partial-update) does not support variant type. This is because  `RowCompactedSerializer` not support variant type.

This PR fixes this.

### Tests

- `VariantTypesTest`.
- `LookupVariantTableTest`.


Zouxxyy (approved):
+1

PR #6620, opened 2025-11-17
[core] format table: fix multi part two-phase stream flush bug

jerry-024:
### Purpose
format table: fix multi part two-phase stream flush bug. For object store, only the last part could be less than the threshold.

### Tests

- MultiPartUploadTwoPhaseOutputStreamTest.testFlushWhenBufferSizeIsSmallerThanThresholdDoesNotUpload


JingsongLi (approved):
+1

PR #6621, opened 2025-11-17
[core][spark] Introduce lastScannedSnapshotId in ScanMetrics

Zouxxyy:
### Purpose

The description method is invoked heavily during Spark execution, and calling latestSnapshot or currentSnapshot inside it would cause severe performance degradation.

Introduce lastScannedSnapshotId in ScanMetrics

image

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonBaseScan.scala
+++ b/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonBaseScan.scala
@@ -57,24 +57,6 @@ abstract class PaimonBaseScan(
 
   override val coreOptions: CoreOptions = CoreOptions.fromMap(table.options())
 
-  val latestSnapshot: Option[Snapshot] = table.latestSnapshot()
-
-  val currentSnapshot: Option[Snapshot] =
-    table match {
-      case dataTable: DataTable =>
-        try {
-          TimeTravelUtil
-            .tryTravelToSnapshot(
-              coreOptions.toConfiguration,
-              dataTable.snapshotManager(),
-              dataTable.tagManager())
-            .orElse(latestSnapshot)
-        } catch {
-          case _: Exception => None
-        }
-      case _ => None
-    }
-
   lazy val statistics: Optional[stats.Statistics] = table.statistics()
 
   private lazy val paimonMetricsRegistry: SparkMetricRegistry = SparkMetricRegistry()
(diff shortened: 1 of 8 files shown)


paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonBaseScan.scala
```diff
@@ -56,6 +57,24 @@ abstract class PaimonBaseScan(
 
   override val coreOptions: CoreOptions = CoreOptions.fromMap(table.options())
 
+  val latestSnapshot: Option[Snapshot] = table.latestSnapshot()
+
+  val currentSnapshot: Option[Snapshot] =
```
JingsongLi: Maybe you can get current snapshot from `getOriginSplits`.

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonBaseScan.scala
```diff
@@ -141,37 +160,14 @@ abstract class PaimonBaseScan(
...
 
-    val latestSnapshotId = if (table.latestSnapshot().isPresent) {
-      Some(table.latestSnapshot().get.id)
-    } else {
-      None
-    }
-
-    val latestSnapshotIdStr = if (latestSnapshotId.isDefined) {
-      s", LatestSnapshotId: [${latestSnapshotId.get}]"
+    val latestSnapshotIdStr = if (latestSnapshot.nonEmpty) {
```
JingsongLi: Why we need to get latest snapshot?

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonBaseScan.scala
Zouxxyy: Great, I will add `lastScannedSnapshotId` in `ScanMetrics`

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonBaseScan.scala
Zouxxyy: It's not very useful and will increase request, I'm inclined to remove it.

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonBaseScan.scala
JingsongLi: Oh nice, let metric come from metric.

PR #6625, opened 2025-11-18
[spark] support CopyFilesProcedure in spark

shyjsarah:
### Purpose

Linked issue: close #xxx

support CopyFilesProcedure in spark

Main Process
1. create target table and get latest snapshot
2. list data and index files
3. copy data and index files
4. commit table

### Tests

org.apache.paimon.spark.procedure.CopyFilesProcedureTest

### Documentation

docs/content/spark/procedures.md

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/copy/CopyDataFilesOperator.java
+++ b/paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/copy/CopyDataFilesOperator.java
@@ -78,7 +78,7 @@ public class CopyDataFilesOperator extends CopyFilesOperator {
                 Path sourcePath = new Path(file.sourceFilePath());
                 Path targetPath = new Path(file.targetFilePath());
                 CopyFilesUtil.copyFiles(
-                        sourceTable.fileIO(), targetTable.fileIO(), sourcePath, targetPath, true);
+                        sourceTable.fileIO(), targetTable.fileIO(), sourcePath, targetPath, false);
                 result.add(file);
             }
             return result.iterator();
--- a/paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/copy/ListDataFilesOperator.java
+++ b/paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/copy/ListDataFilesOperator.java
@@ -98,7 +98,7 @@ public class ListDataFilesOperator extends CopyFilesOperator {
                 targetFileStorePathFactory
                         .createDataFilePathFactory(
                                 manifestEntry.partition(), manifestEntry.bucket())
-                        .toPath(manifestEntry);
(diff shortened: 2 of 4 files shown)


paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/copy/CopyDataFilesOperator.java
```diff
@@ -0,0 +1,89 @@
...
+        @Override
+        public Iterator<CopyFileInfo> call(Iterator<CopyFileInfo> dataFileIterator)
+                throws Exception {
+            List<CopyFileInfo> result = new ArrayList<>();
+            Path targetTableRootPath = targetTable.location();
+            while (dataFileIterator.hasNext()) {
+                CopyFileInfo dataFile = dataFileIterator.next();
+                String filePathExcludeTableRoot = dataFile.filePathExcludeTableRoot();
+                Path sourcePath = new Path(dataFile.sourceFilePath());
+                Path targetPath = new Path(targetTableRootPath + filePathExcludeTableRoot);
```
JingsongLi: Possible path inconsistency due to table options reasons

JingsongLi: Copy the same name from the past, duplicate copies may result in duplicate commit files, cause issues with the table

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/copy/CopyDataFilesOperator.java
```diff
@@ -0,0 +1,89 @@
...
+                CopyFilesUtil.copyFiles(
```
JingsongLi, line 80: We should check that the schema corresponding to the file is consistent with the schema of the new table, otherwise it will cause inconsistency issues.

It's not a problem when the fields are missing.

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/copy/CopySchemaOperator.java
```diff
@@ -0,0 +1,85 @@
...
+        Preconditions.checkState(
+                originalSourceTable instanceof FileStoreTable,
+                String.format(
+                        "Only support copy FileStoreTable, but this table %s is %s.",
+                        sourceIdentifier, sourceIdentifier.getClass()));
+        FileStoreTable sourceTable = (FileStoreTable) originalSourceTable;
+
+        // 1. create target table
+        targetCatalog.createDatabase(targetIdentifier.getDatabaseName(), true);
+        targetCatalog.createTable(
```
JingsongLi: We should support existing table.

JingsongLi (approved):
+1

PR #6627, opened 2025-11-18
[flink] Set max parallelism for batch append table compact source

yuzelin:
### Purpose

For unaware append, allow to set unawareAppendPerTaskDataSize for infer parallelism;
For bucketed append, allow to set bucketedAppendPerTaskBuckets for infer parallelism;
And allow to set appendMaxParallelism for limit the max parallelism.


JingsongLi (changes requested):
Thanks @yuzelin 
Left comments:
1. You should modify `clustering.per-subtask.data-size` to `compaction.infer-data-size-per-task` and apply this option into here. The default should be none, by default, no infer.
2. You should introduce `compaction.infer-parallelism.max`, default none, no limit.
3. Just use bucket number to infer parallelism for bucketed append table.

PR #6628, opened 2025-11-18
[core] Introduce basic structure for global index write

leaves12138:
### Purpose


paimon-common/src/main/java/org/apache/paimon/globalindex/GlobalIndexWriter.java
```diff
@@ -0,0 +1,54 @@
...
+public interface GlobalIndexWriter {
+
+    void write(Object key);
+
+    List<ResultEntry> finish();
+
+    /** Write result meta. */
+    class ResultEntry {
+        private final String fileName;
+        @Nullable private final String meta;
```
JingsongLi: byte[]

PR #6641, opened 2025-11-20
[core] Custom Partition expire factory should be invoked first

JingsongLi:
### Purpose

Let custom factory control all strategies.


XiaoHongbo-Hope:
+1

PR #6642, opened 2025-11-20
[core] Let short streaming job execute partition expire

JingsongLi:
### Purpose

Avoid the execution time of stream jobs from being too short and preventing partition expiration.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/PartitionExpire.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/PartitionExpire.java
@@ -79,8 +79,9 @@ public class PartitionExpire {
         // Avoid the execution time of stream jobs from being too short and preventing partition
         // expiration
         long rndSeconds = 0;
-        if (!checkInterval.isZero()) {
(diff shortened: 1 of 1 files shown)


paimon-core/src/main/java/org/apache/paimon/operation/PartitionExpire.java
```diff
@@ -75,7 +76,13 @@ public PartitionExpire(
         this.scan = scan;
         this.commit = commit;
         this.partitionHandler = partitionHandler;
-        this.lastCheck = LocalDateTime.now();
+        // Avoid the execution time of stream jobs from being too short and preventing partition
+        // expiration
+        long rndSeconds = 0;
+        if (!checkInterval.isZero()) {
+            rndSeconds = ThreadLocalRandom.current().nextLong(checkInterval.toMillis() / 1000);
```
tsreaper: extreme case: `0 < checkInterval < 1000`.

JingsongLi: Let it be zero always.

JingsongLi: I will fix it.

PR #6647, opened 2025-11-21
[core] filter irrelevant column files in DataEvolutionFileStoreScan

steFaiz:
### Purpose
Currently in DataEvolutionFileStoreScan, we filter datafiles by rowIdList. I notice that we can also only return datafiles containing target projected columns.

Linked issue: None

### Tests
see org.apache.paimon.table.DataEvolutionTableTest

### API and Format
Do not modify any existing public user API.

### DIFF
--- a/paimon-core/src/test/java/org/apache/paimon/table/DataEvolutionTableTest.java
+++ b/paimon-core/src/test/java/org/apache/paimon/table/DataEvolutionTableTest.java
@@ -827,7 +827,6 @@ public class DataEvolutionTableTest extends TableTestBase {
                                         table.rowType().project("f1")));
         RecordReader<InternalRow> reader =
                 readBuilder.newRead().createReader(readBuilder.newScan().plan());
-        Assertions.assertThat(reader).isInstanceOf(DataEvolutionFileReader.class);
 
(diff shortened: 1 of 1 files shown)


steFaiz:
converting to draft to fix spark test.

paimon-core/src/test/java/org/apache/paimon/table/DataEvolutionTableTest.java
```diff
@@ -796,7 +827,6 @@ public void testGlobalIndex() throws Exception {
                                         table.rowType().project("f1")));
         RecordReader<InternalRow> reader =
                 readBuilder.newRead().createReader(readBuilder.newScan().plan());
-        Assertions.assertThat(reader).isInstanceOf(DataEvolutionFileReader.class);
```
steFaiz, line 799: remove this assertion because in https://github.com/apache/paimon/blob/74cf00b981917f327fc5bb9d3682ce7bf1d9b62b/paimon-core/src/main/java/org/apache/paimon/operation/DataEvolutionSplitRead.java#L157

we will only create a UnionReader if we have more than one file to read. However in this scenario, only one column is projected and we only need to read one File, so the created reader is just a FileReader.

PR #6649, opened 2025-11-21
[spark] Support data evolution for bucket table

shidayang:
### Purpose
Support data-evolution for append only fixed hash table


leaves12138:
@JingsongLi

paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
```diff
@@ -632,7 +632,7 @@ private static void validateRowTracking(TableSchema schema, CoreOptions options)
         boolean rowTrackingEnabled = options.rowTrackingEnabled();
         if (rowTrackingEnabled) {
             checkArgument(
```
JingsongLi: remove whole `checkArgument`.

docs/content/append-table/row-tracking.md
```diff
@@ -47,7 +47,7 @@ CREATE TABLE part_t (
 WITH ('row-tracking.enabled' = 'true');
 ```
 Notice that:
-- Row tracking is only supported for unaware append tables, not for primary key tables. Which means you can't define `bucket` and `bucket-key` for the table.
+- Row tracking is only supported for unaware or hash_fixed bucket append tables, not for primary key tables.
```
JingsongLi: `supported for append tables`

paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
```diff
@@ -1284,8 +1283,7 @@ private long assignRowTrackingMeta(
             checkArgument(
                     entry.file().fileSource().isPresent(),
                     "This is a bug, file source field for row-tracking table must present.");
-            if (entry.file().fileSource().get().equals(FileSource.APPEND)
```
JingsongLi: Why modify here?

shidayang: Because when performing compact in writer, the file will be rewritten by compact before it is committed and the rowId is generated. This scenario will result in the file generated by compact not having the rowid

JingsongLi: For row-tracking only tables, compact files should not be assigned new row ids, row id already in Data File.

JingsongLi: Maybe we should have a new bucketed append table mode here. Without considering sequence and compaction. Just like normal Append tables.

shidayang: > Maybe we should have a new bucketed append table mode here. Without considering sequence and compaction. Just like normal Append tables.

good idea

JingsongLi:
The `bucket-append-ordered` is for this mode.

shidayang:
> The `bucket-append-ordered` is for this mode.

get,I'll do some fixes

shidayang:
@JingsongLi ，PTAL，I only disable compaction for row-tracking bucket table in this pr. I will handle flink and Spark compaction behaviors in other PR

JingsongLi:
Let's start new PR if you think this is a good direction.

PR #6652, opened 2025-11-21
[variant] Introduce assemble shredded variant struct function

Zouxxyy:
### Purpose

Subtask of https://github.com/apache/paimon/issues/4471

Add in `Variant`

```java
Object variantGet(String path, DataType dataType, VariantCastArgs castArgs);
```

Add in `PaimonShreddingUtils`
```java
/**
 * Represent a single field in a variant struct, that is a single requested field that the scan
 * should produce by extracting from the variant column.
 */
public static class FieldToExtract {

    private final SchemaPathSegment[] path;
    private final BaseVariantReader reader;
}

/** Assemble a variant struct, in which each field is extracted from the variant value. */
public static InternalRow assembleVariantStruct(
        InternalRow inputRow, VariantSchema schema, FieldToExtract[] fields)
）
```

Usage
```java
VariantCastArgs castArgs = new VariantCastArgs(true, ZoneOffset.UTC);

// shreddedType: ROW<a INT, b STRING>
RowType shreddedType =
        RowType.of(
                new DataType[] {DataTypes.INT(), DataTypes.STRING()},
                new String[] {"a", "b"});
RowType shreddingSchema = variantShreddingSchema(shreddedType);
VariantSchema variantSchema = buildVariantSchema(shreddingSchema);

// fieldsToExtract: $.a : int, $.b : string
FieldToExtract f1 = buildFieldsToExtract(DataTypes.INT(), "$.a", castArgs, variantSchema);
FieldToExtract f2 = buildFieldsToExtract(DataTypes.STRING(), "$.b", castArgs, variantSchema);
FieldToExtract[] fields = new FieldToExtract[] {f1, f2};

GenericVariant v = GenericVariant.fromJson("{\"a\": 1, \"b\": \"hello\"}");
assertThat(assembleVariantStruct(castShredded(v, variantSchema), variantSchema, fields))
        .isEqualTo(GenericRow.of(1, BinaryString.fromString("hello")));
```

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/data/variant/GenericVariant.java
+++ b/paimon-common/src/main/java/org/apache/paimon/data/variant/GenericVariant.java
@@ -138,11 +138,6 @@ public final class GenericVariant implements Variant, Serializable {
         }
     }
 
-    @Override
-    public String toJson() {
-        return toJson(ZoneOffset.UTC);
-    }
-
     // Stringify the variant in JSON format.
     // Throw `MALFORMED_VARIANT` if the variant is malformed.
     @Override
--- a/paimon-common/src/main/java/org/apache/paimon/data/variant/Variant.java
+++ b/paimon-common/src/main/java/org/apache/paimon/data/variant/Variant.java
@@ -21,6 +21,7 @@ package org.apache.paimon.data.variant;
 import org.apache.paimon.types.DataType;
 
 import java.time.ZoneId;
+import java.time.ZoneOffset;
 
 /**
  * A Variant represents a type that contain one of: 1) Primitive: A type and corresponding value
@@ -47,7 +48,9 @@ public interface Variant {
     byte[] value();
 
     /** Parses the variant to json. */
-    String toJson();
+    default String toJson() {
+        return toJson(ZoneOffset.UTC);
+    }
 
     /** Parses the variant to json with zoneId. */
     String toJson(ZoneId zoneId);


paimon-common/src/main/java/org/apache/paimon/data/variant/VariantPathSegment.java
```diff
@@ -26,38 +26,9 @@
...
-    }
-
-    public String getKey() {
-        return key;
-    }
-
-    public Integer getIndex() {
-        return index;
-    }
+    public VariantPathSegment() {}
```
yunfengzhou-hub: Can we remove this empty constructor?

paimon-common/src/main/java/org/apache/paimon/data/variant/VariantPathSegment.java
```diff
@@ -99,6 +70,36 @@ public static PathSegment[] parse(String str) {
             throw new IllegalArgumentException("Invalid path: " + str);
         }
 
-        return segments.toArray(new PathSegment[0]);
+        return segments.toArray(new VariantPathSegment[0]);
+    }
+
+    /** A path segment for object extraction. */
+    public static class ObjectExtraction extends VariantPathSegment {
```
yunfengzhou-hub: The parent abstract class is all static, and this class needs not inherit anything from the parent class. Not sure why this inheritance is applied.

PR #6659, opened 2025-11-22
[core] Abstract LookupSerializerFactory and extract classes from LookupLevels

JingsongLi:
### Purpose

Refactor codes for remote sst files. Abstract `LookupSerializerFactory` and extract classes from LookupLevels to clean LookupLevels.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/mergetree/lookup/PersistPositionProcessor.java
+++ b/paimon-core/src/main/java/org/apache/paimon/mergetree/lookup/PersistPositionProcessor.java
@@ -20,34 +20,18 @@ package org.apache.paimon.mergetree.lookup;
 
 import org.apache.paimon.KeyValue;
 import org.apache.paimon.data.InternalRow;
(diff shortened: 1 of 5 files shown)


paimon-core/src/main/java/org/apache/paimon/mergetree/lookup/PersistPositionProcessor.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    @Override
+    public byte[] persistToDisk(KeyValue kv) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public byte[] persistToDisk(KeyValue kv, long rowPosition) {
+        if (persistValue) {
```
tsreaper: Both `persistToDisk` and `readFromDisk` need to check `persistValue`. It seems that this class actually combines two `PersistProcessor`. Split them into two classes.

tsreaper (approved):

PR #6662, opened 2025-11-24
[spark] fix drop partition when partition value is null or empty string

zhongyujiang:
### Purpose

Currently, when using Spark to delete Paimon partitions, deletion fails or incorrectly removes mismatched data if the partition value is NULL or an empty string. This issue occurs across `ALTER TABLE ... DROP PARTITION`, `DELETE FROM`, and `TRUNCATE PARTITION`

**ALTER TABLE T DROP PARTITION and DELETE FROM**
Dropping a partition with partitionField = '' erroneously deletes data where partitionField = null.

The root cause is that during deletion, Spark SQL passes partition values that are replaced by `InternalRowPartitionComputer` with the default partition value. However, the actual partition filter should use the original (real) partition values for matching, because the manifest files store the real partition values—not the default ones.

This PR resolves the issue by preserving the original partition values (instead of replacing them with the default partition value) when computing partitions for data deletion: `InternalRowPartitionComputer preserveNullOrEmptyValue`

**TRUNCATE PARTITION**
Throws a NullPointerException (NPE) when the partition value is null.

Fixed by safely handling null values.

### Tests

Added in the PR.

### API and Format

No.

### Documentation

No.

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/utils/InternalRowPartitionComputer.java
+++ b/paimon-common/src/main/java/org/apache/paimon/utils/InternalRowPartitionComputer.java
@@ -31,6 +31,7 @@ import org.apache.paimon.types.VarCharType;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.function.Function;
 import java.util.stream.IntStream;
 
 import static org.apache.paimon.utils.InternalRowUtils.createNullCheckingFieldGetter;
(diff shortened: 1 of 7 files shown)


JingsongLi (commented):
You mean `HiveCatalog`? Can we fix `HiveCatalog.dropPartitions` to deal with default values?

zhongyujiang:
@JingsongLi Thanks for looking at this.

I think this issue occurs for all catalogs. The root cause lies in Spark’s current partition-dropping logic, which incorrectly transforms partition values using InternalRowPartitionComputer—not in the catalog implementation itself. Here's how the issue unfolds:

When executing ALTER TABLE T DROP PARTITION(pt=''), PaimonPartitionManagement#toPaimonPartitions uses InternalRowPartitionComputer to convert the Spark Row into a partition spec of type Map<String, String>. During this conversion, the empty string ('') is replaced with the default partition value, resulting in (pt='__DEFAULT_PARTITION__').

This spec is then passed to FileStoreCommitImpl#dropPartitions to delete the partition.

Inside dropPartitions, the partition spec (pt='__DEFAULT_PARTITION__') is converted back into a Paimon internal row via `InternalRowPartitionComputer#convertSpecToInternalRow` to construct the partition predicate. However, at this stage, '__DEFAULT_PARTITION__' is converted to null (see code below):

https://github.com/apache/paimon/blob/4a7cdf445f0a726266ecb3dd36b36c823c79e615/paimon-common/src/main/java/org/apache/paimon/utils/InternalRowPartitionComputer.java#L135-L153

As a result, data with pt = '' is not deleted; instead, data with pt = null gets deleted (if it exists), leading to incorrect behavior.

This also happens in `DELETE FROM`.

JingsongLi:
This may be a tricky issue, as Hive's previous definition was to treat empty strings and null as equivalent.

zhongyujiang:
> @zhongyujiang Are you referring to the issue of mixing empty strings and null?

@JingsongLi  Is not that the values are "mixed", but that the data with partition column values equal to an empty string cannot be deleted when using `DELETE FROM` or `ALTER TABLE T DROP PARTITION`.

Reproduce:
```scala
  test("Paimon Partition Management: partition values are empty string") {
    spark.sql(s"""
                 |CREATE TABLE T (pt STRING, data STRING)
                 |PARTITIONED BY (pt)
                 |""".stripMargin)

    sql("INSERT INTO T VALUES('', 'a'), ('2', 'b')")

    sql("ALTER TABLE T DROP PARTITION (pt = '')")

    spark.sql("SELECT * FROM T").show(false)
    //+---+----+
    //|pt |data|
    //+---+----+
    //|2  |b   |
    //|   |a   |
    //+---+----+
  }
```



> This may be a tricky issue, as Hive's previous definition was to treat empty strings and null as equivalent.

Oh, I wasn't aware of that. I noticed that Paimon currently places data with partition values of both null and empty string  into the same partition—the default partition. Is Paimon intended to be consistent with Hive in this behavior?

JingsongLi:
> Oh, I wasn't aware of that. I noticed that Paimon currently places data with partition values of both null and empty string into the same partition—the default partition. Is Paimon intended to be consistent with Hive in this behavior?

My answer is, if possible, it's best for us to maintain the status quo. Can this solve your business problem?

zhongyujiang:
@JingsongLi Do you have time to take another look? Thanks!

JingsongLi:
I merged your first commit.

JingsongLi:
Maybe can we provide a `void truncatePartitions(List<BinaryRow> partitions)` for `BatchTableCommit`?

PR #6663, opened 2025-11-24
[spark] push down partition filter when compactUnAwareBucketTable

summaryzb:
### Purpose
push down partition filter to reduce partition merge during read `partitionEntries`

### Tests

Pass existing test `Paimon Procedure: compact unaware bucket append table with option`

### API and Format

No

### Documentation

No


JingsongLi:
+1 thanks @summaryzb @zhongyujiang

PR #6666, opened 2025-11-25
[core] Introduce 'deletion-vectors.modifiable' to limit dv alter

### LINKED ISSUE #5874: [Bug] Enable deletion vector may lead to the wrong result

JingsongLi:
### Purpose

Cannot change deletion vectors mode . If modifying table deletion-vectors mode without full-compaction, this may result in data duplication.  If you are confident, you can set table option 'deletion-vectors.modifiable' = 'true' to allow deletion vectors modification.

This also fixes #5874 , in a workaround way...

### DIFF
--- a/paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/DeletionVectorITCase.java
+++ b/paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/DeletionVectorITCase.java
@@ -400,6 +400,7 @@ public class DeletionVectorITCase extends CatalogITCaseBase {
         // full compact
         sql("CALL sys.compact(`table` => 'default.TT')");
         // disable dv and select
+        sql("ALTER TABLE TT SET('deletion-vectors.modifiable' = 'true')");
         sql("ALTER TABLE TT SET('deletion-vectors.enabled' = 'false')");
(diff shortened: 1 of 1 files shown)


yuzelin (approved):
+1

PR #6670, opened 2025-11-25
[core] strict-mode check APPEND snapshot with fixed bucket files

yuzelin:
### Purpose

For example, to avoid rescale job overwrite fixed bucket files.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/commit/ConflictDetection.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/commit/ConflictDetection.java
@@ -531,8 +531,7 @@ public class ConflictDetection {
         if (baseEntries.size() > maxEntry || changes.size() > maxEntry) {
             baseEntriesString =
                     "Base entries are:\n"
-                            + baseEntries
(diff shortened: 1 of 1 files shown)


paimon-core/src/main/java/org/apache/paimon/operation/commit/ConflictDetection.java
```diff
@@ -91,12 +99,61 @@ public ConflictDetection(
...
+                                            + "commit user %s has found a %s snapshot (id: %d) by another user %s. "
+                                            + "Giving up committing as %s is set.",
+                                    newSnapshotId,
+                                    commitUser,
+                                    snapshot.commitKind().name(),
+                                    id,
+                                    snapshot.commitUser(),
+                                    CoreOptions.COMMIT_STRICT_MODE_LAST_SAFE_SNAPSHOT.key()));
+                }
+                if (snapshot.commitKind() == CommitKind.APPEND) {
```
JingsongLi: Maybe add this check for OVERWRITE commit only.

JingsongLi:
+1

PR #6685, opened 2025-11-26
[variant] Introduce withVariantAccess in ReadBuilder

Zouxxyy:
### Purpose

Subtask of https://github.com/apache/paimon/issues/4471, new function in `ReadBuilder` for variant push down

```java
/**
 * Push variant access to the reader.
 *
 * @param variantAccessInfo variant access info
 * @since 1.4.0
 */
ReadBuilder withVariantAccess(VariantAccessInfo[] variantAccessInfo);
```

`VariantAccessInfo`

```java
/** Variant access information for a variant column. */
public class VariantAccessInfo implements Serializable {

    // The name of the variant column.
    private final String columnName;

    // Extracted fields from the variant.
    private final List<VariantField> variantFields;

    /** Variant field extracted from the variant. */
    public static class VariantField implements Serializable {

        // The data field of the variant field.
        private final DataField dataField;

        // The `path` parameter of VariantGet.
        private final String path;

        private final VariantCastArgs castArgs;
}
```

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/source/ReadBuilderImpl.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/source/ReadBuilderImpl.java
@@ -83,7 +83,8 @@ public class ReadBuilderImpl implements ReadBuilder {
         RowType finalReadType = readType != null ? readType : table.rowType();
         // When variantAccessInfo is not null, replace the variant with the actual readType.
         if (variantAccessInfo != null) {
-            finalReadType = VariantAccessInfoUtils.buildReadRowType(finalReadType, variantAccessInfo);
+            finalReadType =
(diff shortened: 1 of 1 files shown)


paimon-core/src/main/java/org/apache/paimon/table/source/ReadBuilder.java
```diff
@@ -79,6 +80,11 @@ public interface ReadBuilder extends Serializable {
     /** Returns read row type. */
     RowType readType();
 
+    /** Returns actual read row type. */
+    default RowType actualReadType() {
```
JingsongLi: This interface is a bit confusing to me. What is the difference in usage between it and readType?

PR #6686, opened 2025-11-26
[python] support paimon as  ray datasource

XiaoHongbo-Hope:
### Purpose
support paimon as ray datasource

### Tests
`ray_data_test.py`

### API and Format
`TableRead.to_ray`:  Convert Paimon table data to Ray Dataset.

### Documentation
optimize python-api.md

### DIFF
--- a/docs/content/program-api/python-api.md
+++ b/docs/content/program-api/python-api.md
@@ -376,16 +376,7 @@ print(ray_dataset.to_pandas())
 # ...
 ```
 
-#### Parallelism Control
-
-The `to_ray()` method supports a `parallelism` parameter to control distributed reading:
-
-- `parallelism=1` (default): Single-task read (simple mode), suitable for small datasets
-- `parallelism > 1`: Distributed read with specified parallelism, enabling parallel processing across multiple Ray workers
-
-{{< hint info >}}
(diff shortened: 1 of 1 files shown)


paimon-python/pypaimon/read/table_read.py
```diff
@@ -126,10 +126,24 @@ def to_duckdb(self, splits: List[Split], table_name: str,
         con.register(table_name, self.to_arrow(splits))
         return con
 
-    def to_ray(self, splits: List[Split]) -> "ray.data.dataset.Dataset":
+    def to_ray(self, splits: List[Split], use_distributed_read: bool = True) -> "ray.data.dataset.Dataset":
```
chenghuichen: Suggestion: Replace `use_distributed_read: bool` with `parallelism: int =1 `. This way, parallelism=1 indicates a single-task read, and other values can be passed directly to `ray.data.read_datasource(datasource, parallelism)`. `parallelism` is more useful, instead of just serving as a binary switch.

XiaoHongbo-Hope: > Suggestion: Replace `use_distributed_read: bool` with `parallelism: int =1 `. This way, parallelism=1 indicates a single-task read, and other values can be passed directly to `ray.data.read_datasource(datasource, parallelism)`. `parallelism` is more useful, instead of just serving as a binary switch.

Updated

XiaoHongbo-Hope:
> It would be great if test cases for PrimaryKey tables could be added as well.

Added

PR #6690, opened 2025-11-27
[flink] support detecting fields changes for dedicated compaction in streaming mode

LsomeYeah:
### Purpose

Linked issue: close #xxx

If new columns have been added to the table and new files containing these columns have been written while the dedicated compaction job in streaming mode has not been restarted, it may cause data loss. Specifically, the data from the new columns may be lost after compaction.

This pr will detect schema changes for newly added files for dedicated compaction in streaming mode, if schema of files has changed, the `write` will be refreshed.

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/CompactRefresher.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/CompactRefresher.java
@@ -40,7 +40,7 @@ public class CompactRefresher {
 
     private FileStoreTable table;
     private final WriteRefresher refresher;
-    private final ConfigRefresher configRefresher;
+    private final @Nullable ConfigRefresher configRefresher;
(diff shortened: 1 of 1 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/WriterRefresher.java
```diff
@@ -60,6 +64,15 @@ private WriterRefresher(FileStoreTable table, Refresher refresher, Set<String> c
     @Nullable
     public static WriterRefresher create(
             boolean isStreaming, FileStoreTable table, Refresher refresher) {
+        return create(isStreaming, false, table, refresher);
+    }
+
+    @Nullable
+    public static WriterRefresher create(
```
JingsongLi: Introduce a `CompactRefresher` which can wrap `WriterRefresher`.

PR #6692, opened 2025-11-27
[core] Introduce range for GlobalIndexResult and support with ranges filter push down

leaves12138:
Replace all row id with List<Long>/Iterator<Long> to List<Range>

### Purpose

Related pip: https://cwiki.apache.org/confluence/display/PAIMON/PIP-38%3A+Introduce+Global+Index+for+Paimon+Table?src=contextnavpagetreemode

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/globalindex/GlobalIndexBuilder.java
+++ b/paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/globalindex/GlobalIndexBuilder.java
@@ -36,6 +36,7 @@ import org.apache.paimon.utils.Range;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 
 /** This is a class who truly build index file and generate index metas. */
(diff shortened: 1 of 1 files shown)


paimon-common/src/main/java/org/apache/paimon/globalindex/GlobalIndexResult.java
```diff
@@ -18,54 +18,108 @@
...
 /**
- * Global index result represents row ids.
+ * Global index result represents row ids as a list of ranges.
  *
- * <p>TODO introduce ranges interface
+ * <p>Invariant: The list returned by {@link #results()} must be sorted by range start position and
+ * contain no overlapping ranges.
  */
-public interface GlobalIndexResult extends Iterable<Long> {
+public interface GlobalIndexResult {
```
lxy-9602, line 33: RoaringBitmap appears to strike a reasonable balance between query performance and storage efficiency across different use cases—such as range, a small or large number of id—so it might be worth considering a bitmap-based approach as the foundation for GlobalIndexResult.

P.S. RoaringBitmap includes a runOptimized feature that automatically enables run-length encoding when it’s expected to save space.

leaves12138: It's too big, you can try:
        Roaring64NavigableMap roaring64NavigableMap = new Roaring64NavigableMap();
        for (long i = 0; i < 1000000; i++) {
            roaring64NavigableMap.add(i);
        }

        System.out.println(roaring64NavigableMap.serializedSizeInBytes());
  It returns 131217 means 131KB.
Range just two longs with class head.

leaves12138: Every long value in roaring64NavigableMap, it takes place

leaves12138: roaring64NavigableMap.runOptimize takes time

leaves12138: It will visit all cardinality:
image

lxy-9602: I’m not very familiar with the performance of the Java version of RoaringBitmap, but in my local C++ tests, constructing a bitmap with 1,000,000 consecutive row IDs using the addRange() API results in a serialized size of only about 200 bytes, and range-based intersection operations typically complete in the microsecond (μs) range.

PR #6697, opened 2025-11-28
[spark] paimon-spark supports row id push down

Kkkaneki-k:
### Purpose

This PR is about to support row id push down for spark, following https://github.com/apache/paimon/pull/6483

Linked issue: None

### Tests

org.apache.paimon.spark.sql.RowIdPushDownTestBase

### DIFF
--- a/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
@@ -2146,6 +2146,14 @@ public class CoreOptions implements Serializable {
                     .withDescription(
                             "Whether to try upgrading the data files after overwriting a primary key table.");
 
+    public static final ConfigOption<Boolean> ROW_ID_PUSH_DOWN_ENABLED =
+            key("row-id-push-down.enabled")
+                    .booleanType()
+                    .defaultValue(false)
+                    .withDescription(
+                            "Whether to enable row id push down for scan."
+                                    + " Currently, only the data evolution table supports row id push down.");
+
     private final Options options;
 
     public CoreOptions(Map<String, String> options) {
@@ -3328,6 +3336,10 @@ public class CoreOptions implements Serializable {
 
     public boolean overwriteUpgrade() {
         return options.get(OVERWRITE_UPGRADE);
+    }
+
+    public boolean rowIdPushDownEnabled() {
+        return options.get(ROW_ID_PUSH_DOWN_ENABLED);
     }
 
     /** Specifies the merge engine for table with primary key. */
--- a/paimon-common/src/main/java/org/apache/paimon/predicate/RowIdPredicateVisitor.java
+++ b/paimon-common/src/main/java/org/apache/paimon/predicate/RowIdPredicateVisitor.java
@@ -1,0 +1,109 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.paimon.predicate;
+
+import org.apache.paimon.utils.Range;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.apache.paimon.table.SpecialFields.ROW_ID;
+
+/**
+ * The {@link PredicateVisitor} to extract a list of row id ranges from predicates. The returned row
+ * id ranges can be pushed down to manifest readers and file readers to enable efficient random
+ * access.
+ *
+ * <p>Note that there is a significant distinction between returning {@code null} and returning an
+ * empty list:
+ *
+ * <ul>
+ *   <li>{@code null} indicates that the predicate cannot be converted into a random-access pattern,
+ *       meaning the filter is not consumable by this visitor.
+ *   <li>An empty list indicates that no rows satisfy the predicate (e.g. {@code WHERE _ROW_ID = 3
+ *       AND _ROW_ID IN (1, 2)}).
+ * </ul>
+ */
+public class RowIdPredicateVisitor implements PredicateVisitor<List<Range>> {
+
+    @Override
+    public List<Range> visit(LeafPredicate predicate) {
+        if (ROW_ID.name().equals(predicate.fieldName())) {
+            LeafFunction function = predicate.function();
+            if (function instanceof Equal || function instanceof In) {
+                ArrayList<Long> rowIds = new ArrayList<>();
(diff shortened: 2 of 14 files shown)


Kkkaneki-k:
> It seems your modification breaks some tests.

Yes, I am working on it.

JingsongLi:
Thanks @Kkkaneki-k

Kkkaneki-k:
> Hello, now row id pushdown only work for data evolution table. Can you continue to finish this?

OK, I will continue to complete this PR.

Kkkaneki-k:
@JingsongLi I've finished the modifications. PTAL if you have some time, thanks!

JingsongLi:
@Kkkaneki-k Can you rebase latest master?

paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -2112,6 +2112,14 @@ public InlineElement getDescription() {
                             "The maximum number of concurrent scanner for global index."
                                     + "By default is the number of processors available to the Java virtual machine.");
 
+    public static final ConfigOption<Boolean> ROW_ID_PUSH_DOWN_ENABLED =
+            key("row-id-push-down.enabled")
```
JingsongLi, line 2150: Remove this option, it is useless.

paimon-spark/paimon-spark-3.3/src/main/scala/org/apache/paimon/spark/PaimonScan.scala
```diff
@@ -29,12 +29,15 @@ import org.apache.spark.sql.connector.read.partitioning.{KeyGroupedPartitioning,
...
 
+import java.lang.{Long => JLong}
+
 import scala.collection.JavaConverters._
 
 case class PaimonScan(
     table: InnerTable,
     requiredSchema: StructType,
     filters: Seq[Predicate],
+    override val rowIds: Seq[JLong],
```
JingsongLi: rowIds -> pushedRowIds

paimon-common/src/main/java/org/apache/paimon/predicate/RowIdPredicateVisitor.java
```diff
@@ -0,0 +1,103 @@
...
+ * empty set:
+ *
+ * <ul>
+ *   <li>{@code null} indicates that the predicate cannot be converted into a random-access pattern,
+ *       meaning the filter is not consumable by this visitor.
+ *   <li>An empty set indicates that no rows satisfy the predicate (e.g. {@code WHERE _ROW_ID = 3
+ *       AND _ROW_ID IN (1, 2)}).
+ * </ul>
+ */
+public class RowIdPredicateVisitor implements PredicateVisitor<Set<Long>> {
```
JingsongLi: Can we return `Set<Range>`? In this way, we can support min max pushdown.

paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
Kkkaneki-k: > Remove this option, it is useless.

I think this option is necessary, it disables rowId pushdown for non-data-evolution tables. Currently, due to the issue in #6747, rowId pushdown should not be enabled for non-data-evolution tables.

paimon-spark/paimon-spark-3.3/src/main/scala/org/apache/paimon/spark/PaimonScan.scala
Kkkaneki-k: > rowIds -> pushedRowIds

done

paimon-common/src/main/java/org/apache/paimon/predicate/RowIdPredicateVisitor.java
Kkkaneki-k: > Can we return `Set<Range>`? In this way, we can support min max pushdown.

done

paimon-spark/paimon-spark-3.2/src/main/scala/org/apache/paimon/spark/PaimonBaseScanBuilder.scala
```diff
@@ -71,7 +77,28 @@ abstract class PaimonBaseScanBuilder
       filter =>
         val predicate = converter.convertIgnoreFailure(filter)
         if (predicate == null) {
-          postScan.append(filter)
+          val rowTypeWithRowId = new RowType(
+            false,
+            Collections.singletonList(new DataField(-1, ROW_ID.name(), DataTypes.BIGINT())))
+          val converterWithRowId = new SparkFilterConverter(rowTypeWithRowId)
+          val newPredicate = converterWithRowId.convertIgnoreFailure(filter)
```
Zouxxyy: Can we pass the filter containing RowId into PaimonScanBuilder, and let Paimon Core parse it and generate the corresponding range itself? It feels redundant to reimplement this logic in every engine.

Kkkaneki-k: > Can we pass the filter containing RowId into PaimonScanBuilder, and let Paimon Core parse it and generate the corresponding range itself? It feels redundant to reimplement this logic in every engine.

Thanks for your review! I've carefully considered your suggestion and think that this change might introduce the following two problems:
1. When a filter containing `_ROW_ID` cannot be consumed, we need to return it to the engine as a post-scan filter. This may be difficult to achieve if Paimon Core itself consumes it and generates the corresponding range.
2. Currently, ReadBuilder requires separate inputs of filters containing `_ROW_ID` and filters without `_ROW_ID` during the build process. This means we need to differentiate between these two types of filters in the engine and input them separately (unless we modify ReadBuilder to handle this differentiation automatically during the build process).

Zouxxyy: 1. It should pass the filter containing _ROW_ID as `pushedDataFilters` to Paimon, while still adding it to the postFilter for Spark to handle. Therefore, almost no changes are needed in Paimon's Spark connector—except possibly update `new SparkFilterConverter(rowType)` with requiredSchema or rowType with row id.

2. Yes, that’s exactly where I intend to put it. CC @JingsongLi 

ReadBuilderImpl:

```java
private InnerTableScan configureScan(InnerTableScan scan) {
  scan.withFilter(filter)
          .withReadType(readType)
          .withPartitionFilter(partitionFilter)
          // calculate rowRanges from filter
          .withRowRanges(rowRanges);
}
```

I believe users would much prefer writing filters that include _ROW_ID when using the ReadBuilder API, rather than having to understand what List<Range> rowRanges is and convert to it.

Kkkaneki-k: > 1. It should pass the filter containing _ROW_ID as `pushedDataFilters` to Paimon, while still adding it to the postFilter for Spark to handle. Therefore, almost no changes are needed in Paimon's Spark connector—except possibly update `new SparkFilterConverter(rowType)` with requiredSchema or rowType with row id.
> 2. Yes, that’s exactly where I intend to put it. CC @JingsongLi
> 
> ReadBuilderImpl:
> 
> ```java
> private InnerTableScan configureScan(InnerTableScan scan) {
>   scan.withFilter(filter)
>           .withReadType(readType)
>           .withPartitionFilter(partitionFilter)
>           // calculate rowRanges from filter
>           .withRowRanges(rowRanges);
> }
> ```
> 
> I believe users would much prefer writing filters that include _ROW_ID when using the ReadBuilder API, rather than having to understand what List rowRanges is and convert to it.

Thanks for your reply! I think your point is correct, and I will modify my code based on your suggestions.

PR #6699, opened 2025-11-28
[core] Fix incremental query unable to read changelog when batchScanSkipLevel0 is enabled

### LINKED ISSUE #6698: [Bug] Incremental query failing to read changelog when batchScanSkipLevel0 is true  and changelog producer is set to 'input'
### Search before asking

### Paimon version

master commit id:e14a8111

### Compute Engine

all

### Minimal reproduce step

paimon table
```
CREATE TABLE T (
a INT, 
b INT, 
PRIMARY KEY (a) NOT ENFORCED
) WITH (
    'changelog-producer' = 'lookup',
    'bucket' = '1',
    'deletion-vectors.enabled' = 'true'
)

INSERT INTO T VALUES (1, 2), (2, 2), (3, 3);

SELECT * FROM T /*+ OPTIONS( 'scan.mode' = 'incremental',  'incremental-between' = '0,1'*/;

```
The result is EMPTY

[+I[1, 2], +I[2, 2], +I[3, 3]]

### Anything else?

_No response_

JQ-Cao:
### Purpose

Linked issue: close #6698

If the table is dv enabled or using first-row as merge engine, DataTableBatchScan filters out records with level = 0. This breaks incremental queries using the input changelog producer, as changelog entries are incorrectly excluded. This change ensures changelog records are preserved in such scenarios.

### Tests

IncrementalTableTest#testIncrementalWithInputChangeLogAndMoW: Added in the PR.

### API and Format

No API changes.
No storage format changes.

### Documentation

None

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/mergetree/MergeTreeWriter.java
+++ b/paimon-core/src/main/java/org/apache/paimon/mergetree/MergeTreeWriter.java
@@ -215,7 +215,7 @@ public class MergeTreeWriter implements RecordWriter<KeyValue>, MemoryOwner {
 
             final RollingFileWriter<KeyValue, DataFileMeta> changelogWriter =
                     changelogProducer == ChangelogProducer.INPUT
-                            ? writerFactory.createRollingChangelogFileWriter(0)
+                            ? writerFactory.createRollingChangelogFileWriter(1)
                             : null;
             final RollingFileWriter<KeyValue, DataFileMeta> dataWriter =
                     writerFactory.createRollingMergeTreeFileWriter(0, FileSource.APPEND);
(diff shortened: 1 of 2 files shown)


JingsongLi (commented):
Important bug fix for incremental changelog reading with `batchScanSkipLevel0` enabled.

**Root cause**: When DV is enabled or first-row merge engine is used, `DataTableBatchScan` filters out level-0 records. The changelog writer was using level 0 for INPUT changelog files, causing them to be incorrectly excluded from incremental queries.

**Fix**: Change changelog file level from 0 to 1 in `MergeTreeWriter.flushWriteBuffer()`:
```java
- writerFactory.createRollingChangelogFileWriter(0)
+ writerFactory.createRollingChangelogFileWriter(1)
```

**Review**:
1. The fix is minimal and targeted - single line change in production code.
2. The test `testIncrementalWithInputChangeLogAndMoW` correctly reproduces the scenario: DV-enabled table with INPUT changelog producer, verifying that changelog is readable after compaction.
3. **Question**: Does changing the changelog level from 0 to 1 have any other side effects? Level is typically used for compaction priority - does level 1 for changelog files affect any compaction decisions or other level-based filtering logic?
4. The test covers the key path: write → compact → incremental read. Good coverage.

LGTM - clean fix with good test coverage. Just verify there are no other places that filter by level that could be affected.

JingsongLi:
The right solution should be fixing incremental query to not skip level 0.

PR #6706, opened 2025-11-28
[flink-cdc] Fix partition key conflicts and support append-only parti…

### LINKED ISSUE #6705: [Bug] PaimonMetadataApplier prevents partitioning in append-only tables


### Paimon version

1.20.0

### Compute Engine

Flink 1.20.0

### Minimal reproduce step

1. Set up Flink CDC pipeline with MySQL source and Paimon sink
2. Source table has no primary key                                                                                                                                                                                                                                                                    
3. Configure Paimon sink with partition key.                                                                                                                                                                                                                                                                 
4. Start the pipeline                                                                                                                                                                                                                                                                                                       
5. Pipeline fails when creating table with error:                                                                                                                                                                                                                                                                           
   "Primary key constraint [partition_key] should not be same with partition fields [partition_key]"  

PaimonMetadataApplier.applyCreateTable() (lines 199-203) automatically adds 
partition keys to primary keys without checking if this violates Paimon's own                                                                                                                                                                                                                                               
constraint.                                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                            
Expected: Should allow partitioning by primary key, or at minimum not modify                                                                                                                                                                                                                                                
user's schema in a way that causes validation failure.                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                            
Actual: Code forces partition keys into primary keys, then Paimon's                                                                                                                                                                                                                                                         
TableSchema.trimmedPrimaryKeys() rejects it because trimmed primary keys                                                                                                                                                                                                                                                    
would be empty.                                                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                            
This makes it impossible to partition by the same field as primary key in                                                                                                                                                                                                                                                   
CDC scenarios.  

### Anything else?


Root cause code in PaimonMetadataApplier.java lines 199-203:
                                                                                                                                                                                                                                                                                                                            
for (String partitionColumn : partitionKeys) {                                                                                                                                                                                                                                                                              
    if (!primaryKeys.contains(partitionColumn)) {                                                                                                                                                                                                                                                                           
        primaryKeys.add(partitionColumn);                                                                                                                                                                                                                                                                                   
    }                                                                                                                                                                                                                                                                                                                       
}                                                                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                            
This logic always assumes partition keys must always be in primary keys even in case of append only tables.                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                            
Issues:                                                                                                                                                                                                                                                                                                                     
1. Breaks when partition key equals primary key (current bug)                                                                                                                                                                                                                                                               
2. Prevents append-only tables (no primary key) from having partitions                                                                                                                                                                                                                                                      
3. Modifies user's explicit schema definition without consent                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                            
Suggested fix:                                                                                                                                                                                                                                                                                                              
- Remove automatic modification, OR                                                                                                                                                                                                                                                                                         
- Only apply for bucketed tables with composite primary keys, OR                                                                                                                                                                                                                                                            
- Skip this logic when primaryKeys is empty (append-only tables)                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                            
Append-only tables SHOULD support partitioning without requiring primary keys.                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                            
Workaround: Currently must use different fields for partition and primary key,                                                                                                                                                                                                                                            or avoid partitioning in append-only tables.    

prashant8530:
### Purpose 
Linked issue: close #6705                                                                                                         
                                                                                                                                                                                                                          
Fix `PaimonMetadataApplier` to support append-only partitioned tables and prevent validation failures when partition key equals primary key.                                                                              
                                                                                                                                                                                                                          
**Problem:**                                                                                                                                                                                                              
`PaimonMetadataApplier.applyCreateTable()` unconditionally adds partition keys to primary keys, causing:                                                                                                                  
1. Append-only tables (no primary key) cannot have partitions                                                                                                                                                             
2. When `partitionKeys = primaryKeys`, validation fails with: "Primary key constraint should not be same with partition fields"                                                                                           
                                                                                                                                                                                                                          
**Solution:**                                                                                                                                                                                                             
Only add partition keys to primary keys when safe (non-empty PK and no conflict). This preserves existing behavior for composite keys while enabling append-only and partition-by-PK scenarios.                           
                                                                                                                                                                                                                          
### Tests                                                                                                                                                                                                                 
                                                                                                                                                                                                                          
Existing tests validate the fix doesn't break current behavior. The change makes behavior less restrictive by allowing previously invalid schemas.                                                                        
                                                                                                                                                                                                                          
### API and Format                                                                                                                                                                                                        
                                                                                                                                                                                                                          
No API or storage format changes. This is a bug fix in schema creation logic.                                                                                                                                             
                                                                                                                                                                                                                          
### Documentation                                                                                                                                                                                                         
                                                                                                                                                                                                                          
No new feature introduced. This fixes existing functionality to work as expected.

### DIFF
--- a/paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/pipeline/cdc/schema/PaimonMetadataApplier.java
+++ b/paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/pipeline/cdc/schema/PaimonMetadataApplier.java
@@ -51,6 +51,7 @@ import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -189,15 +190,18 @@ public class PaimonMetadataApplier implements MetadataApplier {
                                                             .getLogicalType()),
                                             column.getComment()));
             List<String> partitionKeys = new ArrayList<>();
-            List<String> primaryKeys = schema.primaryKeys();
+            List<String> primaryKeys = new ArrayList<>(schema.primaryKeys());
             if (partitionMaps.containsKey(event.tableId())) {
                 partitionKeys.addAll(partitionMaps.get(event.tableId()));
             } else if (schema.partitionKeys() != null && !schema.partitionKeys().isEmpty()) {
                 partitionKeys.addAll(schema.partitionKeys());
             }
-            for (String partitionColumn : partitionKeys) {
-                if (!primaryKeys.contains(partitionColumn)) {
-                    primaryKeys.add(partitionColumn);
+            // Only add partition keys to primary keys if primary keys exist and won't cause conflicts
+            if (!primaryKeys.isEmpty() && !new HashSet<>(primaryKeys).equals(new HashSet<>(partitionKeys))) {
+                for (String partitionColumn : partitionKeys) {
+                    if (!primaryKeys.contains(partitionColumn)) {
+                        primaryKeys.add(partitionColumn);
+                    }
                 }
             }
             builder.primaryKey(primaryKeys)
--- a/paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/pipeline/cdc/schema/PaimonMetadataApplier.java
+++ b/paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/pipeline/cdc/schema/PaimonMetadataApplier.java
@@ -51,7 +51,6 @@ import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


JingsongLi (commented):
Good bug fix for PaimonMetadataApplier. The change correctly handles two cases:

1. **Append-only partitioned tables**: When `primaryKeys` is empty, the old code would still add partition keys to the (empty) primary key list, effectively converting an append-only table into a primary key table. The fix correctly skips this logic when `primaryKeys.isEmpty()`.

2. **Mutable primary keys list**: Changed from `schema.primaryKeys()` (potentially immutable) to `new ArrayList<>(schema.primaryKeys())` to safely allow modifications.

The fix is minimal (+7/-4), well-targeted, and the logic is clear. The guard `if (!primaryKeys.isEmpty())` is the right condition.

One suggestion: consider adding a test case in the CDC test suite that creates an append-only partitioned table via the pipeline to prevent regression.

LGTM.

JingsongLi:
+1

PR #6715, opened 2025-12-01
[python/doc] Add options supported by the python API

discivigour:
### Purpose

- add options supported by the python API.


JingsongLi (commented):
We should document supported or unsupported table mode. Instead of options.

PR #6720, opened 2025-12-01
[variant] Extract only required columns when reading shredded variants

Zouxxyy:
### Purpose

Subtask of https://github.com/apache/paimon/issues/4471, we can read only the necessary fields when reading shredded variants


JingsongLi:
+1

PR #6722, opened 2025-12-02
[spark] Update the display of size and timing metrics

Zouxxyy:
### Purpose

use PaimonSizeSummaryMetric for size based summary metric
use PaimonTimingSummaryMetric for timming based summary metric

image


JingsongLi:
+1

PR #6723, opened 2025-12-02
[core] Append commit should check bucket number if latest commit user is different

yuzelin:
### Purpose

The case is, if user1 batch write the postpone table T, and user2 compact and rescale T. If they start the job at the same time and user2 commits first, the user1's append commit should also be checked because the bucket number might be changed.


Zouxxyy (approved):
+1

zhuangchong (approved):
+1

PR #6724, opened 2025-12-02
[core] ShardScanner should not keep partition parameter

leaves12138:
### Purpose


lxy-9602:
+1

PR #6725, opened 2025-12-02
[core] Intruduce bitmap64 for GlobalIndexResult and support with ranges filter push down

leaves12138:
### Purpose

We use bitmap to store Iterator<Long>.

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/utils/RoaringNavigableMap64.java
+++ b/paimon-common/src/main/java/org/apache/paimon/utils/RoaringNavigableMap64.java
@@ -35,11 +35,6 @@ import java.util.Objects;
 /** A compressed bitmap for 64-bit integer aggregated by tree. */
 public class RoaringNavigableMap64 {
 
-    static {
-        Roaring64NavigableMap.SERIALIZATION_MODE =
-                Roaring64NavigableMap.SERIALIZATION_MODE_PORTABLE;
-    }
-
     private final Roaring64NavigableMap roaring64NavigableMap;
 
     public RoaringNavigableMap64() {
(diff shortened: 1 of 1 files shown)


paimon-common/src/main/java/org/apache/paimon/utils/RoaringNavigableMap64.java
```diff
@@ -0,0 +1,169 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Objects;
+
+/** A compressed bitmap for 64-bit integer aggregated by tree. */
+public class RoaringNavigableMap64 {
+
+    private final Roaring64NavigableMap roaring64NavigableMap;
```
lxy-9602, line 38: May need to set the serialization mode to portable for cross-language compatibility:

```java
Roaring64NavigableMap.SERIALIZATION_MODE = Roaring64NavigableMap.SERIALIZATION_MODE_PORTABLE;
```

ref：[compatibility issue](https://github.com/RoaringBitmap/RoaringBitmap/issues/366) and [compatibility pr](https://github.com/RoaringBitmap/RoaringBitmap/pull/598)

leaves12138: OK

JingsongLi: Why not just use `serializePortable`.

leaves12138: Reasonable

PR #6728, opened 2025-12-02
[flink] Flink batch job support specify partition with max_pt() and max_two_pt()

yuzelin:
### Purpose

Support `scan.partitions = max_pt()` in flink batch job.

### DIFF
--- a/docs/layouts/shortcodes/generated/flink_connector_configuration.html
+++ b/docs/layouts/shortcodes/generated/flink_connector_configuration.html
@@ -180,7 +180,7 @@ under the License.
             <td><h5>scan.partitions</h5></td>
             <td style="word-wrap: break-word;">(none)</td>
             <td>String</td>
-            <td>Specify the partitions to scan. Partitions should be given in the form of key1=value1,key2=value2. Partition keys not specified will be filled with the value of partition.default-name. Multiple partitions should be separated by semicolon (;). This option can support normal source tables and lookup join tables. For lookup joins, two special values max_pt() and max_two_pt() are also supported, specifying the (two) partition(s) with the largest partition value.</td>
+            <td>Specify the partitions to scan. Partitions should be given in the form of key1=value1,key2=value2. Partition keys not specified will be filled with the value of partition.default-name. Multiple partitions should be separated by semicolon (;). This option can support normal source tables and lookup join tables. There are two special values max_pt() and max_two_pt() are also supported to specify the (two) partition(s) with the largest partition value. For lookup source, the max partition(s) will be periodically refreshed; for normal source, the max partition(s) will be determined before job running without refreshing even for streaming jobs.</td>
         </tr>
         <tr>
             <td><h5>scan.remove-normalize</h5></td>


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/FlinkTableSource.java
```diff
@@ -135,45 +139,59 @@ public Result applyFilters(List<ResolvedExpression> filters) {
         return Result.of(filters, unConsumedFilters);
     }
 
+    private PartitionPredicate getPartitionPredicate() {
```
yunfengzhou-hub: Methods `getPartitionPredicate` and `getPartitionPredicateWithOptions` seems not differ by options. `getPartitionPredicateWithOptions` is only invoked by `getPartitionPredicate`, and `getPartitionPredicate` contains codes that were previously located in `getPartitionPredicateWithOptions`. Thus it might be better to merge these two methods back to simplify commit history, or rename these two methods to better reflect their difference.

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/BatchFileStoreITCase.java
```diff
@@ -902,6 +902,46 @@ public void testScanWithSpecifiedPartitionsWithFieldMapping() {
...
+    public void testScanWithSpecifiedPartitionsWithLevelMaxPt() throws Exception {
+        sql(
+                "CREATE TABLE P (id INT, v INT, pt1 STRING, pt2 STRING, pt3 STRING) PARTITIONED BY (pt1, pt2, pt3)");
+        sql("CREATE TABLE Q (id INT)");
+        sql(
+                "INSERT INTO P VALUES (1, 10, 'a', '2025-10-01', '1'), (2, 20, 'a', '2025-10-01', '2'), (3, 30, 'a', '2025-10-02', '1'), (4, 40, 'a', '2025-10-02', '2'), "
+                        + "(1, 11, 'b', '2025-10-01', '1'), (2, 21, 'b', '2025-10-01', '2'), (3, 31, 'b', '2025-10-02', '1'), (4, 41, 'b', '2025-10-02', '2')");
+        sql("INSERT INTO Q VALUES (1), (2), (3), (4)");
+        String query =
+                "SELECT Q.id, P.v FROM Q INNER JOIN P /*+ OPTIONS('scan.partitions' = 'pt1=max_pt(),pt2=max_pt()') */ ON Q.id = P.id ORDER BY Q.id, P.v";
```
yunfengzhou-hub: According offline discussions, the initial purpose of batch job users is to use lookup join, but it will be optimized to inner join in Flink batch mode. Thus how about directly testing the lookup join queries in these test cases, to have tests cover the Flink join optimization logic?

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/BatchFileStoreITCase.java
yunfengzhou-hub, line 1: Let's also update the description of the configuration `scan.partitions` to show the following information:

* max_pt can be used on scan sources(or "normal sources" if more proper) due to flink batch job optimizations for lookup join. But it is not recommended to use it in this way for streaming jobs, as the max_pt will not be updated.

yunfengzhou-hub (commented):
Thanks for the PR. Left some comments as below

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lookup/PartitionLoader.java
```diff
@@ -76,7 +76,7 @@ public void addPartitionKeysTo(List<String> joinKeys, List<String> projectFields
         partitionKeys.stream().filter(k -> !projectFields.contains(k)).forEach(projectFields::add);
     }
 
-    public Predicate createSpecificPartFilter() {
+    public Predicate createSpecificPartFilter(List<BinaryRow> partitions) {
```
yunfengzhou-hub: It might be better to make this method static, or further move this method to some utility class.

PR #6730, opened 2025-12-03
[core] format table: throw exception when get partiiton from file system

jerry-024:
### Purpose

Linked issue: close #xxx


JingsongLi:
+1

Issue #6735, opened 2025-12-03
[Feature] PyPaimon Real RowKind Support


kaori-seasons:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### 1.1 Problem Description

The PyPaimon `KeyValueDataWriter` was hardcoding all rows' `_VALUE_KIND` to 0 (INSERT), losing information about different row operation types. This prevented proper handling of CDC (Change Data Capture) scenarios with UPDATE and DELETE operations.

**Example Scenario:**
- Write operation 1: User 'Alice' with RowKind=INSERT
- Write operation 2: User 'Alice' updated to 'Alice_updated' with RowKind=UPDATE_AFTER
- Write operation 3: User 'Alice' deleted with RowKind=DELETE
- **Problem**: All three records stored with _VALUE_KIND=0 (INSERT)
- **Expected**: Each record should have correct _VALUE_KIND (0, 2, 3)

### 1.2 Root Cause Analysis

**Write-side Issue**: The `_add_system_fields()` method in `KeyValueDataWriter` hardcoded `_VALUE_KIND` values:

```python
# Before: Always INSERT
value_kind_column = pa.array([0] * num_rows, type=pa.int32())
```

**Impact Areas:**
- **CDC scenarios**: Cannot distinguish UPDATE_BEFORE/UPDATE_AFTER operations
- **Delete handling**: Cannot mark DELETE rows (RowKind=3)
- **Read-side consistency**: Read-side `DropDeleteReader` depends on correct RowKind values
- **Data semantics**: Loss of operational metadata

### 1.3 RowKind Definition

PyPaimon defines four RowKind types:
- **INSERT (0)**: +I - New record insertion
- **UPDATE_BEFORE (1)**: -U - Previous value before update
- **UPDATE_AFTER (2)**: +U - New value after update
- **DELETE (3)**: -D - Record deletion

### 2.1 Core Design Approach

Implement real RowKind extraction from input data by:
1. Reading optional `__row_kind__` column from input RecordBatch
2. Validating RowKind values (0-3 range)
3. Using extracted values for `_VALUE_KIND` field
4. Maintaining backward compatibility (defaulting to INSERT)

### 2.2 Solution Advantages

1. **Full CDC Support**: Correctly handles all row operation types
2. **Backward Compatible**: No breaking changes to existing API
3. **Type-Safe**: Validates data types and value ranges
[...]

PR #6736, opened 2025-12-03
enhance: support row kind

kaori-seasons:
### Purpose

Related to [issue-6735](https://github.com/apache/paimon/issues/6735)


JingsongLi (commented):
Adding row kind support to the key-value writer in pypaimon. Review:

1. **Purpose**: Supporting RowKind (INSERT, UPDATE_BEFORE, UPDATE_AFTER, DELETE) is essential for CDC and changelog semantics. Good addition.

2. **Implementation scope**: +330 lines across 3 files is reasonable for this feature.

3. **Testing**: Good that there's a dedicated test file (`key_value_writer_row_kind_test.py`). Ensure it covers all RowKind variants and validates the serialized output format.

4. **PR description**: Like the other PRs from this author, the description is very sparse. Please add details about the API surface and how row kinds are represented in the write path.

5. **Compatibility**: Ensure the row kind encoding matches what Java Paimon expects when reading these files.

Looks reasonable in scope. Would benefit from a more detailed description.

TakodaS:
Why was this closed? This is a critical missing feature from pypaimon without which performing deletions on large tables is very expensive.

Issue #6734, opened 2025-12-03
[Feature] Introduce SST File Format base on SortLookupStore


steFaiz:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Motivation

The SST file format can be used in the following scenarios:
1. As the local storage format for LookupFile.
2. In the B-tree global index.
3. When query performance is a higher priority—for example, when used as a Flink dimension table—the data can be directly stored in SST format.

### Solution

Based on current LookupStore implementation, we can introduce a generalized SST FileFormat which can store any binary key-value pairs. The file format is expected to provide below functionality：
1. fast lookup
2. fast range query
3. fast random access
4. efficient full-file scan

### Anything else?

Please refer to [pip-25]( https://cwiki.apache.org/confluence/display/PAIMON/PIP-25%3A+Introduce+a+key-value+file+format+for+paimon+primary+key+table) as the related proposal and refer to [PR](https://github.com/apache/paimon/pull/6702) as current draft implementation.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #6738, opened 2025-12-03
[core] Fix BlockIterator may not move forward after calling `next()`

steFaiz:
### Purpose
This PR is a part of https://github.com/apache/paimon/issues/6734
#### Problem
Currently, the `BlockIterator` can't deal with 'seek + scan' scenario correctly. The bug can be illustrated as below, assuming that we are seeking to 25:
When the `MID` pointer pointing at the block exactly smaller than 25, it works normally: 
```
values:   ... 10, 20, 30, 40, 50 ...
                   |    
                   V    
                  MID 
           
At that case, the final pointers' positions are：

                       Iter Pos
                    Polled ∆
                       ∆   |
                       |   | 
values:   ... 10, 20, 30, 40, 50 ...
                   |   |
                   V   V
                 HIGH MID
                      LOW
```
However, when the `MID` pointer pointing at the block exactly greater than 25, some problems will happen:
```
values:    10, 20, 30, 40, 50
                    |   
                    V   
                   MID 
           
At that case, the final pointers' positions are：

                 Iter Pos
                  Polled 
                    ∆   
                    |   
values:    10, 20, 30, 40, 50
                |   |
                V   V
               MID LOW
              HIGH
```
The `Polled` pointer and the underlying `MemorySliceInput`'s pointer are pointing at the same place, which means the next calling of `next()` method won't move iterator position forward.

#### Resolution
The resolution is quite easy, the target position is always:
1. if found an equal key, the `MID` pointer's position
2. else, the `LEFT` pointer's position

So we just need to find the target position, then seek to the target position.

### Tests
Please see org.apache.paimon.lookup.sort.BlockIteratorTest

### API and Format
None

### Documentation
None

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/lookup/sort/BlockIterator.java
+++ b/paimon-common/src/main/java/org/apache/paimon/lookup/sort/BlockIterator.java
@@ -36,6 +36,8 @@ public abstract class BlockIterator implements Iterator<Map.Entry<MemorySlice, M
     private final int recordCount;
     private final Comparator<MemorySlice> comparator;
 
+    private BlockEntry polled;
+
     public BlockIterator(
             MemorySliceInput data, int recordCount, Comparator<MemorySlice> comparator) {
         this.data = data;
@@ -45,13 +47,19 @@ public abstract class BlockIterator implements Iterator<Map.Entry<MemorySlice, M
 
     @Override
     public boolean hasNext() {
-        return data.isReadable();
+        return polled != null || data.isReadable();
     }
 
     @Override
     public BlockEntry next() {
         if (!hasNext()) {
             throw new NoSuchElementException();
+        }
+
+        if (polled != null) {
+            BlockEntry result = polled;
+            polled = null;
+            return result;
         }
 
         return readEntry();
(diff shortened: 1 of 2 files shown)


paimon-common/src/main/java/org/apache/paimon/lookup/sort/BlockIterator.java
```diff
@@ -70,32 +62,58 @@ public void remove() {
...
 
             if (compare == 0) {
-                polled = midEntry;
-                return true;
+                break;
             } else if (compare > 0) {
-                polled = midEntry;
                 right = mid - 1;
             } else {
                 left = mid + 1;
```
JingsongLi, line 92: Why not just modify here: polled = null;

steFaiz: ## **Quick Answer**
It can't pass the Test QAQ
image
## **Reason**
if we just let `polled` = null here, it means that if we are seeking to some value exactly not present in current block, the `polled` will always be null. At that case, the next iterator position after seek will always be the position after the last `MID` position (because we will seek to mid and call readNext for each loop).
Let's recall the example before:
First case: 
* At that case, the target element is skipped!
```
                       Iter Pos
                           ∆
                           |
                           | 
values:   ... 10, 20, 30, 40, 50 ...
                   |   |
                   V   V
                 HIGH MID
                      LOW
```
* the test result also prove that: the actual value is greater than our expectation
image

Second case:
* At that case, the iteration process is expected.
```
                 Iter Pos
                    ∆   
                    |   
values:    10, 20, 30, 40, 50
                |   |
                V   V
               MID LOW
              HIGH
```
## **Thinking**
I think if we really want to keep the `polled` field, which is always the target element (if has some), we need to seek to the next position of polled before returning. I think it will be equilvalent to current implementation. And removing `polled` field makes the code more clean and readable.

PR #6741, opened 2025-12-03
[core] Make bucketed append table write initial lighter

JingsongLi:
### Purpose

In the past, we had to forcefully assign sequenceNumbers to the append table of bucketed, which made it very difficult to initialize the bucket table during write operations, even in write only mode.

We have modified it to only compare the file creation time to avoid difficulties in writing initialization.

### Tests

- `AppendOnlySimpleTableTest.testBucketedAppendTableWriteWithInit`
- `AppendOnlySimpleTableTest.testBucketedAppendTableWriteNoInit`

### DIFF
--- a/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
@@ -2929,6 +2929,10 @@ public class CoreOptions implements Serializable {
         return consumerId;
     }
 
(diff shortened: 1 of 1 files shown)


JingsongLi:
TODO: only works when `bucket-append-ordered` is false.

Aitozi:
The `bucket-append-ordered=false` would skip the bucket checker which would lead to unexpected error when user rescale the bucket number @JingsongLi

PR #6743, opened 2025-12-04
[spark] Reduce redundant loadTable

summaryzb:
### Purpose
Previously `modifyPaimonTable` and `createRelation` in `BaseProcedure` both loadTable, this should be optimized by reduce redundant loadTable action

### Tests
pass existing test

### API and Format
No

### Documentation
No

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/procedure/BaseProcedure.java
+++ b/paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/procedure/BaseProcedure.java
@@ -76,22 +76,17 @@ abstract class BaseProcedure implements Procedure {
 
     protected <T> T modifyPaimonTable(
             Identifier ident, Function<org.apache.paimon.table.Table, T> func) {
-        return execute(ident, true, func);
-    }
-
-    private <T> T execute(
-            Identifier ident,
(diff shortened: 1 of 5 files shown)


JingsongLi (commented):
Good optimization to reduce redundant `loadTable` calls in Spark procedures.

**Changes:**
1. `modifyPaimonTable` now loads the SparkTable once and extracts the Paimon table from it.
2. New `modifySparkTable` method gives procedures access to the SparkTable directly (needed when `createRelation` is called inside the lambda).
3. `createRelation(Identifier, Table)` overload avoids re-loading the table for relation creation.
4. Procedures (Compact, CreateGlobalIndex, Rescale, RewriteFileIndex) updated to use `modifySparkTable` and pass the already-loaded table to `createRelation`.

**Review:**
- The refactoring is clean. Each procedure that previously called `createRelation(tableIdent)` (which loaded the table again) now passes the already-loaded `sparkTable` directly.
- +29/-28 — nearly the same line count, just reorganized. Good.
- No functional change, purely performance (avoids catalog round-trips for each procedure call).

Minor: The `RescaleProcedure` change also refactors the `execute()` method signature to take `DataSourceV2Relation` instead of `Identifier` — this is a nice coupling reduction.

LGTM.

JingsongLi:
+1

PR #6745, opened 2025-12-04
[spark] Merge into supports _ROW_ID shortcut

wayneli-vt:
### Purpose

This PR enhances the `MERGE INTO` command by adding a specialized execution path for `_ROW_ID`-based joins.

Currently, when performing a `MERGE INTO` operation, the process to find the relevant `DataSplit`s for modification involves a full join between the target and source tables. This PR introduces a shortcut optimization when the merge condition is a simple equality on the target's `_ROW_ID` (e.g., `ON target._ROW_ID = source.col`). 

If so, it directly scans the source table's `col` to identify the relevant `_ROW_ID`s to determine the affected splits. This avoids the need for a full join. For all other merge conditions, the existing join-based strategy is used, preserving the original behavior.

### Tests

A new test case has been added to `RowTrackingTestBase` to specifically verify this pr:

* `org.apache.paimon.spark.sql.RowTrackingTestBase#Data Evolution: merge into table with data-evolution with _ROW_ID shortcut`

### API and Format

No.

### Documentation

No.

### DIFF
--- a/paimon-spark/paimon-spark-ut/pom.xml
+++ b/paimon-spark/paimon-spark-ut/pom.xml
@@ -105,18 +105,6 @@ under the License.
                     </execution>
                 </executions>
             </plugin>
-            <!-- 在这里添加以下新的插件配置 -->
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-compiler-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


JingsongLi (approved):
Thanks @wayneli-vt for the contribution. Looks good to me!

PR #6749, opened 2025-12-04
[python/rest] support dlf.oss-endpoint override endpoint in rest_token_file_io

XiaoHongbo-Hope:
### Purpose
support dlf.oss-endpoint override endpoint in rest_token_file_io which follows rest token file io behivour in java

Linked issue: close #xxx


JingsongLi (approved):
+1

PR #6750, opened 2025-12-04
[hotfix/python] fix pypaimin timestamp non free time zone issue 

XiaoHongbo-Hope:
### Purpose
pypaimon should follow the convention which timestamp should be free timezone


JingsongLi (approved):
+1

PR #6753, opened 2025-12-05
[core] Should prevent users from specifying partition columns as blob field

steFaiz:
### Purpose
Sometimes user may specify a partition field as the blob field. We should check it on creating tables.

Linked issue: none #xxx

### Tests
Please see org.apache.paimon.schema.SchemaValidationTest

### API and Format
None

### Documentation
None


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkCatalog.java
```diff
@@ -1060,6 +1062,9 @@ public static Schema fromCatalogTable(CatalogBaseTable catalogTable) {
                             + "', you must also set '"
                             + CoreOptions.DATA_EVOLUTION_ENABLED.key()
                             + "'");
+            checkArgument(
```
JingsongLi: This check should be in `SchemaValidation`.

steFaiz: @JingsongLi Thanks for your advise! I've moved the check logic to SchemaValidation, so that all engines will share it.

PR #6755, opened 2025-12-05
[core] Extract an SST File Format from LookupStore.

steFaiz:
### Purpose
This PR is a part of https://github.com/apache/paimon/issues/6734

This PR do **NOT** change any file structure, mainly moving classes under `paimon-common/lookup/sort` to `paimon-common/sst`.
Other changes can be summarized as below:
1. Rename previous `SortLookupStoreReader` and `SortLookupStoreWriter` to `SstFileReader` and `SstFileWriter`
2. Slightly change current `SstFileReader` and `SstFileWriter` to directly interact with `SeekableInputStream` and `PositionOutputStream`. Also modified related classes such as `BlockCache` and `FileBasedBloomFilter`.
3. Create new `SortLookupStoreReader` and `SortLookupStoreWriter` which just delegate reads and writes to inner Sst File
Reader/Writer

I'll try to modify SST File structure to make it more generalize in future PRs. i.e. support range query and sequential scan.

### Tests
Please see:
* `org.apache.paimon.lookup.sort.SortLookupStoreFactoryTest` for LookupStore Test
* `org.apache.paimon.sst.SstFileTest` for SST File Test

### API and Format
None

### Documentation
None for now

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/utils/FileBasedBloomFilter.java
+++ b/paimon-common/src/main/java/org/apache/paimon/utils/FileBasedBloomFilter.java
@@ -19,7 +19,8 @@
 package org.apache.paimon.utils;
 
 import org.apache.paimon.annotation.VisibleForTesting;
-import org.apache.paimon.io.PageFileInput;
+import org.apache.paimon.fs.Path;
+import org.apache.paimon.fs.SeekableInputStream;
 import org.apache.paimon.io.cache.CacheCallback;
 import org.apache.paimon.io.cache.CacheKey;
 import org.apache.paimon.io.cache.CacheManager;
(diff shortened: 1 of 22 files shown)


paimon-common/src/main/java/org/apache/paimon/utils/FileBasedBloomFilter.java
```diff
@@ -34,16 +35,18 @@
...
 
-    private final PageFileInput input;
+    private final SeekableInputStream input;
     private final CacheManager cacheManager;
     private final BloomFilter filter;
     private final long readOffset;
-    private final int readLength;
     private final CacheKey cacheKey;
+    // each bloom filter is only used by a single file reader, so we can safely reuse here
+    private final byte[] reusedPageBuffer;
```
JingsongLi, line 44: Do not introduce this, when memory preemption occurs, it's precisely when memory usage needs to be cleaned up. Your way of referencing it makes it impossible to clean up.

JingsongLi (approved):
+1 I will refactor the codes.

PR #6759, opened 2025-12-05
[python/hotfix] update data type of _VALUE_KIND to int8

XiaoHongbo-Hope:
### Purpose
The data type of _VALUE_KIND should be int8 instead of int32.


JingsongLi (approved):
+1

Issue #6760, opened 2025-12-06
[Feature] Pypaimon support compaction task


kaori-seasons:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

## Overview

1. [Executive Summary](#executive-summary)
2. [Overview](#overview)
3. [Design Goals & Rationale](#design-goals--rationale)
4. [Architecture](#architecture)
5. [Core Components](#core-components)
6. [API Design & Changes](#api-design--changes)
7. [User Examples](#user-examples)
8. [Data Impact Analysis](#data-impact-analysis)
9. [Problems Solved](#problems-solved)
10. [Improvements & Benefits](#improvements--benefits)

---

## Executive Summary

The Paimon Python Compaction module introduces a comprehensive file consolidation mechanism for paimon-python, enabling efficient management of small files in Paimon tables. This document describes **Phase 1: Foundation Framework**, which establishes the core compaction capabilities with production-ready quality.

### Key Highlights

- **8 Core Modules** with 2,107 lines of production-grade Python code
- **71 Comprehensive Unit Tests** with 100% pass rate
- **Flexible API Design** supporting multiple compaction strategies and partition filtering
- **Memory-Efficient Implementation** using batch scanning (100K file batches)
- **Full Code Quality Compliance** (flake8, type hints, detailed logging)

### Implementation Timeline

| Phase | Component | Status | Completion Date |
|-------|-----------|--------|-----------------|
| Phase 1 | Core Framework (Builder, Executor, Coordinator, Manager) | ✅ Complete | Dec 2024 |
| Phase 2 | Automatic Trigger Mechanism | 📋 Planned | Q1 2025 |
| Phase 3 | Optimization & Advanced Features | 📋 Planned | Q2 2025 |

---

### What is Compaction?

Compaction is a critical maintenance operation in LSM-tree based storage systems that merges multiple small data files into larger, optimized files. Benefits include:

- **Improved Query Performance**: Fewer files to scan
- **Reduced I/O Overhead**: Better cache utilization
- **Disk Space Optimization**: Eliminates duplicate entries
- **System Stability**: Prevents runaway small file accumulation

### Scope of Phase 1

Phase 1 implements the **foundation framework** for manual and semi-automatic compaction:

```
┌─────────────────────────────────────────────────────┐
│         Paimon Python Compaction Phase 1             │
├─────────────────────────────────────────────────────┤
│                                                      │
[...]

PR #6761, opened 2025-12-06
feat: Pypaimon support compaction task

kaori-seasons:
### Purpose

Related to [issue-6760](https://github.com/apache/paimon/issues/6760)


JingsongLi (commented):
Same concerns as #6770:

1. **4755 additions** with minimal PR description ("Related to issue-6760").
2. This introduces an entire compaction framework in Python (coordinator, executor, manager, strategy, tasks, partition predicates) — this is a major feature that needs design discussion.
3. **No tests** listed in the PR template.
4. The relationship to the existing Java-side compaction (which handles most production workloads) is unclear. When would users choose Python compaction over Flink/Spark-based compaction?
5. Does this PR have any relationship to #7771 (which adds Ray-based distributed compaction) by TheR1sing3un? If so, these should be coordinated.

Please start with a design proposal on the mailing list. A Python compaction implementation needs to be compatible with the Java-side's file format, manifest handling, and commit protocol — this requires careful design review.

PR #6764, opened 2025-12-06
[lance] Do not replace oss scheme to s3 scheme

XiaoHongbo-Hope:
### Purpose
We should not replace oss scheme to s3 scheme, and should keep the original oss config into storage options for lance.

### Tests
`lance_utils_test.py`
`LanceUtilsTest`


ColdL:
+1

JingsongLi:
+1

PR #6766, opened 2025-12-07
[Python] Support read deletion vector pk table

discivigour:
### Purpose

Support read deletion vector pk table

### Tests

- python: test_pk_dv_read
- java: testPKDeletionVectorWrite


paimon-python/pypaimon/read/split_read.py
```diff
@@ -281,17 +285,33 @@ def _construct_partition_mapping(self) -> List[int]:
 
         return mapping
 
+    def _genarate_deletion_file_readers(self):
```
XiaoHongbo-Hope, line 288: typo: genarate -> generate

paimon-python/pypaimon/deletionvectors/deletion_vector.py
```diff
@@ -0,0 +1,142 @@
...
+
+                # Skip CRC (4 bytes)
+                f.read(4)
+
+                return BitmapDeletionVector.deserialize_from_bytes(data)
+            else:
+                raise RuntimeError(
+                    f"Invalid magic number: {magic_number}, "
+                    f"expected: {BitmapDeletionVector.MAGIC_NUMBER}"
+                )
```
XiaoHongbo-Hope, line 142: Do we need support 64-bit bitmap?

paimon-python/pypaimon/deletionvectors/apply_deletion_vector_reader.py
```diff
@@ -0,0 +1,129 @@
...
+        if arrow_batch is None:
+            return None
+        # Remove the deleted rows from the batch
+        range_bitmap = BitMap(
+            range(self._reader.return_batch_pos() - arrow_batch.num_rows, self._reader.return_batch_pos()))
+        intersection_bitmap = range_bitmap - self._deletion_vector.bit_map()
+        added_row_list = [x - (self._reader.return_batch_pos() - arrow_batch.num_rows) for x in
+                          list(intersection_bitmap)]
+        return arrow_batch.take(pyarrow.array(added_row_list, type=pyarrow.int32()))
+
```
XiaoHongbo-Hope, line 66: We can add empty check for self._deletion_vector, if empty, then return arrow_batch directly

paimon-python/pypaimon/deletionvectors/deletion_vector.py
JingsongLi: Not now, it is just for iceberg compatibility.

paimon-python/pypaimon/deletionvectors/apply_deletion_vector_reader.py
JingsongLi: It is not easy to get an empty deletion vector, so I think it is OK without empty check.

paimon-python/pypaimon/deletionvectors/deletion_vector.py
XiaoHongbo-Hope: > Not now, it is just for iceberg compatibility.

👌

paimon-python/pypaimon/deletionvectors/apply_deletion_vector_reader.py
XiaoHongbo-Hope: > It is not easy to get an empty deletion vector, so I think it is OK without empty check.

👌

paimon-python/dev/run_mixed_tests.sh
```diff
@@ -240,6 +268,12 @@ main() {
         echo -e "${RED}✗ Java Read Test (Parquet + Lance): FAILED${NC}"
     fi
 
+     if [[ $pk_dv_result -eq 0 ]]; then
```
XiaoHongbo-Hope, line 271: We should add pk_dv_result return value check in line 282 too

paimon-core/src/test/java/org/apache/paimon/JavaPyE2ETest.java
```diff
@@ -198,6 +220,139 @@ public void testJavaWriteReadPkTable() throws Exception {
...
+                                row,
+                                DataTypes.ROW(
+                                        DataTypes.INT(), DataTypes.INT(), DataTypes.BIGINT()));
+        List<String> result =
+                getResult(table.newRead(), table.newScan().plan().splits(), rowDataToString);
+
+        assertThat(result).hasSize(9999);
+
+        assertThat(result).doesNotContain("+I[1, 81930, 819300L]");
+
```
XiaoHongbo-Hope, line 351: remove "L" here

paimon-python/dev/run_mixed_tests.sh
discivigour: ok

paimon-python/pypaimon/deletionvectors/apply_deletion_vector_reader.py
```diff
@@ -0,0 +1,129 @@
...
+        if batch is None:
+            return None
+
+        return ApplyDeletionRecordIterator(batch, self._deletion_vector)
+
+    def close(self):
+        self._reader.close()
+
+
+class ApplyDeletionRecordIterator(RecordIterator):
```
kaori-seasons, line 85: Problem: The `ApplyDeletionRecordIterator`, `RowPositionRecordIterator`, and `KeyValueWrapIterator` classes are inconsistent in how they return the position (`return_pos`): some classes use the name `returned_position`, while others use `return_pos`.

This can easily lead to `TypeError` / `AttributeError` at runtime, and makes it difficult to guarantee that wrappers (such as `ApplyDeletionVectorReader`, `RowPositionReader`) will work correctly in all scenarios.

Suggestion: Implement compatibility handling in all wrappers/adapters for cases where the object might be a Python iterator or a custom `RecordIterator`. The best practice is to require `RecordIterator` to implement `next` (to achieve fundamental consistency), but if it's not possible to modify all implementations at once, use a compatibility layer (checking `hasattr(next)` / `hasattr(next)`, etc.) to avoid crashes.

Examples:

```
# pypaimon/read/reader/iface/record_iterator.py

from abc import ABC, abstractmethod
from typing import Generic, Optional, TypeVar, Iterator

T = TypeVar("T")

class RecordIterator(Iterator[T], Generic[T], ABC):
    """
    Record iterator: must implement Python iterator protocol and provide return_pos().

    Implementations should:
    - implement __next__() returning Optional[T] (or raise StopIteration)
    - implement return_pos() returning current position index for the last returned element
    """

    def __iter__(self):
        return self

    @abstractmethod
    def __next__(self) -> T:
        """Return next record or raise StopIteration"""
        ...

    # backward-compat helper for code that used next()
    def next(self) -> Optional[T]:
        try:
            return self.__next__()
        except StopIteration:
            return None

    @abstractmethod
    def return_pos(self) -> int:
        """Return current position of the iterator (index of the last returned element)."""
        ...

```

```
class RowPositionRecordIterator(RecordIterator[tuple]):
    def __init__(self):
        self.reused_iterator: Optional[Iterator[tuple]] = None
        self.pos = -1

    def __next__(self) -> tuple:
        if self.reused_iterator is None:
            raise StopIteration
        row_tuple = next(self.reused_iterator, None)
        if row_tuple is None:
            raise StopIteration
        self.pos += 1
        return row_tuple

    def next(self) -> Optional[tuple]:
        try:
            return self.__next__()
        except StopIteration:
            return None

    def return_pos(self) -> int:
        return self.pos
```


Then update all custom implementations (e.g., RowPositionRecordIterator, InternalRowWrapperIterator, etc.):

Implement `next`, and maintain the position counter `self.pos` within `next` (the logic of the original `next` method is moved into the new `next`), and retain `next()` as a compatibility alias (as above).

paimon-python/pypaimon/deletionvectors/bitmap_deletion_vector.py
```diff
@@ -0,0 +1,165 @@
...
+        """
+        return zlib.crc32(data) & 0xffffffff
+
+    def __eq__(self, other):
+        if not isinstance(other, BitmapDeletionVector):
+            return False
+        return self._bitmap == other._bitmap
+
+    def __hash__(self):
+        return hash(tuple(self._bitmap))
```
kaori-seasons, line 165: BitmapDeletionVector.hash: Returns hash(tuple(self._bitmap)). This is inefficient on large bitmaps as it iterates through all bits, and it's rarely used. Consider removing the hash function or replacing it with one based on CRC/hash of serialized bytes (if a hash is truly needed).

paimon-python/pypaimon/deletionvectors/apply_deletion_vector_reader.py
```diff
@@ -0,0 +1,129 @@
...
+    def deletion_vector(self) -> DeletionVector:
+        return self._deletion_vector
+
+    def read_arrow_batch(self) -> Optional[RecordBatch]:
+        self._reader: RecordBatchReader
+        arrow_batch = self._reader.read_arrow_batch()
+        if arrow_batch is None:
+            return None
+        # Remove the deleted rows from the batch
+        range_bitmap = BitMap(
```
kaori-seasons, line 60: `apply_deletion_vector_reader.ApplyDeletionVectorReader.read_arrow_batch` uses `BitMap(range(batch_start, batch_end))` and then performs a set difference operation to obtain the non-deleted positions.
When the batch is very large (or the number of batches is very large), creating such a large range BitMap consumes a significant amount of memory/CPU. However, the actual deletion bitmap is usually sparse, so iterating through the deletion bitmap should be prioritized.

Impact: Performance degradation and OOM risk, especially in large tables or long intervals.

Suggested fix: Instead of generating a BitMap representing the entire batch, directly filter the positions within the current batch range from the deletion vector's bitmap (which is usually sparse and more efficient).
`pyroaring.BitMap` is iterable, so it can be iterated bit by bit and filtered.

paimon-python/pypaimon/manifest/index_manifest_file.py
```diff
@@ -0,0 +1,95 @@
...
+class IndexManifestFile:
+    """Index manifest file reader for reading index manifest entries."""
+
+    DELETION_VECTORS_INDEX = "DELETION_VECTORS"
+
+    def __init__(self, table):
+        from pypaimon.table.file_store_table import FileStoreTable
+
+        self.table: FileStoreTable = table
+        manifest_path = table.table_path.rstrip('/')
```
kaori-seasons, line 38: Problem: In `_to_deletion_files`, the `index_path` is constructed directly using `self.table.table_path + '/index'`, without considering `index_file.external_path` (the `IndexFileMeta` object has an `external_path` field).

Impact: When parsing `dv_meta_record` using fastavro, hardcoded field names like `dv_meta_record['f0']`, 'f1', 'f2', etc., are used. If the Avro schema field names are different, an error will occur.

Suggestion: When constructing `dv_file_path`, if `index_file.external_path` exists, it should be used preferentially.
When parsing `dv_meta_record`, use a defensive approach with key lookup (`get`) and be compatible with multiple naming conventions (or use schema-driven parsing with the fastavro reader).

Examples:
```
def _to_deletion_files(self, index_entry) -> Dict[str, DeletionFile]:
    deletion_files = {}
    index_file = index_entry.index_file

    if not index_file.dv_ranges:
        return deletion_files

    # prefer external_path when present
    index_path = index_file.external_path if index_file.external_path else self.table.table_path.rstrip('/') + '/index'
    dv_file_path = f"{index_path}/{index_file.file_name}"

    for key, dv_meta in index_file.dv_ranges.items():
        # dv_meta is expected to be DeletionVectorMeta
        deletion_file = DeletionFile(
            dv_index_path=dv_file_path,
            offset=dv_meta.offset,
            length=dv_meta.length,
            cardinality=dv_meta.cardinality
        )
        deletion_files[key] = deletion_file

    return deletion_files
```

XiaoHongbo-Hope:
+1

PR #6768, opened 2025-12-07
[arrow] Fix incorrect index usage in ArrowFieldWriters.TimeWriter

### LINKED ISSUE #6767: [Bug] `ArrowFieldWriters.TimeWriter` ignores startIndex and reads incorrect rows
### Search before asking

### Paimon version

master

### Compute Engine

None

### Minimal reproduce step

Description:
In `ArrowFieldWriters.TimeWriter`, the `doWrite` method uses wrong index `i` instead of `row` when reading from the source columnVector.

Impact:
When `startIndex > 0`, the writer reads incorrect values from the source column vector, causing data corruption in the Arrow output.

Reproduction:
```java
public static void main(String[] args) {
        RowType rowType = new RowType(
                Collections.singletonList(new DataField(0, "time_field", DataTypes.TIME())));

        int startIndex = 2;
        final int batchRows = 3;

        try (RootAllocator allocator = new RootAllocator();
             VectorSchemaRoot vsr = ArrowUtils.createVectorSchemaRoot(rowType, allocator)) {

            ArrowFieldWriter[] fieldWriters = ArrowUtils.createArrowFieldWriters(vsr, rowType);

            IntColumnVector timeVec = new IntColumnVector() {
                final int[] values = new int[] {0, 1000, 2000, 3000, 4000};
                @Override public int getInt(int i) { return values[i]; }
                @Override public boolean isNullAt(int i) { return false; }
            };

            fieldWriters[0].write(timeVec, null, 2, batchRows);
            vsr.setRowCount(batchRows);

            TimeMilliVector timeMilliVector = (TimeMilliVector) vsr.getVector("time_field");

            for (int i = 0; i < batchRows; i++) {
                System.out.printf("paimonValue: %s, <--> arrowValue: %s%n",
                        timeVec.getInt(i + startIndex), timeMilliVector.get(i));
            }
        } catch (Exception e) {
            e.printStackTrace();
        }
    }
```
Print output:  
```bash
paimonValue: 2000, <--> arrowValue: 0
paimonValue: 3000, <--> arrowValue: 1000
paimonValue: 4000, <--> arrowValue: 2000
```

None

### Anything else?

_No response_

guluo2016:
### Purpose

Linked issue: close #6767

### DIFF
--- a/paimon-arrow/src/test/java/org/apache/paimon/arrow/vector/ArrowFormatWriterTest.java
+++ b/paimon-arrow/src/test/java/org/apache/paimon/arrow/vector/ArrowFormatWriterTest.java
@@ -440,15 +440,15 @@ public class ArrowFormatWriterTest {
 
     @Test
     public void testTimeFieldWriterWithOffset() {
+        String fieldName = "time_field";
         RowType rowType =
                 new RowType(
-                        Collections.singletonList(
-                                new DataField(0, "time_field", DataTypes.TIME())));
+                        Collections.singletonList(new DataField(0, fieldName, DataTypes.TIME())));
         try (RootAllocator allocator = new RootAllocator();
                 VectorSchemaRoot vsr = ArrowUtils.createVectorSchemaRoot(rowType, allocator)) {
             ArrowFieldWriter[] fieldWriters = ArrowUtils.createArrowFieldWriters(vsr, rowType);
 
-            IntColumnVector timeVec =
+            IntColumnVector paimonTimeVector =
                     new IntColumnVector() {
                         final int[] values = new int[] {0, 1000, 2000, 3000, 4000};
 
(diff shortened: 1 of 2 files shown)


paimon-arrow/src/test/java/org/apache/paimon/arrow/vector/ArrowFormatWriterTest.java
```diff
@@ -433,6 +438,55 @@ public void testWriteRowArrayTwice() {
...
+                new RowType(
+                        Collections.singletonList(
+                                new DataField(0, "time_field", DataTypes.TIME())));
+        try (RootAllocator allocator = new RootAllocator();
+                VectorSchemaRoot vsr = ArrowUtils.createVectorSchemaRoot(rowType, allocator)) {
+            ArrowFieldWriter[] fieldWriters = ArrowUtils.createArrowFieldWriters(vsr, rowType);
+
+            IntColumnVector timeVec =
+                    new IntColumnVector() {
+                        final int[] values = new int[] {0, 1000, 2000, 3000, 4000};
```
kaori-seasons, line 453: Current situation: Similar index mixing (using `i` to read source data, or using `row` to check the Arrow vector) can easily occur repeatedly in multiple field writers, leading to hidden bugs.

Recommendation:

Recommendation: Define and strictly distinguish between `sourceIndex` (e.g., `rowIndex = startIndex + i`) and `targetIndex(i)` in the implementation.

Within the loop body of each writer, use `sourceIndex` to read data from the `ColumnVector`, and use `targetIndex` to write data to the `ArrowVector`; use `sourceIndex` to check for null values.

Example:
```
// Assume startIndex, batchRows, columnVector, and timeMilliVector are known.
for (int i = 0; i < batchRows; i++) {
    int sourceIndex = startIndex + i; // 从 columnVector 读取的索引
    int targetIndex = i;              // 写到 Arrow 向量的位置
    if (columnVector.isNullAt(sourceIndex)) {
        timeMilliVector.setNull(targetIndex);
    } else {
        int value = ((IntColumnVector) columnVector).getInt(sourceIndex);
        timeMilliVector.setSafe(targetIndex, value);
    }
}
```

guluo2016: I didn’t modify `ArrowFieldWriters.java` to keep it consistent with the rest of the code. 
Updating the test case to improve readability, thanks for the review

JingsongLi (commented):
Classic off-by-one index bug. The TimeWriter was using the loop variable `i` (output index) instead of `row` (input index from the column vector).

```java
// Before (wrong): reads wrong position from source vector
int value = ((IntColumnVector) columnVector).getInt(i);
// After (correct): reads from the actual row position
int value = ((IntColumnVector) columnVector).getInt(row);
```

This would produce incorrect TIME values when writing with an offset (e.g., processing a subset of rows starting from a non-zero position). The bug is silent — wrong values written without error.

The test `testTimeFieldWriterWithOffset` verifies correct behavior with non-zero `startIndex`. Good regression test.

LGTM — straightforward one-line fix with proper test.

JingsongLi:
Already fixed.

PR #6773, opened 2025-12-08
[lucene] Support vector index using Lucene

jerry-024:
### Purpose
Introduce vector index support for Paimon using Apache Lucene 9.x to enable:
- Approximate nearest neighbor (ANN) search for vector embeddings

### Implementation
- Uses Lucene's HNSW algorithm for efficient ANN search
- Supports float32 and byte vectors
- Supports EUCLIDEAN, COSINE, DOT_PRODUCT, and MAXIMUM_INNER_PRODUCT similarity metrics

### Tests
- VectorGlobalIndexTest

### DIFF
--- a/paimon-lucene/src/main/java/org/apache/paimon/lucene/index/LuceneByteVectorIndex.java
+++ b/paimon-lucene/src/main/java/org/apache/paimon/lucene/index/LuceneByteVectorIndex.java
@@ -18,23 +18,18 @@
 
 package org.apache.paimon.lucene.index;
 
-import org.apache.lucene.document.KnnFloatVectorField;
+import org.apache.lucene.document.KnnByteVectorField;
 import org.apache.lucene.index.IndexableField;
 import org.apache.lucene.index.VectorSimilarityFunction;
 
-/** Vector index for float vector. */
-public class FloatVectorIndex extends VectorIndex<float[]> {
+/** Lucene vector index for byte vector. */
+public class LuceneByteVectorIndex extends LuceneVectorIndex<byte[]> {
     private final long rowId;
-    private final float[] vector;
+    private final byte[] vector;
 
-    public FloatVectorIndex(long rowId, float[] vector) {
+    public LuceneByteVectorIndex(long rowId, byte[] vector) {
         this.rowId = rowId;
         this.vector = vector;
-    }
-
-    @Override
-    public float[] vector() {
-        return vector;
     }
 
     @Override
@@ -49,6 +44,11 @@ public class FloatVectorIndex extends VectorIndex<float[]> {
 
     @Override
     public IndexableField indexableField(VectorSimilarityFunction similarityFunction) {
-        return new KnnFloatVectorField(VECTOR_FIELD, this.vector(), similarityFunction);
+        return new KnnByteVectorField(VECTOR_FIELD, this.vector, similarityFunction);
+    }
+
+    @Override
+    public byte[] vector() {
+        return vector;
     }
 }
--- a/paimon-lucene/src/main/java/org/apache/paimon/lucene/index/LuceneFloatVectorIndex.java
+++ b/paimon-lucene/src/main/java/org/apache/paimon/lucene/index/LuceneFloatVectorIndex.java
@@ -18,18 +18,23 @@
 
 package org.apache.paimon.lucene.index;
 
-import org.apache.lucene.document.KnnByteVectorField;
+import org.apache.lucene.document.KnnFloatVectorField;
 import org.apache.lucene.index.IndexableField;
 import org.apache.lucene.index.VectorSimilarityFunction;
 
-/** Vector index for byte vector. */
-public class ByteVectorIndex extends VectorIndex<byte[]> {
+/** Lucene vector index for float vector. */
+public class LuceneFloatVectorIndex extends LuceneVectorIndex<float[]> {
     private final long rowId;
-    private final byte[] vector;
+    private final float[] vector;
 
-    public ByteVectorIndex(long rowId, byte[] vector) {
+    public LuceneFloatVectorIndex(long rowId, float[] vector) {
         this.rowId = rowId;
         this.vector = vector;
+    }
+
+    @Override
+    public float[] vector() {
+        return vector;
     }
 
     @Override
@@ -44,11 +49,6 @@ public class ByteVectorIndex extends VectorIndex<byte[]> {
 
     @Override
     public IndexableField indexableField(VectorSimilarityFunction similarityFunction) {
-        return new KnnByteVectorField(VECTOR_FIELD, this.vector, similarityFunction);
-    }
-
-    @Override
-    public byte[] vector() {
-        return vector;
+        return new KnnFloatVectorField(VECTOR_FIELD, this.vector(), similarityFunction);
     }
 }
--- a/paimon-lucene/src/main/java/org/apache/paimon/lucene/index/LuceneIndexMMapDirectory.java
+++ b/paimon-lucene/src/main/java/org/apache/paimon/lucene/index/LuceneIndexMMapDirectory.java
@@ -34,15 +34,15 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.UUID;
 
-/** A wrapper of MMapDirectory for vector index. */
-public class IndexMMapDirectory implements AutoCloseable {
+/** A wrapper of Lucene MMapDirectory for vector index. */
+public class LuceneIndexMMapDirectory implements AutoCloseable {
 
     private static final int VERSION = 1;
 
     private final Path path;
     private final MMapDirectory mmapDirectory;
 
-    public IndexMMapDirectory() throws IOException {
+    public LuceneIndexMMapDirectory() throws IOException {
         this.path = Files.createTempDirectory("paimon-lucene-" + UUID.randomUUID());
         this.mmapDirectory = new MMapDirectory(path);
     }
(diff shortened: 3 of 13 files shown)


paimon-vector-index/pom.xml
```diff
@@ -0,0 +1,124 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <artifactId>paimon-parent</artifactId>
+        <groupId>org.apache.paimon</groupId>
+        <version>1.4-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>paimon-vector-index</artifactId>
```
JingsongLi: Maybe `paimon-lucene` is better.

paimon-lucene/src/main/java/org/apache/paimon/lucene/index/VectorGlobalIndexReader.java
```diff
@@ -0,0 +1,341 @@
...
+
+    private void loadIndices(GlobalIndexFileReader fileReader, List<GlobalIndexIOMeta> files)
+            throws IOException {
+        for (GlobalIndexIOMeta meta : files) {
+            try (SeekableInputStream in = fileReader.getInputStream(meta.fileName())) {
+                IndexMMapDirectory directory = null;
+                IndexReader reader = null;
+                boolean success = false;
+                try {
+                    directory = deserializeDirectory(in);
```
kaori-seasons: Current situation: `serializeDirectory` / `deserializeDirectory` only write the number of files, the length of each filename, the filename itself, and the file length, then directly write the bytes.  There is no magic number, version number, or verification/checksum mechanism.

Impact: If the serialization structure is modified in the future, older versions will not be able to recognize it; it's difficult to pinpoint which part is corrupted if the stream is damaged; security/robustness is poor (prone to silent corruption).

Recommendation: Add a magic number, version number, and CRC32 for each file (or an overall checksum), and use standard `DataOutputStream`/`DataInputStream` to write `int`/`long`, avoiding manual `ByteBuffer` wrapping. This will be more robust and easier to maintain backward compatibility (using a version field).


examples:
```
private static final int MAGIC = 0x50414D4F; // 'PAMO' or any magic
private static final short VERSION = 1;

private void serializeDirectory(Directory directory, OutputStream out) throws IOException {
    DataOutputStream dos = new DataOutputStream(new BufferedOutputStream(out));
    dos.writeInt(MAGIC);
    dos.writeShort(VERSION);

    String[] files = directory.listAll();
    dos.writeInt(files.length);

    for (String fileName : files) {
        byte[] nameBytes = fileName.getBytes(StandardCharsets.UTF_8);
        dos.writeInt(nameBytes.length);
        dos.write(nameBytes);

        long fileLength = directory.fileLength(fileName);
        dos.writeLong(fileLength);

        try (IndexInput input = directory.openInput(fileName, IOContext.DEFAULT)) {
            byte[] buffer = new byte[32 * 1024];
            long remaining = fileLength;
            CRC32 crc = new CRC32();
            while (remaining > 0) {
                int toRead = (int) Math.min(buffer.length, remaining);
                input.readBytes(buffer, 0, toRead);
                dos.write(buffer, 0, toRead);
                crc.update(buffer, 0, toRead);
                remaining -= toRead;
            }
            dos.writeLong(crc.getValue()); // per-file checksum
        }
    }
    dos.flush();
}
```

And the corresponding `deserializeDirectory` function (which reads the magic number/version and verifies the CRC).


```
private IndexMMapDirectory deserializeDirectory(SeekableInputStream in) throws IOException {
    DataInputStream dis = new DataInputStream(new BufferedInputStream(in));
    int magic = dis.readInt();
    if (magic != MAGIC) {
        throw new IOException("Invalid vector index file magic: " + Integer.toHexString(magic));
    }
    short version = dis.readShort();
    if (version != VERSION) {
        throw new IOException("Unsupported vector index version: " + version);
    }

    IndexMMapDirectory indexMMapDirectory = new IndexMMapDirectory();
    try {
        int numFiles = dis.readInt();
        byte[] buffer = new byte[BUFFER_SIZE];

        for (int i = 0; i < numFiles; i++) {
            int nameLength = dis.readInt();
            byte[] nameBytes = new byte[nameLength];
            dis.readFully(nameBytes);
            String fileName = new String(nameBytes, StandardCharsets.UTF_8);
            long fileLength = dis.readLong();

            long expectedCrc = -1L;
            // We need to stream file data to index output, and compute CRC while reading
            try (IndexOutput output = indexMMapDirectory.directory().createOutput(fileName, IOContext.DEFAULT)) {
                long remaining = fileLength;
                CRC32 crc = new CRC32();
                while (remaining > 0) {
                    int toRead = (int) Math.min(buffer.length, remaining);
                    dis.readFully(buffer, 0, toRead);
                    output.writeBytes(buffer, 0, toRead);
                    crc.update(buffer, 0, toRead);
                    remaining -= toRead;
                }
                expectedCrc = dis.readLong();
                if (crc.getValue() != expectedCrc) {
                    throw new IOException("CRC mismatch for file " + fileName);
                }
            }
        }
        return indexMMapDirectory;
    } catch (Exception e) {
        try {
            indexMMapDirectory.close();
        } catch (Exception ee) {
            // ignore or add suppressed
            e.addSuppressed(ee);
        }
        if (e instanceof IOException) {
            throw (IOException) e;
        } else {
            throw new IOException("Failed to deserialize directory", e);
        }
    }
}
```

leaves12138: @kaori-seasons Please do not copy all AI answer to the review comment, it will cost author a lot to understand a simple point of view. You should summarize them into one or two words, like "I think add a MAGIC number is better".

JingsongLi: +1 to @leaves12138 This is really a waste of everyone's time.

.github/workflows/utitcase-paimon-lucene.yml
```diff
@@ -0,0 +1,62 @@
...
+# limitations under the License.
+################################################################################
+
+name: UTCase and ITCase Paimon Lucene
+
+on:
+  push:
+  pull_request:
+    paths:
+      - 'paimon-lucene/**'
```
JingsongLi: Why not cover this in utitcase-jdk11?

paimon-lucene/src/main/java/org/apache/paimon/lucene/index/IndexMMapDirectory.java
```diff
@@ -0,0 +1,136 @@
...
+/** A wrapper of MMapDirectory for vector index. */
+public class IndexMMapDirectory implements AutoCloseable {
+
+    private static final int VERSION = 1;
+
+    private final Path path;
+    private final MMapDirectory mmapDirectory;
+
+    public IndexMMapDirectory() throws IOException {
+        this.path = java.nio.file.Files.createTempDirectory("paimon-lucene-" + UUID.randomUUID());
```
leaves12138: Please do not use java.nio.file.Files in class, import them in head.

paimon-lucene/src/main/java/org/apache/paimon/lucene/index/IndexMMapDirectory.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    public MMapDirectory directory() {
+        return mmapDirectory;
+    }
+
+    public void close() throws Exception {
+        mmapDirectory.close();
+        if (java.nio.file.Files.exists(path)) {
+            java.nio.file.Files.walk(path)
```
leaves12138: ditto

paimon-lucene/src/main/java/org/apache/paimon/lucene/index/IndexMMapDirectory.java
```diff
@@ -0,0 +1,136 @@
...
+                    .sorted(java.util.Comparator.reverseOrder())
```
leaves12138: ditto

paimon-lucene/src/main/java/org/apache/paimon/lucene/index/IndexMMapDirectory.java
```diff
@@ -0,0 +1,136 @@
...
+                    .forEach(
+                            p -> {
+                                try {
+                                    java.nio.file.Files.delete(p);
```
leaves12138: ditto

paimon-lucene/src/main/java/org/apache/paimon/lucene/index/IndexMMapDirectory.java
leaves12138: We don't need to sort

paimon-lucene/src/main/java/org/apache/paimon/lucene/index/VectorGlobalIndexerFactory.java
```diff
@@ -0,0 +1,40 @@
...
+
+import org.apache.paimon.globalindex.GlobalIndexer;
+import org.apache.paimon.globalindex.GlobalIndexerFactory;
+import org.apache.paimon.options.Options;
+import org.apache.paimon.types.DataType;
+
+/** Factory for creating vector global indexers. */
+public class VectorGlobalIndexerFactory implements GlobalIndexerFactory {
+
+    public static final String IDENTIFIER = "vector";
```
leaves12138: lucene-hnw? "vector" is too wide.

paimon-lucene/src/main/java/org/apache/paimon/lucene/index/VectorGlobalIndexReader.java
```diff
@@ -0,0 +1,277 @@
...
+                    }
+                }
+            }
+        }
+    }
+
+    // Implementation of FunctionVisitor methods
+    @Override
+    public GlobalIndexResult visitIsNotNull(FieldRef fieldRef) {
+        throw new UnsupportedOperationException(
```
leaves12138: Do not throw Exception. Please return `GlobalIndexResult.fromRange(<the range pass to you in GlobalIndexIOMeta>)`

paimon-lucene/src/main/java/org/apache/paimon/lucene/index/VectorGlobalIndexerFactory.java
jerry-024: how about lucene-hnsw? (ps: HNSW (Hierarchical Navigable Small World))

JingsongLi:
+1

PR #6776, opened 2025-12-08
[core] Introduce Like LeafFunction for Predicate

JingsongLi:
### Purpose

Introduce Like, the execution of Like has been highly optimized; it can be optimized to equal, starsWith, endsWith, or Contains.

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/predicate/Like.java
+++ b/paimon-common/src/main/java/org/apache/paimon/predicate/Like.java
@@ -65,6 +65,8 @@ public class Like extends NullFalseLeafBinaryFunction {
             Object literal = optimized.get().getValue();
             return field -> func.test(type, field, literal);
         }
+        // TODO optimize for chain checkers when there is no '_'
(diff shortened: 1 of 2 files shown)


paimon-common/src/main/java/org/apache/paimon/predicate/LikeOptimization.java
```diff
@@ -0,0 +1,74 @@
...
+    private static final Pattern NONE_PATTERN = Pattern.compile("[^%]+");
+
+    public static Optional<Pair<NullFalseLeafBinaryFunction, Object>> tryOptimize(
+            Object patternLiteral) {
+        if (patternLiteral == null) {
+            throw new IllegalArgumentException("Pattern can not be null.");
+        }
+
+        String pattern = patternLiteral.toString();
+        if (pattern.contains("_")) {
```
leaves12138, line 49: pattern.replace("\\\\_", "\u0000").contains("\_") to enable pattern like "%my\\_house"?

PR #6778, opened 2025-12-09
[core] add snapshot existence check for failed rename operation

### LINKED ISSUE #6777: [Bug] Snapshot file is erroneously deleted during FS rename retry, causing FileNotFoundException upon Flink failover
### Search before asking

### Paimon version

0.8.2

### Compute Engine

Flink

### Minimal reproduce step

1. Run a Flink job writing to Paimon on Object Storage (S3).

2. Simulate a network timeout specifically during the "Commit/Rename" phase of a checkpoint.

3. Trigger the file system client to retry the rename operation.

4. Observe the deletion of the target snapshot file.

5. Wait for Flink failover; the job will get stuck in a restart loop with FileNotFoundException.
`2025-11-12 18:05:09
java.lang.RuntimeException: java.io.FileNotFoundException: File 's3://dt-warehouse/paimon-warehouse/nd_game_sjmy_cdm.db/dwd_sjmy_gmlog_boss_user_bekilllist/manifest/manifest-list-20f67c4d-ba05-4dbf-a26a-24f229b949cc-32' not found, Possible causes: 1.snapshot expires too fast, you can configure 'snapshot.time-retained' option with a larger value. 2. consumption is too slow, you can improve the performance of consumption (For example, increasing parallelism).
	at org.apache.paimon.flink.sink.AsyncLookupSinkWrite.<init>(AsyncLookupSinkWrite.java:75)
	at org.apache.paimon.flink.sink.FlinkSink.lambda$createWriteProvider$672a9d60$1(FlinkSink.java:147)
	at org.apache.paimon.flink.sink.TableWriteOperator.initializeState(TableWriteOperator.java:78)
	at org.apache.flink.streaming.api.operators.StreamOperatorStateHandler.initializeOperatorState(StreamOperatorStateHandler.java:122)
	at org.apache.flink.streaming.api.operators.AbstractStreamOperator.initializeState(AbstractStreamOperator.java:316)
	at org.apache.flink.streaming.api.operators.AbstractStreamOperator.initializeState(AbstractStreamOperator.java:306)
	at org.apache.flink.streaming.runtime.tasks.RegularOperatorChain.initializeStateAndOpenOperators(RegularOperatorChain.java:107)
	at org.apache.flink.streaming.runtime.tasks.StreamTask.restoreGates(StreamTask.java:759)
	at org.apache.flink.streaming.runtime.tasks.StreamTaskActionExecutor$1.call(StreamTaskActionExecutor.java:55)
	at org.apache.flink.streaming.runtime.tasks.StreamTask.restoreInternal(StreamTask.java:734)
	at org.apache.flink.streaming.runtime.tasks.StreamTask.restore(StreamTask.java:699)
	at org.apache.flink.runtime.taskmanager.Task.runWithSystemExitMonitoring(Task.java:971)
	at org.apache.flink.runtime.taskmanager.Task.restoreAndInvoke(Task.java:940)
	at org.apache.flink.runtime.taskmanager.Task.doRun(Task.java:764)
	at org.apache.flink.runtime.taskmanager.Task.run(Task.java:574)
	at java.base/java.lang.Thread.run(Thread.java:829)
Caused by: java.lang.RuntimeException: java.io.FileNotFoundException: File 's3://dt-warehouse/paimon-warehouse/nd_game_sjmy_cdm.db/dwd_sjmy_gmlog_boss_user_bekilllist/manifest/manifest-list-20f67c4d-ba05-4dbf-a26a-24f229b949cc-32' not found, Possible causes: 1.snapshot expires too fast, you can configure 'snapshot.time-retained' option with a larger value. 2. consumption is too slow, you can improve the performance of consumption (For example, increasing parallelism).
	at org.apache.paimon.utils.ObjectsCache.readSegments(ObjectsCache.java:143)
	at org.apache.paimon.utils.ObjectsCache.read(ObjectsCache.java:93)
	at org.apache.paimon.utils.ObjectsFile.readWithIOException(ObjectsFile.java:149)
	at org.apache.paimon.utils.ObjectsFile.read(ObjectsFile.java:134)
	at org.apache.paimon.utils.ObjectsFile.read(ObjectsFile.java:105)
	at org.apache.paimon.manifest.ManifestList.readDataManifests(ManifestList.java:90)
	at org.apache.paimon.operation.ManifestsReader.readManifests(ManifestsReader.java:128)
	at org.apache.paimon.operation.ManifestsReader.read(ManifestsReader.java:114)
	at org.apache.paimon.operation.AbstractFileStoreScan.readManifests(AbstractFileStoreScan.java:417)
	at org.apache.paimon.operation.AbstractFileStoreScan.plan(AbstractFileStoreScan.java:257)
	at org.apache.paimon.operation.AbstractFileStoreWrite.scanExistingFileMetas(AbstractFileStoreWrite.java:491)
	at org.apache.paimon.operation.AbstractFileStoreWrite.createWriterContainer(AbstractFileStoreWrite.java:440)`

The rename operation should be idempotent or check for the existence of the target file before considering the operation failed. If the source is missing but the target exists and matches expectations during a retry, it should be treated as a success, or at least the target file should not be deleted.

### Anything else?

_No response_

aidadenski:
### Purpose

Linked issue: close [#6777](https://github.com/apache/paimon/issues/6777)

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/catalog/RenamingSnapshotCommit.java
+++ b/paimon-core/src/main/java/org/apache/paimon/catalog/RenamingSnapshotCommit.java
@@ -25,6 +25,7 @@ import org.apache.paimon.operation.Lock;
 import org.apache.paimon.partition.PartitionStatistics;
 import org.apache.paimon.utils.SnapshotManager;
 
+import java.io.IOException;
 import java.util.List;
 import java.util.concurrent.Callable;
 
@@ -57,6 +58,20 @@ public class RenamingSnapshotCommit implements SnapshotCommit {
         Callable<Boolean> callable =
                 () -> {
                     boolean committed = fileIO.tryToWriteAtomic(newSnapshotPath, snapshot.toJson());
+                    if (!committed) {
+                        if (!fileIO.exists(newSnapshotPath)) {
+                            throw new IOException(
+                                    "Commit snapshot "
+                                            + snapshot.id()
+                                            + " failed and "
+                                            + newSnapshotPath
+                                            + " not found");
+                        }
+                        committed =
+                                snapshot.equals(
+                                        Snapshot.fromJson(fileIO.readFileUtf8(newSnapshotPath)));
+                    }
+
                     if (committed) {
                         snapshotManager.commitLatestHint(snapshot.id());
                     }
--- a/paimon-core/src/test/java/org/apache/paimon/catalog/RenamingSnapshotCommitTest.java
+++ b/paimon-core/src/test/java/org/apache/paimon/catalog/RenamingSnapshotCommitTest.java
@@ -1,0 +1,176 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.paimon.catalog;
+
+import org.apache.paimon.Snapshot;
+import org.apache.paimon.fs.FileIO;
+import org.apache.paimon.fs.Path;
+import org.apache.paimon.fs.local.LocalFileIO;
+import org.apache.paimon.operation.Lock;
+import org.apache.paimon.utils.HintFileUtils;
+import org.apache.paimon.utils.SnapshotManager;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.Collections;
+import java.util.HashMap;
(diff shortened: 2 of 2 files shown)


aidadenski:
@JingsongLi  hi. could you take a quick look at the failing CI (it seems flaky/unrelated) and rerun if needed? This PR adds a cheap guard for the case where rename already succeeded but the ACK was lost (unstable network/client), preventing wrong manifest-list deletion. Thanks!

JingsongLi (commented):
Good resilience improvement for object storage environments where rename can fail but the target file still exists.

**The scenario**: On some object stores (S3, Azure), `rename` can timeout or return false even though the underlying PUT/COPY succeeded. The snapshot file exists but `tryToWriteAtomic` reports failure.

**Fix**: After `tryToWriteAtomic` returns false:
1. Check if the snapshot file exists
2. If yes, read it and compare with what we tried to write
3. If content matches → treat as committed (our write succeeded despite the false return)
4. If file doesn't exist → throw IOException (genuine failure)

**Review:**

1. **Correctness**: The `snapshot.equals(Snapshot.fromJson(...))` comparison is the right approach — it confirms our exact content was written, not someone else's conflicting snapshot.

2. **Race condition**: Between `tryToWriteAtomic` returning false and `fileIO.exists()`, another committer could write a different snapshot with the same ID. The `equals()` check handles this correctly — if content differs, we don't claim success.

3. **Performance**: Reading the file back adds latency on the commit path, but only in the failure case. Acceptable.

4. **Test coverage**: The `AlwaysFalseRenameLocalFileIO` mock elegantly tests both cases:
   - Rename returns false but file exists (success case)
   - Rename returns false and file doesn't exist (failure case)

5. **Missing case**: What if `fileIO.readFileUtf8(newSnapshotPath)` throws IOException (e.g., file exists but is corrupted/partial)? This would propagate as an exception — which is correct (fail loudly on ambiguous state).

LGTM. Clean implementation with good test coverage.

JingsongLi:
+1

PR #6780, opened 2025-12-09
[core] Fix retry in Consumer and SnapshotManager

JingsongLi:
### Purpose

Retry not works, because `Consumer.fromJson` throws `UncheckIOException` instead of `MismatchedInputException`.


yuzelin (approved):
+1

PR #6783, opened 2025-12-09
[hive] Fix hive catalog lock may encounter deadlock.

### LINKED ISSUE #6782: [Bug] HiveCatalogLock may experience deadlock
### Search before asking

### Paimon version

1.2.0

### Compute Engine

flink 1.20.3

### Minimal reproduce step

This can occur with a small probability during Flink task runtime and is difficult to reproduce.

I've found that when using Flink tasks to write to the Paimon table, there's a chance of a large number of checkpoint failures, and the table gets locked. I'm using object storage OSS, and the error is as follows:

Image

I guess the problem is here:
https://github.com/apache/paimon/blob/69b9f0dc363f1b9e8d67b3b32d36a947ccfe33aa/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalogLock.java#L64-L72

### Anything else?

_No response_

zhuangchong:
### Purpose

Linked issue: close #6782 

There are many possible causes for “hive lock may encounter deadlock”, for example:
1.The first table task is still running, and subsequent tasks cannot acquire the Hive table lock, leading to a timeout.
2.Delays in acquiring the Hive metastore lock, which also cause timeouts.
And so on.

**In my latest changes:**

1.Added detailed logs to show whether a lock acquisition failure was caused by a timeout or by another lock state.
https://github.com/apache/paimon/blob/3342ef9fa1a78d5a7d41424689fed5dbdf210986/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalogLock.java#L116-L120

2.Fixed an issue where lockResponse = clients.run(client -> client.checkLock(lockId)); would throw an exception and the lock would not be released, preventing subsequent tasks from acquiring the lock.

https://github.com/apache/paimon/blob/3342ef9fa1a78d5a7d41424689fed5dbdf210986/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalogLock.java#L95-L112

### DIFF
--- a/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalogLock.java
+++ b/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalogLock.java
@@ -66,22 +66,11 @@ public class HiveCatalogLock implements CatalogLock {
 
     @Override
     public <T> T runWithLock(String database, String table, Callable<T> callable) throws Exception {
-        Long lockId = null;
+        long lockId = lock(database, table);
         try {
-            lockId = lock(database, table);
             return callable.call();
         } finally {
-            if (lockId != null) {
-                safeUnlock(lockId);
-            }
-        }
-    }
-
-    private void safeUnlock(long lockId) {
-        try {
-            clients.execute(client -> client.unlock(lockId));
-        } catch (Exception e) {
-            LOG.warn("Unlock failed for lockId={}", lockId, e);
+            unlock(lockId);
         }
     }
 
@@ -91,6 +80,8 @@ public class HiveCatalogLock implements CatalogLock {
                 new LockComponent(LockType.EXCLUSIVE, LockLevel.TABLE, database);
         lockComponent.setTablename(table);
         lockComponent.unsetOperationType();
+
+        long startMs = System.currentTimeMillis();
         final LockRequest lockRequest =
                 new LockRequest(
                         Collections.singletonList(lockComponent),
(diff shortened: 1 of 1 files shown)


paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalogLock.java
```diff
@@ -63,11 +66,22 @@ public HiveCatalogLock(
 
     @Override
     public <T> T runWithLock(String database, String table, Callable<T> callable) throws Exception {
-        long lockId = lock(database, table);
+        Long lockId = null;
         try {
+            lockId = lock(database, table);
```
JingsongLi: Just for `runWithLock` method:
I cannot get it why modify here. What is difference?

zhuangchong: There are many possible causes for “hive lock may encounter deadlock”, for example:

The first table task is still running, and subsequent tasks cannot acquire the Hive table lock, leading to a timeout.

Delays in acquiring the Hive metastore lock, which also cause timeouts.
And so on.

In my latest changes:

Added detailed logs to show whether a lock acquisition failure was caused by a timeout or by another lock state.

Fixed an issue where `lockResponse = clients.run(client -> client.checkLock(lockId));` would throw an exception and the lock would not be released, preventing subsequent tasks from acquiring the lock.

zhuangchong:
> Can you update the Purpose in the description? I cannot get information in `try lock method`.

done.

PR #6784, opened 2025-12-09
[core] add audit fields into partitions sys table

XiaoHongbo-Hope:
### Purpose
Add audit fields into partitions sys table

### Tests
- `RestPartitionsTableTest`
- `RestCatalogTest.testReadPartitionsTable`

### API and Format
For rest catalog, partitions sys table should be like this: 
image

For other catalog, partitions sys table should be like this: 
image

### Documentation
`system-tables.md`

### Other fix
Fix list partitions return empty list issue when all files are level 0 with dv enabled, check https://github.com/apache/paimon/pull/6531 for details

### DIFF
--- a/paimon-api/src/main/java/org/apache/paimon/partition/Partition.java
+++ b/paimon-api/src/main/java/org/apache/paimon/partition/Partition.java
@@ -23,6 +23,7 @@ import org.apache.paimon.annotation.Public;
 import org.apache.paimon.shade.jackson2.com.fasterxml.jackson.annotation.JsonCreator;
 import org.apache.paimon.shade.jackson2.com.fasterxml.jackson.annotation.JsonGetter;
 import org.apache.paimon.shade.jackson2.com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import org.apache.paimon.shade.jackson2.com.fasterxml.jackson.annotation.JsonInclude;
 import org.apache.paimon.shade.jackson2.com.fasterxml.jackson.annotation.JsonProperty;
 
 import javax.annotation.Nullable;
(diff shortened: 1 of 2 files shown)


paimon-core/src/main/java/org/apache/paimon/table/system/PartitionsTable.java
```diff
@@ -250,6 +254,14 @@ private InternalRow toRow(
                         .append(partitionValueString);
             }
 
+            String createdBy = fileStoreTable.options().get(FIELD_CREATED_BY);
```
JingsongLi: This is wrong, we should get real partition createdBy from catalog server.

We should modify rest protocol to add these information.

XiaoHongbo-Hope: > This is wrong, we should get real partition createdBy from catalog server.
> 
> We should modify rest protocol to add these information.

Got it

docs/static/rest-catalog-open-api.yaml
```diff
@@ -2990,6 +2990,17 @@ components:
           format: int64
         done:
           type: boolean
+        createdAt:
+          type: integer
+          format: int64
+        createdBy:
+          type: string
+        updatedBy:
```
JingsongLi, line 3256: Why not add updatedAt?

XiaoHongbo-Hope: > Why not add updatedAt?

Got it.

XiaoHongbo-Hope: > Why not add updatedAt?

Added

paimon-api/src/main/java/org/apache/paimon/partition/Partition.java
```diff
@@ -25,38 +25,127 @@
...
-    private static final long serialVersionUID = 2L;
+    private static final long serialVersionUID = 3L;
 
     public static final String FIELD_DONE = "done";
+    public static final String FIELD_OPTIONS = "options";
 
     @JsonProperty(FIELD_DONE)
     private final boolean done;
 
+    @Nullable
```
JingsongLi: `@JsonInclude(JsonInclude.Include.NON_NULL)` for all new fields.

XiaoHongbo-Hope: > `@JsonInclude(JsonInclude.Include.NON_NULL)` for all new fields.

👌

XiaoHongbo-Hope: > `@JsonInclude(JsonInclude.Include.NON_NULL)` for all new fields.

added

PR #6786, opened 2025-12-10
[spark] Introduce `read.allow.fullScan`

Zouxxyy:
### Purpose

Whether to allow full scan when reading a partitioned table.


paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonBaseScan.scala
```diff
@@ -121,6 +121,23 @@ abstract class PaimonBaseScan(
...
 
+  private def ensureNoFullScan(): Unit = {
+    if (OptionUtils.readAllowFullScan()) {
+      return
+    }
+
+    table match {
+      case t: FileStoreTable if !t.partitionKeys().isEmpty =>
+        val skippedFiles = paimonMetricsRegistry.buildSparkScanMetrics().collectFirst {
+          case m: PaimonSkippedTableFilesTaskMetric => m.value
```
JingsongLi, line 132: If data skipping using non-partition cols?

Zouxxyy: Any file skipped is not considered a full scan, we are not currently distinguishing between partitioned and non-partitioned filtering.

PR #6790, opened 2025-12-10
[core] fix the wrong parsing for map type with string sub type

YannByron:
### Purpose

`col MAP<BIGINT, STRING> DEFAULT map(9999, "hello")`

Without this pr, this value part of the default value is `"hello"`, but it should be `hello`.

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/casting/StringToMapCastRule.java
+++ b/paimon-common/src/main/java/org/apache/paimon/casting/StringToMapCastRule.java
@@ -176,7 +176,7 @@ class StringToMapCastRule extends AbstractCastRule<BinaryString, InternalMap> {
                 : castExecutor.cast(BinaryString.fromString(valueStr));
     }
 
-    private List<String> splitMapEntries(String content) {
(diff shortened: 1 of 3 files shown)


paimon-common/src/main/java/org/apache/paimon/casting/StringToMapCastRule.java
```diff
@@ -176,7 +176,7 @@ private Object parseValue(String valueStr, CastExecutor<BinaryString, Object> ca
                 : castExecutor.cast(BinaryString.fromString(valueStr));
     }
 
-    private List<String> splitMapEntries(String content) {
+    public List<String> splitMapEntries(String content) {
```
JingsongLi, line 179: Maybe we can introduce something like Spark `CollationAwareUTF8String.splitSQL`.

YannByron: Let's research it, but can we deal with it with this tiny change?

JingsongLi:
+1

PR #6792, opened 2025-12-10
[rest] Add tag rest api definition

qingwei727:
### Purpose
Milvus integrates with Paimon via REST API. To resolve the snapshot expiration issue, REST APIs for tag management need to be provided.


JingsongLi (approved):
+1

PR #6795, opened 2025-12-11
[core] If overwrite an empty partition, change commit kind to APPEND

yuzelin:
### Purpose

Linked issue: close #xxx


paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
```diff
@@ -845,6 +845,27 @@ private int tryCommit(
         while (true) {
             Snapshot latestSnapshot = snapshotManager.latestSnapshot();
             CommitChanges changes = changesProvider.provide(latestSnapshot);
+            // If overwrite an empty partition, it can be considered as append
+            if (commitKind == CommitKind.OVERWRITE) {
+                boolean noDelete = true;
```
JingsongLi: Use `containsFileDeletionOrDeletionVectors`.

JingsongLi (approved):
+1

PR #6798, opened 2025-12-11
[spark] Add read batch time metric

Zouxxyy:
### Purpose

image


JingsongLi:
What does Spark look like when reading a Parquet Table? (You can take a screenshot)

Zouxxyy:
> What does Spark look like when reading a Parquet Table? (You can take a screenshot)

Like this, I fix the split and partition size metrics for the format table too

image

PR #6806, opened 2025-12-12
[Flink] Add some debug logs.

wwj6591812:
### Purpose

When we use flink batch + Paimon in olap. We found that there is a significant difference in query time for different partitions of a table. We attempted to analyze the reasons through logs and querying system tables, but were unsuccessful. Therefore, we added some debug logs.

`/*+ config(cluster=yinxia-test, job_name=20251202) */
SELECT * FROM paimon_ea120.sample.xfc_monitor_cbu_algo_mainse_sa_union_ctr_compare_new_diff 
/*+ OPTIONS('scan.dedicated-split-generation'='false', 'scan.partitions'='ds=20251202') */ limit 10

/*+ config(cluster=yinxia-test, job_name=20251203) */
SELECT * FROM paimon_ea120.sample.xfc_monitor_cbu_algo_mainse_sa_union_ctr_compare_new_diff 
/*+ OPTIONS('scan.dedicated-split-generation'='false', 'scan.partitions'='ds=20251203') */ limit 10`

Linked issue: close #xxx

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/source/DataTableBatchScan.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/source/DataTableBatchScan.java
@@ -31,6 +31,9 @@ import org.apache.paimon.table.source.snapshot.StartingScanner;
 import org.apache.paimon.table.source.snapshot.StartingScanner.ScannedResult;
 import org.apache.paimon.types.DataType;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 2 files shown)


JingsongLi (commented):
Adding debug logs for troubleshooting partition-specific query latency differences is understandable, but:

1. **233 additions for "some debug logs"** seems excessive. What exactly is being logged? If this is adding detailed per-split timing, per-file statistics, or per-source logging, that's substantial observability work — not just "some debug logs."

2. **Log levels**: Ensure these are at DEBUG level (not INFO). Production environments with many partitions could be flooded if these are at INFO.

3. **PR description includes internal query details** (cluster names, table names) — consider removing these from the public PR.

4. **Missing linked issue**: The PR template shows `close #xxx` without an actual issue. Consider creating one for traceability.

5. **No tests**: Debug logging changes typically don't need tests, but if the logs include computed metrics (timing, counts), consider testing the metric computation logic separately.

Please clarify the scope and verify all logs are at DEBUG level.

PR #6807, opened 2025-12-12
[vector] support vector search

jerry-024:
### Purpose

The PR implements vector search.
- usage
```java
    float[] queryVector = new float[] {0.85f, 0.15f};
    VectorSearch vectorSearch = new VectorSearch(queryVector, 2, vectorFieldName);
    ReadBuilder readBuilder = table.newReadBuilder().withVectorSearch(vectorSearch);
    TableScan scan = readBuilder.newScan();
    readBuilder.newRead().createReader(scan.plan())
```

### Tests
- LuceneVectorGlobalIndexTest
- LuceneVectorGlobalIndexScanTest


paimon-lucene/src/test/java/org/apache/paimon/lucene/index/LuceneVectorGlobalIndexScanTest.java
```diff
@@ -0,0 +1,224 @@
...
+
+        // 3. Manually build vector index
+        List<IndexFileMeta> indexFiles = buildIndexManually(vectors);
+
+        // 4. Commit index files to the Table (Update Index Manifest)
+        commitIndex(indexFiles);
+
+        // 5. Use GlobalIndexScanBuilder to get scanner
+        GlobalIndexScanBuilder scanBuilder = table.store().newGlobalIndexScanBuilder();
+        List<org.apache.paimon.utils.Range> ranges = scanBuilder.shardList();
```
leaves12138: just Range

jerry-024:
> @jerry-024 Vector search and future Full-text search should inherit from TopN

@hang8929201 Thanks for the proposal. However, regarding the class hierarchy, I wonder if it might be better to keep VectorSearch (and future Full-text search) distinct from TopN, rather than inheriting directly. Here are a few thoughts for your consideration:

1. Semantics of orders vs. similarity TopN is typically defined by explicit orders (SortValues). Vector Search, however, operates a bit differently—it focuses on finding the "nearest" matches based on distance calculations rather than sorting by scalar columns. If we inherit from TopN, we might end up carrying the orders field, which isn't strictly necessary for vector queries and might obscure the intent of the search.

2. Architecture in other systems (Lance & PG) Looking at how other systems handle this might offer some perspective:

LanceDB: Treats vector search as a specialized operation. The API usually separates the search(vector) logic from standard sorting or limiting to allow for vector-specific parameters (like probe counts or refinement steps).

PostgreSQL (pgvector): While it uses the ORDER BY syntax, the internal execution plan typically utilizes a specialized k-NN Index Scan operator, which functions quite differently from a standard Heap Sort TopN.

It seems that treating VectorSearch as an independent query type (sibling to TopN), while still accepting the predicate result for filtering, might offer us more flexibility in the future.

paimon-core/src/main/java/org/apache/paimon/globalindex/GlobalIndexScanBuilder.java
```diff
@@ -53,6 +54,7 @@ static Optional<GlobalIndexResult> parallelScan(
             final List<Range> ranges,
             final GlobalIndexScanBuilder globalIndexScanBuilder,
             final Predicate filter,
+            final VectorSearch vectorSearch,
```
leaves12138: @Nullable

paimon-common/src/main/java/org/apache/paimon/fileindex/FileIndexReader.java
```diff
@@ -117,6 +118,11 @@ public FileIndexResult visitNotIn(FieldRef fieldRef, List<Object> literals) {
         return fileIndexResult;
     }
 
+    @Override
+    public FileIndexResult visitVectorSearch(VectorSearch vectorSearch) {
+        throw new UnsupportedOperationException("Should not invoke this");
+    }
```
leaves12138: Why add visitVectorSearch to FunctionVisitor?

paimon-common/src/main/java/org/apache/paimon/predicate/OnlyPartitionKeyEqualVisitor.java
```diff
@@ -117,6 +117,11 @@ public Boolean visitNotIn(FieldRef fieldRef, List<Object> literals) {
         return false;
     }
 
+    @Override
+    public Boolean visitVectorSearch(VectorSearch vectorSearch) {
+        return false;
```
leaves12138: Add this to GlobalIndexReader?

PR #6811, opened 2025-12-12
[core] Allow change ignore-delete/ignore-update-before from false to true

yuzelin:
### Purpose

When users created a partial-update/aggregation table, they might not notice that some aggregate function cannot accept -D record, and if the `ignore-delete` option is not changeable, they have to recreate the table. This PR allow them to change the option to true from false.

Also change the `ignore-update-before`.


JingsongLi:
+1

PR #6814, opened 2025-12-15
[spark] Fix spark write metrics with spark adaptive plan

Zouxxyy:
### Purpose

executionMetrics may have same ids which will make metrics multiply many times


JingsongLi:
+1

PR #6817, opened 2025-12-15
[spark] Split pushedFilters into dataFilters and partitionFilters in ScanBuilder

Zouxxyy:
### Purpose

image

- Split `pushedFilters` into `pushedPartitionFilters` and `pushedDataFilters` in ScanBuilder
- Add equals and hashCode in `PartitionPredicate ` and `TopN `
- Use `rowCount` to Compute `sizeInBytes` when analyzed, so that we can remove `filterStatistics(stats, reservedFilters)` (Because this is still not very accurate and is very much a hack.)

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/ColumnPruningAndPushDown.scala
+++ b/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/ColumnPruningAndPushDown.scala
@@ -20,7 +20,6 @@ package org.apache.paimon.spark
 
 import org.apache.paimon.CoreOptions
 import org.apache.paimon.partition.PartitionPredicate
-import org.apache.paimon.partition.PartitionPredicate.AndPartitionPredicate
 import org.apache.paimon.predicate.{Predicate, TopN}
 import org.apache.paimon.spark.schema.PaimonMetadataColumn
 import org.apache.paimon.spark.schema.PaimonMetadataColumn._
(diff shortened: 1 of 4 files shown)


paimon-core/src/main/java/org/apache/paimon/partition/PartitionValuesTimeExpireStrategy.java
```diff
@@ -117,5 +118,20 @@ public boolean test(
                 InternalArray nullCounts) {
             return true;
         }
+
+        @Override
+        public boolean equals(Object o) {
```
JingsongLi, line 123: Why add this?

Zouxxyy: Because Spark relies on the equals method of `PaimonScan` to determine whether it can be reused, all of its fields must properly implement equals. The `Predicate` has already been implemented equals, and PartitionPredicate needs to be handled as well. 

```scala
case class PaimonScan(
    table: InnerTable,
    requiredSchema: StructType,
    pushDownPartitionFilters: Seq[PartitionPredicate],
    pushDownDataFilters: Seq[Predicate],
    override val pushDownLimit: Option[Int],
    override val pushDownTopN: Option[TopN],
    bucketedScanDisabled: Boolean = false)
```

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/ColumnPruningAndPushDown.scala
```diff
@@ -82,12 +92,14 @@ trait ColumnPruningAndPushDown extends Scan with Logging {
 
   lazy val readBuilder: ReadBuilder = {
     val _readBuilder = table.newReadBuilder().withReadType(readTableRowType)
-    if (filters.nonEmpty) {
-      val pushedPredicate = PredicateBuilder.and(filters: _*)
-      _readBuilder.withFilter(pushedPredicate)
+    if (pushedPartitionFilters.nonEmpty) {
+      _readBuilder.withPartitionFilter(new AndPartitionPredicate(pushedPartitionFilters.asJava))
```
JingsongLi: Do not wrap to `AndPartitionPredicate`, if size is 1, we can introduce `PartitionPredicate.and` static method.

JingsongLi (approved):
+1

PR #6820, opened 2025-12-16
[core] Combine range for global index reader

leaves12138:
### Purpose

We need to combine the same range for global index result reader, then pass them to. We should not pass all different ranges index meta to global index reader


JingsongLi (approved):
+1

PR #6821, opened 2025-12-16
[core] Support ignore corrupt or lost files during read

baiyangtx:
### Purpose

Introduce options `scan.ignore-corrupt-files` and `scan.ignore-lost-files` to skip data files with problems.

### Tests

- paimon-spark/paimon-spark-ut/src/test/scala/org/apache/paimon/spark/sql/PaimonQueryTest.scala

### API and Format

No

### Documentation

No


baiyangtx:
@Aitozi @JingsongLi  PTAL

In production environments, we may encounter various unexpected situations that cause file corruption or loss. To quickly resume downstream tasks, we introduce parameters to ignore these files.

Aitozi (approved):
LGTM +1

PR #6823, opened 2025-12-16
[spark] Add custom topology builder for global index builder

leaves12138:
### Purpose

Linked issue: close #xxx


paimon-common/src/main/java/org/apache/paimon/globalindex/GlobalIndexWriter.java
```diff
@@ -29,6 +29,10 @@ public interface GlobalIndexWriter {
 
     void write(@Nullable Object key);
 
+    default void write(@Nullable Object key, long rowId) {
+        write(key);
```
JingsongLi: Do not write by default, it is wrong.

leaves12138: OK

PR #2749, opened 2024-01-20
[Flink]add range strategy for sort compaction

wg1026688210:
### Purpose

Improve the processing efficiency of sort compaction operations when there is data size skewness on range phase.

1. add  the  size based rangestrategy config.
2. add sample ratio config for per task and range num config.

### Tests

SortCompactActionForUnawareBucketITCase
RangeShuffleTest
InternalRowToSizeVisitorTest

### API and Format

add config  
``` --table_conf sort-compaction.range-strategy= SIZE ```

PR #6826, opened 2025-12-18
[core] change range strategy of clustering to 'SIZE'

LsomeYeah:
### Purpose

Linked issue: close #xxx

As https://github.com/apache/paimon/pull/2749 mentioned, if the sizes of records vary significantly, range partitioning based on the number of records may cause data skew, which further reduces the processing efficiency of individual concurrency. In extreme cases, especially in Flink scenarios, this could fill up the TaskManager's local disk and cause task failure. 

Considering record size during range partitioning can alleviate this issue. This PR sets `SIZE` as the default strategy for range partitioning.


JingsongLi:
+1

PR #6827, opened 2025-12-18
[spark] Optimize MERGE INTO self-merge updates on dataEvolution table

wayneli-vt:
### Purpose

This PR optimizes `MERGE INTO` on Paimon row-tracking (data-evolution) append tables for the self-merge pattern:

```SQL
MERGE INTO target
USING target AS source
ON target._ROW_ID = source._ROW_ID
WHEN MATCHED THEN UPDATE SET col2 = col1 + 1
```

When `source` and `target` are the same table and the merge condition is `_ROW_ID` equality, the joined result is identical to scanning the target table itself. Therefore, we can avoid all `join/shuffle/sort` and only write the updated columns.

Changes included:

1. `targetRelatedSplits`: for self-merge on `_ROW_ID`, directly scan all splits instead of computing touched splits via source scan/join.
2. **Join Free**`updateActionInvoke`: for self-merge on `_ROW_ID`, remove the target-source join. Read the required columns from the target scan once, and rewrite all source-side attribute references in merge actions/conditions to target attributes so `MergeRows` can be evaluated on the target scan output.
3. **Reduce write-side shuffle/sort**: In this self-merge pattern, the scan output is already partitioned by `_FIRST_ROW_ID`: rows with the same `_FIRST_ROW_ID` are contiguous within a Spark partition, and within each `_FIRST_ROW_ID` group `_ROW_ID` is strictly increasing. Therefore, no extra shuffle or sort is needed and we can write partial column updates directly.

### Tests

A new test case has been added to `RowTrackingTestBase` to specifically verify this pr:

* `org.apache.paimon.spark.sql.RowTrackingTestBase#Data Evolution: merge into table with data-evolution with _ROW_ID shortcut`

### API and Format

No

### Documentation

No

### DIFF
--- a/paimon-spark/paimon-spark-ut/src/test/scala/org/apache/paimon/spark/sql/RowTrackingTestBase.scala
+++ b/paimon-spark/paimon-spark-ut/src/test/scala/org/apache/paimon/spark/sql/RowTrackingTestBase.scala
@@ -459,9 +459,6 @@ abstract class RowTrackingTestBase extends PaimonSparkTestBase {
              |WHEN MATCHED AND source.c > 'c2' THEN UPDATE SET b = source.b * 3,
              |c = concat(target.c, source.c)
              |""".stripMargin)
-      // Assert that job was triggered by
-      // `org.apache.paimon.spark.commands.MergeIntoPaimonDataEvolutionTable.targetRelatedSplits`
-      assert(capturedPlans.length == 2)
       // Assert no shuffle/join/sort was used in
       // 'org.apache.paimon.spark.commands.MergeIntoPaimonDataEvolutionTable.updateActionInvoke'
       assert(


wayneli-vt:
Thank you for the reply! @JingsongLi 

These optimizations are mainly for `computed-column backfill` scenarios. For example, users may have already ingested multi-modal data into a Paimon table and they want to use Spark to process the data, derive some features, and write them back to the same table. 

In this case, it should be possible to do `Scan -> Calculate -> Write` directly, because the scan output already matches the required layout for partial column updates, so we can write directly without extra `shuffle/sort/join`. This PR is focused on optimizing this scenario.

JingsongLi (approved):
+1

wayneli-vt:
Thank you for your review!

PR #6829, opened 2025-12-18
[hive] Preserve default partition name when copying Hive tables

Zouxxyy:
### Purpose

hive's default partition name is `__HIVE_DEFAULT_PARTITION__`, which is different from paimon, need to preserve it


JingsongLi:
+1

PR #6830, opened 2025-12-18
[core] Minor refactor the maintenance order in TableCommitImpl

yuzelin:
### Purpose

The auto tag creation is time sensitive, for example, creating every day at 24:00. But currently the snapshot expiration and partition expiration are performed before auto tag creation, and they might take a long time.

So it's better to place the auto tag management to the first.


JingsongLi:
+1

PR #6835, opened 2025-12-19
[core][flink] Support incremental clustering for append bucketed table

LsomeYeah:
### Purpose

Linked issue: close #xxx

By default, Paimon's append bucketed tables maintain data ordering. However, this ordering requirement can be relaxed to enable additional optimizations.

This PR introduces the ability to disable ordering requirements for append bucketed tables, allowing incremental clustering within buckets. When ordering is not strictly required, data can be incrementally clustered within each bucket, significantly improving query performance for `bucket-key + clustering-key` combinations.

Unlike append-unaware tables that require range partitioning, bucketed tables only need to shuffle by partition + bucket and perform local clustering within each bucket partition, making this approach much more efficient and resource-friendly.

### DIFF
--- a/docs/content/append-table/incremental-clustering.md
+++ b/docs/content/append-table/incremental-clustering.md
@@ -39,12 +39,15 @@ gradually converges to an optimal state, significantly reducing the decision-mak
 
 
 Incremental Clustering supports:
-- Support incremental clustering; minimizing write amplification as possible.
-- Support small-file compaction; during rewrites, respect target-file-size.
-- Support changing clustering keys; newly ingested data is clustered according to the latest clustering keys.
-- Provide a full mode; when selected, the entire dataset will be reclustered.
+- Support incremental clustering: minimizing write amplification as possible.
+- Support small-file compaction: during rewrites, respect target-file-size.
+- Support changing clustering keys: newly ingested data is clustered according to the latest clustering keys.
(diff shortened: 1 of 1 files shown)


paimon-core/src/main/java/org/apache/paimon/append/cluster/IncrementalClusterManager.java
```diff
@@ -123,196 +128,260 @@ public IncrementalClusterManager(
...
-                                                                lsr.run().files().size()))
-                                        .collect(Collectors.joining(","));
-                        LOG.debug(
-                                "Partition {} has {} runs: [{}]",
-                                partitionComputer.generatePartValues(partition),
-                                levelSortedRuns.size(),
-                                runsInfo);
-                    });
-        }
+        partitionLevels.forEach(
```
JingsongLi, line 345: Too deep. Use for loop.

JingsongLi (commented):
My idea is that the clustering of this Bucketed table is like merging small files of Append+Bucket-1 table, directly generating tasks for different concurrent writer to reads and writes separately.

LsomeYeah:
> Quote reply
> Refer

Good idea! I will encapsulate the local-sort operation within a single bucket into a Task and move it to paimon-core, which will also facilitate integration with multiple other engines.

PR #6844, opened 2025-12-19
[Fix]partial-update still keeps row when the last message is -U only

### LINKED ISSUE #6862: [Bug] Partial-update merge-function keeps stale row when the last message is a lone UPDATE_BEFORE (-U)
### Search before asking

### Paimon version

1.2.0

### Compute Engine

flink 1.20

##  Create the source changelog table
```
CREATE TABLE testlog(
  id STRING PRIMARY KEY NOT ENFORCED,
  f1 STRING,
  delete INT
) WITH (
  'merge-engine' = 'deduplicate',
  'changelog-producer' = 'lookup'
);

-- seed data
INSERT INTO testlog VALUES ('11', '11', 0), ('12', '12', 0);

-- update id=12 to be logically deleted
INSERT INTO testlog VALUES ('12', '12', 1), ('13', '13', 0);
```

## Query source table 
```
SELECT * FROM testlog;

op   id   f1   delete
+I   11   11   0
+I   12   12   0
-U   12   12   0
+U   12   12   1
+I   13   13   0
```

## Query source table with filter
```
SELECT * FROM testlog WHERE delete = 0;

op   id   f1   delete
+I    11   11   0
+I    12   12   0
-U    12   12   0
+I    13   13   0
```
Note:
After applying the filter delete = 0, the +U  12  12  1 is completely dropped.
Consequently, the only message about id = 12 that reaches the downstream Paimon table is -U 12 12 0.

## Filter out logically-deleted rows and write into a partial-update table
```
CREATE TABLE testlog01 (
  id STRING PRIMARY KEY NOT ENFORCED,
  f1 STRING,
  delete INT
) WITH (
  'merge-engine' = 'partial-update',
  'partial-update.remove-record-on-delete' = 'true',
  'changelog-producer' = 'lookup'
);

INSERT INTO testlog01
SELECT * FROM testlog WHERE delete = 0;
```

## Query the target table
```
SELECT * FROM testlog01;

```

## Expected result
```
op   id   f1   delete
+I   11   11   0
+I   12   12   0
-D   12   12   0
+I   13   13   0
```

(id=12 should delete because its last message is -U and no +U reaches the sink)

## Actual result
```
op   id   f1   delete
+I   11   11   0
+I   12   12   0
-U   12   12   0
+U   12   12   0
+I    13   13   0
```
The stale row id=12 is still present, breaking data correctness.

### Anything else?

_No response_

sandyfog:
### Purpose
In streaming pipelines, a common Flink SQL pattern is to filter rows before they reach the Paimon sink:

```
INSERT INTO paimon_table
SELECT * FROM upstream WHERE delete = 0;
```

When a row flips from delete = 0 to delete = 1, Flink first emits an UPDATE_BEFORE (-U) for the previously qualified row.
The succeeding **UPDATE_AFTER** (+U) is dropped by the filter, so only the -U record arrives at Paimon.
The current PartialUpdateMergeFunction does not treat this lone -U as a deletion.
Consequently, the stale row remains in the table, breaking correctness.

Linked issue: close #6862

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/PartialUpdateMergeFunction.java
+++ b/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/PartialUpdateMergeFunction.java
@@ -92,6 +92,18 @@ public class PartialUpdateMergeFunction implements MergeFunction<KeyValue> {
      */
     private boolean meetInsert;
 
+    /**
+     * Flag indicating whether the latest processed record is a retract operation. Used to determine
+     * the correct behavior when the last record in a sequence is a retract.
+     */
+    private boolean latestRetract = false;
+
+    /**
+     * Stores the latest KeyValue processed. This is used when we need to initialize the row based
+     * on the last retract record in case the final result should be a deletion.
+     */
+    private KeyValue latestKV;
+
     protected PartialUpdateMergeFunction(
             InternalRow.FieldGetter[] getters,
             boolean ignoreDelete,
@@ -118,6 +130,8 @@ public class PartialUpdateMergeFunction implements MergeFunction<KeyValue> {
         this.notNullColumnFilled = false;
         this.row = new GenericRow(getters.length);
         this.latestSequenceNumber = 0;
+        this.latestRetract = false;
+        this.latestKV = null;
         fieldAggregators.forEach(w -> w.getValue().reset());
     }
 
@@ -125,6 +139,11 @@ public class PartialUpdateMergeFunction implements MergeFunction<KeyValue> {
     public void add(KeyValue kv) {
         // refresh key object to avoid reference overwritten
         currentKey = kv.key();
+
+        // Update latestKV and latestRetract status
+        latestKV = kv;
+        latestRetract = false;
+
         currentDeleteRow = false;
         if (kv.valueKind().isRetract()) {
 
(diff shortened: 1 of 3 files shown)


JingsongLi (commented):
Important correctness fix for a real-world streaming pattern. The scenario where Flink's filter drops the +U but passes the -U is well-documented in the PR description.

**Review:**

1. **Root cause is clear**: When `partial-update.remove-record-on-delete=true` and only a -U arrives (no subsequent +U or DELETE), the old code didn't treat this as a deletion.

2. **Fix approach**: Adding `latestRetract` flag and deferring row initialization to `getResult()`:
   - `latestRetract = true` when any retract is the final record in the merge sequence
   - In `getResult()`, if `latestRetract` is true, emit a DELETE row
   
3. **Concern — interaction with subsequent INSERTs**: The `latestRetract` flag is set per `add()` call but never reset when a subsequent INSERT arrives (it's only reset in `reset()`). Wait — looking more carefully, I see `latestRetract = false` is set at the start of every `add()` call, then set to `true` only in the retract path. So it reflects only the *latest* record's retract status. This is correct.

4. **`latestKV` field**: Tracks the most recent KeyValue for row initialization. This adds a reference per merge function instance — negligible memory.

5. **Sequence group interaction**: The fix also handles `retractWithSequenceGroup` — setting `latestRetract = true` when the sequence group contains a partial-delete field. The test `testUpdateBeforeRetractSequenceGroupPartialDelete` covers this.

6. **Test coverage**: Three test cases covering:
   - Simple UPDATE_BEFORE retract (no sequence groups)
   - Sequence group retract (partial fields)
   - Sequence group with partial-delete

**One edge case question**: What happens if `add()` is called with only a single -U record (no prior INSERT)? Should this still produce a DELETE row? The `meetInsert` flag would be false, so `getResult()` returns `RowKind.DELETE` anyway. Seems correct.

LGTM.

JingsongLi:
Please rebase master

PR #6850, opened 2025-12-20
[common] Unify hash function num calculation in BloomFilterFileIndex using bitwise operations

xiaochen-zhou:
### Purpose

Unify hash function num calculation in BloomFilterFileIndex using bitwise operations

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/fileindex/bloomfilter/BloomFilterFileIndex.java
+++ b/paimon-common/src/main/java/org/apache/paimon/fileindex/bloomfilter/BloomFilterFileIndex.java
@@ -120,9 +120,9 @@ public class BloomFilterFileIndex implements FileIndexer {
             // big endian
             int numHashFunctions =
                     ((serializedBytes[0] << 24)
-                            + (serializedBytes[1] << 16)
-                            + (serializedBytes[2] << 8)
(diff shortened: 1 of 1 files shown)


JingsongLi (commented):
This changes `+` to `|` for big-endian byte reconstruction of `numHashFunctions`.

**Analysis:** For unsigned byte assembly, `+` and `|` produce identical results when each component occupies a distinct byte position (no overlap). The change is cosmetically cleaner (bitwise OR is the idiomatic pattern for bit-field assembly) but functionally equivalent.

However, there's a subtle correctness issue with the current code that this PR doesn't fix: `serializedBytes[1]`, `serializedBytes[2]`, and `serializedBytes[3]` are Java `byte` values (signed, -128 to 127). When a byte is negative and shifted left, the sign extension fills high bits with 1s. For example:
- `serializedBytes[3]` with value `0xFF` (-1 as signed byte) becomes `0xFFFFFFFF` as an int, not `0x000000FF`.

The original code with `+` has the same issue. Both `+` and `|` need a `& 0xFF` mask on each byte before shifting. Was this intentional to leave unfixed, or is it a known constraint (e.g., numHashFunctions is always small)?

If `numHashFunctions` is always < 128, this is fine. Otherwise, both the old and new code have the same sign-extension bug. Please verify.

PR #6855, opened 2025-12-22
[arrow] Introduce Arrow native read & write interface

yuzelin:
### Purpose

Apache Arrow is the universal columnar format and multi-language toolbox for fast data interchange, and paimon has introduced some converters for transform Paimon data from/to Arrow format data. But Paimon still lack utils for directly reading arrow data from file or writing arrow data to file.

This PR introduce two native interface:
1. `NativeReader`: Read arrow data from file. Then you can use converter to convert arrow data to Paimon data;
2. `NativeWriter`: Write arrow data to file. You can first use converter to convert Paimon data to arrow data, then use NativeWriter to write file (this PR also provide a ArrowBundleWriter implementation).

With this PR, users can implement faster file reader and writer based on Apache Arrow with other language (like C++) and integrate with Paimon.

This PR also add a `JNIUtils` for helping loading the JNI files.

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/utils/JNIUtils.java
+++ b/paimon-common/src/main/java/org/apache/paimon/utils/JNIUtils.java
@@ -28,7 +28,7 @@ import java.util.logging.Logger;
 
 /**
  * Utils for loading jni. The jni file is placed at {@code /osName/osArch/jniName.libExtension}.
- * Uses should pass jniName through {@code load}.
+ * User should pass jniName through {@code load}.
  */
 public class JNIUtils {
 


paimon-common/src/main/java/org/apache/paimon/utils/JNIUtils.java
```diff
@@ -0,0 +1,141 @@
...
+import java.io.InputStream;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+import java.util.logging.Logger;
+
+/**
+ * Utils for loading jni. The jni file is placed at {@code /osName/osArch/jniName.libExtension}.
+ * Uses should pass jniName through {@code load}.
+ */
+public class JNIUtils {
```
JingsongLi, line 33: Move this to paimon-common?

JingsongLi:
+1

PR #6858, opened 2025-12-22
[core] Support reading sequence_number in AuditLogTable and BinlogTable

wzhero1:
### Purpose

This PR adds support for accessing system fields (specifically `sequence_number`) in `AuditLogTable` and `BinlogTable`.

Currently, downstream consumers often need the `sequence_number` to ensure strict ordering or to implement exactly-once processing when reading from audit logs or binlogs. However, these system tables previously did not expose this field.

This change implements `KeyValueSystemFieldsRecordReader` to enable the projection of system fields for KeyValue records, allowing users to query `sequence_number` directly from audit log and binlog tables.

### Tests

- Added `IncrementalReadSystemFieldsTest` to verify functionality.
- Updated `AuditLogTableTest` and `BinlogTableTest` to cover `sequence_number` read cases.

### API and Format

- No change to storage format.
- No public API change.

### Documentation

- `docs/content/concepts/system-tables.md` needs to change

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/MergeFileSplitRead.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/MergeFileSplitRead.java
@@ -50,7 +50,7 @@ import org.apache.paimon.table.BucketMode;
 import org.apache.paimon.table.source.ChainSplit;
 import org.apache.paimon.table.source.DataSplit;
 import org.apache.paimon.table.source.DeletionFile;
-import org.apache.paimon.table.source.KeyValueSystemFieldsRecordReader;
+import org.apache.paimon.table.source.KeyValueSpecialFieldsRecordReader;
 import org.apache.paimon.table.source.Split;
 import org.apache.paimon.types.DataField;
 import org.apache.paimon.types.RowType;
@@ -99,7 +99,7 @@ public class MergeFileSplitRead implements SplitRead<KeyValue> {
     @Nullable private int[][] outerProjection;
     @Nullable private VariantAccessInfo[] variantAccess;
 
-    private List<KeyValueSystemFieldsRecordReader.SystemFieldExtractor> systemFieldExtractors =
+    private List<KeyValueSpecialFieldsRecordReader.SpecialFieldExtractor> specialFieldExtractors =
             Collections.emptyList();
 
     @Nullable private int[] projection = null;
@@ -144,8 +144,9 @@ public class MergeFileSplitRead implements SplitRead<KeyValue> {
         return this;
     }
 
-    public List<KeyValueSystemFieldsRecordReader.SystemFieldExtractor> getSystemFieldExtractors() {
-        return systemFieldExtractors;
+    public List<KeyValueSpecialFieldsRecordReader.SpecialFieldExtractor>
+            getSystemFieldExtractors() {
+        return specialFieldExtractors;
     }
 
     @Nullable
@@ -155,7 +156,7 @@ public class MergeFileSplitRead implements SplitRead<KeyValue> {
 
     @Override
     public MergeFileSplitRead withReadType(RowType readType) {
-        this.systemFieldExtractors = collectSystemFieldExtractors(readType);
+        this.specialFieldExtractors = collectSystemFieldExtractors(readType);
         this.projection = createProjection(readType);
 
         // todo: replace projectedFields with readType
@@ -435,16 +436,17 @@ public class MergeFileSplitRead implements SplitRead<KeyValue> {
      * @param readType the requested read type (may contain system fields)
      * @return list of extractors for system fields present in readType
      */
-    private List<KeyValueSystemFieldsRecordReader.SystemFieldExtractor>
+    private List<KeyValueSpecialFieldsRecordReader.SpecialFieldExtractor>
             collectSystemFieldExtractors(RowType readType) {
         if (readType == null) {
             return Collections.emptyList();
         }
 
-        List<KeyValueSystemFieldsRecordReader.SystemFieldExtractor> extractors = new ArrayList<>();
+        List<KeyValueSpecialFieldsRecordReader.SpecialFieldExtractor> extractors =
+                new ArrayList<>();
         for (String fieldName : readType.getFieldNames()) {
-            KeyValueSystemFieldsRecordReader.SystemFieldExtractor extractor =
-                    KeyValueSystemFieldsRecordReader.getExtractor(fieldName);
+            KeyValueSpecialFieldsRecordReader.SpecialFieldExtractor extractor =
+                    KeyValueSpecialFieldsRecordReader.getExtractor(fieldName);
             if (extractor != null) {
                 extractors.add(extractor);
             }
@@ -455,7 +457,7 @@ public class MergeFileSplitRead implements SplitRead<KeyValue> {
     /**
      * Creates a projection array to reorder fields from natural order to requested order.
      *
-     * <p>Example: readType = [pt, rowkind, col1], systemFieldExtractors = [rowkind] Natural order:
+     * <p>Example: readType = [pt, rowkind, col1], specialFieldExtractors = [rowkind] Natural order:
      * [rowkind(0), pt(1), col1(2)] (physical fields pt, col1 in readType order) Requested order:
      * [pt, rowkind, col1] Projection: [1, 0, 2]
      *
@@ -464,7 +466,7 @@ public class MergeFileSplitRead implements SplitRead<KeyValue> {
      */
     @Nullable
     private int[] createProjection(RowType readType) {
-        if (readType == null || systemFieldExtractors.isEmpty()) {
+        if (readType == null || specialFieldExtractors.isEmpty()) {
             return null;
         }
 
@@ -473,13 +475,13 @@ public class MergeFileSplitRead implements SplitRead<KeyValue> {
         // System fields are first in natural order
         int systemIdx = 0;
         // Physical fields follow system fields in natural order
-        int physicalIdx = systemFieldExtractors.size();
+        int physicalIdx = specialFieldExtractors.size();
         boolean needsProjection = false;
 
         for (int i = 0; i < readFieldNames.size(); i++) {
             String fieldName = readFieldNames.get(i);
             // Check if it's a system field
-            if (KeyValueSystemFieldsRecordReader.getExtractor(fieldName) != null) {
+            if (KeyValueSpecialFieldsRecordReader.getExtractor(fieldName) != null) {
                 projection[i] = systemIdx++;
             } else {
                 projection[i] = physicalIdx++;
--- a/paimon-core/src/main/java/org/apache/paimon/table/source/KeyValueSpecialFieldsRecordReader.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/source/KeyValueSpecialFieldsRecordReader.java
@@ -35,44 +35,44 @@ import java.util.List;
 import java.util.Map;
 
 /**
- * A decorator for {@link RecordReader} that injects system fields into the output rows for
+ * A wrapper for {@link RecordReader} that injects special fields into the output rows for
  * KeyValue-based data sources.
  *
  * <p>This reader wraps a {@code RecordReader<KeyValue>} and produces {@code
- * RecordReader<InternalRow>} with additional system fields extracted from the KeyValue metadata.
+ * RecordReader<InternalRow>} with additional special fields extracted from the KeyValue metadata.
  *
  * <p><b>Naming:</b> This class is specifically designed for KeyValue format data (e.g.,
- * MergeFileSplitRead). For InternalRow readers (e.g., RawFileSplitRead), system fields are handled
+ * MergeFileSplitRead). For InternalRow readers (e.g., RawFileSplitRead), special fields are handled
  * differently in {@link org.apache.paimon.io.DataFileRecordReader} using file metadata.
  *
  * <p><b>Field Ordering:</b> The output schema supports arbitrary field ordering. Internally, fields
- * are assembled as [system fields... + physical fields...], then reordered using {@link
+ * are assembled as [special fields... + physical fields...], then reordered using {@link
  * ProjectedRow} to match the requested field order.
  *
  * <p><b>Performance:</b> Implementation uses {@link JoinedRow} for zero-copy concatenation of
- * system fields and physical fields, then {@link ProjectedRow} for zero-copy field reordering.
+ * special fields and physical fields, then {@link ProjectedRow} for zero-copy field reordering.
  */
-public class KeyValueSystemFieldsRecordReader implements RecordReader<InternalRow> {
+public class KeyValueSpecialFieldsRecordReader implements RecordReader<InternalRow> {
 
     private final RecordReader<KeyValue> wrapped;
-    private final List<SystemFieldExtractor> systemFieldExtractors;
+    private final List<SpecialFieldExtractor> specialFieldExtractors;
     @Nullable private final int[] projection;
 
     /**
-     * Creates a KeyValueSystemFieldsRecordReader.
+     * Creates a KeyValueSpecialFieldsRecordReader.
      *
      * @param wrapped the underlying KeyValue reader
-     * @param systemFieldExtractors extractors for system fields, in order
+     * @param specialFieldExtractors extractors for special fields, in order
      * @param projection optional projection to reorder fields. If null, fields are in natural order
-     *     [system fields... + physical fields...]. If provided, projects from the natural order to
+     *     [special fields... + physical fields...]. If provided, projects from the natural order to
      *     the desired order.
      */
-    public KeyValueSystemFieldsRecordReader(
+    public KeyValueSpecialFieldsRecordReader(
             RecordReader<KeyValue> wrapped,
-            List<SystemFieldExtractor> systemFieldExtractors,
+            List<SpecialFieldExtractor> specialFieldExtractors,
             @Nullable int[] projection) {
         this.wrapped = wrapped;
-        this.systemFieldExtractors = systemFieldExtractors;
+        this.specialFieldExtractors = specialFieldExtractors;
         this.projection = projection;
     }
 
@@ -83,7 +83,7 @@ public class KeyValueSystemFieldsRecordReader implements RecordReader<InternalRo
         if (batch == null) {
             return null;
         }
-        return new SystemFieldsRecordIterator(batch);
+        return new SpecialFieldsRecordIterator(batch);
     }
 
     @Override
@@ -91,19 +91,14 @@ public class KeyValueSystemFieldsRecordReader implements RecordReader<InternalRo
         wrapped.close();
     }
 
-    private class SystemFieldsRecordIterator implements RecordIterator<InternalRow> {
+    private class SpecialFieldsRecordIterator implements RecordIterator<InternalRow> {
 
         private final RecordIterator<KeyValue> kvIterator;
-        private final JoinedRow joinedRow;
-        private final GenericRow systemFieldsRow;
-        @Nullable private final ProjectedRow projectedRow;
+        private final GenericRow specialFieldsRow;
 
-        private SystemFieldsRecordIterator(RecordIterator<KeyValue> kvIterator) {
+        private SpecialFieldsRecordIterator(RecordIterator<KeyValue> kvIterator) {
             this.kvIterator = kvIterator;
-            this.joinedRow = new JoinedRow();
-            this.systemFieldsRow = new GenericRow(systemFieldExtractors.size());
-            // If projection is provided, use ProjectedRow to reorder fields
-            this.projectedRow = projection != null ? ProjectedRow.from(projection) : null;
+            this.specialFieldsRow = new GenericRow(specialFieldExtractors.size());
         }
 
         @Nullable
@@ -116,20 +111,22 @@ public class KeyValueSystemFieldsRecordReader implements RecordReader<InternalRo
 
             InternalRow value = kv.value();
 
-            // Extract system fields into the reusable row
-            for (int i = 0; i < systemFieldExtractors.size(); i++) {
-                SystemFieldExtractor extractor = systemFieldExtractors.get(i);
-                Object systemValue = extractor.extract(kv);
-                systemFieldsRow.setField(i, systemValue);
+            // Extract special fields into the reusable row
+            for (int i = 0; i < specialFieldExtractors.size(); i++) {
+                SpecialFieldExtractor extractor = specialFieldExtractors.get(i);
+                Object specialValue = extractor.extract(kv);
+                specialFieldsRow.setField(i, specialValue);
             }
 
-            // Join system fields first, then physical fields
-            // Natural order: [system fields...] + [physical fields...]
-            joinedRow.replace(systemFieldsRow, value);
+            // Join special fields first, then physical fields
+            // Natural order: [special fields...] + [physical fields...]
+            JoinedRow joinedRow = new JoinedRow();
+            joinedRow.replace(specialFieldsRow, value);
             joinedRow.setRowKind(kv.valueKind());
 
             // If projection is provided, reorder to match requested order
-            if (projectedRow != null) {
+            if (projection != null) {
+                ProjectedRow projectedRow = ProjectedRow.from(projection);
                 return projectedRow.replaceRow(joinedRow);
             }
             return joinedRow;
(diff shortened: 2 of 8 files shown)


paimon-core/src/main/java/org/apache/paimon/table/source/KeyValueSystemFieldsRecordReader.java
```diff
@@ -0,0 +1,254 @@
...
+            // Extract system fields into the reusable row
+            for (int i = 0; i < systemFieldExtractors.size(); i++) {
+                SystemFieldExtractor extractor = systemFieldExtractors.get(i);
+                Object systemValue = extractor.extract(kv);
+                systemFieldsRow.setField(i, systemValue);
+            }
+
+            // Join system fields first, then physical fields
+            // Natural order: [system fields...] + [physical fields...]
+            joinedRow.replace(systemFieldsRow, value);
```
yunfengzhou-hub: I'm a little concerned about the correctness risk due to object reuse. We can see that in this project, JoinedRow#replace is mainly used in two ways

* `new JoinedRow().replace`: no reuse for this class.
* `joinedRow.replace` soon followed by a serializer or writer, which means there is no more modification to this  row after reused.

But here, the joined row will be output by the source to downstream operators, whose operations might modify the row in place and have `KeyValueSystemFieldsRecordReader` and that operator affects each other.

You can refer to online blogs and stackoverflow questions about the correctness risk of Flink object reuse mechanism, and check whether it is acceptable to enable object reuse here.

paimon-core/src/main/java/org/apache/paimon/table/source/KeyValueSystemFieldsRecordReader.java
```diff
@@ -0,0 +1,254 @@
...
+     * @return a reader producing InternalRow with system fields, or a simple unwrapped reader if no
+     *     system fields
+     */
+    public static RecordReader<InternalRow> wrap(
+            RecordReader<KeyValue> reader,
+            List<SystemFieldExtractor> systemFieldExtractors,
+            @Nullable int[] projection) {
+        if (systemFieldExtractors.isEmpty()) {
+            // No system fields, use the default unwrap logic
+            return KeyValueTableRead.unwrap(reader);
```
yunfengzhou-hub: It might be better to merge this method into `KeyValueTableRead#unwrap`. As a decorator/wrapper to this method, extending the original method might be more beneficial to extensibility.

paimon-core/src/main/java/org/apache/paimon/table/source/KeyValueSystemFieldsRecordReader.java
```diff
@@ -0,0 +1,254 @@
...
+    /**
+     * Creates a KeyValueSystemFieldsRecordReader.
+     *
+     * @param wrapped the underlying KeyValue reader
+     * @param systemFieldExtractors extractors for system fields, in order
+     * @param projection optional projection to reorder fields. If null, fields are in natural order
+     *     [system fields... + physical fields...]. If provided, projects from the natural order to
+     *     the desired order.
+     */
+    public KeyValueSystemFieldsRecordReader(
```
yunfengzhou-hub: better make it private?

paimon-core/src/main/java/org/apache/paimon/table/source/KeyValueSystemFieldsRecordReader.java
```diff
@@ -0,0 +1,254 @@
...
+        Object extract(KeyValue kv);
+
+        // ========== Built-in Extractors ==========
+
+        /**
+         * Extractor for {@code _SEQUENCE_NUMBER} system field.
+         *
+         * <p>Extracts the sequence number from KeyValue metadata.
+         */
+        SystemFieldExtractor SEQUENCE_NUMBER = kv -> kv.sequenceNumber();
```
yunfengzhou-hub:
```suggestion
        SystemFieldExtractor SEQUENCE_NUMBER = KeyValue::sequenceNumber;
```

paimon-core/src/main/java/org/apache/paimon/table/source/KeyValueSystemFieldsRecordReader.java
```diff
@@ -0,0 +1,254 @@
...
+        SystemFieldExtractor ROW_KIND = kv -> BinaryString.fromString(kv.valueKind().shortString());
+
+        /**
+         * Extractor for {@code _LEVEL} system field (LSM tree level).
+         *
+         * <p>Note: Currently not extractable from KeyValue at read time. This is a placeholder for
+         * future implementation where level information would need to be tracked through the read
+         * path.
+         */
+        SystemFieldExtractor LEVEL = kv -> null; // TODO: Level information needs to be propagated
```
yunfengzhou-hub: If this system field cannot be extracted for now, I would prefer to direct throw exceptions in such cases, instead of giving users a false assumption that it could work, but only gets null values.

paimon-core/src/main/java/org/apache/paimon/table/source/KeyValueSystemFieldsRecordReader.java
```diff
@@ -0,0 +1,254 @@
...
+
+        /**
+         * Extractor for {@code _ROW_ID} system field.
+         *
+         * <p>Note: ROW_ID is typically handled by DataFileRecordReader for InternalRow-based
+         * readers. This extractor is provided for completeness but may not be used in KeyValue
+         * scenarios.
+         */
+        SystemFieldExtractor ROW_ID =
+                kv -> null; // ROW_ID is computed from file metadata, not available in KeyValue
```
yunfengzhou-hub: Same as above

paimon-core/src/main/java/org/apache/paimon/operation/MergeFileSplitRead.java
```diff
@@ -408,4 +428,68 @@ public UserDefinedSeqComparator createUdsComparator() {
...
+    private List<KeyValueSystemFieldsRecordReader.SystemFieldExtractor>
+            collectSystemFieldExtractors(RowType readType) {
+        if (readType == null) {
+            return Collections.emptyList();
+        }
+
+        List<KeyValueSystemFieldsRecordReader.SystemFieldExtractor> extractors = new ArrayList<>();
+        for (String fieldName : readType.getFieldNames()) {
+            KeyValueSystemFieldsRecordReader.SystemFieldExtractor extractor =
+                    KeyValueSystemFieldsRecordReader.getExtractor(fieldName);
```
yunfengzhou-hub: Also check data type according to SpecialFields.

paimon-core/src/main/java/org/apache/paimon/operation/MergeFileSplitRead.java
```diff
@@ -408,4 +428,68 @@ public UserDefinedSeqComparator createUdsComparator() {
...
+
+    /**
+     * Collects system field extractors for the requested read type.
+     *
+     * @param readType the requested read type (may contain system fields)
+     * @return list of extractors for system fields present in readType
+     */
+    private List<KeyValueSystemFieldsRecordReader.SystemFieldExtractor>
+            collectSystemFieldExtractors(RowType readType) {
+        if (readType == null) {
```
yunfengzhou-hub, line 441: It might be impossible for this condition to be met, otherwise other codes should have already thrown NPEs.

paimon-core/src/main/java/org/apache/paimon/operation/MergeFileSplitRead.java
```diff
@@ -97,6 +99,11 @@ public class MergeFileSplitRead implements SplitRead<KeyValue> {
     @Nullable private int[][] outerProjection;
     @Nullable private VariantAccessInfo[] variantAccess;
 
+    private List<KeyValueSystemFieldsRecordReader.SystemFieldExtractor> systemFieldExtractors =
+            Collections.emptyList();
+
+    @Nullable private int[] projection = null;
```
yunfengzhou-hub, line 105: There are now at least three variables associated with "projection" in this class now
* pushdownProjection
* outerProjection
* projection

It might be better to merge or clarify between these projections, so as to improve readability.

paimon-core/src/main/java/org/apache/paimon/table/source/splitread/IncrementalChangelogReadProvider.java
```diff
@@ -75,7 +74,8 @@ private SplitRead<InternalRow> create(Supplier<MergeFileSplitRead> supplier) {
                                                     dataSplit.dataFiles(),
                                                     dataSplit.deletionFiles().orElse(null),
                                                     false));
-                    return unwrap(reader);
+                    return KeyValueSystemFieldsRecordReader.wrap(
+                            reader, read.getSystemFieldExtractors(), read.getProjection());
```
yunfengzhou-hub, line 78: LookupCompactDiffRead still uses unwrap. Does it also need to change to this method?

paimon-core/src/main/java/org/apache/paimon/table/source/splitread/IncrementalDiffSplitRead.java
```diff
@@ -76,6 +80,17 @@ public SplitRead<InternalRow> withIOManager(@Nullable IOManager ioManager) {
...
     public SplitRead<InternalRow> withReadType(RowType readType) {
         this.readType = readType;
+
+        List<DataField> fieldsWithAllSchema = new ArrayList<>();
+        for (DataField field : readType.getFields()) {
+            if (SpecialFields.isSystemField(field.name())) {
+                fieldsWithAllSchema.add(field);
+            }
+        }
+        fieldsWithAllSchema.addAll(mergeRead.tableSchema().fields());
```
yunfengzhou-hub: The name `fieldsWithReorderedSchema` might need to be improved for better readability.

paimon-core/src/main/java/org/apache/paimon/table/system/AuditLogTable.java
```diff
@@ -142,6 +144,9 @@ public String name() {
     public RowType rowType() {
         List<DataField> fields = new ArrayList<>();
         fields.add(SpecialFields.ROW_KIND);
+        fields.add(
+                SpecialFields.SEQUENCE_NUMBER.newType(
+                        SpecialFields.SEQUENCE_NUMBER.type().nullable()));
```
yunfengzhou-hub, line 149: Do you think it would be better if we only add sequence number to primary-key tables, not append-only tables, and make this field notnull()?

paimon-core/src/main/java/org/apache/paimon/table/system/AuditLogTable.java
```diff
@@ -722,19 +699,14 @@ public void setRowKind(RowKind kind) {
 
         @Override
         public boolean isNullAt(int pos) {
-            if (indexMapping[pos] < 0) {
-                // row kind is always not null
-                return false;
-            }
-            return super.isNullAt(pos);
+            return pos == rowkindPos ? false : super.isNullAt(pos);
```
yunfengzhou-hub, line 702: Better keep the previous code structure and comment.

yunfengzhou-hub, line 702: Only rowkindPos? not sequenceNumberPos?

paimon-core/src/main/java/org/apache/paimon/table/system/AuditLogTable.java
```diff
@@ -696,17 +666,24 @@ public RecordReader<InternalRow> createReader(Split split) throws IOException {
             return dataRead.createReader(split).transform(this::convertRow);
         }
 
-        private InternalRow convertRow(InternalRow data) {
-            return new AuditLogRow(readProjection, data);
+        private InternalRow convertRow(InternalRow row) {
+            row.setRowKind(org.apache.paimon.types.RowKind.INSERT);
+            // Fallback for append-only tables: rowkind field is null, wrap to return "+I"
```
yunfengzhou-hub, line 671: `row.setRowKind(org.apache.paimon.types.RowKind.INSERT)` is placed above the comment and if condition, which seems that it will always return +I, regardless of whether it is append-only table.

paimon-core/src/main/java/org/apache/paimon/table/system/AuditLogTable.java
```diff
@@ -722,19 +699,14 @@ public void setRowKind(RowKind kind) {
...
         }
 
         @Override
         public BinaryString getString(int pos) {
-            if (indexMapping[pos] < 0) {
-                return BinaryString.fromString(row.getRowKind().shortString());
-            }
-            return super.getString(pos);
+            return pos == rowkindPos
+                    ? BinaryString.fromString(getRowKind().shortString())
```
yunfengzhou-hub, line 708: If rowkind can be covered by SystemFieldExtractor now, do we still need this?

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/scan/BaseScan.scala
```diff
@@ -87,7 +88,8 @@ trait BaseScan extends Scan with SupportsReportStatistics with Logging {
     }
 
     if (ROW_TRACKING_META_COLUMNS.contains(fieldName)) {
-      if (!coreOptions.rowTrackingEnabled()) {
+      val isSystemTable = table.isInstanceOf[AuditLogTable] || table.isInstanceOf[BinlogTable]
```
yunfengzhou-hub: System tables are more than audit log table and binlog table.

paimon-spark/paimon-spark-ut/src/test/scala/org/apache/paimon/spark/sql/TableValuedFunctionsTest.scala
```diff
@@ -289,7 +289,7 @@ class TableValuedFunctionsTest extends PaimonHiveTestBase {
       // after files:  f(-D 999),   f(+I 10001,10002)
       checkAnswer(
         sql("SELECT * FROM paimon_incremental_query('`t$audit_log`', 'tag1', 'tag2') ORDER BY id"),
-        Seq(Row("-D", 999)))
+        Seq(Row("-D", 100002L, 999)))
```
yunfengzhou-hub, line 292: 100002L is the value of `id` column, rather than sequence number. Not sure whether it is correct here.

paimon-core/src/test/java/org/apache/paimon/table/system/AuditLogTableTest.java
```diff
@@ -48,11 +56,176 @@
...
+        write(table, GenericRow.ofKind(RowKind.INSERT, 1, 2, 5)); // Will be updated
+        write(table, GenericRow.ofKind(RowKind.UPDATE_BEFORE, 1, 2, 5));
+        write(table, GenericRow.ofKind(RowKind.UPDATE_AFTER, 1, 4, 6));
+        write(table, GenericRow.ofKind(RowKind.INSERT, 2, 3, 1)); // Remains as insert
+
+        return table;
+    }
+
+    /** Creates a table with changelog producer enabled. */
+    private FileStoreTable createChangelogTable(String tableName) throws Exception {
```
yunfengzhou-hub, line 228: This method is only used in one place. So compared with introducing this method, it might be better to preserve the original code structure and commit history.

paimon-core/src/main/java/org/apache/paimon/table/source/KeyValueSystemFieldsRecordReader.java
```diff
@@ -0,0 +1,254 @@
...
+
+import javax.annotation.Nullable;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * A decorator for {@link RecordReader} that injects system fields into the output rows for
```
yunfengzhou-hub: A decorator pattern should have the same wrapped type and output type, but here the wrapped is `RecordReader<KeyValue>`, and the output is `RecordReader<InternalRow>`. It might be more proper to call this class a wrapper rather than a decorator.

paimon-core/src/main/java/org/apache/paimon/table/source/KeyValueSpecialFieldsRecordReader.java
```diff
@@ -0,0 +1,254 @@
...
+
+    /**
+     * Gets an extractor by field name.
+     *
+     * @param fieldName the system field name
+     * @return the extractor, or null if not a registered system field
+     */
+    @Nullable
+    public static SystemFieldExtractor getExtractor(String fieldName) {
+        return EXTRACTOR_REGISTRY.get(fieldName);
```
yunfengzhou-hub, line 238: This method only checks field names. It might be better to also check field data types.

paimon-core/src/main/java/org/apache/paimon/table/source/KeyValueSystemFieldsRecordReader.java
```diff
@@ -0,0 +1,254 @@
...
+ * differently in {@link org.apache.paimon.io.DataFileRecordReader} using file metadata.
+ *
+ * <p><b>Field Ordering:</b> The output schema supports arbitrary field ordering. Internally, fields
+ * are assembled as [system fields... + physical fields...], then reordered using {@link
+ * ProjectedRow} to match the requested field order.
+ *
+ * <p><b>Performance:</b> Implementation uses {@link JoinedRow} for zero-copy concatenation of
+ * system fields and physical fields, then {@link ProjectedRow} for zero-copy field reordering.
+ */
+public class KeyValueSystemFieldsRecordReader implements RecordReader<InternalRow> {
```
yunfengzhou-hub: According to the existing `SpecialFields` class, Might be better to name them as `KeyValueSpecialFieldsRecordReader` and `SpecialFieldExtractor`.

yunfengzhou-hub (commented):
Thanks for the PR! Left some comments as below.

paimon-core/src/main/java/org/apache/paimon/table/source/KeyValueSystemFieldsRecordReader.java
wzhero1: Great catch! You're right - downstream operator behavior is unpredictable and could mutate the row in-place. I've changed it to `new JoinedRow()`. The minor overhead of object allocation is acceptable to ensure correctness and avoid potential interference between `KeyValueSystemFieldsRecordReader` and downstream operators.

paimon-core/src/main/java/org/apache/paimon/table/source/KeyValueSystemFieldsRecordReader.java
wzhero1: Done. I've moved the logic into `KeyValueTableRead.java`. The method structure is now:
* `unwrap(RecordReader<KeyValue> reader, List<SpecialFieldExtractor> specialFieldExtractors, @Nullable int[] projection)` as public method for extensibility
* `unwrap(RecordReader<KeyValue> reader)` as private method for internal use

paimon-core/src/main/java/org/apache/paimon/table/source/KeyValueSystemFieldsRecordReader.java
wzhero1: done

paimon-core/src/main/java/org/apache/paimon/table/source/KeyValueSystemFieldsRecordReader.java
wzhero1: now `throw new UnsupportedOperationException`

paimon-core/src/main/java/org/apache/paimon/table/source/KeyValueSystemFieldsRecordReader.java
wzhero1: now `throw new UnsupportedOperationException`

paimon-core/src/main/java/org/apache/paimon/table/source/splitread/IncrementalChangelogReadProvider.java
wzhero1: unwrap is also correct, but i changed to `KeyValueTableRead.unwrap(reader, mergeRead.getSystemFieldExtractors(), mergeRead.getProjection())`, so unwrap can be private in `KeyValueTableRead`

paimon-core/src/main/java/org/apache/paimon/table/source/splitread/IncrementalDiffSplitRead.java
wzhero1: changed to `schemaAlignedFields`

paimon-core/src/main/java/org/apache/paimon/table/system/AuditLogTable.java
wzhero1: I prefer keeping the schema consistent across table types. Having `SEQUENCE_NUMBER` only in primary-key tables but not in append-only tables increases user cognitive load when querying audit_log or binlog system tables. A consistent schema with null for unavailable values is semantically correct and easier to understand.

paimon-core/src/main/java/org/apache/paimon/table/system/AuditLogTable.java
wzhero1: For primary-key tables, both `rowKind` and `sequenceNumber` are already extracted in `KeyValueTableRead`. This logic here is specifically for append-only tables, where both fields would be null by default. The special handling for rowKindPos is needed because for `AuditLogTable`, append-only records should explicitly show rowKind as `INSERT` rather than null.

paimon-core/src/main/java/org/apache/paimon/table/system/AuditLogTable.java
wzhero1: There are two distinct concepts here:

1. **Row's metadata** `rowKind` (`row.setRowKind(...)`) - This is always `INSERT` for both pk and ap tables when reading from `AuditLogTable`。
2. The `rowKind` column field (the regular column in schema) - For pk tables, this is already populated by `KeyValueTableRead`. For ap tables, this column needs explicit filling since the underlying data doesn't carry this information.
The if condition below handles case2 specifically for ap tables.

paimon-core/src/main/java/org/apache/paimon/table/system/AuditLogTable.java
wzhero1: For ap table, we need to get the `rowKind` column field (the regular column in schema) from `getRowKind()`

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/scan/BaseScan.scala
wzhero1: changed to isAuditOrBinlogTable

paimon-spark/paimon-spark-ut/src/test/scala/org/apache/paimon/spark/sql/TableValuedFunctionsTest.scala
wzhero1: 999 is the value of `id`, 100002L is the sequence number

paimon-core/src/main/java/org/apache/paimon/table/source/KeyValueSystemFieldsRecordReader.java
wzhero1: changed to wrapped

paimon-core/src/main/java/org/apache/paimon/table/source/KeyValueSpecialFieldsRecordReader.java
wzhero1: For special fields (like _ROWKIND, _SEQUENCE_NUMBER), field name matching is sufficient.

paimon-core/src/main/java/org/apache/paimon/table/source/KeyValueSystemFieldsRecordReader.java
wzhero1: good suggestion. i have changed all terms: `system -> special`

paimon-core/src/test/java/org/apache/paimon/table/system/AuditLogTableTest.java
wzhero1: While it's currently used in one place, I prefer keeping this helper method for single responsibility and better extensibility. It encapsulates the changelog table creation logic cleanly, making future test additions easier without duplicating configuration code.

paimon-core/src/main/java/org/apache/paimon/operation/MergeFileSplitRead.java
wzhero1: Agreed it's unlikely to be `null` in practice, but this check is intentional for method self-containment. It ensures the method handles all input states gracefully without relying on caller guarantees - a minor redundancy for defensive completeness.

PR #6863, opened 2025-12-22
[python] Support read blob row by offsets in with_shard feature

discivigour:
### Purpose

When the current code reads a blob file for 'with shard' feature, it reads the entire blob file and only selects the required data. Since sometimes blob files are very large, the read amplification phenomenon is quite serious. This pr has implemented reading only the required blob records based on the offsets of the blob file.


JingsongLi (approved):
+1

Issue #6834, opened 2025-12-18
[Feature] Introduce BTree global index


steFaiz:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Motivation

At present, for scalar indexes we mainly support the Bitmap index. Many thanks to @leaves12138 for the implementation, which has established a solid framework for scalar indexing. However, a conventional Bitmap index has significant limitations and does not work well for scenarios with high data cardinality, such as int, double, and string types.

A global B-Tree index can be built on any comparable data type, providing efficient point lookups and range queries. In addition, it is more amenable to distributed parallel processing during updates and reads. Therefore, this issue proposes implementing a distributed global B-Tree index.

### Solution

The basic implementation will be built on the SST FileFormat introduced in https://github.com/apache/paimon/issues/6734, providing point lookup and range query capabilities to cover most common SQL filter predicates.
## index construction
Index construction can be efficiently implemented via range shuffle in Flink or Spark: different writer tasks are responsible for writing index files for their assigned key ranges, and a commit task then performs a unified commit of all generated files. As below:

<img width="1498" height="640" alt="Image" src="https://github.com/user-attachments/assets/5fac0cca-6836-4636-a814-72c4061d6830" />

## index query
The metadata of B-Tree index files will record information such as the file’s min key, max key, and whether it contains nulls (hasNulls). During index planning, we can use this metadata to prune candidate files, then query the remaining files in parallel and merge the results.

We will bring more details in future PRs.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #6869, opened 2025-12-23
[core] support btree global index in paimon-common

steFaiz:
### Purpose

This PR is a part of https://github.com/apache/paimon/issues/6834, aiming to provide capability in paimon-common module.
## File Format
The introduced BTree Index File is also base on SST File Format, with slight difference with LookupStore File, as illustrated below:
image

The file is composed by:
1. A normal sst file
2. A null bitmap
3. A Footer containing all handles and magic number
## File Entry
The key-value pair of internal SST File is as below:
image
* key: the serialized key bytes
* value: a list of row id, which is encoded as a varLenInt and following varLenLongs.

### Tests
Please see:
* org.apache.paimon.globalindex.btree.BTreeFileMetaSelectorTest
* org.apache.paimon.globalindex.btree.BTreeGlobalIndexerTest

### API and Format
None

### Documentation
Will be added in the future.


steFaiz:
Test failures are irrelevant to this PR. Will try to reopen after all reviews addressed.

JingsongLi:
@steFaiz Can you open actions in your repo too? This allows for double verification.

JingsongLi:
I think we can just introduce another format.
Lookup Sst file:
```
 *     +-----------------------------------+------+
 *     |             Footer                |      |
 *     +-----------------------------------+      |
 *     |           Index Block             |      +--> Loaded on open
 *     +-----------------------------------+      |
 *     |        Bloom Filter Block         |   +-->  Loaded on first Lookup, memory can be expired
 *     +-----------------------------------+------+
 *     |            Data Block             |      |
 *     +-----------------------------------+      |
 *     |              ......               |      +--> Loaded on requested
 *     +-----------------------------------+      |
 *     |            Data Block             |      |
 *     +-----------------------------------+------+
 ```
BTree Index File:
```
 *     +-----------------------------------+------+
 *     |             Footer                |      |
 *     +-----------------------------------+      |
 *     |           Index Block             |      +--> Loaded on open
 *     +-----------------------------------+      |
 *     |        Bloom Filter Block       |      +--> Loaded on first Lookup, memory can be expired
 *     +-----------------------------------+      |
 *     |         Null bitmap Block    |      +--> Loaded on requested
 *     +-----------------------------------+------+
 *     |            Data Block             |      |
 *     +-----------------------------------+      |
 *     |              ......               |      +--> Loaded on requested
 *     +-----------------------------------+      |
 *     |            Data Block             |      |
 *     +-----------------------------------+------+
 ```

steFaiz:
> I think we can just introduce another format. Lookup Sst file:

Thanks for your inspiration! I will try to refactor the code!

JingsongLi (approved):
Looks good to me!

PR #6875, opened 2025-12-24
[flink] Enable data-evolution compaction on flink

leaves12138:
### Purpose

Enable data-evolution compaction on flink!

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkCatalog.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkCatalog.java
@@ -26,7 +26,6 @@ import org.apache.paimon.catalog.CatalogUtils;
 import org.apache.paimon.catalog.Database;
 import org.apache.paimon.catalog.Identifier;
 import org.apache.paimon.catalog.PropertyChange;
-import org.apache.paimon.flink.function.BuiltInFunctions;
 import org.apache.paimon.flink.procedure.ProcedureUtil;
 import org.apache.paimon.flink.utils.FlinkCatalogPropertiesUtil;
 import org.apache.paimon.flink.utils.FlinkDescriptorProperties;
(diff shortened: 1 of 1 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/function/BuiltInFunctions.java
```diff
@@ -0,0 +1,34 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.paimon.flink.function;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/** Paimon flink built in functions. */
+public class BuiltInFunctions {
```
JingsongLi: Why adding this?

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/action/CompactActionITCase.java
```diff
@@ -825,6 +840,109 @@ public void testStreamingCompactWithEmptyOverwriteUpgrade() throws Exception {
                 60_000);
     }
 
+    public void testDataEvolutionTableCompact() throws Exception {
```
JingsongLi, line 844: No `@Test`?

leaves12138: Thanks, fixed

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/function/BuiltInFunctions.java
leaves12138: Because I think it is a good idea to provide some functions in paimon catalog we need. Do you think so?

leaves12138: For example, path_to_descriptor, users don't have to create function themselves. Just call path_to_descriptor('oss://xx-bucket/xx-path/xx.png'), it could convert the path to blob descriptor

leaves12138: Removed, I will pull request this in another pr

JingsongLi:
+1

PR #6879, opened 2025-12-24
[format] Introduce text format table

Zouxxyy:
### Purpose

```sql
CREATE TABLE t (v STRING) USING text;
INSERT INTO t VALUES ('aaaa');
```

Some tips (follow spark):

- Null values will be written as empty strings and read back as empty strings as well.

```sql
INSERT INTO t VALUES (null);

+----+
|   v|
+----+
|    |
+----+
```

- Values containing the line delimiter will not throw an exception; instead, they will be read as multiple additional lines.

```sql
INSERT INTO t VALUES ('aa'), ('bb?cc');

+---+
|  v|
+---+
| aa|
| bb|
| cc|
+---+
```

### DIFF
--- a/paimon-spark/paimon-spark-ut/src/test/scala/org/apache/paimon/spark/sql/FormatTableTestBase.scala
+++ b/paimon-spark/paimon-spark-ut/src/test/scala/org/apache/paimon/spark/sql/FormatTableTestBase.scala
@@ -259,4 +259,24 @@ abstract class FormatTableTestBase extends PaimonHiveTestBase {
       }
     }
   }
+
+  test("Paimon format table: write text format with value with line-delimiter") {
+    withTable("t") {
+      sql(s"""
+             |CREATE TABLE t (v STRING) USING text
+             |TBLPROPERTIES ('text.line-delimiter'='?')
+             |""".stripMargin)
+      sql("INSERT INTO t VALUES ('aa'), ('bb?cc')")
+
+      for (impl <- Seq("paimon", "engine")) {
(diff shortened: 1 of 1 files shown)


paimon-format/src/main/java/org/apache/paimon/format/text/TextFormatWriter.java
```diff
@@ -0,0 +1,49 @@
...
+            TextOptions textOptions,
+            String compression)
+            throws IOException {
+        super(outputStream, rowType, compression);
+        this.textOptions = textOptions;
+    }
+
+    @Override
+    public void addElement(InternalRow element) throws IOException {
+        if (!element.isNullAt(0)) {
```
Zouxxyy, line 44: Here is follow spark

paimon-spark/paimon-spark-ut/src/test/scala/org/apache/paimon/spark/sql/FormatTableTestBase.scala
```diff
@@ -238,4 +238,25 @@ abstract class FormatTableTestBase extends PaimonHiveTestBase {
...
     }
   }
+
+  test("Paimon format table: text format") {
+    withTable("t") {
+      sql(s"""
+             |CREATE TABLE t (v STRING) USING text
+             |TBLPROPERTIES ('text.line-delimiter'='?')
+             |""".stripMargin)
+      sql("INSERT INTO t VALUES ('aaaa'), ('bbbb'), ('cccc'), (null), ('dddd')")
```
jerry-024, line 248: add value which contain the  `text.line-delimiter`

PR #6882, opened 2025-12-24
[rest] support row filter & column masking in rest catalog

plusplusjiajia:


JingsongLi:
We can split multiple PR parts to advance the progress.

plusplusjiajia:
> We can split multiple PR parts to advance the progress.

Got it! I can split it into multiple PRs.

PR #6884, opened 2025-12-24
[core] Support enable and disable tag time expiration

yuzelin:
### Purpose

Paimon support to disable expiration of snapshot (snapshot.expire.limit = 0), partition (partition.expiration-time = null) and consumer (consumer.expiration-time = null), but the tag expiration are enabled by default. This PR introduce `tag.time-expire-enabled` option for this purpose.


JingsongLi:
Do we need to separate tag auto creation and expire?

JingsongLi:
+1

PR #6896, opened 2025-12-25
[core] Prevent tag deletion when referenced by branches

### LINKED ISSUE #6272: [Bug] Branch Base Snapshot Gets Expired After Deleting Referenced Tag
### Search before asking

### Paimon version

master

### Compute Engine

all

## Summary
When a branch is created based on a tag, deleting that tag can cause the branch's base snapshot to be expired during snapshot expiration operations, making the branch unqueryable.

Currently, Paimon has a protection mechanism that prevents snapshots referenced by tags from being expired (implemented in `ExpireSnapshotsImpl.java`). However, there's no corresponding protection for branches that reference tags.

The issue occurs in the following scenario:
1. Create a tag pointing to a specific snapshot
2. Create a branch based on this tag
3. Delete the tag
4. Run snapshot expiration
5. The base snapshot of the branch gets expired because it's no longer protected by the tag
6. The branch becomes unqueryable

base snapshot not be deleted

### Solution 1: Tag Deletion Protection (Preferred)

Add a protection mechanism when deleting tags to prevent deletion if any branches reference the tag.

**Implementation approach**:
- Modify `TagManager.deleteTag()` method
- Before deleting a tag, check if any branches reference this tag
- If branches exist, either:
  - Reject the deletion with an error message
  - Require a force flag to proceed with deletion

**Advantages**:
- Prevents the problem at the source
- Clear error message to users
- Maintains data integrity

### Solution 2: Branch-aware Snapshot Expiration

Modify the snapshot expiration logic to consider branches when determining which snapshots to protect.

**Implementation approach**:
- Modify `ExpireSnapshotsImpl.expireUntil()` method
- Add branch manager to collect all branch base snapshots
- Include branch-referenced snapshots in the skipping set

**Advantages**:
- More comprehensive protection
- Handles edge cases where branches might reference snapshots directly

userzhy:
### Purpose

Linked issue: close #6272

When a branch is created from a tag, the tag's snapshot information is copied to the branch directory. However, if the tag is deleted, the snapshot expiration process may clean up data files that are still needed by the branch, making the branch unqueryable.

This change implements Tag Deletion Protection: before deleting a tag, we check if any branches were created from that tag. If so, the deletion is blocked with an error message listing the referencing branches.

Changes:
- Added `branchesCreatedFromTag(String tagName)` method to `BranchManager` interface with default empty implementation for backward compatibility
- Implemented `branchesCreatedFromTag()` in `FileSystemBranchManager` to check each branch's tag directory for the specified tag
- Added validation in `AbstractFileStoreTable.deleteTag()` to prevent deletion of tags that are still referenced by branches

### Tests

- Added `testDeleteTagReferencedByBranch()` in `SimpleTableTestBase` to verify:
  - Deleting a tag referenced by a branch throws `IllegalStateException`
  - The error message includes the referencing branch names
  - After deleting all referencing branches, the tag can be deleted successfully

### API and Format

- Added new method `branchesCreatedFromTag(String tagName)` to `BranchManager` interface with default implementation returning empty list, so existing implementations remain compatible
- No storage format changes

### Documentation

This is a bug fix, no new feature documentation needed.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/utils/BranchManager.java
+++ b/paimon-core/src/main/java/org/apache/paimon/utils/BranchManager.java
@@ -42,6 +42,16 @@ public interface BranchManager {
 
     List<String> branches();
 
+    /**
+     * Get all branches that were created based on the given tag.
+     *
+     * @param tagName the name of the tag to check
+     * @return list of branch names that reference the given tag
+     */
+    default List<String> branchesCreatedFromTag(String tagName) {
+        return java.util.Collections.emptyList();
+    }
+
     default boolean branchExists(String branchName) {
         return branches().contains(branchName);
     }
--- a/paimon-core/src/main/java/org/apache/paimon/table/AbstractFileStoreTable.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/AbstractFileStoreTable.java
@@ -656,6 +656,14 @@ abstract class AbstractFileStoreTable implements FileStoreTable {
 
     @Override
     public void deleteTag(String tagName) {
+        List<String> referencingBranches = branchManager().branchesCreatedFromTag(tagName);
+        if (!referencingBranches.isEmpty()) {
+            throw new IllegalStateException(
+                    String.format(
+                            "Cannot delete tag '%s' because it is still referenced by branches: %s. "
+                                    + "Please delete these branches first.",
+                            tagName, referencingBranches));
+        }
         tagManager()
                 .deleteTag(
                         tagName,
--- a/paimon-core/src/main/java/org/apache/paimon/utils/FileSystemBranchManager.java
+++ b/paimon-core/src/main/java/org/apache/paimon/utils/FileSystemBranchManager.java
@@ -215,6 +215,18 @@ public class FileSystemBranchManager implements BranchManager {
         }
     }
 
+    @Override
+    public List<String> branchesCreatedFromTag(String tagName) {
+        List<String> result = new java.util.ArrayList<>();
+        for (String branchName : branches()) {
+            TagManager branchTagManager = tagManager.copyWithBranch(branchName);
+            if (branchTagManager.tagExists(tagName)) {
+                result.add(branchName);
+            }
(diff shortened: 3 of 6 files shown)


JingsongLi (commented):
Good safety mechanism. Preventing tag deletion when branches reference it avoids orphaned data issues.

**Review:**

1. **Implementation approach**: Checking each branch's tag directory via `branchesCreatedFromTag()` in `FileSystemBranchManager` is straightforward. For a small number of branches this is fine, but:
   - How many branches can a table typically have? If it's >100, iterating all branches on every tag delete could be slow.
   - Consider caching or maintaining a reverse index (tag → branches) if this becomes a bottleneck.

2. **Default implementation**: The `BranchManager` interface has a default returning empty list — backward compatible. Good.

3. **Error handling**: `IllegalStateException` with referencing branch names listed is clear and actionable for users.

4. **Test coverage**: The test verifies:
   - Deletion blocked when branch references tag ✓
   - Deletion succeeds after branch is removed ✓
   
   Consider also testing: what if multiple branches reference the same tag?

5. **Race condition**: Between checking `branchesCreatedFromTag()` and actual deletion, a new branch could be created from the tag. This is unlikely in practice but worth noting in the javadoc.

LGTM. Clean implementation for the stated issue.

paimon-core/src/main/java/org/apache/paimon/utils/BranchManager.java
```diff
@@ -42,6 +42,16 @@ public interface BranchManager {
...
     List<String> branches();
 
+    /**
+     * Get all branches that were created based on the given tag.
+     *
+     * @param tagName the name of the tag to check
+     * @return list of branch names that reference the given tag
+     */
+    default List<String> branchesCreatedFromTag(String tagName) {
+        return java.util.Collections.emptyList();
```
JingsongLi, line 52: Please also fix this to `CatalogBranchManager`.

JingsongLi (changes requested):
-1 See comment.

Issue #6897, opened 2025-12-25
[Bug] Correct identifer information in catalog enviroment when using switch branch


Stefanietry:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Paimon version

paimon 1.1

### Compute Engine

Spark 3.2 

### Minimal reproduce step

1. create main table
2. create branch table named "fallback"
3. execute "switch branch to fallback" on the main table 
4. get the identifer by table catalog enviroment 

### What doesn't meet your expectations?

In the spark engine, executing `drop Partition` via `switch branch` deletes the corresponding partition on that branch.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #6898, opened 2025-12-25
[Core] Correct identifer information in catalog enviroment when using…

Stefanietry:
… switch branch
core

### Purpose

https://github.com/apache/paimon/issues/6897

### Tests

org.apache.paimon.table.PrimaryKeySimpleTableTest#testSwitchBranch


Aitozi (approved):
+1

Issue #6899, opened 2025-12-25
[Bug] Fix the exception when building in predicate if the value is empty


Stefanietry:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Paimon version

1.1

### Compute Engine

spark engine

### Minimal reproduce step

1. call  PredicateBuilder#in(int, java.util.List<java.lang.Object>) and pass an empty list to the second parameter

### What doesn't meet your expectations?

When executing join query by spark  and optimized by dynamic filter, the value on filter becomes empty,  ultimately leading to the exception 

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #6900, opened 2025-12-25
[core] fix in predict when value is empty

Stefanietry:
core

### Purpose

https://github.com/apache/paimon/issues/6899

### Tests

org.apache.paimon.predicate.PredicateBuilderTest#testIn


Aitozi (approved):
+1

PR #6905, opened 2025-12-26
[rest] Add fromSnapshot to rollback

JingsongLi:
### Purpose

Add `fromSnapshot` to ensure that there are no other new commits during the rollback period and to avoid unknown errors.


sundapeng:
+1. I understand this PR aims to address the issue of missing commits that were successfully written during rollback, similar to how `git push --force-with-lease` works. If the rollback fails, users would stop the writing jobs first.

PR #6912, opened 2025-12-28
[spark] fix Merge Into unstable tests

wayneli-vt:
### Purpose

This PR fixes a flaky unit test caused by Spark’s asynchronous `QueryExecutionListener` callbacks. The previous implementation could iterate over a mutable collection while the listener thread was still appending, leading to `ConcurrentModificationException`. The fix waits for the expected plan via `CountDownLatch` before asserting.


JingsongLi:
+1 Thanks @wayneli-vt

Issue #6847, opened 2025-12-20
[Feature] Optimize read table with limit.


wwj6591812:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Motivation

一、Problem Background
In our company's Flink Session-based OLAP environment, we observed that a very basic Flink SQL query against a Paimon table with a LIMIT clause took an unexpectedly long time—over 20 seconds—to complete.
（1）The query is as follows:
`select * from `paimon_na61`.`sample`.`s_holo_mainse_rank_xfc_all_features_swift_parsed` limit 10;`

（2）The table location is : 
http://bs.alibaba-inc.com/fsutil?path=dfs%3A//na61dfsalake1--cn-zhangjiakou/alake/omega_na61/sample.db/s_holo_mainse_rank_xfc_all_features_swift_parsed/&isdir=true

二、Root Cause Analysis
Through code review and by attaching to the live Java process with Arthas, we identified the following:
1、The primary bottleneck is the SnapshotReaderImpl#generateSplits method.
（1）We only need select 10 record, but the List<split> is too large.
<img width="1652" height="1580" alt="Image" src="https://github.com/user-attachments/assets/ecb0b910-2186-446e-9a05-51c68226f1bc" />

<img width="1698" height="288" alt="Image" src="https://github.com/user-attachments/assets/13c6f2bb-2406-4ffe-9a02-622d36231d06" />

We found the number of files observed in the manifests system table is also consistent with this large scale.
<img width="2214" height="420" alt="Image" src="https://github.com/user-attachments/assets/0aa864c3-27ef-499a-b52f-d558c06177f8" />

（2）The main time cost:

http://ha3.oss-cn-hangzhou-zmf.aliyuncs.com/052894/paimon_sample_table_slow.arthas.html

<img width="3004" height="1446" alt="Image" src="https://github.com/user-attachments/assets/6083f7e0-30d2-4108-80cf-7a13d36e98ae" />

2、Upon further investigation, we found that when executing a SELECT query, the current implementation of AbstractFileStoreScan#plan does not take the LIMIT clause from the SQL into account. This can cause the splits variable (List<DataSplit>) within SnapshotReaderImpl#read() to become excessively large. Consequently, when the table consists of a large number of files, SnapshotReaderImpl#generateSplits takes a very long time to execute.

三、Conclusion and Plan
To resolve this, we propose to push the LIMIT clause down into the AbstractFileStoreScan logic.
[...]

PR #6914, opened 2025-12-28
[Core] support limit pushdown with pk table

wwj6591812:
### Purpose

一、Background
Issue: https://github.com/apache/paimon/issues/6847
Append table limit pushdown : https://github.com/apache/paimon/pull/6848
This pr is to support limit pushdown with pk table.

二、Code Logic
We try best to filtering manifest entries before data reading. 
Limit pushdown is only enabled for DEDUPLICATE/FIRST_ROW merge engines without deletion vectors, as accurate counting requires no merge operations or deleted rows.
We groups files by (partition, bucket) pairs and processes buckets sequentially. For each bucket, the algorithm checks if safe pushdown is possible: files must have no overlapping (same LSM level, excluding level 0) and no delete rows. 
1、If safe, it accumulates row counts from file metadata until reaching the limit, then stops processing remaining buckets. 2、If unsafe (overlapping files or delete rows exist), all files in that bucket are included.  

Linked issue: close #xxx

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreScan.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreScan.java
@@ -46,6 +46,9 @@ import org.apache.paimon.utils.Pair;
 import org.apache.paimon.utils.Range;
 import org.apache.paimon.utils.SnapshotManager;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import javax.annotation.Nullable;
 
 import java.util.ArrayList;
@@ -67,6 +70,8 @@ import static org.apache.paimon.utils.ThreadPoolUtils.randomlyOnlyExecute;
 
 /** Default implementation of {@link FileStoreScan}. */
 public abstract class AbstractFileStoreScan implements FileStoreScan {
+
+    private static final Logger LOG = LoggerFactory.getLogger(AbstractFileStoreScan.class);
 
     private final ManifestsReader manifestsReader;
     private final SnapshotManager snapshotManager;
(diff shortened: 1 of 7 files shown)


paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreScan.java
```diff
@@ -277,15 +277,17 @@ public Plan plan() {
...
-        if (supportsLimitPushManifestEntries()) {
-            iterator = limitPushManifestEntries(iterator);
-        }
 
         List<ManifestEntry> files = ListUtils.toList(iterator);
         if (postFilterManifestEntriesEnabled()) {
             files = postFilterManifestEntries(files);
         }
 
+        if (supportsLimitPushManifestEntries()) {
```
JingsongLi: You need to do performance test again.

wwj6591812: You are right, I will do it.

docs/layouts/shortcodes/generated/core_configuration.html
```diff
@@ -1001,6 +1001,12 @@
             <td>String</td>
             <td>You can specify a pattern to get a timestamp from partitions. The formatter pattern is defined by 'partition.timestamp-formatter'.<ul><li>By default, read from the first field.</li><li>If the timestamp in the partition is a single field called 'dt', you can use '$dt'.</li><li>If it is spread across multiple fields for year, month, day, and hour, you can use '$year-$month-$day $hour:00:00'.</li><li>If the timestamp is in fields dt and hour, you can use '$dt $hour:00:00'.</li></ul></td>
         </tr>
+        <tr>
+            <td><h5>pk-table-limit-push-down.enabled</h5></td>
```
JingsongLi: Let's remove this config?

wwj6591812: done

JingsongLi (commented):
Can you merge `limitPushManifestEntries` into `postFilterManifestEntries`?

wwj6591812:
> Can you merge `limitPushManifestEntries` into `postFilterManifestEntries`?

done, thx.

please CC .  @JingsongLi

JingsongLi:
Hi @wwj6591812 
1. We can optimize the bucket path function (this is the performance bottleneck) and test its performance to see if the optimization effect can be achieved without pushing down the limit to scan.
2. If the performance improvement is not obvious, it is necessary to merge the two methods, postFilterManifestEntries and limitPushManifestEntries, and only keep one, postFilterManifestEntries.

PR #6917, opened 2025-12-28
 Support generating splits with finer granularity than file level

### LINKED ISSUE #5012: [Feature] Support generating splits with finer granularity than file level
### Search before asking

### Motivation

If we have only one parquet file (4G), but it has 10 rowgroups, if we can generate 10 splits. which will greatly improves concurrency.

### Solution

_No response_

### Anything else?

_No response_

Shekharrajak:
[core] Support generating splits with finer granularity than file level

### Purpose

Linked issue: close #5012

This PR implements support for generating splits at finer granularity than file level (e.g., row groups for Parquet, stripes for ORC) to significantly enhance concurrency when reading large files. This follows the proven pattern used by Spark and Flink, where files are split at natural boundaries (row groups/stripes) rather than file boundaries.

The implementation leverages existing `RawFile` infrastructure with `offset` and `length` fields, ensuring backward compatibility while enabling improved parallelism for large file reads.

### Tests

Unit tests and integration tests are needed for:
- `ParquetMetadataReader`: Verify row group boundary extraction
- `OrcMetadataReader`: Verify stripe boundary extraction
- `FineGrainedSplitGenerator`: Verify split generation logic
- `ParquetReaderFactory.createReader(offset, length)`: Verify range-based reading
- `OrcReaderFactory.createReader(offset, length)`: Verify range-based reading

### API and Format

**New Configuration Options:**
- `source.split.file-enabled`: Enable finer-grained file splitting (default: `false`)
- `source.split.file-threshold`: Minimum file size to consider splitting (default: `128MB`)
- `source.split.file-max-splits`: Maximum splits per file (default: `100`)

**New Interfaces/Classes:**
- `FormatMetadataReader`: Interface for reading format-specific metadata
- `FileSplitBoundary`: Represents split boundaries (offset, length, rowCount)
- `ParquetMetadataReader`: Extracts row group boundaries from Parquet files
- `OrcMetadataReader`: Extracts stripe boundaries from ORC files
- `FineGrainedSplitGenerator`: Decorator for `SplitGenerator` that enables fine-grained splitting

**Extended Interfaces:**
- `FormatReaderFactory.createReader(Context, offset, length)`: Now implemented for Parquet and ORC
- `DataSplit`: Added transient `fileSplitBoundaries` field (not serialized for backward compatibility)

**Storage Format:** No changes to storage format. This is a read-time optimization that doesn't affect how data is written or stored.

### Documentation

This change introduces a new feature that should be documented:

1. **Configuration Guide**: Document the new `source.split.file-enabled` and related options
2. **Performance Tuning Guide**: Explain when and how to use fine-grained splitting for optimal performance
3. **API Documentation**: Document the new `FormatMetadataReader` interface and implementations

The feature is disabled by default to maintain backward compatibility. Users can enable it by setting `source.split.file-enabled=true` in their table options.

### DIFF
--- a/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
@@ -654,6 +654,23 @@ public class CoreOptions implements Serializable {
                             "Open file cost of a source file. It is used to avoid reading"
                                     + " too many files with a source split, which can be very slow.");
 
+    public static final ConfigOption<Boolean> SOURCE_SPLIT_FILE_ENABLED =
+            key("source.split.file-enabled")
+                    .booleanType()
+                    .defaultValue(false)
+                    .withDescription(
+                            "If true, large data files are sub-divided at format-native boundaries "
+                                    + "(Parquet row groups, ORC stripes) during batch scans of "
+                                    + "append-only tables, improving read parallelism. Off by default.");
+
+    public static final ConfigOption<Integer> SOURCE_SPLIT_FILE_MAX_SPLITS =
+            key("source.split.file-max-splits")
+                    .intType()
+                    .defaultValue(100)
+                    .withDescription(
+                            "Upper bound on the number of fine-grained splits produced per data file. "
+                                    + "Any remaining boundaries are absorbed by the last split.");
+
     public static final ConfigOption<MemorySize> WRITE_BUFFER_SIZE =
             key("write-buffer-size")
                     .memoryType()
@@ -2898,6 +2915,14 @@ public class CoreOptions implements Serializable {
 
     public long splitOpenFileCost() {
         return options.get(SOURCE_SPLIT_OPEN_FILE_COST).getBytes();
+    }
+
+    public boolean sourceSplitFileEnabled() {
+        return options.get(SOURCE_SPLIT_FILE_ENABLED);
+    }
+
+    public int sourceSplitFileMaxSplits() {
+        return options.get(SOURCE_SPLIT_FILE_MAX_SPLITS);
     }
 
     public long writeBufferSize() {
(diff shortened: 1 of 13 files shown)


JingsongLi (commented):
Interesting feature for sub-file-level split generation (+1179 lines). Review:

1. **Design**: Splitting at row group (Parquet) and stripe (ORC) boundaries is a well-proven approach for improving read parallelism on large files. Good that this leverages existing `RawFile` offset/length infrastructure.

2. **FormatMetadataReader**: The abstraction for extracting split boundaries from different formats is clean. Make sure:
   - The metadata read is efficient (header-only, no data scanning)
   - It handles corrupted/incomplete files gracefully
   - Column projection metadata is compatible with partial reads

3. **FineGrainedSplitGenerator**: Key questions:
   - What's the minimum split size threshold? Generating too many tiny splits can hurt performance due to scheduling overhead.
   - How does this interact with existing split planning (target split size configuration)?
   - Is it opt-in via `CoreOptions`? It should be, since not all workloads benefit.

4. **DataSplit changes**: Adding boundary information to DataSplit affects serialization. Ensure backward compatibility with existing split formats (older Flink savepoints, etc.).

5. **Reader changes**: `ParquetReaderFactory.createReader(offset, length)` and ORC equivalent - these must correctly handle:
   - Row groups / stripes that span the boundary (read if start is within range)
   - Filter pushdown still working correctly with partial file reads
   - Statistics/min-max indexes still being leveraged

6. **Testing**: The test classes look good in scope. Add tests for edge cases: single row group file, file smaller than target split size, files with many tiny row groups.

This is a solid improvement for scan parallelism. The implementation looks well-structured.

Shekharrajak:
@JingsongLi Thanks for checking - working on comments and planning to have it in phases

JingsongLi:
Modifying the protocol of DataSplit still carries a relatively high risk. Currently, we recommend that the computing engine disassemble individual files and reassemble a new split concept according to its own logic.

PR #6923, opened 2025-12-29
[core][flink] Remove legacy codes for 0.2,0.3 and log system

JingsongLi:
### Purpose

Currently, it has been 3 years since Flink Table Store 0.2 0.3, and at that time, it was completely unpopular, so no one would use these capabilities, including logsystem.

But these codes have caused serious interference to us. Before the release of 2.0, we removed all related codes.

Includes:
- Snapshot: nullable for `totalRecordCount` and `deltaRecordCount`.
- Remove all log offset and log system in Flink Connector.


yuzelin (approved):
+1 for the removal. Waiting for all tests passing.

PR #6924, opened 2025-12-29
[core] Introduce 'nested_partial_update' agg func

JingsongLi:

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/aggregate/FieldNestedPartialUpdateAgg.java
+++ b/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/aggregate/FieldNestedPartialUpdateAgg.java
@@ -19,7 +19,6 @@
 package org.apache.paimon.mergetree.compact.aggregate;
 
 import org.apache.paimon.codegen.Projection;
-import org.apache.paimon.codegen.RecordEqualiser;
 import org.apache.paimon.data.BinaryRow;
 import org.apache.paimon.data.GenericArray;
 import org.apache.paimon.data.GenericRow;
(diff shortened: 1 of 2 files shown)


paimon-core/src/main/java/org/apache/paimon/mergetree/compact/aggregate/FieldNestedPartialUpdateAgg.java
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        InternalArray acc = (InternalArray) accumulator;
+        InternalArray input = (InternalArray) inputField;
+
+        List<InternalRow> rows = new ArrayList<>(acc.size() + input.size());
+        addNonNullRows(acc, rows);
+        addNonNullRows(input, rows);
+
+        if (keyProjection != null) {
```
zhoulii, line 76: when keyProjection is null ?

paimon-core/src/main/java/org/apache/paimon/mergetree/compact/aggregate/FieldNestedPartialUpdateAgg.java
```diff
@@ -0,0 +1,108 @@
...
+        checkArgument(!nestedKey.isEmpty());
+        this.keyProjection = newProjection(nestedType, nestedKey);
+        this.fieldGetters = new FieldGetter[nestedFields];
+        for (int i = 0; i < nestedFields; i++) {
+            fieldGetters[i] = InternalRow.createFieldGetter(nestedType.getTypeAt(i), i);
+        }
+    }
+
+    @Override
+    public Object agg(Object accumulator, Object inputField) {
```
zhoulii, line 64: Maybe add a null check for the key?

PR #6926, opened 2025-12-29
[python] Update PyPaimon Snapshot fields to not null

JingsongLi:
### Purpose

Adjust Python after https://github.com/apache/paimon/pull/6923


discivigour:
+1

PR #6927, opened 2025-12-30
[core] introduce a lazy reader for BTreeIndexReader to filter out files by predicate

steFaiz:
### Purpose

In the original design, each BTreeIndexReader only reads one BTree Index File. So we introduce a lazy reader which can dynamically filter out input files by predicate. For example, consider following files:

```
     +-------------------+        +-------------------+        +-------------------+
     |  min: 1, max: 10  |        |  min: 11, max: 20 |        |  min: 30, max: 40 |
     +-------------------+        +-------------------+        +-------------------+
```
* If the test predicate is EQUAL(15), we only need to read the second file.
* If the test predicate is EQUAL(25), we do not need to read any file.
* so on

This PR also:
1. covers the situation where the entire BTree Index File only contains nulls
2. fixes range query logic

Linked issue: https://github.com/apache/paimon/issues/6834

### Tests
Please see:
* org.apache.paimon.globalindex.btree.BTreeIndexReaderTest for single file reader
* org.apache.paimon.globalindex.btree.LazyFilteredBTreeIndexReaderTest for lazy filtered reader

### API and Format
None

### Documentation
Will be added in the future

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/globalindex/btree/BTreeIndexReader.java
+++ b/paimon-common/src/main/java/org/apache/paimon/globalindex/btree/BTreeIndexReader.java
@@ -366,7 +366,6 @@ public class BTreeIndexReader implements GlobalIndexReader {
         SstFileReader.SstFileIterator fileIter = reader.createIterator();
         fileIter.seekTo(keySerializer.serialize(from));
 
-        boolean skipped = false;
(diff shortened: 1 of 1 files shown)


JingsongLi (approved):
+1 Thanks @steFaiz

PR #6933, opened 2025-12-30
[core] Add option to support reading sequence_number in AuditLogTable and BinlogTable

wzhero1:
### Purpose

This PR adds support for reading `_SEQUENCE_NUMBER` field in `audit_log` and `binlog` system tables for primary key tables.

A new configuration option `table-read.sequence-number.enabled` (default: `false`) is introduced to control whether to include the `_SEQUENCE_NUMBER` field in the output schema of audit_log and binlog system tables. When enabled, users can access the internal sequence number that Paimon uses for ordering records within primary key tables.

Key changes:
- Added new `CoreOptions.TABLE_READ_SEQUENCE_NUMBER_ENABLED` configuration
- Modified `AuditLogTable` and `BinlogTable` to optionally include `_SEQUENCE_NUMBER` field in their row types
- Updated `ValueContentRowDataRecordIterator` to output sequence number when enabled
- Added schema validation to ensure this option is only valid for primary key tables
- Updated predicate conversion logic to handle the additional system field

### Tests

- `AuditLogTableTest` - basic auditlog table reading and sequence number with table option
- `BinlogTableTest` - basic binlog table reading and sequence number with table option
- `BatchFileStoreITCase` - Flink integration test for audit_log and binlog reading sequence number

### API and Format

- New configuration option: `table-read.sequence-number.enabled` (Boolean, default: `false`)
- When enabled, `audit_log` and `binlog` system tables will have an additional `_SEQUENCE_NUMBER` column after `rowkind`
- No storage format changes

### Documentation
- documentation updated in `system-tables.md`
- Configuration documentation updated in `core_configuration.html`

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/source/ValueContentRowDataRecordIterator.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/source/ValueContentRowDataRecordIterator.java
@@ -25,26 +25,21 @@ import org.apache.paimon.data.JoinedRow;
 import org.apache.paimon.reader.RecordReader;
 
 import java.io.IOException;
-import java.util.HashMap;
-import java.util.Map;
 
 /** A {@link RecordReader.RecordIterator} mapping a {@link KeyValue} to its value. */
 public class ValueContentRowDataRecordIterator extends ResetRowKindRecordIterator {
 
-    public static final String KEY_VALUE_SEQUENCE_NUMBER_ENABLED =
-            "key-value.sequence_number.enabled";
     private final boolean keyValueSequenceNumberEnabled;
 
     public ValueContentRowDataRecordIterator(RecordReader.RecordIterator<KeyValue> kvIterator) {
-        this(kvIterator, new HashMap<>(1));
+        this(kvIterator, false);
     }
 
     public ValueContentRowDataRecordIterator(
-            RecordReader.RecordIterator<KeyValue> kvIterator, Map<String, String> schemaOptions) {
+            RecordReader.RecordIterator<KeyValue> kvIterator,
+            boolean keyValueSequenceNumberEnabled) {
         super(kvIterator);
-        this.keyValueSequenceNumberEnabled =
-                Boolean.parseBoolean(
-                        schemaOptions.getOrDefault(KEY_VALUE_SEQUENCE_NUMBER_ENABLED, "false"));
+        this.keyValueSequenceNumberEnabled = keyValueSequenceNumberEnabled;
     }
 
     @Override
--- a/paimon-core/src/main/java/org/apache/paimon/table/system/AuditLogTable.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/system/AuditLogTable.java
@@ -84,7 +84,6 @@ import java.util.stream.Collectors;
 
 import static org.apache.paimon.CoreOptions.TABLE_READ_SEQUENCE_NUMBER_ENABLED;
 import static org.apache.paimon.catalog.Identifier.SYSTEM_TABLE_SPLITTER;
-import static org.apache.paimon.table.source.ValueContentRowDataRecordIterator.KEY_VALUE_SEQUENCE_NUMBER_ENABLED;
 
 /** A {@link Table} for reading audit log of table. */
 public class AuditLogTable implements DataTable, ReadonlyTable {
@@ -104,7 +103,7 @@ public class AuditLogTable implements DataTable, ReadonlyTable {
                 CoreOptions.fromMap(wrapped.options()).tableReadSequenceNumberEnabled();
 
         if (includeSequenceNumber) {
-            this.wrapped.options().put(KEY_VALUE_SEQUENCE_NUMBER_ENABLED, "true");
+            this.wrapped.options().put(CoreOptions.KEY_VALUE_SEQUENCE_NUMBER_ENABLED.key(), "true");
             specialFields.add(SpecialFields.SEQUENCE_NUMBER);
         }
     }
--- a/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
@@ -835,6 +835,16 @@ public class CoreOptions implements Serializable {
                             "Whether to include the _SEQUENCE_NUMBER field when reading the audit_log or binlog "
                                     + "system tables. This is only valid for primary key tables.");
 
+    @ExcludeFromDocumentation("Internal use only")
+    public static final ConfigOption<Boolean> KEY_VALUE_SEQUENCE_NUMBER_ENABLED =
+            key("key-value.sequence_number.enabled")
+                    .booleanType()
+                    .defaultValue(false)
+                    .withDescription(
+                            "Whether to include the _SEQUENCE_NUMBER field when reading key-value data. "
+                                    + "This is an internal option used by AuditLogTable and BinlogTable "
+                                    + "when table.read_sequence_number_enabled is set to true.");
+
     public static final ConfigOption<String> SEQUENCE_FIELD =
             key("sequence.field")
                     .stringType()
--- a/paimon-core/src/main/java/org/apache/paimon/table/source/KeyValueTableRead.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/source/KeyValueTableRead.java
@@ -18,6 +18,7 @@
 
 package org.apache.paimon.table.source;
 
+import org.apache.paimon.CoreOptions;
 import org.apache.paimon.KeyValue;
 import org.apache.paimon.annotation.VisibleForTesting;
 import org.apache.paimon.data.InternalRow;
@@ -169,10 +170,17 @@ public final class KeyValueTableRead extends AbstractDataTableRead {
             @Nullable
             @Override
             public RecordIterator<InternalRow> readBatch() throws IOException {
+                boolean keyValueSequenceNumberEnabled =
+                        Boolean.parseBoolean(
+                                schemaOptions.getOrDefault(
+                                        CoreOptions.KEY_VALUE_SEQUENCE_NUMBER_ENABLED.key(),
+                                        "false"));
+
                 RecordIterator<KeyValue> batch = reader.readBatch();
                 return batch == null
                         ? null
-                        : new ValueContentRowDataRecordIterator(batch, schemaOptions);
+                        : new ValueContentRowDataRecordIterator(
+                                batch, keyValueSequenceNumberEnabled);
             }
 
             @Override


docs/layouts/shortcodes/generated/core_configuration.html
```diff
@@ -146,6 +146,12 @@
             <td>String</td>
             <td>Fields that are ignored for comparison while generating -U, +U changelog for the same record. This configuration is only valid for the changelog-producer.row-deduplicate is true.</td>
         </tr>
+        <tr>
+            <td><h5>changelog-read.sequence-number.enabled</h5></td>
+            <td style="word-wrap: break-word;">false</td>
+            <td>Boolean</td>
+            <td>Whether to include _SEQUENCE_NUMBER field in audit_log and binlog system tables. This is only valid for primary key tables.</td>
```
yunfengzhou-hub: The output result of an audit log table can be controlled by `incremental-between-scan-mode`, where `changelog` is only one of the options. So we might better not include the word "changelog" in the config option name.

paimon-core/src/main/java/org/apache/paimon/table/system/AuditLogTable.java
```diff
@@ -87,25 +87,37 @@
...
-            };
+    protected final FileStoreTable wrapped;
 
-    private final FileStoreTable wrapped;
+    /** Number of special fields (rowkind, and optionally _SEQUENCE_NUMBER). */
+    protected final int specialFieldCount;
 
     public AuditLogTable(FileStoreTable wrapped) {
         this.wrapped = wrapped;
+        this.wrapped.schema().options().put(AUDIT_LOG_ENABLED, "true");
```
yunfengzhou-hub: How about organize the two new configurations in the following way?
1. User sets config A.
2. In `AuditLogTable`, the code detects whether config A is set. If so, set config B to true.
3. In `ValueContentRowDataRecordIterator`, the code only checks config B, instead of configA && configB.

The benefit of such structure is that it can decouple system table concepts from `ValueContentRowDataRecordIterator`. config B need not be related to audit log table.

paimon-core/src/main/java/org/apache/paimon/table/system/AuditLogTable.java
```diff
@@ -87,25 +87,37 @@
...
-                                p.type(),
-                                p.index() - 1,
-                                p.fieldName(),
-                                p.literals()));
-            };
+    protected final FileStoreTable wrapped;
 
-    private final FileStoreTable wrapped;
+    /** Number of special fields (rowkind, and optionally _SEQUENCE_NUMBER). */
+    protected final int specialFieldCount;
```
yunfengzhou-hub: How about introducing `List<SpecialField> specialFields`? If so, we will be able to change
```java
fields.add(SpecialFields.ROW_KIND);
if (specialFieldCount > 1) {
    fields.add(SpecialFields.SEQUENCE_NUMBER);
}
```
into 
```java
fields.add(specialFields);
```

paimon-core/src/main/java/org/apache/paimon/table/system/AuditLogTable.java
```diff
@@ -723,15 +756,26 @@ public void setRowKind(RowKind kind) {
...
         }
 
+        @Override
+        public long getLong(int pos) {
+            int index = indexMapping[pos];
+            if (index == AuditLogRead.SEQUENCE_NUMBER_INDEX) {
+                // _SEQUENCE_NUMBER is at index 0 in bottom output
+                return row.getLong(0);
+            }
+            return super.getLong(pos);
```
yunfengzhou-hub: If _SEQUENCE_NUMBER is at index 0, then the other fields should be acquired through `pos - 1`.

paimon-core/src/main/java/org/apache/paimon/table/system/AuditLogTable.java
```diff
@@ -723,15 +756,26 @@ public void setRowKind(RowKind kind) {
...
                 return false;
             }
             return super.isNullAt(pos);
         }
 
+        @Override
+        public long getLong(int pos) {
+            int index = indexMapping[pos];
+            if (index == AuditLogRead.SEQUENCE_NUMBER_INDEX) {
+                // _SEQUENCE_NUMBER is at index 0 in bottom output
```
yunfengzhou-hub: The reason behind this comment is located in another class(`ValueContentRowDataRecordIterator`), and these two classes are mainly associated with `AUDIT_LOG_ENABLED`, which shows a generic concept not related to sequence number.

Thus it might increase the other developer's burden to understand why "_SEQUENCE_NUMBER is at index 0". We might need to think about how to increase code readability here.

paimon-core/src/test/java/org/apache/paimon/table/system/AuditLogTableTest.java
```diff
@@ -48,60 +48,117 @@
...
+        return (AuditLogTable) catalog.getTable(auditLogTableId);
+    }
 
+    private void writeTestData(FileStoreTable table) throws Exception {
         write(table, GenericRow.ofKind(RowKind.INSERT, 1, 1, 1));
         write(table, GenericRow.ofKind(RowKind.DELETE, 1, 1, 1));
         write(table, GenericRow.ofKind(RowKind.INSERT, 1, 2, 5));
         write(table, GenericRow.ofKind(RowKind.UPDATE_BEFORE, 1, 2, 5));
-        write(table, GenericRow.ofKind(RowKind.UPDATE_AFTER, 1, 4, 6));
+        write(table, GenericRow.ofKind(RowKind.UPDATE_AFTER, 1, 2, 6));
```
yunfengzhou-hub, line 137: This change seems unnecessary.

paimon-core/src/test/java/org/apache/paimon/table/system/AuditLogTableTest.java
```diff
@@ -48,60 +48,117 @@
...
-    }
-
     private List<InternalRow> getExpectedResult() {
         List<InternalRow> expectedRow = new ArrayList<>();
         expectedRow.add(
                 GenericRow.of(BinaryString.fromString(RowKind.DELETE.shortString()), 1, 1, 1));
         expectedRow.add(
                 GenericRow.of(
-                        BinaryString.fromString(RowKind.UPDATE_BEFORE.shortString()), 1, 2, 5));
+                        BinaryString.fromString(RowKind.UPDATE_AFTER.shortString()), 1, 2, 6));
```
yunfengzhou-hub, line 147: In the original code, the expected result contains 4 rows. Here only three rows are left, missing an `UPDATE_BEFORE` row.

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/BatchFileStoreITCase.java
```diff
@@ -1053,6 +1053,108 @@ public void testBinlogTableWithProjection() {
                 .containsExactly(Row.of("+I", new String[] {"A"}));
     }
 
+    @Test
+    public void testAuditLogTableWithSequenceNumberEnabled() {
+        // Create primary key table with changelog-read.sequence-number.enabled option
+        sql(
+                "CREATE TABLE test_table_seq (a int PRIMARY KEY NOT ENFORCED, b int, c AS a + b) "
+                        + "WITH ('changelog-read.sequence-number.enabled'='true');");
```
yunfengzhou-hub: Despite that users are allowed to set `changelog-read.sequence-number.enabled` when creating or altering table, a more recommended way is to configure it dynamically in the SELECT query through Flink [SQL Hints](https://nightlies.apache.org/flink/flink-docs-release-1.20/docs/dev/table/sql/queries/hints/). It should be more suitable for such configurations that are only useful in specific read queries. The usage of SQL hints is like follows
```sql
SELECT * FROM `test_table_seq$audit_log`/*+ OPTIONS('changelog-read.sequence-number.enabled' = 'true') */;
```
Let's add test cases for this use case as well.

yunfengzhou-hub: Let's also add some test cases to verify what will happen if `changelog-read.sequence-number.enabled` is configured on append-only tables.

yunfengzhou-hub (commented):

docs/layouts/shortcodes/generated/core_configuration.html
wzhero1: Changed to `table-read.sequence-number.enabled` for clearer semantics, though I'm not sure if this will be supported on regular tables in the future. If it is, the scope of the parameter would become limited.

paimon-core/src/main/java/org/apache/paimon/table/system/AuditLogTable.java
wzhero1: Done. The parameters are now aggregated and updated in `AuditLogTable`, and `ValueContentRowDataRecordIterator` is only aware of the `addSequenceNumberFirst` behavior. 

The current parameter is `AUDIT_ADD_SEQUENCE_NUMBER_FIRST`, which clearly reflects what it does (the naming can be discussed further, but it's an internal parameter so it's easy to adjust).

paimon-core/src/main/java/org/apache/paimon/table/system/AuditLogTable.java
wzhero1: changed to `List<SpecialField> specialFields`

paimon-core/src/main/java/org/apache/paimon/table/system/AuditLogTable.java
wzhero1: The mapping logic is implemented in `AuditLogRead.defaultProjection()` (for `select *`) and `AuditLogRead.buildProjection(RowType readType)` (for selecting specific fields).

paimon-core/src/main/java/org/apache/paimon/table/system/AuditLogTable.java
wzhero1: I've optimized the `_SEQUENCE_NUMBER` index logic in `defaultProjection()` and `buildProjection(RowType readType)`, so `getLong` no longer needs special handling here.

paimon-core/src/test/java/org/apache/paimon/table/system/AuditLogTableTest.java
wzhero1: The original test case is incorrect - the primary keys for `RowKind.UPDATE_BEFORE` and `RowKind.UPDATE_AFTER` are inconsistent. This scenario should not occur in normal circumstances.

paimon-core/src/test/java/org/apache/paimon/table/system/AuditLogTableTest.java
wzhero1: The original test case is incorrect - the primary keys for `RowKind.UPDATE_BEFORE` and `RowKind.UPDATE_AFTER` are inconsistent. This scenario should not occur in normal circumstances.

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/BatchFileStoreITCase.java
wzhero1: This new parameter cannot be applied to SQL hints, because the `rowType` passed by the operator is determined before obtaining dynamic options. 

* Before processing SQL hint parameters, the Flink job first constructs the `AuditLogTable` to get the `rowtype` for building upstream/downstream `rowtype` transmission information. 
* Later, when reading `AuditLogTable`, it copies the `dynamicOptions`. 
* At this point, the operator's `rowtype` (without `SEQUENCE_NUM` before obtaining dynamic parameters) does not match the task's processing `rowtype` (with `SEQUENCE_NUM`).

wzhero1: In short, for `select *`, the operator's `rowtype` is generated before dynamic options are applied.

wzhero1: I've added `dynamicOptions` detection to prevent users from accidentally misusing it in SQL hints.
```java
public Table copy(Map<String, String> dynamicOptions) {
    if (Boolean.parseBoolean(
            dynamicOptions.getOrDefault(
                    TABLE_READ_SEQUENCE_NUMBER_ENABLED.key(), "false"))) {
                "table-read.sequence-number.enabled is not supported by hint.");
    }
    return new AuditLogTable(wrapped.copy(dynamicOptions));
}
```

docs/content/concepts/system-tables.md
```diff
@@ -136,6 +136,51 @@ SELECT * FROM my_table$audit_log;
 */
 ```
 
+#### Reading with Sequence Number
+
+For primary key tables, you can enable the `table-read.sequence-number.enabled` option to include the `_SEQUENCE_NUMBER` field in the output.
+
+{{< tabs "audit-log-sequence-number" >}}
+
+{{< tab "Enable via CREATE TABLE" >}}
```
yunfengzhou-hub, line 149: It might be better to put tab `Enable via ALTER TABLE` first. Few users should have this requirement when the table was initially created.

docs/content/concepts/system-tables.md
```diff
@@ -136,6 +136,51 @@ SELECT * FROM my_table$audit_log;
 */
 ```
 
+#### Reading with Sequence Number
```
yunfengzhou-hub: It might be better to remove this title to keep the TOC simpler.

yunfengzhou-hub (commented):
Thanks for the PR. Left some comments about the document

docs/content/concepts/system-tables.md
wzhero1: done

docs/content/concepts/system-tables.md
wzhero1: done

paimon-core/src/main/java/org/apache/paimon/table/source/ValueContentRowDataRecordIterator.java
```diff
@@ -19,16 +19,32 @@
...
+    public static final String KEY_VALUE_SEQUENCE_NUMBER_ENABLED =
+            "key-value.sequence_number.enabled";
+    private final boolean keyValueSequenceNumberEnabled;
+
     public ValueContentRowDataRecordIterator(RecordReader.RecordIterator<KeyValue> kvIterator) {
+        this(kvIterator, new HashMap<>(1));
+    }
+
+    public ValueContentRowDataRecordIterator(
+            RecordReader.RecordIterator<KeyValue> kvIterator, Map<String, String> schemaOptions) {
```
yuzelin: I think a boolean value `sequenceNumberEnabled` is enough?

paimon-core/src/main/java/org/apache/paimon/table/system/AuditLogTable.java
```diff
@@ -234,6 +249,11 @@ public InnerTableRead newRead() {
 
     @Override
     public Table copy(Map<String, String> dynamicOptions) {
+        if (Boolean.parseBoolean(
+                dynamicOptions.getOrDefault(TABLE_READ_SEQUENCE_NUMBER_ENABLED.key(), "false"))) {
+            throw new UnsupportedOperationException(
+                    "table-read.sequence-number.enabled is not supported by hint.");
+        }
```
yuzelin, line 255: Why not supported by hint?

paimon-core/src/main/java/org/apache/paimon/table/source/ValueContentRowDataRecordIterator.java
wzhero1: I'd prefer to avoid being too tightly coupled to `sequenceNumberEnabled` here. Passing the entire `schemaOption` map provides better flexibility for future extensions.

paimon-core/src/main/java/org/apache/paimon/table/system/AuditLogTable.java
wzhero1: hint is not supported by this feature, discuss from here，maybe supported later: https://github.com/apache/paimon/pull/6933#discussion_r2675445601

yuzelin: OK

PR #6935, opened 2025-12-31
[flink] fix compact_database parameter mode default value

### LINKED ISSUE #6934: [Bug] compact_database parameter mode default value is wrong
### Search before asking

### Paimon version

1.2.0

### Compute Engine

flink 1.20.3

### Minimal reproduce step

CALL sys.compact_database(
    including_databases = > 'xxx'
  );

The documentation states that the default value for `mode` is `combined`, but the actual default value is `divided`.
Is the documentation wrong, or is the code implementation wrong?

### Anything else?

_No response_

sd4324530:
### Purpose
Linked issue: close https://github.com/apache/paimon/issues/6934
fix compact_database parameter mode default value

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/CompactDatabaseAction.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/CompactDatabaseAction.java
@@ -64,7 +64,7 @@ public class CompactDatabaseAction extends ActionBase {
     @Nullable private Pattern excludingPattern;
     private Pattern databasePattern = Pattern.compile(".*");
 
-    private MultiTablesSinkMode databaseCompactMode = MultiTablesSinkMode.DIVIDED;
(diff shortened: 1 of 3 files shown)


sd4324530:
@JingsongLi
All checks have passed.
Could you please take a look?

JingsongLi (commented):
Clean bug fix. The change makes sense - changing the default `compact_database` mode from DIVIDED to COMBINED.

A few notes:
1. **Breaking change**: This changes the default behavior. Users who relied on the implicit DIVIDED default will now get COMBINED behavior unless they explicitly specify `--mode divided`. This should be documented in release notes.

2. **Test update**: Good that the test is updated to explicitly pass `--mode divided` where DIVIDED behavior is expected, rather than relying on the default.

3. **Linked issue #6934**: The fix aligns with the documented default in the documentation/user-facing docs.

LGTM with the caveat that this is technically a behavior change and should be called out in release notes.

JingsongLi:
re-trigger tests

JingsongLi:
This will cause compatibility issues, and we need to update the documentation. If you're willing, you can submit another PR for the document.

PR #6938, opened 2025-12-31
[pypaimon] Creation_time should be utc timestamp

leaves12138:
### Purpose

Linked issue: close #xxx


XiaoHongbo-Hope:
Does pypaimon create_time match create_time created by java now?

PR #6943, opened 2026-01-02
Fix EqualNullSafe is not correct when column has null value. Bug[#6931]

### LINKED ISSUE #6931: [Bug]Filter pushdown `!(col <=> value)` is not correct when col has null value
### Search before asking

### Paimon version

Latest version,  1.4-SNAPSHOT

### Compute Engine

Spark

### Minimal reproduce step

Step 1: prepare data
```SQL
CREATE TABLE T (a INT, b INT);
INSERT INTO T VALUES (1, null);
```

Step 2: this result is true
```
SELECT !(b <=> 1) FROM T;
```

Step 3: but this return empty rows
```
SELECT a, b FROM T WHERE !(b <=> 1);
```

After filter pushdown, the result is not correct.

Spark supports `Equal (i.e. =)` and `EqualNullSafe (i.e. <=>)`. After pushdown, paimon treat both as `Equal` and simplify `Not Equal (two predicate expressions)` as `NotEqual (one predicate expression)`.

This table shows the differences between spark and paimon when col is null:

||WHERE col = 1|WHERE col <=> 1|WHERE !(col = 1)|WHERE !(col <=> 1)|
|-|-|-|-|-|
|spark|false|false|false|**true**|
|paimon|false|false|false|**false**|

### Anything else?

_No response_

venusMask:
### Purpose

Linked issue: close #6931 

EqualNullSafe supports complete semantics。

### DIFF
--- a/paimon-core/src/test/java/org/apache/paimon/table/PrimaryKeySimpleTableTest.java
+++ b/paimon-core/src/test/java/org/apache/paimon/table/PrimaryKeySimpleTableTest.java
@@ -101,6 +101,7 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Random;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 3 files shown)


JingsongLi (commented):
Thanks @venusMask , can you add a test in `PaimonSourceTest`?

JingsongLi:
+1 thanks @venusMask

PR #6944, opened 2026-01-02
[python] support table scan with row range

universe-hcy:
### Purpose
In model training and inference, it is common to record unread row intervals to checkpoints for recovery after worker failure. Currently, the with_stard method cannot support this feature and needs to support row_range for table scanning.

Linked issue: close #xxx

### Tests
test_data_blob_writer_with_row_range in blob_table_test.py

### DIFF
--- a/paimon-python/pypaimon/read/scanner/full_starting_scanner.py
+++ b/paimon-python/pypaimon/read/scanner/full_starting_scanner.py
@@ -63,6 +63,8 @@ class FullStartingScanner(StartingScanner):
 
         self.idx_of_this_subtask = None
         self.number_of_para_subtasks = None
+        self.start_row_of_this_subtask = None
+        self.end_row_of_this_subtask = None
 
         self.only_read_real_buckets = True if options.bucket() == BucketMode.POSTPONE_BUCKET.value else False
         self.data_evolution = options.data_evolution_enabled()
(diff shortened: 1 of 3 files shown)


paimon-python/pypaimon/read/scanner/full_starting_scanner.py
```diff
@@ -125,33 +127,16 @@ def with_shard(self, idx_of_this_subtask, number_of_para_subtasks) -> 'FullStart
...
-        if self.idx_of_this_subtask < remainder:
-            num_row = base_rows_per_shard + 1
-            start_row = self.idx_of_this_subtask * (base_rows_per_shard + 1)
-        else:
-            num_row = base_rows_per_shard
-            start_row = (remainder * (base_rows_per_shard + 1) +
-                         (self.idx_of_this_subtask - remainder) * base_rows_per_shard)
-
-        end_row = start_row + num_row
+    def with_row_range(self, start_row, end_row) -> 'FullStartingScanner':
```
JingsongLi, line 132: Can we raise a exception for primary key table?

paimon-python/pypaimon/read/table_scan.py
```diff
@@ -71,3 +71,7 @@ def _create_starting_scanner(self) -> Optional[StartingScanner]:
     def with_shard(self, idx_of_this_subtask, number_of_para_subtasks) -> 'TableScan':
         self.starting_scanner.with_shard(idx_of_this_subtask, number_of_para_subtasks)
         return self
+
+    def with_row_range(self, start_row, end_row) -> 'TableScan':
```
JingsongLi, line 75: Are we returning this exact number of lines, or can it be approximate? This needs to be specified clearly in the comments.

paimon-python/pypaimon/read/scanner/full_starting_scanner.py
```diff
@@ -555,7 +579,14 @@ def sort_key(manifest_entry: ManifestEntry) -> tuple:
         for entry in sorted_entries:
             partitioned_files[(tuple(entry.partition.values), entry.bucket)].append(entry)
 
-        if self.idx_of_this_subtask is not None:
+        print("self.start_row_of_this_subtask:{}".format(self.start_row_of_this_subtask))
+        if self.start_row_of_this_subtask is not None:
```
XiaoHongbo-Hope, line 586: logger may be better

paimon-python/pypaimon/read/scanner/full_starting_scanner.py
XiaoHongbo-Hope: Can we use with_row_range and with_shard at the same time or not?

paimon-python/pypaimon/read/scanner/full_starting_scanner.py
universe-hcy: This is for debugging purposes, I will delete it soon.

paimon-python/pypaimon/read/scanner/full_starting_scanner.py
universe-hcy: not, It is difficult for all users to have a completely consistent behavior when using at the same time. Added an exception to avoid this situation.

paimon-python/pypaimon/read/table_scan.py
```diff
@@ -71,3 +71,11 @@ def _create_starting_scanner(self) -> Optional[StartingScanner]:
...
+        """
+        Filter file entries by row range. The row_id corresponds to the row position of the
+        file in all file entries in table scan's partitioned_files.
+        """
```
discivigour, line 79: It might be better to clearly describe the inclusion and exclusion of idx within the range in the comments.

PR #6949, opened 2026-01-04
[core]JindoFileIO support cache using JindoCache

timmyyao:
### Purpose
We tend to persist files of Paimon table on object storage such as S3 and OSS. However the throughput or latency of such storage is usually limited, making IO performance become the bottleneck in ETL or OLAP occasions with high concurrency. Thus we introduce cache when FileIO visits files within a Paimon table. The solution is to support JindoCache in JindoFileIO, which is a distributed filesystem cache system. And table-level cache policy is implemented, which can be supported by REST server such as DLF.

### DIFF
--- a/paimon-api/src/main/java/org/apache/paimon/rest/RESTCatalogOptions.java
+++ b/paimon-api/src/main/java/org/apache/paimon/rest/RESTCatalogOptions.java
@@ -104,20 +104,20 @@ public class RESTCatalogOptions {
                     .noDefaultValue()
                     .withDescription("REST Catalog DLF OSS endpoint.");
 
-    public static final ConfigOption<Boolean> DLF_FILE_IO_CACHE_ENABLED =
-            ConfigOptions.key("dlf.io-cache-enabled")
+    public static final ConfigOption<Boolean> IO_CACHE_ENABLED =
+            ConfigOptions.key("io-cache.enabled")
                     .booleanType()
                     .defaultValue(false)
                     .withDescription(
                             "Enable cache for visiting files using file io (currently only JindoFileIO supports cache).");
-    public static final ConfigOption<String> DLF_FILE_IO_CACHE_WHITELIST_PATH =
-            ConfigOptions.key("dlf.io-cache.whitelist-path")
+    public static final ConfigOption<String> IO_CACHE_WHITELIST_PATH =
+            ConfigOptions.key("io-cache.whitelist-path")
                     .stringType()
                     .defaultValue("bucket-,manifest")
                     .withDescription(
                             "Cache is only applied to paths which contain the specified pattern, and * means all paths.");
-    public static final ConfigOption<String> DLF_FILE_IO_CACHE_POLICY =
-            ConfigOptions.key("dlf.io-cache.policy")
+    public static final ConfigOption<String> IO_CACHE_POLICY =
+            ConfigOptions.key("io-cache.policy")
                     .stringType()
                     .noDefaultValue()
                     .withDescription(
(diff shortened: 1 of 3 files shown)


timmyyao:
> @timmyyao I'm not using DLF, but the data is still stored on OSS. Does this also apply? like this:
> 
> ```sql
> CREATE CATALOG paimon_catalog WITH (
> 'type' = 'paimon',
> 'metastore' = 'hive',
> 'warehouse' = 'oss://xxxx.oss-dls.aliyuncs.com/user/hive/warehouse'
> );
> ```
The major purpose of this PR is to provide a table-level cache policy which can be supported by DLF server-side and cache is backed by JindoCache. So related configs are generated in RESTCatalog and RESTTokenFileIO according to the policy defined at server-side. If you just use filesystem or hive catalog using JindoFileIO, you can simply apply the cache configs to JindoSDK (fs.xengine, fs.jindocache.namespace.rpc.address) as long as you have a JindoCache cluster. But there is no server-side support for cache management, you can only adjust at client-side to decide whether or not to enable cache.

paimon-api/src/main/java/org/apache/paimon/rest/RESTCatalogOptions.java
```diff
@@ -103,4 +103,27 @@ public class RESTCatalogOptions {
                     .stringType()
                     .noDefaultValue()
                     .withDescription("REST Catalog DLF OSS endpoint.");
+
+    public static final ConfigOption<Boolean> DLF_FILE_IO_CACHE_ENABLED =
+            ConfigOptions.key("dlf.io-cache-enabled")
```
JingsongLi: just `io-cache.enabled`.

paimon-api/src/main/java/org/apache/paimon/rest/RESTCatalogOptions.java
```diff
@@ -103,4 +103,27 @@ public class RESTCatalogOptions {
...
+                    .booleanType()
+                    .defaultValue(false)
+                    .withDescription(
+                            "Enable cache for visiting files using file io (currently only JindoFileIO supports cache).");
+    public static final ConfigOption<String> DLF_FILE_IO_CACHE_WHITELIST_PATH =
+            ConfigOptions.key("dlf.io-cache.whitelist-path")
```
JingsongLi: just `io-cache.whitelist-path`

paimon-api/src/main/java/org/apache/paimon/rest/RESTCatalogOptions.java
```diff
@@ -103,4 +103,27 @@ public class RESTCatalogOptions {
...
+                    .stringType()
+                    .defaultValue("bucket-,manifest")
+                    .withDescription(
+                            "Cache is only applied to paths which contain the specified pattern, and * means all paths.");
+    public static final ConfigOption<String> DLF_FILE_IO_CACHE_POLICY =
+            ConfigOptions.key("dlf.io-cache.policy")
```
JingsongLi: just `io-cache.policy`

paimon-api/src/main/java/org/apache/paimon/rest/RESTCatalogOptions.java
timmyyao: fixed

paimon-api/src/main/java/org/apache/paimon/rest/RESTCatalogOptions.java
timmyyao: fixed

paimon-api/src/main/java/org/apache/paimon/rest/RESTCatalogOptions.java
timmyyao: fixed

JingsongLi:
+1

PR #6953, opened 2026-01-05
[variant] Fix reading empty shredded variant via variantAccess

Zouxxyy:
### Purpose

Fix reading empty shredded variant via variantAccess


JingsongLi:
+1

PR #6961, opened 2026-01-06
[core] Support incremental clustering for append bucketed table

LsomeYeah:
### Purpose

Linked issue: close #xxx

https://github.com/apache/paimon/pull/6835 was deprecated.

By default, Paimon's append bucketed tables maintain data ordering. However, this ordering requirement can be relaxed to enable additional optimizations.

This PR introduces the ability to disable ordering requirements for append bucketed tables, allowing incremental clustering within buckets. When ordering is not strictly required, data can be incrementally clustered within each bucket, significantly improving query performance for bucket-key + clustering-key combinations.


JingsongLi (approved):
+1 Thanks @LsomeYeah

PR #6964, opened 2026-01-07
PartitionStatistics support totalBuckets

jianguotian:
### Purpose

PartitionStatistics add totalBuckets field, get value from ManifestEntry.totalBuckets()

### Tests

org.apache.paimon.flink.RESTCatalogITCase#testTotalBucketsStatistics:  
verify totalBuckets stats for 3 bucket mode

### API and Format

Partition and PartitionStatistics add new totalBuckets field

### Documentation

rest-catalog-open-api doc:
Partition and PartitionStatistics add totalBuckets field

### DIFF
--- a/paimon-api/src/main/java/org/apache/paimon/partition/PartitionStatistics.java
+++ b/paimon-api/src/main/java/org/apache/paimon/partition/PartitionStatistics.java
@@ -61,6 +61,8 @@ public class PartitionStatistics implements Serializable {
     @JsonProperty(FIELD_LAST_FILE_CREATION_TIME)
     protected final long lastFileCreationTime;
 
+    // defaults to 0 if this field is absent in the serialized data (e.g., from an older Paimon
+    // version)
     @JsonProperty(FIELD_TOTAL_BUCKETS)
     protected final int totalBuckets;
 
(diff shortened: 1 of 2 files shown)


JingsongLi:
Is this change compatible?
1. Using the old version to read will ignore totalBucket, right?
2. Can the old version of the client, which lacks totalBucket, be deserialized?

Perhaps you need to conduct some tests to ensure

paimon-api/src/main/java/org/apache/paimon/partition/PartitionStatistics.java
```diff
@@ -60,18 +61,23 @@ public class PartitionStatistics implements Serializable {
     @JsonProperty(FIELD_LAST_FILE_CREATION_TIME)
     protected final long lastFileCreationTime;
 
+    @JsonProperty(FIELD_TOTAL_BUCKETS)
+    protected final int totalBuckets;
```
JingsongLi, line 67: Please add comments to explain value when missing.

JingsongLi, line 67: Also add test to missing case.

jianguotian: > Please add comments to explain value when missing.

done

jianguotian: add PartitionStatisticsTest

JingsongLi:
+1

PR #6966, opened 2026-01-07
[flink] Optimize lookup table refresh with full load for snapshot backlog.

### LINKED ISSUE #6965: [Feature] Optimize lookup table refresh with full load for snapshot backlog.
### Search before asking

### Motivation

In Flink lookup joins with Paimon tables, if the lookup table has frequent updates during the refresh interval, a large number of snapshots can accumulate.

The current implementation refreshes the lookup table by incrementally loading each pending snapshot one by one. When there are many pending snapshots (e.g., dozens or hundreds), this incremental approach can be very slow and inefficient.

### Solution

Add a new configuration option `lookup.refresh.full-load-threshold` that allows users to set a threshold for the number of pending snapshots. When the number of snapshots pending for refresh exceeds this threshold, the system will perform a full table reload (close and reopen) instead of incremental refresh.

### Anything else?

_No response_

zhoulii:
### Purpose

Linked issue: close #6965

### DIFF
--- a/docs/layouts/shortcodes/generated/flink_connector_configuration.html
+++ b/docs/layouts/shortcodes/generated/flink_connector_configuration.html
@@ -94,7 +94,7 @@ under the License.
         </tr>
         <tr>
             <td><h5>lookup.refresh.full-load-threshold</h5></td>
-            <td style="word-wrap: break-word;">2147483647</td>
+            <td style="word-wrap: break-word;">(none)</td>
             <td>Integer</td>
             <td>If the pending snapshot count exceeds this threshold, lookup table will discard incremental updates and refresh the entire table from the latest snapshot. This can improve performance when there are many snapshots pending. Set to a reasonable value (e.g., 10) to enable this optimization. Default is Integer.MAX_VALUE (disabled). </td>
         </tr>


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/FlinkConnectorOptions.java
```diff
@@ -292,6 +292,15 @@ public class FlinkConnectorOptions {
                                     + "cache refreshing is forbidden. Blacklist format is start1->end1,start2->end2,... , "
                                     + "and the time format is yyyy-MM-dd HH:mm. Only used when lookup table is FULL cache mode.");
 
+    public static final ConfigOption<Integer> LOOKUP_REFRESH_FULL_LOAD_THRESHOLD =
+            ConfigOptions.key("lookup.refresh.full-load-threshold")
+                    .intType()
+                    .defaultValue(Integer.MAX_VALUE)
```
JingsongLi: Default value NULL is better.

zhoulii: fixed.

PR #6968, opened 2026-01-07
[core] Rollback 'COMPACT' commit for row-level operations

JingsongLi:
### Purpose

To prevent row-level operations (MERGE INTO /  DELETE / UPDATE) from failing, we can force a rollback of COMPACT commits that do not affect correctness.

### Tests

- `RESTCatalogTest.testConflictRollback`


qingwei727:
+1

PR #6976, opened 2026-01-08
[python] Fix KeyError when reading manifest files without _VALUE_STATS_COLS field

XiaoHongbo-Hope:
…S_COLS field

### Purpose

Linked issue: close #xxx

### DIFF
--- a/paimon-python/pypaimon/tests/reader_base_test.py
+++ b/paimon-python/pypaimon/tests/reader_base_test.py
@@ -816,6 +816,23 @@ class ReaderBasicTest(unittest.TestCase):
         manifest_file_name = f"manifest-test-{test_name}"
         manifest_manager.write(manifest_file_name, [entry])
 
+        if value_stats_cols is None:
+            from pathlib import Path
+            import fastavro
+            manifest_path = Path(table.table_path) / "manifest" / manifest_file_name
+            with table.file_io.new_input_stream(str(manifest_path)) as f:
+                reader = fastavro.reader(f)
+                records = list(reader)
+                file_dict = records[0]['_FILE']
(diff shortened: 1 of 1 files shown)


JingsongLi:
> > I'm also curious why the none _VALUE_STATS_COLS is not written out when using Python to write?
> 
> Confused too, User says that table is written by older version pypaimon (0.3), I need to double check it, along with the IndexError.

Not just old version, Python write append table with full metadata.stats-mode should produce none COLS too.

paimon-python/pypaimon/tests/reader_base_test.py
```diff
@@ -813,6 +816,23 @@ def _test_value_stats_cols_case(self, manifest_manager, table, value_stats_cols,
...
+            from pathlib import Path
+            import fastavro
+            manifest_path = Path(table.table_path) / "manifest" / manifest_file_name
+            with table.file_io.new_input_stream(str(manifest_path)) as f:
+                reader = fastavro.reader(f)
+                records = list(reader)
+                file_dict = records[0]['_FILE']
+                try:
+                    if file_dict['_VALUE_STATS_COLS'] is None:
+                        pass
```
XiaoHongbo-Hope, line 829: **Add an assertion to show that None COLS works well with code `if file_dict['_VALUE_STATS_COLS'] is None`.** The root cause of the KeyError is that very older version pypaimon (before [205.09.10](https://github.com/apache/paimon/commit/b7d3e1a37)) . And double checked user's test table is written before that by pypaimon.  



 In short, I suggest to discard this fix and ask user do not use the older table, as the root cause is bug during writing, what's your suggestion?

PR #6980, opened 2026-01-08
[core] Add compaction metrics system table. #6880

### LINKED ISSUE #6880: [Feature] Add compaction duration metrics system table
### Search before asking

### Motivation

While Paimon currently provides metrics reports, they require viewing them through the Flink web UI, and historical compaction metrics become unavailable once a task restarts or stops. Furthermore, Spark engine metrics reports only show scans, leading to inconsistencies when pushing metrics to the engine.

Users need to rely on compaction metrics to determine task performance.

### Solution

A new Compaction Metrics System Table has been added. Each time a compaction snapshot is submitted, a compaction metrics record is written to the metrics directory under the table path.

### Anything else?

This is my first time submitting a PR to the Paimon community, and there may be many things I haven't considered. Please feel free to point them out.

Forus0322:
[core]

Linked issue: close #6880


JingsongLi (commented):
Large PR (+1607 lines) adding compaction metrics system table. Review points:

1. **Scope**: This touches many core files (CompactTask, CompactResult, MergeTreeWriter, FileStoreCommitImpl, CommitMessageSerializer). The blast radius is significant - any bugs here could affect the critical write/compact path.

2. **Serialization compatibility**: Changes to `CommitMessageSerializer` must maintain backward compatibility. Old snapshots with commit messages serialized in the previous format must still be readable. Is there version handling?

3. **CompactMetricMeta / CompactMetricSerializer**: Adding metric metadata to the compact result flow means this data travels through the commit pipeline. Consider:
   - What's the storage overhead per compaction?
   - Is metric data persisted in manifests or a separate location?
   - How is metric data expired (the `ExpireCompactMetrics` classes suggest there's a retention policy)?

4. **System table design**: `CompactionMetricsTable` - what columns does it expose? Key metrics should include: compaction duration, input/output file count, input/output bytes, compaction type (minor/major).

5. **Configuration**: The new `CoreOptions` entry - what's the default? Metrics collection should probably be opt-in to avoid overhead for users who don't need it.

6. **Testing**: I notice the test changes are in `AppendDeletionFileMaintainerTest` - is there a dedicated test for the compaction metrics table itself? This feature deserves its own comprehensive test.

The feature is valuable for observability but the implementation scope is concerning. Consider whether metrics can be collected with less coupling to the core write path.

PR #6982, opened 2026-01-08
[python/hotfix] Fix zstd manifest decompression error on Python 3.6

XiaoHongbo-Hope:
### Purpose

Linked issue: close #xxx

### DIFF
--- a/paimon-python/pypaimon/write/writer/key_value_data_writer.py
+++ b/paimon-python/pypaimon/write/writer/key_value_data_writer.py
@@ -38,10 +38,8 @@ class KeyValueDataWriter(DataWriter):
         num_rows = data.num_rows
         enhanced_table = data
 
-        column_names = data.schema.names
-
         for pk_key in reversed(self.trimmed_primary_keys):
-            if pk_key in column_names:
+            if pk_key in data.column_names:
                 key_column = data.column(pk_key)
                 enhanced_table = enhanced_table.add_column(0, f'_KEY_{pk_key}', key_column)
 


XiaoHongbo-Hope:
> We should add java & python e2e for 3.6 too.

👌

XiaoHongbo-Hope:
> We should add java & python e2e for 3.6 too.

When I tried to enable e2e case for py36, some cases failed. Then I only add current case for py36 in this PR, will create another PR to fix other cases in py36

paimon-python/dev/run_mixed_tests.sh
```diff
@@ -192,13 +192,77 @@ run_pk_dv_test() {
         return 1
     fi
 }
+
+# Function to run zstd manifest e2e test
+run_zstd_manifest_test() {
```
JingsongLi: You don't need to create this, by default, manifest is compressed by zstd.

XiaoHongbo-Hope: > You don't need to create this, by default, manifest is compressed by zstd.

It seems that the default compression is lz4 in python, is it by design?

paimon-core/src/test/java/org/apache/paimon/JavaPyE2ETest.java
```diff
@@ -356,6 +357,7 @@ public void testPKDeletionVectorWriteMultiBatchRawConvertable() throws Exception
 
     @Test
     @EnabledIfSystemProperty(named = "run.e2e.tests", matches = "true")
+    @DisabledIfSystemProperty(named = "python.version", matches = "3.6")
```
JingsongLi, line 360: Why disable for 3.6?

paimon-lance/src/test/java/org/apache/paimon/JavaPyLanceE2ETest.java
```diff
@@ -260,19 +261,9 @@ public void testJavaWriteReadPkTableLance() throws Exception {
 
     @Test
     @EnabledIfSystemProperty(named = "run.e2e.tests", matches = "true")
+    @DisabledIfSystemProperty(named = "python.version", matches = "3.6")
```
JingsongLi, line 264: Why disable for 3.6?

XiaoHongbo-Hope: > Why disable for 3.6?

py36 does not have pylance dependency

paimon-core/src/test/java/org/apache/paimon/JavaPyE2ETest.java
XiaoHongbo-Hope: > Why disable for 3.6?

compatible issue in pk table for py36, will create another  PR to fix it.

PR #6986, opened 2026-01-08
[core] add data-evolution info to Files System Table.

steFaiz:
### Purpose

Add firstRowId, writeCols and IsBlob columns to Files System Table. 

### Tests
None

### API and Format
None

### Documentation
None

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/system/FilesTable.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/system/FilesTable.java
@@ -28,7 +28,6 @@ import org.apache.paimon.data.InternalArray;
 import org.apache.paimon.data.InternalRow;
 import org.apache.paimon.data.LazyGenericRow;
 import org.apache.paimon.disk.IOManager;
-import org.apache.paimon.format.blob.BlobFileFormat;
 import org.apache.paimon.fs.FileIO;
 import org.apache.paimon.io.DataFileMeta;
 import org.apache.paimon.io.DataFilePathFactory;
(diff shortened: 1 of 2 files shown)


paimon-core/src/main/java/org/apache/paimon/table/system/FilesTable.java
```diff
@@ -117,7 +119,10 @@ public class FilesTable implements ReadonlyTable {
                             new DataField(14, "max_sequence_number", new BigIntType(true)),
                             new DataField(15, "creation_time", DataTypes.TIMESTAMP_MILLIS()),
                             new DataField(16, "deleteRowCount", DataTypes.BIGINT()),
-                            new DataField(17, "file_source", DataTypes.STRING())));
+                            new DataField(17, "file_source", DataTypes.STRING()),
+                            new DataField(18, "first_row_id", DataTypes.BIGINT()),
+                            new DataField(19, "write_cols", DataTypes.ARRAY(DataTypes.STRING())),
+                            new DataField(20, "blob", DataTypes.BOOLEAN())));
```
JingsongLi: Remove this, just get info from file name?

steFaiz: Thanks for your advise! I was thinking that this is convenient for sql. For example, if we want to calculate all blob's storage, we just need to write `where T.blob = true` instead of `where T.fileName LIKE '%s.blob'`.
Would you be okay with keeping this boolean field, or do you prefer avoid persisting redundant metadata?

JingsongLi: This system table should just reflect the schema of data file meta, so I think it is better to remove this.

steFaiz: Thanks for your explaination! I've adapted the code.

JingsongLi:
+1

PR #6987, opened 2026-01-08
[python] Support read paimon table as pytorch dataset

discivigour:
### Purpose

Support read paimon table as pytorch dataset

### Tests
- torch_read_test.py

### API and Format

TableRead.to_torch: Convert Paimon table data to pytorch Dataset.

### Documentation

updated python-api.md

### DIFF
--- a/paimon-python/pypaimon/read/datasource.py
+++ b/paimon-python/pypaimon/read/datasource.py
@@ -260,10 +260,7 @@ class TorchDataset(Dataset):
         Returns:
             Total number of rows across all splits
         """
-        if self._data is not None:
-            return len(self._data)
-        else:
-            return 0
+        return len(self._data)
 
     def __getitem__(self, index: int):
         """


docs/content/program-api/python-api.md
```diff
@@ -473,6 +474,38 @@ ray_dataset = table_read.to_ray(splits)
...
+
+This requires `torch` to be installed.
+
+You can read all the data into a `torch.utils.data.Dataset` or `torch.utils.data.IterableDataset`:
+
+```python
+from torch.utils.data import DataLoader
+
+table_read = read_builder.new_read()
+dataset = table_read.to_torch(splits, streaming=True)
```
JingsongLi, line 487: Can we find a reference for this parameter?

discivigour: https://huggingface.co/docs/datasets/loading#webdataset

paimon-python/pypaimon/read/datasource.py
```diff
@@ -216,13 +219,146 @@ def _get_read_task(
...
+
+        Returns:
+            Dictionary containing the row data
+        """
+        if not self._data:
+            return None
+
+        return self._data[index]
+
+    def _load_data(self):
```
JingsongLi: inline this to `__init__ `

discivigour: 👌

PR #6988, opened 2026-01-08
[core] Introduces some transforms for data masking

plusplusjiajia:
### Purpose
Introduces four new Transform implementations for data masking:
1. HashTransform: Masks string columns by hashing with configurable algorithm (default SHA-256) and optional salt
2. PartialMaskTransform: Masks the middle part of string
3. NullTransform: Always returns null for the input field
4. DefaultValueTransform: Returns the default value based on the field's data type

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/predicate/DefaultValueTransform.java
+++ b/paimon-common/src/main/java/org/apache/paimon/predicate/DefaultValueTransform.java
@@ -1,0 +1,92 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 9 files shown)


plusplusjiajia:
> What these transforms refer from?
These transforms reference the implementations of ranger data filter and snowflake mask function.
https://github.com/apache/ranger/blob/master/agents-common/src/main/resources/service-defs/ranger-servicedef-trino.json#L269
https://docs.snowflake.com/en/user-guide/security-column-intro
https://docs.snowflake.com/en/sql-reference/functions/sha2
https://docs.snowflake.com/en/sql-reference/functions/regexp_replace

JingsongLi (commented):
Good feature for data masking transforms. Review comments:

1. **Package placement**: These transforms are in `org.apache.paimon.predicate` package but they're not predicates - they're data transforms/projections. Consider whether a separate `org.apache.paimon.transform` or `org.apache.paimon.masking` package would be more appropriate.

2. **HashTransform**:
   - Using `MessageDigest` is not thread-safe. The transient `digest` field with lazy init needs synchronization if the transform could be shared across threads, or document that instances are not thread-safe.
   - Good that you support configurable algorithm and salt. Consider whether the salt should be stored in plain text or if there are security implications.
   - The dependency on Guava's `HashCode.toString()` just for hex encoding seems heavy - could use a simple hex utility instead.

3. **PartialMaskTransform**: Masking the "middle part" - what defines middle? Is it configurable (start offset, end offset)? Edge cases to consider: strings shorter than mask boundaries, empty strings, null values.

4. **DefaultValueUtils**: Make sure all Paimon data types are covered (including nested types like MAP, ARRAY, ROW).

5. **Security consideration**: For HashTransform, SHA-256 without salt is vulnerable to rainbow table attacks. Consider making salt mandatory or at least strongly recommended in documentation.

6. **Test coverage**: Good that each transform has its own test class. Ensure edge cases are covered (null input, empty string, very long strings, special characters).

Overall a useful feature for data governance. The implementation looks solid but the package organization could be improved.

PR #6991, opened 2026-01-09
[vector] add faiss index and e2e test  to support vector index

jerry-024:
### Purpose

add faiss index and e2e test to support vector index

### Tests
- FaissVectorGlobalIndexScanTest
- FaissVectorGlobalIndexTest
- FaissVectorIndexE2ETest


JingsongLi:
+1 Thanks @jerry-024

PR #6994, opened 2026-01-09
[core] Enable global index external path

leaves12138:
### Purpose

Enable global index store in external path

### DIFF
--- a/paimon-core/src/test/java/org/apache/paimon/table/GlobalIndexTableTest.java
+++ b/paimon-core/src/test/java/org/apache/paimon/table/GlobalIndexTableTest.java
@@ -283,7 +283,7 @@ public class GlobalIndexTableTest extends DataEvolutionTestBase {
         ResultEntry result = results.get(0);
 
         String fileName = result.fileName();
-        long fileSize = fileIO.getFileSize(indexFileReadWrite.filePath(fileName));
+        long fileSize = indexFileReadWrite.fileSize(fileName);
         GlobalIndexMeta globalIndexMeta =
                 new GlobalIndexMeta(0, result.rowCount() - 1, indexField.id(), null, result.meta());
         return Collections.singletonList(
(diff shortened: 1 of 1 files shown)


paimon-common/src/main/java/org/apache/paimon/globalindex/GlobalIndexIOMeta.java
```diff
@@ -27,11 +29,18 @@ public class GlobalIndexIOMeta {
     private final String fileName;
     private final long fileSize;
     private final byte[] metadata;
+    @Nullable private final String externalPath;
```
lxy-9602, line 32: The `GlobalIndexIOMeta` is intended to be a meta interface for plugin authors. However, including an external path in it might be confusing for users, as it adds complexity and is not intuitive in this context.

Would it make more sense to store only the file_path in `GlobalIndexIOMeta`, while leaving the external path to be handled internally by the Paimon framework? This way, the external path could be extracted and managed within the `IndexFileMeta`, simplifying the API for plugin authors and ensuring a clearer separation of concerns.

paimon-core/src/main/java/org/apache/paimon/utils/FileStorePathFactory.java
```diff
@@ -130,6 +135,10 @@ public Path indexPath() {
         return new Path(root, INDEX_PATH);
     }
 
+    public Path globalIndexRootDir() {
+        return globalIndexExternalRootDir != null ? globalIndexExternalRootDir : indexPath();
+    }
```
lxy-9602, line 140: If `globalIndexExternalRootDir` is not null, the index path be `external_index_path/index/` or `external_index_path/`? The first one maybe better? Not sure whether retaining the `index/` is necessary.

paimon-common/src/main/java/org/apache/paimon/globalindex/GlobalIndexIOMeta.java
leaves12138: Accepted

paimon-core/src/main/java/org/apache/paimon/utils/FileStorePathFactory.java
leaves12138: Accepted, it could be like this

PR #6995, opened 2026-01-09
support global index in python

jerry-024:
### Purpose
support global index which is faiss vector index search in python

### Tests
- TestFaissVectorGlobalIndexE2E
- JavaPyFaissE2ETest


JingsongLi (approved):
+1

PR #7003, opened 2026-01-11
[core] support btree indexed-scan in paimon-core and add some tests

steFaiz:
### Purpose

This work is part of https://github.com/apache/paimon/issues/6834

This PR mainly contains two parts:
1. Introduce a `localRowId` method for `GlobalIndexer`, indicating whether the indexer stores original row ids (e.g. BTree index) or local row ids which are self-incremental from 0 (i.e. bitmap index and all vector indices). We should wrap an index reader by an  `OffsetIndexReader` only if the indexer has local row ids.
2. Split a GlobalIndexTest from DataEvolutionTableTest to avoid DataEvolutionTableTest growing too long.

### Tests
Please see `org.apache.paimon.table.GlobalIndexTableTest` for all global-index-related tests.

### API and Format
No public API modified.

### Documentation
No public doc need to be modified.


JingsongLi (changes requested):
I think we should build btree index with local row id.

JingsongLi:
Cool!

PR #7005, opened 2026-01-12
[python] fix cramjam version define and introduce dependency check in CI

XiaoHongbo-Hope:
### Purpose

https://github.com/apache/paimon/pull/6996 introduce a wrong dependency version for cramjam. But CI does not fail due to lack of requirements.txt validation check, so add a check in CI to avoid failure in user side when installing pypaimon


JingsongLi:
+1

PR #7007, opened 2026-01-12
[python] support rest token refresh 

XiaoHongbo-Hope:
### Purpose

Linked issue: close #xxx

### Tests
`rest_token_file_io_test.py`

### DIFF
--- a/paimon-python/pypaimon/api/rest_api.py
+++ b/paimon-python/pypaimon/api/rest_api.py
@@ -46,7 +46,7 @@ class RESTApi:
     PAGE_TOKEN = "pageToken"
     DATABASE_NAME_PATTERN = "databaseNamePattern"
     TABLE_NAME_PATTERN = "tableNamePattern"
-    TOKEN_EXPIRATION_SAFE_TIME_MILLIS = 60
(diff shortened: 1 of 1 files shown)


paimon-python/pypaimon/catalog/rest/rest_token_file_io.py
```diff
@@ -18,27 +18,38 @@
...
 from pypaimon.catalog.rest.rest_token import RESTToken
 from pypaimon.common.file_io import FileIO
 from pypaimon.common.identifier import Identifier
-from pypaimon.common.options import Options
-from pypaimon.common.options.config import CatalogOptions, OssOptions
+
+from cachetools import TTLCache
 
 
 class RESTTokenFileIO(FileIO):
```
JingsongLi, line 36: Maybe we should do a refactor? FileIO should be an interface.

XiaoHongbo-Hope: > Maybe we should do a refactor? FileIO should be an interface.

refactored in https://github.com/apache/paimon/pull/7040, and rebased master. This PR is now ready again.

PR #7009, opened 2026-01-12
[python] fix rest file io and blob-as-descriptor file io token merge issue and add rest catalog blob-as-descriptor sample

XiaoHongbo-Hope:
### Purpose

When using REST catalog with blob-as-descriptor, we have two FileIO instances:
1. Table FileIO (RESTTokenFileIO) - uses REST data token
2. External OSS blob files - uses user OSS credentials

RESTTokenFileIO updates catalog_options dynamically in _initialize_oss_fs() when refreshing token. This would pollute the blob reader's configuration. So try to use a copy of catalog options in uri_reader_factory.

Java: Token updates in RESTTokenFileIO.fileIO() create a new Options object.

### DIFF
--- a/paimon-python/pypaimon/common/file_io.py
+++ b/paimon-python/pypaimon/common/file_io.py
@@ -43,7 +43,7 @@ class FileIO:
         self.properties = catalog_options
         self.logger = logging.getLogger(__name__)
         scheme, netloc, _ = self.parse_location(path)
-        self.uri_reader_factory = UriReaderFactory(catalog_options.copy())
+        self.uri_reader_factory = UriReaderFactory(catalog_options)
         if scheme in {"oss"}:
             self.filesystem = self._initialize_oss_fs(path)
         elif scheme in {"s3", "s3a", "s3n"}:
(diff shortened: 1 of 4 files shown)


paimon-python/pypaimon/common/file_io.py
```diff
@@ -43,7 +43,7 @@ def __init__(self, path: str, catalog_options: Options):
         self.properties = catalog_options
         self.logger = logging.getLogger(__name__)
         scheme, netloc, _ = self.parse_location(path)
-        self.uri_reader_factory = UriReaderFactory(catalog_options)
+        self.uri_reader_factory = UriReaderFactory(catalog_options.copy())
```
JingsongLi: Why do we need copy? What logic would update it?

XiaoHongbo-Hope: > Why do we need copy? What logic would update it?

When using REST catalog with blob-as-descriptor, we have two FileIO instances:
1. Table FileIO (RESTTokenFileIO) - uses REST data token
2. External OSS blob files - uses user OSS credentials

RESTTokenFileIO updates catalog_options dynamically in _initialize_oss_fs() when refreshing token. This would pollute the blob reader's configuration. So try to use a copy of catalog options in uri_reader_factory.

Java: Token updates in RESTTokenFileIO.fileIO() create a new Options object.

Updated this info in PR description too.

JingsongLi: I think `RESTTokenFileIO` should not update `catalog_options`, should copy a new options, I think java also do like this?

PR #7018, opened 2026-01-13
[fix]fix faiss deploy

jerry-024:
### Purpose

fix faiss deploy jar only so file: libpaimon_faiss_jni.so and libgomp.so.1


JingsongLi:
+1

PR #7020, opened 2026-01-13
[core] Add core options blob.file-size-count-in-splitting

leaves12138:
### Purpose

Blob table generates too many splits when scan, cause blob file size counting in splitting. Add a parameter to control this.

### DIFF
--- a/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
@@ -2595,9 +2595,8 @@ public class CoreOptions implements Serializable {
     }
 
     public boolean blobSplitByFileSize() {
-        return options.contains(BLOB_SPLIT_BY_FILE_SIZE)
-                ? options.get(BLOB_SPLIT_BY_FILE_SIZE)
-                : !options.get(BLOB_AS_DESCRIPTOR);
+        return options.getOptional(BLOB_SPLIT_BY_FILE_SIZE)
+                .orElse(!options.get(BLOB_AS_DESCRIPTOR));
     }
 
     public long compactionFileSize(boolean hasPrimaryKey) {


paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -655,6 +655,16 @@ public InlineElement getDescription() {
                                             "Target size of a blob file. Default is value of TARGET_FILE_SIZE.")
                                     .build());
 
+    public static final ConfigOption<Boolean> BLOB_FILE_SIZE_COUNT_IN_SPLITTING =
+            key("blob.file-size-count-in-splitting")
```
JingsongLi: blob.split-by-file-size

paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -2584,6 +2594,12 @@ public long blobTargetFileSize() {
                 .orElse(targetFileSize(false));
     }
 
+    public boolean blobFileSizeCountInSplitting() {
```
JingsongLi: also update method name

leaves12138: OK

paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
leaves12138: OK

paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -2584,6 +2594,12 @@ public long blobTargetFileSize() {
                 .orElse(targetFileSize(false));
     }
 
+    public boolean blobSplitByFileSize() {
+        return options.contains(BLOB_SPLIT_BY_FILE_SIZE)
+                ? options.get(BLOB_SPLIT_BY_FILE_SIZE)
+                : options.get(BLOB_AS_DESCRIPTOR);
```
JingsongLi: reverse? you should add tests.

paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -2584,6 +2594,12 @@ public long blobTargetFileSize() {
                 .orElse(targetFileSize(false));
     }
 
+    public boolean blobSplitByFileSize() {
+        return options.contains(BLOB_SPLIT_BY_FILE_SIZE)
```
JingsongLi: options.getOptional().orElse()

paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
leaves12138: OK

paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
leaves12138: OK

PR #7022, opened 2026-01-13
[python] add error handling to abort pending commits  and serialization support for ray sink

XiaoHongbo-Hope:
### Purpose

Add error handling to abort pending commits on failures and  and write method to create WriteBuilder on each worker.

Changes:
Update write method: create WriteBuilder on worker instead of using driver's builder
Add abort logic in on_write_complete() and on_write_failed() methods
Add _pending_commit_messages tracking and serialization support

### Tests
`ray_sink_test`

### API and Format
pyhton-api.md


JingsongLi:
Please add description.

PR #7023, opened 2026-01-13
[flink] Verify paimon-flink's compatibility with Hadoop 3.x

yunfengzhou-hub:
### Purpose

In order to fix CVE vulnerabilites like [CVE-2025-48924](https://security.snyk.io/vuln/SNYK-JAVA-COMMONSLANG-10734077), this PR adds a profile and CI pipelines to verify paimon-flink's compatibility with Hadoop 3.x, so that if users want to use paimon-flink in a Hadoop 3.x project, they can compile the project with -Phadoop3 on their own and get the needed dependencies.

Given that Paimon still promises compatibility with Hadoop 2.x, Support for Hadoop 3.x is only added as an optional profile, instead of being set as the default dependency.

Currently only Flink 1.x has been verified against the newly introduced profile.

### Tests

* Add new CI workflows to verify the correctness of paimon-flink under the hadoop 3.x profile.

### API and Format

This PR does not change API or format.

### Documentation

It is documented in Flink quickstart how to build paimon-flink with hadoop 3.x.

### DIFF
--- a/.github/workflows/e2e-tests-flink-1.x.yml
+++ b/.github/workflows/e2e-tests-flink-1.x.yml
@@ -27,6 +27,8 @@ on:
       - 'paimon-python/**'
       - '.github/workflows/paimon-python-checks.yml'
       - 'paimon-lucene/**'
+  schedule:
+    - cron: '0 6 * * *'
 
 env:
   JDK_VERSION: 8
@@ -46,6 +48,11 @@ jobs:
         # Only Test Latest Version
         flink_version: [ '1.20' ]
         hadoop_profile: ['', 'hadoop3']
+        include:
+          # Run hadoop3 tests only on daily schedule
+          - hadoop_profile: 'hadoop3'
+            only_schedule: true
+    if: ${{ !matrix.only_schedule || github.event_name == 'schedule' }}
     steps:
       - name: Checkout code
         uses: actions/checkout@v6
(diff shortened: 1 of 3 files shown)


docs/content/flink/quick-start.md
```diff
@@ -70,7 +70,17 @@ You can also manually build bundled jar from the source code.
...
 
 Build bundled jar with the following command.
-- `mvn clean install -DskipTests`
+
+```bash
+# build paimon flink 1.x
+mvn clean install -DskipTests
+
+# build paimon flink 1.x with hadoop 3.x
+mvn clean package -DskipTests -Pflink1,hadoop3
```
JingsongLi, line 79: Why we need to build with -P hadoop3? What changed?

yunfengzhou-hub: Hadoop 2.x relies on commons-lang, which might be vulnerable to [CVE-2025-48924](https://security.snyk.io/vuln/SNYK-JAVA-COMMONSLANG-10734077). Some of our Paimon users have asked to provide a paimon-flink version with CVEs like have been fixed.

I'll update this to the description of this PR.

JingsongLi: Do you mean we will bundle the corresponding dependencies into our JAR file based on the Hadoop version?

yunfengzhou-hub: No, the official Paimon jars will not bundle hadoop 3.x dependencies. 

This PR is more like providing a guarantee that if there is another project that relies on paimon-flink and Hadoop 3.x at the same time, this project is not supposed to have Hadoop version compatibility issue.

PR #7027, opened 2026-01-13
[core][flink] Support parallelism snapshot expire

wzhero1:
### Purpose

This PR implements parallel snapshot expiration using Flink distributed execution to improve the performance of large-scale cleanup operations.

**Motivation:**
- Serial file deletion becomes a performance bottleneck for tables with large amounts of data
- Current implementation cannot leverage Flink's distributed computing capabilities

**Architecture:**

The expire process is split into two phases:
- **Worker phase (parallel flatMap)**: Deletes data files and changelog files in parallel using `RangePartitionedExpireFunction`. Tasks are partitioned by snapshot ID range to maximize cache locality (adjacent snapshots share manifest files and tag data file skippers).
- **Sink phase (serial)**: Deletes manifests and snapshot metadata serially using `SnapshotExpireSink`, because `manifestSkippingSet` is global mutable state and snapshot deletion must be contiguous.

**Key classes:**
- `ExpireSnapshotsPlanner`: Computes expiration plan including snapshot range, four types of tasks, protection set, and snapshot cache
- `SnapshotExpireTask`: Abstract base with 4 polymorphic subclasses (`DeleteDataFilesTask`, `DeleteChangelogFilesTask`, `DeleteManifestsTask`, `DeleteSnapshotTask`)
- `ExpireContext`: Holds runtime dependencies (SnapshotManager, SnapshotDeletion, ChangelogManager) and shared state (taggedSnapshots, snapshotCache, skippingSet)
- `ExpireSnapshotsPlan`: Contains task lists, protection set, snapshot cache, and range-based task partitioning logic
- `DeletionReport`: Carries deletion bucket info from workers to sink for empty directory cleanup

**Execution modes:**
- Local mode: `forceStartFlinkJob=false` → executes locally via `ExpireSnapshotsProcedure`
- Flink mode: `forceStartFlinkJob=true`→ uses Flink distributed execution

**Other changes:**
- Moved planner/plan/task/report classes from paimon-core to paimon-flink (per reviewer suggestion)
- Planner pre-collects snapshots in parallel using `CompletableFuture` and caches them, serialized to both workers and sink to avoid redundant reads
- `ExpireSnapshotsImpl` (serial mode) reuses `ExpireSnapshotsPlanner` for planning, keeping serial execution path unchanged
- Added failure recovery test: verifies that if expire job fails midway (data files partially deleted but snapshot metadata intact), a subsequent run completes successfully

### Tests

**Unit Tests:**
- `ExpireSnapshotsPlanTest` - Tests range-based task partitioning logic
- `DeletionReportTest` - Tests deletion report serialization

**Integration Tests:**
- `ParallelExpireSnapshotsActionITCase` - Extends `ExpireSnapshotsTest` to validate parallel mode produces same results as serial mode, plus failure recovery test
- `ExpireSnapshotsProcedureITCase` - Tests both local and parallel modes via procedure

### API and Format

**New CLI parameter for `expire-snapshots` action:**
| Parameter | Type | Default | Description |
|-----------|------|---------|-------------|
| `--parallelism` | Integer | env parallelism | Parallelism for parallel expire workers |

**No storage format changes.**

**Backward compatible:** Default behavior (local mode) remains unchanged.

### Documentation

Yes, this introduces a new feature: parallel snapshot expiration.

Documentation should cover:
- New `--parallelism` parameter for `expire-snapshots` action
- `--force_start_flink_job` triggers flink mode

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/ExpireSnapshotsImpl.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/ExpireSnapshotsImpl.java
@@ -91,64 +91,14 @@ public class ExpireSnapshotsImpl implements ExpireSnapshots {
     @Override
     public int expire() {
         snapshotDeletion.setChangelogDecoupled(expireConfig.isChangelogDecoupled());
-        int retainMax = expireConfig.getSnapshotRetainMax();
-        int retainMin = expireConfig.getSnapshotRetainMin();
-        int maxDeletes = expireConfig.getSnapshotMaxDeletes();
-        long olderThanMills =
-                System.currentTimeMillis() - expireConfig.getSnapshotTimeRetain().toMillis();
 
-        Long latestSnapshotId = snapshotManager.latestSnapshotId();
-        if (latestSnapshotId == null) {
-            // no snapshot, nothing to expire
+        ExpireRange range =
+                computeSnapshotExpireRange(expireConfig, snapshotManager, consumerManager);
+        if (range.isEmpty()) {
             return 0;
         }
 
-        Long earliest = snapshotManager.earliestSnapshotId();
-        if (earliest == null) {
-            return 0;
-        }
-
-        Preconditions.checkArgument(
-                retainMax >= retainMin,
-                String.format(
-                        "retainMax (%s) must not be less than retainMin (%s).",
-                        retainMax, retainMin));
-
-        // the min snapshot to retain from 'snapshot.num-retained.max'
-        // (the maximum number of snapshots to retain)
-        long min = Math.max(latestSnapshotId - retainMax + 1, earliest);
-
-        // the max exclusive snapshot to expire until
-        // protected by 'snapshot.num-retained.min'
-        // (the minimum number of completed snapshots to retain)
-        long maxExclusive = latestSnapshotId - retainMin + 1;
-
-        // the snapshot being read by the consumer cannot be deleted
-        if (!expireConfig.isConsumerChangelogOnly()) {
-            maxExclusive =
-                    Math.min(
-                            maxExclusive, consumerManager.minNextSnapshot().orElse(Long.MAX_VALUE));
-        }
-
-        // protected by 'snapshot.expire.limit'
-        // (the maximum number of snapshots allowed to expire at a time)
-        maxExclusive = Math.min(maxExclusive, earliest + maxDeletes);
-
-        for (long id = min; id < maxExclusive; id++) {
-            // Early exit the loop for 'snapshot.time-retained'
-            // (the maximum time of snapshots to retain)
-            try {
-                Snapshot snapshot = snapshotManager.tryGetSnapshot(id);
-                if (olderThanMills <= snapshot.timeMillis()) {
-                    return expireUntil(earliest, id);
-                }
-            } catch (FileNotFoundException e) {
-                // ignore
-                // snapshot may have been deleted by another process
-            }
-        }
-
-        return expireUntil(earliest, maxExclusive);
+        return expireUntil(range.earliestId, range.maxExclusive);
     }
 
     @VisibleForTesting
@@ -321,6 +271,95 @@ public class ExpireSnapshotsImpl implements ExpireSnapshots {
     @VisibleForTesting
     public SnapshotDeletion snapshotDeletion() {
         return snapshotDeletion;
+    }
+
+    /** Range boundaries for snapshot expiration. */
+    public static class ExpireRange {
+        private static final ExpireRange EMPTY = new ExpireRange(0, 0);
+
+        public final long earliestId;
+        public final long maxExclusive;
+
+        public ExpireRange(long earliestId, long maxExclusive) {
+            this.earliestId = earliestId;
+            this.maxExclusive = maxExclusive;
+        }
+
+        public static ExpireRange empty() {
+            return EMPTY;
+        }
+
+        public boolean isEmpty() {
+            return maxExclusive <= earliestId;
+        }
+    }
+
+    /** Compute the snapshot expire range, or {@link ExpireRange#empty()} if nothing to expire. */
+    public static ExpireRange computeSnapshotExpireRange(
+            ExpireConfig expireConfig,
+            SnapshotManager snapshotManager,
+            ConsumerManager consumerManager) {
+        int retainMax = expireConfig.getSnapshotRetainMax();
+        int retainMin = expireConfig.getSnapshotRetainMin();
+        int maxDeletes = expireConfig.getSnapshotMaxDeletes();
+        long olderThanMills =
+                System.currentTimeMillis() - expireConfig.getSnapshotTimeRetain().toMillis();
+
+        Long latestSnapshotId = snapshotManager.latestSnapshotId();
+
+        if (latestSnapshotId == null) {
+            // no snapshot, nothing to expire
+            return ExpireRange.empty();
+        }
+
+        Long earliest = snapshotManager.earliestSnapshotId();
+
+        if (earliest == null) {
+            return ExpireRange.empty();
+        }
+
+        Preconditions.checkArgument(
+                retainMax >= retainMin,
+                String.format(
+                        "retainMax (%s) must not be less than retainMin (%s).",
+                        retainMax, retainMin));
+
+        // the min snapshot to retain from 'snapshot.num-retained.max'
+        // (the maximum number of snapshots to retain)
+        long min = Math.max(latestSnapshotId - retainMax + 1, earliest);
+
+        // the max exclusive snapshot to expire until
+        // protected by 'snapshot.num-retained.min'
+        // (the minimum number of completed snapshots to retain)
+        long maxExclusive = latestSnapshotId - retainMin + 1;
+
+        // the snapshot being read by the consumer cannot be deleted
+        if (!expireConfig.isConsumerChangelogOnly()) {
+            maxExclusive =
+                    Math.min(
+                            maxExclusive, consumerManager.minNextSnapshot().orElse(Long.MAX_VALUE));
+        }
+
+        // protected by 'snapshot.expire.limit'
+        // (the maximum number of snapshots allowed to expire at a time)
+        maxExclusive = Math.min(maxExclusive, earliest + maxDeletes);
+
+        for (long id = min; id < maxExclusive; id++) {
+            // Early exit the loop for 'snapshot.time-retained'
+            // (the maximum time of snapshots to retain)
+            try {
+                Snapshot snapshot = snapshotManager.tryGetSnapshot(id);
+                if (olderThanMills <= snapshot.timeMillis()) {
+                    maxExclusive = id;
+                    break;
+                }
+            } catch (FileNotFoundException e) {
+                // ignore
+                // snapshot may have been deleted by another process
+            }
+        }
+
+        return new ExpireRange(earliest, maxExclusive);
     }
 
     /** Find the skipping tags in sortedTags for range of [beginInclusive, endExclusive). */
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/ExpireSnapshotsAction.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/ExpireSnapshotsAction.java
@@ -32,6 +32,8 @@ import org.apache.paimon.options.ExpireConfig;
 import org.apache.paimon.table.FileStoreTable;
 import org.apache.paimon.utils.ProcedureUtils;
 
+import org.apache.flink.api.common.typeinfo.TypeHint;
+import org.apache.flink.api.common.typeinfo.TypeInformation;
 import org.apache.flink.streaming.api.datastream.DataStream;
 import org.apache.flink.streaming.api.datastream.DataStreamSource;
 import org.slf4j.Logger;
@@ -49,10 +51,10 @@ import java.util.Map;
  * <p>This action supports both local and parallel execution modes:
  *
  * <ul>
- *   <li>Local mode: when {@code forceStartFlinkJob} is false and {@code parallelism <= 1}, executes
- *       locally via {@link #executeLocally()}
- *   <li>Parallel mode: when {@code forceStartFlinkJob} is true or {@code parallelism > 1}, uses
- *       Flink distributed execution for deletion
+ *   <li>Local mode: when {@code forceStartFlinkJob} is false, executes locally via {@link
+ *       #executeLocally()}
+ *   <li>Flink mode: when {@code forceStartFlinkJob} is true, uses Flink distributed execution for
+ *       deletion.
  * </ul>
  *
  * <p>In parallel mode:
@@ -108,12 +110,10 @@ public class ExpireSnapshotsAction extends ActionBase implements LocalAction {
 
     @Override
     public void run() throws Exception {
-        if (forceStartFlinkJob || parallelism > 1) {
-            // Parallel mode: build custom multi-parallelism Flink pipeline
+        if (forceStartFlinkJob) {
+            // Flink mode: always build and submit a Flink job
             build();
-            if (!env.getTransformations().isEmpty()) {
-                execute(this.getClass().getSimpleName());
-            }
+            execute(this.getClass().getSimpleName());
         } else {
             // Default: ActionBase handles LocalAction → executeLocally()
             super.run();
@@ -151,15 +151,14 @@ public class ExpireSnapshotsAction extends ActionBase implements LocalAction {
         // Plan the expiration
         ExpireSnapshotsPlan plan = planner.plan(expireConfig);
         if (plan.isEmpty()) {
-            LOG.info("No snapshots to expire, skipping Flink job submission.");
-            return;
+            LOG.info("No snapshots to expire.");
+        } else {
+            LOG.info(
+                    "Planning to expire {} snapshots, range=[{}, {})",
+                    plan.endExclusiveId() - plan.beginInclusiveId(),
+                    plan.beginInclusiveId(),
+                    plan.endExclusiveId());
         }
-
-        LOG.info(
-                "Planning to expire {} snapshots, range=[{}, {})",
-                plan.endExclusiveId() - plan.beginInclusiveId(),
-                plan.beginInclusiveId(),
-                plan.endExclusiveId());
 
         // Build worker phase
         DataStream<DeletionReport> reports = buildWorkerPhase(plan, identifier, expireConfig);
@@ -188,7 +187,10 @@ public class ExpireSnapshotsAction extends ActionBase implements LocalAction {
                 plan.partitionTasksBySnapshotRange(parallelism);
 
         DataStreamSource<List<SnapshotExpireTask>> source =
-                env.fromCollection(partitionedGroups).setParallelism(1);
+                env.fromCollection(
+                                partitionedGroups,
+                                TypeInformation.of(new TypeHint<List<SnapshotExpireTask>>() {}))
+                        .setParallelism(1);
 
         return source.rebalance()
                 .flatMap(
@@ -200,7 +202,6 @@ public class ExpireSnapshotsAction extends ActionBase implements LocalAction {
                                 expireConfig.isChangelogDecoupled()))
                 // Use JavaTypeInfo to ensure proper Java serialization of DeletionReport,
                 // avoiding Kryo's FieldSerializer which cannot handle BinaryRow correctly.
-                // This approach is compatible with both Flink 1.x and 2.x.
                 .returns(new JavaTypeInfo<>(DeletionReport.class))
                 .setParallelism(parallelism)
                 .name("RangePartitionedExpire");
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/ExpireSnapshotsAction.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/ExpireSnapshotsAction.java
@@ -18,30 +18,75 @@
 
 package org.apache.paimon.flink.action;
 
+import org.apache.paimon.CoreOptions;
+import org.apache.paimon.catalog.Identifier;
+import org.apache.paimon.flink.expire.DeletionReport;
+import org.apache.paimon.flink.expire.ExpireSnapshotsPlan;
+import org.apache.paimon.flink.expire.ExpireSnapshotsPlanner;
+import org.apache.paimon.flink.expire.RangePartitionedExpireFunction;
+import org.apache.paimon.flink.expire.SnapshotExpireSink;
+import org.apache.paimon.flink.expire.SnapshotExpireTask;
 import org.apache.paimon.flink.procedure.ExpireSnapshotsProcedure;
+import org.apache.paimon.flink.utils.JavaTypeInfo;
+import org.apache.paimon.options.ExpireConfig;
+import org.apache.paimon.table.FileStoreTable;
+import org.apache.paimon.utils.ProcedureUtils;
 
+import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.streaming.api.datastream.DataStreamSource;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nullable;
+
+import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
-/** Expire snapshots action for Flink. */
+/**
+ * Expire snapshots action for Flink.
+ *
+ * <p>This action supports both local and parallel execution modes:
+ *
+ * <ul>
+ *   <li>Local mode: when {@code forceStartFlinkJob} is false and {@code parallelism <= 1}, executes
+ *       locally via {@link #executeLocally()}
+ *   <li>Parallel mode: when {@code forceStartFlinkJob} is true or {@code parallelism > 1}, uses
+ *       Flink distributed execution for deletion
+ * </ul>
+ *
+ * <p>In parallel mode:
+ *
+ * <ul>
+ *   <li>Worker phase (flatMap): deletes data files/changelog files in parallel
+ *   <li>Sink phase (commit): deletes manifests and snapshot metadata serially to avoid concurrent
+ *       deletion issues
+ * </ul>
+ */
 public class ExpireSnapshotsAction extends ActionBase implements LocalAction {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ExpireSnapshotsAction.class);
 
     private final String database;
     private final String table;
+
     private final Integer retainMax;
     private final Integer retainMin;
     private final String olderThan;
     private final Integer maxDeletes;
     private final String options;
+    private final int parallelism;
 
     public ExpireSnapshotsAction(
             String database,
             String table,
             Map<String, String> catalogConfig,
-            Integer retainMax,
-            Integer retainMin,
-            String olderThan,
-            Integer maxDeletes,
-            String options) {
+            @Nullable Integer retainMax,
+            @Nullable Integer retainMin,
+            @Nullable String olderThan,
+            @Nullable Integer maxDeletes,
+            @Nullable String options,
+            @Nullable Integer parallelism) {
         super(catalogConfig);
         this.database = database;
         this.table = table;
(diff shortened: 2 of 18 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/ExpireSnapshotsAction.java
```diff
@@ -50,12 +95,168 @@ public ExpireSnapshotsAction(
...
+        this.parallelism = parallelism;
     }
 
+    /** Returns true if forceStartFlinkJob is enabled and parallelism is greater than 1. */
+    private boolean isParallelMode() {
+        return forceStartFlinkJob && parallelism != null && parallelism > 1;
+    }
+
+    @Override
+    public void run() throws Exception {
```
yunfengzhou-hub, line 112: In case this method is overridden, this class needs not implement `LocalAction` anymore. Still you can keep the `executeLocally` method for internal use.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/ExpireSnapshotsAction.java
```diff
@@ -50,12 +95,168 @@ public ExpireSnapshotsAction(
         this.olderThan = olderThan;
         this.maxDeletes = maxDeletes;
         this.options = options;
+        this.parallelism = parallelism;
     }
 
+    /** Returns true if forceStartFlinkJob is enabled and parallelism is greater than 1. */
+    private boolean isParallelMode() {
+        return forceStartFlinkJob && parallelism != null && parallelism > 1;
```
yunfengzhou-hub: Forcing users to set the parallelism of this action might increase their burden to understand the logic of Paimon. It might be better to derive the parallelism of the job automatically on its own by default.

I noticed that `parallelism` is mainly used to distribute `SnapshotExpireTask`s evenly between the `RangePartitionedExpireFunction` subtasks, so this PR created a `List<List<SnapshotExpireTask>>` with `parallelism` batches. Compared with this design, a better way for Flink might be to have the subtasks directly fetch from the original list, continuing to consume the next task after completing the former one. On the one hand, this would make it unnecessary to decide the parallelism of the job from the beginning. On the other hand, this could also achieve better dynamic rebalancing during runtime in case different `SnapshotExpireTask`s have different workloads.

Besides, the Flink configuration might be a better place to set the job's parallelism, compared with Action arguments.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/ExpireSnapshotsAction.java
```diff
@@ -50,12 +95,168 @@ public ExpireSnapshotsAction(
...
+    public void build() throws Exception {
+        if (!isParallelMode()) {
+            // Not in parallel mode, nothing to build
+            return;
+        }
+
+        // Prepare table and config using shared method
+        Pair<FileStoreTable, ExpireConfig> prepared =
+                resolveExpireTableAndConfig(
+                        catalog.getTable(Identifier.fromString(database + "." + table)),
```
yunfengzhou-hub: In line 168 there is

```java
Identifier identifier = new Identifier(database, table);
```

We can move this line to above to reuse the created identifier variable here.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/ExpireSnapshotsAction.java
```diff
@@ -50,12 +95,168 @@ public ExpireSnapshotsAction(
...
+    @Override
+    public void build() throws Exception {
+        if (!isParallelMode()) {
+            // Not in parallel mode, nothing to build
+            return;
+        }
+
+        // Prepare table and config using shared method
+        Pair<FileStoreTable, ExpireConfig> prepared =
+                resolveExpireTableAndConfig(
```
yunfengzhou-hub: The introduction of method `resolveExpireTableAndConfig` seems to increase the code complexity a little bit. It might be better to directly put its implementation here.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/expire/ExpireSnapshotsPlan.java
```diff
@@ -0,0 +1,218 @@
...
+     * <ul>
+     *   <li>Worker 0 (snapshot 1-4): [dataTask(2,3,4), changelogTask(1,2,3,4)]
+     *   <li>Worker 1 (snapshot 5-8): [dataTask(5,6,7,8), changelogTask(5,6,7,8)]
+     *   <li>Worker 2 (snapshot 9-11): [dataTask(9,10,11), changelogTask(9,10)]
+     * </ul>
+     *
+     * @param parallelism target parallelism for distribution
+     * @return list of task groups, one per worker
+     */
+    public List<List<SnapshotExpireTask>> partitionTasksBySnapshotRange(int parallelism) {
```
yunfengzhou-hub, line 130: It seems that this PR divides the expire process of a snapshot into different `SnapshotExpireTask`s first, and then sends the tasks of the same snapshot to the same Flink subtask. If this is the case, why should this PR make the division first? Would the following implementation simpler?

```java
// calculates the start and end snapshot id before this part of code

env.fromSequence(startId, endId)
    .flatMap(new SnapshotExpireFunction()) // deletes the data, manifest, and snapshot file sequentially
    .sinkTo(new SnapshotExpireSink());
```

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/expire/SnapshotExpireSink.java
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    /**
+     * Do not annotate with <code>@override</code> here to maintain compatibility with Flink 2.0+.
+     */
+    public SinkWriter<DeletionReport> createWriter(InitContext context) throws IOException {
+        return new ExpireSinkWriter(initExecutor());
+    }
+
+    @Override
```
yunfengzhou-hub: Do we need the following javadoc here?
```java
/**
 * Do not annotate with <code>@override</code> here to maintain compatibility with Flink 1.18-.
 */
```

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/expire/SnapshotExpireSink.java
```diff
@@ -0,0 +1,175 @@
...
+                return;
+            }
+
+            LOG.info(
+                    "Expire sink received: {} manifest tasks, {} snapshot tasks",
+                    manifestTasks.size(),
+                    snapshotFileTasks.size());
+
+            // 1. Clean empty directories
+            executor.cleanEmptyDirectories(globalDeletionBuckets);
```
yunfengzhou-hub: The implementation here should be aware of the configuration `snapshot.clean-empty-directories`.

yunfengzhou-hub (commented):

paimon-core/src/main/java/org/apache/paimon/operation/expire/DeletionReport.java
```diff
@@ -0,0 +1,126 @@
...
+    public long snapshotId() {
+        return snapshotId;
+    }
+
+    public boolean isSkipped() {
+        return skipped;
+    }
+
+    public void setDataFilesDeleted(boolean dataFilesDeleted) {
+        this.dataFilesDeleted = dataFilesDeleted;
```
yunfengzhou-hub: Seems that these fields are only updated, but unused except in `toString` method.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/ExpireSnapshotsAction.java
wzhero1: done

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/ExpireSnapshotsAction.java
wzhero1: done

paimon-core/src/main/java/org/apache/paimon/operation/expire/DeletionReport.java
wzhero1: Removed the unused fields (dataFilesDeleted, changelogDeleted, manifestsDeleted, snapshotDeleted) and their setters from DeletionReport. Also cleaned up the corresponding setter calls in
  ExpireSnapshotsExecutor and updated the test.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/expire/SnapshotExpireSink.java
wzhero1: done

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/expire/SnapshotExpireSink.java
wzhero1: This is already handled. The snapshotDeletion is initialized with options.cleanEmptyDirectories(), so it respects the configuration automatically when cleanEmptyDirectories() is called.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/ExpireSnapshotsAction.java
wzhero1: Thanks for the suggestions! I've made --parallelism optional — when not specified, it falls back to env.getParallelism(), so users can control it via Flink configuration naturally. (Automatic parallelism inference based on workload is out of scope for this PR.)

 Regarding the dynamic task fetching approach (env.fromSequence(startId, endId).flatMap(...)), the current design intentionally separates deletion into two phases:

  - Worker phase (parallel): deletes data files and changelog files
  - Sink phase (serial): deletes manifests and snapshot metadata files

  This separation is necessary because:

1. manifestSkippingSet is global and mutable: Built from all tags + endSnapshot manifests. Parallel manifest deletion would require distributed coordination to safely update this set — significantly more complex.
2. Serial deletion ensures consistency: Parallel snapshot deletion could leave gaps (e.g., snapshots 1, 3, 5 exist but 2, 4 are deleted), which is confusing for users. Serial execution in the sink ensures clean, contiguous
  deletion.
3. Range partitioning maximizes cache locality: createDataFileSkipperForTags caches tag data files internally. Adjacent snapshots typically share the same nearest tag, so processing contiguous ranges avoids expensive cache rebuilds. Dynamic task fetching (each subtask fetching the next available split) would scatter snapshots across subtasks and break this cache benefit.

Data file deletion is the most time-consuming part and is safely parallelizable. Manifest/snapshot deletion is lightweight but requires ordering guarantees, so serial execution in the sink is the right trade-off.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/expire/ExpireSnapshotsPlan.java
wzhero1: The simplified approach (env.fromSequence + flatMap) has two main issues:
                                                                                                                                                                                                                                        
  1. Gap problem: Parallel deletion could leave gaps in snapshot history (e.g., 1, 3, 5 exist but 2, 4 are deleted), which is confusing for users.
  2. Cache locality: Adjacent snapshots typically share the same nearest tag. Range partitioning allows each subtask to reuse createDataFileSkipperForTags cache, while dynamic fetching would scatter snapshots and break this benefit.

  See my response in comment#2 for more details.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/ExpireSnapshotsAction.java
wzhero1: Prefer to keep LocalAction here. By default (without forceStartFlinkJob), the action runs locally via `super.run() → executeLocally()`, which is sufficient for normal scenarios and safer. Only when              
  forceStartFlinkJob is set, we start a Flink job with multi-parallelism for large-scale expiration. This also leaves room for future adjustments.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/ExpireSnapshotsAction.java
```diff
@@ -50,12 +93,135 @@ public ExpireSnapshotsAction(
...
+                        .build();
+
+        // Create planner using factory method
+        ExpireSnapshotsPlanner planner = ExpireSnapshotsPlanner.create(fileStoreTable);
+
+        // Plan the expiration
+        ExpireSnapshotsPlan plan = planner.plan(expireConfig);
+        if (plan.isEmpty()) {
+            LOG.info("No snapshots to expire");
+            return;
```
yunfengzhou-hub: Better to configure an empty Flink job, even if the job would complete immediately after submission. This is what "force_start_flink_job" means.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/ExpireSnapshotsAction.java
```diff
@@ -50,12 +93,135 @@ public ExpireSnapshotsAction(
...
+                        new SnapshotExpireSink(
+                                catalogOptions.toMap(),
+                                identifier,
+                                plan.endExclusiveId(),
+                                plan.protectionSet().manifestSkippingSet(),
+                                plan.manifestTasks(),
+                                plan.snapshotFileTasks(),
+                                expireConfig.isChangelogDecoupled()))
+                .setParallelism(1)
+                .name("SnapshotExpire");
```
yunfengzhou-hub: This operator name can be improved. The upper-stream operator is also doing snapshot-expire jobs.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/expire/RangePartitionedExpireFunction.java
```diff
@@ -0,0 +1,115 @@
...
+    public void flatMap(List<SnapshotExpireTask> tasks, Collector<DeletionReport> out)
+            throws Exception {
+        // Process tasks sequentially in order to maximize cache locality
+        for (SnapshotExpireTask task : tasks) {
+            DeletionReport report = processTask(task);
+            out.collect(report);
+        }
+    }
+
+    private DeletionReport processTask(SnapshotExpireTask task) {
```
yunfengzhou-hub: It's a little over-designed here, introducing a method for two lines of code that is only used once. It can be directly put into the flatMap method.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/expire/RangePartitionedExpireFunction.java
```diff
@@ -0,0 +1,115 @@
...
+        SnapshotDeletion deletion = table.store().newSnapshotDeletion();
+        deletion.setChangelogDecoupled(changelogDecoupled);
+        return new ExpireSnapshotsExecutor(table.snapshotManager(), deletion);
+    }
+
+    @Override
+    public void flatMap(List<SnapshotExpireTask> tasks, Collector<DeletionReport> out)
+            throws Exception {
+        // Process tasks sequentially in order to maximize cache locality
+        for (SnapshotExpireTask task : tasks) {
```
yunfengzhou-hub: Better add some logs to allow users to track the current expiration progress.

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/action/ExpireSnapshotsActionITCase.java
```diff
@@ -0,0 +1,231 @@
...
+/**
+ * IT Case for parallel expire snapshots using Flink execution. This class extends {@link
+ * ExpireSnapshotsTest} and overrides {@link #doExpire} to test parallel execution mode with real
+ * Flink DataStream API.
+ *
+ * <p>This test validates that the Flink-based parallel expire logic produces the same results as
+ * the serial implementation. It extends the production {@link RangePartitionedExpireFunction} and
+ * {@link SnapshotExpireSink} classes, overriding {@code initExecutor} to inject test executors.
+ */
+public class ExpireSnapshotsActionITCase extends ExpireSnapshotsTest {
```
yunfengzhou-hub: Can we add a test case that verifies, if the snapshot expiration job failed accidentally in the middle, it can still complete the expiration job after recovery?

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/action/ParallelExpireSnapshotsActionITCase.java
```diff
@@ -0,0 +1,231 @@
...
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Supplier;
+
+/**
+ * IT Case for parallel expire snapshots using Flink execution. This class extends {@link
```
yunfengzhou-hub, line 64: Given that this ITCase is specifically designed to test parallel mode, it is better to modify the name of this test class to reflect this point.

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/action/ExpireSnapshotsActionITCase.java
```diff
@@ -0,0 +1,231 @@
...
+    public void testExpireExtraFiles() {}
+
+    @Test
+    @Disabled("Tests SnapshotDeletion directly, not the full expire flow")
+    @Override
+    public void testExpireExtraFilesWithExternalPath() {}
+
+    @Test
+    @Disabled(
+            "Tests concurrent expire scenario which has different semantics in Flink parallel mode")
```
yunfengzhou-hub: Could you please elaborate how the two modes are different, and whether this difference would bring an influence on the expected expiration result for users?

paimon-core/src/main/java/org/apache/paimon/operation/expire/ExpireSnapshotsExecutor.java
```diff
@@ -0,0 +1,239 @@
...
+
+        switch (task.taskType()) {
+            case DELETE_DATA_FILES:
+                return executeDeleteDataFiles(task, snapshot, taggedSnapshots);
+            case DELETE_CHANGELOG_FILES:
+                return executeDeleteChangelogFiles(task, snapshot);
+            case DELETE_MANIFESTS:
+                return executeDeleteManifests(task, snapshot, skippingSet);
+            case DELETE_SNAPSHOT:
+                return executeDeleteSnapshot(task, snapshot);
```
yunfengzhou-hub: It might be better to convert 4 `TaskType`s to 4 subclasses of `SnapshotExpireTask`. The 4 switch branches here can be converted to a method in the `SnapshotExpireTask` interface/abstract class, and provide different implementations for this method in the 4 subclasses.

paimon-core/src/main/java/org/apache/paimon/table/ExpireSnapshotsImpl.java
```diff
@@ -18,49 +18,43 @@
...
+ * execution follows four phases:
+ *
+ * <ul>
+ *   <li>Phase 1a: Delete all data files (using dataFileTasks)
+ *   <li>Phase 1b: Delete all changelog files (using changelogFileTasks)
+ *   <li>Phase 2a: Delete all manifests (using manifestTasks)
+ *   <li>Phase 2b: Delete all snapshot files (using snapshotFileTasks)
+ * </ul>
+ */
 public class ExpireSnapshotsImpl implements ExpireSnapshots {
```
yunfengzhou-hub, line 54: It might be better to avoid doing a complete re-structure to ExpireSnapshotsImpl if it has been working well. Instead, we can have most of the logic in paimon-flink.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/ExpireSnapshotsAction.java
wzhero1: Prefer to keep the early return here. Submitting an empty Flink job still involves cluster resource allocation and scheduling overhead for zero useful work. I've improved the log message to make the skip behavior clear to users.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/ExpireSnapshotsAction.java
wzhero1: Renamed to "SnapshotExpireCommit" to distinguish from the upstream worker phase.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/expire/RangePartitionedExpireFunction.java
wzhero1: Done. Inlined processTask into flatMap.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/expire/RangePartitionedExpireFunction.java
wzhero1: Done. Added progress logging: batch-level start/end at INFO with elapsed time, per-task detail with task type, snapshot ID and progress counter.

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/action/ExpireSnapshotsActionITCase.java
wzhero1: Done. Added testExpireRecoveryAfterPartialFailure: uses a failAfterTasks flag in TestExpireFunction to simulate a mid-job crash during the worker phase (some data files deleted, but snapshot metadata intact). Then runs expire again and verifies it completes successfully despite missing files.

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/action/ParallelExpireSnapshotsActionITCase.java
wzhero1: Done. Renamed ExpireSnapshotsActionITCase to ParallelExpireSnapshotsActionITCase to clarify that this test class is specifically for parallel mode.

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/action/ExpireSnapshotsActionITCase.java
wzhero1: done

paimon-core/src/main/java/org/apache/paimon/operation/expire/ExpireSnapshotsExecutor.java
wzhero1: Refactored as suggested. Replaced the ExpireSnapshotsExecutor switch-based dispatch with polymorphic SnapshotExpireTask subclasses (DeleteDataFilesTask, DeleteChangelogFilesTask,
  DeleteManifestsTask, DeleteSnapshotTask). Each subclass encapsulates its own deletion logic via execute(ExpireContext). Also simplified ExpireContext by removing delegate methods — callers now access underlying objects directly (e.g., context.snapshotDeletion().cleanUnusedDataFiles()). taggedSnapshots and snapshotCache are now constructor parameters of ExpireContext instead of setters.

paimon-core/src/main/java/org/apache/paimon/table/ExpireSnapshotsImpl.java
wzhero1: done, remove the complete re-structure to ExpireSnapshotsImpl.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/ExpireSnapshotsAction.java
yunfengzhou-hub: In the current implementation, the DAG of the Flink job would change, if the user configures its parallelism from 1 to 2. This would bring additional complexity to the Job's operation process.

Given the problem above, I prefer to use the same DAG for the job no matter what parallelism the job has.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/ExpireSnapshotsAction.java
```diff
@@ -50,12 +95,139 @@ public ExpireSnapshotsAction(
...
+                .flatMap(
+                        new RangePartitionedExpireFunction(
+                                catalogOptions.toMap(),
+                                identifier,
+                                plan.protectionSet().taggedSnapshots(),
+                                plan.snapshotCache(),
+                                expireConfig.isChangelogDecoupled()))
+                // Use JavaTypeInfo to ensure proper Java serialization of DeletionReport,
+                // avoiding Kryo's FieldSerializer which cannot handle BinaryRow correctly.
+                // This approach is compatible with both Flink 1.x and 2.x.
```
yunfengzhou-hub: This comment is a little redundant. Most Flink APIs used by Paimon are compatible with both Flink 1.x and 2.x, and there is no need to amplify this point.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/expire/DeletionReport.java
```diff
@@ -0,0 +1,86 @@
...
+     * @param snapshotId the snapshot ID
+     * @return a skipped deletion report
+     */
+    public static DeletionReport skipped(long snapshotId) {
+        DeletionReport report = new DeletionReport(snapshotId);
+        report.skipped = true;
+        return report;
+    }
+
+    public long snapshotId() {
```
yunfengzhou-hub, line 61: This field is only used in tests

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/expire/ExpireSnapshotsPlan.java
```diff
@@ -0,0 +1,230 @@
...
+
+    public boolean isEmpty() {
+        return dataFileTasks.isEmpty()
+                && changelogFileTasks.isEmpty()
+                && manifestTasks.isEmpty()
+                && snapshotFileTasks.isEmpty();
+    }
+
+    /** Get data file deletion tasks (Phase 1a). These can be executed in parallel. */
+    public List<SnapshotExpireTask> dataFileTasks() {
```
yunfengzhou-hub: unused method.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/expire/ExpireSnapshotsPlan.java
```diff
@@ -0,0 +1,230 @@
...
+        return dataFileTasks;
+    }
+
+    /** Get changelog file deletion tasks (Phase 1b). These can be executed in parallel. */
+    public List<SnapshotExpireTask> changelogFileTasks() {
```
yunfengzhou-hub: unused method.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/expire/ExpireSnapshotsPlanner.java
```diff
@@ -0,0 +1,326 @@
...
+        long min = Math.max(latestSnapshotId - retainMax + 1, earliestId);
+
+        // the max exclusive snapshot to expire until
+        // protected by 'snapshot.num-retained.min'
+        // (the minimum number of completed snapshots to retain)
+        long endExclusiveId = latestSnapshotId - retainMin + 1;
+
+        // the snapshot being read by the consumer cannot be deleted
+        long consumerProtection = consumerManager.minNextSnapshot().orElse(Long.MAX_VALUE);
+        endExclusiveId = Math.min(endExclusiveId, consumerProtection);
```
yunfengzhou-hub: This logic to compute begin and end snapshot id is also repeated in `ExpireSnapshotsImpl` and `ExpireChangelogImpl`. It might be better to extract a common method for this code part, like whtat this PR has done for `FileDeletionBase#cleanEmptyDirectories`.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/action/ExpireSnapshotsAction.java
wzhero1: Done. Removed the parallelism > 1 condition. Now only forceStartFlinkJob controls whether to use Flink mode. The DAG stays the same regardless of parallelism setting.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/expire/DeletionReport.java
wzhero1: Added @VisibleForTesting annotation to snapshotId().

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/expire/ExpireSnapshotsPlan.java
wzhero1: Removed.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/expire/ExpireSnapshotsPlan.java
wzhero1: Removed.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/expire/ExpireSnapshotsPlanner.java
wzhero1: Done. Extracted `computeSnapshotExpireRange()` as a static method in `ExpireSnapshotsImpl` with an `ExpireRange` return type. Both `ExpireSnapshotsImpl.expire()` and `ExpireSnapshotsPlanner.plan() `now call this shared method instead of duplicating the range computation logic. Also removed the duplicated `findSkippingTags()` from `ExpireSnapshotsPlanner` — it now uses the one in `ExpireSnapshotsImpl`.

yunfengzhou-hub (approved):
+1

JingsongLi (commented):
This is a significant feature (+2347 lines) for parallel snapshot expiration via Flink. Some review points:

1. **Architecture**: Parallelizing snapshot expiration is important for large tables where sequential expiration becomes a bottleneck. However, the approach needs careful consideration:
   - How are snapshots partitioned across parallel tasks? By snapshot ID range?
   - What happens if one parallel task fails while others succeed? Is there a consistent state guarantee?

2. **Concurrency safety**: Snapshot expiration involves deleting data files and updating manifest lists. Concurrent expiration must not:
   - Delete files still referenced by non-expired snapshots
   - Create race conditions with concurrent writers or compactors
   - Leave orphan files if a task fails mid-way

3. **File deletion ordering**: The order of deletion matters - manifest entries should be cleaned up before the snapshot metadata to avoid referencing deleted files.

4. **Configuration**: What controls the parallelism level? Is there a sensible default that won't overwhelm the filesystem with DELETE operations?

5. **Testing**: Given the complexity and risk (data loss potential), this needs thorough integration tests including failure scenarios (task failure mid-expiration, concurrent write during expiration).

6. **Backward compatibility**: Ensure this works correctly alongside the existing sequential expiration path and can be enabled/disabled via configuration.

This is a high-impact feature but also high-risk given it involves data deletion. Please ensure comprehensive test coverage.

Issue #7030, opened 2026-01-13
[Bug] Write and compaction performance significantly degraded when writing to highly partitioned tables


mao-liu:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Paimon version

1.3.1

### Compute Engine

Flink 1.20

### Minimal reproduce step

Disclaimer: performance issues are observed for really big tables

Table:
- Several thousand partitions
- PK table
- Fixed bucket (e.g. 32)
- Partitioned by a string field in the data (not partitioned by time)
- Table stored in cloud object store (e.g. S3)

Flink jobs:
- Streaming, write-only, all partitions are receiving realtime updates
- Batch, full compaction, all partitions require compaction

For the above setup, we noticed that write performance degrades quite significantly, and the performance does **not** correlate with the size of the table or streaming data volume.

After a lot of debugging and logging, we've identified a significant bottleneck for highly partitioned tables:
- The bottleneck appears to be [FileSystemWriteRestore.restoreFiles](https://github.com/apache/paimon/blob/release-1.3/paimon-core/src/main/java/org/apache/paimon/operation/FileSystemWriteRestore.java#L80), which is invoked via [AbstractFileStoreWrite.createWriterContainer](https://github.com/apache/paimon/blob/release-1.3/paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreWrite.java#L415) for each combination of (partition, bucket)
- Each execution of `restoreFiles` would fetch manifest files
  - This leads to API requests to retrieve files from cloud filesystem
  - This is also computationally expensive as manifest files (avro/parquet) are decompressed and deserialized into the right object models
- For a highly partitioned table receiving writes across all partitions, this could lead to the manifest being fetched 100,000s times
  - This results in excessively high API costs for object retrieval
  - This also lead to very long execution time of batch jobs, or unstable streaming jobs that may not keep up with streaming data loads
- Dedicated compaction job performance is particularly impacted by this

### What doesn't meet your expectations?

- Reduce cloud API requests to not fetch manifest in an excessively duplicated manner
- Improve initialization performance for writers

### Anything else?

We have been testing a patch that pre-fetches the entire manifest and caches in memory, to avoid duplicated API requests to retrieve the same object.

Caching also avoids repeated decompression/deserialization of manifest files.

For a batch compaction job:
- Release 1.3: ~3h run time, ~300k API requests
[...]

PR #7031, opened 2026-01-13
[core][flink] WIP for discussion: optimisations for manifest fetching

mao-liu:
### Purpose

https://github.com/apache/paimon/issues/7030

### Tests

WIP

### API and Format

- Adds new writer option - `manifest.prefetch-entries: boolean`

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreScan.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreScan.java
@@ -43,6 +43,9 @@ import org.apache.paimon.utils.Filter;
 import org.apache.paimon.utils.Pair;
 import org.apache.paimon.utils.SnapshotManager;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import javax.annotation.Nullable;
 
 import java.util.ArrayList;
@@ -66,6 +69,8 @@ import static org.apache.paimon.utils.ThreadPoolUtils.randomlyOnlyExecute;
 
 /** Default implementation of {@link FileStoreScan}. */
 public abstract class AbstractFileStoreScan implements FileStoreScan {
+
+    private static final Logger LOG = LoggerFactory.getLogger(AbstractFileStoreScan.class);
 
     private final ManifestsReader manifestsReader;
     private final SnapshotManager snapshotManager;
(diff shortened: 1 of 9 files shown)


paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreScan.java
```diff
@@ -67,6 +70,8 @@
 /** Default implementation of {@link FileStoreScan}. */
 public abstract class AbstractFileStoreScan implements FileStoreScan {
```
mao-liu, line 71: Logging changes to this class only, can ignore

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/StoreCompactOperator.java
```diff
@@ -244,4 +259,65 @@ public Class<? extends StreamOperator> getStreamOperatorClass(ClassLoader classL
             return StoreCompactOperator.class;
         }
     }
+
+    /** {@link StreamOperatorFactory} of {@link StoreCompactOperator}. */
+    public static class CoordinatedFactory
```
mao-liu, line 264: Extending StoreCompactOperator to use WriteOperatorCoordinator

In practice, this did not help with the write performance, as it required very high memory for ManifestEntryCache, and was also repeatedly fetching from cloud filesystem.

JingsongLi (commented):
Interesting optimization for manifest prefetch. A few concerns about the WIP state:

1. **Design question**: What's the prefetch strategy? Is it based on access patterns (LRU-style) or does it eagerly prefetch all manifests in a snapshot? For large tables with thousands of manifests, eager prefetch could cause memory pressure.

2. **Concurrency**: How is the prefetch thread pool managed? Make sure it's bounded and shared appropriately to avoid thread explosion in multi-table scenarios.

3. **Cache invalidation**: When manifests are compacted or expired, how does the prefetch cache stay consistent?

4. **Metrics**: Consider adding metrics for prefetch hit rate to validate the optimization's effectiveness in production.

5. **WIP status**: Since this is marked WIP, I'd suggest documenting the expected performance improvement with benchmarks before merging.

The direction looks promising - manifest reading is often a bottleneck for scan planning. Looking forward to the completed version.

PR #7033, opened 2026-01-13
[Core][Bug-fix] Fix wrong merge order of increment diff split read

LiangDai-Mars:
### Purpose

This pull request aims to fix an issue in the incremental scan where concurrent writes to the same key could finally return wrong merged result. Since records included in merging set are in non-deterministic sequence numbers  could result in unstable merge order, which the incremental scan heavily relies on, would lead to incorrect final results. The fix ensures the correct result by re-sorting the data in the getResult stage of the split diff read process.

### Tests

A new integration test, testIncrementScanModeWithInsertOverwrite, has been added to BatchFileStoreITCase.java. This test simulates the scenario with multiple INSERT OVERWRITE operations with different seq. number on the same key, and then verifies that the diff mode of incremental scan correctly identifies the changes, ensuring the fix is effective.

### API and Format

No API or format changes. This change only involves an internal logic correction and does not affect any external APIs, data storage formats, or configuration files.

### Documentation

This change is an internal bug fix and does not require updates to user-facing documentation.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/source/splitread/IncrementalDiffSplitRead.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/source/splitread/IncrementalDiffSplitRead.java
@@ -206,10 +206,8 @@ public class IncrementalDiffSplitRead implements SplitRead<InternalRow> {
             } else if (kvs.size() == 2) {
                 KeyValue before = kvs.get(0);
                 KeyValue after = kvs.get(1);
-                if (after.level() == AFTER_LEVEL) {
-                    if (!valueAndRowKindEquals(before, after)) {
-                        toReturn = after;
-                    }
+                if (!valueAndRowKindEquals(before, after)) {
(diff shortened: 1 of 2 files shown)


paimon-core/src/main/java/org/apache/paimon/table/source/splitread/IncrementalDiffSplitRead.java
```diff
@@ -213,9 +213,11 @@ public KeyValue getResult() {
             } else if (kvs.size() == 2) {
                 KeyValue before = kvs.get(0);
                 KeyValue after = kvs.get(1);
-                if (after.level() == AFTER_LEVEL) {
-                    if (!valueAndRowKindEquals(before, after)) {
+                if (!valueAndRowKindEquals(before, after)) {
+                    if (after.level() == AFTER_LEVEL) {
```
Aitozi: change to `return after.level() == AFTER_LEVEL ? after : before`?

Aitozi (approved):
+1

PR #7034, opened 2026-01-13
[core] Add row filter & column masking support for table read

plusplusjiajia:
### Purpose
Add row filter & column masking support for table read

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/source/DataSplit.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/source/DataSplit.java
@@ -18,7 +18,6 @@
 
 package org.apache.paimon.table.source;
 
-import org.apache.paimon.catalog.TableQueryAuthResult;
 import org.apache.paimon.data.BinaryRow;
 import org.apache.paimon.data.InternalArray;
 import org.apache.paimon.data.InternalRow;
@@ -59,11 +58,11 @@ import static org.apache.paimon.utils.Preconditions.checkArgument;
 import static org.apache.paimon.utils.Preconditions.checkState;
 
 /** Input splits. Needed by most batch computation engines. */
-public class DataSplit implements QueryAuthSplit {
+public class DataSplit implements Split {
 
-    private static final long serialVersionUID = 8L;
+    private static final long serialVersionUID = 7L;
     private static final long MAGIC = -2394839472490812314L;
-    private static final int VERSION = 9;
+    private static final int VERSION = 8;
 
     private long snapshotId = 0;
     private BinaryRow partition;
@@ -79,8 +78,6 @@ public class DataSplit implements QueryAuthSplit {
 
     private boolean isStreaming = false;
     private boolean rawConvertible;
-
-    @Nullable private TableQueryAuthResult authResult;
 
     public DataSplit() {}
 
@@ -127,12 +124,6 @@ public class DataSplit implements QueryAuthSplit {
 
     public boolean rawConvertible() {
         return rawConvertible;
-    }
-
-    @Override
-    @Nullable
-    public TableQueryAuthResult authResult() {
-        return authResult;
     }
 
     public OptionalLong latestFileCreationEpochMillis() {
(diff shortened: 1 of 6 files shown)


paimon-core/src/main/java/org/apache/paimon/table/source/ReadBuilderImpl.java
```diff
@@ -256,6 +258,20 @@ public TableRead newRead() {
         if (variantAccessInfo != null) {
             read.withVariantAccess(variantAccessInfo);
         }
+        if (table instanceof FileStoreTable) {
+            CoreOptions options = new CoreOptions(table.options());
+            if (options.queryAuthEnabled()) {
+                TableQueryAuth queryAuth =
+                        ((FileStoreTable) table).catalogEnvironment().tableQueryAuth(options);
+                TableQueryAuthResult authResult =
+                        queryAuth.auth(readType == null ? null : readType.getFieldNames());
```
JingsongLi: New Read will invoked in every works. We should not let so many works to do RPC to metastore.

I think maybe we can introduce `QueryAuthSplit` with filter and masking. And here wrapping new TableRead to do filter and masking.

paimon-spark/paimon-spark-ut/src/test/java/org/apache/paimon/spark/SparkCatalogWithRestTest.java
```diff
@@ -237,6 +246,116 @@ public void testMapFunction() throws Exception {
         cleanFunction(functionName);
     }
 
+    @Test
+    public void testColumnMasking() {
+        spark.sql(
+                "CREATE TABLE t_column_masking (id INT, secret STRING, email STRING, phone STRING) TBLPROPERTIES"
+                        + " ('bucket'='1', 'bucket-key'='id', 'file.format'='avro', 'query-auth.enabled'='true')");
```
JingsongLi: Don't need to test bucketed table. Just no bucket.

paimon-spark/paimon-spark-ut/src/test/java/org/apache/paimon/spark/SparkCatalogWithRestTest.java
```diff
@@ -237,6 +246,116 @@ public void testMapFunction() throws Exception {
...
+        spark.sql(
+                "INSERT INTO t_column_masking VALUES (1, 's1', 'user1@example.com', '12345678901'), (2, 's2', 'user2@example.com', '12345678902')");
+
+        // Test single column masking
+        Transform maskTransform =
+                new ConcatTransform(Collections.singletonList(BinaryString.fromString("****")));
+
+        Map<String, Transform> columnMasking = new HashMap<>();
+        columnMasking.put("secret", maskTransform);
+        restCatalogServer.setColumnMaskingAuth(
```
JingsongLi, line 273: Also add case for row filtering.

paimon-core/src/main/java/org/apache/paimon/table/source/DataSplit.java
```diff
@@ -58,11 +59,11 @@
 import static org.apache.paimon.utils.Preconditions.checkState;
 
 /** Input splits. Needed by most batch computation engines. */
-public class DataSplit implements Split {
+public class DataSplit implements QueryAuthSplit {
```
JingsongLi: Don't modify DataSplit, just create a new class to wrap it.
DataSplit has a heavy historical burden, and we need to ensure its compatibility. Its serialization protocol is related to multiple languages.

plusplusjiajia: > Don't modify DataSplit, just create a new class to wrap it.
> DataSplit has a heavy historical burden, and we need to ensure its compatibility. Its serialization protocol is related to multiple languages.

Got it!

paimon-spark/paimon-spark-ut/src/test/java/org/apache/paimon/spark/SparkCatalogWithRestTest.java
plusplusjiajia: > Don't need to test bucketed table. Just no bucket.

done

paimon-spark/paimon-spark-ut/src/test/java/org/apache/paimon/spark/SparkCatalogWithRestTest.java
plusplusjiajia: > Also add case for row filtering.

done

JingsongLi:
+1

Issue #7037, opened 2026-01-14
[Feature] Support re-overwrite on spark for chain table


Stefanietry:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Motivation

When re-overwrite data to delta branch, the corresponding partition on snaopshot branch should  to be deleted.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #7038, opened 2026-01-14
[core] Support re-overwrite by spark for chain table

Stefanietry:
[core]

### Purpose

Linked issue: https://github.com/apache/paimon/issues/7037

When re-overwrite data to delta branch, delete the corresponding partition on snapshot branch.

### Tests

org.apache.paimon.spark.SparkChainTableITCase#testChainTable
org.apache.paimon.spark.SparkChainTableITCase#testHourlyChainTable


paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/commands/WriteIntoPaimonTable.scala
```diff
@@ -82,6 +84,18 @@ case class WriteIntoPaimonTable(
     (dynamicPartitionOverwriteMode, overwritePartition)
   }
 
+  private def postCommit(
```
JingsongLi: I think you should add `CommitCallback` instead of modifying Spark.

Stefanietry: Get, thanks for your suggestion. I have rewritten this function using the CommitCallback mechanism.

JingsongLi (approved):
+1

PR #7040, opened 2026-01-14
[python] refactor file_io to support polymorphism

XiaoHongbo-Hope:
### Purpose

refactor file_io to support polymorphism instead of involve all the logic into one class.

Before:  
- `FileIo`
     - `RestTokenFileIo`

After:
- `FileIo`
    - `LocalFileIo`
    - `PyArrowFileIO`
    - `RESTTokenFileIO`


paimon-python/pypaimon/filesystem/pyarrow_file_io.py
```diff
@@ -0,0 +1,512 @@
...
+        self.logger = logging.getLogger(__name__)
+        scheme, netloc, _ = self.parse_location(path)
+        self.uri_reader_factory = UriReaderFactory(catalog_options)
+        if scheme in {"oss"}:
+            self.filesystem = self._initialize_oss_fs(path)
+        elif scheme in {"s3", "s3a", "s3n"}:
+            self.filesystem = self._initialize_s3_fs()
+        elif scheme in {"hdfs", "viewfs"}:
+            self.filesystem = self._initialize_hdfs_fs(scheme, netloc)
+        elif scheme in {"file"}:
```
discivigour: Maybe local file system is not needed here.

paimon-python/pypaimon/filesystem/local_file_io.py
```diff
@@ -0,0 +1,444 @@
...
+
+
+class LocalFileIO(FileIO):
+    """
+    Local file system implementation of FileIO.
+    """
+    
+    RENAME_LOCK = threading.Lock()
+    
+    INSTANCE = None
```
discivigour: Do we need INSTANCE variable?

paimon-python/pypaimon/filesystem/pyarrow_file_io.py
XiaoHongbo-Hope: > Maybe local file system is not needed here.

removed

paimon-python/pypaimon/filesystem/local_file_io.py
XiaoHongbo-Hope: > Do we need INSTANCE variable?

removed

PR #7051, opened 2026-01-15
[python] support read.batch-size and fix default value

XiaoHongbo-Hope:
### Purpose

Currently, python use default read batch size 4096, which is different with java: 1024. 
It may cause performance issue, especially for the lance format with vector data (array type) with thousands of dimension


JingsongLi:
+1

PR #7055, opened 2026-01-15
[format] Enable blob write/read null

leaves12138:
### Purpose

For the preparation to delete blob


paimon-format/src/main/java/org/apache/paimon/format/blob/BlobFormatWriter.java
```diff
@@ -55,21 +55,21 @@ public BlobFormatWriter(PositionOutputStream out) {
     @Override
     public void addElement(InternalRow element) throws IOException {
         checkArgument(element.getFieldCount() == 1, "BlobFormatWriter only support one field.");
-        checkArgument(!element.isNullAt(0), "BlobFormatWriter only support non-null blob.");
-        Blob blob = element.getBlob(0);
 
         long previousPos = out.getPos();
-        crc32.reset();
-
         write(MAGIC_NUMBER_BYTES);
```
JingsongLi, line 60: Why we need to write magic number for null blob?

PR #7059, opened 2026-01-15
[core] Refactor MergeFunctionFactory to use read rowType

Zouxxyy:
### Purpose

A previous TODO item — now variant push must be used, as the new readType with pushed variant struct is required in MergeFunction


JingsongLi:
Maybe add a case for this? Why to modify this?

JingsongLi:
Thanks @Zouxxyy , +1

PR #7073, opened 2026-01-18
[variant] Refactor VariantAccessInfo to VariantExtraction

Zouxxyy:
### Purpose

Follow up [SPARK-54656](https://issues.apache.org/jira/browse/SPARK-54656):
1. Change `VariantAccessInfo` to `VariantExtraction`
2. Change `String columnName` in `VariantExtraction` to `String[] columnName` to support nested variant pushdown in the future.

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/data/variant/VariantExtraction.java
+++ b/paimon-common/src/main/java/org/apache/paimon/data/variant/VariantExtraction.java
@@ -18,28 +18,36 @@
 
 package org.apache.paimon.data.variant;
 
+import org.apache.paimon.annotation.Experimental;
 import org.apache.paimon.types.DataField;
 
 import java.io.Serializable;
+import java.util.Arrays;
 import java.util.List;
+import java.util.Objects;
 
-/** Variant access information for a variant column. */
-public class VariantAccessInfo implements Serializable {
+/** Variant extraction information that describes fields extraction from a variant column. */
+@Experimental
+public class VariantExtraction implements Serializable {
 
     private static final long serialVersionUID = 1L;
 
-    // The name of the variant column.
-    private final String columnName;
+    /**
+     * Returns the path to the variant column. For top-level variant columns, this is a single
+     * element array containing the column name. For nested variant columns within structs, this is
+     * an array representing the path (e.g., ["structCol", "innerStruct", "variantCol"]).
+     */
+    private final String[] columnName;
 
     // Extracted fields from the variant.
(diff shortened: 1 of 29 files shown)


paimon-common/src/main/java/org/apache/paimon/data/variant/VariantExtraction.java
```diff
@@ -18,28 +18,36 @@
...
     private static final long serialVersionUID = 1L;
 
-    // The name of the variant column.
-    private final String columnName;
+    /**
+     * Returns the path to the variant column. For top-level variant columns, this is a single
+     * element array containing the column name. For nested variant columns within structs, this is
+     * an array representing the path (e.g., ["structCol", "innerStruct", "variantCol"]).
+     */
+    private final String[] columnName;
```
Zouxxyy, line 40: I still believe that making `extractedFields` as a field of `VariantType` is a better approach. This offers the following advantages:

- There's no need to keep adding `withVariantExtractions` in the API, as it will be propagated through the `readType`.
- It natively supports nested types (similar to how I previously changed `projection[][]` into a `RowType`).
- It also makes implementing nested reading on the reader side much cleaner and easier.

Alternatively, we could directly convert `VariantType` into the `RowType` resulting from shredding reads. This approach would require the RowType to be marked as originating from a variant by special method. (Spark' plan)

```scala
val METADATA_KEY = "__VARIANT_METADATA_KEY"
val PATH_KEY = "path"
val FAIL_ON_ERROR_KEY = "failOnError"
val TIME_ZONE_ID_KEY = "timeZoneId"

def isVariantStruct(s: StructType): Boolean =
  s.fields.length > 0 && s.fields.forall(_.metadata.contains(METADATA_KEY))
```

Zouxxyy: Perhaps I can follow Spark's approach and design it this way—by leveraging the description field in Paimon's DataField.The description value could be encoded similarly as follows:

```
__VARIANT_METADATA:<path>;<failOnError>;<timeZoneId>

e.g.
__VARIANT_METADATA:$.a.b;true;UTC
```

I'm fine with using either `,` or `;` as the delimiter, since both are safe (they won't be interpreted as part of a column name—the column names will be present in the `path`).

And introduce a method like this

```scala
def isVariantRow(s: RowType): Boolean =
  s.fields.length > 0 && s.fields.forall(_.description.startWith("__VARIANT_METADATA"))
```

PR #7077, opened 2026-01-19
[BugFix] Avro schema inconsistents between Java and Python

chenghuichen:
### Purpose
- Align Python's Avro schema generation with Java implementation. Currently they are inconsist.
- Full support for `TIMESTAMP_LTZ` in pypaimon.

### Tests

Re-using `run_mixed_tests.sh`, changes inside:
- JavaPyE2ETest.java
- java_py_read_write_test.py

### DIFF
--- a/paimon-python/pypaimon/schema/data_types.py
+++ b/paimon-python/pypaimon/schema/data_types.py
@@ -596,19 +596,19 @@ class PyarrowFieldParser:
         elif pyarrow.types.is_timestamp(field_type):
             unit = field_type.unit
             if field_type.tz is None:
-                if unit == 'us':
-                    return {"type": "long", "logicalType": "timestamp-micros"}
-                elif unit == 'ms':
+                if unit == 'ms':
                     return {"type": "long", "logicalType": "timestamp-millis"}
-                else:
+                elif unit == 'us':
                     return {"type": "long", "logicalType": "timestamp-micros"}
-            else:
-                if unit == 'us':
-                    return {"type": "long", "logicalType": "local-timestamp-micros"}
-                elif unit == 'ms':
-                    return {"type": "long", "logicalType": "local-timestamp-millis"}
                 else:
+                    raise ValueError(f"Avro does not support pyarrow timestamp with unit {unit}.")
+            else:
+                if unit == 'ms':
+                    return {"type": "long", "logicalType": "local-timestamp-millis"}
+                elif unit == 'us':
                     return {"type": "long", "logicalType": "local-timestamp-micros"}
+                else:
+                    raise ValueError(f"Avro does not support pyarrow timestamp with unit {unit}.")
         elif pyarrow.types.is_list(field_type) or pyarrow.types.is_large_list(field_type):
             value_field = field_type.value_field
             return {


paimon-core/src/test/java/org/apache/paimon/JavaPyE2ETest.java
```diff
@@ -106,118 +106,126 @@ public void before() throws Exception {
...
-
-            commit.commit(0, write.prepareCommit(true, 0));
+        for (String format : Arrays.asList("parquet", "orc", "avro")) {
+            Identifier identifier = identifier("mixed_test_append_tablej_" + format);
+            Schema schema =
+                    Schema.newBuilder()
+                            .column("id", DataTypes.INT())
+                            .column("name", DataTypes.STRING())
+                            .column("category", DataTypes.STRING())
+                            .column("value", DataTypes.DOUBLE())
```
JingsongLi, line 116: Can you also add TIMESTAMP and LOCAL_ZONE_TIMESTAMP?

chenghuichen: > Can you also add TIMESTAMP and LOCAL_ZONE_TIMESTAMP?

Sure!

chenghuichen: > Can you also add TIMESTAMP and LOCAL_ZONE_TIMESTAMP?

Updated. This did reveal some issues.

paimon-python/pypaimon/schema/data_types.py
```diff
@@ -589,31 +595,40 @@ def to_avro_type(field_type: pyarrow.DataType, field_name: str) -> Union[str, Di
...
-                return {"type": "long", "logicalType": "timestamp-micros"}
-            elif unit == 'ms':
-                return {"type": "long", "logicalType": "timestamp-millis"}
+            if field_type.tz is None:
+                if unit == 'us':
+                    return {"type": "long", "logicalType": "local-timestamp-micros"}
+                elif unit == 'ms':
+                    return {"type": "long", "logicalType": "local-timestamp-millis"}
+                else:
+                    return {"type": "long", "logicalType": "local-timestamp-micros"}
```
XiaoHongbo-Hope: None tz should be timestamp-micros.

chenghuichen: @JingsongLi hello, I have had a discussion with HongBo with this part of code. 

My understanding is that the type mapping in this code is consistent with the design intent, but it's inconsistent with the Java version. And the Java version, like Flink, may be a historical mistake https://nightlies.apache.org/flink/flink-docs-master/docs/connectors/table/formats/avro/ 
> Before 1.19, The default behavior of Flink wrongly mapped both SQL TIMESTAMP and TIMESTAMP_LTZ type to AVRO TIMESTAMP.
The correct behavior is Flink SQL TIMESTAMP maps Avro LOCAL TIMESTAMP and Flink SQL TIMESTAMP_LTZ maps Avro TIMESTAMP

 If my understanding is correct, the question now becomes whether pypaimon should fix this in advance. I need your help to confirm this.

JingsongLi: We should take a look to Paimon Types, not Flink Types.

JingsongLi: See:
https://paimon.apache.org/docs/master/concepts/spec/fileformat/#avro
And:
org.apache.paimon.format.avro.AvroSchemaConverter

JingsongLi: Whatever Java does, Python should align with. We need to ensure compatibility

JingsongLi:
I checked the current implementation and it seems to be correct. Our Avro format has a very strange mapping relationship, the mapping on the Java side is reversed compared to standard...

paimon-python/pypaimon/schema/data_types.py
chenghuichen: > Whatever Java does, Python should align with. We need to ensure compatibility

Updated. It's now aligned with Java.

PR #7078, opened 2026-01-19
[spark] Support source table can be any type in data-evolution

shidayang:
Currently，source table can only be paimon table in data-evolution


JingsongLi:
Looks good to me! Thanks @shidayang

PR #7081, opened 2026-01-19
[parquet] Bump parquet version to 1.16.0

Zouxxyy:
### Purpose

Parquet added the variant logical type in version 1.16.0, which we need. Although I'd prefer to upgrade to 1.17.0, but it requires at least JDK 11.

Shade `org.locationtech.jts:jts-core` (< 1M), because it is required by the geography type in Parquet 1.16.0.


JingsongLi (approved):
+1

PR #7082, opened 2026-01-20
[arrow] Support arrow variant/shredding writer.

leaves12138:
### Purpose

Support variant/shredding for arrow writer


JingsongLi:
+1

PR #7088, opened 2026-01-20
[python] optimize schema validation and support binary/large_binary type conversion

XiaoHongbo-Hope:
### Purpose

- Optimize schema validation

  - Added `preserve_index=False` in `write_pandas()` to prevent pandas index from creating extra columns
  - support `write_cols=None` and optimize error message

- Binary/Large_Binary Type Compatibility
  - Ray Data converts `large_binary()` to `binary()` when reading from Paimon tables, make it compatibility by supporting binary/large_binary type conversion
  -  

### Tests
`ray_data_test.test_ray_data_read_with_blob`

### Documentation
add note about ray data converts `large_binary()` to `binary()` when reading from Paimon tables in `python-api.md`

### DIFF
--- a/paimon-python/pypaimon/tests/ray_data_test.py
+++ b/paimon-python/pypaimon/tests/ray_data_test.py
@@ -117,125 +117,20 @@ class RayDataTest(unittest.TestCase):
         self.assertIsNotNone(ray_dataset, "Ray dataset should not be None")
         self.assertEqual(ray_dataset.count(), 5, "Should have 5 rows")
 
-    def test_ray_data_read_and_write_with_blob(self):
-        import time
-        pa_schema = pa.schema([
-            ('id', pa.int64()),
-            ('name', pa.string()),
-            ('data', pa.large_binary()),  # Table uses large_binary for blob
-        ])
+        # Test basic operations
+        sample_data = ray_dataset.take(3)
+        self.assertEqual(len(sample_data), 3, "Should have 3 sample rows")
 
-        schema = Schema.from_pyarrow_schema(
-            pa_schema,
-            options={
-                'row-tracking.enabled': 'true',
-                'data-evolution.enabled': 'true',
-                'blob-field': 'data',
-            }
-        )
-
-        table_name = f'default.test_ray_read_write_blob_{int(time.time() * 1000000)}'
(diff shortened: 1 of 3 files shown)


paimon-python/pypaimon/tests/ray_data_test.py
```diff
@@ -115,6 +116,66 @@ def test_basic_ray_data_read(self):
         self.assertIsNotNone(ray_dataset, "Ray dataset should not be None")
         self.assertEqual(ray_dataset.count(), 5, "Should have 5 rows")
 
+    def test_ray_data_read_with_blob(self):
```
JingsongLi: This test can be passed in the master.

XiaoHongbo-Hope: > This test can be passed in the master.

I need to think more about  this PR and get back to you.

XiaoHongbo-Hope: > This test can be passed in the master.

You're right that the direct read → write roundtrip passes on master. The issue occurs when map_batches with dict is involved.  

When a user function in map_batches returns a Python dict (which is the most common pattern), PyArrow re-infers bytes as binary, losing the original large_binary (BLOB) type: 
` large_binary → to_pylist() → Python bytes → from_pydict() → binary`
see the simple case: `test_dict_return_loses_large_binary_type` for better understanding.

So the pipeline to_ray() → map_batches(fn returning dict) → write_ray() fails because _validate_pyarrow_schema rejects binary  against a large_binary table schema. We can do  a schema cast when write ray or let user work around to convert the right schema in user code. Both should be ok, what is your suggestion.

leaves12138 (approved):
LGTM. The fix is scoped appropriately to the Ray datasink path: Ray/PyArrow can infer Python `bytes` as `binary` after `map_batches`, while the Paimon BLOB field expects `large_binary`, so casting only those columns back to the target table schema before `write_arrow` preserves the existing strict schema validation for other paths.

I also verified the intended behavior locally with a minimal write-side reproduction: current master fails with `Input schema ... data: binary` vs `Table schema ... data: large_binary`, while this PR writes successfully and reads back `data: large_binary`.

JingsongLi:
+1

PR #7090, opened 2026-01-21
[core] Introduce mergedRowCount in Split

JingsongLi:
### Purpose

```Java
    /**
     * The row count in files, may be duplicated, such as in the primary key table and the
     * Data-Evolution Append table.
     */
    long rowCount();

    /**
     * Return the merged row count of data files. For example, when the delete vector is enabled in
     * the primary key table, the number of rows that have been deleted will be subtracted from the
     * returned result. In the Data Evolution mode of the Append table, the actual number of rows
     * will be returned.
     */
    OptionalLong mergedRowCount();
```


XiaoHongbo-Hope:
+1

PR #7091, opened 2026-01-21
[test][spark] Add test case for timestamp partition column

Zouxxyy:
### Purpose

Add test case for timestamp partition column


JingsongLi:
+1

PR #7093, opened 2026-01-21
[core] Introduce IncrementalSplit to simplify DataSplit

JingsongLi:
### Purpose

DataSplit is too complex, and we have exposed too many useless fields. We should simplify it and isolate other scenarios.


paimon-core/src/main/java/org/apache/paimon/table/source/snapshot/SnapshotReaderImpl.java
```diff
@@ -505,68 +508,84 @@ private Plan toChangesPlan(
...
-                                    deletionFilesMap.getOrDefault(
-                                            Pair.of(part, bucket), Collections.emptyMap())));
+                                    after,
+                                    afterDeletionFilesMap.getOrDefault(
+                                            Pair.of(part, bucket), Collections.emptyMap()));
                 }
-                splits.add(builder.build());
+
+                if (totalBuckets == null) {
+                    System.out.println("");
```
XiaoHongbo-Hope: debug code left

PR #7097, opened 2026-01-22
[spark] Support UPDATE on Paimon append-only table in spark V2 write

kerwin-zk:
### Purpose

Support UPDATE on Paimon append-only table in spark V2 write

### Tests

CI

### DIFF
--- a/paimon-spark/paimon-spark-3.5/src/test/scala/org/apache/paimon/spark/sql/UpdateTableTest.scala
+++ b/paimon-spark/paimon-spark-3.5/src/test/scala/org/apache/paimon/spark/sql/UpdateTableTest.scala
@@ -25,3 +25,9 @@ class UpdateTableTest extends UpdateTableTestBase {
     super.sparkConf.set("spark.paimon.write.use-v2-write", "false")
   }
 }
+
+class V2UpdateTableTest extends UpdateTableTestBase {
+  override protected def sparkConf: SparkConf = {
+    super.sparkConf.set("spark.paimon.write.use-v2-write", "true")
+  }
+}
--- a/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/catalyst/analysis/PaimonUpdateTable.scala
+++ b/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/catalyst/analysis/PaimonUpdateTable.scala
@@ -22,7 +22,7 @@ import org.apache.paimon.spark.commands.UpdatePaimonTableCommand
 import org.apache.paimon.table.FileStoreTable
 
 import org.apache.spark.sql.catalyst.expressions.Literal.TrueLiteral
-import org.apache.spark.sql.catalyst.plans.logical.{LogicalPlan, UpdateTable}
+import org.apache.spark.sql.catalyst.plans.logical.{Assignment, LogicalPlan, UpdateTable}
 import org.apache.spark.sql.catalyst.rules.Rule
 
 import scala.collection.JavaConverters._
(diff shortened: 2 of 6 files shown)


paimon-spark/paimon-spark-3.5/src/test/scala/org/apache/paimon/spark/sql/UpdateTableTest.scala
```diff
@@ -18,4 +18,12 @@
 
 package org.apache.paimon.spark.sql
 
+import org.apache.spark.SparkConf
+
 class UpdateTableTest extends UpdateTableTestBase {}
```
Zouxxyy: Add super.sparkConf.set("spark.paimon.write.use-v2-write", "false") here

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/catalyst/analysis/RowLevelHelper.scala
```diff
@@ -95,4 +95,13 @@ trait RowLevelHelper extends SQLConfHelper {
       table.getTable.asInstanceOf[FileStoreTable],
       condition)
   }
+
+  /** Determines if DataSourceV2 update is not supported for the given table. */
+  protected def shouldFallbackToV1Update(table: SparkTable, updateTable: UpdateTable): Boolean = {
+    shouldFallbackToV1(table) ||
+    !updateTable.rewritable ||
```
YannByron, line 102: `updateTable.rewritable` should always be true here, right?

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/extensions/PaimonSparkSessionExtensions.scala
```diff
@@ -51,6 +51,8 @@ class PaimonSparkSessionExtensions extends (SparkSessionExtensions => Unit) {
     extensions.injectPostHocResolutionRule(_ => PaimonDeleteTable)
     extensions.injectPostHocResolutionRule(spark => PaimonMergeInto(spark))
 
+    extensions.injectPostHocResolutionRule(_ => RewritePaimonV2UpdateTable)
```
YannByron: I suggest  to use one single-rule to deal with update operations. In it, we can always align assignments first, and then chose the v1/v2 code plan.

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/catalyst/analysis/RowLevelHelper.scala
kerwin-zk: @YannByron When the table contains a CHAR column, it returns false.

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/catalyst/analysis/PaimonUpdateTable.scala
```diff
@@ -45,21 +45,31 @@ object PaimonUpdateTable
...
-            if (paimonTable.coreOptions().dataEvolutionEnabled()) {
-              throw new RuntimeException(
-                "Update operation is not supported when data evolution is enabled yet.")
-            }
+            val alignedUpdateTable = u.copy(assignments = alignedAssignments)
+
+            if (!shouldFallbackToV1Update(table, alignedUpdateTable)) {
+              alignedUpdateTable
+            } else {
+              val primaryKeys = paimonTable.primaryKeys().asScala.toSeq
```
YannByron: I think `update pk columns` and `update table that enable data evolution`, in either dsv1 or dsv2 mode, are not supported, so move these two check before this line:
```if (!shouldFallbackToV1Update(table, alignedUpdateTable)) {```

PR #7099, opened 2026-01-22
[core] Fix Iceberg metadata parsing error by excluding null doc field from JSON serialization

KangJiJi:
### Purpose

When using Paimon table with `metadata.iceberg.storage` option to read metadata as Iceberg table, Iceberg fails to parse the metadata JSON with the following error:

```
Error while reading table: iceberg, error message: Failed to parse Iceberg table schema 
with the following error: [json.exception.type_error] type must be string, but is null
```

This happens because Paimon serializes the `doc` field as `"doc": null` when the field description is not set. According to Iceberg spec, the doc field is optional. It is preferable to omit it rather than set it to null when not present.

This PR adds `@JsonInclude(JsonInclude.Include.NON_NULL)` annotation to the `doc` field in `IcebergDataField`, so that null values are excluded from JSON output.

**Before:**
```json
{
  "id": 1,
  "name": "column_name",
  "required": false,
  "type": "string",
  "doc": null
}
```

**After:**
```json
{
  "id": 1,
  "name": "column_name",
  "required": false,
  "type": "string"
}
```

### Tests

- `IcebergDataFieldTest#testDocFieldSerializationWithNullValue`
- `IcebergDataFieldTest#testDocFieldSerializationWithNonNullValue`

### API and Format

No API changes.

### Documentation

[Iceberg](https://iceberg.apache.org/spec/#schemas)

### DIFF
--- a/paimon-core/src/test/java/org/apache/paimon/iceberg/metadata/IcebergDataFieldTest.java
+++ b/paimon-core/src/test/java/org/apache/paimon/iceberg/metadata/IcebergDataFieldTest.java
@@ -35,6 +35,7 @@ import org.apache.paimon.types.RowType;
 import org.apache.paimon.types.TimestampType;
 import org.apache.paimon.types.VarBinaryType;
 import org.apache.paimon.types.VarCharType;
+import org.apache.paimon.utils.JsonSerdeUtil;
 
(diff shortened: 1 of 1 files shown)


JingsongLi:
+1 thanks @KangJiJi

PR #7110, opened 2026-01-22
[variant] Annotate Variant columns with Variant logical type annotation

Zouxxyy:
### Purpose

Subtask of https://github.com/apache/paimon/issues/4471, annotate Variant columns with Variant logical type annotation


JingsongLi (approved):
+1

PR #7123, opened 2026-01-26
[core] Split `blob-as-descriptor` into `read-blob-as-descriptor` and `write-blob-from-descriptor`

Zouxxyy:
### Purpose

There are two main reasons for this change:

1. `blob-as-descriptor` is highly ambiguous during writes—it actually means that the input for writing is a descriptor, not that the blob itself is being written as a descriptor.

2. A single configuration cannot adequately serve both use cases. I believe, the most common scenario should be:

  - `write-blob-from-descriptor=true` so that data is loaded at write time which is extremely memory-efficient. This maybe can be set to default true, in fact, there has been discussion on this topic: https://github.com/apache/paimon/pull/7021
  - `read-blob-as-descriptor=false` ~~since in most read scenarios~~ in some case, we need the original raw data rather than a descriptor.

### DIFF
--- a/docs/content/append-table/blob.md
+++ b/docs/content/append-table/blob.md
@@ -571,11 +571,10 @@ SELECT * FROM video_table;  -- Returns actual blob bytes from Paimon storage
 
 ## Limitations
 
-1. **Single Blob Field**: Currently, only one blob field per table is supported.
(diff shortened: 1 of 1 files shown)


JingsongLi:
We need to conduct more research.

PR #7125, opened 2026-01-26
[format] Introduce null blob

leaves12138:
### Purpose

Linked issue: close #xxx


paimon-format/src/main/java/org/apache/paimon/format/blob/BlobFormatWriter.java
```diff
@@ -80,7 +80,10 @@ public boolean deleteFileUponAbort() {
     @Override
     public void addElement(InternalRow element) throws IOException {
         checkArgument(element.getFieldCount() == 1, "BlobFormatWriter only support one field.");
-        checkArgument(!element.isNullAt(0), "BlobFormatWriter only support non-null blob.");
+        if (element.isNullAt(0)) {
+            lengths.add(-1L);
+            return;
```
JingsongLi, line 85: It is better to notify this to `writeConsumer`.

Issue #7019, opened 2026-01-13
[Feature] Introduce Data Evolution MergeInto procedure for Flink


steFaiz:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Motivation

As current org.apache.paimon.flink.procedure.MergeIntoProcedure implementation, we could also introduce a DataEvolution MergeInto Action and Procedure, to support light-weight column append and update in Flink.

### Solution

Follow spark's MergeIntoPaimonDataEvolutionTable implementation.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #7128, opened 2026-01-27
[flink] introduce a simplified MERGE INTO procedure on data-evolution-table for flink

steFaiz:
### Purpose
Linked issue: https://github.com/apache/paimon/issues/7019

This PR is about to introduce a simplified MERGE INTO action/procedure on DataEvolutionTable for flink.
The motivation is that for data-evolution tables, we could efficiently update or insert columns without rewriting existing data files. Paimon-spark module implements it through `merge into` syntax, which is not supported by flink. So we introduce this action to simulate `merge into` behavier.

<strong style="color: red;">NOTE</strong>: Due to limitations in Flink’s implementation, compared with paimon-spark we currently only support the MERGE UPDATE SET branch and do not support inserting new records yet.

The process can be illustrated as below:
image

1. Source table INNER JOIN target table on merge condition.
This step will assign `_row_id` for each row in source table.
2. Shuffle the joined table by corresponding `FirstRowId` of each newly assigned `_row_id`.
This step is about to ensure that rows belonging to the same files should be processed by same writer operators.
3. Write updated/inserted columns to new files.
  a. Sort rows by `_row_id`
  b. Read original data from each row ranges, merge original data with new rows
  c. Write out merged data 
4. Commit new files.

This implementation is specially designed for cases where the source table may be much smaller than the target table. Each writer is responsible for reading the original file data. Another possible approach is to perform a left outer join of the target table with the source table, rather than an inner join.

#### Merge Detail
New rows will be merged with existing rows to make new files aligned with existing files. For example, consider existing rows:
| _row_id | value (double) | first_row_id |
|---:|---:|:---|
| 1 | 12.34 | 1 |
| 2 | 0.00 | 1 |
| 3 | -7.50 | 1 |
| 4 | 100.01 | 1 |
| 5 | 3.14 | 1 |

They belong to a same file whose row range is [1, 5]
Then a new updated row comes:
| _row_id | value (double) | first_row_id |
|---:|---:|:---|
| 3 | 10000.00 | 1 |

We will merge exiting file and the new file, write out:
| _row_id | value (double) | first_row_id |
|---:|---:|:---|
| 1 | 12.34 | 1 |
| 2 | 0.00 | 1 |
| 3 | **10000.00** | 1 |
| 4 | 100.01 | 1 |
| 5 | 3.14 | 1 |

### Tests
Please see org.apache.paimon.flink.action.DataEvolutionMergeIntoActionITCase

### API and Format
Do not modify any existing api.

### Documentation
Will be added ASAP

### DIFF
--- a/docs/content/append-table/data-evolution.md
+++ b/docs/content/append-table/data-evolution.md
@@ -57,7 +57,9 @@ CREATE TABLE target_table (id INT, b INT, c INT) TBLPROPERTIES (
 INSERT INTO target_table VALUES (1, 1, 1), (2, 2, 2);
 ```
 
-Now we could only support spark 'MERGE INTO' statement to update partial columns.
+Now we could update partial columns by spark 'MERGE INTO' statement or flink 'data_evolution_merge_into' procedure:
+
+### Spark
 
 ```sql
 CREATE TABLE source_table (id INT, b INT);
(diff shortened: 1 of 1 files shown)


JingsongLi (commented):
Please add documentation in `/append-table/data-evolution` too.

PR #7136, opened 2026-01-27
[iceberg] Support real uuid for iceberg table

jianguotian:
### Purpose
If iceberg table, RESTCatalog  getTable return fullName. Support real uuid for iceberg table.

### Tests

org.apache.paimon.rest.RESTCatalogTest#testCreateIcebergTable, verify uuid


JingsongLi:
+1

PR #7137, opened 2026-01-28
[rest] Fix auth header generation order in HttpClient

sundapeng:
### Purpose

Fix the order of authentication header generation in HttpClient to ensure proper request signing for authentication mechanisms that require complete request body information.

In the original implementation, authentication headers were generated before setting the request entity, which could cause issues for authentication mechanisms (like AlibabaCloud DLF ApiSigner) that need to sign the complete request including the body content.

This PR adjusts the execution order in `post()` and `delete()` methods to:
1. Set the encoded request entity first
2. Generate authentication headers with complete request information
3. Set the headers to the request

### Tests

- Existing unit tests for HttpClient should pass

### API and Format

No API or storage format changes. This is an internal implementation fix that maintains backward compatibility.

### Documentation

No new feature introduced. This is a bug fix for authentication header generation order.

### DIFF
--- a/paimon-api/src/main/java/org/apache/paimon/rest/HttpClient.java
+++ b/paimon-api/src/main/java/org/apache/paimon/rest/HttpClient.java
@@ -89,19 +89,14 @@ public class HttpClient implements RESTClient {
             RESTRequest body,
             Class<T> responseType,
             RESTAuthFunction restAuthFunction) {
-        try {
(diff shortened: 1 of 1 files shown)


sundapeng:
Marked as WIP, GET also need to fix.

JingsongLi:
+1

PR #7143, opened 2026-01-28
[python] Add multi-threaded prefetch for pytorch streaming read

XiaoHongbo-Hope:
### Purpose

Linked issue: close #xxx


JingsongLi (commented):
Maybe async is enough? Multiple threads may consume too many memory. And has there been any performance test to measure the improvement?

JingsongLi:
+1

PR #7153, opened 2026-01-29
[python/hotfix] correct with_slice index for split with multi files

XiaoHongbo-Hope:
### Purpose

Fixes with_slice for splits with multiple data files by computing per-file ranges and correct next_pos instead of using only the first file.

Linked issue: close #xxx


JingsongLi (approved):
+1

PR #7157, opened 2026-01-30
[python] support read after data evolution updating by shard

XiaoHongbo-Hope:
## Problem

When the user updates a column for **only one shard** (e.g. ShardTableUpdator runs shard 0 only and writes new column `d`), **full table read** fails:

```
pyarrow.lib.ArrowInvalid: Schema at index 1 was different: d: int32 vs d: null
```

Only that shard’s files have the new column; other files do not. Concat batches → schema mismatch → crash. To fix the issue, we support data evolution shard read.

### DIFF
--- a/paimon-python/pypaimon/read/reader/format_pyarrow_reader.py
+++ b/paimon-python/pypaimon/read/reader/format_pyarrow_reader.py
@@ -34,18 +34,18 @@ class FormatPyArrowReader(RecordBatchReader):
     and filters it based on the provided predicate and projection.
     """
 
-    def __init__(self, file_io: FileIO, file_format: str, file_path: str, read_fields: List[str],
-                 push_down_predicate: Any, batch_size: int = 1024,
-                 read_schema_fields: Optional[List[DataField]] = None):
+    def __init__(self, file_io: FileIO, file_format: str, file_path: str,
+                 read_fields: List[DataField],
+                 push_down_predicate: Any, batch_size: int = 1024):
         file_path_for_pyarrow = file_io.to_filesystem_path(file_path)
         self.dataset = ds.dataset(file_path_for_pyarrow, format=file_format, filesystem=file_io.filesystem)
         self.read_fields = read_fields
-        self.read_schema_fields = read_schema_fields
+        self._read_field_names = [f.name for f in read_fields]
 
         # Identify which fields exist in the file and which are missing
         file_schema_names = set(self.dataset.schema.names)
-        self.existing_fields = [f for f in read_fields if f in file_schema_names]
-        self.missing_fields = [f for f in read_fields if f not in file_schema_names]
+        self.existing_fields = [f.name for f in read_fields if f.name in file_schema_names]
+        self.missing_fields = [f.name for f in read_fields if f.name not in file_schema_names]
 
         # Only pass existing fields to PyArrow scanner to avoid errors
         self.reader = self.dataset.scanner(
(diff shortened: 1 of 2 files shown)


paimon-python/pypaimon/read/reader/data_file_batch_reader.py
```diff
@@ -38,7 +38,8 @@ def __init__(self, format_reader: RecordBatchReader, index_mapping: List[int], p
                  max_sequence_number: int,
                  first_row_id: int,
                  row_tracking_enabled: bool,
-                 system_fields: dict):
+                 system_fields: dict,
+                 requested_field_names: Optional[List[str]] = None):
```
JingsongLi: You should just use `fields: List[DataField]`?

paimon-python/pypaimon/read/table_read.py
```diff
@@ -54,31 +58,49 @@ def _record_generator():
...
             columns.append(col)
 
         return pyarrow.RecordBatch.from_arrays(columns, schema=target_schema)
 
+    @staticmethod
+    def _schema_with_row_tracking_not_null(schema: pyarrow.Schema) -> pyarrow.Schema:
+        """Ensure _ROW_ID and _SEQUENCE_NUMBER are not null (per SpecialFields)."""
+        fields = []
+        for field in schema:
+            if field.name == SpecialFields.ROW_ID.name or field.name == SpecialFields.SEQUENCE_NUMBER.name:
```
JingsongLi: Why it can be nullable?

XiaoHongbo-Hope: > Why it can be nullable?

A bug here. Nullable info of row-tracking system fields is lost during `_assign_row_tracking`. Opened a separate PR #7174 to fix it.

paimon-python/pypaimon/read/reader/data_file_batch_reader.py
XiaoHongbo-Hope: > You should just use `fields: List[DataField]`?

Updated

docs/content/pypaimon/data-evolution.md
```diff
@@ -204,3 +204,8 @@ commit.close()
 - **Row order matters**: the batches you write must have the **same number of rows** as the batches you read, in the
   same order for that shard.
 - **Parallelism**: run multiple shards by calling `new_shard_updator(shard_idx, num_shards)` for each shard.
+
+## Read After Partial Shard Update
```
JingsongLi: I feel like this document doesn't make much sense

paimon-python/pypaimon/read/reader/concat_batch_reader.py
```diff
@@ -206,9 +208,17 @@ def read_arrow_batch(self) -> Optional[RecordBatch]:
...
             if batch_index >= 0 and batches[batch_index] is not None:
-                columns.append(batches[batch_index].column(field_index).slice(0, min_rows))
+                src_batch = batches[batch_index]
+                if field_name in src_batch.schema.names:
+                    column = src_batch.column(
+                        src_batch.schema.get_field_index(field_name)
+                    ).slice(0, min_rows)
+                    columns.append(column)
+                else:
+                    columns.append(pa.nulls(min_rows, type=self.schema.field(i).type))
```
JingsongLi: This work should be done in `DataFileBatchReader`?

paimon-python/pypaimon/read/reader/data_file_batch_reader.py
```diff
@@ -93,32 +130,127 @@ def read_arrow_batch(self, start_idx=None, end_idx=None) -> Optional[RecordBatch
...
+                        inter_arrays.append(record_batch.column(col_idx))
+                        inter_names.append(name)
+                    elif real_index < record_batch.num_columns:
                         inter_arrays.append(record_batch.column(real_index))
-                        inter_names.append(record_batch.schema.field(real_index).name)
+                        inter_names.append(name)
+                    else:
+                        field = self.schema_map.get(name)
+                        inter_arrays.append(
+                            pa.nulls(num_rows, type=field.type) if field is not None else pa.nulls(num_rows)
```
JingsongLi: I don't get it, `FormatPyArrowReader` have already handled `read_fields`.

docs/content/pypaimon/data-evolution.md
XiaoHongbo-Hope: > I feel like this document doesn't make much sense

Removed

JingsongLi (approved):
Thanks @XiaoHongbo-Hope ! Looks good to me.

PR #7163, opened 2026-01-31
[python] Python BTree index reader supports INT and BIGINT

JingsongLi:
### Purpose

After https://github.com/apache/paimon/pull/7160


XiaoHongbo-Hope:
+1

PR #7172, opened 2026-02-02
[core] Allow chain table to use non-deduplicate merge engines.

JunRuiLee:
### Purpose

Previously chain table was restricted to only use deduplicate merge engine.
This change removes that limitation, allowing chain table to work with other merge engines like partial-update.
### Tests

- Add unit test `testChainTableAllowsNonDeduplicateMergeEngine` to verify chain table can use partial-update merge engine
- Add integration test `testChainTableWithPartialUpdate` to verify chain table read/write with partial-update merge engine


Aitozi (approved):
+1

PR #7175, opened 2026-02-02
[core] Support pushing down Decimal filter in parquet format

kerwin-zk:
### Purpose

Support pushing down Decimal filter in parquet format

### Tests

CI


JingsongLi (approved):
+1

PR #7177, opened 2026-02-02
[python] Refactor to File Scanner and add _filter_manifest_files_by_row_ranges

JingsongLi:
### Purpose

1. Simplify StartingScanner to FileScanner, just one class.
2. Introduce `_filter_manifest_files_by_row_ranges` to accelerate query with global index.

### DIFF
--- a/paimon-python/pypaimon/manifest/manifest_file_manager.py
+++ b/paimon-python/pypaimon/manifest/manifest_file_manager.py
@@ -53,15 +53,15 @@ class ManifestFileManager:
         def _process_single_manifest(manifest_file: ManifestFileMeta) -> List[ManifestEntry]:
             return self.read(manifest_file.file_name, manifest_entry_filter, drop_stats)
 
-        def _entry_identifier(entry: ManifestEntry) -> tuple:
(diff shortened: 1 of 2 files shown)


paimon-python/pypaimon/manifest/manifest_list_manager.py
```diff
@@ -79,6 +81,8 @@ def read(self, manifest_list_name: str) -> List[ManifestFileMeta]:
                 num_deleted_files=record['_NUM_DELETED_FILES'],
                 partition_stats=partition_stats,
                 schema_id=record['_SCHEMA_ID'],
+                min_row_id=record['_MIN_ROW_ID'],
+                max_row_id=record['_MAX_ROW_ID'],
             )
```
XiaoHongbo-Hope, line 86: Should we consider to be compatible with older manifest without  _MIN_ROW_ID and _MAX_ROW_ID?

JingsongLi: This will be None, and this file will not be skipped.

PR #7188, opened 2026-02-03
[hive] Remove duplicate assignments to createTime for Hive table

huangxiaopingRD:
### Purpose
The `getEmptyTable` method of Hive code will call `setCreateTime`, so we can avoid unnecessary call of `setCreateTime`.
https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/metadata/Table.java#L258


JingsongLi (commented):
LGTM. This is a trivial 1-line fix removing a duplicate `createTime` assignment in HiveCatalog. The duplicate assignment is clearly unnecessary since the value is already set. Clean and correct.

PR #7195, opened 2026-02-03
[core] Fix that get format table in rest catalog might throw permission exception

yuzelin:
### Purpose

In some cases we only need the format table object but won't access the data.

The exception stack is like:
```
Caused by: com.aliyun.oss.OSSException: Access denied by authorizer's policy.
xxx 
xxx
	at org.apache.paimon.fs.FileIO.checkAccess(FileIO.java:638) ~[?:?]
	at org.apache.paimon.fs.FileIO.get(FileIO.java:552) ~[?:?]
	at org.apache.paimon.rest.RESTCatalog.fileIOFromOptions(RESTCatalog.java:1130) ~[?:?]
	at org.apache.paimon.catalog.CatalogUtils.toFormatTable(CatalogUtils.java:407) ~[?:?]
	at org.apache.paimon.catalog.CatalogUtils.loadTable(CatalogUtils.java:255) ~[?:?]
	at org.apache.paimon.rest.RESTCatalog.getTable(RESTCatalog.java:297) ~[?:?]
```

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/rest/RESTCatalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/rest/RESTCatalog.java
@@ -73,7 +73,6 @@ import org.apache.paimon.shade.org.apache.commons.lang3.StringUtils;
 import javax.annotation.Nullable;
 
 import java.io.IOException;
-import java.io.UncheckedIOException;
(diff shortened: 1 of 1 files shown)


JingsongLi (approved):
+1

PR #7199, opened 2026-02-03
[python] Data Evolution with_slice should use row id to do slice

JingsongLi:
### Purpose

Only RowId can be trusted for Data Evolution table, otherwise, multiple versions of RowId files will result in data errors

### DIFF
--- a/paimon-python/pypaimon/read/reader/iface/record_batch_reader.py
+++ b/paimon-python/pypaimon/read/reader/iface/record_batch_reader.py
@@ -126,3 +126,12 @@ class RowPositionRecordIterator(RecordIterator[tuple]):
 
     def __next__(self):
         return self.next()
+
+
+class EmptyRecordBatchReader(RecordBatchReader):
+
+    def read_arrow_batch(self) -> Optional[RecordBatch]:
+        return None
+
+    def close(self):
+        return None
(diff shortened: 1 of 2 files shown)


paimon-python/pypaimon/read/scanner/data_evolution_split_generator.py
```diff
@@ -171,76 +171,99 @@ def _build_split_from_pack_for_data_evolution(
...
         """
-        sliced_splits = []
-        file_end_pos = 0  # end row position of current file in all splits data
+        # Collect all Row ID ranges from files
+        list_ranges = []
+        for file_entries in partitioned_files.values():
+            for entry in file_entries:
+                first_row_id = entry.file.first_row_id
+                # Range is inclusive [from_, to], so use row_count - 1
+                list_ranges.append(Range(first_row_id, first_row_id + entry.file.row_count - 1))
```
XiaoHongbo-Hope, line 185: blob is included in the entry here.

JingsongLi: What does a blob have to do with it?

XiaoHongbo-Hope: > What does a blob have to do with it?

It should be ok, I misunderstood.

PR #7200, opened 2026-02-03
[spark] Support MergeInto on Paimon append-only table in spark V2 write

kerwin-zk:
### Purpose

Support MergeInto on Paimon append-only table in spark V2 write

### Tests

CI

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/rowops/PaimonCopyOnWriteScan.scala
+++ b/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/rowops/PaimonCopyOnWriteScan.scala
@@ -43,8 +43,8 @@ import scala.collection.mutable
  * Note: The [[pushedPartitionFilters]] and [[pushedDataFilters]] are intentionally set to empty
  * because file-level filtering is handled through Spark's runtime V2 filtering mechanism.
  *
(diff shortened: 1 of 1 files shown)


paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/rowops/PaimonCopyOnWriteScan.scala
```diff
@@ -51,10 +54,28 @@ case class PaimonCopyOnWriteScan(
...
+      // If filter() was never called, it means Spark decided not to use runtime filtering
+      // (e.g., FULL OUTER JOIN for NOT MATCHED BY SOURCE). Read all data.
+      loadAllSplits()
+    }
+    dataSplits.asInstanceOf[Array[Split]]
+  }
 
   var dataSplits: Array[DataSplit] = Array()
 
+  private def loadAllSplits(): Unit = {
```
YannByron: Extract the initialization  of `snapshotReader` from `loadAllSplits` and `filter` methods, treat it as a member variable, and then define `dataSplits` as a lazy variable. The `filter` method just applies `SnapshotReader#withDataFileNameFilter` if it's called.

kerwin-zk: @YannByron SnapshotReader is not serializable, it cannot be declared as a member variable.

Issue #7011, opened 2026-01-12
[Feature] Introduce a new vector data type


ColdL:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Motivation

[PIP-40](https://cwiki.apache.org/confluence/display/PAIMON/PIP-40%3A+Introduce+a+new+Vector+data+type)

### Solution

As discussed in [PIP-40](https://cwiki.apache.org/confluence/display/PAIMON/PIP-40%3A+Introduce+a+new+Vector+data+type), we propose introducing a dedicated vector data type in Paimon to better support storage and retrieval of vector data for AI workloads.

PIP-40 can be roughly split into two parts:
(1) introducing the vector data type itself;
(2) allowing users to specify the file format for vector data, to further optimize storage/access efficiency in mixed workloads.

For Part (1), a basic implementation is already available and includes:
 - Introducing a new vector type. To avoid confusion with the existing term "Vector" in the codebase, the new type is named VecType.java.
 - Providing a ColumnVector implementation for the vector type, with support in the paimon-arrow module, so Arrow-related file formats (e.g., Lance) can map FixedSizeList to VecType.
 - Adding Flink-side compatibility: via configuration, a Flink Array can be stored as Paimon VecType.

For Part (2) (specifying the file format for vector data), work is still in progress.

- [x] Add a new DataType extension to represent the vector type.
- [x] Map vector type to arrow FixedSizeList, applying to arrow-based file format.
- [x] Provide compatibility at the Flink connector layer, enabling vector type read/write via Flink SQL.
- [x] Add relevant tests, including an end-to-end (E2E) test.
- [x] Support specifying the file format for the vector store.

Although the code is still in a draft state, it changes some basic interfaces (e.g., DataGetters), thus I'd like to discuss it early. Any comments on this @JingsongLi 

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #7204, opened 2026-02-04
add vector type definition and basic APIs / Implements

ColdL:
### Purpose

Linked issue: update #7011


paimon-api/src/main/java/org/apache/paimon/types/VectorType.java
```diff
@@ -0,0 +1,180 @@
...
+
+    public static final String FORMAT = "VECTOR<%s, %d>";
+
+    private final DataType elementType;
+
+    private final int length;
+
+    public VectorType(boolean isNullable, int length, DataType elementType) {
+        super(isNullable, DataTypeRoot.VECTOR);
+        // TODO: should we support nullable for vector type?
```
JingsongLi: Can we remove nullable for the first version?

ColdL: DONE

PR #7205, opened 2026-02-04
[python/hotfix] fix data evolution merge batch size mismatch issue

XiaoHongbo-Hope:
### Purpose/Problem

Reader may return batches of different sizes (e.g. 4096 or 6865). DataEvolutionMergeReader used to merge batch-by-batch and assumed all inner readers return the same number of rows in each round, so ValueError raised `ValueError: Arrays were not all the same length: 6865 vs 4096`
**

Linked issue: close #xxx

### DIFF
--- a/paimon-python/pypaimon/read/reader/concat_batch_reader.py
+++ b/paimon-python/pypaimon/read/reader/concat_batch_reader.py
@@ -17,13 +17,15 @@
 ################################################################################
 
 import collections
-from typing import Callable, List, Optional, Tuple
+from typing import Callable, List, Optional
 
 import pyarrow as pa
 import pyarrow.dataset as ds
 from pyarrow import RecordBatch
 
 from pypaimon.read.reader.iface.record_batch_reader import RecordBatchReader
+
+_MIN_BATCH_SIZE_TO_REFILL = 1024
 
 
 class ConcatBatchReader(RecordBatchReader):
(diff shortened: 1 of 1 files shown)


paimon-python/pypaimon/read/reader/concat_batch_reader.py
```diff
@@ -162,27 +162,56 @@ def __init__(
...
-                if batch is None:
-                    # all readers are aligned, as long as one returns null, the others will also have no data
-                    return None
-                batches[i] = batch
-        # Assemble record batches from batches based on row_offsets and field_offsets
+                if self._buffers[i] is not None:
+                    batch, offset = self._buffers[i]
+                    self._buffers[i] = None
+                    remainder = batch.slice(offset, batch.num_rows - offset)
+                    new_batch = reader.read_arrow_batch()
```
JingsongLi: Why we need to read a new batch if there is a batch? Maybe just check min batch size (for example, 1024), and read a new batch?

paimon-python/pypaimon/read/reader/concat_batch_reader.py
```diff
@@ -162,27 +162,56 @@ def __init__(
         self.field_offsets = field_offsets
         self.readers = readers
         self.schema = schema
+        self._buffers: List[Optional[Tuple[RecordBatch, int]]] = [None] * len(readers)
```
JingsongLi: Why need to store offset? Just store `batch.slice`?

PR #7207, opened 2026-02-04
[python] fix data evolution with_slice out-of-bounds: return empty rows instead of all data

XiaoHongbo-Hope:
### Purpose

Linked issue: close #xxx


paimon-python/pypaimon/read/scanner/data_evolution_split_generator.py
```diff
@@ -72,6 +72,8 @@ def sort_key(manifest_entry: ManifestEntry) -> tuple:
             if slice_row_ranges:
```
JingsongLi: Just remove this if.

PR #7208, opened 2026-02-04
[core] Disable async write when blob type found

leaves12138:
### Purpose

As title, blob write stuck in io mostly, so there is no need to open async write


paimon-core/src/main/java/org/apache/paimon/append/AppendOnlyWriter.java
```diff
@@ -318,7 +318,8 @@ private RollingFileWriter<InternalRow, DataFileMeta> createRollingRowWriter() {
                     statsCollectorFactories,
                     fileIndexOptions,
                     FileSource.APPEND,
-                    asyncFileWrite,
+                    // blob write does not need async write
```
JingsongLi: Write more...

leaves12138: OK

PR #7211, opened 2026-02-04
[python] Fix filter on data evolution table not working issue

XiaoHongbo-Hope:
### Purpose/Problem

Filter not working on data evolution read: when a predicate is provided, all rows are returned.

### DIFF
--- a/paimon-python/pypaimon/read/reader/filter_record_batch_reader.py
+++ b/paimon-python/pypaimon/read/reader/filter_record_batch_reader.py
@@ -17,16 +17,14 @@
 ###############################################################################
 
 import logging
-from typing import List, Optional, Tuple
+from typing import List, Optional
 
 import pyarrow as pa
-import pyarrow.compute as pc
 import pyarrow.dataset as ds
 
 from pypaimon.common.predicate import Predicate
 from pypaimon.read.reader.iface.record_batch_reader import RecordBatchReader
 from pypaimon.schema.data_types import DataField
-from pypaimon.table.row.offset_row import OffsetRow
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


paimon-python/pypaimon/read/reader/filter_record_batch_reader.py
```diff
@@ -0,0 +1,143 @@
...
+            return None
+        col = batch.column(self.predicate.field)
+        mask = pc.is_null(col) if self.predicate.method == 'isNull' else pc.is_valid(col)
+        return batch.filter(mask)
+
+    def _filter_batch(self, batch: pa.RecordBatch) -> Optional[pa.RecordBatch]:
+        simple_null = self._filter_batch_simple_null(batch)
+        if simple_null is not None:
+            return simple_null
+        if not self.predicate.has_null_check():
```
JingsongLi: We can just use `Predicate.to_arrow`, it is same to other table modes.

paimon-python/pypaimon/read/reader/filter_record_batch_reader.py
```diff
@@ -0,0 +1,143 @@
...
+                    return batches[0]
+                concat_batches = getattr(pa, "concat_batches", None)
+                if concat_batches is not None:
+                    return concat_batches(batches)
+                return pa.RecordBatch.from_arrays(
+                    [result.column(i) for i in range(result.num_columns)],
+                    schema=result.schema,
+                )
+            except (TypeError, ValueError, pa.ArrowInvalid) as e:
+                logger.debug(
```
JingsongLi: What exception here?

paimon-python/pypaimon/read/scanner/file_scanner.py
JingsongLi: We can refactor this, this if should be in `is_primary_key_table`.

paimon-python/pypaimon/read/scanner/file_scanner.py
```diff
@@ -434,7 +434,7 @@ def _filter_manifest_entry(self, entry: ManifestEntry) -> bool:
         else:
             if not self.predicate:
                 return True
-            if self.predicate_for_stats is None:
+            if self.predicate_for_stats is None or self.data_evolution:
```
JingsongLi: Create a separate `if`. And we should add comments to this if, explain why there is no filtering done here.

paimon-python/pypaimon/read/split_read.py
```diff
@@ -449,6 +451,10 @@ def __init__(
             actual_split = split.data_split()
         super().__init__(table, predicate, read_type, actual_split, row_tracking_enabled)
 
+    def _push_down_predicate(self) -> Optional[Predicate]:
+        # Do not push predicate to file readers;
```
JingsongLi: Detailed comments, why not push predicate.

paimon-python/pypaimon/read/split_read.py
```diff
@@ -467,7 +473,21 @@ def create_reader(self) -> RecordReader:
                     lambda files=need_merge_files: self._create_union_reader(files)
                 )
 
-        return ConcatBatchReader(suppliers)
+        merge_reader = ConcatBatchReader(suppliers)
+        if self.predicate is not None:
+            # Only apply filter when all predicate columns are in read_type (e.g. projected schema).
```
JingsongLi: What we are returning here is complete row, right? So this check should be applicable to all table modes? That shouldn't be added here, it should be verified in `SplitRead.init`.

paimon-python/pypaimon/read/reader/filter_record_batch_reader.py
XiaoHongbo-Hope: > We can just use `Predicate.to_arrow`, it is same to other table modes.

Updated

paimon-python/pypaimon/read/reader/filter_record_batch_reader.py
XiaoHongbo-Hope: > What exception here?

Used Predicate.to_arrow. This code is removed

paimon-python/pypaimon/read/scanner/file_scanner.py
XiaoHongbo-Hope: > We can refactor this, this if should be in `is_primary_key_table`.

Updated

paimon-python/pypaimon/read/scanner/file_scanner.py
XiaoHongbo-Hope: > Create a separate `if`. And we should add comments to this if, explain why there is no filtering done here.

Updated.

paimon-python/pypaimon/read/split_read.py
XiaoHongbo-Hope: > Detailed comments, why not push predicate.

Added

paimon-python/pypaimon/read/split_read.py
XiaoHongbo-Hope: > What we are returning here is complete row, right? So this check should be applicable to all table modes? That shouldn't be added here, it should be verified in `SplitRead.init`.

Updated

PR #7213, opened 2026-02-05
[core] Clean up constructor for FileStoreCommitImpl

tsreaper:
### Purpose

Many arguments in constructor of `FileStoreCommitImpl` are extracted from `options`. This PR cleans up these arguments.

### Tests

This is a code refactor, no additional tests needed.

### API and Format

No format changes.

### Documentation

No new feature.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
@@ -54,7 +54,6 @@ import org.apache.paimon.operation.commit.StrictModeChecker;
 import org.apache.paimon.operation.commit.SuccessCommitResult;
 import org.apache.paimon.operation.metrics.CommitMetrics;
 import org.apache.paimon.operation.metrics.CommitStats;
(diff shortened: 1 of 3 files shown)


paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
```diff
@@ -236,9 +199,10 @@ public FileStoreCommitImpl(
         this.commitMetrics = null;
         this.statsFileHandler = statsFileHandler;
         this.bucketMode = bucketMode;
-        this.rowTrackingEnabled = rowTrackingEnabled;
-        this.discardDuplicateFiles = discardDuplicateFiles;
-        this.strictModeChecker = strictModeChecker;
+        this.strictModeChecker =
+                options.commitStrictModeLastSafeSnapshot()
+                        .map(id -> new StrictModeChecker(snapshotManager, commitUser, scan, id))
```
JingsongLi: Should not use scan. This may lead to parameter confusion.

JingsongLi (approved):
+1

PR #7217, opened 2026-02-05
[core] BucketedAppendCompactManager should use compactionFileSize instead of targetFileSize

yuzelin:
### Purpose

See #1680


leaves12138 (approved):
+1

JingsongLi (changes requested):
Please add test for this.

paimon-core/src/main/java/org/apache/paimon/append/BucketedAppendCompactManager.java
```diff
@@ -214,7 +214,7 @@ Optional<List<DataFileMeta>> pickCompactBefore() {
             fileNum++;
             if (fileNum >= minFileNum) {
                 return Optional.of(candidates);
-            } else if (totalFileSize >= targetFileSize) {
+            } else if (totalFileSize >= compactionFileSize) {
```
JingsongLi: See `AppendCompactCoordinator`, there is another magic...

JingsongLi (approved):
+1

PR #7218, opened 2026-02-05
[core][filesystems] Add conditional write support for S3 and Azure ABFS

### LINKED ISSUE #6563: [Feature] S3-based lock to mitigate snapshot conflicts
### Search before asking

### Motivation

Current concurrency controls in Paimon rely on an external hive or jdbc metastore to provide locks to mitigate the risk of [snapshot conflicts](https://paimon.apache.org/docs/master/concepts/concurrency-control/).

We would like to run multiple concurrent writers, but without needing to also maintain a metastore service.

Taking inspiration form Hudi, we would like to propose a similar implementation that uses conditional-put semantics, and so the lock can be based on the storage system.

**References**:
- Hudi
  - [AWS S3-based lock using conditional requests](https://hudi.apache.org/docs/next/concurrency_control/#storage-with-conditional-writes-based)
  - [hudi.client.transaction.lock.StorageBasedLockProvider](https://github.com/apache/hudi/blob/f5f0ef6549fedf93863526a2110fe262a3460432/hudi-client/hudi-client-common/src/main/java/org/apache/hudi/client/transaction/lock/StorageBasedLockProvider.java)
  - [hudi.aws.transaction.lock.S3StorageLockClient](https://github.com/apache/hudi/blob/master/hudi-aws/src/main/java/org/apache/hudi/aws/transaction/lock/S3StorageLockClient.java)
  - [hudi.gcp.transaction.lock.GCSStorageLockClient](https://github.com/apache/hudi/blob/master/hudi-gcp/src/main/java/org/apache/hudi/gcp/transaction/lock/GCSStorageLockClient.java)
- AWS S3
  - [S3 If-None-Match header](https://docs.aws.amazon.com/AmazonS3/latest/userguide/conditional-writes.html)
- GCP Storage
    - [Conditional write example](https://docs.cloud.google.com/storage/docs/uploading-objects#storage-upload-object-client-libraries)
    - [Storage.BlobWriteOption.doesNotExist()](https://docs.cloud.google.com/java/docs/reference/google-cloud-storage/latest/com.google.cloud.storage.Storage.BlobWriteOption#com_google_cloud_storage_Storage_BlobWriteOption_doesNotExist__)
    - [Request preconditions](https://docs.cloud.google.com/storage/docs/request-preconditions)
    - [Storage JSON api - objects.insert](https://docs.cloud.google.com/storage/docs/json_api/v1/objects/insert)

### Solution

Intended implementation approach:

- New classes in [org.apache.paimon.s3](https://github.com/apache/paimon/tree/master/paimon-filesystems/paimon-s3-impl/src/main/java/org/apache/paimon/s3) and [org.apache.paimon.gs](https://github.com/apache/paimon/tree/master/paimon-filesystems/paimon-gs-impl/src/main/java/org/apache/paimon/gs)
  - Use existing methods to extract hadoop configurations (particularly for auth)
  - Construct new AWS / GCP storage clients, not relying on Hadoop filesystem (similar to implementation in Hudi)
  - Note: Hadoop now has some support for conditional-write semantics, e.g. [HADOOP-19256](https://issues.apache.org/jira/browse/HADOOP-19256), however this is available in Hadoop 3.4.2 onwards, which is not well supported in Flink/Paimon yet
  - New classes implement  [CatalogLockFactory](https://github.com/apache/paimon/blob/master/paimon-core/src/main/java/org/apache/paimon/catalog/CatalogLockFactory.java#L26)
- Extend [CatalogEnvironment](https://github.com/apache/paimon/blob/master/paimon-core/src/main/java/org/apache/paimon/table/CatalogEnvironment.java) to support adding a lock factory on top of `CatalogEnvironment::empty`, e.g. new method `.withLockFactory(CatalogLockFactory lockFactory)`

To use the lock:
```kotlin
import org.apache.paimon.flink.FlinkFileIO
import org.apache.paimon.fs.Path
import org.apache.paimon.options.Options
import org.apache.paimon.schema.SchemaManager
import org.apache.paimon.table.CatalogEnvironment
import org.apache.paimon.table.FileStoreTableFactory

// assuming an existing table for simplicity
val tablePath = Path("s3a://bucket/prefix")
val fileIO = FlinkFileIO(tablePath)
val writerOptions = Options()
val schema = SchemaManager(fileIO, tablePath).latest().get()
val emptyCatalogWithLock = 
    CatalogEnvironment
    .empty()
    .withLockFactory(S3LockFactory()) // NEW

val table = FileStoreTableFactory.create(fileIO, tablePath, schema, writerOptions, emptyCatalogWithLock)
```

**Guidance needed**
- For the purpose of commit locks, the option `lock.type` does not appear to be used when we construct the table via `FileStoreTableFactory.create(...)`.
- Is it necessary to extend catalogs and the catalog option `lock.type` to also support s3/gs locks?

### Anything else?

_No response_

tub:
### Purpose

Implements native conditional writes using Hadoop 3.4+ API (fs.createFile().overwrite(false).build()) which leverages `If-None-Match: *` for both AWS S3 and Azure ABFS.

Changes:
- Add FileIO.supportsConditionalWrite() and tryToWriteAtomicIfAbsent()
- S3FileIO: Override to use native conditional writes
- AzureFileIO: Override to use native conditional writes
- RenamingSnapshotCommit: Use conditional writes when available, eliminating need for metastore lock on S3/Azure

This allows S3 and Azure users to run Paimon without lock.enabled=true for safe concurrent commits.

Linked issue: close #6563
Builds on: #7187

### Tests

- paimon-common/src/test/java/org/apache/paimon/fs/FileIOTest.java (updated)
- paimon-core/src/test/java/org/apache/paimon/catalog/RenamingSnapshotCommitTest.java (new)

### API and Format

Not directly - changes default safety guarantees.

### Documentation
Added documentation covering both AWS and Azure,

### DIFF
--- a/docs/content/concepts/concurrency-control.md
+++ b/docs/content/concepts/concurrency-control.md
@@ -44,8 +44,21 @@ Paimon's snapshot ID is unique, so as long as the job writes its snapshot file t
 Paimon uses the file system's renaming mechanism to commit snapshots, which is secure for HDFS as it ensures
 transactional and atomic renaming.
 
-But for object storage such as OSS and S3, their `'RENAME'` does not have atomic semantic. We need to configure Hive or
-jdbc metastore and enable `'lock.enabled'` option for the catalog. Otherwise, there may be a chance of losing the snapshot.
+For object storage, atomic commits are handled differently depending on the storage system:
+
+### S3 and Azure ABFS (Conditional Writes)
+
+S3 and Azure ABFS support native conditional writes. When using `paimon-s3` or `paimon-azure`, Paimon automatically
+uses conditional writes for snapshot commits, providing atomic write-if-absent semantics without requiring external
+locking.
+
+When using S3 or Azure ABFS, you do **not** need to configure `'lock.enabled'` for safe concurrent commits.
+
+### Other Object Stores (OSS, GCS, OBS, etc.)
+
+Paimon currently does not support conditional writes for other object storage systems such as Aliyun OSS, Google
+Cloud Storage, and Huawei OBS. You need to configure Hive or JDBC metastore and enable `'lock.enabled'` option for
+the catalog. Otherwise, there may be a chance of losing the snapshot.
 
 ## Files conflict
 
--- a/docs/content/maintenance/filesystems.md
+++ b/docs/content/maintenance/filesystems.md
@@ -389,6 +389,12 @@ Please refer to [Trino S3](https://trino.io/docs/current/object-storage/file-sys
 
 {{< /tabs >}}
 
+### Conditional Writes
+
+S3 supports native conditional writes, which Paimon uses automatically for atomic snapshot commits. This eliminates
+the need to configure `'lock.enabled'` when using S3 for safe concurrent writes. See
+[Concurrency Control]({{< ref "concepts/concurrency-control" >}}) for more details.
+
 ### S3 Compliant Object Stores
 
 The S3 Filesystem also support using S3 compliant object stores such as MinIO, Tencent's COS and IBM’s Cloud Object
@@ -509,6 +515,12 @@ spark-sql \
 {{< /tab >}}
 
 {{< /tabs >}}
+
+### Conditional Writes
+
+Azure ABFS supports native conditional writes, which Paimon uses automatically for atomic snapshot commits. This
+eliminates the need to configure `'lock.enabled'` when using Azure ABFS for safe concurrent writes. See
+[Concurrency Control]({{< ref "concepts/concurrency-control" >}}) for more details.
 
 ## OBS
 


paimon-core/src/main/java/org/apache/paimon/catalog/RenamingSnapshotCommit.java
```diff
@@ -54,6 +55,17 @@ public boolean commit(Snapshot snapshot, String branch, List<PartitionStatistics
                         ? snapshotManager.snapshotPath(snapshot.id())
                         : snapshotManager.copyWithBranch(branch).snapshotPath(snapshot.id());
 
+        // Use native conditional writes if supported
```
tub, line 58: I'm not 100% sure if this is the right approach, or if we should rename `RenamingSnapshotCommit‎` given that it doesn't _always_ use rename now.

junmuz: We also have IcebergCommitCallbacks [here](https://github.com/apache/paimon/blob/master/paimon-core/src/main/java/org/apache/paimon/iceberg/IcebergCommitCallback.java#L385), so we need to support it there as well.

junmuz: I think overriding the method `tryToWriteAtomic` would be better as it is used at different places. Perhaps we can leverage the same property `fs.s3a.create.conditional.enabled` (https://github.com/apache/hadoop/blob/8222ac0f546911b387b6a141e06dd8bf5306d565/hadoop-tools/hadoop-aws/src/site/markdown/tools/hadoop-aws/troubleshooting_s3a.md) for making a decision whether to use fallback.

PR #7221, opened 2026-02-05
[core] Fix LookupMergeFunction to use sequence.field for picking high level records

### LINKED ISSUE #7220: [Bug] LookupMergeFunction.pickHighLevel() ignores sequence.field when selecting high level record
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found no similar issues.

### Paimon version

master (latest)

### Compute Engine

None

### Minimal reproduce step

When using `changelog-producer = lookup` with `sequence.field` configured, `LookupMergeFunction.pickHighLevel()` may select the wrong "old" record when out-of-order data arrives.

**Configuration:**
```sql
CREATE TABLE test (
    id INT PRIMARY KEY NOT ENFORCED,
    value INT,
    update_time BIGINT
) WITH (
    'changelog-producer' = 'lookup',
    'sequence.field' = 'update_time'
);
```

**Scenario:**
```
Initial state after compaction:
  L1: (id=1, value=100, update_time=7)
  L2: (id=1, value=200, update_time=8)  ← Actually newer!

New out-of-order data arrives at L0:
  L0: (id=1, value=50, update_time=6)   ← Old data arriving late
```

**Expected behavior:**
- `pickHighLevel()` should select L2 (update_time=8) as the "latest" high-level record
- Result should reflect the record with highest sequence value

**Actual behavior:**
- `pickHighLevel()` selects L1 (level 1 < level 2) ignoring sequence.field
- Wrong changelog is generated

`LookupMergeFunction.pickHighLevel()` only compares level numbers, ignoring `sequence.field`:

```java
// LookupMergeFunction.java:88 - Current behavior
if (highLevel == null || kv.level() < highLevel.level()) {
    highLevel = kv;  // Always picks lowest level, ignores sequence
}
```

**Reproducible scenario:**
```java
// When candidates contain:
// L1: (key=1, sequence=7)  <- level 1
// L2: (key=1, sequence=8)  <- level 2, but higher sequence (newer!)

// pickHighLevel() returns L1 (because level 1 < 2)
// But should return L2 (because sequence 8 > 7)
```

It should use `sequence.field` comparator when configured, similar to how `SortMergeReaderWithMinHeap` correctly handles it at line 61-67.

### Anything else?

This issue only affects `changelog-producer = lookup` scenario. Normal queries (Batch/Streaming Scan) and Lookup Join are not affected.

I'm working on a fix and will submit a PR shortly. The PR includes a complete unit test to reproduce this issue.

Liulietong:
### Purpose

Fix #7220

When `sequence.field` is configured, `LookupMergeFunction.pickHighLevel()` should select the record with the highest sequence value instead of the lowest level number.

### Problem

Previously, `pickHighLevel()` only compared level numbers:
```java
if (highLevel == null || kv.level() < highLevel.level()) {
    highLevel = kv;
}
```

This could lead to incorrect results when out-of-order data arrives:
- L1 has sequence=7 (older)
- L2 has sequence=8 (newer)
- L0 has sequence=6 (oldest, out-of-order arrival)

The old logic would pick L1 (level 1 < level 2), but the correct behavior should pick L2 (sequence 8 > 7).

### Changes

1. Add `sequenceComparator` field to `LookupMergeFunction`
2. Modify `pickHighLevel()` to use sequence comparator when available
3. Modify `getResult()` to sort records by sequence before adding to merge function
4. Only set `sequenceComparator` when user-defined sequence field is configured, preserving original behavior when `sequence.field` is not set

### Tests

Added 3 test cases:
- `testSequenceFieldWithMultipleLevels` - verifies sequence.field is used correctly
- `testWithoutSequenceFieldPreservesOriginalBehavior` - verifies backward compatibility  
- `testSequenceFieldWithDescendingSortOrder` - verifies descending sort order works

### Impact

Only affects `changelog-producer = lookup` with `sequence.field` configured. No impact on:
- Normal queries (Batch/Streaming Scan)
- Lookup Join
- Tables without `sequence.field`

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/LookupChangelogMergeFunctionWrapper.java
+++ b/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/LookupChangelogMergeFunctionWrapper.java
@@ -88,6 +88,14 @@ public class LookupChangelogMergeFunctionWrapper<T>
         this.lookupStrategy = lookupStrategy;
         this.deletionVectorsMaintainer = deletionVectorsMaintainer;
         this.comparator = createSequenceComparator(userDefinedSeqComparator);
+        // Only set sequence comparator when user-defined sequence field is configured
+        // to preserve original behavior (pick by level) when sequence.field is not set.
+        // Note: We use the same comparator for both insertInto and pickHighLevel.
+        // The comparator's semantics (ascending/descending) are already handled correctly
+        // by UserDefinedSeqComparator based on sequence.field.sort-order configuration.
+        if (userDefinedSeqComparator != null) {
+            this.mergeFunction.setSequenceComparator(this.comparator);
+        }
     }
 
     @Override
--- a/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/LookupMergeFunction.java
+++ b/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/LookupMergeFunction.java
@@ -27,7 +27,9 @@ import org.apache.paimon.utils.CloseableIterator;
 
 import javax.annotation.Nullable;
 
+import java.util.ArrayList;
 import java.util.Comparator;
+import java.util.List;
 
 /**
  * A {@link MergeFunction} for lookup, this wrapper only considers the latest high level record,
@@ -41,6 +43,7 @@ public class LookupMergeFunction implements MergeFunction<KeyValue> {
     private final KeyValueBuffer candidates;
     private boolean containLevel0;
     private InternalRow currentKey;
+    @Nullable private Comparator<KeyValue> sequenceComparator;
 
     public LookupMergeFunction(
             MergeFunction<KeyValue> mergeFunction,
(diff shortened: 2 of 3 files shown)


tub:
+1 - we've seen this behaviour too, would be great to get it fixed!

JingsongLi (commented):
Important correctness fix for `LookupMergeFunction` with `sequence.field`.

**The bug is clear:** Without this fix, `pickHighLevel()` selects by minimum level number, ignoring sequence values. When out-of-order data arrives at different levels, the wrong record (older sequence) can be picked as the "latest."

**Review of the fix:**

1. **`setSequenceComparator`**: Only set when `userDefinedSeqComparator != null` — preserves original behavior for tables without `sequence.field`. Good defensive design.

2. **`pickHighLevel()` logic**:
   - `sequenceComparator != null` → pick by highest sequence value
   - Otherwise → pick by lowest level (original behavior)
   - This is correct. The sequence comparator already handles ascending/descending via `UserDefinedSeqComparator`.

3. **`getResult()` sorting**: Sorting `toMerge` by `sequenceComparator` ensures the highest sequence record is added *last* to the merge function. Since `DeduplicateMergeFunction` keeps the last record, this produces the correct result. **But** — this assumes the merge function's semantics are "last wins." If another merge function is used (e.g., partial-update), does adding in sequence order still produce correct results?

4. **Collecting to ArrayList**: `toMerge.add(kv)` collects into an ArrayList then sorts and iterates. This introduces an allocation per getResult() call. For most workloads this is negligible, but for high-throughput compaction with many keys, it could add up. Consider: could the sort be avoided if there are only 2 elements (just compare and swap order)?

5. **Test coverage**: 3 tests covering:
   - sequence.field with multiple levels (the bug scenario) ✓
   - Without sequence.field (backward compat) ✓
   - Descending sort order ✓

Good fix. Please clarify point 3 — does sorting work correctly with all merge function types that support sequence.field?

JingsongLi:
+1

PR #7223, opened 2026-02-05
Changes for supporting Atomic writes with Hadoop 3.4+

junmuz:
### Purpose

Linked issue: close #xxx

### DIFF
--- a/paimon-filesystems/paimon-azure-impl/src/main/java/org/apache/paimon/azure/AzureFileIO.java
+++ b/paimon-filesystems/paimon-azure-impl/src/main/java/org/apache/paimon/azure/AzureFileIO.java
@@ -20,9 +20,13 @@ package org.apache.paimon.azure;
 
 import org.apache.paimon.catalog.CatalogContext;
 import org.apache.paimon.fs.FileIO;
+import org.apache.paimon.fs.Path;
 import org.apache.paimon.options.Options;
 
 import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FSDataOutputStream;
+import org.apache.hadoop.fs.FSDataOutputStreamBuilder;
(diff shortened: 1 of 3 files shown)


JingsongLi (commented):
This PR adds atomic write support for S3 and Azure using Hadoop 3.4+ `FSDataOutputStreamBuilder` API.

**Review:**

1. **API compatibility**: `fs.createFile(path).must("fs.option.create.conditional.overwrite", true)` requires Hadoop 3.4+. This is a **runtime dependency** — if Paimon is used with Hadoop 3.3.x, the `must()` call will throw `UnsupportedOperationException` or similar. How is this handled?
   - Is there a Hadoop version check before using this path?
   - Is `tryToWriteAtomic` only called when explicitly enabled, or is it a fallback?

2. **PR description is empty** — no Purpose, Tests, or linked issue. This makes it difficult to evaluate:
   - When is `tryToWriteAtomic` called? What replaces the current rename-based atomicity?
   - What's the performance impact? Conditional writes on S3/Azure may have different latency characteristics than rename.
   - Is there a fallback for object stores that don't support conditional writes?

3. **Visibility change in `HadoopCompliantFileIO`**: `path()` and `getFileSystem()` changed from `private` to `protected`. This is a minor API surface expansion — acceptable for subclass reuse, but document that these are now extension points.

4. **No tests**: The PR has zero test additions. This is a significant functional change (atomic write semantics) that needs:
   - Unit tests with mocked FileSystem
   - Integration tests with real S3/Azure (or localstack)
   - Tests for the case where the file already exists (returns false)
   - Tests for Hadoop < 3.4 (graceful fallback or clear error)

5. **Code duplication**: The `tryToWriteAtomic` implementation is nearly identical in `S3FileIO` and `AzureFileIO`. Consider extracting to `HadoopCompliantFileIO`.

Please fill in the PR description and add tests before this can be reviewed in depth.

PR #7227, opened 2026-02-06
[tag] Force creating auto-tag even no data when calling trigger_tag_automatic_creation

JackeyLee007:
### Purpose

Make sure the auto-tag exists as expected even when no data is provided.

The former version of procedure `trigger_tag_automatic_creation` creates the tag only if it finds the right snapshot which is after `tag.creation-delay`. This is not always okay.

As we know, the auto-tag would be created within the Flink CDC job. But it would fail if the Flink job:
* crashes accidentally
* fails to create checkpoint
* is delayed in creating checkpoint

To meet the auto-tag expectation, I enhanced the `trigger_tag_automatic_creation` procedure by adding a new `force` parameter, defaulting to false. When it's true, and the auto-tag still does not exist after the first attempt at auto-creation, we make an empty commit to force the creation of the snapshot which could be used to create the auto-tag during the second run.

### Tests

Flink UT `TriggerTagAutomaticCreationProcedureITCase` and Spark UT `TriggerTagAutomaticCreationProcedureTest` were adapted.

### API and Format

NO

### Documentation

Doc provied for the new parameter.

### DIFF
--- a/docs/content/flink/procedures.md
+++ b/docs/content/flink/procedures.md
@@ -272,14 +272,21 @@ All available procedures are listed below.
    <tr>
       <td>trigger_tag_automatic_creation</td>
       <td>
+         -- Use named argument
+         CALL [catalog.]sys.trigger_tag_automatic_creation(`table` => 'identifier')
+         CALL [catalog.]sys.trigger_tag_automatic_creation(`table` => 'identifier', force => false)<br/><br/>
+         -- Use indexed argument
          CALL [catalog.]sys.trigger_tag_automatic_creation('identifier')
+         CALL [catalog.]sys.trigger_tag_automatic_creation('identifier', false)
       </td>
       <td>
          Trigger the tag automatic creation. Arguments:
             <li>table: the target table identifier. Cannot be empty.</li>
+            <li>force: force creating the auto-tag when it's after tag.creation-delay even no data exits. Default false.</li>
       </td>
       <td>
-         CALL sys.trigger_tag_automatic_creation(table => 'default.T')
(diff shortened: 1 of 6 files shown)


JingsongLi (commented):
Good feature for operational reliability. The use case (Flink CDC job crashes/delays, auto-tag not created) is real.

**Review:**

1. **Empty commit approach**: Using `ManifestCommittable(Long.MAX_VALUE, null, Collections.emptyList())` with `ignoreEmptyCommit(false)` to force a snapshot is a reasonable workaround. But using `Long.MAX_VALUE` as the checkpoint ID is concerning — could this conflict with real checkpoint IDs or cause ordering issues in snapshot metadata?

2. **Two-attempt pattern**: First try without force, then if tag doesn't exist and `forceCreatingSnapshot()` returns true, make an empty commit and try again. This is clear and handles the case where the tag already exists (no unnecessary empty commits).

3. **`isAutoTagExits` typo**: The method name should be `isAutoTagExists` (missing 's').

4. **Time sensitivity**: `LocalDateTime.now()` is used to check if the tag should exist. But between `normalizeToPreviousTag(now)` and the actual tag creation, a few milliseconds pass. For tag periods like "daily" this is fine, but for very short periods (e.g., "hourly" near the boundary), there could be a race. Probably acceptable for the stated use case.

5. **Test change**: The test changes `tag.creation-delay` from `10 m` to `0 m` and uses `force => true`. This tests the force path directly — good. But should there also be a test for `force => false` (original behavior preserved)?

6. **Documentation**: Both Flink and Spark procedure docs updated. Good.

The approach is sound. Please fix the typo and consider the `Long.MAX_VALUE` checkpoint ID concern.

PR #7228, opened 2026-02-06
[core] Introduce supportsPartitionModification to Catalog

JingsongLi:
### Purpose

See `RESTCatalogTest.testPartitionExpire`, partition expire should use commit user prefix.

This PR fix it, and refactoring improves the semantic clarity of the code, and the name PartitionModification more accurately describes the responsibilities of the interface - specifically designed to handle partition modification operations, rather than general partition handling.

### DIFF
--- a/paimon-core/src/test/java/org/apache/paimon/rest/RESTCatalogTest.java
+++ b/paimon-core/src/test/java/org/apache/paimon/rest/RESTCatalogTest.java
@@ -1617,18 +1617,24 @@ public abstract class RESTCatalogTest extends CatalogTestBase {
 
     @Test
     void testPartitionExpire() throws Exception {
+        // create table
         Identifier identifier = Identifier.create("test_db", "test_partition_expire");
         Map<String, String> options = new HashMap<>();
         options.put(PARTITION_EXPIRATION_STRATEGY.key(), "update-time");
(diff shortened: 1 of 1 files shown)


paimon-core/src/test/java/org/apache/paimon/rest/RESTCatalogTest.java
```diff
@@ -1611,6 +1615,23 @@ public void testListPartitionsPagedWithMultiLevel() throws Exception {
         assertPagedPartitions(pagedPartitions, 1, partitionSpecs.get(1));
     }
 
+    @Test
+    void testPartitionExpire() throws Exception {
```
yuzelin, line 1619: Use following option for more accuracy test:
1. options.put(METASTORE_PARTITIONED_TABLE.key(), "true");
2. table = table.copy(Collections.singletonMap(COMMIT_USER_PREFIX.key(), "my_user"));

JingsongLi: I should run test before PR. It should be failed.

JingsongLi: Addressed.

PR #7231, opened 2026-02-06
[core][flink] supports reading blob descriptors from external storage

steFaiz:
### Purpose

BlobDescriptor is an important feature in writing, with which Flink/Spark/users do not have to load the entire blob into memory. Currently, input blob descriptors share a same `FileIO` configuration with Paimon, which means we cannot load some files from a different OSS endpoint.
This PR introduces a config prefix. With which user can specify FileIO related options for input blob descriptor. i.e. oss AK, SK and endpoints.

Linked issue: none

### Tests
For test, we mock different OSS endpoints by a IsolatedDirectoryFileIO, which is a LocalFileIO only permitting accessing the configured 'root directory'.

### API and Format
This PR do not modify any existing APIs.

### Documentation
Updated the Blob's documentation.

### DIFF
--- a/docs/content/append-table/blob.md
+++ b/docs/content/append-table/blob.md
@@ -124,7 +124,7 @@ For details about the blob file format structure, see [File Format - BLOB]({{< r
 
 Specifically, if the storage system of the input BlobDescriptor differs from that used by Paimon, 
 you can specify the storage configuration for the input blob descriptor using the prefix 
-`blob-descriptor.io.`. For example, if the source data is stored in a different OSS endpoint, 
+`blob-descriptor.`. For example, if the source data is stored in a different OSS endpoint, 
 you can configure it as below (using flink sql as an example):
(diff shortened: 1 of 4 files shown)


docs/content/append-table/blob.md
```diff
@@ -122,6 +122,24 @@ For details about the blob file format structure, see [File Format - BLOB]({{< r
 
 *Required for blob functionality to work correctly.
 
+Specifically, if the storage system of the input BlobDescriptor differs from that used by Paimon, 
+you can specify the storage configuration for the input blob descriptor using the prefix 
+`blob-descriptor.io.`. For example, if the source data is stored in a different OSS endpoint,
```
JingsongLi: just `blob-descriptor.`?

steFaiz: Thanks for your suggestion! I've modified this.

PR #7232, opened 2026-02-08
[python] Do not abort on commit failure

XiaoHongbo-Hope:
### Purpose/Problem
TableCommit._commit called abort() on any commit failure, which could delete committed data when the server had already committed but the client received an error (e.g. timeout, close failure).

Linked issue: close #xxx


JingsongLi:
+1 Thanks @XiaoHongbo-Hope

PR #7235, opened 2026-02-08
[core] Introduce 'visibility-callback.enabled'

JingsongLi:
### Purpose

This feature is particularly useful for write-only tables where data should be queryable immediately after insertion. The callback ensures that queries wait for the necessary compaction to complete, providing immediate visibility guarantees.

### Tests

`VisibilityCallbackTest.scala`

### API and Format

Configuration Options
- visibility-callback.enabled: Enable/disable the visibility callback
- visibility-callback.timeout: Maximum time to wait for compaction
- visibility-callback.check-interval: Polling interval when checking compaction status

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/metastore/VisibilityWaitCallback.java
+++ b/paimon-core/src/main/java/org/apache/paimon/metastore/VisibilityWaitCallback.java
@@ -24,6 +24,7 @@ import org.apache.paimon.data.BinaryRow;
 import org.apache.paimon.manifest.FileKind;
 import org.apache.paimon.manifest.ManifestCommittable;
 import org.apache.paimon.manifest.ManifestEntry;
+import org.apache.paimon.table.BucketMode;
 import org.apache.paimon.table.FileStoreTable;
 import org.apache.paimon.table.sink.BatchWriteBuilder;
 import org.apache.paimon.table.sink.CommitCallback;


docs/layouts/shortcodes/generated/core_configuration.html
```diff
@@ -1452,6 +1452,24 @@
...
+            <td><h5>visibility-callback.check-interval</h5></td>
+            <td style="word-wrap: break-word;">10 s</td>
+            <td>Duration</td>
+            <td>The interval for checking visibility when visibility-callback enabled.</td>
+        </tr>
+        <tr>
+            <td><h5>visibility-callback.enabled</h5></td>
+            <td style="word-wrap: break-word;">false</td>
+            <td>Boolean</td>
+            <td>Whether to enable the visibility wait callback that waits for compaction to complete after commit. This is useful for primary key tables with deletion vectors or postpone bucket mode to ensure data visibility.</td>
```
tsreaper: State that it is only used for batch mode or bounded stream.

paimon-core/src/main/java/org/apache/paimon/metastore/VisibilityWaitCallback.java
```diff
@@ -0,0 +1,147 @@
...
+    private boolean shouldBeTracked(ManifestEntry entry) {
+        if (!FileKind.ADD.equals(entry.kind())) {
+            return false;
+        }
+
+        if (deletionVectorsEnabled && entry.level() == 0) {
+            return true;
+        }
+
+        return entry.bucket() == -2;
```
tsreaper:
```suggestion
        return entry.bucket() == BucketMode.POSTPONE_BUCKET;
```

docs/layouts/shortcodes/generated/core_configuration.html
JingsongLi: Addressed

paimon-core/src/main/java/org/apache/paimon/table/sink/CommitCallback.java
```diff
@@ -40,13 +40,40 @@
  */
 public interface CommitCallback extends AutoCloseable {
 
-    void call(
+    default void call(Context context) {
+        call(context.baseFiles, context.deltaFiles, context.indexFiles, context.snapshot);
+    }
+
+    default void call(
```
tsreaper: This interface is not public. Maybe mark this method as deprecated and delete it after one version? Also why add `default`?

JingsongLi: Removed legacy method.

tsreaper (approved):

PR #7240, opened 2026-02-09
[core] Introduce vector-store for data-evolution table

ColdL:
### Purpose

Linked issue: update #7011

The goal of this PR is to optimize storage layout for vector scenarios in the Data Evolution table, specifically by storing vector type columns in specified file formats.

For example, scalar columns can be stored using Parquet format, while vector columns and columns that may require point lookups during vector search can be stored using file formats like Lance.

**1. Configuration**

This PR introduces three new configuration options:
   - `vector-field`: defines the column names for separate storage
   - `vector.file.format`: defines the file format
   - `vector.target-file-size`: specifies the file size threshold for rolling

**2. Storage Layout**

When this feature is enabled, a set of columns specified by `vector-field` will be stored separately in the file format specified by `vector.file.format`, marked by `.vector.` in the data file path.

**File Path Pattern:** `data-xxx-{count}.vector.{file-format}`

This design serves two purposes:
   - (1) The `.vector.` segment identifies these as separately stored column groups
   - (2) The trailing `.{file-format}` follows current conventions, using the file format as the suffix

The final storage layout might be:
   - data-xxx-0.parquet
   - data-xxx-1.blob
   - data-xxx-2.vector.lance
   - data-xxx-3.vector.lance
   - data-xxx-4.vector.lance

These `.vector.` files are associated with regular columns through Row-tracking / Data Evolution.

### DIFF
--- a/docs/content/append-table/vector.md
+++ b/docs/content/append-table/vector.md
@@ -1,0 +1,167 @@
+---
+title: "Vector Storage"
+weight: 7
+type: docs
+aliases:
+- /append-table/vector-storage.html
+---
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+# Vector Storage
+
+## Overview
+
+With the explosive growth of AI scenarios, vector storage has become increasingly important.
+
+Paimon provides optimized storage solutions specifically designed for vector data to meet the needs of various scenarios.
+
+## Vector Data Type
+
+Vector data comes in many types, among which dense vectors are the most commonly used. They are typically expressed as fixed-length, densely packed arrays, generally without `null` elements.
+
+Paimon supports defining columns of type `VECTOR<t, n>`, which represents a fixed-length, dense vector column, where:
+ - **`t`**: The element type of the vector. Currently supports seven primitive types: `BOOLEAN`, `TINYINT`, `SMALLINT`, `INT`, `BIGINT`, `FLOAT`, `DOUBLE`;
+ - **`n`**: The vector dimension, must be a positive integer not exceeding `2,147,483,647`;
+ - **`null constraint`**: `VECTOR` type supports defining `NOT NULL` or the default nullable. However, if a specific `VECTOR` value itself is not `null`, its elements are not allowed to be `null`.
+
+Compared to variable-length arrays, these features make dense vectors more concise in storage and memory representation, with benefits including:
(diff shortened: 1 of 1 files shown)


paimon-core/src/main/java/org/apache/paimon/append/dataevolution/DataEvolutionCompactCoordinator.java
```diff
@@ -203,6 +209,25 @@ List<DataEvolutionCompactTask> compactPlan(List<ManifestEntry> input) {
                             }
                         }
                     }
+                    if (false) {
```
leaves12138: if compactVector?

paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -3469,6 +3492,26 @@ public Duration visibilityCallbackCheckInterval() {
...
+    public String vectorStoreFileFormatString() {
+        return normalizeFileFormat(options.get(VECTOR_STORE_FORMAT));
+    }
+
+    public List<String> vectorStoreFieldNames() {
+        String vectorStoreFields = options.get(CoreOptions.VECTOR_STORE_FIELDS);
+        if (vectorStoreFields == null || vectorStoreFields.trim().isEmpty()) {
+            return new ArrayList<>();
+        } else {
+            return Arrays.asList(vectorStoreFields.split(","));
```
leaves12138: vectorStoreFields.split(",").trim()?

paimon-core/src/main/java/org/apache/paimon/operation/DataEvolutionSplitRead.java
```diff
@@ -467,7 +483,7 @@ public List<DataFileMeta> files() {
     }
 
     @VisibleForTesting
-    static class BlobBunch implements FieldBunch {
+    static class SplitBunch implements FieldBunch {
```
leaves12138: SpecialFieldBunch?

paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -2230,6 +2230,29 @@ public InlineElement getDescription() {
...
+                    .noDefaultValue()
+                    .withDescription("Specify the vector store file format.");
+
+    public static final ConfigOption<String> VECTOR_STORE_FIELDS =
+            key("vector-field")
+                    .stringType()
+                    .noDefaultValue()
+                    .withDescription("Specify the vector store fields.");
+
+    public static final ConfigOption<MemorySize> VECTOR_STORE_TARGET_FILE_SIZE =
```
leaves12138: How about these names? @JingsongLi

paimon-core/src/main/java/org/apache/paimon/append/dataevolution/DataEvolutionCompactCoordinator.java
ColdL: DONE

paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
ColdL: DONE

paimon-core/src/main/java/org/apache/paimon/operation/DataEvolutionSplitRead.java
ColdL: DONE

paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
ColdL: Fixed. Now the config names in code are consistent with the public configuration keys.

PR #7242, opened 2026-02-09
[core] fix btree index row range

steFaiz:
### Purpose
The row range is miscalculated by [first_row_id, first_row_id + row_count], should minus 1


paimon-core/src/main/java/org/apache/paimon/globalindex/btree/BTreeGlobalIndexBuilder.java
```diff
@@ -247,7 +247,7 @@ public static Range calcRowRange(List<DataSplit> dataSplits) {
             for (DataFileMeta file : dataSplit.dataFiles()) {
                 long firstRowId = file.nonNullFirstRowId();
                 start = Math.min(start, firstRowId);
-                end = Math.max(end, firstRowId + file.rowCount());
+                end = Math.max(end, firstRowId + file.rowCount() - 1);
```
JingsongLi, line 250: Can you add a case for this?

PR #7254, opened 2026-02-10
[core] Improve HttpClient error response handling

plusplusjiajia:
### Purpose
This PR improves error response handling in HttpClient by ensuring that error messages are always populated with meaningful information when the response body is available.

### DIFF
--- a/paimon-api/src/main/java/org/apache/paimon/rest/HttpClient.java
+++ b/paimon-api/src/main/java/org/apache/paimon/rest/HttpClient.java
@@ -131,29 +131,14 @@ public class HttpClient implements RESTClient {
                     response -> {
                         String responseBodyStr = RESTUtil.extractResponseBodyAsString(response);
                         if (!RESTUtil.isSuccessful(response)) {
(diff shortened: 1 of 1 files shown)


paimon-api/src/main/java/org/apache/paimon/rest/HttpClient.java
```diff
@@ -144,6 +144,16 @@ private <T extends RESTResponse> T exec(HttpUriRequestBase request, Class<T> res
                                                         : "response body is null",
                                                 response.getCode());
                             }
+                            if (error.getMessage() == null || error.getMessage().isEmpty()) {
+                                error =
+                                        new ErrorResponse(
```
JingsongLi: I believe you can do some refactoring. This code is not good to duplicate with the above code.

plusplusjiajia: > I believe you can do some refactoring. This code is not good to duplicate with the above code.

Thanks for the review, the duplicate code removed, please have a look, thanks!

PR #7255, opened 2026-02-10
[core] RowIdPredicateVisitor supports converting between statement

steFaiz:
### Purpose
After introducing Between leaf function, `RowIdPredicateVisitor` can extract row ranges from between predicate.

Linked issue: none

### Tests
Please see org.apache.paimon.predicate.RowIdPredicateVisitorTest for unit test

### API and Format
none

### Documentation
none


steFaiz:
Paimon-spark do not support converting SparkPredicate to `BETWEEN` leaf predicate now. Will submit a separated PR for this.

PR #7264, opened 2026-02-11
feat(python): add list tag for TagManager

cos120:
### Purpose

Linked issue: close #xxx

Adding list tags for table.

### DIFF
--- a/paimon-python/pypaimon/tag/tag_manager.py
+++ b/paimon-python/pypaimon/tag/tag_manager.py
@@ -16,6 +16,7 @@
 #  under the License.
 
 import logging
+import os
 from typing import Optional
 
 from pypaimon.common.file_io import FileIO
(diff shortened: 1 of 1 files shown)


paimon-python/pypaimon/tag/tag_manager.py
```diff
@@ -140,6 +140,14 @@ def create_tag(
 
         self._create_or_replace_tag(snapshot, tag_name)
 
+    def list_tag(self):
+        """List all tags."""
+        result = []
+        for tag_file in self.file_io.list_status(self.tag_directory()):
+            _, tag = tag_file.base_name.split("-", 1)
+            result.append(tag)
+        return result
```
XiaoHongbo-Hope, line 159: The tag name should be the substring after TAG_PREFIX.

cos120: since name of tag file is `f"{TAG_PREFIX}{tag_name}"`, `split("-", 1)` can handle it.

PR #7265, opened 2026-02-11
[spark] supports converting some SparkPredicate to Paimon between LeafPredicate

steFaiz:
### Purpose
Currently, spark will convert Between predicate to the composition of lessOrEqual and GreaterOrEqual, this PR is about to recognize this pattern, converting some And CompoundedPredicate to a single Between LeafPredicate. 

Linked issue: none 

### Tests
Please see `org.apache.paimon.spark.sql.SparkV2FilterConverterTestBase`

### API and Format
No changes.

### Documentation
No changes.

### Generative AI tooling
This PR is fully hand-written.

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/utils/PredicateUtils.java
+++ b/paimon-common/src/main/java/org/apache/paimon/utils/PredicateUtils.java
@@ -29,6 +29,8 @@ import org.apache.paimon.predicate.Predicate;
 import org.apache.paimon.predicate.PredicateBuilder;
 import org.apache.paimon.predicate.Transform;
 
+import javax.annotation.Nullable;
+
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/util/PredicateUtils.scala
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+package org.apache.paimon.spark.util
+
+import org.apache.paimon.predicate.{Between, CompareUtils, GreaterOrEqual, LeafPredicate, LessOrEqual, Predicate}
+
+import java.util
+import java.util.Objects
+
+object PredicateUtils {
```
JingsongLi: We could have a PredicateRewrite in core. Maybe invoked in `ReadBuilder.withFilter`.

steFaiz: @JingsongLi  Thanks for your advise! I've refactored it. But it's worthy to note that implementing this logic in core is much more complicated than in spark. We have to consider broader scenarios, for example:
1. recursive: OR(AND(a >= 1, a <= 10, a is not null), b > 10, ... )
2. chained AND: AND(a >= 1, AND(a <= 10, b > 10)) (this could be converted to BETWEEN(a, 1, 10))

and more. maybe I've missed some scenario.

Please take a look and i'm pleasant to improve my code and fix potential bugs.

JingsongLi (approved):
+1

PR #7271, opened 2026-02-12
[python] with_shard should be evenly distributed for data evolution mode

JingsongLi:
### Purpose

For machine learning, evenly distribution is very important.


leaves12138 (approved):
+1

PR #7272, opened 2026-02-12
[rest]Improve RestCatalog OpenAPI nonce generation

sundapeng:
### Purpose

Improve Paimon REST OpenAPI nonce generation and auth code(Inspired by aliyun/OpenApiTeaSDK approach)
- Use `UUID + timestamp + thread ID` for nonce instead of `UUID` alone, because UUIDs can collide in VMs;
- Replace `SimpleDateFormat` with `DateTimeFormatter` to improve thread safety;
- Add null checks in `signHeaders` and `authorization`.

### Tests

- `DLFRequestSignerTest#testOpenApiSignHeadersWithEnhancedNonce` — nonce format (UUID, timestamp, thread ID)
- `DLFRequestSignerTest#testConcurrentNonceGeneration` — concurrent nonce uniqueness

### API and Format

No. Internal improvement only; no public API or storage format change.

### Documentation

No. Internal improvement only.

### Generative AI tooling

No.

### DIFF
--- a/paimon-api/src/main/java/org/apache/paimon/rest/auth/DLFOpenApiSigner.java
+++ b/paimon-api/src/main/java/org/apache/paimon/rest/auth/DLFOpenApiSigner.java
@@ -26,13 +26,12 @@ import javax.crypto.spec.SecretKeySpec;
 
 import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
-import java.text.SimpleDateFormat;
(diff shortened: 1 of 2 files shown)


sundapeng:
The background is that we encountered a `Specified signature nonce was used already.` exception in the unit test environment.

sundapeng:
> I remember I have merged this PR...

Yep...

PR #7273, opened 2026-02-12
[spark] Introduce TrimTransform for trim/ltrim/rtrim functions to pushdown

xuzifu666:
### Purpose

Linked issue: close #xxx

The ```trim``` function are frequently used. This PR supports ```trim```/```ltrim```/```rtrim``` functions pushdown in Spark side.

Note the automatic conversion in Spark, as follows:
trim(leading 'abc' from b) => ltrim('abc', b)
trim(tailing 'abc' from b) => rtrim('abc', b)
trim(both 'abc' from b) => trim('abc', b)

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/predicate/TrimTransform.java
+++ b/paimon-common/src/main/java/org/apache/paimon/predicate/TrimTransform.java
@@ -69,6 +69,7 @@ public class TrimTransform extends StringTransform {
         return new TrimTransform(inputs, this.trimFlag);
     }
 
+    /** Enum of trim functions. */
     public enum Flag {
         LEADING,
         TRAILING,


paimon-api/src/main/java/org/apache/paimon/utils/StringUtils.java
```diff
@@ -545,6 +545,36 @@ public static String trim(String value) {
         return value.trim();
     }
 
+    public static String trim(String value, String charsToTrim) {
+        if (value == null || charsToTrim == null) {
```
JingsongLi: Useless if, remove it.

paimon-common/src/main/java/org/apache/paimon/predicate/TrimTransform.java
```diff
@@ -0,0 +1,75 @@
...
+
+    private static final long serialVersionUID = 1L;
+
+    public static final String NAME = "TRIM";
+
+    public static final String LTRIM = "LTRIM";
+
+    public static final String RTRIM = "RTRIM";
+
+    private final String trimWay;
```
JingsongLi: Create an enum `Flag` in this class. And values `LEADING`, `TRAILING`, `BOTH`.

paimon-api/src/main/java/org/apache/paimon/utils/StringUtils.java
xuzifu666: Thanks for reminder，Done.

paimon-common/src/main/java/org/apache/paimon/predicate/TrimTransform.java
xuzifu666: Done.

PR #7298, opened 2026-02-24
[core] Data evolution table limit push down

leaves12138:
### Purpose

Push down limit for data-evolution table


JingsongLi (approved):
+1

PR #7310, opened 2026-02-26
[core] Support schema_id predicate pushdown for ManifestsTable

kerwin-zk:
### Purpose

Support schema_id predicate pushdown for ManifestsTable

### Tests

CI


JingsongLi (approved):
+1

Issue #7306, opened 2026-02-25
[Feature] Support native Flink 2.0 job lineage (FLIP-314) for Paimon tables


jsingh-yelp:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Motivation

Flink 2.0 introduced native job lineage as a first-class feature (FLIP-314). When a job starts, Flink builds a LineageGraph of sources and sinks and fires it to any registered JobStatusChangedListener, enabling external lineage systems (for example open lineage) to consume accurate lineage without custom instrumentation.

The [FLIP-314](https://cwiki.apache.org/confluence/display/FLINK/FLIP-314%3A+Support+Customized+Job+Lineage+Listener) proposal itself uses Paimon as one of its motivating example.

### Solution

To support Flink’s native lineage tracking, implement the `LineageVertexProvider` interface in Paimon’s FlinkSource and FlinkSink classes. The implementation should return the table name in the name() method and the warehouse path in the namespace() method, enabling table-level lineage reporting to external systems. This is a small, self-contained change with no impact on Paimon’s internal data model or APIs.

**Note**: I see there were other similar issues reported related to PIP-5 ( https://cwiki.apache.org/confluence/display/PAIMON/PIP-5%3A+Paimon+Table+And+Data+Lineage+For+Flink), I ended up creating a separate issue because this change is much more contained with in Flink's effort to support native table-level lineage and should require lot less effort.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

jsingh-yelp:
Changes done for this feature:
- Enhanced existing Paimon Flink Table API lineage by adding dataset namespace handling and enriching datasets with Paimon catalog & table config metadata: https://github.com/apache/paimon/pull/7311, https://github.com/apache/paimon/pull/7576
- Added lineage support for Paimon Flink non-table DataStream APIs: https://github.com/apache/paimon/pull/8001

PR #7311, opened 2026-02-27
[flink] Implement FLIP-314 LineageVertexProvider for source and sink connectors

jsingh-yelp:
### Purpose
- Implement `LineageVertexProvider` which helps flink ([FLIP-314](https://cwiki.apache.org/confluence/display/FLINK/FLIP-314%3A+Support+Customized+Job+Lineage+Listener)) to generate lineage events with additional details which flink don't have access to.
- It generates a proper namespace and passes down table config options with additional details such as primary_keys, partition_keys e.t.c
- Linked issue: https://github.com/apache/paimon/issues/7306

### Tests
- Added unit tests for the change
- Also was able to do end to end manual testing for change. This is how the final lineage event will look like using openlineage: https://i.fluffy.cc/7KGHgtJ3x003hvs4lK2rrBDNqgmF9MLZ.html#L87-88,L95-120,L129-131

### API and Format
- This change does not change any API

### Documentation
- Yes I think this is worth adding documentation for, I am happy to add docs for it if someone can point me where to add the specific documentation for this.
- Yes this introduces a new feature of adding support of native lineage generation of Paimon tables for Flink.

**Note**: This change also depends on two other changes and one of them is already merged. Other two changes:
- Flink: https://github.com/apache/flink/pull/27727
- Openlineage: [Already Merged] https://github.com/OpenLineage/OpenLineage/pull/4368


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/FlinkTableSinkBase.java
```diff
@@ -104,13 +104,18 @@ public SinkRuntimeProvider getSinkRuntimeProvider(Context context) {
             throw new UnsupportedOperationException(
                     "Paimon doesn't support streaming INSERT OVERWRITE.");
         }
+        String name = tableIdentifier.asSummaryString();
+        String namespace = CoreOptions.path(table.options()).toString();
```
jsingh-yelp:
```suggestion
        String namespace =
                table instanceof DataTable
                        ? ((DataTable) table).location().toString()
                        : CoreOptions.path(table.options()).toString();
```

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/SystemTableSource.java
```diff
@@ -130,7 +130,9 @@ public ScanRuntimeProvider getScanRuntimeProvider(ScanContext scanContext) {
                         dataStreamSource.setParallelism(parallelism);
                     }
                     return dataStreamSource;
-                });
+                },
+                tableIdentifier.asSummaryString(),
+                CoreOptions.path(table.options()).toString());
```
jsingh-yelp:
```suggestion
                table instanceof DataTable
                        ? ((DataTable) table).location().toString()
                        : CoreOptions.path(table.options()).toString());
```

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/BaseDataTableSource.java
```diff
@@ -202,7 +202,9 @@ public ScanRuntimeProvider getScanRuntimeProvider(ScanContext scanContext) {
                         sourceBuilder
                                 .sourceParallelism(inferSourceParallelism(env))
                                 .env(env)
-                                .build());
+                                .build(),
+                tableIdentifier.asSummaryString(),
+                CoreOptions.path(table.options()).toString());
```
jsingh-yelp:
```suggestion
                table instanceof DataTable
                        ? ((DataTable) table).location().toString()
                        : CoreOptions.path(table.options()).toString());
```

paimon-flink/paimon-flink2-common/pom.xml
```diff
@@ -55,6 +55,12 @@ under the License.
             <version>${flink.version}</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-table-api-java-bridge</artifactId>
```
jsingh-yelp: This is required because since now we have to import `DataStreamScanProvider` in flink 2-common. Code ref: https://github.com/apache/flink/blob/c0479c74f2b736aaed00b4d9fb1b14ad296e562d/flink-table/flink-table-api-java-bridge/src/main/java/org/apache/flink/table/connector/source/DataStreamScanProvider.java#L40

paimon-flink/paimon-flink2-common/src/main/java/org/apache/paimon/flink/lineage/DataStreamProviderFactory.java
```diff
@@ -0,0 +1,117 @@
...
+     * Flink's lineage graph discovers the Paimon sink table.
+     */
+    public static SinkRuntimeProvider getSinkProvider(
+            SinkRuntimeProvider provider, String name, Table table) {
+        return new LineageAwarePaimonDataStreamSinkProvider(
+                (DataStreamSinkProvider) provider, name, table);
+    }
+
+    private static class LineageAwarePaimonDataStreamScanProvider
+            implements DataStreamScanProvider, LineageVertexProvider {
```
yunfengzhou-hub: Compared with delegate pattern, would it be better to directly have `PaimonDataStreamScanProvider` implement `LineageVertexProvider`?

Same for `LineageAwarePaimonDataStreamSinkProvider`.

jsingh-yelp: Thanks for the review @yunfengzhou-hub. The reason for the delegate pattern is backward compatibility. `PaimonDataStreamScanProvider` lives in paimon-flink-common, which is built against multiple Flink versions, including ones older than 1.20. Since `LineageVertexProvider` only exists in Flink 1.20+, having `PaimonDataStreamScanProvider` directly implement it would cause a ClassNotFoundException at build/load time for older Flink versions.

By using the delegate pattern with a version-specific `DataStreamProviderFactory`, the lineage-aware wrapper is only created in the paimon-flink2-common module (which targets Flink 2.x), and the older Flink modules never reference `LineageVertexProvider` at all.

JingsongLi: Another way is copying all `LineageVertexProvider` related classes to lower Paimon-Flink module from Flink.

This may seem like copying a bunch of classes, but it ensures the cleanliness of the new version.

yunfengzhou-hub: +1 with Jingsong's suggestion. It is better to have Paimon's code best fit the latest version of the engine. Compatiblity with older versions should be considered afterwards, if the introduced classes for compatibility are not that much. Existing examples are like
`paimon-flink/paimon-flink1-common/src/main/java/org/apache/flink/types/variant/Variant.java`.

jsingh-yelp: @JingsongLi , @yunfengzhou-hub Suggestion makes sense to me, I have applied the fix to copy over all the missing files for older versions of Flink.

yunfengzhou-hub (approved):
+1

PR #7313, opened 2026-02-27
[spark] Support compact_chain_table procedure

### LINKED ISSUE #7312: [Feature] [spark] Support compact_chain_table procedure
### Search before asking

### Motivation

Chain Compact
Compact the incremental data of the current cycle with the full data of the previous cycle to generate the full data for the day; for example, the full data for date=20250729 is generated by compacting all incremental partition from 20250723 to 20250729 on t$branch_delta and 20250722 on t$branch_snapshot

https://cwiki.apache.org/confluence/display/PAIMON/PIP-37%3A+Introduce+Chain+Table

### Solution

Add the `CompactChainTableProcedure` to implement this feature.

### Anything else?

_No response_

juntaozhang:
### Purpose

Linked issue: close #7312

### Tests

- CompactChainTableProcedureTest.scala
### Documentation
- docs/content/primary-key-table/chain-table.md
- docs/content/spark/procedures.md

### Generative AI tooling

No

### DIFF
--- a/docs/docs/primary-key-table/chain-table.md
+++ b/docs/docs/primary-key-table/chain-table.md
@@ -130,7 +130,6 @@ Notice that:
 - Chain table is only supported for primary key table, which means you should define `bucket` and `bucket-key` for the table.
 - Chain table should ensure that the schema of each branch is consistent.
 - Both Spark and Flink batch read/write are supported. Flink streaming read/write is not supported.
-- Chain compact is not supported for now, and it will be supported later.
 - Deletion vector is not supported for chain table.
 
 After creating a chain table, you can read and write data in the following ways.
@@ -188,6 +187,32 @@ you will get the following result:
 | 1 |   1|   1 |           
 | 2 |   1|   1 |  
 | 2 |   1|   1 |               
++---+----+-----+ 
+```
+
+- Chain Table Compaction: Merge data from snapshot and delta branches into the snapshot branch.
+  This is useful for periodically compacting incremental data into full snapshots.
+  You can use the `compact_chain_table` procedure to merge a specific partition:
+
+```sql
+CALL sys.compact_chain_table(table => 'default.t', partition => 'date="20250811"');
+```
+
+After compaction, the data in the snapshot branch will contain the merged result from both snapshot
+and delta branches, and subsequent queries will benefit from direct snapshot access without
+merge-on-read overhead.
+
+```sql
+select t1, t2, t3 from `default`.`t$branch_snapshot` where date = '20250811';
+```
+
+you will get the following result:
+```text
++---+----+-----+ 
+| t1|  t2|   t3| 
++---+----+-----+ 
+| 1 |   1|   1 | 
+| 2 |   1|   1 |
 +---+----+-----+ 
 ```
 
--- a/docs/docs/primary-key-table/chain-table.md
+++ b/docs/docs/primary-key-table/chain-table.md
@@ -152,12 +152,12 @@ select t1, t2, t3 from default.t where date = '20250811'
 ```
 you will get the following result:
 ```text
-+---+----+-----+ 
-| t1|  t2|   t3| 
-+---+----+-----+ 
-| 1 |   1|   1 |           
-| 2 |   1|   1 |               
-+---+----+-----+ 
++---+----+-----+
+| t1|  t2|   t3|
++---+----+-----+
+| 1 |   1|   1 |
+| 2 |   1|   1 |
++---+----+-----+
 ```
 
 - Incremental Query: Read the incremental partition from t$branch_delta
@@ -166,11 +166,11 @@ select t1, t2, t3 from `default`.`t$branch_delta` where date = '20250811'
 ```
 you will get the following result:
 ```text
-+---+----+-----+ 
-| t1|  t2|   t3| 
-+---+----+-----+      
-| 2 |   1|   1 |               
-+---+----+-----+ 
++---+----+-----+
+| t1|  t2|   t3|
++---+----+-----+
+| 2 |   1|   1 |
++---+----+-----+
 ```
 
 - Hybrid Query: Read both full and incremental data simultaneously.
@@ -181,13 +181,13 @@ select t1, t2, t3 from `default`.`t$branch_delta` where date = '20250811'
 ```
 you will get the following result:
 ```text
-+---+----+-----+ 
-| t1|  t2|   t3| 
-+---+----+-----+ 
-| 1 |   1|   1 |           
-| 2 |   1|   1 |  
-| 2 |   1|   1 |               
-+---+----+-----+ 
++---+----+-----+
+| t1|  t2|   t3|
++---+----+-----+
+| 1 |   1|   1 |
+| 2 |   1|   1 |
+| 2 |   1|   1 |
++---+----+-----+
 ```
 
 - Chain Table Compaction: Merge data from snapshot and delta branches into the snapshot branch.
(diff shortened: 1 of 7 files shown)


JingsongLi (commented):
Thanks for adding the Spark procedure for chain table compaction.

**Comments:**

1. **PR description is minimal** — the "Purpose" section just links issue #7312 without explaining the approach. Please describe:
   - How does `compact_chain_table` differ from regular `compact`?
   - Does it compact both snapshot and delta branches?
   - What's the merge strategy for chain compaction?

2. **File changes**: The PR touches `ChainGroupReadTable`, `FallbackReadFileStoreTable`, and `ChainSplit`. What changes were needed to support compaction vs. just read? Are these refactoring prerequisites or functional changes?

3. **613 additions** is significant. A test file (`CompactChainTableProcedureTest.scala`) + procedure + supporting core changes — consider whether the core changes could be a separate prerequisite PR.

4. **Documentation**: Good that both `chain-table.md` and `procedures.md` are updated.

5. **`ChainSplitTest`**: What cases does it cover? Chain tables have complex split logic due to cross-branch data dependencies.

Please fill in the PR description with the design approach before requesting final review.

juntaozhang:
> Thanks for adding the Spark procedure for chain table compaction.
> 
> **Comments:**
> 
> 1. **PR description is minimal** — the "Purpose" section just links issue #7312 without explaining the approach. Please describe:
>    
>    * How does `compact_chain_table` differ from regular `compact`?
>    * Does it compact both snapshot and delta branches?
>    * What's the merge strategy for chain compaction?
> 2. **File changes**: The PR touches `ChainGroupReadTable`, `FallbackReadFileStoreTable`, and `ChainSplit`. What changes were needed to support compaction vs. just read? Are these refactoring prerequisites or functional changes?
> 3. **613 additions** is significant. A test file (`CompactChainTableProcedureTest.scala`) + procedure + supporting core changes — consider whether the core changes could be a separate prerequisite PR.
> 4. **Documentation**: Good that both `chain-table.md` and `procedures.md` are updated.
> 5. **`ChainSplitTest`**: What cases does it cover? Chain tables have complex split logic due to cross-branch data dependencies.
> 
> Please fill in the PR description with the design approach before requesting final review.

Thank you for the thorough review and valuable feedback.
1. PR description
   Already updated in #7312.
2. Core file changes
  Extracted to prerequisite PR #7950.
5. ~~ChainSplitTest / Split support~~
   `ScanPlanHelper.createNewScanPlan` already supports the `Split` interface, no additional changes needed.

JingsongLi:
Can you rebase master?

juntaozhang:
> Can you rebase master?
Hi @JingsongLi, thanks for the reminder, done.

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/procedure/CompactChainTableProcedure.java
```diff
@@ -0,0 +1,215 @@
...
+                (FallbackReadFileStoreTable.FallbackReadScan) table.newScan();
+        PartitionPredicate partitionPredicate =
+                SparkProcedureUtils.convertToPartitionPredicate(
+                        partition, table.schema().logicalPartitionType(), spark(), relation);
+
+        // Check if target partition already exists in snapshot branch
+        boolean partitionExists = checkPartitionExists(snapshotTable, partition, relation);
+        if (partitionExists) {
+            if (overwrite) {
+                scan.withPartitionFilter(
```
JingsongLi: This overwrite path appears to read too much data. When the target partition already exists, the snapshot-side predicate is changed to `NOT (target partition)`, while the delta side still uses the target partition. However, `ChainGroupReadTable.plan()` adds all `mainScan.plan()` splits directly before planning the delta/anchor splits, and this procedure later rewrites every output row's partition columns to the target partition. That means unrelated snapshot partitions can be copied into the compacted target partition during overwrite. The overwrite path should only read the chain-merge result needed for the target partition, not every snapshot partition except the target.

docs/docs/primary-key-table/chain-table.md
```diff
@@ -191,6 +190,32 @@ you will get the following result:
...
+and delta branches, and subsequent queries will benefit from direct snapshot access without
+merge-on-read overhead.
+
+```sql
+select t1, t2, t3 from `default`.`t$branch_snapshot` where date = '20250811';
+```
+
+you will get the following result:
+```text
++---+----+-----+
```
JingsongLi: Please remove the trailing whitespace in this added result block. `git diff --check origin/master...HEAD` reports trailing whitespace on this block, which will fail whitespace/style checks.

JingsongLi (commented):
I found a couple of issues while reviewing this change.

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/procedure/CompactChainTableProcedure.java
juntaozhang: Thanks a lot for your review. I added a flag of `preloadTargetSnapshot`  to skip executing `mainScan.plan()` when overwriting the target partition. 
Appreciate your reminder, thanks.

docs/docs/primary-key-table/chain-table.md
juntaozhang: Done, thanks for your reminder.

JingsongLi (commented):
I found two correctness issues around overwrite scope and partition parsing.

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/procedure/CompactChainTableProcedure.java
```diff
@@ -0,0 +1,213 @@
...
+
+        Dataset<Row> datasetForWrite =
+                PaimonUtils.createDataset(
+                        spark(),
+                        ScanPlanHelper$.MODULE$.createNewScanPlan(
+                                splits.toArray(new Split[0]), relation));
+
+        PaimonSparkWriter writer = PaimonSparkWriter.apply(snapshotTable);
+        if (partitionExists) {
+            writer.writeBuilder().withOverwrite();
```
JingsongLi: This overwrite path relies on the table-level `dynamic-partition-overwrite` option because it calls `withOverwrite()` with an empty static partition. If the snapshot branch/table has `dynamic-partition-overwrite=false`, `FileStoreCommitImpl` will build a null overwrite filter and may overwrite the whole snapshot branch instead of only the requested `partitionStr`. Could we force dynamic partition overwrite here or pass the target partition map explicitly so the procedure's overwrite scope is independent of table configuration?

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/procedure/CompactChainTableProcedure.java
```diff
@@ -0,0 +1,213 @@
...
+        }
+        Map<String, String> targetPartition = ParameterUtils.getPartitions(partitionStr).get(0);
```
JingsongLi: `SparkProcedureUtils.toWhere(partitionStr)` accepts semicolon-separated partitions, but the write path only uses `get(0)` and then rewrites every row to that first partition via `withColumn`. If a user passes multiple partitions, data from later partitions can be written into the first partition. Please either reject `partitionStr` values containing multiple partitions or handle each target partition separately.

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/procedure/CompactChainTableProcedure.java
juntaozhang: Thanks for your patient guidance. Forcing dynamic partition overwrite here is a great approach. I've revised the code following this solution already.

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/procedure/CompactChainTableProcedure.java
juntaozhang: Thanks for your guidance. Currently this procedure only supports a single partition, and I have already added validation logic to reject partition strings containing multiple partitions.

JingsongLi (approved):
+1

PR #6353, opened 2025-09-29
add --metadata_column support for Kafka CDC connector

gmdfalk:
### Description

Add `--metadata_column` support to Paimon Kafka CDC connector, similar to the already existing options added for MySQL and Postgres: https://github.com/apache/paimon/pull/2077
Also add optional `--metadata_column_prefix` to avoid conflicts with existing Paimon fields like `topic`, `timestamp` etc.

Supported metadata columns are those on `org.apache.kafka.clients.consumer.ConsumerRecord` i.e.:
* topic
* partition
* offset
* timestamp
* timestampType: This is the name of the enum i.e. `NoTimestampType`, `CreateTime` or `LogAppendTime`

The feature is backwards compatible. It's only active when `--metadata_column` is supplied resp. `SynchronizationActionBase.withMetadataColumns` is used.

### Motivation
This is a requested feature: https://github.com/apache/paimon/issues/3210

We primarly use this feature for two purposes:
1. Troubleshooting and data lineage (e.g. where in our Kafka infrastructure does this Paimon row come from?)
2. Mapping large Kafka topic partitions 1:1 to Paimon buckets to avoid reshuffling (see this issue it would solve: https://github.com/apache/paimon/issues/3249)

### Tests
Unit tests: [KafkaMetadataConverter.java](paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/action/cdc/kafka/KafkaMetadataConverter.java)

Integration tests: [KafkaSyncTableActionITCase.java::testMetadataColumn](src/test/java/org/apache/paimon/flink/action/cdc/kafka/KafkaSyncTableActionITCase.java)

### API and Format
No changes to public apis or storage format.

The changes here are contained to the flink cdc package but I did have to update `CdcSourceRecord` since it previously didn't provide a way to surface arbitrary metadata for a record.

The metadata attribute on `CdcSourceRecord` is intentionally a generic Map so that it can potentially be used to add metadata support for other connectors like Pulsar or Mongo that are not yet implemented.

### Documentation

Added the new `--metadata_column` and `--metadata_column_prefix` parameter to Kafka CDC docs.

### Dev notes
For running integration tests on MacOS with Rancher Desktop, i had to properly expose the docker socket to testcontainers e.g. system wide via `sudo ln -sf "$HOME/.rd/docker.sock" /var/run/docker.sock`.

Issue #3249, opened 2024-04-23
[Feature] 


polyzos:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Motivation

When creating an Append for Message Queue table, as depicted in the screenshot here:
![Screenshot 2024-04-23 at 8 33 05 AM](https://github.com/apache/paimon/assets/23555517/d23bf690-7c06-4f90-91da-a4009cbbbb2b)

we can notice the following:
1. 5 buckets are specified, but unless data comes in the bucket is not created
2. If you check the Kafka partitions; partition 3 has keys 2, 3 and 4 
3. These keys though end up in different buckets
4. Paimon does a shuffle, even though the parallelism is the same because it doesn't do 1-1 mapping

Because it is a Kafka-like message queue functionality, some users are confused, as they expect the same partitioning to happen and overall have a 1-1 mapping, between a Kafka partition and a paimon bucket.

At the same time, I believe this is a really good enhancement and should also allow to remove the shuffle between the operators, thus improving performance.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

eric666666:
If you want 1->1 mapping，Paimon's bucket number should bigger than kafka partitions, and their should be shuffle by kafka partition id.
I think paimon already can implement your thoughts.
Here is a demo, you can define ddl like this
Kafka source table:
```
CREATE TABLE KafkaTable (
  `event_time` TIMESTAMP(3) METADATA FROM 'value.source.timestamp' VIRTUAL,  -- from Debezium format
  `origin_table` STRING METADATA FROM 'value.source.table' VIRTUAL, -- from Debezium format
  `partition_id`  int METADATA FROM 'partition' VIRTUAL,  -- from Kafka connector
  `offset` BIGINT METADATA VIRTUAL,  -- from Kafka connector
  `user_id` BIGINT,
  `item_id` BIGINT,
  `behavior` STRING
) WITH (
  'connector' = 'kafka',
  'topic' = 'user_behavior',
  'properties.bootstrap.servers' = 'localhost:9092',
  'properties.group.id' = 'testGroup',
  'scan.startup.mode' = 'earliest-offset',
  'value.format' = 'debezium-json'
);
```
Paimon sink table:
```
CREATE table if not exists sink_paimon_table
        WITH ('connector' = 'paimon',
        'bucket' = '3',  -- bucket number should bigger than kafka partitions 
[...]

Issue #3210, opened 2024-04-15
[Feature] kafka cdc support metadata_column


melin:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Motivation

Get metadata from kafka header, such as database, tablename, and other keys/values

flink cdc kafka pipeline writes database, tablename to kafka header, plan to support custom key/value
https://github.com/apache/flink-cdc/pull/2938

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #2077, opened 2023-09-26
[flink] CDC Ingestion supported metadata columns

MonsterChenzhuo:
### Purpose

close #1985

### Tests

MySqlSyncDatabaseActionITCase#testMetadataColumns()
MySqlSyncTableActionITCase#testMetadataColumns()

PR #7315, opened 2026-02-27
Add metadata column support for Kafka CDC connector

nickdelnano:
This PR supercedes https://github.com/apache/paimon/pull/6353 - I took over this work from my colleague @gmdfalk 

## Description
Add --metadata_column support to Paimon Kafka CDC connector, similar to the already existing options added for MySQL and Postgres: https://github.com/apache/paimon/pull/2077
Also add optional --metadata_column_prefix to avoid conflicts with existing Paimon fields like topic, timestamp etc.

Supported metadata columns are those on org.apache.kafka.clients.consumer.ConsumerRecord i.e.:

- topic
- partition
- offset
- timestamp
- timestampType: This is the name of the enum i.e. NoTimestampType, CreateTime or LogAppendTime

The feature is backwards compatible. It's only active when --metadata_column is supplied resp. SynchronizationActionBase.withMetadataColumns is used.

## Motivation
This is a requested feature: https://github.com/apache/paimon/issues/3210

We primarly use this feature for two purposes:

1. Troubleshooting and data lineage (e.g. where in our Kafka infrastructure does this Paimon row come from?)
2. Mapping large Kafka topic partitions 1:1 to Paimon buckets to avoid reshuffling (see this issue it would solve: https://github.com/apache/paimon/issues/3249)

## Tests
Unit and Integration Tests

## API and Format
No changes to public apis or storage format.

The changes here are contained to the flink cdc package but I did have to update CdcSourceRecord since it previously didn't provide a way to surface arbitrary metadata for a record.

The metadata attribute on CdcSourceRecord is intentionally a generic Map so that it can potentially be used to add metadata support for other connectors like Pulsar or Mongo that are not yet implemented.

## Documentation
Added the new --metadata_column and --metadata_column_prefix parameter to Kafka CDC docs.

## Dev notes
For running integration tests on MacOS with Rancher Desktop, i had to properly expose the docker socket to testcontainers e.g. system wide via sudo ln -sf "$HOME/.rd/docker.sock" /var/run/docker.sock.

### DIFF
--- a/paimon-flink/paimon-flink-cdc/src/test/java/org/apache/paimon/flink/action/cdc/kafka/KafkaAWSDMSSyncTableActionITCase.java
+++ b/paimon-flink/paimon-flink-cdc/src/test/java/org/apache/paimon/flink/action/cdc/kafka/KafkaAWSDMSSyncTableActionITCase.java
@@ -76,6 +76,12 @@ public class KafkaAWSDMSSyncTableActionITCase extends KafkaSyncTableActionITCase
 
     @Test
     @Timeout(60)
+    public void testMetadataColumn() throws Exception {
+        testMetadataColumn(AWSDMS);
+    }
+
+    @Test
+    @Timeout(60)
     public void testFieldValNullSyncTable() throws Exception {
         testTableFiledValNull(AWSDMS);
     }
--- a/paimon-flink/paimon-flink-cdc/src/test/java/org/apache/paimon/flink/action/cdc/kafka/KafkaCanalSyncTableActionITCase.java
+++ b/paimon-flink/paimon-flink-cdc/src/test/java/org/apache/paimon/flink/action/cdc/kafka/KafkaCanalSyncTableActionITCase.java
@@ -841,6 +841,12 @@ public class KafkaCanalSyncTableActionITCase extends KafkaSyncTableActionITCase 
 
     @Test
     @Timeout(60)
+    public void testMetadataColumn() throws Exception {
+        testMetadataColumn(CANAL);
+    }
+
+    @Test
+    @Timeout(60)
     public void testTypeMappingToString() throws Exception {
         final String topic = "map-to-string";
         createTestTopic(topic, 1, 1);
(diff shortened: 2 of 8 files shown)


paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/action/cdc/format/AbstractDataFormat.java
```diff
@@ -55,9 +55,8 @@ public AbstractRecordParser createParser(
             TypeMapping typeMapping,
             List<ComputedColumn> computedColumns,
             CdcMetadataConverter[] metadataConverters) {
-        // Most parsers don't support metadata converters, so we default to the 2-parameter version
-        // Only specific parsers like DebeziumAvroRecordParser will override this
-        return createParser(typeMapping, computedColumns);
```
nickdelnano: @gmdfalk curious why not supported on other debezium formats except debezium avro?

`testComputedColumn` was failing for other formats because:
-  this test runs for many formats other than debezium avro 
- it was waiting for kafka metadata fields to appear yet they weren't configured

paimon-flink/paimon-flink-cdc/src/test/java/org/apache/paimon/flink/action/cdc/kafka/KafkaSyncTableActionITCase.java
```diff
@@ -572,41 +555,16 @@ public void testMetadataColumn(String format) throws Exception {
...
-                        "__kafka_timestamp_type");
-
-        // Verify the data types of metadata columns
-        assertThat(tableRowType.getField("__kafka_topic").type()).isEqualTo(DataTypes.STRING());
-        assertThat(tableRowType.getField("__kafka_partition").type()).isEqualTo(DataTypes.INT());
-        assertThat(tableRowType.getField("__kafka_offset").type()).isEqualTo(DataTypes.BIGINT());
-        assertThat(tableRowType.getField("__kafka_timestamp").type())
-                .isEqualTo(DataTypes.TIMESTAMP_WITH_LOCAL_TIME_ZONE(3));
-        assertThat(tableRowType.getField("__kafka_timestamp_type").type())
-                .isEqualTo(DataTypes.STRING());
```
nickdelnano: these assertions are redundant - `waitForResult` below validates column types

paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/action/cdc/format/AbstractDataFormat.java
gmdfalk: i initially scoped the change just to Debezium Avro format because that's what we were using and I wanted to get tests to pass there first. Since the other formats and tests are so similar, it would then have been a small jump to add it to all of them. 
But turns out this was the exact choice that blocked me from testing it successfully, so thanks a lot!

JingsongLi (approved):
+1

PR #7316, opened 2026-02-28
[core] Remove clone & migrate support for Hudi, Iceberg

JingsongLi:
### Purpose

These features have hardly been used and there are many hidden dangers. After time verification, we have fully proved that they are not very useful.


LsomeYeah (approved):
Looks good to me! Just waiting for the tests to pass.

PR #7318, opened 2026-02-28
[core] Introduce BTreeIndexReader.createIterator to iterator all records in BTree

JingsongLi:
### Purpose

In future development, we can consider introducing the ability to merge multiple BTree files by reading all elements for merging.


steFaiz (approved):
+1 Thanks for this!

PR #7323, opened 2026-02-28
[python] Support conflict detection when updating existing data concurrently.

discivigour:
### Purpose

Support conflict detection when updating the data-evolution table concurrently.

### API and Format

- TableUpdateTest.test_concurrent_updates_with_retry()

### DIFF
--- a/paimon-python/pypaimon/write/commit_message.py
+++ b/paimon-python/pypaimon/write/commit_message.py
@@ -17,7 +17,7 @@
 ################################################################################
 
 from dataclasses import dataclass
-from typing import List, Tuple
+from typing import List, Tuple, Optional
 
 from pypaimon.manifest.schema.data_file_meta import DataFileMeta
 
@@ -27,7 +27,7 @@ class CommitMessage:
     partition: Tuple
     bucket: int
     new_files: List[DataFileMeta]
-    snapshot_for_update: int = -1
+    check_from_snapshot: Optional[int] = -1
 
     def is_empty(self):
         return not self.new_files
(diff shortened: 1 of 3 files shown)


paimon-python/pypaimon/write/commit_message.py
```diff
@@ -27,6 +27,7 @@ class CommitMessage:
     partition: Tuple
     bucket: int
     new_files: List[DataFileMeta]
+    snapshot_for_update: int = -1
```
JingsongLi: just `check_from_snapshot`? and use `Option[long]`

paimon-python/pypaimon/write/file_store_commit.py
```diff
@@ -93,11 +97,30 @@ def __init__(self, snapshot_commit: SnapshotCommit, table, commit_user: str):
...
+
     def commit(self, commit_messages: List[CommitMessage], commit_identifier: int):
         """Commit the given commit messages in normal append mode."""
         if not commit_messages:
             return
 
+        # Extract snapshot_for_update from commit messages
+        for msg in commit_messages:
+            if msg.snapshot_for_update != -1:
+                self.conflict_detection._row_id_check_from_snapshot = msg.snapshot_for_update
```
JingsongLi: Here should do a min computation.

paimon-python/pypaimon/write/commit_message.py
discivigour: I will rename it.

paimon-python/pypaimon/write/file_store_commit.py
discivigour: ok.

PR #7324, opened 2026-03-02
[flink] optimize flink Data-Evolution-Merge-Into

steFaiz:
### Purpose
This PR optimize Data-Evolution-Merge-Into in several aspects:
1. Introduce a MergeIntoUpdateChecker to check if some global-indexed columns are updated (This is same as Spark's implementation)
2. Use calcite to rename target table (current implementation is based on regex, which is very unstable)
3. Use calcite to find _row_id field (if exists) in source table. We can eliminate join process.

Linked issue: None

### Tests
Please see: `org.apache.paimon.flink.action.DataEvolutionMergeIntoActionITCase`.

### API and Format
None

### Documentation
None

### Generative AI tooling
Full hand-writing.


steFaiz:
> Maybe you can just use CommitOperator with `commit.strict-mode.last-safe-snapshot` configured in `DataEvolutionMergeIntoAction`.

@JingsongLi  Thanks for your advise! I've modified my code.

JingsongLi:
> > @steFaiz Sounds good to me.
> 
> @JingsongLi Appreciate for it! Do you mean I should introduce a UpdateCheckerOperator or keep the new committer(just resolve the conflict)?

 introduce a UpdateCheckerOperator

PR #7327, opened 2026-03-02
[vector] remove lucene and faiss

jerry-024:
### Purpose

This PR removes specific indexing modules from the 1.4 release due to technical constraints regarding Python support and storage efficiency.
- Lucene: Removed because it lacks native Python support, which conflicts with our cross-language compatibility goals.

- Faiss: Removed due to the requirement of downloading large index files during query execution, which significantly impacts performance and resource overhead.


JingsongLi:
Please describe clearly why they need to be deleted.

JingsongLi:
+1

PR #7331, opened 2026-03-02
[core] Enable clustering before write phase for incremental clustering table

LsomeYeah:
### Purpose

Linked issue: close #xxx

For Append -1 tables, clustering data before batch write and incremental clustering are mutually exclusive. This PR introduces a parameter `'clustering.pre-write.enabled'` to enable clustering of data before write when incremental clustering is enabled.

### DIFF
--- a/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
@@ -3409,7 +3409,7 @@ public class CoreOptions implements Serializable {
         return options.get(CLUSTERING_INCREMENTAL);
     }
 
-    public boolean preClusteringEnabled() {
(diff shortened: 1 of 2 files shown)


paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -2071,6 +2071,13 @@ public InlineElement getDescription() {
                             "The duration after which a partition without new updates is considered a historical partition. "
                                     + "Historical partitions will be automatically fully clustered during the cluster operation.");
 
+    public static final ConfigOption<Boolean> CLUSTERING_PRE_WRITE_ENABLED =
+            key("clustering.pre-write.enabled")
```
JingsongLi: clustering.incremental.optimize-write

PR #7337, opened 2026-03-03
[filesystem] support report access extended information to DLF OSS

shyjsarah:
### Purpose
This PR adds support for DLF access tracking in Paimon's REST Catalog integration. When enabled, the access tracking extended info returned by getTableToken is propagated to the underlying FileIO (e.g., JindoFileIO) as part of the OSS User-Agent header, enabling partition-level access statistics on the DLF side.

Configuration Options：
- dlf.access-tracking.enabled (boolean, default false): Server-side catalog-level configuration. Should be enabled when partition-level access statistics are needed.

### Tests

org.apache.paimon.jindo.TestJindoDlfAccessTracking


paimon-common/src/main/java/org/apache/paimon/rest/RESTTokenFileIO.java
```diff
@@ -63,6 +64,12 @@ public class RESTTokenFileIO implements FileIO {
                     .defaultValue(false)
                     .withDescription("Whether to support data token provided by the REST server.");
 
+    public static final ConfigOption<String> DLF_ACCESS_TRACKING_EXTENDED_INFO =
+            ConfigOptions.key("dlf.access-tracking.extended-info")
```
JingsongLi: We don't need this option, enable or disable should be configured in remote Rest server.

shyjsarah: update

PR #7338, opened 2026-03-03
[core] Fix integer overflow in FieldSumAgg causing silent data corruption

dubin555:
### Purpose

`FieldSumAgg` performs unchecked arithmetic for TINYINT, SMALLINT, INTEGER, and BIGINT types. When the accumulator overflows, the result silently wraps around (e.g., `Byte.MAX_VALUE + 1` becomes `-128`), producing incorrect aggregation results without any error or warning. This is silent data corruption in the merge-tree SUM aggregation path.

The root cause differs by type:
- **TINYINT/SMALLINT:** Java promotes operands to `int` for arithmetic, then the `(byte)` or `(short)` cast silently truncates the overflow bits.
- **INTEGER/BIGINT:** Standard `+` and `-` wrap around at `MAX_VALUE` / `MIN_VALUE` per Java spec.

This fix adds overflow detection to all three affected methods (`agg`, `retract`, `negative`):
- For INTEGER and BIGINT: uses `Math.addExact` / `Math.subtractExact` / `Math.negateExact`, which throw `ArithmeticException` on overflow.
- For TINYINT and SMALLINT: uses range-checked helper methods that perform arithmetic in `int` width and verify the result fits in `byte` / `short` range before casting.

This is consistent with the overflow-protection pattern already used in `DecimalUtils.add()` and `DecimalUtils.subtract()` in the same codebase.

### Tests

- `testFieldSumByteOverflow` — TINYINT: verifies normal addition works, positive overflow throws `ArithmeticException`, negative overflow throws, and retract overflow throws.
- `testFieldSumShortOverflow` — SMALLINT: same coverage.
- `testFieldSumIntOverflow` — INTEGER: same coverage.
- `testFieldSumLongOverflow` — BIGINT: same coverage.

### API and Format

### Documentation

No new feature introduced.

### Generative AI tooling

Generated-by: Claude Code 1.0.0

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/aggregate/FieldSumAgg.java
+++ b/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/aggregate/FieldSumAgg.java
@@ -55,16 +55,16 @@ public class FieldSumAgg extends FieldAggregator {
                                 mergeFieldDD.scale());
                 break;
             case TINYINT:
-                sum = (byte) ((byte) accumulator + (byte) inputField);
+                sum = addExactByte((byte) accumulator, (byte) inputField);
                 break;
             case SMALLINT:
-                sum = (short) ((short) accumulator + (short) inputField);
+                sum = addExactShort((short) accumulator, (short) inputField);
                 break;
             case INTEGER:
-                sum = (int) accumulator + (int) inputField;
+                sum = Math.addExact((int) accumulator, (int) inputField);
                 break;
             case BIGINT:
-                sum = (long) accumulator + (long) inputField;
+                sum = Math.addExact((long) accumulator, (long) inputField);
                 break;
             case FLOAT:
                 sum = (float) accumulator + (float) inputField;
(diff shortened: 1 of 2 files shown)


JingsongLi (commented):
Good fix for silent data corruption. Using `Math.addExact`/`Math.subtractExact`/`Math.negateExact` for INTEGER and BIGINT, and range-checking helpers for TINYINT/SMALLINT is the correct approach.

**One concern: behavior change for existing users.**

This PR changes SUM aggregation from "silently wrap on overflow" to "throw ArithmeticException". For production tables that currently rely on (perhaps unknowingly) the wrapping behavior, this becomes a hard failure during compaction. Consider:

1. Should there be a configuration option like `fields.*.aggregate-function.sum.overflow-behavior = error|saturate|wrap`? At minimum, the error message should clearly indicate which field overflowed and suggest using BIGINT or DECIMAL if the field needs a wider range.

2. The `ArithmeticException("byte overflow")` message is terse. Consider including the actual values: `"byte overflow: %d + %d"` so users can debug the data that caused the overflow.

3. **FLOAT/DOUBLE** are untouched — they have infinite/NaN overflow behavior per IEEE 754. This is fine and consistent with most systems.

4. Test coverage is thorough — covers all four types, both directions (positive/negative), and retract.

The fix is correct. The behavioral question (fail-loud vs. configurable) is a product decision — if the answer is "always fail," then LGTM as-is.

PR #7340, opened 2026-03-04
[core] Introduce Catalog.listPartitionsByNames

JingsongLi:
### Purpose

Get partitions by names. This can avoid obtaining all partitions.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/rest/RESTCatalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/rest/RESTCatalog.java
@@ -657,14 +657,7 @@ public class RESTCatalog implements Catalog {
         } catch (ForbiddenException e) {
             throw new TableNoPermissionException(identifier, e);
         } catch (NotImplementedException e) {
-            // not a metastore partitioned table, filter from file system partitions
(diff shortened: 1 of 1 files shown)


paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java
```diff
@@ -193,6 +193,13 @@ public PagedList<Partition> listPartitionsPaged(
         return new PagedList<>(listPartitions(identifier), null);
     }
 
+    @Override
+    public List<Partition> listPartitionsByNames(
+            Identifier identifier, List<Map<String, String>> partitions)
+            throws TableNotExistException {
+        return CatalogUtils.listPartitionsFromFileSystem(getTable(identifier), partitions);
+    }
+
```
qingwei727, line 202: Missing the 1000-limit validation that RESTApi.listPartitionsByNames checkde

JingsongLi: The check is just for protect REST server. Lack check here is valid. Let it go.

PR #7352, opened 2026-03-05
[python] support predicate NotBetween, Like function

plusplusjiajia:
### Purpose

- Add predicate negate support
- NotBetween, Like, AlwaysTrue, AlwaysFalse function

### Generative AI tooling
Claude-4.6


JingsongLi (commented):
always_true and always_false are also useless for python. They are just for Spark SQL.

JingsongLi (approved):
+1

PR #7360, opened 2026-03-08
[python] Introduce table import, drop and alter to Python CLI

JingsongLi:
### Purpose

This PR extends the PyPaimon CLI with three new table sub-commands: import, drop, and alter, significantly expanding the CLI's data management capabilities beyond the existing read, get, and create commands.

- table import — Bulk-load data from CSV or JSON files into an existing Paimon table. Reads the file into a pandas DataFrame, converts it to a PyArrow Table matching the target schema, and writes it using the batch write API.

- table drop — Drop a table from the catalog with an optional --ignore-if-not-exists (-i) flag. Implemented in both FileSystemCatalog (deletes the table directory) and the abstract Catalog interface.

- table alter — A rich set of schema alteration sub-commands:
  - set-option / remove-option — Manage table-level key-value options
  - add-column — Add a column with optional --comment, --first, or --after positioning
  - drop-column — Remove a column by name
  - rename-column — Rename an existing column
  - alter-column — Change a column's type, comment, or position (supports combining multiple changes in one command)
  - update-comment — Update the table-level comment

### Generative AI tooling

Claude-4.6

### DIFF
--- a/paimon-python/pypaimon/cli/cli_table.py
+++ b/paimon-python/pypaimon/cli/cli_table.py
@@ -495,7 +495,7 @@ def add_table_subcommands(table_parser):
         help='Table identifier in format: database.table'
     )
     alter_parser.add_argument(
-        '--ignore-if-not-exists',
+        '--ignore-if-not-exists', '-i',
         action='store_true',
         help='Do not raise error if table does not exist'
     )
--- a/docs/content/pypaimon/cli.md
+++ b/docs/content/pypaimon/cli.md
@@ -261,7 +261,7 @@ paimon table alter DATABASE.TABLE [--ignore-if-not-exists] SUBCOMMAND [OPTIONS]
 
 **Global Options:**
 
-- `--ignore-if-not-exists`: Do not raise error if table does not exist
+- `--ignore-if-not-exists, -i`: Do not raise error if table does not exist
 
 #### Set Option
 


paimon-python/pypaimon/cli/cli_table.py
```diff
@@ -185,6 +185,229 @@ def cmd_table_create(args):
...
+        
+        # Convert DataFrame to PyArrow Table with the correct schema
+        table_data = pa.Table.from_pandas(df, schema=pa_schema)
+        
+        # Write data
+        table_write.write_arrow(table_data)
+        
+        # Commit write
+        table_commit.commit(table_write.prepare_commit())
+        table_write.close()
```
plusplusjiajia: It's better to close table_write and table_commit in a finally block to ensure resources are always released.

paimon-python/pypaimon/cli/cli_table.py
```diff
@@ -233,3 +461,89 @@ def add_table_subcommands(table_parser):
...
+    import_parser.set_defaults(func=cmd_table_import)
+    
+    # table alter command
+    alter_parser = table_subparsers.add_parser('alter', help='Alter a table with schema changes')
+    alter_parser.add_argument(
+        'table',
+        help='Table identifier in format: database.table'
+    )
+    alter_parser.add_argument(
+        '--ignore-if-not-exists',
```
plusplusjiajia: --ignore-if-not-exists for table alter doesn't have a -i short option, any particular reason for this?

JingsongLi: Fixed, just because I changed the AI model to claude-4.6.

PR #7364, opened 2026-03-08
[arrow] Fix TIMESTAMP_LTZ Arrow timezone to use UTC instead of system default

plusplusjiajia:
### Purpose
LocalZonedTimestampType stores UTC timestamps by definition. However, ArrowFieldTypeConversion used ZoneId.systemDefault() as the ArrowTimestamp timezone.

### Tests
testTimestampArrowFieldTypeTimezone

### DIFF
--- a/paimon-arrow/src/main/java/org/apache/paimon/arrow/ArrowFieldTypeConversion.java
+++ b/paimon-arrow/src/main/java/org/apache/paimon/arrow/ArrowFieldTypeConversion.java
@@ -49,8 +49,6 @@ import org.apache.arrow.vector.types.Types;
 import org.apache.arrow.vector.types.pojo.ArrowType;
 import org.apache.arrow.vector.types.pojo.FieldType;
 
-import java.time.ZoneId;
-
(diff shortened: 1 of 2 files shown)


leaves12138 (approved):
LGTM. The fix is correct - `LocalZonedTimestampType` semantically stores UTC timestamps, so the Arrow field should use UTC timezone rather than system default. The test also covers both TIMESTAMP_LTZ (should be UTC) and TIMESTAMP (should be null) across multiple precisions.

JingsongLi (commented):
Clean fix. `TIMESTAMP_LTZ` stores UTC by definition, so the Arrow timezone metadata should always be "UTC", not `ZoneId.systemDefault()`.

Using the system default causes incorrect interpretation when the reader is in a different timezone than the writer — a silent data correctness issue.

The fix is minimal (+22/-4) and the test (`testTimestampArrowFieldTypeTimezone`) verifies the correct behavior. LGTM.

JingsongLi:
Please rebase master

plusplusjiajia:
> Please rebase master

@JingsongLi Thanks, done.

PR #7365, opened 2026-03-08
[spark] Fix required configuration check for Paimon SparkSession extensions

lsm1:
### Purpose

   Fix false failures in Paimon required Spark configuration checks by making the check rely on the current SparkSession configuration (instead of a temporary/active SQLConf), so Spark SQL operations (including creating/querying views across session restarts) don’t incorrectly report missing spark.sql.extensions.

### Tests

- Spark SQL manual verification:
```
     create table paimon.default.p1 (
         k int,
         v string
     ) USING paimon
     tblproperties (
         'primary-key' = 'k'
     );
     insert into table paimon.default.p1 select 1,'2';
     create view v2 as  select * from paimon.default.p1;
     --- restart spark-sql session
     select * from v2;
```
 - UT: paimon-spark-ut added/updated cases for required conf checks with
   temporary SQLConf.

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/SparkCatalog.java
+++ b/paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/SparkCatalog.java
@@ -122,8 +122,8 @@ public class SparkCatalog extends SparkBaseCatalog
 
     @Override
     public void initialize(String name, CaseInsensitiveStringMap options) {
-        checkRequiredConfigurations();
         SparkSession sparkSession = PaimonSparkSession$.MODULE$.active();
+        checkRequiredConfigurations(sparkSession);
         this.catalogName = name;
         CatalogContext catalogContext =
                 CatalogContext.create(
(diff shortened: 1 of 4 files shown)


cxzl25:
Could you please help take a look at this issue? It might be preventing Spark from querying views created in Paimon.

@Zouxxyy @JingsongLi

https://github.com/apache/paimon/pull/5327

JingsongLi (commented):
Good fix. The root cause is clear: using a temporary/active `SQLConf` instead of the SparkSession's persistent configuration causes false negatives when checking for Paimon extensions.

**Changes look correct:**
1. `checkRequiredConfigurations()` now takes `SparkSession` parameter and reads from `sparkSession.sessionState.conf` — this survives session restarts and temporary SQLConf scopes.
2. `paimonExtensionEnabled()` also uses SparkSession — consistent.
3. The `requiredSparkConfsCheckEnabled` helper reads from session state conf with proper key prefixing.

**Minor suggestions:**
- The `PAIMON_SPARK_SESSION_EXTENSIONS` constant extraction is a nice cleanup.
- Tests cover both the "extension present but temporary SQLConf" case and the "check disabled" case. Good.

LGTM.

JingsongLi:
+1

PR #7374, opened 2026-03-09
[core] Supoort Option of adding columns before partition

discivigour:
### Purpose

Support 'add-column-before-partition' option to add columns before partition for partition tables.
### Tests
SchemaChangeITCase.testAddColumnBeforePartitionEnabled()


discivigour:
> Also modify Python?

ok.

PR #7377, opened 2026-03-09
[python] Introduce select for CLI table read

JingsongLi:
### Purpose

```shell
paimon table read mydb.users
```

**Options:**

- `--select, -s`: Select specific columns to read (comma-separated)
- `--limit, -l`: Maximum number of results to display (default: 100)

**Examples:**

```shell
# Read with limit
paimon table read mydb.users -l 50

# Read specific columns
paimon table read mydb.users -s id,name,age

# Combine select and limit
paimon table read mydb.users -s id,name -l 50
```

### Generative AI tooling

GLM 5

### DIFF
--- a/paimon-python/pypaimon/tests/cli_table_test.py
+++ b/paimon-python/pypaimon/tests/cli_table_test.py
@@ -154,23 +154,6 @@ class CliTableTest(unittest.TestCase):
         # Simulate CLI command: paimon table read test_db.users --select invalid_column
         with patch('sys.argv',
                    ['paimon', '-c', self.config_file,
-                    'table', 'read', 'test_db.users', '--select', 'invalid_column']):
-            with patch('sys.stderr', new_callable=StringIO) as mock_stderr:
-                try:
-                    main()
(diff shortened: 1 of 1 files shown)


paimon-python/pypaimon/tests/cli_table_test.py
```diff
@@ -128,6 +128,27 @@ def test_cli_table_read_with_limit(self):
...
+                    pass
+                
+                output = mock_stdout.getvalue()
+                
+                # Verify output contains selected columns
+                self.assertIn('id', output.lower())
+                self.assertIn('name', output.lower())
+                # Verify selected data is present
+                self.assertIn('Alice', output)
+                self.assertIn('Bob', output)
```
plusplusjiajia, line 150: better to add some bad cases, such as  the invalid column names are passed

JingsongLi: Good point!

plusplusjiajia:
LGTM, +1

PR #7392, opened 2026-03-10
[core] Support push-down valueStats prune when read manifests.

baiyangtx:
### Purpose

When scanning Paimon tables, each `ManifestEntry` carries per-file column statistics (value stats) — min/max values, null counts, etc. For wide tables with many columns, these statistics can consume significant heap memory during large scans (compaction, queries). In many scan scenarios, the downstream caller does not need these statistics at all.

This PR adds a `dropStats()` option on `FileStoreScan` that prunes value stats from `ManifestEntry` during manifest reading, reducing GC pressure and peak memory usage.

To support this efficiently, two general-purpose improvements are made to the read pipeline:

1. **`loadFilter` for cache read** — `ObjectsFile` / `ObjectsCache` now accept a `loadFilter` parameter. When loading manifest files into the in-memory segment cache, entries not matching the filter are skipped before deserialization, reducing cache memory footprint.

2. **`convertor` for in-place transformation** — A `convertor` function can be passed through the read pipeline, applied to each entry immediately after filtering. This eliminates the need for a secondary allocation-and-copy pass when entries need to be transformed (e.g. dropping stats, projecting fields).

### API Changes

- `AbstractFileStoreScan.readAndMergeFileEntries()` converter signature changed from `Function<List<ManifestEntry>, List<T>>` (batch) to `Function<ManifestEntry, T>` (per-entry). Callers (`readFileEntriesForFullCompaction`, `readFileEntriesForBucketLevel`) are updated accordingly.
- `FileStoreScan.dropStats()` added — when set, `ManifestEntry` value stats (min/max keys, null counts, row count) are cleared during read.

### Changes

| File | Change |
|------|--------|
| `ObjectsFile.java` | Add `loadFilter` + `convertor` to `read()` and `readFromIterator()` |
| `ObjectsCache.java` | Add `loadFilter` to `createSegments()`, propagate `convertor` through cache read path |
| `SimpleObjectsCache.java` | Thread `loadFilter` through segment creation; skip non-matching rows during serialization |
| `ManifestEntryCache.java` | Accept `loadFilter` in `createSegments()` to filter before partitioning into indexed segments |
| `AbstractFileStoreScan.java` | Use `convertor` to apply `dropStats()` during manifest read; pass partition/bucket filters through ManifestFile.read() for correct cache behavior |
| `FileEntry.java` | Update `readDeletedEntries` to new read API |
| `ObjectsCacheTest.java` | Tests for `loadFilter` behavior in cache read path |

### Tests

All existing tests in `paimon-core` pass. `ObjectsCacheTest` extended with new test cases covering `loadFilter` filtering during cache segment creation and cache-hit scenarios.

### Compatibility

No storage format change. `dropStats` defaults to `false`.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreScan.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreScan.java
@@ -491,8 +491,10 @@ public abstract class AbstractFileStoreScan implements FileStoreScan {
             @Nullable Filter<ManifestEntry> additionalTFilter) {
 
         Filter<InternalRow> entryRowFilter = createEntryRowFilter();
+        Function<ManifestEntry, T> finalConverter =
+                dropStats ? e -> converter.apply(dropStats(e)) : converter;
 
-        List<ManifestEntry> entries =
+        List<T> entries =
                 manifestFileFactory
                         .create()
                         .withCacheMetrics(
@@ -507,14 +509,10 @@ public abstract class AbstractFileStoreScan implements FileStoreScan {
                                         (additionalTFilter == null || additionalTFilter.test(entry))
                                                 && (manifestEntryFilter == null
                                                         || manifestEntryFilter.test(entry))
-                                                && filterByStats(entry));
-
-        List<T> result = new ArrayList<>(entries.size());
-        for (ManifestEntry entry : entries) {
-            result.add(converter.apply(dropStats ? dropStats(entry) : entry));
-        }
-        LOG.info("Read {} manifest entries from {}", result.size(), manifest.fileName());
-        return result;
+                                                && filterByStats(entry),
+                                finalConverter);
+        LOG.info("Read {} manifest entries from {}", entries.size(), manifest.fileName());
+        return entries;
     }
 
     protected ManifestEntry dropStats(ManifestEntry entry) {
--- a/paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreScan.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreScan.java
@@ -492,6 +492,8 @@ public abstract class AbstractFileStoreScan implements FileStoreScan {
         Function<ManifestEntry, T> finalConverter =
                 dropStats ? e -> converter.apply(dropStats(e)) : converter;
 
+        Filter<InternalRow> entryRowFilter = createEntryRowFilter();
+
         List<T> entries =
                 manifestFileFactory
                         .create()
@@ -500,9 +502,8 @@ public abstract class AbstractFileStoreScan implements FileStoreScan {
                         .read(
                                 manifest.fileName(),
                                 manifest.fileSize(),
-                                manifestsReader.partitionFilter(),
-                                createBucketFilter(),
-                                createEntryRowFilter().and(additionalFilter),
+                                entryRowFilter,
+                                entryRowFilter.and(additionalFilter),
                                 entry ->
                                         (additionalTFilter == null || additionalTFilter.test(entry))
                                                 && (manifestEntryFilter == null
--- a/paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreScan.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreScan.java
@@ -489,12 +489,10 @@ public abstract class AbstractFileStoreScan implements FileStoreScan {
             Function<ManifestEntry, T> converter,
             @Nullable Filter<InternalRow> additionalFilter,
             @Nullable Filter<ManifestEntry> additionalTFilter) {
-        Function<ManifestEntry, T> finalConverter =
-                dropStats ? e -> converter.apply(dropStats(e)) : converter;
 
         Filter<InternalRow> entryRowFilter = createEntryRowFilter();
 
-        List<T> entries =
+        List<ManifestEntry> entries =
                 manifestFileFactory
                         .create()
                         .withCacheMetrics(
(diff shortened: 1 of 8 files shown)


JingsongLi (commented):
Interesting optimization. Pruning valueStats during manifest reads reduces memory when manifests contain large statistics (e.g., tables with many columns).

**Review:**

1. **Problem**: Manifest entries contain per-column min/max stats (`valueStats`). For wide tables (hundreds of columns), these stats dominate memory even though queries only need stats for a few filter columns.

2. **Solution**: Push-down projection of valueStats to only load statistics for columns referenced in scan predicates.

3. **Files changed**: `ObjectsFile`, `ObjectsCache`, `SimpleObjectsCache`, `ManifestEntryCache`, `FileEntry` — this touches the caching layer, which is sensitive. Ensure:
   - Cached entries with full stats aren't served when pruned stats are requested
   - The cache key accounts for the projection (otherwise a pruned entry could be returned for a different query that needs different stats)

4. **+195/-47** — moderate change. The refactoring of `ObjectsCache` to support projection is the core change.

5. **Correctness risk**: If pruning is too aggressive (e.g., removes stats needed by file-index evaluation or partition predicate push-down), it could cause incorrect scan results. Ensure the pruning only removes stats for columns NOT referenced anywhere in the read path.

6. **PR title typo**: "purne" → "prune"

Please verify cache key correctness with projections and add a test that validates stats are complete for filter-referenced columns.

baiyangtx:
Thanks for the review, @JingsongLi 

**Cache key:** The cache stores raw `InternalRow` bytes, not deserialized objects. Deserialization, filtering, and conversion run fresh on every cache hit. `dropStats` is applied as a post-processing loop after the cache read returns, so the same cached bytes are safe for both `dropStats=true` and `dropStats=false` queries.

**Correctness:** `filterByStats(entry)` executes inside `ManifestFile.read()` before the post-processing loop, so it always sees complete stats. And `dropStats` defaults to `false` — unaffected paths remain untouched.

Fixed the PR title typo as well.

Aitozi (approved):
Thanks @baiyangtx for the PR. LGTM +1

baiyangtx:
@JingsongLi Thanks for the review — addressed all the comments. Could you take another look when you have time?

paimon-core/src/main/java/org/apache/paimon/utils/ObjectsCache.java
```diff
@@ -78,22 +98,38 @@ public List<V> read(K key, @Nullable Long fileSize, Filters<V> filters) throws I
                 fileSize = fileSizeFunction.apply(key);
             }
             if (fileSize <= cache.maxElementSize()) {
-                segments = createSegments(key, fileSize);
+                segments = createSegments(key, fileSize, loadFilter);
```
JingsongLi: This makes the cache unsafe for any selective `loadFilter`: the cache key is still only `key`, so a cold read can store a filtered subset of the file, and a later read of the same key with a broader/different filter will hit that partial cache and silently miss rows. The new test currently asserts the first filtered read only; it should also read the same key again with `Filter.alwaysTrue()` and verify all rows are still returned. Could we avoid putting filtered segments into this cache, or include the load filter/projection in the cache identity?

paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreScan.java
```diff
@@ -493,20 +502,19 @@ private List<ManifestEntry> readManifest(
...
-        if (dropStats) {
-            List<ManifestEntry> copied = new ArrayList<>(entries.size());
-            for (ManifestEntry entry : entries) {
-                copied.add(dropStats(entry));
-            }
-            entries = copied;
+
+        List<T> result = new ArrayList<>(entries.size());
+        for (ManifestEntry entry : entries) {
+            result.add(converter.apply(dropStats ? dropStats(entry) : entry));
```
JingsongLi: This still materializes the full `List<ManifestEntry>` from `ManifestFile.read()` before applying `dropStats`/`converter`, so peak memory during manifest read still includes all `valueStats`. That seems to miss the main goal of pushing the prune down into manifest reading: callers using `dropStats()` can still OOM before this loop gets a chance to copy the entries without stats.

JingsongLi (commented):
Took another look at the latest revision. Two concerns below.

JingsongLi (commented):
Let me take a look again.

JingsongLi (commented):
I think there is still a cache-semantics problem in the new `loadFilter` path.

`ObjectsCache.read(key, fileSize, loadFilter, readFilter, readVFilter, convertor)` applies `loadFilter` while building the cached segments, but the cache key is still only `key`. This means the first filtered read can populate the shared cache with only a subset of the object, and a later broader read of the same key will hit that incomplete cache.

A concrete example is the new `ObjectsCacheTest` setup: if the first read of `k3` uses `loadFilter = row -> row.endsWith("2")`, the cache stores only `v2` under `k3`. A following read of `k3` with `loadFilter/readFilter = alwaysTrue()` will hit the cache and return only `v2`, even though the underlying object contains `v1, v2, v3`. That changes the old cache contract from “cache the full object and apply per-read filters on top” to “cache the first reader's filtered subset”.

The current `ManifestFile.read(...)` path avoids the worst version of this by passing `Filter.alwaysTrue()` as `loadFilter`, but the public `ObjectsFile` / `ObjectsCache` API now exposes the unsafe behavior. I think we should either avoid populating the shared cache when `loadFilter` is not full-scan, include the load-filter identity in the cache key, or remove `loadFilter` from cached segment creation and keep it only on non-cached reads.

Related: `ManifestFile.read(...)` now constructs a `ManifestEntryFilters`, but then calls the new `cache.read(...)` overload with only `filters.readFilter()` and `filters.readVFilter()`. `ObjectsCache` wraps those back into a plain `Filters`, so `ManifestEntryCache.readFromSegments` no longer sees the original `partitionFilter` / `bucketFilter` and cannot use its indexed segment lookup. `AbstractFileStoreScan` duplicates partition/bucket checks in `readFilter`, so the common scan path should still be correct, but this loses the previous manifest-cache fast path and makes the cache-specific filtering semantics more fragile.

baiyangtx:
@JingsongLi Good points. Updated:

- Removed the `loadFilter` parameter from `ObjectsCache.createSegments` — the cache now always stores full objects, so filtered subsets can never pollute the shared cache.
- Reverted the 6-param `read()` overload that was unwrapping filters. Added a minimal `read(key, fileSize, Filters, convertor)` overload instead, so `ManifestEntryFilters` passes through intact and `ManifestEntryCache` retains its indexed partition/bucket lookup.

PTAL when you have time.

JingsongLi (approved):
+1

PR #7407, opened 2026-03-11
[core] Introduce SimpleLsmKvDb to put and get based on Paimon SST

JingsongLi:
### Purpose

```
/**
 * A simple LSM-Tree based KV database built on top of {@link SortLookupStoreFactory}.
 *
 * <p>Architecture (Universal Compaction, inspired by RocksDB):
 *
 * <pre>
 *     ┌──────────────────────────────────────────────┐
 *     │            MemTable (SkipList)                │  ← Active writes
 *     ├──────────────────────────────────────────────┤
 *     │  Sorted Runs (newest → oldest):              │
 *     │    [Run-0] [Run-1] [Run-2] ... [Run-N]       │  ← Each run is a sorted SST file set
 *     └──────────────────────────────────────────────┘
 * </pre>
 *
 * <p>Compaction is triggered when the number of sorted runs exceeds a threshold. Runs are selected
 * for merging based on size ratios between adjacent runs, following RocksDB's Universal Compaction
 * strategy.
 *
 * <p>Note: No WAL is implemented. Data in the MemTable will be lost on crash.
 *
 * <p>This class is <b>not</b> thread-safe. External synchronization is required if accessed from
 * multiple threads.
 */
```

### Generative AI tooling

Claude 4.6 Opus

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/lookup/sort/db/SimpleLsmKvDb.java
+++ b/paimon-common/src/main/java/org/apache/paimon/lookup/sort/db/SimpleLsmKvDb.java
@@ -484,25 +484,25 @@ public class SimpleLsmKvDb implements Closeable {
             this.dataDirectory = dataDirectory;
         }
 
-        /** Set the MemTable flush threshold in bytes. Default is 4 MB. */
+        /** Set the MemTable flush threshold in bytes. Default is 64 MB. */
         public Builder memTableFlushThreshold(long thresholdBytes) {
             this.memTableFlushThreshold = thresholdBytes;
             return this;
         }
 
-        /** Set the maximum SST file size produced by compaction in bytes. Default is 64 MB. */
+        /** Set the maximum SST file size produced by compaction in bytes. Default is 8 MB. */
         public Builder maxSstFileSize(long maxSstFileSize) {
             this.maxSstFileSize = maxSstFileSize;
             return this;
         }
 
-        /** Set the SST block size in bytes. Default is 4 KB. */
+        /** Set the SST block size in bytes. Default is 32 KB. */
         public Builder blockSize(int blockSize) {
             this.blockSize = blockSize;
             return this;
         }
 
-        /** Set the block cache size in bytes. Default is 8 MB. */
+        /** Set the block cache size in bytes. Default is 128 MB. */
         public Builder cacheSize(long cacheSize) {
             this.cacheSize = cacheSize;
             return this;


paimon-common/src/main/java/org/apache/paimon/lookup/sort/db/SimpleLsmKvDb.java
```diff
@@ -0,0 +1,573 @@
...
+        private int level0FileNumCompactTrigger = 4;
+        private int sizeRatio = 10;
+        private CompressOptions compressOptions = CompressOptions.defaultOptions();
+        private Comparator<MemorySlice> keyComparator = MemorySlice::compareTo;
+
+        Builder(File dataDirectory) {
+            this.dataDirectory = dataDirectory;
+        }
+
+        /** Set the MemTable flush threshold in bytes. Default is 4 MB. */
```
leaves12138: Default is 64 MB?

paimon-common/src/main/java/org/apache/paimon/lookup/sort/db/SimpleLsmKvDb.java
```diff
@@ -0,0 +1,573 @@
...
+        public Builder memTableFlushThreshold(long thresholdBytes) {
+            this.memTableFlushThreshold = thresholdBytes;
+            return this;
+        }
+
+        /** Set the maximum SST file size produced by compaction in bytes. Default is 64 MB. */
```
leaves12138: Default is 8MB?

paimon-common/src/main/java/org/apache/paimon/lookup/sort/db/SimpleLsmKvDb.java
```diff
@@ -0,0 +1,573 @@
...
+        public Builder maxSstFileSize(long maxSstFileSize) {
+            this.maxSstFileSize = maxSstFileSize;
+            return this;
+        }
+
+        /** Set the SST block size in bytes. Default is 4 KB. */
```
leaves12138: Default is 32 KB?

paimon-common/src/main/java/org/apache/paimon/lookup/sort/db/SimpleLsmKvDb.java
```diff
@@ -0,0 +1,573 @@
...
+        public Builder blockSize(int blockSize) {
+            this.blockSize = blockSize;
+            return this;
+        }
+
+        /** Set the block cache size in bytes. Default is 8 MB. */
```
leaves12138: 128 MB

PR #7411, opened 2026-03-12
[core] Add Compaction Sort Buffer metrics

mikedias:
### Purpose
Adding Compaction Sort Buffer metrics to gain visibility in the current sort buffer usage to assist with tuning the `[sort-spill-threshold](https://paimon.apache.org/docs/1.3/primary-key-table/compaction/#compaction-options) option.

### Tests

Added unit tests 

### API and Format

No changes in the API or Format

### Documentation

Updated [metrics.md](docs/content/maintenance/metrics.md) with the new metrics

### Generative AI tooling

Assisted by Rovodev and Claude Sonnet 4.6

### DIFF
--- a/docs/content/maintenance/metrics.md
+++ b/docs/content/maintenance/metrics.md
@@ -294,6 +294,26 @@ Below is lists of Paimon built-in metrics. They are summarized into types of sca
             <td>Gauge</td>
             <td>The average total file size of all active (currently being written) buckets.</td>
         </tr>
+        <tr>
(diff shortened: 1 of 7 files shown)


JingsongLi (approved):
+1

steFaiz:
@mikedias  Hi, do you find these metrics useful?  I casually found `memoryPool` of `MergeSorter` is actually a legacy code, which is never used. I think the metrics would be always zero.

Issue #5179, opened 2025-02-28
[Bug] aggregate function in partial update inconsistency


fantasy2100:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Paimon version

1.0.0

### Compute Engine

Flink

### Minimal reproduce step

CREATE TABLE t1
(
    k   INT,
    a   INT,
    b   INT,
    g_1 INT,
    PRIMARY KEY (k) NOT ENFORCED
) WITH (
   'merge-engine' = 'partial-update',
   'fields.a.aggregate-function' = 'sum',
   'fields.b.aggregate-function' = 'last_non_null_value',
   'fields.g_1.sequence-group' = 'a,b'
);

INSERT INTO t1 VALUES (1, 1, 1, 1);
INSERT INTO t1 VALUES (1, 2, 2, 2);
INSERT INTO t1 VALUES (1, 3, 3, 1);

### What doesn't meet your expectations?

when g_1 of third record is smaller,  third record should be ignored。
I expect the result to be: 1,3,2,2

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

fantasy2100:
"SELECT * FROM t1" 
current output: 1,6,2,2
 expect output: 1,3,2,2

JingsongLi:
It is sum... should be 6...

fantasy2100:
> It is sum... should be 6...

sum and last_non_null_value are both aggregate functions.
The result of sum is 6, why is the result of last_non_null_value not 3?

LinMingQiang:
> > It is sum... should be 6...
> 
> sum and last_non_null_value are both aggregate functions. The result of sum is 6, why is the result of last_non_null_value not 3?

`sum` agg does not care about the order, but `last_non_null_value ` needs.

dark2momo:
My understanding of aggregation for partial update is as follows: 
aggregate the specified fields using the aggregate function (fields.<field-name>.aggregate-function) after reordering the rows based on the sequence-group field (fields.<field-name>.sequence-group). If the sequence-group field is NULL, discard the corresponding record.
And, when the value of the sequence-group fields is equal to that of the preceding row, the corresponding field value to aggregate takes precedence over the previous one. A typical example of this behavior is the first_value function. 
Perhaps the documentation and test cases should be updated.

fantasy2100:
> > > It is sum... should be 6...
> > 
> > 
[...]

PR #7420, opened 2026-03-13
[docs] Clarify sequence-group behavior when combined with aggregate functions

dyzcs:
### Purpose

Linked issue: related to  #5179

Users are frequently confused when `sequence-group` is combined with aggregate functions (e.g. `sum`, `last_non_null_value`), expecting it to act as a version filter (records with a smaller sequence value are ignored). However, when aggregate functions are involved, the sequence-group field acts as an **ordering key**: every record with a non-NULL sequence value participates in the aggregation, regardless of whether its value is larger or smaller than the currently stored one.

This PR clarifies the distinction by:
1. Adding an informational note at the top of the "Aggregation For Partial Update" section that explicitly explains the behavioral difference between using sequence-group with and without aggregate functions.
2. Fixing a misleading code comment `-- a should not be updated` in an existing example, which contradicted the shown output (`a=6` via sum aggregation).

### Tests

Documentation-only change. No tests required.

### API and Format

### Documentation

This PR itself is a documentation improvement. No additional documentation needed.

### Generative AI tooling

Generated-by: Claude (Cursor AI assistant), used to help draft and review the documentation wording.

### DIFF
--- a/docs/content/primary-key-table/merge-engine/partial-update.md
+++ b/docs/content/primary-key-table/merge-engine/partial-update.md
@@ -152,6 +152,22 @@ FROM SG;
 You can specify aggregation function for the input field, all the functions in the
 [Aggregation]({{< ref "primary-key-table/merge-engine/aggregation" >}}) are supported.
 
+{{< hint info >}}
+**Sequence-group behavior changes when aggregate functions are involved.**
+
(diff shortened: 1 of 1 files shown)


dyzcs:
Hi @JingsongLi, could you take a look when you have time? It's a small docs clarification for #5179.

Issue #7421, opened 2026-03-13
[Feature] Support table properties in jdbc metastore


nickdelnano:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Motivation

Similar to HiveCatalog, it is helpful for a platform owner to see table properties for all tables

HiveCatalog:  https://github.com/apache/paimon/pull/3973

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #7423, opened 2026-03-13
[core] Add table properties sync to JDBC catalog

nickdelnano:
### Purpose
Linked issue:  https://github.com/apache/paimon/issues/7421

Supporting table properties in JdbcCatalog similar to HiveCatalog support. Some implementation details are different due to differences in Hive vs Jdbc.

Table properties are stored in SQL table `paimon_table_properties` controlled by the existing `sync-all-properties` option.

### Tests

Unit tests in JdbcCatalogTest. These use in memory sqlite database so they function like integration tests.

### API and Format

N/A

### Documentation

Updated docs

### Generative AI tooling
Generated-by: claude code

I have reviewed the code.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/jdbc/JdbcCatalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/jdbc/JdbcCatalog.java
@@ -18,6 +18,7 @@
 
 package org.apache.paimon.jdbc;
 
+import org.apache.paimon.CoreOptions;
 import org.apache.paimon.annotation.VisibleForTesting;
 import org.apache.paimon.catalog.AbstractCatalog;
 import org.apache.paimon.catalog.CatalogContext;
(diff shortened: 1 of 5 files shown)


paimon-core/src/main/java/org/apache/paimon/jdbc/JdbcCatalog.java
```diff
@@ -486,14 +554,58 @@ public void repairTable(Identifier identifier) throws TableNotExistException {
...
+                    CoreOptions.PARTITION.key(),
+                    String.join(",", tableSchema.partitionKeys()));
+        }
+        if (!tableSchema.bucketKeys().isEmpty()) {
+            properties.put(
+                    CoreOptions.BUCKET_KEY.key(),
+                    String.join(",", tableSchema.bucketKeys()));
+        }
+        return properties;
+    }
```
nickdelnano, line 594: https://github.com/apache/paimon/blob/master/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java#L1682-L1698

nickdelnano:
@JingsongLi can you review?

PR #7429, opened 2026-03-16
[core] Do not clean base and delta manifests in ChangelogDeletion

tsreaper:
### Purpose

Currently, `ChangelogDeletion` will clean base and delta manifests. However these manifests are also cleaned in `SnapshotDeletion`.

Consider the following scenario:

* `snapshot-1` is changed to `changelog-1`. However due to some issue, its base and delta manifests are not cleaned.
* `snapshot-1` is changed to `changelog-2`. Its base and delta manifests are cleaned.
* When expiring `changelog-1`, as `changelog-2` has no manifests, all manifests in `changelog-1` will be cleaned, even if they're still in use by later snapshots.

This PR fixes the bug by not cleaning base and delta manifests in `ChangelogDeletion`.

### Tests

* `ChangelogExpireTest#testChangelogExpireWithUncleanedManifestLists`.

### API and Format

No format changes.

### Documentation

No new feature.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/ChangelogDeletion.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/ChangelogDeletion.java
@@ -78,6 +78,8 @@ public class ChangelogDeletion extends FileDeletionBase<Changelog> {
             if (manifestList.exists(changelog.deltaManifestList())) {
                 cleanUnusedManifestList(changelog.deltaManifestList(), skippingSet);
             }
+            // See FileDeletionBase#cleanUnusedManifests
+            // about why we need to clean base manifest
             if (manifestList.exists(changelog.baseManifestList())) {
                 cleanUnusedManifestList(changelog.baseManifestList(), skippingSet);
             }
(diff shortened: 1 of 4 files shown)


paimon-core/src/main/java/org/apache/paimon/operation/ChangelogDeletion.java
```diff
@@ -63,25 +63,24 @@ public ChangelogDeletion(
...
-        }
-
-        if (manifestList.exists(changelog.baseManifestList())) {
-            cleanUnusedManifestList(changelog.baseManifestList(), skippingSet);
+        } else {
+            if (manifestList.exists(changelog.deltaManifestList())) {
+                cleanUnusedManifestList(changelog.deltaManifestList(), skippingSet);
+            }
+            if (manifestList.exists(changelog.baseManifestList())) {
+                cleanUnusedManifestList(changelog.baseManifestList(), skippingSet);
```
JingsongLi, line 84: Why clean base here, it is very dangerous.

PR #7445, opened 2026-03-17
[core] refactor flink create global index

jerry-024:
### Purpose

  Support building generic global indexes (e.g. lumina-vector-ann) through Flink with parallel execution topology, aligned with the existing Spark implementation. This PR introduces `GenericIndexTopoBuilder` which builds a Flink streaming topology:
  `Source(1) → BuildIndexOperator(N) → CommitOperator(1)`, where each shard is an independent Flink task.

  **Key design decisions:**
  - **File-level sharding** (aligned with Spark's `groupFilesIntoShardsByPartition`): files are assigned to shards by `firstRowId / rowsPerShard`, avoiding redundant I/O
  - **Per-shard sub-range**: each shard gets its own `Range` so the reader creates one index file per shard, merged via `UnionGlobalIndexReader` at query time
  - **Defensive checks**: validates `bucket=-1` and `deletion-vectors.enabled=false`, since the sequential row-id design assumes append-only unaware-bucket tables

  ### Tests

  - `LuminaVectorGlobalIndexITCase` 
  - `GenericIndexTopoBuilderTest` 

  ### API and Format

  - New config option `global-index.build.max-parallelism` (default: 4096) — caps the number of parallel Flink tasks for generic index building
  - No storage format changes

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/globalindex/GenericIndexTopoBuilder.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/globalindex/GenericIndexTopoBuilder.java
@@ -89,7 +89,7 @@ public class GenericIndexTopoBuilder {
 
     private static final Logger LOG = LoggerFactory.getLogger(GenericIndexTopoBuilder.class);
 
-    public static void buildIndex(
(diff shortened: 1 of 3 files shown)


leaves12138 (approved):
+1

PR #7456, opened 2026-03-17
[python] Add streaming reads to paimon CLI (`table stream` command)

tub:
## Summary

- Adds `paimon table stream <db.table>` CLI subcommand that continuously polls a Paimon table and prints new rows as they arrive until Ctrl+C
- Adds `StreamReadBuilder.with_scan_from()` to the library so programmatic users can also control starting position (`"latest"`, `"earliest"`, or a snapshot ID integer)
- Timestamp support in `--from` is resolved to a snapshot ID at the CLI layer (no timestamps in the library API)
  - I can push this down into the library if we think it'll be useful

## Flags

| Flag | Default | Description |
|---|---|---|
| `--from` | `latest` | Starting position: `latest`, `earliest`, snapshot ID, or timestamp (`YYYY-MM-DD`, ISO 8601) |
| `--select` | — | Column projection (comma-separated) |
| `--where` | — | SQL-subset filter predicate |
| `--format` | `table` | Output format: `table` or `json` |
| `--poll-interval-ms` | `1000` | Milliseconds between snapshot polls |
| `--include-row-kind` | off | Prepend `_row_kind` column (`+I`, `-U`, `+U`, `-D`) |
| `--consumer-id` | — | Persist scan progress for at-least-once resume |

## Changes

- `paimon-python/pypaimon/cli/cli_table_stream.py` — new command handler + `parse_from_position()` timestamp resolver
- `paimon-python/pypaimon/cli/cli_table.py` — register `stream` subparser
- `paimon-python/pypaimon/read/stream_read_builder.py` — `with_scan_from()`
- `paimon-python/pypaimon/read/streaming_table_scan.py` — `scan_from` startup resolution (consumer restore always wins)

## Test plan

- [x] `pypaimon/tests/stream_read_builder_test.py` — 7 new unit tests for `with_scan_from()`
- [x] `pypaimon/tests/streaming_table_scan_test.py` — 4 new integration tests (`earliest`, numeric ID, `latest`, consumer-overrides-scan_from)
- [x] `pypaimon/tests/cli_table_stream_test.py` — 20 new tests (CLI integration + `parse_from_position` unit tests)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/paimon-python/pypaimon/cli/cli_table_stream.py
+++ b/paimon-python/pypaimon/cli/cli_table_stream.py
@@ -23,52 +23,38 @@ Continuously polls a table for new snapshots and prints rows as they arrive.
 
 import json
 import sys
-from datetime import datetime, timezone, timedelta
 from typing import Union
 
 from pypaimon.snapshot.snapshot_manager import SnapshotManager
 
 
-# Timestamp format strings tried in order during parse_from_position.
-_TIMESTAMP_FORMATS = [
-    "%Y-%m-%dT%H:%M:%S%z",  # ISO 8601 with offset (e.g. +05:30 or Z via strptime %z)
-    "%Y-%m-%dT%H:%M:%S",    # ISO 8601 naive
-    "%Y-%m-%d %H:%M:%S",    # space-separated naive
-    "%Y-%m-%d",             # date only
-]
-
-
 def _parse_timestamp(value: str) -> int:
     """Parse a human-readable datetime string to epoch milliseconds.
 
+    Accepts ISO 8601, space-separated datetimes, date-only strings, and named
+    timezones (e.g. "2025-01-15T10:30:00 EST", "2025-01-15T10:30:00 Europe/London").
     Naive datetimes (no timezone) are treated as local machine time.
     Returns epoch ms as an integer.
     Raises ValueError on unrecognised format.
     """
-    # Handle the 'Z' suffix — Python <3.11 strptime doesn't recognise 'Z' as UTC
-    normalised = value
-    if normalised.endswith("Z"):
-        normalised = normalised[:-1] + "+00:00"
+    from dateutil import parser as dateutil_parser
+    from dateutil.parser import ParserError
 
-    for fmt in _TIMESTAMP_FORMATS:
-        try:
-            dt = datetime.strptime(normalised, fmt)
-            break
-        except ValueError:
-            continue
-    else:
+    try:
+        dt = dateutil_parser.parse(value)
+    except (ParserError, OverflowError):
         raise ValueError(
             f"Unrecognised timestamp format: '{value}'. "
             "Expected formats: YYYY-MM-DD, YYYY-MM-DD HH:MM:SS, "
-            "YYYY-MM-DDTHH:MM:SS, YYYY-MM-DDTHH:MM:SSZ, or YYYY-MM-DDTHH:MM:SS+HH:MM"
+            "YYYY-MM-DDTHH:MM:SS, YYYY-MM-DDTHH:MM:SSZ, YYYY-MM-DDTHH:MM:SS+HH:MM, "
+            "or a datetime with a named timezone (e.g. '2025-01-15T10:30:00 EST')"
         )
 
     if dt.tzinfo is None:
         # Treat naive datetime as local time
         dt = dt.astimezone()
 
-    epoch_ms = int(dt.timestamp() * 1000)
-    return epoch_ms
+    return int(dt.timestamp() * 1000)
 
 
 def parse_from_position(value: str, snapshot_manager: SnapshotManager) -> Union[str, int]:
--- a/paimon-python/pypaimon/cli/cli_table_stream.py
+++ b/paimon-python/pypaimon/cli/cli_table_stream.py
@@ -1,0 +1,230 @@
+################################################################################
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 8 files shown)


JingsongLi (commented):
Nice CLI addition for pypaimon. The `table stream` command provides a user-friendly way to tail table changes.

**Review:**

1. **Feature set**: `--from`, `--select`, `--where`, `--format`, `--poll-interval-ms`, `--include-row-kind`, `--consumer-id` — comprehensive flag set covering the common streaming read scenarios.

2. **Library API addition**: `StreamReadBuilder.with_scan_from()` is useful beyond CLI. Good separation between library and CLI layers.

3. **+859 additions** is substantial. The split between CLI handler, library API, and tests is clean.

4. **Consumer-id for at-least-once resume**: Persisting scan progress enables resumable tailing. Important for operational use.

5. **Timestamp resolution at CLI layer**: Making the library API accept snapshot IDs while the CLI resolves timestamps is the right layering. Pushing timestamp resolution down to the library is optional.

6. **Test coverage**: `cli_table_stream_test.py`, `stream_read_builder_test.py`, `streaming_table_scan_test.py` — three test files covering different layers. Good.

7. **Poll interval**: Default 1000ms is reasonable. Consider documenting that very low intervals (< 100ms) may cause unnecessary load on the metadata path.

LGTM. Well-designed feature.

JingsongLi:
Please make this not a draft.

leaves12138 (commented):
Thanks for the update. I reviewed the latest version again.

The CLI layering looks reasonable now: `table stream` keeps the human-friendly parsing in the command layer, resolves timestamp-based `--from` values to snapshot IDs, and passes a normalized scan position down to `StreamReadBuilder`. The streaming scan behavior also looks consistent to me: consumer restore takes precedence over `scan_from`, `earliest` emits an initial full-scan plan from the earliest snapshot, and numeric snapshot IDs rely on the follow-up scan path.

The new tests cover the main CLI paths, output formats, `--from` parsing, builder propagation, and the streaming scan startup semantics. I do not see a blocking code issue from this pass.

One small follow-up you may consider (not blocking from my side): since `StreamReadBuilder.with_scan_from()` is a public programmatic API and its docstring restricts values to `"latest"`, `"earliest"`, or a positive integer snapshot ID, it would be a bit nicer to fail fast there for invalid strings or non-positive integers instead of letting the scan loop fail later. The CLI already normalizes its inputs, so this is mostly API hardening.

I noticed the latest Python lint jobs are still in progress, so I am leaving this as a review comment rather than approval for now. If those checks turn green, this looks good to merge from my side.

tub:
Rebased & fixed up the tests, sorry for the delay, was out of the office for a while.

paimon-python/pypaimon/cli/cli_table_stream.py
```diff
@@ -0,0 +1,216 @@
...
+            print(f"Error: {e}", file=sys.stderr)
+            sys.exit(1)
+    elif from_position.strip().isdigit():
+        from_position = int(from_position.strip())
+
+    # --- Build StreamReadBuilder ---
+    builder = table.new_stream_read_builder()
+
+    if user_columns:
+        builder = builder.with_projection(user_columns)
```
JingsongLi, line 175: When --select is combined with a --where predicate that references a non-selected column, this projection drops the predicate column from the read schema. The reader then cannot evaluate the row predicate and returns unfiltered rows. For example, --from earliest --select id --where "age > 18" returns both ids for rows (1, age=10) and (2, age=30). Please mirror table read here: add WHERE-referenced fields to the internal projection, then drop those extra columns from the printed output.

paimon-python/pypaimon/read/streaming_table_scan.py
```diff
@@ -130,13 +134,28 @@ async def stream(self) -> AsyncGenerator[Plan, None]:
...
             consumer = self._consumer_manager.consumer(self._consumer_id)
             if consumer:
                 self.next_snapshot_id = consumer.next_snapshot
 
-        # Initial scan
+        # Resolve scan_from if no position has been set yet (no consumer restore)
+        if self.next_snapshot_id is None and self._scan_from is not None:
+            scan_from = self._scan_from
+            if scan_from == "earliest":
+                earliest_snapshot = self._snapshot_manager.try_get_earliest_snapshot()
```
JingsongLi, line 147: try_get_earliest_snapshot() raises RuntimeError when the table has no snapshots; it does not return None. That means table stream --from earliest on a newly created empty table exits before reaching the polling loop, despite the comment below. Please handle the no-snapshot case before or around this call so the stream waits for the first snapshot.

paimon-python/pypaimon/cli/cli_table_stream.py
```diff
@@ -0,0 +1,216 @@
...
+    # Anything containing '-', 'T', ':', or '/' is treated as a timestamp
+    if any(c in value for c in ("-", "T", ":", "/")):
+        epoch_ms = _parse_timestamp(value)
+        snapshot = snapshot_manager.earlier_or_equal_time_mills(epoch_ms)
+        if snapshot is None:
+            raise ValueError(
+                f"No snapshot found at or before '{value}'. "
+                "The timestamp may predate all stored snapshots."
+            )
+        return snapshot.id
```
JingsongLi, line 89: `AsyncStreamingTableScan` treats a numeric `scan_from` as the next snapshot to emit, so resolving a timestamp with `earlier_or_equal_time_mills` and returning that same snapshot id makes `--from <timestamp>` include data from before the requested time. For example, with S1 at 10:00 and S2 at 11:00, `--from 10:30` resolves to S1 and then streams S1. Please either return `snapshot.id + 1` for timestamp inputs, or resolve to the first snapshot at/after the timestamp so timestamp starts do not replay an older commit.

JingsongLi:

PR #7460, opened 2026-03-18
[python] Fix blob descriptor conversion missing in to_arrow_batch_reader

XiaoHongbo-Hope:
### Purpose

Linked issue: close #xxx

### DIFF
--- a/paimon-python/pypaimon/read/table_read.py
+++ b/paimon-python/pypaimon/read/table_read.py
@@ -20,7 +20,6 @@ from typing import Any, Dict, Iterator, List, Optional
 import pandas
 import pyarrow
 
-from pypaimon.common.options.core_options import CoreOptions
 from pypaimon.common.predicate import Predicate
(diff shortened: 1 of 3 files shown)


paimon-python/pypaimon/read/table_read.py
```diff
@@ -159,6 +159,20 @@ def _convert_descriptor_stored_fields_for_read(self, table: pyarrow.Table) -> py
             )
         return result
 
+    def _convert_descriptor_stored_fields_for_read_batch(
```
JingsongLi: Maybe we can move this to `DataEvolutionSplitRead` now.

XiaoHongbo-Hope: > Maybe we can move this to `DataEvolutionSplitRead` now.

Moved the conversion logic into a separate class `DataEvolutionSplitRead`, which wraps the reader in `create_reader` of `DataEvolutionSplitRead`. Do you think it looks good?

PR #7461, opened 2026-03-18
[core][python] Add a version number in BtreeFileFooter

steFaiz:
### Purpose
This PR introduce a version number in BTreeFileFooter.  This is for future BTree File Format compatibility. Since no release version has been published yet, we can add a new version field to BTreeFileFooter.

For safety, also modified the magic number of BTreeIndexFile to 0x50425449, which is the ascii code of 'PBTI'(Paimon BTree Index)

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/globalindex/btree/BTreeFileFooter.java
+++ b/paimon-common/src/main/java/org/apache/paimon/globalindex/btree/BTreeFileFooter.java
@@ -31,7 +31,7 @@ import static org.apache.paimon.utils.Preconditions.checkArgument;
 /** The Footer for BTree file. */
 public class BTreeFileFooter {
 
-    public static final int MAGIC_NUMBER = 198732882;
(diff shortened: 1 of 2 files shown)


paimon-common/src/main/java/org/apache/paimon/globalindex/btree/BTreeFileFooter.java
```diff
@@ -32,8 +32,10 @@
 public class BTreeFileFooter {
 
     public static final int MAGIC_NUMBER = 198732882;
```
JingsongLi: Even no release, we should modify MAGIC_NUMBER to ensure safety.

steFaiz: @JingsongLi  Thanks! I've modified the magic number to 0x50425449, which is the ascii code of 'PBTI'(Paimon BTree Index)

PR #7470, opened 2026-03-18
[core] Ensuring JDBC connection is not stale before using it

junmuz:
### Purpose

This change ensure connection is valid before it is used for locking or cataloging purposes. Changes made include

  1. paimon-common/.../client/ClientPool.java

  - Added a reconnect(C client) hook in ClientPoolImpl that subclasses can override
  - Called it in run() before executing the action — the reassignment to client ensures the valid connection is what goes back into the deque via addFirst

  2. paimon-core/.../jdbc/JdbcClientPool.java

  - Stored connectionSupplier as a field for creating fresh connections
  - Overrode reconnect() to validate via isClosed() and isValid(5), replacing stale connections
  - Added closeQuietly() helper for safe cleanup of dead connections

  3. paimon-core/.../jdbc/JdbcClientPoolTest.java (new)

  - testValidConnectionIsReused — a healthy connection is returned as-is from the pool (same object identity)
  - testClosedConnectionIsReplaced — a closed connection triggers creation of a fresh one
  - testReplacedConnectionIsReturnedToPool — the fresh replacement goes back into the pool and is reused on subsequent calls
  - testActionIsExecutedOnValidConnection — after replacing a stale connection, the action can execute real SQL successfully

Issue: [7474](https://github.com/apache/paimon/issues/7474)

### Tests

There are new tests added that will ensure the change works as expected

### API and Format

N/A

### Documentation

No

### Generative AI tooling
Generated-by: Claude with opus-4.6

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/client/ClientPool.java
+++ b/paimon-common/src/main/java/org/apache/paimon/client/ClientPool.java
@@ -65,7 +65,7 @@ public interface ClientPool<C, E extends Exception> {
                     continue;
                 }
                 try {
-                    client = reconnect(client);
+                    client = ensureActiveClient(client);
                     return action.run(client);
                 } finally {
                     clients.addFirst(client);
@@ -73,8 +73,11 @@ public interface ClientPool<C, E extends Exception> {
             }
         }
 
-        /** Validate and potentially replace a client before use. */
-        protected C reconnect(C client) {
+        /**
+         * Validates that the client is still active before use. Subclasses should override this
+         * method to provide protocol-specific validation and reconnection logic.
+         */
+        protected C ensureActiveClient(C client) {
             return client;
         }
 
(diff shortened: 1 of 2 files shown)


junmuz:
@JingsongLi can you review the PR? The details are mentioned in the issue and there are test cases for covering the scenario.

paimon-common/src/main/java/org/apache/paimon/client/ClientPool.java
```diff
@@ -65,13 +65,19 @@ public <R> R run(Action<R, C, E> action) throws E, InterruptedException {
...
+                    client = reconnect(client);
                     return action.run(client);
                 } finally {
                     clients.addFirst(client);
                 }
             }
         }
 
+        /** Validate and potentially replace a client before use. */
+        protected C reconnect(C client) {
```
JingsongLi: Maybe it is better to change a method name?

junmuz: Sure, I can rename it to `ensureActiveClient`.

junmuz: @JingsongLi The PR is ready for review.

JingsongLi (approved):
+1

PR #7475, opened 2026-03-18
[core] Support custom table paths in JdbcCatalog

tub:
## Summary
- Enable users to specify custom storage locations for individual tables in JdbcCatalog via `CoreOptions.PATH`, replicating the pattern already used by HiveCatalog
- The catalog-level `warehouse` becomes a default, but individual tables can live at arbitrary paths persisted in `paimon_table_properties`
- Custom-path tables behave as external tables: drop skips filesystem deletion, rename skips filesystem move

## Test plan
- [x] `testCreateTableWithCustomPath` — schema at custom location, `getTableLocation()` returns it, `path` stored in JDBC
- [x] `testCreateTableWithCustomPathSyncDisabled` — `path` stored even when `syncTableProperties=false`
- [x] `testDropTableWithCustomPath` — JDBC cleaned up, data NOT deleted
- [x] `testDropTableWithDefaultPath` — data IS deleted (existing behavior preserved)
- [x] `testRenameTableWithCustomPath` — metadata updated, path preserved, no filesystem rename
- [x] `testAlterTableWithCustomPath` — path preserved after alter
- [x] `testLoadTableSchemaWithCustomPath` — schema loads from custom location
- [x] `testGetTableLocationFallback` — `getTableLocation()` returns default when no stored path
- [x] All 59 JdbcCatalogTest tests pass (including inherited CatalogTestBase tests)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/jdbc/JdbcCatalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/jdbc/JdbcCatalog.java
@@ -294,8 +294,9 @@ public class JdbcCatalog extends AbstractCatalog {
     protected void dropTableImpl(Identifier identifier, List<Path> externalPaths) {
         try {
             // Retrieve table location and custom-path flag BEFORE deleting JDBC metadata
-            Path path = getTableLocation(identifier);
-            boolean customPath = syncTableProperties() && isCustomTablePath(identifier);
+            Optional<String> storedPath = fetchStoredPathIfSyncEnabled(identifier);
+            Path path = storedPath.map(Path::new).orElse(super.getTableLocation(identifier));
+            boolean customPath = storedPath.isPresent();
 
             int deletedRecords =
                     execute(
@@ -402,8 +403,9 @@ public class JdbcCatalog extends AbstractCatalog {
     protected void renameTableImpl(Identifier fromTable, Identifier toTable) {
         try {
             // Check custom path BEFORE renaming metadata
-            boolean customPath = syncTableProperties() && isCustomTablePath(fromTable);
-            Path fromPath = getTableLocation(fromTable);
+            Optional<String> storedPath = fetchStoredPathIfSyncEnabled(fromTable);
+            boolean customPath = storedPath.isPresent();
+            Path fromPath = storedPath.map(Path::new).orElse(super.getTableLocation(fromTable));
 
             // update table metadata info
             updateTable(connections, catalogKey, fromTable, toTable);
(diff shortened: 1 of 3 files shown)


JingsongLi (commented):
Well-documented PR with comprehensive test plan. Supporting custom table paths in JdbcCatalog aligns with HiveCatalog behavior.

**Review:**

1. **External table semantics**: Custom-path tables treated as external (drop skips filesystem deletion, rename skips filesystem move) — correct and consistent with Hive conventions.

2. **Path persistence**: Stored in `paimon_table_properties` — the existing JDBC metadata table. No schema migration needed.

3. **Fallback**: `getTableLocation()` returns default warehouse path when no custom path stored. Backward compatible.

4. **Test coverage**: 9 specific tests plus all 59 inherited `CatalogTestBase` tests passing — excellent.

5. **388 additions** across JdbcCatalog, JdbcUtils, and tests. Reasonable for the scope.

6. **Edge cases to verify**:
   - Custom path points to a non-existent location → create table should fail or create directory?
   - Custom path is on a different filesystem than warehouse → does FileIO resolve correctly?
   - Two tables pointing to the same custom path → should be rejected?

LGTM. Clean implementation following established patterns.

nickdelnano:
@tub can you mark this PR as ready? Looks like @JingsongLi has commented and approved although will need an actual approval. I would like to build https://github.com/apache/paimon/pull/8111 on top of it

tub:
@nickdelnano done.

paimon-core/src/main/java/org/apache/paimon/jdbc/JdbcCatalog.java
```diff
@@ -334,12 +344,17 @@ protected void dropTableImpl(Identifier identifier, List<Path> externalPaths) {
...
         try {
-            // create table file
-            SchemaManager schemaManager = getSchemaManager(identifier);
+            // Determine table location before table exists in JDBC
+            Path tableLocation = initialTableLocation(schema.options(), identifier);
+            boolean externalTable =
+                    syncTableProperties() && schema.options().containsKey(CoreOptions.PATH.key());
+
+            // create table file using the determined location
+            SchemaManager schemaManager = new SchemaManager(fileIO, tableLocation);
```
JingsongLi, line 353: Custom paths should use a FileIO resolved from the custom path, not the catalog warehouse. With a warehouse on one filesystem and `path` on another (for example local/HDFS warehouse plus `s3://...` table path), this `SchemaManager` writes/validates schema through the warehouse FileIO. The same problem shows up when the table is loaded or altered because `JdbcCatalog` still uses the inherited non-external metadata path and does not override `fileIO(Path)` like `HiveCatalog` does. That breaks the "multiple warehouses/buckets" use case. Please resolve FileIO from `tableLocation` and make loaded custom-path tables use that path-based/external FileIO.

JingsongLi:

Issue #7490, opened 2026-03-20
[Bug] [core] changing 'dynamic-bucket.initial-buckets' on existing tables should be forbidden


jacklong319:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Paimon version

1.3.1

### Compute Engine

flink

### Minimal reproduce step

We have an existing primary key table with dynamic bucket / cross-partition upsert enabled, for example:
`WITH (
  'bucket' = '-1',
  'dynamic-bucket.initial-buckets' = '64',
  'dynamic-bucket.max-buckets' = '256',
  'deletion-vectors.enabled' = 'true',
  'merge-engine' = 'deduplicate'
)`
The table has already been written with data and is used by a running Flink job.

Later, we altered:
`ALTER TABLE ... SET (
  'dynamic-bucket.initial-buckets' = '128'
);`
Then after restarting or recovering the Flink writing job, checkpoint failed during cross-partition bootstrap with errors like:

`java.io.IOException: Could not perform checkpoint ... for operator cross-partition-bucket-assigner
...
Caused by: java.lang.RuntimeException: Exception in bulkLoad, the most suspicious reason is that your data contains duplicates, please check your sink table. (The likelihood of duplication is that you used multiple jobs to write the same dynamic bucket table, it only supports single write)
...
Caused by: org.rocksdb.RocksDBException: Keys must be added in strict ascending order.`

### What doesn't meet your expectations?

dynamic-bucket.initial-buckets seems to participate in assigner/bucket distribution logic for dynamic bucket tables.
After the table already contains data, altering this option changes the assigner/bucket mapping used by bootstrap / checkpoint path, and may finally trigger duplicate-key related failures during bootstrap bulk load.

From the user perspective, this option behaves like an immutable table option once the table is created and has data.

However, it is currently possible to change it on an existing table, and the failure only appears later when the Flink job runs and checkpoints, which is confusing and dangerous.

### Anything else?

I think Paimon should reject altering dynamic-bucket.initial-buckets for existing dynamic bucket tables (bucket = -1), especially for cross-partition upsert scenarios.

Suggested behavior:

Treat dynamic-bucket.initial-buckets as an immutable option once the table is created.
Reject ALTER TABLE ... SET ('dynamic-bucket.initial-buckets' = ...) with a clear validation error.
Ideally document that this option is only intended for table creation / initial topology setup, not online adjustment for existing tables.
[...]

PR #7491, opened 2026-03-20
[core] make dynamic-bucket.initial-buckets immutable

jacklong319:
### Purpose

[ISSUE](https://github.com/apache/paimon/issues/7490)

`dynamic-bucket.initial-buckets` affects assigner/bootstrap semantics for dynamic bucket tables.
For existing tables, changing this option may break Flink recovery/checkpoint for cross-partition upsert jobs, and may finally fail during bootstrap with duplicate-key related exceptions.
This PR marks `dynamic-bucket.initial-buckets` as immutable so that users get a clear validation error earlier instead of hitting runtime failures later.

## Changes
- add `@Immutable` to `CoreOptions.DYNAMIC_BUCKET_INITIAL_BUCKETS`
- add tests to reject altering this option on existing tables

### Tests

- manual test

### DIFF
--- a/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
@@ -1391,6 +1391,7 @@ public class CoreOptions implements Serializable {
                             "If the bucket is -1, for primary key table, is dynamic bucket mode, "
                                     + "this option controls the target row number for one bucket.");
 
+    @Immutable
     public static final ConfigOption<Integer> DYNAMIC_BUCKET_INITIAL_BUCKETS =
             key("dynamic-bucket.initial-buckets")
                     .intType()


jacklong319:
One more concern is the production impact of this issue.

For an already-used dynamic bucket table with a large amount of existing data, if `dynamic-bucket.initial-buckets` is changed by mistake, the Flink writing job may fail during checkpoint/bootstrap and become unable to run normally.

In practice, this is a serious online risk because:

- the table may already contain a large amount of production data
- the failure happens at runtime instead of being rejected at alter time
- the job may not be able to recover quickly
- rollback / repair / data cleanup may take a long time for large tables

So this is not only a usability issue, but also a stability and operability issue in production environments.

It would be much safer to reject changing `dynamic-bucket.initial-buckets` immediately for existing tables, instead of allowing the change and exposing users to a potentially major online incident later.

JingsongLi (commented):
Simple and correct. Adding `@Immutable` to `DYNAMIC_BUCKET_INITIAL_BUCKETS` prevents users from changing this option on existing tables, which would break Flink checkpoint recovery.

+1/-0 (just the annotation). The PR description clearly explains the failure mode (duplicate-key exceptions during bootstrap).

**One question**: Is there a migration path for users who already have tables with the wrong `initial-buckets` setting? If they can't change it now, they'd need to recreate the table. Consider adding a note in the error message: "To change this option, create a new table and migrate data."

LGTM.

JingsongLi:
+1

PR #7492, opened 2026-03-20
[python] Introduce changelog_manager for pypaimon

xuzifu666:
### Purpose

Support changelog manager API in pypaimon.


JingsongLi (commented):
Hi @xuzifu666 , why we need to add a so complex doc for changelog management? Is is so useful?

PR #7494, opened 2026-03-20
[core] Add catalog name to Catalog

Zouxxyy:
### Purpose

Add catalog name to Spark table display (`catalog.db.table` format) for easier identification in multi-catalog environments. Catalog name is propagated via `CatalogContext` → `CatalogEnvironment`, and Spark side uses `BaseTable.tableNameWithCatalog()` to build the display name. `Table.fullName()` is not changed.

### Tests

`DescribeTableTestBase`, `DDLWithHiveCatalogTestBase`, `V2WriteRequireDistributionTest`

### API and Format

- `Catalog` interface: added `@Nullable String name()`

### Documentation

No.

### Generative AI tooling

Generated-by: Kiro (Claude Opus 4.6)

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
@@ -1166,6 +1166,10 @@ public interface Catalog extends AutoCloseable {
 
     // ==================== Catalog Information ==========================
 
+    /** The name of this catalog. */
+    @Nullable
+    String name();
+
     /** Catalog options for re-creating this catalog. */
     Map<String, String> options();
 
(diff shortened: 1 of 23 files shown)


paimon-core/src/main/java/org/apache/paimon/table/Table.java
```diff
@@ -54,7 +54,10 @@ public interface Table extends Serializable {
     /** A name to identify this table. */
     String name();
 
-    /** Full name of the table, default is database.tableName. */
+    /**
+     * Full name of the table, default is database.tableName or catalog.database.tableName if
```
JingsongLi: I think this is a bit of a big change. It's not compatible and some systems already rely on it. Can we just modify the Spark side and pass the catalogName?

paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
```diff
@@ -1166,6 +1166,10 @@ TableQueryAuthResult authTableQuery(Identifier identifier, @Nullable List<String
 
     // ==================== Catalog Information ==========================
 
+    /** The name of this catalog. */
+    @Nullable
+    String name();
```
JingsongLi, line 1171: Can we just store this name to `SparkTable`?

PR #7497, opened 2026-03-21
[core] Add bulk load for bootstrapping in clustering compact

JingsongLi:
### Purpose

Adds a bulkLoad() method to `SimpleLsmKvDb` that writes globally sorted entries directly into SST files at the deepest level (L3), bypassing MemTable/flush/compaction. This is used by `ClusteringKeyIndex.bootstrap()` to efficiently populate the key index during restore.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/clustering/ClusteringCompactManager.java
+++ b/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/clustering/ClusteringCompactManager.java
@@ -27,63 +27,29 @@ import org.apache.paimon.compact.CompactDeletionFile;
 import org.apache.paimon.compact.CompactFutureManager;
 import org.apache.paimon.compact.CompactResult;
 import org.apache.paimon.compact.CompactTask;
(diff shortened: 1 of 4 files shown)


paimon-common/src/main/java/org/apache/paimon/lookup/sort/db/SimpleLsmKvDb.java
```diff
@@ -225,6 +228,89 @@ public void delete(byte[] key) throws IOException {
...
+        } catch (IOException | RuntimeException e) {
+            if (currentWriter != null) {
+                try {
+                    currentWriter.close();
+                } catch (IOException suppressed) {
+                    e.addSuppressed(suppressed);
+                }
+            }
+            throw e;
+        }
```
LsomeYeah, line 306: If bootstrap fails repeatedly (e.g. during failover), could the orphaned SST files accumulate and eventually blow up the disk?

JingsongLi: This Job will fail and k8s will clean the disk.

PR #7499, opened 2026-03-21
[core] Introduce 'consumer.changelog-only' to keep less snapshots

JingsongLi:
### Purpose

Introduce `consumer.changelog-only`: If true, consumer will only affect changelog expiration and will not prevent snapshot from being expired.

### Tests

`ExpireSnapshotsTest.testConsumerChangelogOnly`

### DIFF
--- a/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
@@ -3378,7 +3378,7 @@ public class CoreOptions implements Serializable {
     }
 
     public boolean consumerChangelogOnly() {
-        return options.get(CONSUMER_CHANGELOG_ONLY);
+        return options.get(CONSUMER_CHANGELOG_ONLY) && changelogLifecycleDecoupled();
     }
 
     public boolean partitionedTableInMetastore() {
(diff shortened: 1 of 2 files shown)


paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -3368,6 +3377,10 @@ public boolean consumerIgnoreProgress() {
         return options.get(CONSUMER_IGNORE_PROGRESS);
     }
 
+    public boolean consumerChangelogOnly() {
+        return options.get(CONSUMER_CHANGELOG_ONLY);
```
tsreaper: Also check that changelog decouple is set?

paimon-core/src/test/java/org/apache/paimon/operation/ExpireSnapshotsTest.java
```diff
@@ -661,6 +663,48 @@ public void testManifestFileSkippingSetFileNotFoundException() throws Exception
         assertSnapshot(latestSnapshotId, allData, snapshotPositions);
     }
 
+    @Test
+    public void testConsumerChangelogOnly() throws Exception {
```
tsreaper, line 667: Test that with this option, changelog expiration can still be protected by consumer.

PR #7516, opened 2026-03-24
[python] Fix signing algorithm selection by host name

tonymtu:
### Purpose
Python signature algorithm selection align with Java behavior.


JingsongLi:
+1

PR #7527, opened 2026-03-25
[python] Add comment method for file_store_table

xuzifu666:
### Purpose
Currently Python API can not directly retrieve the `comment` from table; a `comment` method provided to make it easier for users.


JingsongLi:
+1

PR #7537, opened 2026-03-26
[core][flink] Add option to skip expired partitions in compaction job

### LINKED ISSUE #7531: [Feature] add option to skip expired partitions in compaction job
### Search before asking

### Motivation

Compact job may attempt to compact files in partitions that are concurrently being expired. When committing, the compactBefore files  are deleted by partition expiration, causing a conflict and the error: "You are writing data to expired partitions".

ERROR message like this:
java.lang.RuntimeException: You are writing data to expired partitions, and you can filter this data to avoid job failover. Otherwise, continuous expired records will cause the job to failover restart continuously. Expired partitions are: \[20260311, 20260318\] at org.apache.paimon.operation.FileStoreCommitImpl.assertConflictForPartitionExpire(FileStoreCommitImpl.java:1445) at org.apache.paimon.operation.FileStoreCommitImpl.assertNoDelete(FileStoreCommitImpl.java:1424) at org.apache.paimon.operation.FileStoreCommitImpl.noConflictsOrFail(FileStoreCommitImpl.java:1368) at org.apache.paimon.operation.FileStoreCommitImpl.tryCommitOnce(FileStoreCommitImpl.java:948) at org.apache.paimon.operation.FileStoreCommitImpl.tryCommit(FileStoreCommitImpl.java:782) at org.apache.paimon.operation.FileStoreCommitImpl.commit(FileStoreCommitImpl.java:361) at org.apache.paimon.table.sink.TableCommitImpl.commitMultiple(TableCommitImpl.java:217) at

### Solution

Add a new table option `compaction.skip-expired-partitions` (default false). When enabled, the compaction job source filters out partitions  that have already expired, preventing compaction from generating DELETE  manifest entries for expired partitions and avoiding the commit  conflict.

### Anything else?

No

lilei1128:
### Purpose
  Add a new option `compaction.skip-expired-partitions` (default: false)
  that allows the Flink compaction job to skip partitions already expired
  under the `values-time` partition expiration strategy, avoiding
  unnecessary compaction work on soon-to-be-deleted dataPlease continue.

- Add CoreOptions.COMPACTION_SKIP_EXPIRED_PARTITIONS config option
- Filter expired partitions in CompactorSourceBuilder.build() when the option enabled.

Close #7531 
### Tests
- Add testSkipExpiredPartitions, testNotSkipExpiredPartitionsByDefault

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/CompactorSourceBuilder.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/CompactorSourceBuilder.java
@@ -25,6 +25,7 @@ import org.apache.paimon.flink.LogicalTypeConversion;
 import org.apache.paimon.manifest.PartitionEntry;
 import org.apache.paimon.options.Options;
 import org.apache.paimon.partition.PartitionPredicate;
+import org.apache.paimon.partition.PartitionValuesTimeExpireStrategy;
 import org.apache.paimon.table.FileStoreTable;
 import org.apache.paimon.table.source.ReadBuilder;
 import org.apache.paimon.table.system.CompactBucketsTable;
@@ -138,6 +139,26 @@ public class CompactorSourceBuilder {
                             rowData -> {
                                 BinaryRow partition = deserializeBinaryRow(rowData.getBinary(1));
                                 return partitionInfo.get(partition) <= historyMilli;
+                            });
+            dataStream = new DataStreamSource<>(filterStream);
+        }
+        CoreOptions coreOptions = table.coreOptions();
+        if (coreOptions.compactionSkipExpiredPartitions()
+                && coreOptions.partitionExpireTime() != null
+                && CoreOptions.PartitionExpireStrategy.VALUES_TIME
+                        .toString()
+                        .equals(coreOptions.partitionExpireStrategy())) {
+            RowType partitionType = table.schema().logicalPartitionType();
+            Duration expireTime = coreOptions.partitionExpireTime();
+            PartitionValuesTimeExpireStrategy expireStrategy =
+                    new PartitionValuesTimeExpireStrategy(coreOptions, partitionType);
+            SingleOutputStreamOperator<RowData> filterStream =
+                    dataStream.filter(
+                            rowData -> {
+                                LocalDateTime expireDateTime =
+                                        LocalDateTime.now().minus(expireTime);
+                                BinaryRow partition = deserializeBinaryRow(rowData.getBinary(1));
+                                return !expireStrategy.isExpired(expireDateTime, partition);
                             });
             dataStream = new DataStreamSource<>(filterStream);
         }
--- a/paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/action/CompactActionITCase.java
+++ b/paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/action/CompactActionITCase.java
@@ -947,6 +947,8 @@ public class CompactActionITCase extends CompactActionITCaseBase {
                 CoreOptions.PARTITION_EXPIRATION_STRATEGY.key(),
                 CoreOptions.PartitionExpireStrategy.VALUES_TIME.toString());
         tableOptions.put(CoreOptions.PARTITION_TIMESTAMP_FORMATTER.key(), "yyyyMMdd");
+        // Prevent partition expiration from running during the compact commit.
+        tableOptions.put(CoreOptions.PARTITION_EXPIRATION_CHECK_INTERVAL.key(), "999 d");
 
         FileStoreTable table =
                 prepareTable(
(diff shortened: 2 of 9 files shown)


lilei1128:
> I have a quick solution: see `AbstractFileStore#newPartitionExpire`, you can use sql hint or other way to pass dynamic option like
> 
> ```
> dynamicOptions.put(CoreOptions.PARTITION_EXPIRATION_TIME.key(), null);
> newTable = table.copy(dynamicOptions)
> -- use newTable to build job
> ```

Thank you for your reply, sql hints can indeed dynamically pass in parameters, but for compact jobs(like the issue), it seems that parameters cannot be passed in through hints.

JingsongLi (commented):
Useful optimization. Skipping compaction for already-expired partitions avoids wasted work.

**Review:**

1. **Config**: `compaction.skip-expired-partitions` (default: false) — conservative default is correct. Users opt in explicitly.

2. **Implementation**: Filtering in `CompactorSourceBuilder.build()` using the existing `PartitionExpireStrategy` — reuses the partition expiration evaluation logic. Clean.

3. **Scope**: Only applies when `partition.expiration-strategy=values-time`. Does it also work with `update-time` strategy? If not, document the limitation.

4. **Edge case**: A partition that is "about to expire" might still receive late-arriving data that resets its expiration time. If we skip compaction, that late data won't be compacted until the partition is no longer expired. Is this acceptable?

5. **+199/-8** with tests. Tests cover both enabled/disabled cases. Good.

6. **Interaction with partition expiration**: If the compaction job skips expired partitions, but the expiration job hasn't run yet, the data files remain uncompacted until they're deleted. This is fine — no point compacting data that will be deleted soon.

LGTM.

JingsongLi:
Please rebase master.

lilei1128:
> Please rebase master.

Thanks a lot for taking your time to review. I have rebased this PR onto the latest master branch.
Currently only values-time strategy is supported and update-time will be added in a follow-up PR(Including documentation instructions) to keep this change focused.

lilei1128:
Hi @JingsongLi,  I have added documentation for this feature. Could you please take a look when you have time?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/CompactorSourceBuilder.java
```diff
@@ -141,6 +142,26 @@ public DataStreamSource<RowData> build() {
                             });
             dataStream = new DataStreamSource<>(filterStream);
         }
+        CoreOptions coreOptions = table.coreOptions();
```
JingsongLi, line 145: This filter only removes expired partitions from the compaction source, but the compactor sink still uses the table with `write-only=false`, so `table.newCommit(...)` will create and run `PartitionExpire` during the compact commit. As a result, a job that is supposed to skip expired partitions can still delete those partitions as a side effect of compacting the remaining active partitions. Could we either disable partition expiration for this compaction commit path, or document that enabling this option may also expire the skipped partitions during the commit?

lilei1128: Thanks for the review!
You're right that the compactor sink still runs PartitionExpire during commit. However, we intentionally kept this behavior rather than disabling partition expiration in the commit path.

The reason: in the typical deployment pattern where write jobs use write-only=true, the compaction job's commit is the only path that  triggers PartitionExpire. If we disable it here too, expired partitions would never be cleaned up automatically — users would have to  manually run ExpirePartitionsAction to delete them. That feels worse than the current side effect.

So the goal of this option is specifically to **avoid unnecessary compaction IO** on partitions that are about to be deleted (no file merging, no writing compacted files that get immediately dropped).

I have updated the documentation (both CoreOptions.java and dedicated-compaction.mdx) to explicitly note that expired partitions may still be deleted during the commit phase as a side effect, so users are aware of this behavior.

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/action/CompactActionITCase.java
```diff
@@ -930,6 +932,127 @@ public void testDataEvolutionTableCompact() throws Exception {
...
+                        database,
+                        "--table",
+                        tableName,
+                        "--table_conf",
+                        CoreOptions.COMPACTION_SKIP_EXPIRED_PARTITIONS.key() + "=true");
+        StreamExecutionEnvironment env = streamExecutionEnvironmentBuilder().batchMode().build();
+        action.withStreamExecutionEnvironment(env).build();
+        env.execute();
+
+        checkLatestSnapshot(table, 3, Snapshot.CommitKind.COMPACT);
```
JingsongLi, line 987: This assertion assumes partition expiration does not run during the compact commit, but that is not guaranteed here. `CompactAction` copies the table with `write-only=false`, so `TableCommitImpl.maintain` may call `PartitionExpire.expire(...)` after the compaction snapshot. `NormalPartitionExpire` also initializes `lastCheck` with a random offset within `partition.expiration-check-interval`, so even with the default 1h interval this test can occasionally expire `expiredDt`, creating an extra OVERWRITE snapshot before this check (or removing the expired split before the later file-count assertion). Could we make the test deterministic, for example by setting a very large `partition.expiration-check-interval`, or by explicitly asserting the documented delete-on-commit behavior?

lilei1128: Thanks for the review!  I've added partition.expiration-check-interval=999 d to  both test cases to make the tests deterministic and prevent partition expiration from being triggered during the compact commit.

JingsongLi:
+1

PR #7542, opened 2026-03-26
[spark] Fix nested struct field mapping for V2 write merge schema

Zouxxyy:
### Purpose

Fix nested struct field mapping in SparkInternalRowWrapper for V2 write merge schema.

### Tests

Added test cases for nested struct field evolution.


JingsongLi:
+1

PR #7543, opened 2026-03-27
[vortex] Introduce vortex file format integration

JingsongLi:
### Purpose

See https://vortex.dev/

> 100x faster random access
> 10-20x faster scans5x faster writes
> Similar compression ratio
> (vs. Apache Parquet)

Vortex performs well in terms of vectors, and introducing it can significantly help us gain significant advantages in vector storage. (https://paimon.apache.org/docs/master/append-table/vector/)

### Tests

`VortexFileFormatReadWriteTest` and more.

### DIFF
--- a/paimon-vortex/paimon-vortex-format/src/main/java/org/apache/paimon/format/vortex/VortexPredicateConverter.java
+++ b/paimon-vortex/paimon-vortex-format/src/main/java/org/apache/paimon/format/vortex/VortexPredicateConverter.java
@@ -38,6 +38,8 @@ import org.apache.paimon.predicate.Predicate;
 import org.apache.paimon.predicate.PredicateBuilder;
 import org.apache.paimon.predicate.PredicateVisitor;
 import org.apache.paimon.types.DataType;
+import org.apache.paimon.types.LocalZonedTimestampType;
+import org.apache.paimon.types.TimestampType;
 
 import dev.vortex.api.Expression;
 import dev.vortex.api.expressions.Binary;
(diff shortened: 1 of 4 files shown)


paimon-vortex/paimon-vortex-format/src/main/java/org/apache/paimon/format/vortex/VortexRecordsReader.java
```diff
@@ -0,0 +1,148 @@
...
+                return filePath;
+            }
+
+            @Nullable
+            @Override
+            public InternalRow next() {
+                if (!rows.hasNext()) {
+                    return null;
+                }
+                currentPosition++;
```
leaves12138: We should not simple increase currentPosition if rowIndices is not none. Please refer to ColumnarRowIterator.returnedPosition(), use LongIterator to memorize the returned position

paimon-vortex/paimon-vortex-format/src/main/java/org/apache/paimon/format/vortex/VortexPredicateConverter.java
```diff
@@ -0,0 +1,192 @@
...
+            case VARCHAR:
+                return Literal.string(
+                        value instanceof BinaryString ? value.toString() : (String) value);
+            case DECIMAL:
+                Decimal decimal = (Decimal) value;
+                return Literal.decimal(
+                        decimal.toBigDecimal(), decimal.precision(), decimal.scale());
+            case TIMESTAMP_WITHOUT_TIME_ZONE:
+                Timestamp ts = (Timestamp) value;
+                return Literal.timestampMicros(
```
leaves12138: If timestamp precision is not 3, we should make the converter match the precision

PR #7545, opened 2026-03-27
[core] Skip schema update when no changes detected

zhongyujiang:
## Purpose

This PR fixes an issue where Paimon creates a new schema version even when there are no actual changes to the schema content.

Previously, when executing schema update operations like `setOption("foo", "bar")` where the option already had the same value, Paimon would still create a new schema version with an incremented ID. This resulted in unnecessary schema versions being created.

## Changes

1. Added `sameContent()` method to `TableSchema` to compare schema content while ignoring version and timeMillis
2. Modified `commitChanges()` to check if new schema has same content as old schema before committing
3. Fixed `equals()` and `hashCode()` in `TableSchema` to include `highestFieldId` for proper comparison

## Tests

Added `testNoChangeCommitDoesNotCreateNewSchema()` in `SchemaManagerTest`:
- Creates a table with an option `foo=bar`
- Calls `commitChanges(SchemaChange.setOption("foo", "bar"))` with the same value
- Verifies that no new schema version is created (ID remains the same)
- Also tests `updateComment` with unchanged comment value

### DIFF
--- a/paimon-api/src/main/java/org/apache/paimon/schema/TableSchema.java
+++ b/paimon-api/src/main/java/org/apache/paimon/schema/TableSchema.java
@@ -342,6 +342,7 @@ public class TableSchema implements Serializable {
         }
         TableSchema tableSchema = (TableSchema) o;
         return version == tableSchema.version
+                && highestFieldId == tableSchema.highestFieldId
                 && Objects.equals(fields, tableSchema.fields)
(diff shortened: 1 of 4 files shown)


JingsongLi (commented):
I think we should modify `generateTableSchema` remove useless changes and return a `Optional<TableSchema>`.

PR #7546, opened 2026-03-27
[doc] Fix master docs version to 1.5-SNAPSHOT

leaves12138:
## Why
The master branch has already moved to 1.5-SNAPSHOT in pom.xml, but docs/config.toml still uses 1.4-SNAPSHOT for `Version` and `VersionTitle`. This makes the master docs site render outdated artifact and version references.

## What is changed
- Update `docs/config.toml` `Version` to `1.5-SNAPSHOT`
- Update `docs/config.toml` `VersionTitle` to `1.5-SNAPSHOT`

`StableDocs` remains unchanged because this PR only fixes the master docs snapshot version.


JingsongLi:
+1

PR #7549, opened 2026-03-30
[core] Support rename branch API and procedures

xuzifu666:
### Purpose
Currently user can't change branch name directly, need to create a new branch and then delete the old one, which is inconvenient. Therefore provide an API for rename branch. This PR provided:

1. Experimental rename branch API in Table
2. RenameBranchProcedure in Flink/Spark
3. RestCatalog api for rename branch

### DIFF
--- a/docs/static/rest-catalog-open-api.yaml
+++ b/docs/static/rest-catalog-open-api.yaml
@@ -1214,8 +1214,7 @@ paths:
         content:
           application/json:
             schema:
-              type: string
-            example: "new_branch_name"
+              $ref: '#/components/schemas/RenameBranchRequest'
       responses:
         "200":
           description: Success, no content
@@ -3294,6 +3293,11 @@ components:
           type: string
         fromTag:
           nullable: true
+          type: string
+    RenameBranchRequest:
+      type: object
+      properties:
+        toBranch:
           type: string
     ForwardBranchRequest:
       type: object


paimon-api/src/main/java/org/apache/paimon/rest/requests/RenameBranchRequest.java
```diff
@@ -0,0 +1,58 @@
...
+
+/** Request for renaming branch. */
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class RenameBranchRequest implements RESTRequest {
+
+    private static final String FIELD_FROM_BRANCH = "fromBranch";
+    private static final String FIELD_TO_BRANCH = "toBranch";
+
+    @JsonProperty(FIELD_FROM_BRANCH)
+    private final String fromBranch;
```
JingsongLi: from is already in path.

paimon-api/src/main/java/org/apache/paimon/rest/ResourcePaths.java
```diff
@@ -262,6 +262,19 @@ public String forwardBranch(String databaseName, String tableName, String branch
                 "forward");
     }
 
+    public String renameBranch(String databaseName, String tableName, String branch) {
```
JingsongLi, line 265: Please also modify rest-catalog-open-api.yaml

paimon-api/src/main/java/org/apache/paimon/rest/requests/RenameBranchRequest.java
xuzifu666: Had removed fromBranch.

paimon-api/src/main/java/org/apache/paimon/rest/ResourcePaths.java
xuzifu666: Done ~

docs/static/rest-catalog-open-api.yaml
```diff
@@ -1183,6 +1183,50 @@ paths:
...
+          in: path
+          required: true
+          schema:
+            type: string
+      requestBody:
+        content:
+          application/json:
+            schema:
+              type: string
+            example: "new_branch_name"
```
JingsongLi: Why add example?

docs/static/rest-catalog-open-api.yaml
```diff
@@ -1183,6 +1183,50 @@ paths:
...
+        - name: branch
+          in: path
+          required: true
+          schema:
+            type: string
+      requestBody:
+        content:
+          application/json:
+            schema:
+              type: string
```
JingsongLi: This should be a struct, and has a field toBranch

docs/static/rest-catalog-open-api.yaml
xuzifu666: It's definitely not needed,  I've deleted it.

docs/static/rest-catalog-open-api.yaml
xuzifu666: Yes, a struct description has been added; it has been modified.

PR #7551, opened 2026-03-30
[tantivy] Introduce paimon-tantivy for full text search index

JingsongLi:
### Purpose

Introduced a complete Tantivy full text search module, including Rust JNI layer, Java integration layer, archive format CI workflow.

### Tests

`TantivyFullTextGlobalIndexTest`.

### DIFF
--- a/paimon-tantivy/paimon-tantivy-index/src/main/java/org/apache/paimon/tantivy/index/TantivyFullTextGlobalIndexWriter.java
+++ b/paimon-tantivy/paimon-tantivy-index/src/main/java/org/apache/paimon/tantivy/index/TantivyFullTextGlobalIndexWriter.java
@@ -110,6 +110,10 @@ public class TantivyFullTextGlobalIndexWriter implements GlobalIndexSingletonWri
         } catch (IOException e) {
             throw new RuntimeException("Failed to write Tantivy full-text global index", e);
         } finally {
+            if (writer != null) {
+                writer.close();
+                writer = null;
+            }
             closed = true;
             deleteTempDir();
         }


jerry-024 (commented):
createFileReader already receives an absolute meta.filePath(), so wrapping it in new Path(path, meta.filePath()) is misleading (and only “works” because Path ignores the parent when the child is absolute). Prefer the clearer version:
```java
  private GlobalIndexFileReader createFileReader() {
      return meta -> fileIO.newInputStream(meta.filePath());
  }
```

paimon-tantivy/paimon-tantivy-index/src/main/java/org/apache/paimon/tantivy/index/TantivyFullTextGlobalIndexWriter.java
```diff
@@ -0,0 +1,213 @@
...
+            writer.close();
+            writer = null;
+
+            return Collections.singletonList(packIndex());
+        } catch (IOException e) {
+            throw new RuntimeException("Failed to write Tantivy full-text global index", e);
+        } finally {
+            closed = true;
+            deleteTempDir();
+        }
```
jerry-024, line 119: Native resource leak in `finish()`:

1. **`rowId == 0` path**: returns early without calling `writer.close()`, then `finally` sets `closed = true` and deletes the temp dir. The subsequent `close()` method skips cleanup because `closed` is already `true`, so the native `TantivyIndexWriter` handle is never freed.

2. **Exception path**: if `writer.commit()` or `packIndex()` throws, same issue — `writer` is not closed but `closed` is set to `true`.

Suggested fix — close the writer in `finally` before setting `closed`:

```java
} finally {
    if (writer != null) {
        writer.close();
        writer = null;
    }
    closed = true;
    deleteTempDir();
}
```

PR #7563, opened 2026-03-31
[core] Support btree global index with embedded file metadata

lilei1128:
### Purpose
  Add with-file-meta option for btree global index to embed ManifestEntry
  data directly in index files, enabling manifest-skip query planning.

  Key changes:
  - New index type "btree_file_meta": SST file mapping fileName -> ManifestEntry bytes
  - BTreeWithFileMetaBuilder: builds btree key-index + btree_file_meta atomically
  - BTreeWithFileMetaReader: reads both indexes and returns FilePathGlobalIndexResult
  - DataEvolutionBatchScan: fast path using file-meta to build DataSplits without
    manifest reads, with staleness detection via fileIO.exists()
  - BTreeIndexOptions.BTREE_WITH_FILE_META: config to enable the feature
  
  When enabled, query planning reads only:
  - btree key-index SST (for matching row IDs)
  - btree_file_meta SST (for ManifestEntry data)
  
  See: https://cwiki.apache.org/confluence/display/PAIMON/PIP-41%3A+Introduce+FilePath+Global+Index+And+Optimizations+For+Lookup+In+Append+Table

### Tests
CI


steFaiz (commented):
Hi, thanks for this PR! 
I'm just wondering that if it's necessary to reuse current BTree codebase? For example:
1.  in the btree index build topo, current implementation will decide  the partition num by records number per range, and split ranges by partition, which may be not suitable for your case. 
2. And also, it seems that the `BTREE_WITH_FILE_META` option will create a total different index type compared to BTree.

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/globalindex/btree/BTreeIndexTopoBuilder.java
```diff
@@ -126,6 +140,31 @@ public List<CommitMessage> buildIndex(
                 int partitionNum = Math.max((int) (range.count() / recordsPerRange), 1);
                 partitionNum = Math.min(partitionNum, maxParallelism);
 
+                // Pre-serialize ManifestEntries for file-meta index (if withFileMeta is enabled)
```
steFaiz, line 143: It seems that the key(i.e. the filename) is just used for deduplication? Can I image this index as actually a Range to Collection\<ManifestEntry\> index?

lilei1128: Yes, currently fileName is mainly for deduplication; the runtime path does not do fileName point lookup yet.
 
 This is an optimization point that follows

JingsongLi (commented):
Significant feature for PIP-41. Embedding ManifestEntry data in the btree global index enables manifest-skip planning — a major performance win for lookup queries.

**Review:**

1. **1685 additions** is large but justified for a PIP-level feature. The approach (btree key-index + btree_file_meta SST) is architecturally sound.

2. **Staleness detection via `fileIO.exists()`**: This is the main concern. Checking file existence for every returned ManifestEntry is expensive on object storage (one HEAD request per file). Consider:
   - Batch existence checks
   - Using snapshot ID comparison instead (if embedded entry's snapshot < current, it's stale)
   - LRU cache for known-valid file paths

3. **Atomicity**: Building btree key-index + btree_file_meta atomically is critical. If one is written without the other, readers could get partial results. How is this ensured?

4. **Backward compatibility**: Tables with existing btree indexes (without file meta) must continue to work. Is there a schema/version check?

5. **Memory**: Embedding full ManifestEntry bytes in index files increases index size. What's the expected overhead? (e.g., for a table with 1M files, how large does the btree_file_meta SST become?)

Please link to PIP-41 in the PR for full design context.

PR #7566, opened 2026-03-31
[flink] Support dropping primary keys for empty tables in Flink

tsreaper:
### Purpose

Paimon has supported adding primary keys for empty tables before. So in this PR, we support the symmetrical operation: dropping primary keys for empty tables.

### Tests

* `SchemaManagerTest`
* `SchemaChangeITCase`


paimon-api/src/main/java/org/apache/paimon/schema/SchemaChange.java
```diff
@@ -164,6 +167,10 @@ static SchemaChange updateColumnPosition(Move move) {
         return new UpdateColumnPosition(move);
     }
 
+    static SchemaChange dropPrimaryKeys() {
```
JingsongLi: dropPrimaryKey, in SQL, always is primary key not primary keys.

PR #7571, opened 2026-04-01
[python] Fix token cache isolation by adding user identity to cache key

shyjsarah:
### Purpose

Revert #7562 and fix token cache pollution with a minimal approach.

**Problem:**
In #7562, the class-level `_TOKEN_CACHE` was completely removed to fix token pollution across different catalog instances. However, this also removed the ability to reuse tokens for the same user on the same table, causing unnecessary token refresh requests.

The root cause of the original pollution was that `_TOKEN_CACHE` was keyed only by table identifier (`str(self.identifier)`), so different users (different AK/SK credentials) operating on the same table within one process would share the same cached data token.

**Fix:**
Instead of removing the cache entirely, this PR adds **user identity** to the cache key to isolate tokens across different credentials while preserving token reuse for the same user.

- Revert commit `8a310149` to restore the class-level `_TOKEN_CACHE` mechanism
- Add `_build_cache_key()` and `_get_user_identity()` methods
- Cache key format changes from `"{identifier}"` to `"{user_identity}:{identifier}"`
- User identity is obtained from the existing `api_instance`'s auth provider:
  - **DLF auth** (AK/SK, STS, ECS Role): `auth_provider.get_token().access_key_id`
  - **Bear Token auth**: `auth_provider.token`
  - **Unknown auth**: `'anonymous'`

No new instance fields are added, so `__getstate__`/`__setstate__` serialization remains unchanged.

### Tests

Added 4 new test cases in `rest_token_file_io_test.py`:
- `test_build_cache_key_with_dlf_auth`: verifies DLF cache key includes `access_key_id`
- `test_build_cache_key_with_bear_auth`: verifies Bear cache key includes bearer token
- `test_different_ak_same_table_token_isolation`: verifies different AKs on the same table produce isolated cache entries
- `test_same_ak_same_table_token_reuse`: verifies same AK on the same table can reuse cached tokens

All 13 tests pass.

### DIFF
--- a/paimon-python/pypaimon/catalog/rest/rest_token_file_io.py
+++ b/paimon-python/pypaimon/catalog/rest/rest_token_file_io.py
@@ -29,7 +29,7 @@ from pypaimon.common.file_io import FileIO
 from pypaimon.filesystem.pyarrow_file_io import PyArrowFileIO
 from pypaimon.api.auth.bearer import BearTokenAuthProvider
 from pypaimon.api.auth.dlf_provider import DLFAuthProvider
-from pypaimon.common.identifier import Identifier
+from pypaimon.common.identifier import Identifier, SYSTEM_TABLE_SPLITTER
 from pypaimon.common.options import Options
 from pypaimon.common.options.config import CatalogOptions, OssOptions
 from pypaimon.common.uri_reader import UriReaderFactory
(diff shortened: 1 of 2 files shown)


shyjsarah:
> 1. Why removing SYSTEM_TABLE_SPLITTER in refresh_token?
> 2. should_refresh is dead code.
> 3. self.lock is uesless.

updated

PR #7576, opened 2026-04-01
[flink] fix namespace for lineage from table path to warehouse path

jsingh-yelp:
### Purpose
Fix the namespace for Paimon table lineage datasets. Previously the namespace was set to the full table path which cannot be used to group datasets meaningfully, changed it to use the warehouse path to better follow namespace conventions. This warehouse path is extracted from `CatalogContext` attached to `Table`.

### Tests
- Did manual testing and fixed existing tests.

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lineage/LineageUtils.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lineage/LineageUtils.java
@@ -19,8 +19,10 @@
 package org.apache.paimon.flink.lineage;
 
 import org.apache.paimon.CoreOptions;
-import org.apache.paimon.catalog.Catalog;
+import org.apache.paimon.catalog.CatalogContext;
 import org.apache.paimon.options.CatalogOptions;
+import org.apache.paimon.table.FileStoreTable;
+import org.apache.paimon.table.FormatTable;
 import org.apache.paimon.table.Table;
 
 import org.apache.flink.api.connector.source.Boundedness;
@@ -57,12 +59,25 @@ public class LineageUtils {
     private static final Set<String> PAIMON_OPTION_KEYS =
             CoreOptions.getOptions().stream().map(opt -> opt.key()).collect(Collectors.toSet());
 
+    /** Extracts the {@link CatalogContext} from a table, or null if not available. */
+    @Nullable
+    private static CatalogContext catalogContext(Table table) {
+        if (table instanceof FileStoreTable) {
+            return ((FileStoreTable) table).catalogEnvironment().catalogContext();
+        }
+        if (table instanceof FormatTable) {
+            return ((FormatTable) table).catalogContext();
+        }
+        return null;
+    }
+
     /**
-     * Builds the config map for a dataset facet from a {@link Table} and {@link Catalog}. Includes
-     * filtered Paimon {@link CoreOptions}, partition keys, primary keys, and a safe subset of
-     * catalog-level options (warehouse, metastore) prefixed with {@code "catalog."}.
+     * Builds the config map for a dataset facet. Includes filtered Paimon {@link CoreOptions},
+     * partition keys, primary keys, and a safe subset of catalog-level options (warehouse,
+     * metastore) prefixed with {@code "catalog."}.
      */
-    private static Map<String, String> buildConfigMap(Table table, @Nullable Catalog catalog) {
+    private static Map<String, String> buildConfigMap(
+            Table table, @Nullable CatalogContext catalogContext) {
         Map<String, String> config = new HashMap<>();
         config.put("type", "paimon");
         config.put("partition-keys", String.join(",", table.partitionKeys()));
@@ -72,8 +87,10 @@ public class LineageUtils {
                 .filter(e -> PAIMON_OPTION_KEYS.contains(e.getKey()))
                 .forEach(e -> config.put(e.getKey(), e.getValue()));
 
-        if (catalog != null) {
-            catalog.options()
+        if (catalogContext != null) {
+            catalogContext
+                    .options()
+                    .toMap()
                     .forEach(
                             (k, v) -> {
                                 if (CATALOG_OPTION_ALLOWLIST.contains(k)) {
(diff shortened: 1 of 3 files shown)


jsingh-yelp:
@JingsongLi / @XiaoHongbo-Hope Sorry for the late addition, but is it possible to get a review on this and include it with for the latest 1.4 release?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lineage/LineageUtils.java
```diff
@@ -60,12 +60,12 @@ private static Map<String, String> buildConfigMap(Table table) {
...
-     * Returns the lineage namespace for a Paimon table. The namespace uses the {@code paimon://}
-     * scheme followed by the table's physical warehouse path, e.g. {@code
-     * "paimon://s3://my-bucket/warehouse/mydb.db/mytable"}.
+     * Returns the lineage namespace for a Paimon table. The namespace is the warehouse path derived
+     * via {@link CatalogUtils#warehouse(String)}, e.g. {@code "s3://my-bucket/warehouse"} for
+     * object stores or {@code "file:/tmp/warehouse"} for local paths.
      */
     public static String getNamespace(Table table) {
-        return PAIMON_DATASET_PREFIX + CoreOptions.path(table.options());
+        return CatalogUtils.warehouse(CoreOptions.path(table.options()).toString());
```
JingsongLi, line 68: CatalogUtils.warehouse() is fragile — it assumes the standard `<warehouse>/<db>.db/<table>` layout, maybe we should pass meaningful message from Catalog.

jsingh-yelp: @JingsongLi Thanks this is a valid concern. 

I was thinking of using the warehouse property from Catalog.options() ([Code Ref](https://github.com/apache/paimon/blob/master/paimon-api/src/main/java/org/apache/paimon/options/CatalogOptions.java#L31)) directly as the namespace, with a fallback to a constant "paimon" string when it's not available.

Is this something in line what you had in mind or you were thinking something else?

JingsongLi: +1 to warehouse option.

XiaoHongbo-Hope:
> @JingsongLi / @XiaoHongbo-Hope Sorry for the late addition, but is it possible to get a review on this and include it with the latest 1.4 release?

Sorry for late response, My github count had some issues. And it gets back in this Friday.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lineage/LineageUtils.java
jsingh-yelp: @JingsongLi Done, couple of files were changed but it is mostly passing `Catalog` param to `PaimonDataStreamScanProvider` and `PaimonDataStreamSinkProvider` no logic was changed.

Here is an example of a lineage event with warehouse info: https://i.fluffy.cc/xDPlNsmb9vk8QxDVt5xs6PM6GdgSht8P.html#L38,L75-79

jsingh-yelp: @JingsongLi I have updated the PR to use `CatalogContext` from Table instead of threading
Catalog through constructors. This should make this PR change a lot simpler, can I please get a re-review on this PR. Thanks

jsingh-yelp: cc: @JingsongLi friendly bump can I please have a review on this.

jsingh-yelp:
cc: @JingsongLi friendly reminder, can I please have a review on this PR? Example of lineage emitted by this change: https://i.fluffy.cc/xDPlNsmb9vk8QxDVt5xs6PM6GdgSht8P.html#L38,L75-79

jsingh-yelp:
@JingsongLi Can I please have a review on this PR.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lineage/LineageUtils.java
```diff
@@ -19,66 +19,115 @@
...
-    public static String getNamespace(Table table) {
-        return PAIMON_DATASET_PREFIX + CoreOptions.path(table.options());
+    public static String getNamespace(@Nullable CatalogContext catalogContext) {
+        if (catalogContext != null) {
+            String warehouse = catalogContext.options().get(CatalogOptions.WAREHOUSE);
+            if (warehouse != null) {
+                return warehouse;
+            }
+        }
+        return DEFAULT_NAMESPACE;
```
XiaoHongbo-Hope, line 122: For direct Flink connector tables, `CatalogContext` is built from table options and may have `path` but
  no `warehouse`. This makes all such tables fall back to the same namespace `"paimon"`; can we fall back to table path instead?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lineage/LineageUtils.java
```diff
@@ -19,66 +19,115 @@
...
-     * {@link CoreOptions}, partition keys, primary keys, and the table comment (if present).
+     * Builds the config map for a dataset facet. Includes filtered Paimon {@link CoreOptions},
+     * partition keys, primary keys, and a safe subset of catalog-level options (warehouse,
+     * metastore) prefixed with {@code "catalog."}.
      */
-    private static Map<String, String> buildConfigMap(Table table) {
+    private static Map<String, String> buildConfigMap(
+            Table table, @Nullable CatalogContext catalogContext) {
         Map<String, String> config = new HashMap<>();
+        config.put("type", "paimon");
```
XiaoHongbo-Hope: `type` conflicts with `CoreOptions.TYPE.key()` and can be overwritten when table options are copied
  below. If this means connector type, can we use a non-conflicting key or write it after table options?

jsingh-yelp: @XiaoHongbo-Hope fixed as suggested

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/lineage/LineageUtils.java
jsingh-yelp: @XiaoHongbo-Hope Yes, I have fixed this as you suggested.

jsingh-yelp:
@XiaoHongbo-Hope I have added the fixes, can you please give a re-review thanks.

JingsongLi:
+1

PR #7577, opened 2026-04-02
[api] Fix locale-sensitive date formatting in DLFOpenApiSigner

plusplusjiajia:
Pin DateTimeFormatter to Locale.ENGLISH so RFC 1123 date headers are always in English regardless of JVM default locale. Add tests for Chinese and Japanese locales.

### Purpose                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                        
Fix locale-sensitive date formatting in `DLFOpenApiSigner`. The `DateTimeFormatter` for RFC 1123 dates was not pinned to `Locale.ENGLISH`, causing day/month names to be localized on non-English JVM locales (e.g. zh_CN, ja_JP), which breaks the Aliyun OpenAPI signature.                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                        
### Tests                                                       
 Added two unit tests verifying correct English date output under Chinese (`Locale.CHINA`) and Japanese (`Locale.JAPAN`) locales.

### DIFF
--- a/paimon-api/src/main/java/org/apache/paimon/rest/auth/DLFOpenApiSigner.java
+++ b/paimon-api/src/main/java/org/apache/paimon/rest/auth/DLFOpenApiSigner.java
@@ -31,6 +31,7 @@ import java.time.ZoneId;
 import java.time.format.DateTimeFormatter;
 import java.util.Base64;
 import java.util.HashMap;
+import java.util.Locale;
 import java.util.Map;
 import java.util.TreeMap;
 import java.util.UUID;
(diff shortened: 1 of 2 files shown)


paimon-api/src/main/java/org/apache/paimon/rest/auth/DLFOpenApiSigner.java
```diff
@@ -65,7 +66,7 @@ public class DLFOpenApiSigner implements DLFRequestSigner {
     private static final String API_VERSION = "2026-01-18";
 
     private static final DateTimeFormatter GMT_DATE_FORMATTER =
-            DateTimeFormatter.ofPattern("EEE, dd MMM yyyy HH:mm:ss 'GMT'")
+            DateTimeFormatter.ofPattern("EEE, dd MMM yyyy HH:mm:ss 'GMT'", Locale.ENGLISH)
```
JingsongLi, line 69: Do we need to modify python too?

plusplusjiajia: > Do we need to modify python too?

Python's strftime is locale-safe by default, unlike Java.

Here's the key difference:

Python: On startup, LC_TIME defaults to the C locale ((None, None)), so strftime("%a") always outputs English (e.g., Mon, Tue) regardless of the OS system locale.
Java: DateTimeFormatter uses Locale.getDefault(), which follows the system locale. On a Chinese system, it produces 星期一 instead of Mon.

JingsongLi:
+1

PR #7598, opened 2026-04-05
[core] Add coverage for dedicated-format bundle fallback

QuakeWang:
### Purpose

Add regression coverage for the dedicated-format `writeBundle` fallback path.

The dedicated bundle pass-through approach was removed because it added too much complexity for limited benefit and could regress rolling behavior for large bundles. The existing row-by-row fallback is kept so blob/vector/main child writers can continue to roll within a bundle and preserve row-level side effects.

### Tests

Added/updated tests in paimon-core:

- DedicatedFormatRollingFileWriterTest
- DedicatedFormatRollingFileWriterVectorTest

These cover:

- blob target-file-size rolling during bundle writes;
- vector target-file-size rolling during bundle writes;
- sequence-number side effects for bundle writes;
- main-file index side effects for bundle writes;
- external-storage blob fallback for bundle writes;
- single-use bundle records to ensure the fallback consumes the bundle only once.


JingsongLi:
Can you explain what went wrong?

QuakeWang:
> Can you explain what went wrong?

The old dedicated-format `writeBundle` path was still functionally fine in the simple row-by-row fallback, but it dropped bundle semantics completely by iterating the bundle and calling `write(row)`.

That becomes problematic in the dedicated fan-out path, where one logical bundle has to be written to projected main/blob/vector writers. `BundleRecords` itself does not guarantee replayability, so the same bundle cannot be safely reused across multiple child writers. At the same time, if we forward the bundle directly to restore pass-through, we can bypass row-level side effects in `RowDataFileWriter`, such as sequence-number updates and main-file index writing. We also lose typed / bundle-aware fast paths such as `ArrowBundleRecords`.

This patch makes those constraints explicit: replayable bundles can be passed through safely, non-replayable bundles are materialized once, and dedicated row-data writers preserve the row-level side effects after bundle writes.

So the issue was not that the old row-by-row fallback always corrupted data. The issue was that the dedicated-format bundle path could not preserve bundle semantics safely and correctly once a single bundle needed to be fanned out to multiple writers.

JingsongLi (commented):
Thanks for fixing the dedicated-format bundle write path. This is a complex change touching the append writer hierarchy. Comments:

**Design:**

1. **`ReplayableBundleRecords` / `ProjectableBundleRecords` interfaces**: The opt-in capability pattern is clean. It avoids strengthening the base `BundleRecords` contract and allows runtime capability checks. This is the right extension pattern for the existing architecture.

2. **Materialization strategy**: "Materialize only non-replayable bundles" is correct — if the bundle supports replay, we can iterate it multiple times without copying. The `MaterializedBundleRecords` fallback for non-replayable bundles ensures correctness.

3. **`ArrowBundleRecords` implementing `ProjectableBundleRecords`**: This preserves the Arrow fast path through projection, which avoids row-by-row materialization. The `project()` implementation using `rowType.project(projection)` is elegant.

**Concerns:**

4. **`BundlePassThroughWriter` interface**: This introduces yet another interface in the writer hierarchy. The writer abstraction is getting quite layered: `SingleFileWriter` → `RowDataFileWriter` → `BundleAwareRowDataFileWriter` → implements `BundlePassThroughWriter`. Ensure this doesn't become a maintenance burden.

5. **Sequence number side effects**: The PR description mentions "preserves row-level side effects in bundle writes, including sequence number updates." This is critical — if the sequence number counter isn't advanced correctly for bundle writes, it could cause data consistency issues during compaction. The test `BundleAwareRowDataRollingFileWriterTest` should explicitly verify sequence numbers are correct after bundle writes.

6. **File index writing**: "file-index writing" side effects — does the main-file index get updated correctly when the blob writer uses the bundle path? The `DedicatedFormatRollingFileWriterVectorTest` should cover this.

7. **1054 lines added**: The change is large. Could the interface additions (`ReplayableBundleRecords`, `ProjectableBundleRecords`) be a separate preparatory PR?

Good test coverage with 4 dedicated test classes. Please confirm sequence number correctness in tests.

QuakeWang:
@JingsongLi Thanks for the review.

I added explicit coverage for the side effects:

  - `BundleAwareRowDataRollingFileWriterTest` verifies the sequence counter and min/max sequence numbers after bundle writes.
  - `DedicatedFormatRollingFileWriterVectorTest` verifies sequence side effects for the normal/blob/vector fan-out path.
  - It also verifies the main-file index path by enabling bitmap index on `f0`, writing a single-use bundle, and evaluating the embedded index.

I kept `ReplayableBundleRecords` / `ProjectableBundleRecords` in this PR because they are part of the fix: replayability is needed for safe fan-out, and projectability keeps the Arrow bundle fast path.

The remaining failed CI is a timeout in `SinkSavepointITCase.testRecoverFromSavepoint`, which looks unrelated to this change.

JingsongLi:

JingsongLi (commented):
I think there is still a correctness issue in the projected Arrow bundle fast path.

`ArrowBundleRecords.project(int[] projection)` only changes the `rowType`, but it keeps the original `VectorSchemaRoot`. This is fine for iterator-based consumers, because `ArrowBatchReader` uses the projected `rowType` to map fields by name. However, after this PR, `ProjectedFileWriter.writeBundle` preserves the projected bundle as an `ArrowBundleRecords` and passes it through to the underlying bundle-aware writer.

For real bundle format writers such as Lance/Mosaic/Vortex, `writeBundle` does not consume the bundle through `iterator()`. It detects `ArrowBundleRecords` and writes `getVectorSchemaRoot()` directly. In that path the projection is effectively lost, so a dedicated vector child writer may receive the full original root instead of only its projected physical columns. This can either write the wrong schema/columns or fail with a schema mismatch, depending on the format.

The current `ProjectedFileWriterTest` does not catch this because the mock writer iterates over the bundle. I think we need either:

1. make `ArrowBundleRecords.project(...)` return a truly projected `VectorSchemaRoot`, or
2. avoid preserving the `ArrowBundleRecords` fast-path type after projection unless the root itself is projected.

Could you add a test that exercises a projected `ArrowBundleRecords` with a real `BundleFormatWriter` path, not only an iterator-based mock writer?

QuakeWang:
> I think there is still a correctness issue in the projected Arrow bundle fast path.
> 
> `ArrowBundleRecords.project(int[] projection)` only changes the `rowType`, but it keeps the original `VectorSchemaRoot`. This is fine for iterator-based consumers, because `ArrowBatchReader` uses the projected `rowType` to map fields by name. However, after this PR, `ProjectedFileWriter.writeBundle` preserves the projected bundle as an `ArrowBundleRecords` and passes it through to the underlying bundle-aware writer.
> 
> For real bundle format writers such as Lance/Mosaic/Vortex, `writeBundle` does not consume the bundle through `iterator()`. It detects `ArrowBundleRecords` and writes `getVectorSchemaRoot()` directly. In that path the projection is effectively lost, so a dedicated vector child writer may receive the full original root instead of only its projected physical columns. This can either write the wrong schema/columns or fail with a schema mismatch, depending on the format.
> 
> The current `ProjectedFileWriterTest` does not catch this because the mock writer iterates over the bundle. I think we need either:
> 
> 1. make `ArrowBundleRecords.project(...)` return a truly projected `VectorSchemaRoot`, or
> 2. avoid preserving the `ArrowBundleRecords` fast-path type after projection unless the root itself is projected.
> 
> Could you add a test that exercises a projected `ArrowBundleRecords` with a real `BundleFormatWriter` path, not only an iterator-based mock writer?

@JingsongLi  Thanks for catching this.

I fixed the issue by making `ArrowBundleRecords.project(...)` project the underlying `VectorSchemaRoot` as well, instead of only changing the `rowType`.

I also added a test that goes through the `ArrowBundleWriter.writeBundle` fast path, so it verifies the projected Arrow root that a real bundle writer receives, not just iterator-based consumption.

paimon-core/src/main/java/org/apache/paimon/append/DedicatedFormatRollingFileWriter.java
```diff
@@ -394,16 +384,50 @@ private void handleWriteException(Throwable e) {
...
+        // only explicitly replayable bundles can be forwarded safely without copying.
+        ReplayableBundleRecords replayableBundle =
+                bundle instanceof ReplayableBundleRecords
+                        ? (ReplayableBundleRecords) bundle
+                        : MaterializedBundleRecords.from(bundle, writeSchema);
+
+        try {
+            openWritersIfNeeded();
+            if (blobWriter != null) {
+                blobWriter.writeBundle(replayableBundle);
```
JingsongLi: The new fan-out path writes the whole bundle into each rolling child writer before any target-size check runs. `MultipleBlobFileWriter.writeBundle` delegates to `RollingFileWriterImpl.writeBundle`, and that only calls `reachTargetSize(true)` after `currentWriter.writeBundle(bundle)` has consumed the entire batch. This regresses the old dedicated `writeBundle` path, which iterated rows through `write(row)` and let blob/vector/main writers roll within the bundle; a large bundle with blobs can now produce a single blob file far beyond `blobTargetFileSize` (and similarly oversized main/vector files). Could we preserve the fast path only for writers that can split/check inside the bundle, or add chunking / a fallback so target-file-size rolling still happens during large bundle writes?

JingsongLi:
I found that we paid a relatively high price for the path of the bundle. Perhaps a simple solution is to disable this path, as its effectiveness is limited.

QuakeWang:
> I found that we paid a relatively high price for the path of the bundle. Perhaps a simple solution is to disable this path, as its effectiveness is limited.

@JingsongLi Thanks for pointing this out. I agree that the dedicated bundle pass-through path is not worth the complexity.

I removed the pass-through implementation from this PR and kept the existing row-by-row fallback for `writeBundle`. This avoids the replayability/projection issues and preserves rolling checks inside large bundle writes.

The PR is now intentionally reduced to regression coverage for the fallback behavior:
  - blob target-file-size rolling during bundle writes
  - vector target-file-size rolling during bundle writes
  - sequence-number side effects
  - main-file index side effects
  - external-storage blob fallback

JingsongLi (approved):
+1

PR #7602, opened 2026-04-07
[core] Support BlobView for shared blob data

leaves12138:
## Purpose

This PR adds a `BlobView` read-time reference mode for sharing blob payloads from upstream tables without copying the payload into downstream table files.

The feature stays within the existing `BLOB` type instead of introducing a new logical type:

- `blob-field` marks `BYTES` columns as `BLOB` columns.
- `blob-descriptor-field` stores serialized `BlobDescriptor` bytes inline in data files.
- `blob-view-field` stores serialized `BlobViewStruct` bytes inline in data files and resolves the referenced upstream blob at read time.

## Changes

- Add `BlobView`, `BlobViewStruct`, and shared BLOB byte encode/decode helpers.
- Add `blob-view-field` as a secondary option under `blob-field`.
- Validate descriptor/view fields as subsets of `blob-field`, and reject overlap between descriptor and view fields.
- Reuse the existing BLOB storage paths to write inline descriptor/view metadata.
- Resolve `BlobView` values during table reads using upstream table name, field id, and row id.
- Add Flink and Spark `blob_view(tableName, fieldId, rowId)` functions.
- Update mixed BLOB storage, external storage, Python options, and generated option docs.
- Add common/core/Flink coverage for BlobView serialization, validation, and end-to-end reads.


leaves12138 (commented):
I found a few runtime gaps where `BLOB_REF` support is still incomplete.

paimon-common/src/main/java/org/apache/paimon/data/BlobViewStruct.java
```diff
@@ -0,0 +1,143 @@
...
+ * resolved at read time by scanning the upstream table.
+ */
+public class BlobViewStruct implements Serializable {
+
+    private static final long serialVersionUID = 1L;
+
+    private static final long MAGIC = 0x424C4F4256494557L; // "BLOBVIEW"
+    private static final byte CURRENT_VERSION = 1;
+
+    private final String tableName;
```
JingsongLi: Is this just tableName, or can be a identifier?

paimon-core/src/main/java/org/apache/paimon/table/source/AbstractDataTableRead.java
```diff
@@ -96,6 +121,19 @@ public final RecordReader<InternalRow> createReader(Split split) throws IOExcept
             split = authSplit.split();
             authResult = authSplit.authResult();
         }
+
+        if (catalogContext != null) {
+            RowType rowType = this.readType == null ? schema.logicalRowType() : this.readType;
```
JingsongLi: Why not put these to `DataEvolutionTableRead`?

paimon-common/src/main/java/org/apache/paimon/data/BlobViewStruct.java
leaves12138: Fixed

paimon-core/src/main/java/org/apache/paimon/table/source/AbstractDataTableRead.java
leaves12138: I wish someday, maybe, primary-key table could use blob (blob_view and blob_descriptor) as well~

leaves12138: Fixed

XiaoHongbo-Hope:
+1

XiaoHongbo-Hope (approved):

paimon-common/src/main/java/org/apache/paimon/data/BlobUtils.java
```diff
@@ -0,0 +1,98 @@
...
+            byte[] bytes, @Nullable UriReaderFactory uriReaderFactory, @Nullable FileIO fileIO) {
+        return fromBytes(bytes, uriReaderFactory, fileIO, true);
+    }
+
+    public static Blob fromBytes(
+            byte[] bytes,
+            @Nullable UriReaderFactory uriReaderFactory,
+            @Nullable FileIO fileIO,
+            boolean allowBlobData) {
+        return fromBytes(
```
JingsongLi: Move these methods into `Blob`.

leaves12138: OK

paimon-core/src/main/java/org/apache/paimon/table/source/AbstractDataTableRead.java
```diff
@@ -88,14 +90,37 @@ public final InnerTableRead withReadType(RowType readType) {
         return this;
     }
 
+    protected void configurePrescanRead(InnerTableRead prescanRead) {}
```
JingsongLi: Remove this, just let `DataEvolutionTableRead` handle this.

leaves12138: OK

JingsongLi (approved):
+1

PR #7608, opened 2026-04-07
[python] Fix non-compact decimal and timestamp serialization in BinaryRow

plusplusjiajia:
### Purpose
Previously python's `GenericRowSerializer` treated all decimal and timestamp fields as compact format (value stored directly in the 8-byte fixed slot). Java's BinaryRow uses two formats:

  | Type | Compact | Non-compact |
  |------|---------|-------------|
  | DECIMAL | precision ≤ 18: unscaled long in fixed slot | precision > 18: big-endian bytes in variable area, `(offset << 32 \| length)` in fixed slot |
  | TIMESTAMP | precision ≤ 3: epoch millis in fixed slot | precision > 3: epoch millis in variable area, `(offset << 32 \| nanoOfMillisecond)` in fixed slot |

### DIFF
--- a/paimon-python/pypaimon/table/row/generic_row.py
+++ b/paimon-python/pypaimon/table/row/generic_row.py
@@ -49,17 +49,30 @@ def _decimal_to_unscaled_with_check(d: Decimal, precision: int, scale: int):
 
 
 def _parse_type_precision_scale(data_type):
-    """Parse precision and scale from type string like DECIMAL(38, 10)."""
-    type_str = str(data_type)
+    """Parse precision and scale from type string like DECIMAL(38, 10).
+
+    Falls back to Java-side defaults for parameter-less types:
+      - DECIMAL / NUMERIC -> (10, 0)  (DecimalType.DEFAULT_PRECISION / DEFAULT_SCALE)
+      - TIMESTAMP / TIMESTAMP_LTZ / TIMESTAMP WITH LOCAL TIME ZONE -> (6, 0)
+        (TimestampType.DEFAULT_PRECISION)
+    """
+    type_str = str(data_type).upper().strip()
     if '(' in type_str and ')' in type_str:
         try:
-            params_str = type_str.split('(')[1].split(')')[0]
+            params_str = type_str.split('(', 1)[1].split(')', 1)[0]
             parts = [p.strip() for p in params_str.split(',')]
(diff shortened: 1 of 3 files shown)


JingsongLi (commented):
Important compatibility fix for Python BinaryRow serialization. The non-compact format handling is critical for correctness.

**Review comments:**

1. **Decimal non-compact path**: The `(cursor << 32 | byte_length)` encoding matches Java's `BinaryRow.getDecimal()`. The big-endian signed byte interpretation is correct. Good.

2. **Timestamp non-compact path**: `(cursor << 32 | nanoOfMillisecond)` — note that in Java, `Timestamp.getNanoOfMillisecond()` returns the nanoseconds portion *within* the millisecond (0-999999). Your `_datetime_to_millis_and_nanos` converts microseconds to nanos: `(value.microsecond % 1000) * 1000`. This loses sub-microsecond precision, which is fine since Python's datetime only has microsecond resolution. Document this.

3. **`_parse_type_precision_scale` fallback**: Defaulting to `(10, 0)` for bare `DECIMAL` and `(6, 0)` for bare `TIMESTAMP` matches Java defaults. Good.

4. **Precision overflow → None**: `_parse_decimal` returns `None` on precision overflow for the non-compact path. Is this consistent with how Java handles it? Java's `BinaryRow.getDecimal()` returns null for overflow, so this is correct. But ensure callers handle `None` properly (e.g., don't crash on `None` in downstream operations).

5. **`_decimal_to_unscaled_with_check` rounding**: Using `ROUND_HALF_UP` matches Java's `DecimalData.fromBigDecimal()` behavior. Good.

6. **Test coverage**: Having separate `decimal_test.py` and `timestamp_test.py` is appropriate. Make sure there are tests for:
   - precision=18 (compact boundary)
   - precision=19 (non-compact start)
   - precision=38 (max)
   - Negative decimals in non-compact format
   - Timestamp precision=3 (compact) vs precision=6 (non-compact)

7. **Performance note**: The `_decimal_to_unscaled_with_check` function does string manipulation (`int(''.join(str(x) for x in digits))`). For high-throughput serialization this could be a bottleneck. Consider `int(Decimal.scaleb(scale))` or similar numeric approach.

LGTM with the above addressed.

JingsongLi:
+1

Issue #7617, opened 2026-04-09
[Bug] cdc filter pattern does not handle special character


ziyanTOP:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Paimon version

1.4.1-rc1

### Compute Engine

flink 1.20.3

### Minimal reproduce step

create a mysql table with `$` in name

```sql
create database if not exists `test_database`;

use `test_database`;

create table if not exists `test_table_0` (
    id int not null primary key,
    key varchar(50) null
);
create table if not exists `test_table_1_$` (
    id int not null primary key,
    key varchar(50) null
);
```

then use paimon flink action mysql_sync_database combine mode to synchronized mysql database
```shell
flink run paimon-flink-action-1.4.1.jar \
    mysql_sync_database \
    --mode combined \
    --warehouse s3://bucket/path --database test_db \
    --mysql_conf hostname=mysql_host \
    --mysql_conf username=mysql_user --mysql_conf password=mysql_pass \
    --mysql_conf database-name=test_database \
    --catalog_conf metastore=filesystem \
    --catalog_conf s3.endpoint=s3_endpoint --catalog_conf s3.access-key=s3_ak \
    --catalog_conf s3.secret-key=s3_sk --catalog_conf s3.region=s3_region \
    --excluding_tables test_table_1.*
```

### What doesn't meet your expectations?

the table which name matches the `excluding_tables` pattern still be synchronized

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

ziyanTOP:
fixed with pr merged

PR #7619, opened 2026-04-10
[cdc] fix special regex characters unescaped in excluded table patterns

ziyanTOP:
### Purpose
In combined mode, when building the excluding pattern for CDC table monitoring, special regex characters (e.g., `$`) in table names were not properly escaped. This caused incorrect table matching.

**Fix:** Use `Pattern.quote()` to escape the full qualified table name before inserting it into the regex pattern, ensuring exact literal matching.

#7617


JingsongLi:
+1

PR #7622, opened 2026-04-10
[core] Allow pk-clustering-override without explicit DV for first-row merge engine

JingsongLi:
### Purpose

First-row merge engine already has built-in deletion vector semantics, so requiring users to explicitly enable deletion-vectors is unnecessary.

### Tests

`ClusteringTableTest.testFirstRowWithoutDeletionVectors`


LsomeYeah (approved):
+1

PR #7624, opened 2026-04-10
[spark] Fix exprId mismatch in MERGE INTO by constructing Join plan directly

Zouxxyy:
### Purpose

Fix exprId mismatch bug in MERGE INTO when action expressions reference both source
and target columns (e.g., `COALESCE(src.b, dest.b)`).

Replace Dataset API join (`sourceDS.join(targetDS, ...)`) with manual `Join` plan node
construction, following the same pattern as Spark's `RewriteMergeIntoTable`. This
preserves original exprIds and eliminates the need for position-based rebinding.

### Tests

Added 3 test cases in `MergeIntoTableTestBase` covering coalesce with dual-column
references: temp view source, Paimon table source, and subquery source.


JingsongLi:
+1

PR #7628, opened 2026-04-11
[core] Support partition predicate pushdown for PartitionsTable

sundapeng:
### Purpose

`PartitionsTable.withFilter()` was a no-op (`// TODO` at line 189), causing full manifest scans when querying with partition filters like `SELECT * FROM t$partitions WHERE partition = 'dt=20260410'`. This adds partition predicate pushdown following the same pattern established by BucketsTable (#7592), FilesTable (#7376), ManifestsTable (#7310), and ConsumersTable (#7329).

The implementation uses a dual-path filtering strategy:
- **Catalog path**: preserves `catalog.listPartitions()` call and filters results in memory, keeping metadata columns (`created_at`, `created_by`, `updated_by`, `options`) intact
- **TableScan fallback path**: pushes predicate down to `InnerTableScan.withPartitionFilter()` for manifest-level pruning

Also refactors `PartitionPredicateHelper.applyPartitionFilter()` into a two-step build+apply pattern (`buildPartitionPredicate()` + apply), and extends `parsePartitionSpec()` to support PartitionsTable's `key=value/key=value` format in addition to the existing `{value1, value2}` format.

### Tests

- Equal filter on single partition key
- IN filter on multiple partition values
- No-match filter returns empty result
- Non-partition column filter safely ignored
- Multi-column partition keys with Equal and IN filters
- Existing `BucketsTableTest` passes (verifies refactored helper is backward-compatible)

### API and Format

### Documentation

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/utils/PartitionPredicateHelper.java
+++ b/paimon-core/src/main/java/org/apache/paimon/utils/PartitionPredicateHelper.java
@@ -18,6 +18,7 @@
 
 package org.apache.paimon.utils;
 
+import org.apache.paimon.predicate.AlwaysFalse;
 import org.apache.paimon.predicate.Equal;
 import org.apache.paimon.predicate.In;
 import org.apache.paimon.predicate.LeafBinaryFunction;
@@ -44,26 +45,33 @@ public class PartitionPredicateHelper {
      * Build a partition-typed predicate from a string-based leaf predicate on the "partition"
      * column.
      *
-     * @return the predicate on partition fields, or {@code null} if the partition spec is invalid
-     *     (indicating no results should be returned)
+     * @return {@code null} if the predicate type is unsupported for pushdown (caller should skip
+     *     pushdown), {@link PredicateBuilder#alwaysFalse()} if no partition can match (caller
+     *     should return empty), or a normal predicate to push down.
      */
     @Nullable
     public static Predicate buildPartitionPredicate(
-            LeafPredicate partitionPredicate, List<String> partitionKeys, RowType partitionType) {
+            LeafPredicate partitionPredicate,
+            List<String> partitionKeys,
+            RowType partitionType,
+            String defaultPartitionName) {
         if (partitionPredicate.function() instanceof Equal) {
             LinkedHashMap<String, String> partSpec =
                     parsePartitionSpec(
                             partitionPredicate.literals().get(0).toString(), partitionKeys);
             if (partSpec == null) {
-                return null;
+                return PredicateBuilder.alwaysFalse();
             }
             PredicateBuilder partBuilder = new PredicateBuilder(partitionType);
             List<Predicate> predicates = new ArrayList<>();
             for (int i = 0; i < partitionKeys.size(); i++) {
-                Object value =
-                        TypeUtils.castFromString(
-                                partSpec.get(partitionKeys.get(i)), partitionType.getTypeAt(i));
-                predicates.add(partBuilder.equal(i, value));
+                String strValue = partSpec.get(partitionKeys.get(i));
+                if (defaultPartitionName.equals(strValue)) {
+                    predicates.add(partBuilder.isNull(i));
+                } else {
+                    Object value = TypeUtils.castFromString(strValue, partitionType.getTypeAt(i));
+                    predicates.add(partBuilder.equal(i, value));
+                }
             }
             return PredicateBuilder.and(predicates);
         } else if (partitionPredicate.function() instanceof In) {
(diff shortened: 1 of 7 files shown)


paimon-core/src/main/java/org/apache/paimon/utils/PartitionPredicateHelper.java
```diff
@@ -36,27 +36,36 @@
...
+                return null;
             }
-            snapshotReader.withPartitionFilter(partSpec);
+            PredicateBuilder partBuilder = new PredicateBuilder(partitionType);
+            List<Predicate> predicates = new ArrayList<>();
+            for (int i = 0; i < partitionKeys.size(); i++) {
+                Object value =
+                        TypeUtils.castFromString(
+                                partSpec.get(partitionKeys.get(i)), partitionType.getTypeAt(i));
+                predicates.add(partBuilder.equal(i, value));
```
littlecoder04: We should consider default partitions here. If the value is the default partition name, this should become `isNull(...)` rather than `equal(...)`.

sundapeng:
## E2E Benchmark v2: 3 Rounds × 9 Samples (with bug fixes)

### Environment
- ECS EMR cn-hongkong (16 vCPU, 32GB), DLF REST Catalog, Spark 3.5.3, Java 11
- 4 tables: 100 / 1,000 / 5,000 / 10,000 partitions
- 3 full rounds per scenario, 3 internal runs per query = **9 samples per data point**

### Bug Fixes in Latest Push
1. `buildPartitionPredicate()` handles `__DEFAULT_PARTITION__` → `isNull()` (fixes `NumberFormatException` on INT partitions)
2. Unsupported predicates skip pushdown instead of returning empty results
3. `defaultPartitionName` passed through all system table callers

### Catalog Path (DLF REST API)

| Partitions | Query | Before (ms) | After (ms) | Speedup |
|------------|-------|-------------|------------|---------|
| 100 | no_filter | 150 | 141 | 1.06x |
| 100 | equal_filter | 180 | 171 | 1.05x |
| 100 | in_filter | 132 | 126 | 1.05x |
| 1,000 | no_filter | 152 | 157 | 0.97x |
| 1,000 | equal_filter | 127 | 112 | **1.14x** |
| 1,000 | in_filter | 121 | 103 | **1.18x** |
| 5,000 | no_filter | 251 | 300 | 0.84x |
| 5,000 | equal_filter | 247 | 229 | **1.08x** |
| 5,000 | in_filter | 240 | 221 | **1.09x** |
| 10,000 | no_filter | 410 | 404 | 1.02x |
| 10,000 | equal_filter | 402 | 374 | **1.08x** |
| 10,000 | in_filter | 401 | 360 | **1.11x** |

**Filtered avg at 10K: +8.6%**

### Scan Path (Time-Travel `VERSION AS OF 1`)

| Partitions | Query | Before (ms) | After (ms) | Speedup |
|------------|-------|-------------|------------|---------|
| 100 | no_filter | 223 | 223 | 1.00x |
| 100 | equal_filter | 242 | 239 | 1.01x |
| 100 | in_filter | 207 | 222 | 0.93x |
| 1,000 | no_filter | 206 | 208 | 0.99x |
| 1,000 | equal_filter | 167 | 156 | **1.07x** |
| 1,000 | in_filter | 168 | 161 | 1.04x |
| 5,000 | no_filter | 235 | 256 | 0.92x |
| 5,000 | equal_filter | 179 | 154 | **1.16x** |
| 5,000 | in_filter | 193 | 152 | **1.27x** |
| 10,000 | no_filter | 240 | 234 | 1.02x |
| 10,000 | equal_filter | 199 | 142 | **1.40x** |
| 10,000 | in_filter | 201 | 156 | **1.29x** |

**Filtered avg at 10K: +25.5%**

### Summary

| Path | Best Single Speedup | 10K Filtered Avg |
|------|---------------------|------------------|
| Catalog (DLF) | 1.18x (1K in_filter) | **+8.6%** |
| Scan (time-travel) | **1.40x** (10K equal_filter) | **+25.5%** |

No regression in unfiltered baselines. Improvement scales with partition count.

paimon-core/src/main/java/org/apache/paimon/utils/PartitionPredicateHelper.java
```diff
@@ -97,29 +121,69 @@ public static boolean applyPartitionFilter(
...
+            String defaultPartitionName) {
+        if (partitionPredicate == null) {
+            return true;
+        }
+
+        Predicate predicate =
+                buildPartitionPredicate(
+                        partitionPredicate, partitionKeys, partitionType, defaultPartitionName);
+        if (predicate == null) {
+            return false;
```
JingsongLi, line 149: false? No record will be returned? {@code null} if the predicate cannot be pushed down (unsupported predicate type or invalid partition spec.

sundapeng: Good catch. The issue is that `buildPartitionPredicate` returned `null` for two different reasons:

1. **Unsupported predicate type** (e.g., `isNotNull`) → should skip pushdown, return all data
2. **Invalid partition spec** (e.g., partial key `[2]` on a 2-column partition table) → no match, return empty is correct

Simply changing `return false` to `return true` would break `FilesTableTest.testReadWithNotFullPartitionKey` (case 2).

Fixed by using `@Nullable Predicate` with clear semantics: `null` means unsupported (skip pushdown), `PredicateBuilder.alwaysFalse()` means no match (return empty), and a normal predicate means apply pushdown. This reuses the existing `AlwaysFalse` predicate instead of introducing a wrapper class.

Added `testBucketsTableUnsupportedPredicateFallsBackToFullScan` to cover the regression.

paimon-core/src/main/java/org/apache/paimon/utils/PartitionPredicateHelper.java
```diff
@@ -68,27 +85,36 @@ public static boolean applyPartitionFilter(
...
+                return PredicateBuilder.alwaysFalse();
+            }
+            PredicateBuilder partBuilder = new PredicateBuilder(partitionType);
+            List<Predicate> predicates = new ArrayList<>();
+            for (int i = 0; i < partitionKeys.size(); i++) {
+                String strValue = partSpec.get(partitionKeys.get(i));
+                if (defaultPartitionName.equals(strValue)) {
+                    predicates.add(partBuilder.isNull(i));
+                } else {
+                    Object value = TypeUtils.castFromString(strValue, partitionType.getTypeAt(i));
```
JingsongLi, line 117: This conversion is not semantically equivalent for `PartitionsTable.partition`, which is a string column like `pt=10` or `dt=20260410/region=1`. For non-equality predicates the SQL predicate is evaluated lexicographically on that full string, but this branch casts each partition value to its physical type and ANDs the same binary function across partition columns. For example, on an INT partition, `partition > "pt=2"` should not match `pt=10` under string ordering (`"pt=10" < "pt=2"`), while the pushed predicate becomes `pt > 2` and does match it. The catalog path filters by testing the original string predicate, so the manifest fallback/time-travel path can return different results. I think only equality/IN full-spec predicates are safe to translate this way; other partition-string predicates should skip pushdown unless the string semantics can be preserved.

Akash3121 (changes requested):
@sundapeng there are conflicts please rebase

PR #7632, opened 2026-04-13
[common] move FileType from paimon-core to paimon-common

LsomeYeah:
### Purpose
Move `FileType` from paimon-core to paimon-common since it only depends on Path (from paimon-api). This allows users who depend on paimon-common to classify files without pulling in the heavier paimon-core module.


JingsongLi:
+1

PR #7633, opened 2026-04-13
[core] Improve error message for fallback branch schema mismatch

LsomeYeah:
## Summary
- Improve the error message in `FallbackReadFileStoreTable.validateSchema()` to explicitly mention that the validation is triggered by the `scan.fallback-branch` option, helping users quickly identify the root cause of the schema mismatch.

## Test plan
- [x] Compile verified

🤖 Generated with [Qoder][https://qoder.com]


JingsongLi:
+1

PR #7634, opened 2026-04-13
[core] Fix DataEvolutionCompactTask potential resource leak

yuzelin:
### Purpose
DataEvolutionCompactTask should close writer after compact.


JingsongLi:
+1

PR #7641, opened 2026-04-14
[rest]  Align view permission handling with table in RESTCatalog.

shyjsarah:
### Purpose

  Align view permission handling with table in RESTCatalog.

  Currently, table operations in `RESTCatalog` catch `ForbiddenException` and throw `TableNoPermissionException`, but view operations
  (`getView`, `createView`, `dropView`, `renameView`, `alterView`, `listViews`, etc.) do not handle `ForbiddenException` at all.

  This PR adds:
  - `ViewNoPermissionException` in `Catalog.java`, mirroring `TableNoPermissionException`
  - `ForbiddenException` handling for all view operations in `RESTCatalog`
  - `DatabaseNoPermissionException` handling for list view operations (consistent with list table behavior)

  ### Tests

  - Added `testApiWhenViewNoPermission` in `RESTCatalogTest` to verify that `getView`, `dropView`, `renameView`, and `alterView` throw
  `ViewNoPermissionException` when permission is revoked
  - Updated `RESTCatalogServer` to support `noPermissionViews` and return 403 for forbidden view operations

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
@@ -1436,7 +1436,7 @@ public interface Catalog extends AutoCloseable {
     /** Exception for trying to operate on the view that doesn't have permission. */
     class ViewNoPermissionException extends RuntimeException {
 
-        private static final String MSG = "View %s has no permission. Cause by %s.";
+        private static final String MSG = "View %s has no permission. Caused by %s.";
 
         private final Identifier identifier;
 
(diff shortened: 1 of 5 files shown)


paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
```diff
@@ -1433,6 +1433,33 @@ public String getTableId() {
         }
     }
 
+    /** Exception for trying to operate on the view that doesn't have permission. */
+    class ViewNoPermissionException extends RuntimeException {
+
+        private static final String MSG = "View %s has no permission. Cause by %s.";
```
JingsongLi: Caused

shyjsarah: updated

JingsongLi:
+1

PR #7642, opened 2026-04-14
[core] Avoid key bytes OOM in ClusteringFileRewriter.sortAndRewriteFile

JingsongLi:
### Purpose

Avoid key bytes OOM in ClusteringFileRewriter.sortAndRewriteFile. Removing the in-memory List<byte[]> collectedKeys and the batchPutIndex method eliminates the unbounded memory accumulation.

### Tests

Existing Tests.


paimon-core/src/main/java/org/apache/paimon/mergetree/compact/clustering/ClusteringFileRewriter.java
```diff
@@ -115,20 +112,16 @@ public ClusteringFileRewriter(
     }
 
     /**
-     * Sort and rewrite unsorted files by clustering columns. Reads all KeyValue records, sorts them
+     * Sort and rewrite unsorted file by clustering columns. Reads all KeyValue records, sorts them
      * using an external sort buffer, and writes to new level-1 files. Checks the key index inline
      * during writing to handle deduplication (FIRST_ROW skips duplicates, DEDUPLICATE marks old
      * positions in DV) and updates the index without re-reading the output files.
```
LsomeYeah: Doc seems to be not accurate. When `rebuildIndex`, the file now will be re-read.

LsomeYeah (approved):
+1

PR #7645, opened 2026-04-14
[core] Fix initRow in AggregateMergeFunction and PartialUpdateMergeFunction to set all fields including nullable ones

lxy-9602:
### Purpose
Fix `initRow` in `AggregateMergeFunction` and `PartialUpdateMergeFunction` to always call `setField` for all fields, including nullable ones. Previously, nullable fields were skipped during `initRow`, which could cause inconsistent aggregation results when delete records contain null values in nullable columns.

### Tests
AggregateMergeFunctionTest
PartialUpdateMergeFunctionTest


leaves12138 (approved):
+1

PR #7648, opened 2026-04-15
[spark] Add paimon spark4.1 module

kerwin-zk:
### Purpose
Add paimon spark4.1 module

### Tests
CI


pom.xml
```diff
@@ -427,6 +427,7 @@ under the License.
             <modules>
                 <module>paimon-spark/paimon-spark4-common</module>
                 <module>paimon-spark/paimon-spark-4.0</module>
+                <module>paimon-spark/paimon-spark-4.1</module>
```
YannByron, line 430: Should we consider to update `<paimon-spark-common.spark.version>` to 4.1.1 so that `paimon-spark4-common` and `paimon-spark-ut` are compiled against the latest Spark version in the profile?

paimon-spark/paimon-spark-4.1/src/main/scala/org/apache/paimon/spark/catalyst/optimizer/MergePaimonScalarSubqueries.scala
YannByron, line 1: The `MergePaimonScalarSubqueries` in `paimon-spark-4.1` is identical to the one in `paimon-spark-4.0`. Would it be better to move these version-agnostic implementations into `paimon-spark4-common` instead of duplicating them in each version-specific module?

YannByron: Maybe `hive-test-udfs.jar`、`hive-site.xml` and `log4j2-test.properties`, consider these as well.

kerwin-zk: @YannByron  `log4j2-test.properties` and `function/hive-test-udfs.jar` have been removed from `paimon-spark-4.1` — `paimon-spark-ut` already ships identical copies on the test classpath. `hive-site.xml` is kept per-module on purpose. `paimon-spark-3.2` through `paimon-spark-4.0` each carry their own copy with a distinct `hive.metastore.uris` port (3.2→9087, 3.3→9088, 3.4→9089, 3.5→9090, 4.0→9090) so the ITCases can run in parallel without port collisions; `paimon-spark-4.1` follows the same pattern with 9091.

YannByron:
@kerwin-zk I have some thoughts which may be worth considering.
1. These classes (`PaimonFormatTable`, `ScanPlanHelper`, `PaimonBatchWrite`, `AbstractPaimonSparkSqlExtensionsParser`, `PaimonStrategyHelper`, `PaimonCreateTableAsSelectStrategy` and `PaimonRelation`) maybe don't have to copy to `paimon-spark-4.0` cause these use the stable api. Please check.

2. And For classes (`AssignmentAlignmentHelper`, `PaimonMergeIntoResolver`, `PaimonMergeIntoResolverBase`, `RewriteUpsertTable`, `MergeIntoPaimonTable`and `MergeIntoPaimonDataEvolutionTable`), can we identify specific incompatibilities and then draw them into 'SparkShim' to solve them?

kerwin-zk:
> @kerwin-zk I have some thoughts which may be worth considering.
> 
> 1. These classes (`PaimonFormatTable`, `ScanPlanHelper`, `PaimonBatchWrite`, `AbstractPaimonSparkSqlExtensionsParser`, `PaimonStrategyHelper`, `PaimonCreateTableAsSelectStrategy` and `PaimonRelation`) maybe don't have to copy to `paimon-spark-4.0` cause these use the stable api. Please check.
> 2. And For classes (`AssignmentAlignmentHelper`, `PaimonMergeIntoResolver`, `PaimonMergeIntoResolverBase`, `RewriteUpsertTable`, `MergeIntoPaimonTable`and `MergeIntoPaimonDataEvolutionTable`), can we identify specific incompatibilities and then draw them into 'SparkShim' to solve them?

@YannByron `PaimonRelation` has been removed, and the rest need some refactoring. After this PR is merged, I will submit follow-up PRs to refactor and optimize this part of the logic.

YannByron:
+1

PR #7651, opened 2026-04-15
[python] Fix Identifier.is_system_table()

plusplusjiajia:
### Purpose
 The previous implementation checked `self.object.startswith('$')`, but Paimon system-table object names are `<base>$<system_table>` (e.g.`orders$snapshots`), not `$<system_table>`.


JingsongLi:
+1

PR #7652, opened 2026-04-15
[core] Fix FieldListaggAgg distinct mode incorrectly skipping values …

lszskye:
### Purpose
Fix a bug in `FieldListaggAgg.agg()` where the distinct deduplication logic
incorrectly uses substring matching (`BinaryString.contains()`) instead of
exact token matching, causing valid values to be silently dropped.

For example, with delimiter `,`:
- accumulator = `"abc,def,asd"`
- inputField = `"ab,xy"`
- Token `"ab"` is incorrectly skipped because `"abc,def,asd".contains("ab")`
  returns `true`
- Result: `"abc,def,asd,xy"` (missing `"ab"`)
- Expected: `"abc,def,asd,ab,xy"`

### Tests
testFieldListAggDistinctShouldNotMatchSubstring
testFieldListAggDistinctSubstringWithCustomDelimiter


JingsongLi (approved):
+1

PR #7653, opened 2026-04-15
[core] fix BlobDescriptor toString format

steFaiz:
### Purpose
The toString method of BlobDescriptor has a isolated `'`. It makes parsing the string hard.
image

### Tests
None


JingsongLi:
+1

PR #7654, opened 2026-04-15
[core] Optimize the printing of error logs for snapshot expiration

linchen1101:
### Purpose
When throwing a snapshot expiration exception, log the earliest and latest snapshot information


Aitozi (approved):
+1

Issue #7052, opened 2026-01-15
[Bug] Failed to Compaction due to the expection: 'Failed to read 4 bytes'


hexiaoshu96:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Paimon version

PaimonVersion: 0.9
Storage: LocalFile

### Compute Engine

FlinkVersion: 1.17.2

### Minimal reproduce step

When using Paimon for partial column updates, we mistakenly wrote the primary key to the sequence group. This caused some interesting phenomena. If the version column of the sequence group is null and an update occurs, a ParquetDecodingException will result. Below is the process we reproduced in our local test.

- Create Table DDL

CREATE TABLE if not exists test.paimon_test_partical
(
  order_id BIGINT,
  order_key STRING,
  send_order_name STRING,
  get_order_name STRING,
  send_order_version BIGINT,
  get_order_version BIGINT,
  PRIMARY KEY (order_id,order_key) NOT ENFORCED
)
WITH
(
    'connector' = 'paimon',
    'bucket' = '1',
    'file.format' = 'parquet',
    'file.compression' = 'zstd',
    'merge-engine' = 'partial-update',
    'fields.send_order_version.sequence-group' = 'send_order_name,order_key',
    'fields.get_order_version.sequence-group' = 'get_order_name',
    'snapshot.num-retained.min' = '1',
    'snapshot.num-retained.max' = '1',
    'full-compaction.delta-commits' = '10'
);
- Batch Insert SQL
1.  First, execute the first statement to write
insert into test.paimon_test_partical(order_id,order_key,send_order_name,get_order_name,get_order_version) 
values 
(1,'a','test_send','test_get',1766632140),
(2,'a','test_send','test_get',1766632140),
(3,'a','test_send','test_get',1766632140),
(1,'b','test_send','test_get',1766632140),
(2,'b','test_send','test_get',1766632140),
(3,'b','test_send','test_get',1766632140),
(1,'c','test_send','test_get',1766632140),
(2,'c','test_send','test_get',1766632140),
(3,'c','test_send','test_get',1766632140)

QueryResult: 

<img width="2196" height="378" alt="Image" src="https://github.com/user-attachments/assets/ae00701f-5ada-4808-8409-a20563d3ecf7" />

2. Then write a portion of the data to overwrite.
insert into test.paimon_test_partical(order_id,order_key,send_order_name,get_order_name,get_order_version) 
values 
[...]

PR #7656, opened 2026-04-15
[core] Add validation to prevent primary key in sequence-group (#7052)

dark2momo:
When a primary key field is configured in a sequence-group of partial-update merge engine, it causes Parquet decoding failures during compaction because the key field may be set to null. This commit adds early validation at configuration parsing time to reject such invalid configurations with a clear error message.


JingsongLi (commented):
Review: The validation logic is correct, covering both value fields and sequence fields themselves. Good refactor with protectedFieldName. Tests cover the three distinct scenarios well. Minor nits: consider extracting the duplicated error message string, and the test name testMultiPrimaryKeyCannotBeInSequenceGroup is slightly misleading since it only uses a single PK. Overall looks good.

JingsongLi (commented):
## Review: [core] Add validation to prevent primary key in sequence-group (#7052)

**Correctness:** The validation logic is correct. It covers both places where a PK could appear: (1) value fields in the forEach lambda, and (2) sequence fields themselves in the "add self" loop. The reordering of allSequenceFields.add after the PK check avoids partial state mutations on exception.

**Design:** The protectedFieldName variable is a good refactor eliminating a redundant fieldNames.get(field) call. Error messages are clear and actionable.

**Tests:** Good coverage of three distinct scenarios -- PK as sequence field, PK in multi-field sequence key, and PK in value-field list.

**Minor suggestions (non-blocking):**
1. The duplicated error format string could be extracted to a constant.
2. primaryKeys.contains() on a List is O(n) per call; wrapping in a HashSet would be O(1) but is not necessary given small PK counts at config time.
3. Test name testMultiPrimaryKeyCannotBeInSequenceGroup implies multiple PKs but only uses one (ImmutableList.of("f2")).

Overall this is a well-scoped fix that moves failure detection from a confusing Parquet runtime crash to a clear config-time validation error. Looks good.

JingsongLi (changes requested):
-1 Please fix the tests.

dark2momo:
@JingsongLi 
Thanks for the review! Sorry for the oversight — since the broken tests were all **existing** test cases (not newly added ones in this PR), I should have fixed them in the original commit.

To highlight: this PR modifies two **existing** test cases to adapt to the new validation:

- **`PrimaryKeySimpleTableTest#testPartialUpdateWithAgg`**: changed `fields.a.sequence-group` → `fields.b.sequence-group` (since `a` is a primary key)
- **`PartialUpdateITCase#testProjectPushDownWithLookupChangelogProducer`**: changed `fields.a.sequence-group='j'` → `fields.a.sequence-group='b,c'` (since `j` is a primary key)

The test intent remains the same — only the invalid PK-in-sequence-group config was replaced with a valid non-PK alternative.

Regarding the minor suggestions:

1. **Extract error string to constant** — done. Added `SEQUENCE_GROUP_PK_ERROR` to eliminate the duplicated format string.
2. **`HashSet` for `primaryKeys.contains()`** — left as `List` as you noted, since PK counts are small at config time.
3. **Renamed misleading test** — `testMultiPrimaryKeyCannotBeInSequenceGroup` → `testPrimaryKeyCannotBeInSequenceGroupValue` to accurately reflect what it tests.

JingsongLi (commented):
LGTM, thanks.

PR #7657, opened 2026-04-15
[spark] Push down variant_get for Spark 4.0

chenghuichen:
### Purpose
Queries like `SELECT variant_get(v, '$.age', 'int') FROM T` on a shredded Variant column still read all sub-columns and reassemble the full binary Variant, leaving Paimon's `VariantRowType` / `clipVariantType` infrastructure unused.

This PR adds `PushDownVariantExtract` (Spark 4 only), a Catalyst optimizer rule that replaces `VariantGet` with `GetStructField` and sets `variantProjections` on `PaimonScan`, so only the accessed `typed_value.*` Parquet sub-columns are read.

The rule runs in the "User Provided Optimizers" batch (via `experimentalMethods.extraOptimizations`) to ensure it fires after `V2ScanRelationPushDown` has built the scan relation.

Part of #4471 

Note: Spark 4.0 lacks a V2-compatible variant push-down interface (`SupportsPushDownVariantExtractions` was introduced in 4.1), so registering a custom optimizer rule via `experimentalMethods.extraOptimizations` is the right fit for 4.0. For a future `paimon-spark-4.1` module, a cleaner approach would be implementing `SupportsPushDownVariantExtractions` on `PaimonScan` and letting Spark's built-in `V2ScanRelationPushDown` handle the rewrite natively.

### Tests
VariantTest.scala::VariantPushDownPlanTest (paimon-spark-4.0)


JingsongLi:
cc @Zouxxyy to take a look

chenghuichen:
The variant_get pushdown approach is too hacky for Spark 4.0. The community has decided to target Spark 4.1 (and later) only. Close this PR.

PR #7663, opened 2026-04-16
[spark] Support CREATE TABLE LIKE

Zouxxyy:
### Purpose

Support `CREATE TABLE LIKE` in Paimon Spark integration on Spark 3.4+.

Behavior:
- `SparkCatalog`: rewrite `CREATE TABLE LIKE` to the Paimon command, except `STORED AS`, which remains unsupported.
- If `USING` is omitted in `SparkCatalog`, inherit the source table provider.
- `SparkGenericCatalog`: rewrite only when `USING paimon` is specified; otherwise keep Spark native behavior.
- Copy source table comment and properties only when source and target providers match. If providers differ, keep only the comment and log a warning.

### Tests

Updated/covered by:
- `DDLTest`
- `DDLWithHiveCatalogTest`
- `PaimonFormatTableTest`

### DIFF
--- a/paimon-spark/paimon-spark-ut/src/test/scala/org/apache/paimon/spark/sql/DDLTestBase.scala
+++ b/paimon-spark/paimon-spark-ut/src/test/scala/org/apache/paimon/spark/sql/DDLTestBase.scala
@@ -31,6 +31,8 @@ import org.junit.jupiter.api.Assertions
 import java.sql.{Date, Timestamp}
 import java.time.LocalDateTime
 
+import scala.collection.JavaConverters._
+
 abstract class DDLTestBase extends PaimonSparkTestBase {
 
   import testImplicits._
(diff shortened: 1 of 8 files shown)


LinMingQiang:
Is it possible to support creating a new table from a branch?

Zouxxyy:
This is supported now. I added a test to verify CREATE TABLE LIKE works with a branch source like `source_tbl$branch_test_branch` under `SparkCatalog`, and the target table is created from the branch metadata rather than the main table metadata.

paimon-spark/paimon-spark-ut/src/test/scala/org/apache/paimon/spark/sql/DDLTestBase.scala
```diff
@@ -133,6 +135,83 @@ abstract class DDLTestBase extends PaimonSparkTestBase {
...
+            |""".stripMargin)
+      sql("INSERT INTO source_tbl VALUES (1, 'a', 'p1')")
+
+      checkAnswer(
+        sql("CALL paimon.sys.create_branch(table => 'test.source_tbl', branch => 'test_branch')"),
+        Row(true) :: Nil)
+      sql("ALTER TABLE `source_tbl$branch_test_branch` ADD COLUMNS(extra STRING)")
+
+      sql("""
+            |CREATE TABLE target_tbl
```
JingsongLi, line 197: Can you also add tests for `IF NOT EXISTS` and `STORED AS` error path?

JingsongLi (approved):
+1

PR #7665, opened 2026-04-17
[core] Do not pushdown limit when non partition filter is present

XiaoHongbo-Hope:
### Purpose
If limit pushdown is applied before non-partition filters are evaluated, scanning may stop too early and return fewer rows than requested. To fix this, limit pushdown is disabled when non-partition filters exist.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/source/DataTableBatchScan.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/source/DataTableBatchScan.java
@@ -21,7 +21,6 @@ package org.apache.paimon.table.source;
 import org.apache.paimon.CoreOptions;
 import org.apache.paimon.manifest.PartitionEntry;
 import org.apache.paimon.predicate.Predicate;
-import org.apache.paimon.predicate.PredicateVisitor;
 import org.apache.paimon.predicate.SortValue;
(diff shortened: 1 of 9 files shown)


XiaoHongbo-Hope:
@wwj6591812 Could you please help take a look at this PR.

paimon-core/src/main/java/org/apache/paimon/table/source/DataTableBatchScan.java
```diff
@@ -126,7 +132,7 @@ public List<PartitionEntry> listPartitionEntries() {
     }
 
     private Optional<StartingScanner.Result> applyPushDownLimit() {
```
JingsongLi, line 128: Maybe this should be in `SnapshotReaderImpl`, `withFilter` already know if there is non-part filter. (`!pair.getRight().isEmpty()`)

XiaoHongbo-Hope: > Maybe this should be in `SnapshotReaderImpl`, `withFilter` already know if there is non-part filter. (`!pair.getRight().isEmpty()`)

Updated

JingsongLi:
+1

wwj6591812:
+1

PR #7666, opened 2026-04-17
[python] Push down IndexedSplit row ranges to Lance reader

chenghuichen:
### Purpose
After a global index search returns a sparse set of matching row IDs, the data fetch path previously used RowIdFilterRecordBatchReader for Lance — reading the entire Lance file and discarding non-matching rows in memory. 

This change converts the global row ID ranges to local file offsets and pushes them down to lance.file.LanceFileReader.take_rows(), so only the matched rows are physically read from disk, leveraging Lance's native random-access capability.

### Tests
- paimon-python/pypaimon/tests/data_evolution_test.py

### DIFF
--- a/paimon-python/pypaimon/read/reader/format_lance_reader.py
+++ b/paimon-python/pypaimon/read/reader/format_lance_reader.py
@@ -16,7 +16,7 @@
 #  limitations under the License.
 ################################################################################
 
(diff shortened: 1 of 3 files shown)


XiaoHongbo-Hope:
+1

JingsongLi:
+1

PR #7668, opened 2026-04-18
[jindo] Extract doDelete() template method from HadoopCompliantFileIO

sundapeng:
### Purpose

Refactor `HadoopCompliantFileIO.delete()` to delegate to a new `protected doDelete(Path, boolean)` template method. This allows subclasses (e.g. JindoFileIO) to intercept delete operations (trash/recycle-bin, audit logging, etc.) while retaining access to the real filesystem delete via `doDelete()`.

### Changes

- Extract filesystem delete logic into `protected doDelete(Path, boolean)`
- `delete()` now delegates to `doDelete()` — no behavior change for existing code
- Subclasses can override `delete()` with custom logic and call `doDelete()` for the actual delete

### Tests

- Unit tests verifying the template method pattern works correctly

### API and Format

### Documentation


XiaoHongbo-Hope:
+1

PR #7669, opened 2026-04-19
[deploy] Fix JDK 17 deploy script to use renamed Spark4 module artifactIds

XiaoHongbo-Hope:
### Purpose
paimon-spark4-common and paimon-spark-4.0 were renamed with _2.13 suffix in #6128, but the deploy script
   was not updated accordingly, causing JDK 17 deploy to fail with "Could not find the selected project in
   the reactor".


yuzelin (approved):
+1

PR #7670, opened 2026-04-19
[tantivy] Fix unnecessary .store file generation in full-text index

chenghuichen:
### Purpose
Tantivy is used purely as an inverted index in Paimon, so `.store` files (raw field values) are never needed or read. The original implementation mistakenly set `row_id` as `stored=True`, wasting 30% or even more of archive size per index file.

This PR removes `.set_stored()` from the schema and filters `.store` files when packing the archive, and updates the Python reader accordingly.

### Tests
Re-using `JavaPyTantivyE2ETest#testTantivyFullTextIndexWrite` +`test_read_tantivy_full_text_inde`

### DIFF
--- a/paimon-python/pypaimon/globalindex/tantivy/tantivy_full_text_global_index_reader.py
+++ b/paimon-python/pypaimon/globalindex/tantivy/tantivy_full_text_global_index_reader.py
@@ -153,26 +153,17 @@ class TantivyFullTextGlobalIndexReader(GlobalIndexReader):
         searcher = self._searcher
         query = self._index.parse_query(query_text, ["text"])
 
-        scored_results = searcher.search(query, limit)
-        if not scored_results.hits:
+        results = searcher.search(query, limit)
+        if not results.hits:
             return DictBasedScoredIndexResult({})
 
-        addr_to_score: Dict[tuple, float] = {
-            (addr.segment_ord, addr.doc): score
-            for score, addr in scored_results.hits
-        }
+        doc_addresses = [addr for score, addr in results.hits]
+        scores = [score for score, addr in results.hits]
+        row_ids = searcher.fast_field_values("row_id", doc_addresses)
 
-        # This can be replaced by https://github.com/quickwit-oss/tantivy-py/pull/641 when it's released
-        all_by_rowid = searcher.search(query, searcher.num_docs, order_by_field='row_id')
         id_to_scores: Dict[int, float] = {}
-        remaining = len(addr_to_score)
-        for row_id, addr in all_by_rowid.hits:
-            score = addr_to_score.get((addr.segment_ord, addr.doc))
-            if score is not None:
-                id_to_scores[row_id] = score
-                remaining -= 1
-                if remaining == 0:
-                    break
+        for row_id, score in zip(row_ids, scores):
+            id_to_scores[row_id] = score
 
         return DictBasedScoredIndexResult(id_to_scores)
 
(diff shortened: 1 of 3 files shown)


paimon-python/pypaimon/globalindex/tantivy/tantivy_full_text_global_index_reader.py
```diff
@@ -152,13 +152,27 @@ def visit_full_text_search(self, full_text_search) -> Optional[ScoredGlobalIndex
...
 
+        scored_results = searcher.search(query, limit)
+        if not scored_results.hits:
+            return DictBasedScoredIndexResult({})
+
+        addr_to_score: Dict[tuple, float] = {
+            (addr.segment_ord, addr.doc): score
+            for score, addr in scored_results.hits
+        }
+
```
jerry-024, line 164: This fallback looks like a potentially large performance regression for broad queries. We only need the `row_id` for the top-`limit` hits in `scored_results`, but this second search asks tantivy-py to collect up to `searcher.num_docs` matches ordered by `row_id`. For a common term on a large shard, a `limit=10` lookup can now degenerate into scanning/materializing almost the full match set just to recover 10 ids. Could we keep `row_id` stored until batch fast-field access is available in the shipped tantivy-py version, or add a direct fast-field read path instead?

chenghuichen: Thanks for the review! Based on feedback from the tantivy-py community, v0.26 with fast-field access should be released in a few weeks (see https://github.com/quickwit-oss/tantivy-py/pull/641). I think it makes sense to hold this PR for now — I'll track the community progress and ping you once the code is updated.

chenghuichen:
`tantivy-py` 0.26 is out. Updated to use `fast_field_values()`, the workaround is removed.

chenghuichen:
@JingsongLi The [support_directory](https://github.com/JingsongLi/tantivy-py/tree/support_directory) branch is currently 29 commits behind upstream 0.26.0, which means it's missing the `fast_field_values` API that our code depends on. Could we rebase it onto 0.26.0?

Alternatively we could wait for 0.27 and switch to the official release — upstream is going to replace the custom directory approach with BlobStore (https://github.com/quickwit-oss/tantivy-py/pull/632). But the timeline is uncertain, and if this code ships before that, backward compatibility becomes a headache.

JingsongLi (commented):
Review: Fix unnecessary .store file generation in full-text index

Good optimization. Since the JNI searchIndex path already reads row_id via fast fields, stored fields were pure waste on the Rust/Java side. Aligning the Python reader to also use fast fields is the right approach.

CORRECTNESS:

1. Rust schema change -- Removing .set_stored() from row_id and replacing TEXT with explicit TextOptions (indexing-only, no storage) is correct. The custom TextOptions preserves WithFreqsAndPositions and the default tokenizer, matching TEXT minus storage.

2. Python reader -- Switching from searcher.doc(doc_address) to searcher.fast_field_values is both correct and more efficient (batch columnar read vs per-doc stored-field deserialization). The early return on empty hits is a nice touch.

3. Backward compatibility -- Old indexes already have fast field data for row_id since fast=True was always set. The new reader only depends on fast fields, so it reads old indexes correctly. Orphaned .store files in existing archives are safely ignored.

ISSUES / QUESTIONS:

1. tantivy-py version requirement -- searcher.fast_field_values is not available in tantivy-py <= 0.22. Based on the discussion thread this depends on tantivy-py >= 0.26. Is there a version pin that should be bumped?

2. Return type assumption of fast_field_values -- The code assumes it returns an iterable of scalar values matching the input doc_addresses order. If the API returns list-of-lists (fast fields can be multivalued), this would silently produce wrong results. Please confirm the return contract or defensively index (e.g. row_ids[i][0]).

3. Rust TEXT import is now unused -- The diff adds IndexRecordOption, TextFieldIndexing, TextOptions but the original TEXT import should be removed. Please verify.

SUGGESTION: Add a comment in the Rust TextOptions construction explaining why storage is intentionally omitted for future contributors.

Overall a clean change. The 30%+ archive size reduction is significant.

JingsongLi:
+1

PR #7673, opened 2026-04-19
[python] Push down shard range to Vortex and unify Lance range parameter

chenghuichen:
### Purpose
Push down SlicedSplit shard range to Vortex format reader for native row selection, and unify the range-read parameter naming across Lance and Vortex.

Other formats (Parquet/ORC/Avro/Blob) fall back to ShardBatchReader as before.

### Tests
- paimon-python/pypaimon/tests/reader_append_only_test.py

### DIFF
--- a/paimon-python/pypaimon/read/reader/data_file_batch_reader.py
+++ b/paimon-python/pypaimon/read/reader/data_file_batch_reader.py
@@ -43,7 +43,8 @@ class DataFileBatchReader(RecordBatchReader):
                  system_fields: dict,
                  blob_as_descriptor: bool = False,
                  blob_descriptor_fields: Optional[set] = None,
-                 file_io: Optional[FileIO] = None):
+                 file_io: Optional[FileIO] = None,
+                 row_id_offsets: Optional[List[int]] = None):
         self.format_reader = format_reader
         self.index_mapping = index_mapping
         self.partition_info = partition_info
@@ -51,6 +52,8 @@ class DataFileBatchReader(RecordBatchReader):
         self.schema_map = {field.name: field for field in PyarrowFieldParser.from_paimon_schema(fields)}
         self.row_tracking_enabled = row_tracking_enabled
         self.first_row_id = first_row_id
+        self.row_id_offsets = row_id_offsets
+        self._row_id_cursor = 0
         self.max_sequence_number = max_sequence_number
         self.system_fields = system_fields
         self.blob_as_descriptor = blob_as_descriptor
@@ -209,9 +212,14 @@ class DataFileBatchReader(RecordBatchReader):
         # Handle _ROW_ID field
         if SpecialFields.ROW_ID.name in self.system_fields.keys():
             idx = self.system_fields[SpecialFields.ROW_ID.name]
-            # Create a new array that fills with computed row IDs
-            arrays[idx] = pa.array(range(self.first_row_id, self.first_row_id + record_batch.num_rows), type=pa.int64())
-            self.first_row_id += record_batch.num_rows
+            if self.row_id_offsets is not None:
+                end = self._row_id_cursor + record_batch.num_rows
+                row_ids = [self.first_row_id + o for o in self.row_id_offsets[self._row_id_cursor:end]]
+                arrays[idx] = pa.array(row_ids, type=pa.int64())
+                self._row_id_cursor = end
+            else:
+                arrays[idx] = pa.array(range(self.first_row_id, self.first_row_id + record_batch.num_rows), type=pa.int64())
+                self.first_row_id += record_batch.num_rows
 
         # Handle _SEQUENCE_NUMBER field
         if SpecialFields.SEQUENCE_NUMBER.name in self.system_fields.keys():
--- a/paimon-python/pypaimon/read/reader/data_file_batch_reader.py
+++ b/paimon-python/pypaimon/read/reader/data_file_batch_reader.py
@@ -218,7 +218,8 @@ class DataFileBatchReader(RecordBatchReader):
                 arrays[idx] = pa.array(row_ids, type=pa.int64())
                 self._row_id_cursor = end
             else:
-                arrays[idx] = pa.array(range(self.first_row_id, self.first_row_id + record_batch.num_rows), type=pa.int64())
+                row_id_range = range(self.first_row_id, self.first_row_id + record_batch.num_rows)
+                arrays[idx] = pa.array(row_id_range, type=pa.int64())
                 self.first_row_id += record_batch.num_rows
 
         # Handle _SEQUENCE_NUMBER field
--- a/paimon-python/pypaimon/read/reader/format_lance_reader.py
+++ b/paimon-python/pypaimon/read/reader/format_lance_reader.py
@@ -37,7 +37,6 @@ class FormatLanceReader(RecordBatchReader):
 
     def __init__(self, file_io: FileIO, file_path: str, read_fields: List[DataField],
                  push_down_predicate: Any, batch_size: int = 1024,
-                 row_range: Optional[Tuple[int, int]] = None,
                  row_indices: Optional[List[int]] = None,
                  shard_range: Optional[Tuple[int, int]] = None):
         """Initialize Lance reader."""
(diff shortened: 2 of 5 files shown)


leaves12138:
Thanks for the update. I reviewed the latest patch (`9e596a1f3337`). The overall direction makes sense: pushing `SlicedSplit` ranges into Lance/Vortex readers avoids reading the full file and then slicing in Python. However, I think there is still a correctness issue before merging.

**Blocking issue: native partial reads can produce wrong `_ROW_ID` values.**

For Vortex, `FormatVortexReader` now reads only the selected local offsets through `scan(..., indices=...)`. After that, the batch still goes through `DataFileBatchReader`, which assigns `_ROW_ID` as a continuous sequence starting from `file.first_row_id`:

```python
arrays[idx] = pa.array(range(self.first_row_id, self.first_row_id + record_batch.num_rows), type=pa.int64())
self.first_row_id += record_batch.num_rows
```

This was correct when Vortex read the full file and `ShardBatchReader` sliced afterwards, because row-id assignment happened before slicing. With the new native pushdown, a shard like `[200, 700)` will read only 500 rows, but `_ROW_ID` will be generated as `[file.first_row_id, file.first_row_id + 500)`, instead of `[file.first_row_id + 200, file.first_row_id + 700)`. So scans that project `_ROW_ID` together with `with_shard` / `with_slice` can return incorrect metadata.

The same class of issue can also affect Lance `read_range`, and it is even more subtle for `IndexedSplit` / `take_rows`, because the result rows may be sparse and cannot be represented by a single shifted `first_row_id`. For the sparse case, we probably need either a row-id vector aware path, or fall back to reading/filtering in a way that preserves row-position semantics when `_ROW_ID` is requested or deletion-vector position semantics depend on original file offsets.

Could you please add a regression test for this? For example, use a row-tracking-enabled Vortex/Lance table, write one file with many rows, run `with_shard(1, 3)` or `with_slice(200, 700)` with projection including `_ROW_ID`, and assert that the returned `_ROW_ID` values are shifted by the slice start. This test should fail with the current implementation.

A smaller follow-up concern: `vortex.array(range(shard_range[0], shard_range[1]))` materializes an index array for a contiguous shard. That may be expensive for large files. If Vortex does not expose a native range/slice scan API, this may be acceptable for now, but it would be good to document the trade-off or add a benchmark/guardrail.

I only ran syntax validation locally because this environment does not have the required Python dependencies (`pyarrow`, `pandas`, `lance`, `vortex`) and the Vortex tests require Python 3.11+. `compileall` passed for the modified reader files and `split_read.py`.

chenghuichen:
@leaves12138 Thanks for catching this! The `_ROW_ID` correctness issue makes total sense. 

Fixed both cases:
- SlicedSplit: offset `first_row_id` by `shard_range[0]` for correct `_ROW_ID`
- IndexedSplit: pass local offsets to `DataFileBatchReader` for per-row `_ROW_ID` assignment

Added e2e regression tests (with_shard + `_ROW_ID` projection, IndexedSplit + `_ROW_ID`). 

For the `vortex.array(range(...))` concern: benchmarked with 20M rows, native pushdown shows overall ~5% speedup over ShardBatchReader. 
[bench_shard_pushdown.py](https://github.com/user-attachments/files/27792184/bench_shard_pushdown.py)
The gain is modest since Vortex readers still fetch full files before selecting rows, but this path will benefit directly once fine-grained byte-range reads land.

leaves12138:
Thanks for the update. I re-checked the latest patch (06650b515329). The no-predicate cases look much better now: native Lance shard/range pushdown and IndexedSplit sparse _ROW_ID assignment are covered by the new tests, and I was able to run the Lance affected tests locally after installing pylance.

However, I think there is still a blocking correctness/runtime issue when native row slicing is combined with a file-level predicate and _ROW_ID projection.

The current flow is:

- split_read.py passes row_indices / shard_range into the Lance/Vortex format readers.
- FormatLanceReader applies read_range() / take_rows() first, and then applies the Arrow predicate on the sliced in-memory table.
- FormatVortexReader passes both expr and indices into scan().
- DataFileBatchReader then assigns _ROW_ID by assuming the returned rows are still the original contiguous slice, or the prefix of row_id_offsets.

That assumption no longer holds after predicate filtering. For example, with a Lance append-only file containing ids 0..999 and row tracking enabled, reading with projection [id, _ROW_ID] and predicate id >= 500 currently returns rows like (id=500, _ROW_ID=0), (id=501, _ROW_ID=1), ... instead of preserving the original row ids. When this is combined with with_shard(), another failure mode appears: if a shard is fully filtered out by the predicate, DataFileBatchReader._assign_row_tracking() builds an empty table and then calls table.to_batches()[0], which raises IndexError.

So this is not only a missing test case; it can return wrong _ROW_ID values or fail at runtime for valid scans.

Could you please fix this before merge? Possible directions:

1. If native row slicing and predicate filtering are both enabled and _ROW_ID is requested, avoid pushing the predicate into the format reader; assign _ROW_ID using the original row positions first, then apply the predicate after row tracking is available.
2. Or make the format reader return the selected local row offsets after predicate filtering, and let DataFileBatchReader assign _ROW_ID from that filtered offset vector.
3. Also make _assign_row_tracking() handle zero-row batches safely instead of indexing table.to_batches()[0].

Please also add regression tests for Lance/Vortex covering projection [id, _ROW_ID] + predicate + with_shard()/with_slice(), including both a non-empty filtered shard and an empty filtered shard.

JingsongLi (commented):
## Review: [python] Push down shard range to Vortex and unify Lance range parameter

Overall this is a well-structured change. The unification of `row_range` -> `shard_range` naming and the extension of native pushdown to Vortex are sensible. A few observations:

### Correctness concerns

**1. No bounds check on `_row_id_cursor` in `DataFileBatchReader._assign_row_tracking`**

In `data_file_batch_reader.py`, when `row_id_offsets` is not None, the code slices:
```python
row_ids = [self.first_row_id + o for o in self.row_id_offsets[self._row_id_cursor:end]]
```
If the format reader yields more rows than `len(row_id_offsets)` (e.g., due to a bug in the upstream reader or a mismatch between the computed indices and actual rows returned), this will silently produce an empty list rather than raising an error. Consider adding an assertion like:
```python
assert end <= len(self.row_id_offsets), "row_id_offsets exhausted"
```

**2. `row_ranges` and `shard_range` mutual exclusivity is not enforced**

`file_reader_supplier` accepts both `row_ranges` (from IndexedSplit) and `shard_range` (from SlicedSplit). These are semantically mutually exclusive since they originate from different split types, but there is no guard. If both were provided, `effective_first_row_id` would incorporate `shard_range[0]` while `row_id_offsets` (derived from `row_ranges`) are also passed to `DataFileBatchReader`, producing incorrect row IDs. A defensive assertion at the top of the method would prevent subtle bugs:
```python
assert not (row_ranges is not None and shard_range is not None), \
    "row_ranges and shard_range are mutually exclusive"
```

### Design observations

**3. Vortex shard via materialized index array**

The Vortex path materializes `range(shard_range[0], shard_range[1])` into a full array. For extremely large files (e.g., hundreds of millions of rows per shard), this may allocate significant memory. The trade-off comment is appreciated, but it might be worth adding a note about potential future improvement if Vortex adds a native slice/range scan API.

**4. `first_row_id` state after offset-based assignment**

When `row_id_offsets` is used, `self.first_row_id` is never advanced (unlike the else branch). This is correct for the current usage but makes the object's post-iteration state differ between the two paths. If `DataFileBatchReader` were ever reused or its `first_row_id` inspected after reading, this inconsistency could cause confusion. Not a bug today, but worth a brief comment explaining this intentional divergence.

### Positives

- The refactoring of `RawFileSplitRead.raw_reader_supplier` removes format-specific branching and delegates correctly to `file_reader_supplier`, which is a nice simplification.
- The `ShardBatchReader` fallback for non-native formats (Parquet/ORC/Avro/Blob) preserves backward compatibility cleanly.
- Test coverage is thorough: shard correctness tests for both Lance and Vortex, plus row ID correctness with IndexedSplit, are exactly the right tests for this change.
- The docstring-style comments distinguishing `row_indices` (discrete offsets from ANN) vs `shard_range` (contiguous range from parallel scan) improve readability.

JingsongLi:
+1

PR #7674, opened 2026-04-20
[WIP] [lumina] Reuse LuminaSearcher across queries via searcher pool

chenghuichen:

### DIFF
--- a/paimon-lumina/src/main/java/org/apache/paimon/lumina/index/LuminaSearcherPool.java
+++ b/paimon-lumina/src/main/java/org/apache/paimon/lumina/index/LuminaSearcherPool.java
@@ -1,0 +1,116 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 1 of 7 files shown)


JingsongLi (commented):
This PR is marked as [WIP]. A few observations:

1. **Searcher Pool design**: The `LuminaSearcherPool` concept is valuable for reusing expensive searcher instances across queries. Key questions:
   - What's the eviction strategy? LRU? Time-based?
   - Is the pool bounded? What happens when it fills up?
   - How does refresh work when the underlying index is updated (e.g., after compaction adds new segments)?

2. **Thread safety**: Vector index searchers may hold file handles or memory-mapped regions. Ensure the pool handles concurrent access correctly — particularly the case where a searcher is returned to the pool while still being used by another thread.

3. **Resource cleanup**: When the pool is closed, all pooled searchers must release their native resources (e.g., Lucene IndexReader handles, mmap regions). Is there a `close()` method on the pool?

4. **PR description is empty** ("Purpose: / Tests:") — please fill in the design rationale and test plan before requesting review.

Happy to look more closely once this is ready for review.

PR #7675, opened 2026-04-20
[common] handle tmp file when classifying file type

LsomeYeah:
### Purpose
`FileIO#tryToWriteAtomic` will write tmp file first and rename it to correct name, the type of tmp file may be classified incorrectly, this pr fix it.


JingsongLi:
+1

PR #7677, opened 2026-04-21
[docs] warn user the potential inaccuracy of global index.

steFaiz:
### Purpose
We should warn user the global-indexed queries may return inaccurate results.

### Tests
None


JingsongLi:
+1

PR #7679, opened 2026-04-21
[core] Optimize error message when creating PK table with row-tracking enabled 

XiaoHongbo-Hope:
### Purpose
When creating a PK table with `row-tracking.enabled=true` with setting bucket = -2,
  the error reports "Cannot define bucket for row tracking table, it only support
  bucket = -1" instead of "Cannot define primary-key for row tracking table". Which will make user confused. This PR optimize the error message.


JingsongLi:
+1

PR #7681, opened 2026-04-21
[core] Skip partition-specific table-default options for non-partitioned tables

XiaoHongbo-Hope:
### Purpose
When catalog-level table-default options include partition-specific options (e.g. table-default.partition.expiration-time), creating a non-partitioned table fails with "Can not set 'partition.expiration-time' for non-partitioned table".

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
+++ b/paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
@@ -204,16 +204,6 @@ public class SchemaValidation {
                             CoreOptions.STREAMING_READ_APPEND_OVERWRITE.key()));
         }
 
-        if (schema.partitionKeys().isEmpty()) {
-            for (String key : schema.options().keySet()) {
-                if (!CoreOptions.requiresPartitionedTable(key)) {
-                    continue;
-                }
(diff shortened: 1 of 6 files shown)


paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
```diff
@@ -204,10 +204,13 @@ public static void validateTableSchema(TableSchema schema) {
...
 
-        if (schema.options().containsKey(CoreOptions.PARTITION_EXPIRATION_TIME.key())) {
-            if (schema.partitionKeys().isEmpty()) {
+        if (schema.partitionKeys().isEmpty()) {
+            for (String key : schema.options().keySet()) {
+                if (!CoreOptions.requiresPartitionedTable(key)) {
+                    continue;
+                }
                 throw new IllegalArgumentException(
-                        "Can not set 'partition.expiration-time' for non-partitioned table.");
```
JingsongLi, line 210: Just remove this?

XiaoHongbo-Hope: > Just remove this?

Updated

JingsongLi:
+1

PR #7686, opened 2026-04-22
[hive] Add hive.skip-update-stats config option to control DO_NOT_UPDATE_STATS in HiveCatalog alter table

cxzl25:
### Purpose

Commit 989a433 introduced a fix to avoid excessive HMS memory usage during alter table by setting DO_NOT_UPDATE_STATS in the EnvironmentContext. However, commit 8688206 later rewrote HiveAlterTableUtils for Hive 1.x compatibility and hardcoded DO_NOT_UPDATE_STATS to "false", breaking the previous fix.

https://github.com/apache/paimon/pull/4549

### Tests
Fix HiveTableStatsTest, the originally implemented Test does not take effect.

### DIFF
--- a/docs/generated/hive_catalog_configuration.html
+++ b/docs/generated/hive_catalog_configuration.html
@@ -39,6 +39,12 @@ If not configured, try to load from 'HIVE_CONF_DIR' env.
 </td>
         </tr>
         <tr>
+            <td><h5>hive.skip-update-stats</h5></td>
+            <td style="word-wrap: break-word;">false</td>
+            <td>Boolean</td>
+            <td>If true, sets DO_NOT_UPDATE_STATS in the Hive EnvironmentContext when altering tables, preventing Hive from updating table statistics.</td>
+        </tr>
+        <tr>
             <td><h5>location-in-properties</h5></td>
             <td style="word-wrap: break-word;">false</td>
             <td>Boolean</td>
--- a/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveAlterTableUtils.java
+++ b/paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveAlterTableUtils.java
@@ -29,20 +29,22 @@ import org.apache.thrift.TException;
 /** Utils for hive alter table. */
 public class HiveAlterTableUtils {
 
-    public static void alterTable(IMetaStoreClient client, Identifier identifier, Table table)
+    public static void alterTable(
+            IMetaStoreClient client, Identifier identifier, Table table, boolean skipUpdateStats)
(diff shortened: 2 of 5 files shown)


cxzl25:
@GangYang-HX @JingsongLi  Can you help review this PR, thanks

huyuanfeng2018:
@cxzl25  Why can't this default value be directly true?

cxzl25:
> Why can't this default value be directly true?

The default value can be changed from false to true. In theory, true would be more reasonable. I just used the default value of the original PR.

JingsongLi (commented):
## Review of hive.skip-update-stats config option

Overall this is a reasonable fix for the regression described in the PR. A few observations:

### Correctness

1. **Logic in `alterTableWithEnv` is correct**: The `|| skipUpdateStats` ensures that either the table-level parameter or the catalog-level config can suppress stats updates. Good.

2. **Fallback path (`alterTableWithoutEnv`) does not use the flag**: When `NoClassDefFoundError` or `NoSuchMethodError` is caught (Hive 1.x compatibility path), `skipUpdateStats` has no effect because `EnvironmentContext` is unavailable. This is acceptable but worth documenting in the config description (e.g., "This option has no effect on Hive 1.x environments where EnvironmentContext is not supported").

3. **Test assertion may be fragile**: The new assertion `assertThat(table1.getParameters()).isEqualTo(table2.getParameters())` compares the entire parameter map before and after `alterTable`. In a real HMS (and possibly in embedded HMS), parameters like `transient_lastDdlTime` are typically updated on DDL operations. If the embedded HMS updates such metadata parameters during alter, this assertion would fail. The original assertion (`COLUMN_STATS_ACCURATE` is null) was more targeted. Consider either:
   - Keeping a targeted assertion on `COLUMN_STATS_ACCURATE`, or
   - Asserting that specific stats-related keys did not change, rather than comparing the full map.

### Design

4. **Default `false` is a safe choice** for backward compatibility, but it means users affected by the regression must explicitly opt in. Since the original commit 989a433 intended to always skip stats updates, you might consider whether `true` would be a better default to restore the original intended behavior. This is a product decision though.

5. **Single caller of `HiveAlterTableUtils.alterTable`** confirmed - the signature change is safe.

6. **Option naming** (`hive.skip-update-stats`) is clear and follows the existing `hive.*` prefix convention in the project.

### Minor

7. The test previously set `StatsSetupConst.DO_NOT_UPDATE_STATS` as a raw string option on `CatalogContext`, which would never have been picked up by the catalog as a typed `ConfigOption`. Good catch fixing this - the original test was effectively a no-op for validating the skip-stats behavior.

JingsongLi:
Please rebase master.

JingsongLi (approved):
+1

PR #7688, opened 2026-04-22
[iceberg] Fix NPE in IcebergRestMetadataCommitter when table has no snapshots

JTaky:
## Motivation

When committing Iceberg metadata via the REST catalog, `commitMetadataImpl` takes the `!withBase` code path if the base metadata check fails. 
In this path, a log statement calls `currentSnapshot().snapshotId()` on both the existing and new table metadata. However, `currentSnapshot()` returns null when a table exists but has no snapshots (e.g., a freshly created or empty table), causing a `NullPointerException` that fails the entire commit. Since it happens only for an empty iceberg table, we can manually delete the table and be unblocked. 

Meanwhile we should not have NPE in the log method.

```
Caused by: java.lang.NullPointerException
    at org.apache.paimon.iceberg.IcebergRestMetadataCommitter.commitMetadataImpl(IcebergRestMetadataCommitter.java:161)
```

## Implementation

Added null guards for `metadata.currentSnapshot()` and `newMetadata.currentSnapshot()` in the log statement, logging `"No snapshot"` when the snapshot is absent.


JingsongLi:
+1 Thanks @JTaky

PR #7689, opened 2026-04-23
[core] Support BlobRef in internal blob serializer

leaves12138:
## What
- support serializing both BlobData and BlobRef in BlobSerializer
- serialize BlobRef as descriptor bytes without materializing blob payload
- make internal binary blob write/read paths consistently use BlobSerializer
- reject BlobStream in BlobSerializer because it cannot be safely materialized here

## Why
BlobSerializer and internal binary writers previously called `blob.toData()` unconditionally. For BlobRef this forces loading the full blob payload, which defeats the lazy reference semantics and can be prohibitively expensive.

## Tests
- mvn -pl paimon-common -Pfast-build -Dtest=BlobSerializerTest,BinaryRowTest test

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/data/AbstractBinaryWriter.java
+++ b/paimon-common/src/main/java/org/apache/paimon/data/AbstractBinaryWriter.java
@@ -18,6 +18,7 @@
 
 package org.apache.paimon.data;
 
+import org.apache.paimon.data.serializer.BlobSerializer;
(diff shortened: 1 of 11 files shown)


JingsongLi (commented):
Why not using `isBlobDescriptor` to deserialize BlobDescriptor?

leaves12138:
> Why not using `isBlobDescriptor` to deserialize BlobDescriptor?

Reasonable

leaves12138:
Don't fix this until we meet it in reality

PR #7690, opened 2026-04-23
[core] Introduce 'compaction.small-file-ratio'

JingsongLi:
### Purpose

The ratio of target file size. Files whose size is smaller than target-file-size compaction.small-file-ratio will be picked for compaction rewriting. This avoids compacting the same file repeatedly due to compression inaccuracy causing output files to be slightly smaller than the target size.


yuzelin (approved):
+1

PR #7693, opened 2026-04-23
[test] Fix FormatTable directory creation in RESTFileSystemCatalog

Zouxxyy:
### Purpose

Fix format table tests when using REST catalog. Add `createFormatTable` override in `RESTFileSystemCatalog` to automatically create the table directory, and remove redundant manual `mkdirs` calls from `PaimonFormatTableTest`.

### Tests

Existing `PaimonFormatTableTest`.


JingsongLi:
+1

Issue #7694, opened 2026-04-24
[Feature] [Iceberg] Support writing metadata for multiple Iceberg tables


nickdelnano:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Motivation

Paimon's Iceberg compatibility supports a single Iceberg table in the catalog with `metadata.iceberg.database` and `metadata.iceberg.table` values. I'd like it to support the option of multiple db.table references so that I can register the Iceberg metadata to both a legacy db.table value and a new one. This will allow me to migrate my platform onto this feature without changing existing queries while also using my new db.table naming standard.

I don't have a view implementation available in my environment so I cannot use views to solve it.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #7696, opened 2026-04-25
[core] [iceberg] Support syncing Iceberg metadata to multiple Hive databases

nickdelnano:
Issue:  https://github.com/apache/paimon/issues/7694

> Paimon's Iceberg compatibility supports a single Iceberg table in the catalog with metadata.iceberg.database and metadata.iceberg.table values. I'd like it to support the option of multiple db.table references so that I can register the Iceberg metadata to both a legacy db.table value and a new one. This will allow me to migrate my platform onto this feature without changing existing queries while also using my new db.table naming standard.

> I don't have a view implementation available in my environment so I cannot use views to solve it.

Support registering Iceberg metadata in multiple Hive databases by allowing semicolon-delimited values in metadata.iceberg.database (e.g. 'db1;db2;db3').

## Tests
Unit and integration tests

### DIFF
--- a/docs/layouts/shortcodes/generated/iceberg_configuration.html
+++ b/docs/layouts/shortcodes/generated/iceberg_configuration.html
@@ -42,7 +42,7 @@ under the License.
             <td><h5>metadata.iceberg.database</h5></td>
             <td style="word-wrap: break-word;">(none)</td>
             <td>String</td>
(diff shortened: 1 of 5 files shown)


nickdelnano:
@JingsongLi can you review please

JingsongLi (approved):
+1

Issue #5820, opened 2025-06-30
[Feature] Introduce FilesCache to cache the manifest in local disk


Aitozi:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Motivation

In ad-hoc use case, the scan process will have to access the remote manifest in remote filesystem, which may be unstable time, we could support the cache based on local file to speed up the scan processing

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

Aitozi:
This should be covered by the block cache introduced in #7699

PR #7699, opened 2026-04-26
[java][python] Add block-level local disk cache for file reads

JingsongLi:
### Purpose

Introduce a CachingFileIO wrapper that transparently caches remote file reads at block granularity on local disk. Files are classified by FileType and only META, GLOBAL_INDEX, BUCKET_INDEX types are cached; DATA and FILE_INDEX are read directly.

See `file-cache.md`.

### DIFF
--- a/docs/content/program-api/file-cache.md
+++ b/docs/content/program-api/file-cache.md
@@ -123,7 +123,7 @@ table = catalog.get_table("db.my_table")
 
 ## Cache Lifecycle
 
-The cache is created and managed by the Catalog. All tables obtained from the same catalog share a single cache instance. When the catalog is closed, the cache is closed as well.
+The cache is created and managed by the Catalog. All tables obtained from the same catalog share a single cache instance. The cache lives as long as the Catalog object is reachable — no explicit close is needed.
 
 In distributed computing frameworks (Flink, Spark), the `FileIO` is serialized and shipped to task managers. After deserialization, the cache is **not** recreated — reads fall through directly to the remote storage. This is by design: the cache lifecycle is bound to the Catalog that created it, and a deserialized `FileIO` is no longer managed by any Catalog.
 
--- a/docs/content/program-api/file-cache.md
+++ b/docs/content/program-api/file-cache.md
@@ -1,0 +1,122 @@
+---
+title: "Local Cache"
+weight: 8
+type: docs
+aliases:
+- /program-api/file-cache.html
+- /pypaimon/file-cache.html
+---
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+# Local Cache
+
+When reading files from remote storage (S3, OSS, HDFS, etc.), each seek+read goes over the network. Paimon provides a block-level local cache that transparently caches file reads, significantly reducing remote I/O for repeated access patterns.
+
+The cache supports two modes:
+- **Disk cache**: when `local-cache.dir` is configured, blocks are cached on local disk.
+- **Memory cache**: when `local-cache.dir` is not configured, blocks are cached in memory.
+
+## Cached File Types
+
+The cache classifies files by type. By default, only `meta` and `global-index` types are cached. You can customize this via the `local-cache.whitelist` option.
+
+| File Type | Config Name | Examples | Default Cached |
+|-----------|-------------|----------|----------------|
+| META | meta | snapshot, schema, manifest, statistics, tag | Yes |
+| GLOBAL_INDEX | global-index | BTree, Lumina, Tantivy index files | Yes |
+| BUCKET_INDEX | bucket-index | Hash, deletion vector index files | No |
+| DATA | data | Data files (ORC, Parquet, etc.) | No |
+| FILE_INDEX | file-index | Data-file level bloom filter, bitmap | No |
+
+All file types can be added to the whitelist. The default whitelist is `meta,global-index`.
+
+## Enable Cache
+
+This is a catalog-level option. Configure it when creating the catalog:
+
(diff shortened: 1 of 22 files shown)


JingsongLi:
This configure option should be catalog-level, and it is better to name to local-disk.cache.

docs/content/program-api/file-cache.md
```diff
@@ -0,0 +1,112 @@
...
+
+The cache classifies files by type. By default, only `meta` and `global-index` types are cached. You can customize this via the `file-cache.whitelist` option.
+
+| File Type | Config Name | Examples | Default Cached |
+|-----------|-------------|----------|----------------|
+| META | meta | snapshot, schema, manifest, statistics, tag | Yes |
+| GLOBAL_INDEX | global-index | BTree, Lumina, Tantivy index files | Yes |
+| BUCKET_INDEX | bucket-index | Hash, deletion vector index files | No |
+| DATA | data | Data files (ORC, Parquet, etc.) | No |
+| FILE_INDEX | file-index | Data-file level bloom filter, bitmap | No |
```
Aitozi, line 46: Is this block cache suitable for manifests reading?

JingsongLi: You mean META?

Aitozi: Yes，sorry for missing it.

JingsongLi:
This PR is ready to be reviewed. @Aitozi Hope you can take a look if you have time.

paimon-common/src/main/java/org/apache/paimon/fs/cache/LocalDiskCacheManager.java
```diff
@@ -0,0 +1,287 @@
...
+            needEvict = maxSizeBytes < Long.MAX_VALUE && currentSize > maxSizeBytes;
+        }
+        if (needEvict) {
+            evict();
+        }
+    }
+
+    private void evict() {
+        List<CacheEntry> entries = new ArrayList<>();
+        File[] prefixDirs = cacheDir.listFiles();
```
Aitozi: Why aren't we maintaining the sizes/access time of these cache files in memory to avoid these frequent list operations?

paimon-common/src/main/java/org/apache/paimon/fs/cache/LocalDiskCacheManager.java
```diff
@@ -0,0 +1,287 @@
...
+            path.setLastModified(System.currentTimeMillis());
+            return data;
+        } catch (IOException e) {
+            LOG.debug("Failed to read cache block: {}", path, e);
+            return null;
+        }
+    }
+
+    public void putBlock(String filePath, int blockIndex, byte[] data) {
+        File path = cachePath(filePath, blockIndex);
```
Aitozi, line 96: Here, we are storing each block as an individual file. Will this lead to an excessive number of local files? AFAIK, engines like StarRocks handle this differently; they generally store serval blocks to a big single file.

Another issue to consider here is the access pattern of these files. For example, data files might have only certain blocks accessed due to column pruning, But regarding other files like the Manifest and Index, will we only access a portion of them? I'm not entirely sure.If we aren't accessing only a portion of the data, does it still make sense for us to break this down into blocks?

Internally, we adopted a different approach. For Meta files, we store a one-to-one mapping between the local file and the remote metadata file. https://github.com/apache/paimon/issues/5820

paimon-common/src/main/java/org/apache/paimon/fs/cache/CachingSeekableInputStream.java
```diff
@@ -0,0 +1,153 @@
...
+import org.apache.paimon.fs.FileIO;
+import org.apache.paimon.fs.Path;
+import org.apache.paimon.fs.SeekableInputStream;
+
+import javax.annotation.Nullable;
+
+import java.io.IOException;
+
+/** A {@link SeekableInputStream} that caches reads at block granularity on local disk. */
+public class CachingSeekableInputStream extends SeekableInputStream {
```
Aitozi, line 30: Has the performance improvement of CachingInputStream been measured yet?

Aitozi (commented):
Thanks @JingsongLi for the nice work, I left some comments.

paimon-common/src/main/java/org/apache/paimon/fs/cache/LocalDiskCacheManager.java
JingsongLi: First comment: 'store serval blocks to a big single file' is better, we should improve this in the next step.

Second comment: I don't think it's necessary to handle the manifest file separately. Universal block level caching is fully applicable to it, and its design has almost no impact on performance. (If the design is reasonable, there are still many areas that need improvement)

paimon-common/src/main/java/org/apache/paimon/fs/cache/CachingSeekableInputStream.java
JingsongLi: Only performance testing for object storage, the improvement is significant, but it will continue to be tested in the future.

paimon-common/src/main/java/org/apache/paimon/fs/cache/CachingFileIO.java
```diff
@@ -0,0 +1,195 @@
...
+        if (!options.get(CatalogOptions.LOCAL_CACHE_ENABLED)) {
+            return fileIO;
+        }
+
+        MemorySize maxSizeOpt = options.get(CatalogOptions.LOCAL_CACHE_MAX_SIZE);
+        long maxSize = maxSizeOpt == null ? Long.MAX_VALUE : maxSizeOpt.getBytes();
+        int blockSize = (int) options.get(CatalogOptions.LOCAL_CACHE_BLOCK_SIZE).getBytes();
+
+        String cacheDir = options.get(CatalogOptions.LOCAL_CACHE_DIR);
+        LocalCacheManager cache;
```
JingsongLi: cache may need to be managed by catalog.

JingsongLi: Fixed: Cache Lifecycle

The cache is created and managed by the Catalog. All tables obtained from the same catalog share a single cache instance. The cache lives as long as the Catalog object is reachable — no explicit close is needed.

In distributed computing frameworks (Flink, Spark), the `FileIO` is serialized and shipped to task managers. After deserialization, the cache is **not** recreated — reads fall through directly to the remote storage. This is by design: the cache lifecycle is bound to the Catalog that created it, and a deserialized `FileIO` is no longer managed by any Catalog.

If you need caching on task managers, create a new Catalog with cache options enabled on each worker node.

Aitozi (approved):
+1

paimon-common/src/main/java/org/apache/paimon/fs/cache/LocalDiskCacheManager.java
Aitozi: That sounds great, thanks for the explanation.

PR #7701, opened 2026-04-26
[python] Support value stats with truncate mode by default

XiaoHongbo-Hope:
### Purpose
Python-written append tables have no value stats in data files, preventing predicate pushdown from skipping irrelevant files during upsert-by-key reads. This PR enables default value stats for append table pruning. A follow-up PR will use these stats in the `upsert_by_key` lookup path.

 Skip us/ns/tz timestamps: _serialize_timestamp only supports ms precision (8-byte millis). Java's TIMESTAMP(4-9) uses a compound millis+nanos format that requires a different serialization path. Timezone is also not yet supported in serialization.

### DIFF
--- a/paimon-core/src/test/java/org/apache/paimon/JavaPyE2ETest.java
+++ b/paimon-core/src/test/java/org/apache/paimon/JavaPyE2ETest.java
@@ -466,17 +466,18 @@ public class JavaPyE2ETest {
             assertThat(table.rowType().getFieldTypes().get(5))
                     .isEqualTo(DataTypes.TIMESTAMP_WITH_LOCAL_TIME_ZONE());
             assertThat(table.rowType().getFieldTypes().get(6)).isEqualTo(DataTypes.TIME());
-            assertThat(table.rowType().getFieldTypes().get(7)).isInstanceOf(RowType.class);
-            RowType metadataType = (RowType) table.rowType().getFieldTypes().get(7);
+            assertThat(table.rowType().getFieldTypes().get(7)).isEqualTo(DataTypes.BYTES());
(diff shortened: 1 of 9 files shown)


JingsongLi (changes requested):
The binary column in Java does not generate min/max stats. If Java reads the manifest written in Python and pushes predicates down to the binary column, it may result in incorrect file skipping.

XiaoHongbo-Hope:
> The binary column in Java does not generate min/max stats. If Java reads the manifest written in Python and pushes predicates down to the binary column, it may result in incorrect file skipping.

Thanks, fixed

paimon-python/pypaimon/write/file_store_write.py
```diff
@@ -78,7 +78,8 @@ def max_seq_number():
                 partition=partition,
                 bucket=bucket,
                 max_seq_number=max_seq_number(),
-                options=options)
+                options=options,
+                write_cols=self.write_cols)
```
SteNicholas: The Java KeyValueDataFileWriter does not support `writeCols`, therefore is it necessary to pass `write_cols`?

XiaoHongbo-Hope: > The Java KeyValueDataFileWriter does not support `writeCols`, therefore is it necessary to pass `write_cols`?

Removed

JingsongLi (changes requested):
We should not enable this stats by default. Ideally, these statistical information should be obtained from the format.

XiaoHongbo-Hope:
> We should not enable this stats by default. Ideally, these statistical information should be obtained from the format.

Updated default config as none

PR #7703, opened 2026-04-26
[lance] Support timestamp_ltz by normalizing timezone strings for Arrow-Rust compatibility

### LINKED ISSUE #6648: [Feature] Paimon lance format should support timestamp_ltz
### Search before asking

### Motivation

Currently, Paimon lance format doesn't support timestamp_ltz,  when set timezone to `GMT-10:00`, it will throw `Unsupported timestamp type: timestamp:us:GMT-10:00`.

### Solution

The exception is thrown in https://github.com/lance-format/lance/blob/main/rust/lance-core/src/datatypes.rs#L357. 
The core reason is arrow-rust can't recognize `GMT-10:00`, it can only recognize IANA timezone like `America/New_York` or UTC offset format like `+07:30`. So users follow the two ways to set timezone, no exception will happen.    

### Anything else?

_No response_

0dunay0:
## Summary

Fixes #6648.

When using Lance format with `TIMESTAMP WITH LOCAL TIME ZONE`, Lance's Rust-side schema parser fails on timezone strings like `GMT-10:00` with `Unsupported timestamp type: timestamp:us:GMT-10:00`. This happens because Arrow-Rust only accepts IANA timezone names (e.g. `America/New_York`) or standard UTC offset format (e.g. `+05:30`, `-10:00`, `Z`), but Java's `ZoneId.systemDefault().toString()` produces the `GMT-10:00` prefix form.

The fix uses `ZoneId.normalized()` before converting to string. This converts `GMT-10:00` to `-10:00`, `GMT+05:30` to `+05:30`, `UTC` to `Z`, and leaves IANA names like `America/New_York` unchanged. All of these are formats that Arrow-Rust accepts.

Changes:
- In `ArrowFieldTypeConversion`, call `.normalized()` on the ZoneId before `.toString()` when creating Arrow Timestamp types
- In `LanceFileFormat`, remove the `UnsupportedOperationException` that blocked `LOCAL_ZONED_TIMESTAMP` entirely
- Add `timestamp_ltz` coverage to `LanceFileFormatTest` (validation) and `LanceFileFormatReadWriteTest` (round-trip read/write)

### DIFF
--- a/paimon-arrow/src/main/java/org/apache/paimon/arrow/ArrowFieldTypeConversion.java
+++ b/paimon-arrow/src/main/java/org/apache/paimon/arrow/ArrowFieldTypeConversion.java
@@ -149,7 +149,8 @@ public class ArrowFieldTypeConversion {
             int precision = localZonedTimestampType.getPrecision();
             TimeUnit timeUnit = getTimeUnit(precision);
             ArrowType arrowType =
-                    new ArrowType.Timestamp(timeUnit, ZoneId.systemDefault().toString());
+                    new ArrowType.Timestamp(
+                            timeUnit, ZoneId.systemDefault().normalized().toString());
             return new FieldType(localZonedTimestampType.isNullable(), arrowType, null);
         }
 
--- a/paimon-lance/src/main/java/org/apache/paimon/format/lance/LanceFileFormat.java
+++ b/paimon-lance/src/main/java/org/apache/paimon/format/lance/LanceFileFormat.java
@@ -169,13 +169,7 @@ public class LanceFileFormat extends FileFormat {
 
         @Override
         public Void visit(LocalZonedTimestampType localZonedTimestampType) {
-            // lance has a bug here, if the local in GMT-10:00, it failed by reson: called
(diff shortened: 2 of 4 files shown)


0dunay0:
Closing this PR. While investigating, I found that the root cause of #6648 goes deeper than the timezone string format.

Lance v0.39.0 has a bug in its type serialization. It serializes timestamp types as colon-delimited strings like `timestamp:us:+05:00`, then deserializes by splitting on `:`. Any timezone containing a colon (which includes all UTC offset formats like `+05:00`, `-10:00`, `+05:30`) produces too many segments and causes a Rust panic/SIGABRT. This bug is still present on Lance's main branch.

IANA timezone names like `America/New_York` work fine since they don't contain colons, but there's no colon-free format that covers all timezone offsets. So timestamp_ltz can't be fully supported in Lance until this is fixed upstream.

I'll open a separate PR for the `ArrowFieldTypeConversion.normalized()` improvement, which benefits other Arrow consumers but is unrelated to #6648.

PR #7705, opened 2026-04-27
[arrow] Normalize timezone strings for arrow-rust compatibility

### LINKED ISSUE #7704: [Bug] ArrowFieldTypeConversion produces GMT-prefix timezone strings that arrow-rust cannot parse
### Search before asking

### Paimon version

master (1.5-SNAPSHOT)

### Compute Engine

Any (affects paimon-arrow module used by Spark, Flink, etc.)

### Minimal reproduce step

Set JVM timezone to a GMT-prefix offset (e.g. `-Duser.timezone=GMT+5:00`), then use any code path that converts `LocalZonedTimestampType` through `ArrowFieldTypeConversion.ArrowFieldTypeVisitor`.

`ZoneId.systemDefault().toString()` produces `GMT+05:00`, which arrow-rust rejects because it doesn't recognize the `GMT` prefix.

This is reproducible in CI when the JVM timezone is randomized to a GMT-prefix offset.

`ArrowFieldTypeConversion` should produce timezone strings that arrow-rust can parse. Using `ZoneId.systemDefault().normalized().toString()` converts GMT-prefix offsets to the standard UTC offset format (e.g. `+05:00`), GMT/UTC to `Z`, and leaves IANA names like `America/New_York` unchanged. Arrow-rust accepts all of these.

### Anything else?

_No response_

0dunay0:
## Summary

Fixes #7704.

`ZoneId.systemDefault().toString()` can produce GMT-prefix timezone strings like `GMT-10:00` when the JVM timezone is set to a GMT-prefix offset. Arrow-rust doesn't recognize the `GMT` prefix and rejects these strings.

`normalized()` converts them to formats arrow-rust accepts:
- `GMT-10:00` becomes `-10:00`
- `GMT+05:30` becomes `+05:30`
- `UTC` / `GMT` becomes `Z`
- IANA names like `America/New_York` pass through unchanged

This was discovered while investigating #6648. The CI randomizes `-Duser.timezone` and hit `GMT+5:00`, which triggered a crash in the Lance native writer.

### DIFF
--- a/paimon-arrow/src/main/java/org/apache/paimon/arrow/ArrowFieldTypeConversion.java
+++ b/paimon-arrow/src/main/java/org/apache/paimon/arrow/ArrowFieldTypeConversion.java
@@ -149,7 +149,8 @@ public class ArrowFieldTypeConversion {
             int precision = localZonedTimestampType.getPrecision();
             TimeUnit timeUnit = getTimeUnit(precision);
             ArrowType arrowType =
-                    new ArrowType.Timestamp(timeUnit, ZoneId.systemDefault().toString());
+                    new ArrowType.Timestamp(
+                            timeUnit, ZoneId.systemDefault().normalized().toString());
             return new FieldType(localZonedTimestampType.isNullable(), arrowType, null);
         }
 


0dunay0:
@plusplusjiajia @JingsongLi Please have a look

JingsongLi (approved):
+1

PR #7706, opened 2026-04-27
[core] Check next snapshot's time for snapshotTimeRetain expiration protection

JingsongLi:
### Purpose

Previously, a snapshot was expired if its own timestamp exceeded snapshotTimeRetain. This could cause premature expiration when snapshotRetainMin is small. Now we check the next snapshot's timestamp instead, ensuring each snapshot is protected for the full snapshotTimeRetain duration.

### Tests

ExpireSnapshotsTest.java

### DIFF
--- a/paimon-core/src/test/java/org/apache/paimon/operation/ExpireSnapshotsTest.java
+++ b/paimon-core/src/test/java/org/apache/paimon/operation/ExpireSnapshotsTest.java
@@ -538,13 +538,63 @@ public class ExpireSnapshotsTest {
         expire.config(builder.snapshotTimeRetain(Duration.ofMillis(1000)).build()).expire();
 
         int latestSnapshotId = requireNonNull(snapshotManager.latestSnapshotId()).intValue();
-        for (int i = 1; i <= latestSnapshotId; i++) {
-            if (snapshotManager.snapshotExists(i)) {
-                assertThat(snapshotManager.snapshot(i).timeMillis())
-                        .isBetween(expireMillis - 1000, expireMillis);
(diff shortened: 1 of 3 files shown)


paimon-core/src/test/java/org/apache/paimon/operation/ExpireSnapshotsTest.java
```diff
@@ -740,6 +799,34 @@ private TestFileStore createStore() {
...
+                        old.commitKind(),
+                        newTimeMillis,
+                        old.totalRecordCount(),
+                        old.deltaRecordCount(),
+                        old.changelogRecordCount(),
+                        old.watermark(),
+                        old.statistics(),
+                        old.properties(),
+                        old.nextRowId());
+        fileIO.overwriteFileUtf8(snapshotManager.snapshotPath(snapshotId), updated.toJson());
```
tsreaper: It is likely that `Snapshot` constructor will have more arguments in the future. For testing purpose, we can just rewrite the json string, so we won't need to update this call in the future.

```suggestion
        String oldJson = fileIO.readFileUtf8(snapshotManager.snapshotPath(snapshotId));
        ObjectNode node = (ObjectNode) JsonSerdeUtil.OBJECT_MAPPER_INSTANCE.readTree(oldJson);
        node.put(Snapshot.FIELD_TIME_MILLIS, newTimeMillis);
        String newJson = JsonSerdeUtil.OBJECT_MAPPER_INSTANCE.writeValueAsString(node);
        fileIO.overwriteFileUtf8(snapshotManager.snapshotPath(snapshotId), newJson);
```

tsreaper (approved):

PR #7707, opened 2026-04-27
[lumina] Rename index identifier `lumina-vector-ann` to `lumina

jerry-024:
### Purpose

  Rename the Lumina vector global index identifier from `lumina-vector-ann` to
  `lumina` for a cleaner, shorter name. Existing tables created with the old
  identifier continue to work via a `LegacyLuminaVectorGlobalIndexerFactory`
  registered through SPI, so no on-disk migration is required. File naming on
  disk is unchanged (the writer has always used the `lumina-` prefix).

  ### Tests

  - `LuminaVectorGlobalIndexerFactoryTest` — verifies both identifiers route to
    the correct factory class.
  - `IndexFileHandlerTest#testExistsAndDeleteIndexFile` — parameterized over both
    `lumina` and `lumina-vector-ann`.
  - `FileTypeTest#testGlobalIndexFiles` — adds the new `lumina-global-index-*`
    path to the classification check.
  - `LuminaVectorIndexTest` (Spark) — adds a "create lumina vector index - legacy
    index type" case using `lumina-vector-ann`.
  - `vector_search_filter_test.VectorReaderFactoryTest` — Python reader accepts
    both identifiers.

### DIFF
--- a/docs/content/append-table/global-index.md
+++ b/docs/content/append-table/global-index.md
@@ -97,10 +97,12 @@ Generation) applications.
 CALL sys.create_global_index(
     table => 'db.my_table',
     index_column => 'embedding',
-    index_type => 'lumina-vector-ann',
+    index_type => 'lumina',
     options => 'lumina.index.dimension=128'
 );
 ```
+
+The legacy index type `lumina-vector-ann` is still accepted for existing tables and SQL compatibility.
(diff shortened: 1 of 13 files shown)


JingsongLi (approved):
+1

PR #7709, opened 2026-04-27
[core] Skip tryUpgrade for pkClusteringOverride since sort order differs from pk order

JingsongLi:
### Purpose

The sorting rule is not Level 0, so it cannot be directly upgraded


LsomeYeah (approved):
+1

PR #7711, opened 2026-04-27
[python] Support FileScanner partition_predicate to fix FileStoreCommit overwrite for sanity check

### LINKED ISSUE #7708: [Bug] The `overwrite()` of `file_store_commit.py` should use full table fields for `FullStartingScanner`'s `trim_and_transform_predicate`
### Search before asking

### Paimon version

The `overwrite()` of `file_store_commit.py` should use full table fields for `FullStartingScanner`'s `trim_and_transform_predicate`, which aligin with `drop_partitions`. Therefore, `FullStartingScanner`'s `trim_and_transform_predicate` could map indices correctly (full schema index -> partition index).

### Compute Engine

PythonAPI

### Minimal reproduce step

Insert static partition commit could reproduce the issue.

The `overwrite()` of `file_store_commit.py` mismatches index space for static partition.

### Anything else?

_No response_

SteNicholas:
### Purpose

Add `partition_predicate` support to `FileScanner`, matching the Java and Rust implementations:

- Previously, callers like `FileStoreCommit.overwrite()`, `drop_partitions()`, and `CommitScanner` had to build full-schema-indexed predicates (`PredicateBuilder(table.fields)`) that `FileScanner` would internally re-trim to partition keys via `trim_and_transform_predicate`.
- Now `FileScanner` accepts an optional `partition_predicate` parameter indexed directly by partition keys. When provided, it is used as `partition_key_predicate` without any index translation. The existing predicate parameter is preserved for the read path. All write-path callers are simplified to use `PredicateBuilder(partition_keys_fields)` and pass the result as `partition_predicate`.

Close #7708.
                                             
### Tests

- `TestFileScannerPartitionPredicate`: direct use, fallback derivation, no-filter pass-through, manifest file/entry filtering.
- `TestOverwritePartitionPredicate`: partition mismatch rejection, partition-scoped predicate passing, `drop_partitions` OR-predicate.
- `TestCommitScannerPartitionPredicate`: partition-key-indexed predicate construction, unpartitioned table returns None, keyword argument forwarding to FileScanner.

### DIFF
--- a/paimon-python/pypaimon/write/file_store_commit.py
+++ b/paimon-python/pypaimon/write/file_store_commit.py
@@ -204,9 +204,7 @@ class FileStoreCommit:
                         f"Partition keys are: {list(self.table.partition_keys)}."
                     )
 
-        # Use full table fields so FullStartingScanner's trim_and_transform_predicate
-        # maps indices correctly (full schema index -> partition index).
-        predicate_builder = PredicateBuilder(self.table.fields)
+        predicate_builder = PredicateBuilder(self.table.partition_keys_fields)
         partition_predicates = []
         for part in partitions:
             sub_predicates = [
@@ -524,7 +522,7 @@ class FileStoreCommit:
         """Generate commit entries for OVERWRITE mode based on latest snapshot."""
         entries = []
         current_entries = [] if latest_snapshot is None \
-            else (FileScanner(self.table, lambda: ([], None), partition_filter).
+            else (FileScanner(self.table, lambda: ([], None), partition_predicate=partition_filter).
                   read_manifest_entries(self.manifest_list_manager.read_all(latest_snapshot)))
         for entry in current_entries:
             entry.kind = 1  # DELETE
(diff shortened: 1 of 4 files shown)


chenghuichen:
LGTM.

paimon-python/pypaimon/write/file_store_commit.py
```diff
@@ -170,15 +170,23 @@ def overwrite(self, overwrite_partition, commit_messages: List[CommitMessage], c
         partition_filter = None
         # sanity check, all changes must be done within the given partition, meanwhile build a partition filter
         if len(overwrite_partition) > 0:
-            predicate_builder = PredicateBuilder(self.table.partition_keys_fields)
-            sub_predicates = []
+            # Build two predicates from overwrite_partition in one pass:
+            # - check_filter: indexed by partition keys only, used for the sanity check
+            #   below because msg.partition only contains partition key values.
+            # - partition_filter: indexed by full table schema, used by FileScanner's
```
JingsongLi: Can we just let `FileScanner` supports partition_predicate just like Java and Rust? In the future, if there are too many partitions, we can also carry out targeted optimization.

XiaoHongbo-Hope:
+1

SteNicholas:
Ping @JingsongLi. Addressed above comment.

JingsongLi (approved):
+1

PR #7713, opened 2026-04-27
[spark] supports reading blob descriptors from different FS with paimon

steFaiz:
### Purpose
This PR is a mirror of [flink's version](https://github.com/apache/incubator-paimon/pull/7231)

### Tests
See `org.apache.paimon.spark.sql.BlobTestBase`.


JingsongLi (approved):
+1

PR #7714, opened 2026-04-27
[lance] Fix zero-column read error in data evolution with BTree global index

XiaoHongbo-Hope:
### Purpose

Building BTree global index on Lance tables with data evolution enabled throws `java.lang.IllegalArgumentException: Invalid user input: Attempt to read zero columns from the file, at least one column must be specified`. This PR fixes the above zero-column read error.
### Tests
LanceBTreeGlobalIndexTest


JingsongLi:
+1

PR #7715, opened 2026-04-27
[core] Fix orphan files clean deleting data files during concurrent snapshot expiration

### LINKED ISSUE #7710: [Bug] remove_orphan_files may accidentally delete data in multitasking concurrent scenarios
### Search before asking

### Paimon version

version 1.2

### Compute Engine

Spark 3.5.6

### Minimal reproduce step

table max retained snapshot is 1
'snapshot.num-retained.max' = '1'

2026-04-25 04:37:12,116, Task 1 completed sys.compact operation and submitted snapshot 4959
2026-04-25 04:37:13,195, Task 1 completed snapshot expiration. Mark snapshot 4958 as expired and delete it
2026-04-25 04:37:37,605, Task 2 completed writing and submitted snapshot 4960
2026-04-25 04:37:38,185, Task 2 completed snapshot expiration. Mark snapshot 4959 as expired and delete it
At the same time, Task 1 started executing the remove_orphan_files, but the current snapshot 4959 was not read because 4959 had already been deleted by Task 2, resulting in usedFiles being empty, which in turn caused deleted=candidates, meaning that all files written one day ago were marked as files to be deleted and ultimately deleted.

Incorrect deletion of historical data

### Anything else?

_No response_

heye1005:
### Purpose

Fix #7710. 

`LocalOrphanFilesClean.clean()` collects used files by reading all snapshots via `safelyGetAllSnapshots()`. However, this method silently catches `FileNotFoundException` — if concurrent snapshot expiration deletes all snapshots between listing and reading, `usedFiles` ends up empty, and all candidate data files get deleted as orphans.

This patch skips orphan file deletion when `usedFiles` is empty to prevent accidental data loss.

### Tests

Added `LocalOrphanFilesCleanTest#testSkipCleaningWhenAllSnapshotsDeleted`.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/LocalOrphanFilesClean.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/LocalOrphanFilesClean.java
@@ -47,6 +47,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Consumer;
 import java.util.function.Function;
@@ -108,11 +109,42 @@ public class LocalOrphanFilesClean extends OrphanFilesClean {
         }
         candidateDeletes = new HashSet<>(candidates.keySet());
 
+        AtomicLong missingManifestListCount = new AtomicLong(0);
+        AtomicLong missingManifestCount = new AtomicLong(0);
+
         // find used files
         Set<String> usedFiles =
                 branches.stream()
-                        .flatMap(branch -> getUsedFiles(branch).stream())
+                        .flatMap(
+                                branch ->
+                                        getUsedFiles(
+                                                        branch,
+                                                        missingManifestListCount,
+                                                        missingManifestCount)
+                                                .stream())
                         .collect(Collectors.toSet());
+
+        if (usedFiles.isEmpty()) {
+            LOG.warn(
+                    "Collected used files is empty while {} candidates are present, "
+                            + "aborting orphan files clean to prevent data loss.",
+                    candidateDeletes.size());
+            return new CleanOrphanFilesResult(
+                    deleteFiles.size(), deletedFilesLenInBytes.get(), deleteFiles);
+        }
+
+        if (missingManifestListCount.get() > 0 || missingManifestCount.get() > 0) {
+            LOG.warn(
+                    "Detected {} missing manifest-list/index-manifest and {} missing manifest "
+                            + "file(s) during used-files collection while {} candidates are "
+                            + "present; this indicates a concurrent snapshot expiration. "
+                            + "Aborting orphan files clean to prevent data loss.",
+                    missingManifestListCount.get(),
+                    missingManifestCount.get(),
+                    candidateDeletes.size());
+            return new CleanOrphanFilesResult(
+                    deleteFiles.size(), deletedFilesLenInBytes.get(), deleteFiles);
+        }
 
         // delete unused files
         candidateDeletes.removeAll(usedFiles);
@@ -157,14 +189,21 @@ public class LocalOrphanFilesClean extends OrphanFilesClean {
     }
 
     private void collectWithoutDataFile(
-            String branch, Consumer<String> usedFileConsumer, Consumer<String> manifestConsumer)
+            String branch,
+            Consumer<String> usedFileConsumer,
+            Consumer<String> manifestConsumer,
+            Consumer<String> missingManifestListConsumer)
             throws IOException {
         randomlyOnlyExecute(
                 executor,
                 snapshot -> {
                     try {
                         collectWithoutDataFile(
-                                branch, snapshot, usedFileConsumer, manifestConsumer);
+                                branch,
+                                snapshot,
+                                usedFileConsumer,
+                                manifestConsumer,
+                                missingManifestListConsumer);
                     } catch (IOException e) {
                         throw new RuntimeException(e);
                     }
@@ -172,20 +211,29 @@ public class LocalOrphanFilesClean extends OrphanFilesClean {
                 safelyGetAllSnapshots(branch));
     }
 
-    private Set<String> getUsedFiles(String branch) {
+    private Set<String> getUsedFiles(
+            String branch,
+            AtomicLong missingManifestListCount,
+            AtomicLong missingManifestCount) {
         Set<String> usedFiles = ConcurrentHashMap.newKeySet();
         ManifestFile manifestFile =
                 table.switchToBranch(branch).store().manifestFileFactory().create();
         try {
             Set<String> manifests = ConcurrentHashMap.newKeySet();
-            collectWithoutDataFile(branch, usedFiles::add, manifests::add);
+            collectWithoutDataFile(
+                    branch,
+                    usedFiles::add,
+                    manifests::add,
+                    name -> missingManifestListCount.incrementAndGet());
             randomlyOnlyExecute(
                     executor,
                     manifestName -> {
                         try {
+                            AtomicBoolean manifestMissing = new AtomicBoolean(false);
                             retryReadingFiles(
                                             () -> manifestFile.readWithIOException(manifestName),
-                                            Collections.<ManifestEntry>emptyList())
+                                            Collections.<ManifestEntry>emptyList(),
+                                            manifestMissing)
                                     .stream()
                                     .map(ManifestEntry::file)
                                     .forEach(
@@ -197,6 +245,9 @@ public class LocalOrphanFilesClean extends OrphanFilesClean {
                                                         .filter(candidateDeletes::contains)
                                                         .forEach(usedFiles::add);
                                             });
+                            if (manifestMissing.get()) {
+                                missingManifestCount.incrementAndGet();
+                            }
                         } catch (IOException e) {
                             throw new RuntimeException(e);
                         }
--- a/paimon-core/src/main/java/org/apache/paimon/operation/LocalOrphanFilesClean.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/LocalOrphanFilesClean.java
@@ -109,39 +109,22 @@ public class LocalOrphanFilesClean extends OrphanFilesClean {
         }
         candidateDeletes = new HashSet<>(candidates.keySet());
 
-        AtomicLong missingManifestListCount = new AtomicLong(0);
-        AtomicLong missingManifestCount = new AtomicLong(0);
+        AtomicBoolean missingManifest = new AtomicBoolean(false);
 
         // find used files
         Set<String> usedFiles =
                 branches.stream()
-                        .flatMap(
-                                branch ->
-                                        getUsedFiles(
-                                                        branch,
-                                                        missingManifestListCount,
-                                                        missingManifestCount)
-                                                .stream())
+                        .flatMap(branch -> getUsedFiles(branch, missingManifest).stream())
                         .collect(Collectors.toSet());
 
         if (usedFiles.isEmpty()) {
-            LOG.warn(
-                    "Collected used files is empty while {} candidates are present, "
-                            + "aborting orphan files clean to prevent data loss.",
-                    candidateDeletes.size());
+            LOG.warn("Collected used files is empty, aborting orphan files clean.");
             return new CleanOrphanFilesResult(
                     deleteFiles.size(), deletedFilesLenInBytes.get(), deleteFiles);
         }
 
-        if (missingManifestListCount.get() > 0 || missingManifestCount.get() > 0) {
-            LOG.warn(
-                    "Detected {} missing manifest-list/index-manifest and {} missing manifest "
-                            + "file(s) during used-files collection while {} candidates are "
-                            + "present; this indicates a concurrent snapshot expiration. "
-                            + "Aborting orphan files clean to prevent data loss.",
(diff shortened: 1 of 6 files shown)


JingsongLi:
Hi @heye1005  It is better to fix it reading full ref files for latest snapshot.

heye1005:
> Hi @heye1005 It is better to fix it reading full ref files for latest snapshot.


Thanks for the review! Updated the PR based on your suggestion.

The root cause here is that `safelyGetAllSnapshots()` lists the snapshot directory first, then reads each file. If expiration deletes the snapshot file (along with its manifest-list) between these two steps, the read silently fails and returns an empty list — all old data files then get treated as orphans.

This PR has two changes:

1. Replaced the list-then-read in `safelyGetAllSnapshots()` with `latestSnapshot()` + ID range traversal. Expiration never deletes the latest snapshot, so the snapshot file read is safe. But the manifest-list of that snapshot can still be deleted before we read it.

2. Added `collectLatestSnapshotFiles()` — re-reads the latest snapshot right before deletion and removes its referenced files from candidates. By then expiration is usually done, so the current latest snapshot's manifest-list is intact. This is applied to Local, Flink, and Spark subclasses.

This doesn't 100% eliminate the race — if expiration happens to delete the latest snapshot's manifest-list during the pre-deletion check, it would still fail. But that window is just milliseconds, so practically very unlikely.

To fully solve this, I think we'd need either some form of locking between orphan clean and expiration, or delay manifest-list deletion (similar to how orphan clean already requires data files to be older than `olderThanMillis`). Happy to discuss further.

JingsongLi (commented):
## Review: [core] Fix orphan files clean deleting data files during concurrent snapshot expiration

### Overall Assessment

The defense-in-depth approach (modifying `safelyGetAllSnapshots` + adding `collectLatestSnapshotFiles` as a second barrier) is a reasonable strategy for narrowing the race window.

### Correctness Concerns

**1. Residual race still exists without a hard guard**

The TOCTOU window is narrowed but not eliminated. Consider adding an explicit empty-set guard: if `usedFiles` is empty after collecting from all branches, abort with a warning rather than proceeding to delete. This is simple, cheap, and guarantees no data loss regardless of timing.

**2. Broad exception swallowing in `safelyGetAllSnapshots`**

The outer `catch (Exception e)` catches all Exception types. If `latestSnapshot()` throws due to corruption, permissions, or network timeout, it silently proceeds with an empty snapshot set. Consider catching only `FileNotFoundException` and letting other exceptions propagate.

**3. `collectLatestSnapshotFiles` failure path**

If it fails for a branch (manifest file deleted by expiration), the catch-all means that branch's latest snapshot files are NOT protected.

### Performance

**4. Redundant manifest reading** -- `collectLatestSnapshotFiles` re-reads the latest snapshot's entire manifest tree. For large tables this is not trivial.

### Minor

**5.** The test verifies sequential (expire then clean) but not true concurrent race. A clarifying comment would help.

### Summary

Clear improvement. Main recommendation: add an explicit empty-set guard (`if (usedFiles.isEmpty()) abort`) as an absolute safety net.

heye1005:
Thanks for thorough review, @JingsongLi.
Fix by your feedback. only retain add an explicit empty-set guard, and tests adjusted. PTAL.

paimon-core/src/main/java/org/apache/paimon/operation/LocalOrphanFilesClean.java
```diff
@@ -114,6 +114,15 @@ public CleanOrphanFilesResult clean()
                         .flatMap(branch -> getUsedFiles(branch).stream())
                         .collect(Collectors.toSet());
 
+        if (usedFiles.isEmpty()) {
```
JingsongLi: This guard still does not protect the race described in the PR. `collectWithoutDataFileWithManifestFlag` adds the snapshot manifest-list names to `usedFiles` before reading the manifest-list contents, and `retryReadingFiles(..., emptyList())` returns an empty list when those manifest-lists are concurrently removed. In that case `usedFiles` is not empty, so this check is bypassed, but none of the data files referenced by the expired manifests are protected and `candidateDeletes.removeAll(usedFiles)` can still delete them. Can we guard the data-file collection result instead, or treat a missing manifest-list as an abort for the branch/table rather than only checking whether the overall used-file set is empty?

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/procedure/SparkOrphanFilesClean.scala
```diff
@@ -88,6 +88,18 @@ case class SparkOrphanFilesClean(
       .toDS()
       .cache()
 
+    if (usedManifestFiles.isEmpty) {
```
JingsongLi: Same issue here: this only checks whether any snapshot/manifest-list names were collected, not whether the referenced data files were protected. `collectWithoutDataFileWithManifestFlag` emits the manifest-list names before reading their contents; if concurrent expiration removes those manifest-lists, `retryReadingFiles(..., emptyList())` yields no manifest entries, but `usedManifestFiles` is still non-empty. The job then builds an empty `dataFiles` dataset and anti-joins candidates only against the manifest-list names, so old data files can still be deleted. Please make the abort condition cover the data-file collection failure/missing manifest-list case as well.

heye1005:
Thanks for the review, @lijingsong! 
I fixed it to only early-abort when a manifest / manifest-list is missing for a live snapshot on the main branch — missing manifests on tags and other branches are ignored as before.
I also updated an existing test whose assumption conflicted with this new abort behavior.

JingsongLi:
The current `early-abort` behavior, which lacks a manifest, only protects `main`: `LocalOrphanFilesClean` constructs `liveSnapshots` only for `DEFAULT_MAIN_BRANCH`, and Flink/Spark also use the same `liveOnMainBranch` gating.
- However, the orphan cleaner’s `validBranches()` scans all active branches; since branches have independent snapshot directories, writes and snapshot expiration can also be performed via the branch table. When new snapshots are generated concurrently on branches and old snapshots are deleted due to expiration, stale-list/missing-manifest race conditions may also occur.
- Because a branch’s `missingManifest` is `null`, a read failure is treated as an empty result, which may still cause data files referenced by active branches to be marked as orphans. Recommendation: **All active branch snapshots should adopt the same `missing-manifest` abort behavior as the main branch; tags can remain on a best-effort basis**, and additional regression tests for concurrent branch commits and expirations should be implemented.

heye1005:
> The current `early-abort` behavior, which lacks a manifest, only protects `main`: `LocalOrphanFilesClean` constructs `liveSnapshots` only for `DEFAULT_MAIN_BRANCH`, and Flink/Spark also use the same `liveOnMainBranch` gating.
> 
> * However, the orphan cleaner’s `validBranches()` scans all active branches; since branches have independent snapshot directories, writes and snapshot expiration can also be performed via the branch table. When new snapshots are generated concurrently on branches and old snapshots are deleted due to expiration, stale-list/missing-manifest race conditions may also occur.
> * Because a branch’s `missingManifest` is `null`, a read failure is treated as an empty result, which may still cause data files referenced by active branches to be marked as orphans. Recommendation: **All active branch snapshots should adopt the same `missing-manifest` abort behavior as the main branch; tags can remain on a best-effort basis**, and additional regression tests for concurrent branch commits and expirations should be implemented.

Thanks for the review — fixed by building liveSnapshots for every active branch across local/Flink/Spark so all branch snapshots share the same missing-manifest abort behavior, kept tags best-effort also, all tests and spot passed

JingsongLi:
In Flink's distributed orphan file cleanup, the `missing-manifest` sentinel generated by the `keyBy` operation is sent to only one join sub-task. Unless that sub-task has `abortDeletion` enabled, other parallel sub-tasks may still delete the file, making it impossible to prevent accidental deletions globally.

heye1005:
@JingsongLi Thanks for this. I  broadcast the missing-manifest flag to all subtasks before candidate files are processed. I also added a parallelism-5 regression test to cover this case. Could you please take another look again? thanks a lot

paimon-core/src/main/java/org/apache/paimon/operation/LocalOrphanFilesClean.java
```diff
@@ -157,35 +173,59 @@ private void cleanEmptyDataDirectory(List<Path> deleteFiles) {
...
 
     private void collectWithoutDataFile(
-            String branch, Consumer<String> usedFileConsumer, Consumer<String> manifestConsumer)
+            String branch,
+            Consumer<String> usedFileConsumer,
+            Consumer<String> manifestConsumer,
+            Consumer<String> liveManifestConsumer,
+            AtomicBoolean missingManifest)
             throws IOException {
+        Set<Snapshot> liveSnapshots =
```
JingsongLi: [P1] Reuse one active-snapshot view instead of rescanning it

This captures `liveSnapshots` here, but line 207 independently calls `safelyGetAllSnapshots(branch)` again. That helper fixes the path list first and silently skips a snapshot whose file is deleted before it is read. A concurrent S1 -> S2 commit plus retention-1 expiration can therefore make the second scan drop S1 without seeing S2 or setting `missingManifest`. If another branch contributes any used files, the global `usedFiles.isEmpty()` guard is bypassed and old files still referenced by S2 can be deleted. Please process the captured live snapshot objects (adding tags/changelogs separately), or propagate a per-branch incomplete-enumeration signal and abort globally. A deterministic two-branch race test would cover this gap.

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/procedure/SparkOrphanFilesClean.scala
```diff
@@ -102,18 +136,41 @@ case class SparkOrphanFilesClean(
...
+                  .flatMap {
+                    manifestEntry =>
+                      manifestEntry.fileName() +: manifestEntry.file().extraFiles().asScala
+                  }
+                  .map(name => (name, false))
+                  .iterator
               }
           }
       }
+      .cache()
```
JingsongLi, line 166: [P2] Release this cached dataset

`dataFilesWithFlag` is materialized by the `isEmpty` check and later reused by deletion planning, but `doOrphanClean` returns only `usedManifestFiles`; `executeDatabaseOrphanFiles` therefore unpersists only that dataset in its `finally` block. Normal runs and the early return at line 163 leave this cache registered in the Spark session, so repeated or database-wide cleanup can retain large executor memory/local-disk blocks. Please return and release all cached datasets, including every early-return path.

paimon-core/src/main/java/org/apache/paimon/operation/LocalOrphanFilesClean.java
```diff
@@ -108,12 +110,26 @@ public CleanOrphanFilesResult clean()
...
+        AtomicBoolean missingManifest = new AtomicBoolean(false);
+
         // find used files
         Set<String> usedFiles =
                 branches.stream()
-                        .flatMap(branch -> getUsedFiles(branch).stream())
+                        .flatMap(branch -> getUsedFiles(branch, missingManifest).stream())
                         .collect(Collectors.toSet());
 
+        if (usedFiles.isEmpty()) {
```
JingsongLi: [P2] Preserve cleanup for a valid zero-snapshot table

An empty `usedFiles` set is also the successful result for a never-committed table. If an aborted first write leaves old bucket or manifest files, local cleanup now returns here forever, while Flink and Spark still clean against zero references. Please base the abort on an explicit per-branch incomplete or missing signal instead of the empty set itself, so legitimate zero-reference cleanup still works.

JingsongLi (commented):
Inline findings from the review.

heye1005:
> Inline findings from the review.

@JingsongLi  thanks，fixed three issues: 
1.split two functions (get active snapshots per branch,get tags and changelog snapshot), and reused liveSnapshots; 
2. all spark cached datasets are free on exit path
3. local cleanup now supports valid zero-snapshot tables by aborting only on an explicit missing-manifest signal.

paimon-core/src/main/java/org/apache/paimon/operation/OrphanFilesClean.java
```diff
@@ -243,22 +243,26 @@ protected boolean isManagedBlobPack(Path path) {
...
+    protected Set<Snapshot> safelyGetLiveSnapshots(String branch) throws IOException {
         FileStoreTable branchTable = table.switchToBranch(branch);
-        SnapshotManager snapshotManager = branchTable.snapshotManager();
-        ChangelogManager changelogManager = branchTable.changelogManager();
-        TagManager tagManager = branchTable.tagManager();
-        Set<Snapshot> readSnapshots = new HashSet<>(snapshotManager.safelyGetAllSnapshots());
-        readSnapshots.addAll(tagManager.taggedSnapshots());
-        readSnapshots.addAll(changelogManager.safelyGetAllChangelogs());
-        return readSnapshots;
+        return new HashSet<>(branchTable.snapshotManager().safelyGetAllSnapshots());
```
JingsongLi, line 248: [P1] Detect an incomplete live-snapshot enumeration before deleting files

This helper still delegates to `safelyGetAllSnapshots()`, which first freezes the snapshot-path list and then silently skips any path that disappears while it is read. For example, it can list only S1, then a concurrent commit creates S2 and retention-1 expiration deletes S1; the method returns an empty set, never scans S2, and never sets `missingManifest` because no `Snapshot` reaches the new guarded manifest readers. Local/Flink/Spark can then proceed with no live files from this branch and delete old manifests/data files still inherited by S2. This is the original data-loss interleaving one level earlier than the new sentinel. Please make live enumeration itself complete (for example, anchor/revalidate a latest snapshot and abort on a changed/incomplete view) and add a deterministic test where commit+expiration occurs between path listing and snapshot reading.

PR #7716, opened 2026-04-27
[python] Support Kerberos authentication for HDFS

chenghuichen:
### Purpose
Add Kerberos support for HDFS in PyPaimon, aligned with Java Paimon's `SecurityConfiguration`. Supports keytab login (automatic `kinit`) and existing ticket cache mode, with fallback keys `security.principal` / `security.keytab` for Java compatibility.

Includes `SecurityOptions` config class, unit tests, sample script, and documentation.

### Tests
- pypaimon/tests/kerberos_test.py
- **Manually E2E tested on Aliyun EMR**

### DIFF
--- a/paimon-python/pypaimon/filesystem/pyarrow_file_io.py
+++ b/paimon-python/pypaimon/filesystem/pyarrow_file_io.py
@@ -32,7 +32,7 @@ from pyarrow._fs import FileSystem
 
 from pypaimon.common.file_io import FileIO
 from pypaimon.common.options import Options
-from pypaimon.common.options.config import OssOptions, S3Options
+from pypaimon.common.options.config import OssOptions, S3Options, SecurityOptions
 from pypaimon.common.options.options_utils import OptionsUtils
 from pypaimon.common.uri_reader import UriReaderFactory
 from pypaimon.filesystem.jindo_file_system_handler import JindoFileSystemHandler, JINDO_AVAILABLE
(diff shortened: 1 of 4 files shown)


chenghuichen:
Hi @JingsongLi , could you please review this PR when you have a chance?

paimon-python/pypaimon/filesystem/pyarrow_file_io.py
```diff
@@ -256,13 +256,61 @@ def _initialize_hdfs_fs(self, scheme: str, netloc: Optional[str]) -> FileSystem:
...
+    def _get_ticket_cache_path() -> Optional[str]:
+        cc = os.environ.get('KRB5CCNAME')
+        if cc:
+            if cc.startswith('FILE:'):
+                return cc[5:]
+            return cc
+        default_path = f'/tmp/krb5cc_{os.getuid()}'
+        if os.path.exists(default_path):
+            return default_path
+        return None
```
JingsongLi, line 316: After calling kinit, the code calls _get_ticket_cache_path() to find the resulting ticket. But if KRB5CCNAME is not set and the default `/tmp/krb5cc_{os.getuid()}` doesn't exist yet (e.g., first-time kinit on a fresh container), _get_ticket_cache_path returns None, and the code silently falls through to SIMPLE auth. This is a real bug — kinit succeeded but the connection won't use Kerberos.

Suggestion: after a successful kinit, either trust that the default cache path exists (skip the os.path.exists check) or raise an error if no ticket cache is found.

JingsongLi:
+1

PR #7717, opened 2026-04-27
[Flink] Add option to avoid shuffle in bucket unaware append sink

nickdelnano:
### Purpose

Bucket unaware append table [1] is a great choice for streaming events into Paimon format for batch consumers. These types of streams can be very high throughput like clickstream data. Currently there is shuffle in the writer https://github.com/apache/paimon/pull/4203 and in my production use cases (Kafka --> Paimon) this shuffles a _lot_ of data.

rebalance was added to avoid chaining, instead we can use [startNewChain](https://nightlies.apache.org/flink/flink-docs-stable/docs/dev/datastream/operators/overview/#task-chaining-and-resource-groups). This can avoid the deadlock issue described in that PR without shuffle.

[1] https://cwiki.apache.org/confluence/display/PAIMON/PIP-6%3A+Unaware-Bucket+Table

### Tests

FlinkCdcSyncTableSinkITCase.java has tests for schema evolution, however #4203 mentions this is hard to exercise in tests

### DIFF
--- a/docs/layouts/shortcodes/generated/flink_connector_configuration.html
+++ b/docs/layouts/shortcodes/generated/flink_connector_configuration.html
@@ -362,5 +362,11 @@ under the License.
             <td>Integer</td>
             <td>Defines a custom parallelism for the unaware-bucket table compaction job. By default, if this option is not defined, the planner will derive the parallelism for each statement individually by also considering the global configuration.</td>
         </tr>
+        <tr>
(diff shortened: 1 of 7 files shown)


nickdelnano:
@JingsongLi I saw you authored https://github.com/apache/paimon/pull/4203 which relates to this PR, can you review it?

JingsongLi:
+1

PR #7718, opened 2026-04-28
[format] Update Vortex JNI to 0.69.0

jerry-024:
### Purpose

This PR upgrades the Vortex JNI integration from 0.24.0 to 0.69.0 and adapts Paimon's Vortex format implementation to the updated Java API and protobuf layout.

It also fixes several correctness issues found during the upgrade:
- Write record batches through Arrow IPC byte arrays instead of Arrow FFI, so Java can reset Arrow writers after each flush without keeping native-borrowed Arrow buffers alive until file close.
- Keep the Vortex writer close path robust by attempting to close both the native writer and Arrow writer even when flush fails, while preserving suppressed cleanup exceptions.
- Support projected reads that include virtual row-tracking fields by reading only physical columns and mapping them back by field id/name.
- Update Vortex scalar, dtype, expression, temporal metadata, and JNI wrappers for 0.69.0.


JingsongLi (approved):
+1

PR #7719, opened 2026-04-28
[lance] Fix ArrayIndexOutOfBoundsException when reading vector search results with Lance format

XiaoHongbo-Hope:
### Purpose
Reading vector search results on lance format throws `java.lang.ArrayIndexOutOfBoundsException: Index 2 out of bounds for length 2`. This PR fixes the issue so that lance format can work well in vector search


JingsongLi (approved):
+1

PR #7722, opened 2026-04-28
[core] Replace reflective FileIO lookups with HadoopOptionsProvider interface

sundapeng:
### Purpose

Add `HadoopOptionsProvider` so format modules (Lance, Vortex) can pull hadoop-style fs options from a `FileIO` through a typed contract instead of `Class.forName` + reflective `Method.invoke`.

The reflection silently masked signature drift — `VortexUtils` was calling the no-arg `hadoopOptions()` on `JindoFileIO` and never picked up `jindocache` read/write config. With the interface, dispatch is checked at compile time and collapses to a single `instanceof` check.

### Changes

- Add `HadoopOptionsProvider` in `paimon-common`
- Implement it on `HadoopFileIO`, `OSSFileIO`, `JindoFileIO`, `PluginFileIO`
- `LanceUtils` / `VortexUtils`: one `instanceof` check, drop `Class.forName` blocks
- Split `VortexUtils.toVortexSpecified` into `forReader` / `forWriter` so the writer path passes the correct `opType`

### Tests

Existing tests in affected modules pass.

### API and Format

No format changes. Adds one new interface `HadoopOptionsProvider`.

### Documentation

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/fs/HadoopOptionsProvider.java
+++ b/paimon-common/src/main/java/org/apache/paimon/fs/HadoopOptionsProvider.java
@@ -1,0 +1,38 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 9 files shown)


sundapeng:
One thing I am not sure about: the naming. Two options here, I think we need to pick one as a whole.

**Option A: `HadoopOptionsProvider` + `hadoopOptions(path, opType)`**
- Pros: matches the existing no-arg `hadoopOptions()` already on `HadoopFileIO` / `OSSFileIO`, so the convention is consistent. The `fs.*` key style does come from the Hadoop ecosystem, so the name is not wrong.
- Cons: the return type is actually `Options` (paimon-internal map), not Hadoop `Configuration`, the interface itself does not depend on Hadoop. So the name is a bit narrower than what it really is. Also `provider.hadoopOptions(...)` reads a little redundant with the interface name.

**Option B: `FileIOOptionsProvider` + `options(path, opType)`**
- Pros: more accurate, the contract is just key/value options scoped to a `FileIO`, no Hadoop type involved. `provider.options(path, opType)` reads cleanly. Also leaves room if we later want to pass non-`fs.*` style options through the same hook.
- Cons: breaks consistency with the existing `hadoopOptions()` method name, and `options()` alone does not tell the reader that it is `fs.*` style — need Javadoc to clarify.

I am leaning a little to Option A for now, mainly because of the consistency with the existing `hadoopOptions()` method and the small surface (only 4 impls). But Option B feels more accurate long-term. No strong opinion. WDYT?

jerry-024:
+1

PR #7723, opened 2026-04-28
[spark] Refactor BatchWrite subclasses into base logic + per-version wrappers

kerwin-zk:
### Purpose

Follow-up of #7648 (Spark 4.1 module) and a sibling of #7721. After landing the reverse-shim layout, two of the files under `paimon-spark-4.0/src/main` only existed as shadows because their compilation unit defined a Scala class that **`extends BatchWrite`**. Spark 4.1 added a default method `BatchWrite.commit(WriterCommitMessage[], WriteSummary)` whose `WriteSummary` parameter type does not exist on Spark 4.0; a class compiled against 4.1 that mixes in `BatchWrite` carries the inherited `commit(.., WriteSummary)` signature in its method table, which JVM `ObjectStreamClass.getPrivateMethod` lazy-links during Spark task serialization and crashes 4.0 with `ClassNotFoundException: WriteSummary`.

This PR refactors both affected classes into the same **base + per-version wrapper** pattern:

- `PaimonBatchWrite` (used by V2 writes)
- `FormatTableBatchWrite` (used by `FormatTable` V2 writes — was previously a `private case class` inside `PaimonFormatTable.scala`)

For each, the body lives in a new abstract base in `paimon-spark-common` that deliberately does *not* extend `BatchWrite` (renamed protected helpers: `commitMessages`, `abortMessages`, `createPaimonDataWriterFactory`, `createFormatTableDataWriterFactory`). Each per-version module (`paimon-spark3-common`, `paimon-spark4-common`, `paimon-spark-4.0/src/main`) ships a thin wrapper that mixes in `BatchWrite` and forwards the four `BatchWrite` methods to the base helpers. Routing happens through two new `SparkShim` factories so each Spark version's scalac compiles the right `extends BatchWrite` mixin.

The Spark 4.0 shadow of `PaimonFormatTable.scala` is no longer needed and is deleted; only the new thin `FormatTableBatchWrite.scala` wrapper remains under `paimon-spark-4.0/src/main`.

### Tests

CI

### API and Format

No new public API. Two internal factories added to `org.apache.spark.sql.paimon.shims.SparkShim`:

- `createPaimonBatchWrite(table, writeSchema, dataSchema, overwritePartitions, copyOnWriteScan)`
- `createFormatTableBatchWrite(table, overwriteDynamic, overwritePartitions, writeSchema)`

### Documentation

No user-facing changes.

### DIFF
--- a/paimon-spark/paimon-spark-4.0/src/main/scala/org/apache/paimon/spark/format/FormatTableBatchWrite.scala
+++ b/paimon-spark/paimon-spark-4.0/src/main/scala/org/apache/paimon/spark/format/FormatTableBatchWrite.scala
@@ -16,47 +16,34 @@
  * limitations under the License.
  */
 
-package org.apache.paimon.spark.write
+package org.apache.paimon.spark.format
 
-import org.apache.paimon.spark.rowops.PaimonCopyOnWriteScan
-import org.apache.paimon.table.FileStoreTable
+import org.apache.paimon.table.FormatTable
 
 import org.apache.spark.sql.connector.write.{BatchWrite, DataWriterFactory, PhysicalWriteInfo, WriterCommitMessage}
 import org.apache.spark.sql.types.StructType
 
 /**
- * Spark-4.0 shadow wrapper. Source-identical to the `paimon-spark4-common` version but compiled
- * against Spark 4.0.2; the maven shade order picks `paimon-spark-4.0/target/classes` ahead of the
- * shaded 4-common copy, so the class metadata loaded at runtime does not include the 4.1-only
- * `BatchWrite.commit(.., WriteSummary)` signature that triggers `ClassNotFoundException` via
- * `ObjectStreamClass.getPrivateMethod` during Spark task serialization.
+ * Spark 4.0-compatible shadow of the `paimon-spark4-common` `FormatTableBatchWrite`. Compiled
+ * against 4.0.2 so its class file's method table does not carry the `commit(.., WriteSummary)`
+ * signature added by Spark 4.1's `BatchWrite` default method, avoiding `ClassNotFoundException:
+ * WriteSummary` lazy-linking on 4.0 task serialization.
  */
-class PaimonBatchWrite(
-    table: FileStoreTable,
-    writeSchema: StructType,
-    dataSchema: StructType,
+class FormatTableBatchWrite(
+    table: FormatTable,
+    overwriteDynamic: Option[Boolean],
     overwritePartitions: Option[Map[String, String]],
-    copyOnWriteScan: Option[PaimonCopyOnWriteScan])
-  extends PaimonBatchWriteBase(table, writeSchema, dataSchema, overwritePartitions, copyOnWriteScan)
+    writeSchema: StructType)
+  extends FormatTableBatchWriteBase(table, overwriteDynamic, overwritePartitions, writeSchema)
   with BatchWrite
   with Serializable {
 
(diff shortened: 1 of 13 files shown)


Zouxxyy:
Let me check it

JingsongLi (commented):
## Review: base + per-version wrapper refactoring for BatchWrite

### Overall

The approach is sound. Spark 4.1's addition of a default method `BatchWrite.commit(.., WriteSummary)` means any class compiled against 4.1 that mixes in `BatchWrite` will carry the `WriteSummary` reference in its method table, breaking Spark 4.0 at runtime via lazy-linking. Splitting the implementation into a base class (no `extends BatchWrite`) in `paimon-spark-common` plus thin per-version wrappers compiled against each Spark version is the correct fix. The code is well-documented with clear scaladoc explaining the rationale.

### Correctness

1. **`commitStarted` visibility change (private -> protected):** Correct and necessary since `abortMessages` now lives in the base class and checks this flag.

2. **`FormatTableBatchWriteBase.abortMessages` has no `commitStarted` guard:** This is consistent with the original `FormatTableBatchWrite` which also lacked the guard. The `PaimonBatchWriteBase.abortMessages` retains it. No regression here.

3. **`FormatTableDataWriter` relocation:** The class moved out of `PaimonFormatTable.scala` (in both the 4.0 shadow and common) into `FormatTableBatchWriteBase.scala` in common. This keeps the writer co-located with the batch-write base logic which makes sense, but please verify it is visible to per-version wrappers that call `createFormatTableDataWriterFactory()`.

### Design observations

4. **Companion `object PaimonBatchWrite.apply` in per-version modules:** After this PR, all instantiation routes through `SparkShimLoader.shim.createPaimonBatchWrite(...)` which calls `new PaimonBatchWrite(...)` directly. The companion `apply` methods appear unused. If they exist solely for binary compatibility with call-sites that used the former `case class` constructor syntax, please confirm whether any such call-sites remain; otherwise these could be removed to avoid confusion.

5. **`with Serializable` on both base classes:** Since `BatchWrite` instances live exclusively on the driver (only `DataWriterFactory` is serialized to executors), `Serializable` is not strictly required. It is harmless and preserves the implicit contract of the former `case class`, but worth noting. The transitive serializability of `SparkMetricRegistry` and `BatchWriteBuilder` held by `PaimonBatchWriteBase` could cause surprising `NotSerializableException`s if something does accidentally attempt serialization.

6. **Source duplication across wrappers:** The three `PaimonBatchWrite` wrappers (spark3-common, spark4-common, spark-4.0) and three `FormatTableBatchWrite` wrappers are near-identical. This is unavoidable given the design constraint. A brief comment at the top of each (already present -- good) explaining this is a deliberate compile-target copy helps future maintainers.

### Minor nits

- In the `paimon-spark-4.0` `PaimonBatchWrite.scala` scaladoc, the phrase "maven shade order picks `paimon-spark-4.0/target/classes` ahead of the shaded 4-common copy" is an implementation detail that may become stale; consider referencing the build mechanism more generically (e.g., "classpath ordering ensures the 4.0 shadow takes precedence").
- The `FormatTableBatchWriteBase` constructor exposes `batchWriteBuilder` as `protected val`. If no subclass other than the thin wrappers needs it, `private[format]` would tighten visibility.

### Summary

Well-structured refactoring that correctly isolates the Spark version-specific mixin. The pattern is consistent across both affected classes. No functional regressions observed. Only cosmetic/hygiene items noted above.

JingsongLi:
+1

PR #7727, opened 2026-04-29
[python] Add PyPaimon maintenance operations

HansiChan:
### Purpose

This PR adds native PyPaimon APIs and CLI commands for table maintenance and snapshot inspection.

### Changes

- Add SnapshotManager.list_snapshots() and extend table snapshot CLI with --id, --earliest, --time-millis, and --all.
- Add table.new_maintenance() with compact(partition=None) and rescale_bucket(bucket_num, partition=None).
- Add CLI commands for table compact and table rescale.
- Document the new snapshot and maintenance APIs.

Native Python maintenance currently rejects row tracking, data evolution, and deletion-vector tables because those modes require extra metadata handling beyond a plain data rewrite.

### Tests

- PYTHONPATH=. python -m pytest pypaimon/tests/table/maintenance_test.py
- PYTHONPATH=. python -m pytest pypaimon/tests/cli_table_test.py -k snapshot
- python -m flake8 --config=paimon-python/dev/cfg.ini paimon-python/pypaimon/cli/cli_table.py paimon-python/pypaimon/common/json_util.py paimon-python/pypaimon/snapshot/snapshot_manager.py paimon-python/pypaimon/table/file_store_table.py paimon-python/pypaimon/table/maintenance.py paimon-python/pypaimon/write/file_store_commit.py paimon-python/pypaimon/write/table_commit.py paimon-python/pypaimon/tests/cli_table_test.py paimon-python/pypaimon/tests/table/maintenance_test.py

### DIFF
--- a/docs/docs/pypaimon/python-api.mdx
+++ b/docs/docs/pypaimon/python-api.mdx
@@ -527,6 +527,30 @@ arrow_table = table_read.to_arrow(splits)
 pandas_df = table_read.to_pandas(splits)
 ```
 
+### Snapshot Query
+
+You can inspect table snapshots through `SnapshotManager`:
+
+```python
+snapshot_manager = table.snapshot_manager()
+
+latest = snapshot_manager.get_latest_snapshot()
+earliest = snapshot_manager.try_get_earliest_snapshot()
+snapshot_10 = snapshot_manager.get_snapshot_by_id(10)
+snapshots = snapshot_manager.list_snapshots()
+snapshot_at_time = snapshot_manager.earlier_or_equal_time_mills(1700000000000)
+```
+
+The CLI also supports querying snapshots:
+
+```shell
+paimon -c catalog.yaml table snapshot default.my_table
+paimon -c catalog.yaml table snapshot default.my_table --id 10
+paimon -c catalog.yaml table snapshot default.my_table --earliest
+paimon -c catalog.yaml table snapshot default.my_table --time-millis 1700000000000
+paimon -c catalog.yaml table snapshot default.my_table --all
+```
+
 ### Shard Read
 
 Shard Read allows you to read data in parallel by dividing the table into multiple shards. This is useful for
@@ -758,6 +782,50 @@ table.rollback_to('v3')  # tag name
 
 The `rollback_to` method accepts either an `int` (snapshot ID) or a `str` (tag name) and automatically dispatches
 to the appropriate rollback logic.
+
+## Maintenance
+
+PyPaimon supports native maintenance helpers for rewriting current table data.
+
+### Compact
+
+Use `compact` to rewrite the current visible records and commit a `COMPACT` snapshot. You can compact the whole table
+or a specific partition:
+
+```python
+table = catalog.get_table('database_name.table_name')
+result = table.new_maintenance().compact({'dt': '2024-01-01'})
+
+print(result.snapshot_id)
+print(result.rewritten_record_count)
+print(result.rewritten_file_count)
+```
+
+You can also trigger compaction from the CLI:
+
+```shell
+paimon -c catalog.yaml table compact default.my_table --partition dt=2024-01-01
+```
+
+### Rescale Bucket
+
+Use `rescale_bucket` to rewrite data with a new bucket number. For partitioned tables, specify the target partition:
+
+```python
+table = catalog.get_table('database_name.table_name')
+result = table.new_maintenance().rescale_bucket(
+    bucket_num=4,
+    partition={'dt': '2024-01-01'}
+)
+```
+
+The same operation is available from the CLI:
+
+```shell
+paimon -c catalog.yaml table rescale default.my_table --bucket-num 4 --partition dt=2024-01-01
+```
+
+Native Python maintenance does not yet support row tracking, data evolution, or deletion vectors.
 
 ## Streaming Read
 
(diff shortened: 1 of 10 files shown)


JingsongLi (commented):
## Review: [python] Add PyPaimon maintenance operations

Overall this is a well-structured addition. The design is clean — `TableMaintenance` acts as a facade that reads existing data through the standard read path and rewrites it through the write path, keeping the maintenance logic separate from core table internals. A few observations and suggestions:

---

### Correctness

1. **`BatchTableCommit.compact()` sets `batch_committed = True` via `_check_committed()`, but the flag name is misleading for a compact operation.** After calling `compact()`, any subsequent call to `commit()` or other mutating methods on the same `BatchTableCommit` instance will raise, which is fine for the current use in `TableMaintenance` (a fresh commit is created per operation). Just noting this coupling.

2. **`_rewrite` does not close `table_read`** — only `table_write` is closed. If `ReadBuilder.new_read()` holds resources (file handles, buffers), they could leak on large tables. Consider adding `table_read.close()` in a `finally` block (or using a context manager if available).

3. **`list_snapshots()` in the diff loads every snapshot file sequentially.** For tables with hundreds of snapshots this will be slow and memory-intensive. The existing `get_snapshots_batch` uses parallel I/O — consider either reusing it here or at minimum documenting that this method is not intended for production use on long-lived tables. Alternatively, add an optional `limit` parameter.

4. **Race between commit and reading the latest snapshot** — In `compact()`, after `table_commit.compact(...)` succeeds, you read `get_latest_snapshot()` again. If another writer commits between those two operations, the returned `snapshot_id` will belong to the other writer's snapshot, not the compaction snapshot. The `FileStoreCommit._try_commit` path likely has the committed snapshot ID available; surfacing it would be more robust. For now, this is an edge case but worth a TODO comment.

---

### Design

5. **`JSON.__to_dict` list handling** — The new guard at the top of `__to_dict` that checks `isinstance(obj, list)` changes the method's contract: previously it only accepted dataclass objects, now it silently handles lists. However, the return type annotation still says `Dict[str, Any]` while it may now return a `List`. This could confuse callers. Consider renaming the top-level dispatch or making the list serialization a separate helper (e.g., `__serialize(obj)` that handles both scalars, lists, and dataclasses).

6. **`copy(options, allow_bucket_change=False)`** — Adding `allow_bucket_change` is pragmatic, but it makes the public API of `FileStoreTable.copy()` aware of a single caller's requirements. An alternative would be a private `_copy_with_bucket(bucket_num)` helper on `TableMaintenance` itself, keeping `copy()` strict. Minor design preference.

7. **CLI `_load_file_store_table` duplicates catalog loading logic** that already exists in `cmd_table_snapshot`. Since the snapshot command already does the same catalog load + table fetch + isinstance check, consider extracting the shared logic into `_load_file_store_table` and having `cmd_table_snapshot` use it as well, to reduce duplication.

---

### Minor / Nits

8. The PR diff shows `list_snapshots()` defined twice — once in the diff's new addition and once apparently already on `master` (lines 227-258 in the file). If the master version was merged separately, this PR should not re-add it; if it is the same PR, the diff shown is confusing. Please verify there is no duplicate definition.

9. In `maintenance.py`, the `_rewrite` method builds a `read_builder` from `self.table` but writes to `write_table` (which may be a rescaled copy). The asymmetry is correct, but a brief inline comment explaining "read from original table, write to target table" would improve readability for future maintainers.

10. Tests are clear and cover the two main operations well. Consider adding a test for the error paths: compact on a table with no snapshot, rescale with `bucket_num=0`, and compact on a deletion-vector table to verify the `NotImplementedError` guard.

---

Nice work. The guarded rejection of row-tracking / data-evolution / deletion-vector tables is the right call for now. The core read-then-rewrite approach is sound.

HansiChan:
Thanks for the detailed review. I pushed a follow-up commit to address the actionable items:

- Closed both `table_read` and `table_write` in `_rewrite`.
- Changed `list_snapshots()` to reuse `get_snapshots_batch()` instead of loading snapshots sequentially.
- Split JSON serialization into a top-level `__serialize()` helper so `__to_dict()` keeps a dictionary-only contract.
- Updated `cmd_table_snapshot` to reuse `_load_file_store_table()`.
- Added a short comment in `_rewrite` to clarify that maintenance reads from the original table and writes to the target table.
- Added TODOs for returning the committed snapshot ID directly from the commit path to avoid the latest-snapshot race.
- Added error-path tests for compacting an empty table, `bucket_num=0`, and deletion-vector table rejection.

I also verified that `list_snapshots()` is only defined once in the current branch.

Validated with:

`PYTHONPATH=. python -m pytest pypaimon/tests/table/maintenance_test.py`

`PYTHONPATH=. python -m pytest pypaimon/tests/cli_table_test.py -k snapshot`

`python -m flake8 --config=dev/cfg.ini pypaimon/common/json_util.py pypaimon/snapshot/snapshot_manager.py pypaimon/table/maintenance.py pypaimon/cli/cli_table.py pypaimon/tests/table/maintenance_test.py`

HansiChan:
The current CI failure seems unrelated to this PR. It fails during the Java build in paimon-core with `FileStoreCommitImpl` unable to resolve `cleanUpNoReuseTmpManifests(...)`. I also see the same compilation failure on the latest apache/paimon master, while this PR only changes PyPaimon/docs files.

JingsongLi (commented):
Please remove the maintenance related code. We do not need a universal concept of maintenance, and we do not need to add COMPACT related capabilities for now. Additional discussion is needed.

HansiChan:
Thanks for the clarification. I will close this PR for now since the maintenance / COMPACT scope needs further discussion before being added to PyPaimon.

PR #7730, opened 2026-04-29
[python] add vector type in DataType system

steFaiz:
### Purpose
This PR adds the Vector Type for python module's DataType System. The related java version is: https://github.com/apache/paimon/pull/7204

This PR do not contain vector store nor vector-search-related parts. I'll try to add those in future work.

### Tests
Added unit test and e2e test.


JingsongLi (approved):
+1

PR #7734, opened 2026-04-29
[iceberg] Support timestamp conversion for IcebergDataField

ArnavBalyan:
### Purpose
 - Paimon table containing a TIME column fails when Iceberg is enabled.
 - Support TIME_WITHOUT_TIME_ZONE case, all values are scaled by 1000 since Paimon stores milliseconds internally.
 - Output follows Iceberg convention for storing timestamp type.

### Tests
 - Unit tests


JingsongLi (approved):
+1

PR #7737, opened 2026-04-29
[Flink] Fix PostponeFixedBucketChannelComputer routing all records to same channel

leaves12138:
## Summary

Fix a bug in `PostponeFixedBucketChannelComputer` where all records in the same partition are routed to the same downstream channel in batch mode, causing only one subtask to actually process data.

## Root Cause

In the `channel()` method, the variable `bucket` was set to the total number of buckets for the partition (from `knownNumBuckets`). Since this value is the same for all records in the same partition, `ChannelComputer.select(partition, bucket, numChannels)` always returns the same channel — all records go to one subtask.

## Fix

Compute a per-record bucket by hashing the `trimmedPrimaryKey` and taking modulo `numBuckets`. This distributes records with different primary keys across different channels/subtasks, similar to how `ChannelComputer.startChannel` handles `Integer.MIN_VALUE`.

## Testing

Verified the logic matches the pattern used in `RowDataChannelComputer` (fixed bucket tables) and `PostponeBucketChannelComputer` (streaming postpone tables).

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/PostponeFixedBucketChannelComputer.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/PostponeFixedBucketChannelComputer.java
@@ -55,7 +55,12 @@ public class PostponeFixedBucketChannelComputer implements ChannelComputer<Inter
     @Override
     public int channel(InternalRow record) {
         BinaryRow partition = partitionKeyExtractor.partition(record);
-        int bucket = knownNumBuckets.computeIfAbsent(partition, p -> numChannels);
+        int numBuckets = knownNumBuckets.computeIfAbsent(partition, p -> numChannels);
+        int hash = partitionKeyExtractor.trimmedPrimaryKey(record).hashCode();
+        if (hash == Integer.MIN_VALUE) {
+            hash = Integer.MAX_VALUE;
+        }
+        int bucket = Math.abs(hash) % numBuckets;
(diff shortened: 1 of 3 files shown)


JingsongLi:
@leaves12138  Add the test.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/PostponeFixedBucketChannelComputer.java
```diff
@@ -55,7 +55,12 @@ public void setup(int numChannels) {
     @Override
     public int channel(InternalRow record) {
         BinaryRow partition = partitionKeyExtractor.partition(record);
-        int bucket = knownNumBuckets.computeIfAbsent(partition, p -> numChannels);
+        int numBuckets = knownNumBuckets.computeIfAbsent(partition, p -> numChannels);
+        int hash = partitionKeyExtractor.trimmedPrimaryKey(record).hashCode();
```
JingsongLi: Just use `FixedBucketRowKeyExtractor`?

leaves12138: Fixed automatically by Codex. `PostponeFixedBucketChannelComputer` now reuses `FixedBucketRowKeyExtractor`, with `FixedBucketRowKeyExtractor#bucket(int numBuckets)` added so postpone fixed-bucket writes can still use the per-partition bucket count from `knownNumBuckets`.

JingsongLi (approved):
+1

PR #7739, opened 2026-04-29
[python] Generate input changelogs from Python writer

junmuz:
### Purpose

- Add input changelog producer support to the Python Paimon library, including DataChangelogWriter, ManifestCommittable, and integration into the write/commit pipeline
- Reject changelog-producer on tables without primary keys, since changelog generation requires PK-based deduplication
- Decouple changelog file format from data file format via a new changelog-file.format config option

### Tests

There are new tests added. Python scripts have been executed manually and the generated changelogs are verified to be readable from a FlinkSQL job.

### Limitation

Changelogs are currently only generated for inserts.

### DIFF
--- a/paimon-python/pypaimon/common/options/core_options.py
+++ b/paimon-python/pypaimon/common/options/core_options.py
@@ -470,6 +470,14 @@ class CoreOptions:
                           "Options: none, input, full-compaction, lookup.")
     )
 
+    CHANGELOG_FILE_FORMAT: ConfigOption[str] = (
+        ConfigOptions.key("changelog-file.format")
+        .string_type()
+        .no_default_value()
+        .with_description("Specify the file format of changelog files. "
+                          "Currently parquet, avro and orc are supported.")
+    )
+
     MERGE_ENGINE: ConfigOption[MergeEngine] = (
         ConfigOptions.key("merge-engine")
         .enum_type(MergeEngine)
@@ -958,6 +966,9 @@ class CoreOptions:
 
     def changelog_producer(self, default=None):
         return self.options.get(CoreOptions.CHANGELOG_PRODUCER, default)
+
+    def changelog_file_format(self, default=None):
+        return self.options.get(CoreOptions.CHANGELOG_FILE_FORMAT, default)
 
     def merge_engine(self, default=None):
         return self.options.get(CoreOptions.MERGE_ENGINE, default)
--- a/paimon-python/pypaimon/write/commit_message.py
+++ b/paimon-python/pypaimon/write/commit_message.py
@@ -29,9 +29,9 @@ class CommitMessage:
     partition: Tuple
     bucket: int
     new_files: List[DataFileMeta]
-    changelog_files: List[DataFileMeta] = field(default_factory=list)
     check_from_snapshot: Optional[int] = -1
     index_deletes: List['IndexManifestEntry'] = field(default_factory=list)
+    changelog_files: List[DataFileMeta] = field(default_factory=list)
 
     def is_empty(self):
         return not self.new_files and not self.index_deletes and not self.changelog_files
--- a/paimon-python/pypaimon/write/commit_message.py
+++ b/paimon-python/pypaimon/write/commit_message.py
@@ -29,8 +29,9 @@ class CommitMessage:
     partition: Tuple
     bucket: int
     new_files: List[DataFileMeta]
+    changelog_files: List[DataFileMeta] = field(default_factory=list)
     check_from_snapshot: Optional[int] = -1
     index_deletes: List['IndexManifestEntry'] = field(default_factory=list)
 
     def is_empty(self):
-        return not self.new_files and not self.index_deletes
+        return not self.new_files and not self.index_deletes and not self.changelog_files
--- a/paimon-python/pypaimon/schema/schema.py
+++ b/paimon-python/pypaimon/schema/schema.py
@@ -48,6 +48,13 @@ class Schema:
         self.options = options if options is not None else {}
         self.comment = comment
 
+        changelog_producer = self.options.get(CoreOptions.CHANGELOG_PRODUCER.key(), 'none')
+        if changelog_producer != 'none' and not self.primary_keys:
+            raise ValueError(
+                f"Cannot set 'changelog-producer' to '{changelog_producer}' on a table without primary keys. "
+                f"Changelog producer requires primary keys to be defined."
+            )
+
     @staticmethod
     def from_pyarrow_schema(pa_schema: pa.Schema, partition_keys: Optional[List[str]] = None,
                             primary_keys: Optional[List[str]] = None, options: Optional[Dict] = None,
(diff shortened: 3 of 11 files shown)


paimon-python/pypaimon/common/options/core_options.py
```diff
@@ -292,6 +292,14 @@ class CoreOptions:
                           "Options: none, input, full-compaction, lookup.")
     )
 
+    CHANGELOG_FILE_FORMAT: ConfigOption[str] = (
+        ConfigOptions.key("changelog-file.format")
+        .string_type()
+        .no_default_value()
```
junmuz, line 476: This is set as defined in [the docs](https://paimon.apache.org/docs/master/maintenance/configurations/)

paimon-python/pypaimon/write/commit_message.py
```diff
@@ -27,7 +27,8 @@ class CommitMessage:
     partition: Tuple
     bucket: int
     new_files: List[DataFileMeta]
+    changelog_files: List[DataFileMeta] = field(default_factory=list)
```
junmuz: Should we move it to the end for backward compatibility?

junmuz:
@JingsongLi @XiaoHongbo-Hope The PR adds support for generating changelogs for input changelog producer table (primarily for inserts). Can you review it?

XiaoHongbo-Hope:
> @JingsongLi @XiaoHongbo-Hope The PR adds support for generating changelogs for input changelog producer table (primarily for inserts). Can you review it?

sure

XiaoHongbo-Hope:
Could you add coverage for input changelog with row tracking enabled?

junmuz:
> Could you add coverage for input changelog with row tracking enabled?

@XiaoHongbo-Hope   [Row tracking ](https://paimon.apache.org/docs/master/append-table/row-tracking/)requires tables without primary keys (row-tracking.enabled is only valid on append-only tables with bucket=-1), while changelog-producer requires primary keys to be defined. These two features looks mutually exclusive by design. I don't see a valid table configuration where both can be active simultaneously. The existing test [test_reject_changelog_producer_on_append_only_table](https://github.com/apache/paimon/pull/7739/changes#diff-38a7ba6c1e4012afdf0076beb2394f16b652f917146b12a3c29c7076e1edcc00R269-R282) already verifies that we reject changelog-producer on tables without primary keys, which should cover  all row-tracking tables.

XiaoHongbo-Hope:
> > Could you add coverage for input changelog with row tracking enabled?
> 
> @XiaoHongbo-Hope [Row tracking ](https://paimon.apache.org/docs/master/append-table/row-tracking/)requires tables without primary keys (row-tracking.enabled is only valid on append-only tables with bucket=-1), while changelog-producer requires primary keys to be defined. These two features looks mutually exclusive by design. I don't see a valid table configuration where both can be active simultaneously. The existing test [test_reject_changelog_producer_on_append_only_table](https://github.com/apache/paimon/pull/7739/changes#diff-38a7ba6c1e4012afdf0076beb2394f16b652f917146b12a3c29c7076e1edcc00R269-R282) already verifies that we reject changelog-producer on tables without primary keys, which should cover all row-tracking tables.

This validation can still be bypassed because it only runs in Schema.from_pyarrow_schema(). Directly constructing Schema(fields=..., primary_keys=[], options={'changelog-producer': 'input'}) and passing it to catalog.create_table() still creates an append-only table

junmuz:
> > > Could you add coverage for input changelog with row tracking enabled?
> > 
> > 
> > @XiaoHongbo-Hope [Row tracking ](https://paimon.apache.org/docs/master/append-table/row-tracking/)requires tables without primary keys (row-tracking.enabled is only valid on append-only tables with bucket=-1), while changelog-producer requires primary keys to be defined. These two features looks mutually exclusive by design. I don't see a valid table configuration where both can be active simultaneously. The existing test [test_reject_changelog_producer_on_append_only_table](https://github.com/apache/paimon/pull/7739/changes#diff-38a7ba6c1e4012afdf0076beb2394f16b652f917146b12a3c29c7076e1edcc00R269-R282) already verifies that we reject changelog-producer on tables without primary keys, which should cover all row-tracking tables.
> 
> This validation can still be bypassed because it only runs in Schema.from_pyarrow_schema(). Directly constructing Schema(fields=..., primary_keys=[], options={'changelog-producer': 'input'}) and passing it to catalog.create_table() still creates an append-only table

@XiaoHongbo-Hope I have tried addressing your comment. I have moved the guard-rail to the constructor to prevent it from being by-passed.

JingsongLi (commented):
## Review: [python] Generate input changelogs from Python writer

Overall this is a well-structured addition that follows the existing patterns in the codebase. The test coverage is solid. A few observations:

### Correctness Issues

1. **`DataBlobWriter` does not receive `changelog_producer`** — In `file_store_write.py`, `KeyValueDataWriter` and `AppendOnlyDataWriter` are updated to accept the new `changelog_producer` parameter, but `DataBlobWriter` is not:
   ```python
   if self._has_blob_columns():
       return DataBlobWriter(
           table=self.table, partition=partition, bucket=bucket,
           max_seq_number=0, options=options, write_cols=self.write_cols,
       )
   ```
   This silently falls back to `ChangelogProducer.NONE` via the base class default. If someone enables `changelog-producer=input` on a blob table, no changelogs will be produced and no error will be raised. Consider either passing the parameter for consistency, or adding an explicit rejection similar to the append-only table validation.

2. **Shared mutable state in `_current_external_path`** — In `_write_changelog_file`, the changelog path is generated via `self._generate_file_path(changelog_file_name)` which mutates `self._current_external_path`. The subsequent read of `self._current_external_path` is correct in the current sequential flow, but this coupling between `_generate_file_path`'s side-effect and the consumer is fragile. If the method ordering ever changes, or if `_generate_file_path` is called elsewhere between write and metadata creation, the wrong path could be recorded. Returning the external path from `_generate_file_path` as a second value (or using a local variable) would be safer.

### Design Observations

3. **Schema-level validation placement** — The changelog-producer validation in `Schema.__init__` catches the case at schema-creation time, which is good for early failure. However, if a table is loaded from disk (not newly created), this validation does not run. This is acceptable since the Java side would have validated at table creation, but it is worth documenting this assumption.

4. **`CommitMessage` field ordering** — Adding `changelog_files` (with a default) after `check_from_snapshot` (also with a default) is valid Python, but it means the two positional-only fields must always be passed by name when changelog_files is needed. This is fine for the existing call sites but be careful about positional instantiation going forward.

5. **`changelog_record_count` filter is redundant** — In `_try_commit_once`:
   ```python
   changelog_record_count = sum(
       entry.file.row_count for entry in changelog_entries if entry.kind == 0)
   ```
   Since `_collect_changelog_entries` always creates entries with `kind=0`, the filter is a no-op. It reads as defensive coding which is fine, but a comment explaining the intent (e.g., "only count ADD entries") would improve clarity.

### Minor Items

6. The removed `import logging` statement in the `abort` method of `file_store_commit.py` is safe because `logger` is already defined at module level — good cleanup.

7. Tests are comprehensive: they cover none/input/full-compaction/lookup modes, abort cleanup, multiple commits, file format inheritance, and schema rejection. The `test_abort_cleans_up_changelog_files` test is particularly valuable for ensuring no orphan files are left.

### Summary

The implementation correctly limits changelog production to `INPUT` mode (leaving `FULL_COMPACTION` and `LOOKUP` to dedicated compaction processes), which aligns with the Java implementation's behavior. The cleanup paths (both abort and preparation failure) properly handle the new changelog artifacts. The main actionable item is the `DataBlobWriter` gap and the fragile external path coupling.

junmuz:
@JingsongLi I have tried addressing your comments and rebasing the branch as well. Can you re-review the changes? Thanks.

JingsongLi:

JingsongLi (approved):
+1

PR #7742, opened 2026-04-29
[python] Fix limit push-down discarding non-raw_convertible splits

TheR1sing3un:
## How

  `_apply_push_down_limit` now mirrors Java: short-circuit when the
  predicate references any non-partition column, otherwise accumulate
  `split.merged_row_count()` and stop at the limit. Splits with unknown
  merged count fall through to the reader unchanged.

  ## Tests

  `ApplyPushDownLimitUnitTest` drives `_apply_push_down_limit` with
  synthetic splits.

  ## Compatibility

  No public API change. No file format change.

### DIFF
--- a/paimon-python/pypaimon/read/scanner/file_scanner.py
+++ b/paimon-python/pypaimon/read/scanner/file_scanner.py
@@ -31,7 +31,8 @@ from pypaimon.manifest.schema.manifest_entry import ManifestEntry
 from pypaimon.manifest.schema.manifest_file_meta import ManifestFileMeta
 from pypaimon.manifest.simple_stats_evolutions import SimpleStatsEvolutions
 from pypaimon.read.plan import Plan
-from pypaimon.read.push_down_utils import (remove_row_id_filter,
+from pypaimon.read.push_down_utils import (_get_all_fields,
+                                           remove_row_id_filter,
                                            trim_and_transform_predicate)
 from pypaimon.read.scanner.append_table_split_generator import \
     AppendTableSplitGenerator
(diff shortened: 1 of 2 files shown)


XiaoHongbo-Hope:
Could you please double-check the PR description and the change? I ran both newly added tests on master,
  and they pass there as well, so they don't seem to reproduce the described pre-fix issue.

TheR1sing3un:
@XiaoHongbo-Hope you're right — the two earlier tests didn't actually distinguish the buggy and fixed implementations. Both used inputs (same-key-twice on a single bucket) where every split ended up non-raw_convertible, which means the pre-fix loop body never ran and the fallback `return splits` returned everything anyway. Thanks for catching it.

I've replaced them with a single, deterministic reproducer that does exercise the bug:

- PK table partitioned on `dt`, `bucket=1`.
- Partition `p1` — two overlapping writes on the same PK → **non-raw_convertible** split.
- Partition `p2` — one write → **raw_convertible** split with `row_count=1`.

`PrimaryKeyTableSplitGenerator` walks partitions in order, so the plan is `[non-raw (p1), raw (p2)]`. With `with_limit(1)` the pre-fix loop skips the non-raw split, then immediately early-returns after the raw one — `limited_splits=[raw]`, p1's data is silently dropped.

End-to-end check:

```
$ git checkout origin/master -- pypaimon/read/scanner/file_scanner.py
$ pytest ...test_limit_drops_non_raw_split_after_raw_budget_is_met
FAILED ... AssertionError: 1 != 2
$ git checkout HEAD -- pypaimon/read/scanner/file_scanner.py
$ pytest ...test_limit_drops_non_raw_split_after_raw_budget_is_met
1 passed
```

Force-pushed [3b7c7484b](https://github.com/apache/paimon/pull/7742/commits/3b7c7484b) with the new reproducer and an updated commit message / PR description that walks through why the bug requires `[non-raw, raw]` ordering. PTAL.

TheR1sing3un:
Update — there was a second divergence from Java's implementation that the previous version of this PR didn't address.

Java's `applyPushDownLimit` accumulates `split.partialMergedRowCount()` (file row_count **minus** deletion-vector cardinality already recorded in the manifest), while Python was using `split.row_count` (the pre-DV upper bound). On a DV-aware raw split, the pre-DV accumulator over-counts and the early-return fires before the reader can produce `limit` rows — even with the "always append" fix from the previous round, a plan like `[raw_with_DV(row_count=10, merged=4), non-raw, non-raw]` with `limit=5` would still early-return on the raw split and leave the reader with only 4 rows, despite the trailing non-raw splits being able to provide more.

`DataSplit.merged_row_count()` already exposes the post-DV value, so the fix is one line — accumulate `merged_row_count()` instead of `row_count`, with a fallback to `row_count` when the merged count is unavailable (older snapshots / certain data-evolution shapes).

Force-pushed [a57200ebb](https://github.com/apache/paimon/pull/7742/commits/a57200ebb):

* `_apply_push_down_limit` now uses `merged_row_count() if not None else row_count` as the budget accumulator.
* New unit tests in `ApplyPushDownLimitUnitTest` drive the accumulator directly with synthetic splits — pypaimon's writer doesn't promote L0 → L1+, and the DV-enabled PK read path skips L0 files, so producing a true DV-aware raw split end-to-end isn't really feasible from a pure-Python fixture. The synthetic-split tests pin down the accumulator semantics without depending on storage layout.
* Both regression tests (`test_limit_drops_non_raw_split_after_raw_budget_is_met` and `test_dv_aware_accumulator_uses_merged_row_count`) fail on `origin/master` and pass post-fix.
* PR description rewritten to walk through both divergences and how they interact.

PTAL.

paimon-python/pypaimon/read/scanner/file_scanner.py
```diff
@@ -371,10 +371,30 @@ def _apply_push_down_limit(self, splits: List[DataSplit]) -> List[DataSplit]:
...
+        # the split's contribution entirely.
+        # Non-raw_convertible splits cannot be cheaply counted ahead of
+        # read, so we keep them around so the reader still has data to
+        # drain.
         for split in splits:
+            limited_splits.append(split)
             if split.raw_convertible:
-                limited_splits.append(split)
-                scanned_row_count += split.row_count
+                merged = split.merged_row_count()
```
XiaoHongbo-Hope: Could you double-check this against current Java semantics? Or the PR description is not latest? It says this mirrors Java

TheR1sing3un: > Could you double-check this against current Java semantics? Or the PR description is not latest? It says this mirrors Java

Hi, How about now? Thanks for your reminder!

XiaoHongbo-Hope:
Thanks for the update - the latest code looks good to me. Two small follow-ups, just suggestions:
 1. The PR description still seems not correct, could you please double check again? 
 2. Could you please keep the comments a bit lighter?

TheR1sing3un:
> Thanks for the update - the latest code looks good to me. Two small follow-ups, just suggestions:
> 
> 1. The PR description still seems not correct, could you please double check again?
> 2. Could you please keep the comments a bit lighter?

You're right. My opus is a bit wordy, lol. Let me update it

TheR1sing3un:
ready for the final review! @XiaoHongbo-Hope

paimon-python/pypaimon/tests/reader_split_generator_test.py
```diff
@@ -330,5 +330,130 @@ def test_sliced_split_merged_row_count(self):
         # This test ensures that if SlicedSplit is created, merged_row_count() works correctly
 
 
+class ApplyPushDownLimitUnitTest(unittest.TestCase):
```
XiaoHongbo-Hope, line 333: Thanks for trimming the production-code comments. Could we also make the new unit-test docstrings a bit lighter? Also some comments seem to be slightly out of sync with the latest implementation

TheR1sing3un: > Thanks for trimming the production-code comments. Could we also make the new unit-test docstrings a bit lighter? Also some comments seem to be slightly out of sync with the latest implementation

done~

XiaoHongbo-Hope:
Thanks for the updates — this looks much cleaner now. One tiny wording nit: the current Java code seems
  to use `split.mergedRowCount()`, while the PR description / docstring mentions `partialMergedRowCount`.
 This will make other developers a little confusion. Otherwise LGTM from my side.

TheR1sing3un:
> Thanks for the updates — this looks much cleaner now. One tiny wording nit: the current Java code seems to use `split.mergedRowCount()`, while the PR description / docstring mentions `partialMergedRowCount`. This will make other developers a little confusion. Otherwise LGTM from my side.

done~

XiaoHongbo-Hope:
+1

TheR1sing3un:
> +1

cc @JingsongLi

JingsongLi (approved):
+1

PR #7743, opened 2026-04-29
[spark] Case-insensitive field matching when casting nested struct on write.

JunRuiLee:
### Purpose

 When writing BY NAME into a Paimon table whose nested struct field names differ from the source only in case, the insert fails with "field does not exist", because the lookup is case-sensitive and ignores spark.sql.caseSensitive.

  Example:

  ```
  CREATE TABLE t1 (id INT, info STRUCT<Age: INT, Name: STRING>);
  CREATE TABLE t2 (id INT, info STRUCT<name: STRING, age: INT>);
  INSERT INTO t1 VALUES (1, struct(30, 'Alice'));
  INSERT INTO t2 BY NAME SELECT * FROM t1;  -- fails
```

### Tests
- insert by name with case-insensitive nested struct field matching
- insert by name with case-insensitive matching inside array<struct<...>>
- insert by name with case-insensitive matching inside nested struct
- insert by name rejects ambiguous source nested struct fields
- insert by name rejects nested struct target fields colliding under resolver


JingsongLi:
cc @Zouxxyy to take a review.

JunRuiLee:
Hi @Zouxxyy , could you help take a look? Many thanks.

JingsongLi (approved):
+1

PR #7745, opened 2026-04-29
[python] Implement partial-update merge engine in pypaimon

TheR1sing3un:
## Purpose

`pypaimon` exposes `MergeEngine.PARTIAL_UPDATE` in `core_options.py` and accepts `merge-engine: partial-update` as a table option, but the read path never reads that option — `pypaimon/read/reader/sort_merge_reader.py` hardcodes `DeduplicateMergeFunction()`. So a user who:

1. Creates a PK table with `merge-engine: partial-update`,
2. Writes one batch with `[{id: 1, a: 'A', b: null}]`, then another with `[{id: 1, a: null, b: 'B'}]`,
3. Reads back,

gets `[{id: 1, a: null, b: 'B'}]` (silently deduplicated to the latest row) instead of the expected `[{id: 1, a: 'A', b: 'B'}]` (per-field merge of non-null values). **No error, no warning, just wrong data.** The same is true today for `merge-engine: aggregation` and `merge-engine: first-row` — both are silently degraded to dedupe.

This PR ports the core `PartialUpdateMergeFunction` semantics from Java (`paimon-core/.../mergetree/compact/PartialUpdateMergeFunction.java`) and wires the Python read path to dispatch on `merge-engine`. End state:

* `merge-engine: partial-update` works correctly for the common case (no DELETE rows, no sequence-group config, no per-field aggregator overrides).
* `merge-engine: deduplicate` is unchanged.
* `merge-engine: aggregation` / `first-row` raise an explicit `NotImplementedError` instead of silently behaving as dedupe — fail loud now that we have an obvious place for users to escalate from.

### New: `pypaimon/read/reader/partial_update_merge_function.py`

```python
class PartialUpdateMergeFunction:
    def __init__(self, key_arity, value_arity): ...
    def reset(self): ...
    def add(self, kv):
        if not RowKind.is_add_byte(kv.value_row_kind_byte):
            raise NotImplementedError(...)  # DELETE / UPDATE_BEFORE
        # accumulator[i] = kv.value.get_field(i) if non-null
    def get_result(self) -> Optional[KeyValue]:
        # Build a fresh KeyValue from a fresh tuple — never aliased to
        # upstream's reused input kv.
```

The `reset` / `add` / `get_result` protocol matches `DeduplicateMergeFunction` exactly so `SortMergeReader` doesn't change.

### Modified: `pypaimon/read/reader/sort_merge_reader.py`

`SortMergeReaderWithMinHeap.__init__` gains an optional `merge_function` kwarg (defaults to `DeduplicateMergeFunction()` so any direct callers are unchanged).

### Modified: `pypaimon/read/split_read.py`

`MergeFileSplitRead.section_reader_supplier` now picks the merge function based on `self.table.options.merge_engine()`:

```python
DEDUPLICATE     -> DeduplicateMergeFunction()
PARTIAL_UPDATE  -> PartialUpdateMergeFunction(key_arity, value_arity)
AGGREGATE / FIRST_ROW -> NotImplementedError(...)
```

## Out of scope (deliberate, called out in code comments)

* Per-field aggregator overrides (`fields.<name>.aggregate-function=...`) — needs a `FieldAggregator` framework.
* Sequence-group support (`fields.<name>.sequence-group=...`) — needs a `UserDefinedSeqComparator`-equivalent.
* `ignore-delete`, `partial-update.remove-record-on-delete`, `partial-update.remove-record-on-sequence-group` — depend on the above.
* AGGREGATE / FIRST_ROW merge engines — separate PRs each. This PR just upgrades them from "silent dedupe" to "explicit error".
* DELETE / UPDATE_BEFORE row kinds — raise `NotImplementedError` so we never silently corrupt data with a half-implemented contract.

## Linked issue

N/A — surfaced when verifying that `merge-engine: partial-update` actually does what the option name implies in pypaimon.

## Tests

* **`pypaimon/tests/test_partial_update_merge_function.py`** — 11 unit cases driving the merge function with synthetic `KeyValue` instances. Covers: single insert, two-way overlapping merge (overwrite vs fill-null), three-way merge composition, later-null-does-not-clobber, reset between keys, `get_result` before any add, UPDATE_AFTER acceptance, DELETE / UPDATE_BEFORE refusal, **and result decoupling from input kv** (proves the result is built into a fresh tuple so upstream's reused-KeyValue pattern doesn't corrupt us).
* **`pypaimon/tests/test_partial_update_e2e.py`** — 8 end-to-end cases on real PK tables. Covers: two-write merge (`A,_` + `_,B` → `A,B`), three-write left-to-right composition, disjoint keys unaffected, later-non-null wins over earlier non-null, later-null preserves earlier value, deduplicate-engine-unchanged (regression), aggregation / first-row raise `NotImplementedError`.

**Master-vs-fix verification** — checked out `origin/master`'s `sort_merge_reader.py` / `split_read.py` over the new tests:

```
$ git show origin/master:.../sort_merge_reader.py > .../sort_merge_reader.py
$ git show origin/master:.../split_read.py > .../split_read.py
$ pytest pypaimon/tests/test_partial_update_e2e.py
... 6 failed, 2 passed
   FAILED test_partial_update_two_writes_merges_non_null
   FAILED test_partial_update_three_writes_merges_left_to_right
   FAILED test_partial_update_later_value_wins_over_earlier_non_null
   FAILED test_partial_update_later_null_does_not_clobber_earlier_value
   FAILED test_aggregation_engine_raises_not_implemented
   FAILED test_first_row_engine_raises_not_implemented
$ git checkout HEAD -- .../sort_merge_reader.py .../split_read.py
$ pytest pypaimon/tests/test_partial_update_e2e.py
... 8 passed
```

Regression: `pytest pypaimon/tests/{reader_primary_key_test,reader_split_generator_test,reader_append_only_test,test_partial_update_*}.py` → 50 passed, 7 failed. The 7 failures are all pre-existing lance / vortex environment issues unrelated to this PR; the dedupe path is unchanged.

`flake8 --config=dev/cfg.ini` clean.

## API and format

Public API additions:
* `pypaimon.read.reader.partial_update_merge_function.PartialUpdateMergeFunction` — new class.
* `SortMergeReaderWithMinHeap.__init__` gets an optional `merge_function` kwarg (back-compat default).

No public API removals or signature breaks. No file format change.

Behaviour change: tables with `merge-engine: partial-update` now produce per-field merged results (was: silently deduplicated). Tables with `merge-engine: aggregation` or `first-row` now raise `NotImplementedError` (was: silently deduplicated). Both are correctness fixes — the previous behaviour was producing wrong data with no signal to the user.

## Documentation

The new module carries a docstring covering the algorithm, the out-of-scope list, and the link back to the Java reference. The dispatch in `MergeFileSplitRead._build_merge_function` carries an inline comment explaining why we now raise on AGGREGATE / FIRST_ROW.

## Generative AI disclosure

Drafted with assistance from an AI coding tool; the algorithm follows `org.apache.paimon.mergetree.compact.PartialUpdateMergeFunction` and the soundness contract is exercised end-to-end by the regression tests above (which fail on master and pass post-fix).

### DIFF
--- a/paimon-python/pypaimon/read/split_read.py
+++ b/paimon-python/pypaimon/read/split_read.py
@@ -103,6 +103,10 @@ class SplitRead(ABC):
         self.row_tracking_enabled = row_tracking_enabled
         self.value_arity = len(read_type)
         self.nested_name_paths = nested_name_paths
+        # Snapshot the raw value-side schema before _create_key_value_fields
+        # wraps it, so MergeFileSplitRead can hand per-value-field nullable
+        # flags to merge functions that mirror Java's NOT-NULL check.
+        self.value_fields = list(read_type)
 
         self.trimmed_primary_key = self.table.trimmed_primary_keys
         self.read_fields = read_type
@@ -623,6 +627,7 @@ class MergeFileSplitRead(SplitRead):
             return PartialUpdateMergeFunction(
                 key_arity=len(self.trimmed_primary_key),
                 value_arity=self.value_arity,
+                nullables=[f.type.nullable for f in self.value_fields],
             )
         raise NotImplementedError(
             "merge-engine '{}' is not implemented in pypaimon yet "
--- a/paimon-python/pypaimon/read/split_read.py
+++ b/paimon-python/pypaimon/read/split_read.py
@@ -591,95 +591,28 @@ class MergeFileSplitRead(SplitRead):
 
     def _build_merge_function(self):
         """Pick the right MergeFunction implementation for the table's
-        ``merge-engine`` option. ``DEDUPLICATE`` is the default and the
-        only engine supported on the Python read path historically;
-        ``PARTIAL_UPDATE`` is now wired up to its dedicated
-        implementation. The remaining engines (``AGGREGATE`` /
-        ``FIRST_ROW``) used to silently degrade to dedupe — that quietly
-        produced wrong data — so we now raise an explicit
-        ``NotImplementedError`` instead, until they're ported.
+        ``merge-engine`` option.
 
-        For ``PARTIAL_UPDATE``, we also refuse to run when the table
-        configures any option whose semantics this port does not yet
-        implement (sequence-group, per-field aggregator overrides,
-        ignore-delete and friends). Without this guard those options
-        would be silently ignored and produce subtly wrong results —
-        the same anti-pattern this PR exists to close.
+        The pre-flight checks that reject unsupported engines or option
+        combinations live in
+        :func:`pypaimon.read.merge_engine_support.check_supported` and
+        run at ``TableRead.__init__`` time, so by the point this method
+        executes only the supported engines are reachable.
         """
         engine = self.table.options.merge_engine()
         if engine == MergeEngine.DEDUPLICATE:
             return DeduplicateMergeFunction()
         if engine == MergeEngine.PARTIAL_UPDATE:
-            unsupported = self._partial_update_unsupported_options()
-            if unsupported:
-                raise NotImplementedError(
-                    "merge-engine 'partial-update' is enabled together "
-                    "with options that pypaimon does not yet implement: "
-                    "{}. The supported subset is per-key last-non-null "
-                    "merge with no sequence-group, no per-field "
-                    "aggregator override, no ignore-delete and no "
-                    "partial-update.remove-record-on-* flags. Use the "
-                    "Java client for the full feature set, or open an "
-                    "issue to track Python support.".format(
-                        ", ".join(sorted(unsupported))
-                    )
-                )
             return PartialUpdateMergeFunction(
                 key_arity=len(self.trimmed_primary_key),
                 value_arity=self.value_arity,
                 nullables=[f.type.nullable for f in self.value_fields],
             )
-        raise NotImplementedError(
-            "merge-engine '{}' is not implemented in pypaimon yet "
-            "(supported: deduplicate, partial-update). Use the Java "
-            "client or open an issue to track support.".format(engine.value)
+        # check_supported() rejects everything else at TableRead.__init__.
+        raise AssertionError(
+            "unreachable: merge-engine '{}' should have been rejected by "
+            "merge_engine_support.check_supported".format(engine.value)
         )
-
-    # Boolean-valued options that, when truthy, opt the table into
-    # behaviour the Python PartialUpdateMergeFunction does not implement.
-    # Mirrors org.apache.paimon.CoreOptions and the fallback keys in
-    # PartialUpdateMergeFunction.java.
-    _PARTIAL_UPDATE_UNSUPPORTED_BOOLEAN_OPTIONS = (
-        "ignore-delete",
-        "partial-update.ignore-delete",
-        "first-row.ignore-delete",
-        "deduplicate.ignore-delete",
-        "partial-update.remove-record-on-delete",
-        "partial-update.remove-record-on-sequence-group",
-    )
-    _FIELDS_PREFIX = "fields."
-    _FIELD_SEQUENCE_GROUP_SUFFIX = ".sequence-group"
-    _FIELD_AGGREGATE_FUNCTION_SUFFIX = ".aggregate-function"
-    _DEFAULT_AGGREGATE_FUNCTION_KEY = "fields.default-aggregate-function"
-
-    def _partial_update_unsupported_options(self):
-        """Return the set of option keys configured on this table that
-        ``PartialUpdateMergeFunction`` does not yet support. Empty set
-        means we can safely run the simple last-non-null merge.
-        """
-        flagged = set()
-        raw = self.table.options.options.to_map()
-        for key, value in raw.items():
-            if (key in self._PARTIAL_UPDATE_UNSUPPORTED_BOOLEAN_OPTIONS
-                    and self._option_is_truthy(value)):
-                flagged.add(key)
-            elif key == self._DEFAULT_AGGREGATE_FUNCTION_KEY:
-                flagged.add(key)
-            elif key.startswith(self._FIELDS_PREFIX) and (
-                    key.endswith(self._FIELD_SEQUENCE_GROUP_SUFFIX)
-                    or key.endswith(self._FIELD_AGGREGATE_FUNCTION_SUFFIX)):
-                flagged.add(key)
-        return flagged
-
-    @staticmethod
-    def _option_is_truthy(raw):
-        if raw is None:
-            return False
-        if isinstance(raw, bool):
-            return raw
-        if isinstance(raw, str):
-            return raw.strip().lower() in ("true", "1", "yes", "on")
-        return bool(raw)
 
     def create_reader(self) -> RecordReader:
         # Create a dict mapping data file name to deletion file reader method
--- a/paimon-python/pypaimon/read/split_read.py
+++ b/paimon-python/pypaimon/read/split_read.py
@@ -594,11 +594,32 @@ class MergeFileSplitRead(SplitRead):
         ``FIRST_ROW``) used to silently degrade to dedupe — that quietly
         produced wrong data — so we now raise an explicit
         ``NotImplementedError`` instead, until they're ported.
+
+        For ``PARTIAL_UPDATE``, we also refuse to run when the table
+        configures any option whose semantics this port does not yet
+        implement (sequence-group, per-field aggregator overrides,
+        ignore-delete and friends). Without this guard those options
+        would be silently ignored and produce subtly wrong results —
+        the same anti-pattern this PR exists to close.
         """
         engine = self.table.options.merge_engine()
         if engine == MergeEngine.DEDUPLICATE:
             return DeduplicateMergeFunction()
         if engine == MergeEngine.PARTIAL_UPDATE:
+            unsupported = self._partial_update_unsupported_options()
+            if unsupported:
+                raise NotImplementedError(
+                    "merge-engine 'partial-update' is enabled together "
+                    "with options that pypaimon does not yet implement: "
+                    "{}. The supported subset is per-key last-non-null "
+                    "merge with no sequence-group, no per-field "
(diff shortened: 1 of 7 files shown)


paimon-python/pypaimon/read/split_read.py
```diff
@@ -508,7 +485,33 @@ def section_reader_supplier(self, section: List[SortedRun]) -> RecordReader:
...
+        if engine == MergeEngine.PARTIAL_UPDATE:
+            return PartialUpdateMergeFunction(
+                key_arity=len(self.trimmed_primary_key),
+                value_arity=self.value_arity,
+            )
+        raise NotImplementedError(
+            "merge-engine '{}' is not implemented in pypaimon yet "
+            "(supported: deduplicate, partial-update). Use the Java "
+            "client or open an issue to track support.".format(engine.value)
+        )
```
XiaoHongbo-Hope, line 616: The PR says sequence groups and per-field aggregate overrides are out of scope, but
  `_build_merge_function()` still enables `PartialUpdateMergeFunction` when those options are configured. Since the Python merge function ignores them, it can silently return wrong results.

TheR1sing3un: > The PR says sequence groups and per-field aggregate overrides are out of scope, but `_build_merge_function()` still enables `PartialUpdateMergeFunction` when those options are configured. Since the Python merge function ignores them, it can silently return wrong results.

Do you mean it would be best to check again here whether these parameters have been configured? What you said makes sense. I can add some parameters for judgment

XiaoHongbo-Hope:
Could you add coverage for partial-update rows that land in the same data file, e.g. two `write_arrow()`
  calls before a single `prepare_commit()`?  I think this currently still bypasses the new merge-engine dispatch

XiaoHongbo-Hope:
Nit: thanks for adding the coverage. I wonder if you can make it a bit more focused though — some unit and
e2e cases seem to overlap on the same last-non-null semantics, and the unsupported-option cases could probably be table-driven.

TheR1sing3un:
> Could you add coverage for partial-update rows that land in the same data file, e.g. two `write_arrow()` calls before a single `prepare_commit()`? I think this currently still bypasses the new merge-engine dispatch

Thank you for your suggestion. I will pin you again after it's done

TheR1sing3un:
> Nit: thanks for adding the coverage. I wonder if you can make it a bit more focused though — some unit and e2e cases seem to overlap on the same last-non-null semantics, and the unsupported-option cases could probably be table-driven.

Verified -- and the gap is independent of this PR.

  I added two coverage cases per your suggestion (commit 0d58859d0):
  `test_partial_update_two_write_arrows_single_commit` /
  `test_partial_update_three_write_arrows_single_commit`. Both fail, so
  they're marked `unittest.expectedFailure` for now.

  To rule out anything this PR introduced, I ran the same workload on
  plain `origin/master` (b4e54ada3) using the **default `deduplicate`**
  merge engine -- two `write_arrow` calls + one `prepare_commit` with
  the same PK in both batches:

```
  w.write_arrow(pa.Table.from_pylist([{'id': 1, 'a': 'first',  'b': 'old'}], ...))
  w.write_arrow(pa.Table.from_pylist([{'id': 1, 'a': 'second', 'b': 'new'}], ...))
  c.commit(w.prepare_commit())
  # Expected: 1 row -- {'id': 1, 'a': 'second', 'b': 'new'} (dedupe to latest)
  # Actual on master: 2 rows -- {'id': 1, 'a': 'first', ...}, {'id': 1, 'a': 'second', ...}
```

  The flushed data file contains 2 rows for the same PK; on read,
  `_build_split_from_pack` (split_generator.py:99-100) marks any
  single-file PK split as `raw_convertible=True`, which routes through
  the raw-convertible fast path and skips `SortMergeReader`. Even on
  master the dedupe path silently returns both rows -- so what this PR
  exposes is a pre-existing user-facing data-quality issue that affects
  the default PK-table configuration. It's not a regression introduced
  by the partial-update dispatch.

  Root cause split across the write and read paths:

  * `KeyValueDataWriter._merge_data` only does `concat + sort` -- no
    merge function is applied at flush time, so the file holds duplicate
    primary keys (violates the LSM "PK unique within a file" invariant
    that Java enforces in `SortBufferWriteBuffer.forEach` /
    `MergeTreeWriter.flushWriteBuffer`).
  * The read-side `raw_convertible` shortcut assumes that invariant
    holds, so it skips `SortMergeReader` -- where the merge-engine
    dispatch (this PR) and the existing dedupe both live.

  Fixing it requires either (a) an in-memory merge buffer in
  `KeyValueDataWriter` mirroring Java's `SortBufferWriteBuffer` (which
  applies the table's `MergeFunction` during flush -- the proper fix,
  also benefits dedupe/aggregation/first-row), or (b) a stricter
  `raw_convertible` check that proves intra-file PK uniqueness. Both
  are write-path / scan-path restructuring, well outside the read-side
  scope of this PR. I'll open a separate PR for it -- the dedupe data-
  quality issue is severe enough on its own to warrant a dedicated fix
  rather than buried in this merge-engine port.

  The expectedFailure cases here will turn into passing regressions
  once that follow-up lands.

TheR1sing3un:
@XiaoHongbo-Hope Hi, thank you for your point. However, this issue is not strictly related to this pr. It is a general problem. I would rather propose a separate pr to solve it. These two PRS are independent and can be advanced in parallel

TheR1sing3un:
@XiaoHongbo-Hope Let's land this pr. Looking forward to your review!

TheR1sing3un:
@leaves12138 @JingsongLi @XiaoHongbo-Hope Looking forward to your review, lets land it!

paimon-python/pypaimon/read/split_read.py
```diff
@@ -578,7 +585,101 @@ def section_reader_supplier(self, section: List[SortedRun]) -> RecordReader:
                 supplier = partial(self.kv_reader_supplier, file, self.deletion_file_readers.get(file.file_name, None))
                 data_readers.append(supplier)
             readers.append(ConcatRecordReader(data_readers))
-        return SortMergeReaderWithMinHeap(readers, self.table.table_schema)
+        merge_function = self._build_merge_function()
```
XiaoHongbo-Hope, line 588: This guard is bypassed for raw-convertible PK splits because those use `RawFileSplitRead`, not
  `MergeFileSplitRead`.

XiaoHongbo-Hope (changes requested):

XiaoHongbo-Hope:
> @XiaoHongbo-Hope Hi, thank you for your point. However, this issue is not strictly related to this pr. It is a general problem. I would rather propose a separate pr to solve it. These two PRS are independent and can be advanced in parallel

You can clean test in this PR.

paimon-python/pypaimon/read/split_read.py
TheR1sing3un: > This guard is bypassed for raw-convertible PK splits because those use `RawFileSplitRead`, not `MergeFileSplitRead`.

done

TheR1sing3un:
> > @XiaoHongbo-Hope Hi, thank you for your point. However, this issue is not strictly related to this pr. It is a general problem. I would rather propose a separate pr to solve it. These two PRS are independent and can be advanced in parallel
> 
> You can clean test in this PR.

done

TheR1sing3un:
@XiaoHongbo-Hope Hi, thanks for your review, any issue to fix? If not, lets land it.

paimon-python/pypaimon/read/split_read.py
XiaoHongbo-Hope: This fixes merge-engine dispatch for MergeFileSplitRead, but partial-update semantics can still be bypassed when the split is raw_convertible. A same-commit multi-write case still returns duplicate unmerged rows.

XiaoHongbo-Hope:
> @XiaoHongbo-Hope Hi, thanks for your review, any issue to fix? If not, lets land it.

Commented again.

paimon-python/pypaimon/read/split_read.py
TheR1sing3un: >  A same-commit multi-write case still returns duplicate unmerged rows.

Track in this pr：https://github.com/apache/paimon/pull/7759

JingsongLi (commented):
## Review: [python] Implement partial-update merge engine in pypaimon

Overall this is a well-structured PR that fixes a real correctness issue (silent data corruption via deduplicate fallback). The algorithm matches the Java reference, the test coverage is thorough, and the fail-loud strategy for unsupported engines/options is the right approach.

### Correctness

1. **NOT NULL check matches Java** -- `updateNonNullFields` in the Java source raises `IllegalArgumentException` when `!nullables[i]` and the field is null. The Python port faithfully mirrors this.

2. **Result decoupling is correct and important** -- The `get_result()` method builds a fresh tuple and a fresh `KeyValue`, avoiding aliasing with upstream's reused kv. The dedicated test nails this contract down.

3. **Potential edge case with column projection** -- When a NOT NULL field is *not* projected, the merge function won't see it, so its constraint won't fire. This is fine semantically but worth a brief comment for future maintainers.

### Design

4. **Merge function protocol could benefit from a Protocol/ABC** -- `SortMergeReaderWithMinHeap.__init__` types the parameter as `Optional[Any]`. Since the contract (`reset`, `add`, `get_result`) is shared across implementations, a `typing.Protocol` would give IDEs something to latch onto.

5. **`value_fields` on the base `SplitRead`** -- Set in base class but consumed only by `MergeFileSplitRead._build_merge_function`. Consider scoping it to `MergeFileSplitRead.__init__`.

6. **Double `.options` access** -- `raw = table.options.options.to_map()` reads oddly; a one-line comment would reduce confusion.

### Nit

7. In `partial_update_merge_function.py`, the `NotImplementedError` for DELETE/UPDATE_BEFORE constructs `RowKind(row_kind_byte).to_string()`. If an unexpected byte arrives, this raises a secondary `ValueError` masking the original problem. A defensive fallback would be more robust.

Summary: core semantics are sound, guard rails are well-placed, tests are comprehensive.

JingsongLi (approved):
+1

PR #7752, opened 2026-04-30
[python] Fix vector search RecursionError with many index readers

XiaoHongbo-Hope:
### Purpose
Vector search throws `RecursionError: maximum recursion depth exceeded` when there are many index readers. This PR fixes the above issue by changing `or_`/`and_` in `GlobalIndexResult` from lazy to eager evaluation

### Tests
`GlobalIndexTest`


JingsongLi (approved):
+1

PR #7754, opened 2026-05-01
[flink] Fix StackOverflowError when building global index with many index columns and partitions

XiaoHongbo-Hope:


jerry-024:
+1

PR #7757, opened 2026-05-01
[core] Reject dot-paths in fields.* options on row columns

### LINKED ISSUE #6471: [Bug] Paimon ROW Type Internal Field Sum Aggregation Not Working
### Search before asking

### Paimon version

1.2

### Compute Engine

Flink 1.20

### Minimal reproduce step

Hi,
I'm encountering an issue where the sum aggregation on internal fields of a ROW type in Paimon doesn't work as expected. Instead of aggregating values, the data is overwritten (only the last value is retained). Here's a reproducible test case:

1. my DDL
CREATE TABLE "test_row_agg"
(
    "id" INT,
    "data" ROW("num" INT, "info" VARCHAR)
)
WITH (
    options='{
        "merge-engine":"aggregation",
        "fields.data.num.aggregate-function":"sum",
        "fields.data.sequence-group":"data"
    }',
    primary_key=ARRAY['id'],
    bucket=4,
    bucket_key=ARRAY['id']
);
2. Insert Test Data (Two Writes with the Same Primary Key)
INSERT INTO "test_row_agg" ("id", "data")
VALUES (1, ROW(10, 'info1'));

INSERT INTO "test_row_agg" ("id", "data")
VALUES (1, ROW(20, 'info2'));
3.  Query Result
SELECT "id", "data"."num" FROM "test_row_agg" WHERE "id" = 1;

I expected data.num = 30 ,But，Actual Result is 20
why? I create table and config "fields.data.num.aggregate-function":"sum",
But not sum

Expected: (1, 30, 'info2')
Actual: (1, 20, 'info2')

### Anything else?

_No response_

ArnavBalyan:
### Purpose
 - `fields.<row>.<inner>.aggregate-function` is silently ignored at write time, the parent row overwrites instead of aggregating. (during write time)
 - Reject the misconfig at ddl when the parent field is a RowType. Ensure we can strictly reject this case instead of allowing silent overwrites. 
 - Fixes: #6471

### Tests
 - UT

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
+++ b/paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
@@ -487,7 +487,8 @@ public class SchemaValidation {
                 .forEach(
                         k -> {
                             if (k.startsWith(FIELDS_PREFIX)) {
-                                String[] fields = k.split("\\.")[1].split(FIELDS_SEPARATOR);
+                                String[] segments = k.split("\\.");
+                                String[] fields = segments[1].split(FIELDS_SEPARATOR);
                                 for (String field : fields) {
                                     checkArgument(
                                             DEFAULT_AGG_FUNCTION.equals(field)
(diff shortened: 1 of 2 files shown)


ArnavBalyan:
cc @JingsongLi thanks!

JingsongLi (approved):
+1

PR #7759, opened 2026-05-01
[python] In-memory merge buffer for primary-key writer

TheR1sing3un:
## Background

`KeyValueDataWriter._merge_data` only did `concat + sort` before flush — multiple `write_arrow` calls inside one `prepare_commit` could land same-PK rows in a single data file. On the read side, single-snapshot single-file splits are marked `raw_convertible=True` and bypass `SortMergeReader`, so the merge function never runs. Net effect: `deduplicate` / `partial-update` semantics were silently lost whenever a single commit contained multiple writes hitting the same key.

## Solution

`KeyValueDataWriter` now applies the table's `MergeFunction` at flush time:

- `write()` batches accumulate in `pending_data` without sorting or folding (1 sort per flush, not 1 per write).
- On overflow / `prepare_commit` / `close`, `_flush_all` runs `sort → fold (MergeFunction.reset → add* → get_result) → roll-write`. The folded result is split into one or more files each ≤ `target_file_size`, and the buffer is fully drained.
- Each flushed file is PK-unique, restoring the LSM invariant the read-side `raw_convertible` fast path depends on. The read path is unchanged.

`pypaimon/common/merge_engine_dispatch.py` is the single dispatch shared by the read and write sides, so the two cannot drift:

- `deduplicate` / `partial-update` (without sequence-group, per-field aggregator, ignore-delete or remove-record-on-* options) are fully supported.
- `partial-update` paired with any of those options raises `NotImplementedError` on the first flush (matches the read-side guard).
- `aggregation` / `first-row` fall back to deduplicate with a `logging.warning` so the semantic mismatch is visible to downstream readers.

## Tests

- `tests/test_write_merge_buffer.py` — fold algorithm, `_flush_all` lifecycle, `_roll_write` multi-file split, KeyValue pooling.
- `tests/test_partial_update_merge_function.py` — `PartialUpdateMergeFunction` contract incl. NOT-NULL with field names.
- `tests/test_merge_engine_dispatch.py` — option parsing + dispatch routing.
- `tests/test_partial_update_e2e.py` — end-to-end on real PK tables: same-commit multi-write regressions, unsupported-options guards, aggregation fallback warning.


TheR1sing3un:
block on https://github.com/apache/paimon/pull/7745 .

JingsongLi (commented):
## Review: In-memory merge buffer for primary-key writer

Overall this is a well-structured fix for a real correctness issue (silent PK-uniqueness violation on multi-batch writes). The shared dispatch design prevents read/write drift, the tests are thorough, and the scoping choices (fallback to dedupe for unsupported engines, explicit rejection of partial-update with out-of-scope options) are sound. A few observations:

### Correctness

1. **Redundant merge in `prepare_commit`**: Every `write()` call ends with `_check_and_roll_if_needed()` which already merges the buffer. By the time `prepare_commit()` is called, `pending_data` is guaranteed PK-unique. The merge in `prepare_commit` is therefore always a no-op that still pays the `to_pylist()`/`from_pylist()` round-trip cost on the entire remaining buffer. Consider guarding it with a dirty flag or removing it (since `_check_and_roll_if_needed` is the last thing `write()` does).

2. **`_option_is_truthy` accepts `"1"`, `"yes"`, `"on"`**: Java's `Boolean.parseBoolean` only recognizes `"true"` (case-insensitive). If a user has a property set to `"yes"` that Java treats as false, the Python dispatch would reject the write while Java would proceed. This is unlikely in practice but diverges from Java semantics. Suggest aligning to only `"true"` (case-insensitive) for parity.

3. **Type fidelity through `to_pylist()`/`from_pylist()`**: Arrow types with complex representations (timestamps with timezone, Decimal128 with specific precision/scale, durations) can lose precision or change representation when round-tripped through Python dicts. For tables with such column types, the merged data could silently differ from the input. This is an inherent limitation of the approach. Worth a comment or a tracked follow-up to use columnar operations instead.

### Performance

4. **O(n^2) on buffer size (acknowledged)**: Each `write()` re-merges the full buffer via row-by-row `to_pylist` materialization. For streaming workloads with many small appends, this dominates. A straightforward mitigation would be to maintain an append-only list of `pa.Table` batches and only merge (once) at flush time -- the same approach Java takes with `SortBufferWriteBuffer` only folding on `forEach`.

5. **Single-row fast path is good**: Skipping the merge for `n < 2` avoids the materialization overhead in the common streaming case. Consider extending this to also skip when the buffer has no duplicate PKs (a single `pc.value_counts` on the PK column could detect this cheaply in Arrow without materializing to Python).

### Design

6. **Writer fallback for `aggregation`/`first-row`**: Falling back to deduplicate on the write side means files on disk were processed with wrong-engine semantics. If the read-side port is ever added, those historical files would produce incorrect results unless compacted. A `logging.warning` at the fallback site would make this visible in production.

7. **`DeduplicateMergeFunction` imported from `pypaimon.read.reader.*` by the write path**: The module path suggests this is a read-side concern. Since it is now shared infrastructure, consider relocating `DeduplicateMergeFunction` and `PartialUpdateMergeFunction` into `pypaimon/common/merge_functions/` (or next to `merge_engine_dispatch.py`) to clarify ownership. Not blocking, just organizational.

### Minor

- `_merge_pending_by_pk` constructs a fresh `KeyValue` per row, but `KeyValue.__init__` allocates two `OffsetRow` objects each time. For large buffers with many PKs, a pooled/reused `KeyValue` (reset via `replace`) would reduce GC pressure.
- The `PartialUpdateMergeFunction` NOT-NULL check raises `ValueError` on the first `add()` if the initial write has a null for a non-nullable field. This matches Java, but users of partial-update tables may find it unintuitive that partially-written fields must be declared nullable. A note in the error message suggesting this would improve DX.

Good work closing the silent data-quality gap. The core algorithm is correct and well-tested.

TheR1sing3un:
> Good work closing the silent data-quality gap. The core algorithm is correct and well-tested.

Some of the issues you mentioned have been fixed. Please review them again~ Thanks !

TheR1sing3un:
@JingsongLi Hi, could u review again, thanks!

TheR1sing3un:
@JingsongLi @leaves12138 Hi, any suggestions? Lets land it! Thanks!

TheR1sing3un:
first-row was not supported on master when this PR was opened; now that it has landed on master, this PR also adapts the write-side merge buffer to support first-row.

Ready for review again, cc @JingsongLi @leaves12138

JingsongLi (commented):
LGTM, thanks.

PR #7762, opened 2026-05-03
[python] Add end-to-end SQL tests for Paimon system tables

plusplusjiajia:
### Purpose
Add pypaimon/tests/sql_system_table_test.py covering the <table>$<system_name> syntax via SQLContext for $options, $schemas, $snapshots, $tags, $branches, $manifests.


plusplusjiajia:
CI builds pypaimon-rust from paimon-rust main, which doesn't yet include the FFI fix in apache/paimon-rust#295. Will re-trigger here once that merges.

JingsongLi:
+1

PR #7764, opened 2026-05-03
[iceberg] Fix ClassCastException reading timestamp with local timezone columns

### LINKED ISSUE #7763: [Bug] Fix ClassCastException reading TIMESTAMP_WITH_LOCAL_TIME_ZONE columns
### Search before asking

- Reading a Paimon timestamp with local timezone column from the Iceberg manifest throws ClassCastException.  
- In IcebergConversions.toPaimonObject the two timestamp variants share a switch arm and cast the data type
   to TimestampType, but LocalZonedTimestampType is not a subtype of TimestampType, so the cast blows up at runtime.
 - The write side already splits the two cases and casts each one correctly. 

### Anything else?

_No response_

ArnavBalyan:
### Purpose
   - Reading a timestamp with local timezone column from the Iceberg manifest throws ClassCastException. In IcebergConversions.toPaimonObject the two timestamp variants share a switch arm that casts the data type to TimestampType, but LocalZonedTimestampType is not a subtype of TimestampType.
   - Split the arm so each case casts to its own type. Write side was already doing this.
   - Fixes: #7763 

### Tests
 - UT

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/iceberg/manifest/IcebergConversions.java
+++ b/paimon-core/src/main/java/org/apache/paimon/iceberg/manifest/IcebergConversions.java
@@ -126,6 +126,14 @@ public class IcebergConversions {
                 .putLong(0, timestamp.toMicros());
     }
 
+    private static Timestamp timestampFromBytes(byte[] bytes, int precision) {
+        Preconditions.checkArgument(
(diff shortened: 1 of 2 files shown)


ArnavBalyan:
cc @JingsongLi thanks!

JingsongLi:
+1

PR #7765, opened 2026-05-03
[core] Add eager failure for fallback branch read failures

### LINKED ISSUE #7761: [Bug] Fallback branch read silently swallows exceptions with no fail-fast option
### Search before asking

### Paimon version

master

### Compute Engine

all

### Minimal reproduce step

.

### Description

When reading from a fallback branch in `FallbackReadFileStoreTable`, if the read operation fails, the exception is silently swallowed:

```java
} catch (Exception ignored) {
    LOG.error(
            "Reading from supplemental branch has problems: {}",
            fallbackSplit.wrapped());
}
```

This has two problems:

1. **No fail-fast option**: There is no way to let the read operation fail fast when the fallback branch is unavailable. The reader always silently falls back to the current branch, which can lead to incorrect or incomplete query results without any warning to the user.

2. **Loss of exception stack trace**: The `LOG.error` call does not include the exception object as a parameter, so the log only contains the split description but not the actual cause of the failure, making it very difficult to diagnose issues.

### Expected Behavior

- Add a config option `scan.fallback-branch.read-fail-fast` (default `false`) to allow users to opt into failing fast when the fallback branch read fails.
- When `fail-fast=false` (default, preserving backward compatibility), log the full exception stack trace so that failures are observable.
- When `fail-fast=true`, propagate the exception to the caller so the read operation fails immediately.

### Anything else?

_No response_

ArnavBalyan:
### Purpose
   - When reading from a fallback branch fails, the exception is swallowed and the reader falls through to the current branch. This can mask data issues, and the LOG.error does not include the exception object so the stack trace was lost.
   - Added `scan.fallback-branch.read-fail-fast` (default false). When enabled the exception is propagated to the caller so the read fails immediately. When disabled the failure is logged with the full stack trace and the reader still falls through to the current branch.
   - Fixes: #7761

### Tests
 - UT

### DIFF
--- a/docs/layouts/shortcodes/generated/core_configuration.html
+++ b/docs/layouts/shortcodes/generated/core_configuration.html
@@ -1188,6 +1188,12 @@ This config option does not affect the default filesystem metastore.</td>
             <td>When a batch job queries from a table, if a partition does not exist in the current branch, the reader will try to get this partition from this fallback branch.</td>
         </tr>
         <tr>
+            <td><h5>scan.fallback-branch.read-fail-fast</h5></td>
+            <td style="word-wrap: break-word;">false</td>
+            <td>Boolean</td>
+            <td>Whether to fail the read immediately when reading from a fallback branch throws. By default the failure is logged with the full stack trace and the reader falls through to the current branch, which can mask data issues. Set this to true to surface fallback branch errors to the caller instead.</td>
+        </tr>
+        <tr>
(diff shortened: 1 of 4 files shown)


ArnavBalyan:
cc @JingsongLi can you pls take a look thanks!

JingsongLi:
Please update docs. See paimon-docs/README

ArnavBalyan:
updated thanks

JingsongLi (approved):
+1

PR #7767, opened 2026-05-04
[iceberg] Fix UnsupportedOperationException reading tinyint and smallint columns

ArnavBalyan:
### Purpose
   - Iceberg toByteBuffer encodes tiny and small int as 4 byte little endian int, but IcebergConversions.toPaimonObject has no matching case and throws UnsupportedOperationException.
   - Occurs when the type is used as a partition column and decoded back into Paimon objects.
   - Add support for the same.

### Tests
 - Unit tests


ArnavBalyan:
cc @JingsongLi thanks! :)

JingsongLi:
+1

Issue #7768, opened 2026-05-04
[Feature] PyArrowFileIO handling for Google Cloud Storage


larssk:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Motivation

Use Google Cloud Storage as warehouse / object storage for pypaimon (without JDK dependencies).

### Solution

Add scheme `gs` to `PyArrowFileIO`, and use `pyarrow.fs.GcsFileSystem` from the `pyarrow` package to support GCS.  

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #7769, opened 2026-05-05
[python] Use pypaimon with Google Cloud Storage

larssk:
### Purpose

Use Google Cloud Storage (GCS) as warehouse / object storage for pypaimon (without java dependencies). This PR adds PyArrowFileIO handling for GCS.

#### 1. Scheme dispatch (`pyarrow_file_io.py`)

Added a `gs` branch in `__init__` alongside the existing `s3` and `hdfs` branches:

```python
elif scheme == "gs":
    self.filesystem = self._initialize_gcs_fs()
```

#### 2. `_initialize_gcs_fs()` method (`pyarrow_file_io.py`)

Uses `pyarrow.fs.GcsFileSystem`, which is already available via `pyarrow[gcs]`.

With no arguments, `GcsFileSystem` picks up credentials automatically via Application Default Credentials (ADC): `GOOGLE_APPLICATION_CREDENTIALS`, the GCP metadata server, or Workload Identity on GKE/GCE. Three optional properties allow explicit credential passing:

| Property | Description |
|---|---|
| `gcs.access-token` | OAuth2 access token. If unset, ADC is used. |
| `gcs.access-token.expiration` | ISO 8601 expiry datetime for the token. |
| `gcs.project-id` | GCP project ID for billing/quota. |

```python
def _initialize_gcs_fs(self) -> FileSystem:
    access_token = self._get_property("gcs.access-token")
    token_expiry = self._get_property("gcs.access-token.expiration")
    project_id = self._get_property("gcs.project-id")

    kwargs = {}
    if access_token:
        from datetime import datetime
        kwargs["access_token"] = access_token
        kwargs["credential_token_expiration"] = (
            datetime.fromisoformat(token_expiry) if token_expiry
            else datetime(9999, 12, 31)
        )
    if project_id:
        kwargs["project_id"] = project_id

    # With no kwargs, GcsFileSystem uses ADC automatically
    return pafs.GcsFileSystem(**kwargs)
```

#### 3. Path fix in `to_filesystem_path()` (`pyarrow_file_io.py`)

`to_filesystem_path()` had a catch-all for non-S3/non-HDFS schemes that returned only `uri.path` — e.g. for `gs://my-bucket/data/table` it returned `/data/table`, stripping the bucket name. `GcsFileSystem` (like `S3FileSystem`) expects paths in `bucket/key` form without a leading slash.

Added a `GcsFileSystem` branch mirroring the existing S3 logic:

```python
from pyarrow.fs import GcsFileSystem
if isinstance(self.filesystem, GcsFileSystem):
    if parsed.scheme and parsed.netloc:
        path_part = normalized_path.lstrip('/')
        return f"{parsed.netloc}/{path_part}" if path_part else parsed.netloc
    return str(path)
```

#### 4. `GcsOptions` class (`config.py`)

Added a `GcsOptions` class alongside the existing `S3Options` documenting the three new properties.

```
class GcsOptions:
    GCS_ACCESS_TOKEN = ConfigOptions.key("gcs.access-token").string_type().no_default_value().with_description(
        "GCS access token. If not set, ADC (Application Default Credentials) is used automatically.")
    ...
```

### Linked issue

https://github.com/apache/paimon/issues/7768

#### Unit tests — `pypaimon/tests/file_io_test.py`

A new `GCSFileIOPathTest` class is added to the existing `file_io_test.py`, following the same pattern as the existing S3/OSS path conversion tests. These tests require no GCS credentials.

| Test | What it checks |
|---|---|
| `test_gcs_filesystem_type` | `PyArrowFileIO("gs://...")` produces a `pafs.GcsFileSystem` instance |
| `test_gcs_path_conversion` | `gs://bucket/key` maps to `bucket/key` (bucket prepended, no leading slash) |
| `test_gcs_path_bucket_only` | `gs://bucket` with no path component maps to `bucket` |
| `test_gcs_path_normalization` | Consecutive slashes in the path are collapsed (e.g. `gs://bucket///a///b` → `bucket/a/b`) |
| `test_gcs_path_idempotency` | Already-converted `bucket/key` paths pass through unchanged |
| `test_gcs_path_no_leading_slash` | `to_filesystem_path` never returns a path starting with `/` for any GCS URI |

```bash
python -m pytest pypaimon/tests/file_io_test.py::GCSFileIOPathTest -v
```

#### Integration tests — `pypaimon/tests/gcs_file_io_test.py`

A new `GCSFileIOTest` class is added following the pattern of `oss_file_io_test.py`. All tests are skipped automatically when `GCS_TEST_BUCKET` is not set. Credentials are picked up via ADC — no explicit credential properties are required.

| Test | What it checks |
|---|---|
| `test_gcs_filesystem_type` | Live `PyArrowFileIO` with `gs://` uses `GcsFileSystem` |
| `test_exists` | `exists()` returns `False` for non-existent paths; `get_file_status()` raises `FileNotFoundError` |
| `test_write_and_read_file` | `write_file()` / `read_file_utf8()` round-trip |
| `test_write_file_overwrite` | `write_file(..., overwrite=False)` raises `FileExistsError`; `overwrite=True` replaces content |
| `test_new_input_stream_read` | `new_output_stream()` / `new_input_stream()` binary round-trip; `FileNotFoundError` for missing file |
| `test_get_file_status_directory` | `get_file_status()` returns `FileType.Directory` for a directory |
| `test_get_file_status_file` | `get_file_status()` returns `FileType.File` with non-None size |
| `test_delete_returns_false_when_not_exists` | `delete()` returns `False` for non-existent file/directory |
| `test_delete_non_empty_directory_raises_error` | `delete(..., recursive=False)` raises `OSError` for non-empty directory |
| `test_rename_returns_false_when_dst_exists` | `rename()` returns `False` when destination already exists |
| `test_copy_file` | `copy_file(..., overwrite=False)` raises `FileExistsError`; `overwrite=True` replaces content |
| `test_try_to_write_atomic` | `try_to_write_atomic()` writes content and returns `True` on success |
| `test_mkdirs_raises_error_when_path_is_file` | `mkdirs()` raises `FileExistsError` when path is an existing file |

```bash
# Skips all tests (no bucket configured)
python -m pytest pypaimon/tests/gcs_file_io_test.py -v

# Runs all tests against a real GCS bucket using ADC
GCS_TEST_BUCKET=my-bucket python -m pytest pypaimon/tests/gcs_file_io_test.py -v
```

### DIFF
--- a/paimon-python/pypaimon/common/options/config.py
+++ b/paimon-python/pypaimon/common/options/config.py
@@ -43,6 +43,15 @@ class S3Options:
     S3_REGION = ConfigOptions.key("fs.s3.region").string_type().no_default_value().with_description("S3 region")
 
 
+class GcsOptions:
+    GCS_ACCESS_TOKEN = ConfigOptions.key("gcs.access-token").string_type().no_default_value().with_description(
+        "GCS access token. If not set, ADC (Application Default Credentials) is used automatically.")
+    GCS_ACCESS_TOKEN_EXPIRATION = ConfigOptions.key("gcs.access-token.expiration").string_type().no_default_value().with_description(
+        "ISO 8601 expiration datetime for the GCS access token. Required when gcs.access-token is set.")
+    GCS_PROJECT_ID = ConfigOptions.key("gcs.project-id").string_type().no_default_value().with_description(
+        "GCP project ID for GCS requests.")
+
+
 class PVFSOptions:
     CACHE_ENABLED = ConfigOptions.key("cache-enabled").boolean_type().default_value("true").with_description(
         "Enable cache")
--- a/paimon-python/pypaimon/common/options/config.py
+++ b/paimon-python/pypaimon/common/options/config.py
@@ -44,12 +44,19 @@ class S3Options:
 
 
 class GcsOptions:
-    GCS_ACCESS_TOKEN = ConfigOptions.key("gcs.access-token").string_type().no_default_value().with_description(
-        "GCS access token. If not set, ADC (Application Default Credentials) is used automatically.")
-    GCS_ACCESS_TOKEN_EXPIRATION = ConfigOptions.key("gcs.access-token.expiration").string_type().no_default_value().with_description(
-        "ISO 8601 expiration datetime for the GCS access token. Required when gcs.access-token is set.")
-    GCS_PROJECT_ID = ConfigOptions.key("gcs.project-id").string_type().no_default_value().with_description(
-        "GCP project ID for GCS requests.")
+    GCS_ACCESS_TOKEN = (
+        ConfigOptions.key("gcs.access-token").string_type().no_default_value()
+        .with_description(
+            "GCS access token. If not set, ADC (Application Default Credentials) is used "
+            "automatically."))
+    GCS_ACCESS_TOKEN_EXPIRATION = (
+        ConfigOptions.key("gcs.access-token.expiration").string_type().no_default_value()
+        .with_description(
+            "ISO 8601 expiration datetime for the GCS access token. "
+            "Required when gcs.access-token is set."))
+    GCS_PROJECT_ID = (
+        ConfigOptions.key("gcs.project-id").string_type().no_default_value()
+        .with_description("GCP project ID for GCS requests."))
 
 
 class PVFSOptions:
--- a/paimon-python/pypaimon/filesystem/pyarrow_file_io.py
+++ b/paimon-python/pypaimon/filesystem/pyarrow_file_io.py
@@ -78,6 +78,8 @@ class PyArrowFileIO(FileIO):
             self.filesystem = self._initialize_s3_fs()
         elif scheme in {"hdfs", "viewfs"}:
             self.filesystem = self._initialize_hdfs_fs(scheme, netloc)
+        elif scheme == "gs":
+            self.filesystem = self._initialize_gcs_fs()
         else:
             raise ValueError(f"Unrecognized filesystem type in URI: {scheme}")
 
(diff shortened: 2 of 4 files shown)


JingsongLi (commented):
## Review: [python] Use pypaimon with Google Cloud Storage

Overall this is a clean, well-structured addition that follows the existing S3/HDFS patterns closely. Tests are comprehensive (both unit and integration). A few observations:

### 1. Timezone-naive fallback datetime (potential bug)

In `_initialize_gcs_fs()`:

```python
kwargs["credential_token_expiration"] = (
    datetime.fromisoformat(token_expiry) if token_expiry
    else datetime(9999, 12, 31)
)
```

The fallback `datetime(9999, 12, 31)` is timezone-naive. If `datetime.fromisoformat(token_expiry)` returns a timezone-aware datetime (e.g., input is `"2025-12-31T00:00:00+00:00"`), PyArrow may reject mixing aware and naive datetimes, or comparisons downstream may raise `TypeError`. Consider:

```python
from datetime import datetime, timezone
# ...
else datetime(9999, 12, 31, tzinfo=timezone.utc)
```

### 2. `GCS_ACCESS_TOKEN_EXPIRATION` description is misleading

The option description states *"Required when gcs.access-token is set"*, but the code silently falls back to a far-future expiry when the value is absent. Either:
- Enforce the requirement by raising `ValueError` when `access_token` is set but `token_expiry` is not, or
- Update the description to say something like *"If not set when gcs.access-token is provided, a far-future expiry is assumed."*

### 3. Minor: `to_filesystem_path` – scheme-without-netloc edge case

The S3 path handling has an explicit branch for `parsed.scheme and not parsed.netloc`:

```python
# S3 version
if parsed.scheme:
    if parsed.netloc:
        ...
    else:
        result = normalized_path.lstrip('/')
        return result if result else '.'
```

The GCS block collapses this into a single `return str(path)` for anything that doesn't match `scheme and netloc`. If a bare `gs:some/relative/path` is ever passed (unlikely but possible), it would be returned as-is instead of being cleaned up. For defensive consistency, consider mirroring the S3 structure:

```python
if isinstance(self.filesystem, GcsFileSystem):
    if parsed.scheme and parsed.netloc:
        path_part = normalized_path.lstrip('/')
        return f"{parsed.netloc}/{path_part}" if path_part else parsed.netloc
    elif parsed.scheme:
        result = normalized_path.lstrip('/')
        return result if result else '.'
    return str(path)
```

### 4. Positive notes

- The `GcsOptions` class and property keys follow the existing `S3Options` convention well.
- Integration tests in `gcs_file_io_test.py` are thorough and gracefully skip when no bucket is configured.
- Path normalization (collapsing multiple slashes) and the idempotency test are good defensive additions.
- The `to_filesystem_path` placement after the S3 block and before the generic fallback is correct.

Nice contribution overall — just the timezone issue (point 1) seems worth fixing before merge.

JingsongLi:
Please rebase master.

JingsongLi:
+1

PR #7771, opened 2026-05-06
[python] Add compaction module with Ray distributed executor

TheR1sing3un:
## Summary

Brings Apache Paimon's compaction story to pypaimon end-to-end:

- **Phase 1**: Extend `CommitMessage` with `compact_before` / `compact_after`; `FileStoreCommit` emits ADD + DELETE manifest entries and a new `commit_compact()` helper produces snapshots with `commit_kind=COMPACT`. `DataFileMeta` gains JSON-friendly `to_dict` / `from_dict` and a `CommitMessageSerializer` for cross-process transport.
- **Phase 2**: Append-only compaction end-to-end via `table.new_compact_job(...).execute()`, plumbed through a `Coordinator → Task → Executor → Driver-commit` shape that mirrors Spark `CompactProcedure`. Ships a `LocalExecutor` for in-process / test usage.
- **Phase 3**: Primary-key (merge-tree) compaction. Direct port of `Levels` + `UniversalCompaction` (size-amp / size-ratio / file-num three-stage decision). New abstract `MergeFunction` + factory; `DeduplicateMergeFunction` migrated, `PartialUpdate` / `Aggregate` / `FirstRow` stubbed so configured tables fail loudly with a Phase 6 message.
- **Phase 4**: `RayExecutor` wires the same `CompactJob` to Ray. Driver serializes each `CompactTask` to JSON (table + payload + catalog loader spec), workers rebuild their `FileStoreTable` via the catalog and run the rewriter, driver collects messages for one atomic commit.

Each phase landed as a separate commit, with a follow-up `*-fixup` commit addressing the review findings inline. Eight commits total — a single PR keeps the design coherent for review while commits stay small enough to walk through.

Out of scope (later PRs):

- Phase 5 — CLI + `python -m pypaimon.compact.entrypoint` for `ray job submit`
- Phase 6 — Real `PartialUpdate` / `Aggregate` / `FirstRow` `MergeFunction` bodies
- Phase 7 — Sort compact (zorder/hilbert), Deletion Vectors, Changelog producer, streaming Ray-Actor coordinator

Plan / design doc: \`/Users/lcy/.claude/plans/paimon-compaction-java-spark-python-com-cached-cosmos.md\` (local).

## Test plan

- [x] Unit: \`Levels\` semantics, \`UniversalCompaction\` three-stage decision (size-amp, size-ratio, file-num, force-pick-L0), \`MergeFunction\` registry + Phase-6 stubs, \`CompactOptions\` validation, \`CommitMessageSerializer\` JSON round-trip, \`CompactTask\` serde with non-JSON-native partitions
- [x] Append-only e2e: file-count drops, data identity preserved, \`snapshot.commit_kind == "COMPACT"\`, partitioned per-partition messages, no-op on below-trigger
- [x] Primary-key e2e: full-compaction dedup keeps latest, level promotion to >0, no-op below trigger
- [x] Rewriter contracts: does not mutate manifest-owned \`DataFileMeta\`, aborts partial output on failure
- [x] Ray e2e (skipped if \`ray\` not installed): \`ray.init(local_mode=True)\` runs Append-only compaction end-to-end through real \`ray.remote\` task dispatch + \`ray.get\` collection
- [ ] Integration with a Java-written table (manual; left for follow-up before un-drafting)

### DIFF
--- a/paimon-python/pypaimon/compact/__init__.py
+++ b/paimon-python/pypaimon/compact/__init__.py
@@ -1,0 +1,17 @@
+################################################################################
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+# limitations under the License.
+################################################################################
--- a/paimon-python/pypaimon/compact/coordinator/__init__.py
+++ b/paimon-python/pypaimon/compact/coordinator/__init__.py
@@ -1,0 +1,17 @@
+################################################################################
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+# limitations under the License.
+################################################################################
--- a/paimon-python/pypaimon/compact/coordinator/append_compact_coordinator.py
+++ b/paimon-python/pypaimon/compact/coordinator/append_compact_coordinator.py
@@ -108,8 +108,13 @@ class AppendCompactCoordinator(CompactCoordinator):
         # if the executor later relies on file ordering for something.
         candidates.sort(key=lambda f: f.min_sequence_number)
 
+        # NOTE: under non-full compaction, the trailing chunk may have fewer
+        # than min_file_num files and is dropped here — those files stay live
+        # and will be picked up next time more small files accumulate. This is
+        # a deliberate trade-off (vs. Java, which always picks them up); a
+        # future change can revisit it.
         chunks: List[List[DataFileMeta]] = []
-        max_per_task = max(self.options.max_file_num, self.options.min_file_num)
+        max_per_task = self.options.max_file_num
         for start in range(0, len(candidates), max_per_task):
             chunk = candidates[start:start + max_per_task]
             if len(chunk) >= self.options.min_file_num or self.options.full_compaction:
--- a/paimon-python/pypaimon/compact/coordinator/append_compact_coordinator.py
+++ b/paimon-python/pypaimon/compact/coordinator/append_compact_coordinator.py
@@ -30,16 +30,29 @@ from pypaimon.read.scanner.file_scanner import FileScanner
 class AppendCompactCoordinator(CompactCoordinator):
     """Plan compaction tasks for append-only tables (HASH_FIXED or BUCKET_UNAWARE).
 
-    For each (partition, bucket) we compact files smaller than the table's
-    target_file_size. A bucket is eligible only when it has at least
-    `min_file_num` such files (default 5) — matching the Java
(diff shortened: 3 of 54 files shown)


TheR1sing3un:
This pr is for tracking the full of work about compaction. I will initiate and apply for reviews based on the minimum separable pr

JingsongLi (commented):
## Review: [python] Add compaction module with Ray distributed executor

This is a substantial, well-structured contribution (5k+ lines) that brings end-to-end compaction to pypaimon through a clean Coordinator -> Task -> Executor -> Driver-commit architecture that mirrors Spark's CompactProcedure. The phased approach (append-only, then PK merge-tree, then Ray distribution) keeps the design coherent. Below are findings grouped by category.

---

### Correctness

1. **`UniversalCompaction._pick_for_size_ratio_with_count` uses floating-point arithmetic**
   ```python
   if candidate_size * (100.0 + self.size_ratio) / 100.0 < nxt.total_size():
   ```
   Java uses integer arithmetic (`candidateSize * (100 + sizeRatio) / 100`). For large file sizes (petabyte-scale warehouses), floating-point precision loss could cause a different pick decision than Java. Consider:
   ```python
   if candidate_size * (100 + self.size_ratio) < nxt.total_size() * 100:
   ```
   This avoids division entirely and keeps everything in integer domain.

2. **`RayExecutor`: no cleanup of orphan files on partial failure**
   If one Ray task fails, `ray.get(futures)` raises the first exception, but other tasks may have already written output files. Unlike `LocalExecutor` where the rewriter's `abort()` cleans up, there is no mechanism to abort the remaining tasks or clean their partial output. Consider wrapping `ray.get` with `ray.cancel` on the remaining futures in a try/except, or document this as a known limitation that relies on snapshot consistency (orphan files never become visible because the commit never happens).

3. **`CompactTask` class-level mutable defaults**
   ```python
   _catalog_loader_options: Optional[Dict[str, str]] = None
   _table_identifier: Optional[str] = None
   ```
   These are class variables, not instance variables. While `self._catalog_loader_options = ...` in `with_table_loader()` correctly creates an instance attribute that shadows the class variable, the pattern is fragile. If a subclass ever reads via `cls._catalog_loader_options` or forgets `self.`, it would see stale state. Consider initializing these in an `__init__` or `__init_subclass__`.

4. **`Levels.non_empty_highest_level()` edge case**
   Returns `0` when only L0 has files, `-1` when everything is empty. `_should_drop_delete` does `unit.output_level >= levels.non_empty_highest_level()`. If `non_empty_highest_level()` returns -1, drop_delete would be True for any output_level. This path is unreachable (no files = no compaction), but a defensive guard or assertion would clarify intent.

---

### Design

5. **`CommitMessage` breaking change is well-handled**
   Converting `new_files` from a constructor field to a `@property` delegating to `data_increment.new_files` is a clean migration. All internal callers are updated. Consider adding a deprecation note in the class docstring for any external pypaimon users who might construct `CommitMessage` directly.

6. **`CompactRewriter` as a marker base class**
   The docstring explains why there is no abstract `rewrite()` method (different signatures for append vs merge-tree). This is reasonable, but an alternative would be to make it a `Protocol` or simply remove the base class entirely, since it provides no contract enforcement. Not blocking, just a style observation.

7. **Ray worker registry bootstrapping via side-effect imports**
   ```python
   from pypaimon.compact.task import append_compact_task as _append_task  # noqa: F401
   from pypaimon.compact.task import merge_tree_compact_task as _mt_task  # noqa: F401
   ```
   This relies on `ray_executor.py` being imported in the worker process so the task subclasses register themselves. If Ray ever changes its serialization strategy (e.g., cloudpickle the function without importing the module), the registry would be empty on the worker. A safer pattern is to have `CompactTask.deserialize()` perform a lazy import based on the `type` field, eliminating the need for a global registry. That said, the current approach works with Ray's actual behavior today.

8. **`build_kv_file_fields` centralization** -- Good refactoring. Extracting this from `_create_key_value_fields` into a shared utility prevents schema drift between the read and compact paths.

---

### Minor / Nits

9. **`_level0_compare` treats `None` creation_time as smallest** -- Java's TreeSet comparator for L0 uses `Long.compare(creationTime, ...)` which would NPE on null. If Java never stores null creation times this is fine, but worth a comment explaining the divergence.

10. **`MergeTreeRollingWriter._extract_seq_bounds`** -- If `data` is empty (0 rows), `pc.min(seq).as_py()` returns `None`, which would fail downstream. The caller guards with `if data.num_rows == 0: return` but the static method itself is unguarded.

11. **Test `_make_coord` bypasses `__init__`** via `AppendCompactCoordinator.__new__(...)` -- This is fine for unit-testing the pure algorithm but is brittle if the constructor ever adds required initialization. Consider a lightweight mock/stub table instead.

---

### Summary

Overall this is a high-quality contribution with solid test coverage (unit, integration, e2e, and Ray). The architecture cleanly separates planning from execution, and the commit protocol correctly mirrors Java's ADD/DELETE manifest entry semantics. The main actionable items are (1) switching the size-ratio comparison to integer arithmetic for Java parity, (2) documenting or addressing the Ray partial-failure orphan-file scenario, and (3) converting class-level loader fields to instance variables in `CompactTask`.

JingsongLi:
Super big, split this to multiple PRs.

PR #7772, opened 2026-05-06
[iceberg] Support TagCallback.notifyCreation without snapshot id

ArnavBalyan:
### Purpose
 - notifyCreation(String) throws UnsupportedOperationException breaking the TagCallback contract.
 - The snapshot id is available to the iceberg compat layer, TagManager persists the tag file before invoking a callback.
 - Ensure we can support Tagmanager notify creation based on the tag id.

### Tests
 - UT


ArnavBalyan:
cc @JingsongLi thanks :)

JingsongLi:
+1

PR #7776, opened 2026-05-07
[spark] Preserve table options for RTAS saveAsTable overwrite.

### LINKED ISSUE #7687: [Bug] Saving dataframe in overwrite mode using Spark doesn't set primary key
### Search before asking

### Paimon version

3.4

### Compute Engine

Spark

### Minimal reproduce step

Saving dataframe with append mode is correctly setting the primary key in the schema:
Image

Schema:
Image

But when I try to save dataframe with overwrite mode then primary key is NOT set in the schema:
Image

Schema:
Image

Primary key in is NOT set in paimon table schema

### Anything else?

_No response_

Stephen0421:
### Purpose
fix #7687 

Route ReplaceTableAsSelect through the Paimon CTAS strategy so table
options such as primary-key are split from write options and applied to
the created table. Adapt RTAS cache invalidation across Spark 3.5 and
Spark 4.x callback signatures.

### Tests
Add DataFrameWrite tests for overwrite saveAsTable preserving primary
keys and table options.


Zouxxyy:
We are refactoring RTAS to prevent it from dropping tables and to ensure that snapshots are retained

Stephen0421:
> We are refactoring RTAS to prevent it from dropping tables and to ensure that snapshots are retained

OK

JingsongLi:
See https://github.com/apache/paimon/pull/7860

PR #7778, opened 2026-05-07
[core] Parallelize manifest full compaction reads

leaves12138:
### What changed

- Parallelize manifest reads during manifest full compaction using the existing `ManifestReadThreadPool` and `scan.manifest.parallelism`.
- Keep compaction output ordering stable by reading in parallel batches and writing results sequentially.
- Add a regression test that blocks manifest reads until two readers enter concurrently.

### Why

Commit full compaction previously read each manifest file serially in the second read pass. On object storage or Jindo-backed file systems, one slow manifest read could hold the global committer for a long time. This change lets the existing manifest read parallelism speed up that IO-bound phase.

### Validation

- `mvn -pl paimon-core -Pfast-build -Dtest=ManifestFileMetaTest test`
- `mvn -pl paimon-core -DskipTests compile`


XiaoHongbo-Hope:
+1

XiaoHongbo-Hope (approved):

PR #7779, opened 2026-05-07
[core] fix inaccurate record count in snapshot for data-evolution tables

steFaiz:
### Purpose
For data evolution table, record count in snapshot should have different computation logic.

### Tests
See `org.apache.paimon.table.DataEvolutionTableTest`


JingsongLi:
The recordCount in the snapshot should be calculated strictly according to the file and should not be merged. Additionally, the primary key table has not been merged.

steFaiz:
> The recordCount in the snapshot should be calculated strictly according to the file and should not be merged. Additionally, the primary key table has not been merged.

Thanks! I got that!

PR #7780, opened 2026-05-07
[infra] Add stale PR policy via GitHub Actions

ArnavBalyan:
### Purpose
 - Adds an actions/stale workflow, give warning and mark stale after 60d of inactivity.
 - Currently only a comment is applied and PR marked as stale, does not auto close, can be considered in future.
 - Runs every 1 day.

### Tests
 - N/A


JingsongLi (approved):
+1

PR #7785, opened 2026-05-08
[core] Expose total_buckets and done columns in PartitionsTable system table

plusplusjiajia:
### Purpose
Add two new columns to the $partitions system table:
    
    - num_buckets (INT): number of buckets in the partition, sourced from PartitionStatistics.totalBuckets()
    
    - done (BOOLEAN): whether the partition has been marked done, sourced from Partition.done()


JingsongLi (approved):
+1

PR #7786, opened 2026-05-08
[codex] Support catalog-aware blob view function

leaves12138:
## Summary

- Inject Paimon catalog context into catalog-aware Flink functions.
- Make `sys.blob_view` resolve blob fields by name through the catalog while keeping the legacy `(identifier, fieldId, rowId)` overload.

## Root Cause

The stash conflicted with an upstream BlobViewStruct change that now stores table identifiers as `Identifier`. The catalog-aware BlobView path also needed the Flink catalog context during function instantiation so it can resolve field IDs from table metadata at runtime.

## Validation

- `mvn -pl paimon-flink/paimon-flink-common -am -Pfast-build -DskipTests compile`
- `mvn -pl paimon-flink/paimon-flink-common -DfailIfNoTests=false -Dtest=BlobTableITCase#testWriteBlobViewWithBuiltInFunction,BTreeIndexTopoBuilderTest test`

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/function/BlobViewFunction.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/function/BlobViewFunction.java
@@ -18,18 +18,168 @@
 
 package org.apache.paimon.flink.function;
 
+import org.apache.paimon.catalog.Catalog;
+import org.apache.paimon.catalog.CatalogContext;
+import org.apache.paimon.catalog.CatalogFactory;
 import org.apache.paimon.catalog.Identifier;
 import org.apache.paimon.data.BlobViewStruct;
+import org.apache.paimon.flink.FlinkFileIOLoader;
+import org.apache.paimon.options.Options;
+import org.apache.paimon.table.Table;
+import org.apache.paimon.types.DataField;
+import org.apache.paimon.types.DataTypeRoot;
 
+import org.apache.flink.table.functions.FunctionContext;
 import org.apache.flink.table.functions.ScalarFunction;
 
+import javax.annotation.Nullable;
+
+import java.util.HashMap;
+import java.util.Map;
+
 /** Flink scalar function that constructs a serialized {@link BlobViewStruct}. */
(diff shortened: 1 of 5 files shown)


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/function/BlobViewFunction.java
```diff
@@ -18,18 +18,157 @@
...
+            String catalogName, String defaultDatabase, Options catalogOptions) {
+        this.catalogName = catalogName;
+        this.catalogOptions = new Options(catalogOptions.toMap());
+        this.catalog = null;
+        this.blobViewFieldCache = null;
+    }
+
+    @Override
+    public void open(FunctionContext context) {
+        openCatalog(context.getUserCodeClassLoader());
```
XiaoHongbo-Hope, line 60: This will fail if blob view function without catalog context?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/function/BlobViewFunction.java
```diff
@@ -18,18 +18,157 @@
...
+            blobViewFieldCache().put(tableName, tableCache);
+        }
+        tableCache.put(fieldName, field);
+        return field;
+    }
+
+    private int fieldId(Identifier identifier, String fieldName) {
+        try {
+            Table table = catalog().getTable(identifier);
+            if (!table.rowType().containsField(fieldName)) {
```
XiaoHongbo-Hope, line 109: Should we also check that the resolved field is a BLOB field here?

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/function/BlobViewFunction.java
leaves12138: It is CatalogAwareFunction, so it must set catalog context. 
image

XiaoHongbo-Hope:
+1

XiaoHongbo-Hope (approved):

PR #7787, opened 2026-05-08
[python] Use SQLContext from pypaimon-rust directly

JingsongLi:
### Purpose

Remove the Python SQLContext wrapper (sql_context.py) and use the native Rust-based SQLContext from pypaimon_rust.datafusion instead. Update imports, CLI, tests, and docs accordingly.


paimon-python/pypaimon/cli/cli_sql.py
```diff
@@ -151,12 +153,13 @@ def cmd_sql(args):
 def _execute_query(ctx, query, output_format):
     """Execute a single SQL query and print the result."""
     try:
-        table = ctx.sql(query)
+        batches = ctx.sql(query)
     except Exception as e:
         print(f"Error: {e}", file=sys.stderr)
```
jerry-024, line 158: When `ctx.sql()` returns no batches (DDL/DML statements like `CREATE TABLE`, `INSERT INTO`, `DROP TABLE`), this silently produces no output. Users in single-query mode will see zero feedback — silent exit with code 0 and nothing printed.

Consider adding an else branch to confirm execution, e.g.:
```python
if batches:
    _print_table(pa.Table.from_batches(batches), output_format)
else:
    print("OK")
```

paimon-python/pypaimon/cli/cli_sql.py
```diff
@@ -255,9 +258,10 @@ def _interactive_repl(ctx, output_format):
 
             try:
                 start = time.time()
-                table = ctx.sql(query)
+                batches = ctx.sql(query)
                 elapsed = time.time() - start
-                _print_table(table, output_format, elapsed)
+                if batches:
+                    _print_table(pa.Table.from_batches(batches), output_format, elapsed)
                 print()
```
jerry-024, line 269: Same issue here in the interactive REPL path. When a user types `INSERT INTO t VALUES (1, 'x')` and gets zero feedback, they'll wonder whether the statement ran at all.

Consider:
```python
if batches:
    _print_table(pa.Table.from_batches(batches), output_format, elapsed)
else:
    print(f"OK ({elapsed:.2f}s)")
print()
```

jerry-024:
+1

PR #7788, opened 2026-05-08
[iceberg] Fix contains_nan in partition summary for float/double

ArnavBalyan:
### Purpose
  - IcebergManifestFile hardcodes contains_nan = false in `IcebergPartitionSummary`.
  - This value is used by readers for predicate pruning, hardcoding to false causes the files to get skipped that contain NaN matches.
  - Detect NaN via the existing min/max and pass it correctly. Issue only emerges for float/double values.
### Tests
 - UT


JingsongLi (approved):
+1

PR #7789, opened 2026-05-08
[spark] Support merge schema in MERGE INTO

Zouxxyy:
### Purpose

Add schema evolution support for MERGE INTO and fix nested-field alignment.

- With `spark.paimon.write.merge-schema=true`, `UPDATE *` / `INSERT *` evolves target schema with new source columns. Star clauses pull from source by name; explicit clauses fill NULL.
- A `FROM_STAR` `TreeNodeTag` preserves the original star intent, so a fully-listed explicit clause is not mistaken for `*`.
- `AssignmentAlignmentHelper` now reorders nested struct / array<struct> / map fields by name.

### Scope

- `UPDATE *` / `INSERT *` → evolve
- Explicit clauses → no evolve
- Mixed → evolve, star pulls source, explicit fills NULL
- Nested struct / array<struct> new fields

### Tests

13 new cases in `MergeIntoTableTestBase` plus `WHEN NOT MATCHED BY SOURCE` coverage in `MergeIntoNotMatchedBySourceTest`.


paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/catalyst/analysis/AssignmentAlignmentHelper.scala
```diff
@@ -177,4 +175,72 @@ trait AssignmentAlignmentHelper extends SQLConfHelper with ExpressionHelper {
...
+        val newVals =
+          ArrayTransform(MapValues(expression), LambdaFunction(reorderedVal, Seq(valVar)))
+        MapFromArrays(newKeys, newVals)
+      case _ =>
+        expression
+    }
+  }
+
+  /** Reorder source struct fields to match target field order by name, recursing into nested types. */
+  private def reorderStructByName(
```
JingsongLi, line 227: reorderStructByName crashes when target struct has fields absent from source

Should we support this?

The same issue applies to MapType value reordering in reorderFieldsByName.

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/commands/SchemaHelper.scala
```diff
@@ -41,21 +41,66 @@ private[spark] trait SchemaHelper extends WithFileStoreTable {
...
+      None
+    }
+  }
+
   /**
    * Recursively align columns from dataSchema to targetSchema by name. For nested struct fields,
    * reorder and fill nulls for missing sub-fields.
    */
-  private def alignColumns(
+  def alignColumns(
```
JingsongLi, line 103: SchemaHelper.scala now handles ArrayType<StructType> alignment via transform, but MapType<StructType> is not handled. Meanwhile, AssignmentAlignmentHelper.reorderFieldsByName does handle MapType. This inconsistency means the DataFrame write path won't align map values while the MERGE path will.

Is this a problem?

JingsongLi (approved):
+1

PR #7790, opened 2026-05-09
[flink][spark] fix incorrect first_row_id range in DataEvolution MergeInto

steFaiz:
### Purpose
Current first row id check in DataEvolutionPartialWriter maybe incorrect because of special files i.e. Blob Files and Vector FIles, which may cause:
```text
java.lang.AssertionError: assertion failed: Number of written records 2419 does not match expected number 244 for first row ID 19352.
```

This is because the blob file's record count override the normal file's record count:
image

We should filter out special files when calculating first_row_id to record_count mapping
### Tests
See :
`org.apache.paimon.flink.action.DataEvolutionMergeIntoActionITCase` for flink test
`org.apache.paimon.spark.sql.BlobTestBase` for spark test


JingsongLi:
Thanks @steFaiz  +1

PR #7791, opened 2026-05-09
[core] Implement withFilter for PartitionsTable and FilesTable

### LINKED ISSUE #7792: [Feature] Implement withFilter for PartitionsTable and FilesTable
### Search before asking

## Search before asking

- [x] I had searched in the [issues](https://github.com/apache/paimon/issues) and found no similar issues.

## Motivation

`PartitionsRead.withFilter` and `FilesRead.withFilter` are currently no-op
stubs marked `// TODO`. As a result, predicates on system tables such as
`SELECT * FROM t$files WHERE schema_id = 0` rely entirely on the upstream
engine to filter the result, even though the source could prune rows
much earlier.

## Solution

Apply the predicate as a post-filter on the materialized row before projection.

For `FilesRead`, leaves referencing `partition` / `bucket` / `level` need to
be stripped because they are already pushed down by `FilesScan` — re-evaluating
them on the materialized row may throw ClassCastException due to type mismatch
(e.g. a String literal vs the BinaryString in the row).

`PartitionsRead` keeps the predicate as-is because `PartitionsScan` does not
push anything down and `PartitionsTable` schema does not contain `bucket`
/ `level` columns.

POC ready (already implemented locally with tests, will open PR once assigned).

## Anything else?

No.

## Are you willing to submit a PR?

### Solution

_No response_

### Anything else?

_No response_

heye1005:
close #7792
### Purpose

Implement `PartitionsRead.withFilter` and `FilesRead.withFilter`, which were
previously no-op stubs marked `// TODO`. Predicates are now applied as a
post-filter on the materialized row, so `WHERE` clauses on these system tables
take effect at the source level.
For `FilesRead`, predicates on `partition`, `bucket` and `level` are excluded
from the post-filter because they are already handled by `FilesScan`.

### Tests

Added unit tests for equal / in / range / combined pushdown + post-filter
scenarios in `PartitionsTableTest` and `FilesTableTest`. All existing
system-table tests pass.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/system/FilesTable.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/system/FilesTable.java
@@ -35,6 +35,7 @@ import org.apache.paimon.manifest.FileSource;
 import org.apache.paimon.predicate.LeafPredicate;
 import org.apache.paimon.predicate.LeafPredicateExtractor;
 import org.apache.paimon.predicate.Predicate;
+import org.apache.paimon.predicate.PredicateBuilder;
 import org.apache.paimon.reader.RecordReader;
 import org.apache.paimon.schema.SchemaManager;
 import org.apache.paimon.schema.TableSchema;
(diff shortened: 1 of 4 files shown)


JingsongLi:
+1

PR #7794, opened 2026-05-09
[core] add an option to group delta files by partition for row tracking table

steFaiz:
### Purpose
For row tracking tables, we can group delta files before assigning row ids to reduce row id fraction within a single partition.
This can greatly reduce the cost of building global indices and the number of generated index files.

### Tests
See `paimon-core/src/test/java/org/apache/paimon/table/DataEvolutionTableTest.java`


JingsongLi (commented):
I think this should be enabled by default.

steFaiz:
> I think this should be enabled by default.

Thanks for your advise! Modified.

JingsongLi (approved):
+1

PR #7795, opened 2026-05-09
[core] Fix bug of bootstrap in ClusteringKeyIndex

JingsongLi:

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/lookup/sort/db/SimpleLsmKvDb.java
+++ b/paimon-common/src/main/java/org/apache/paimon/lookup/sort/db/SimpleLsmKvDb.java
@@ -247,11 +247,13 @@ public class SimpleLsmKvDb implements Closeable {
 
         int targetLevel = MAX_LEVELS - 1;
         List<SstFileMetadata> targetLevelFiles = levels.get(targetLevel);
+        List<SstFileMetadata> bulkLoadFiles = new ArrayList<>();
 
         SortLookupStoreWriter currentWriter = null;
         File currentSstFile = null;
         MemorySlice currentFileMinKey = null;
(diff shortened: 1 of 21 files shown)


leaves12138 (commented):
Reviewed the PR end-to-end. The two sort-order fixes look correct to me.

- `ClusteringKeyIndex.bootstrap` now bulk-loads key-index entries in the same ascending order required by `SimpleLsmKvDb`'s key comparator. Without this, SST min/max key ranges can be inverted, so first-row duplicate lookup may miss existing records after bootstrap.
- `ClusteringFileRewriter.sortAndRewriteFile` now rewrites clustering files in ascending clustering-key order, so the generated data file `minKey` / `maxKey` range remains valid for scan pruning and later clustering merge grouping.

I also added regression coverage for both paths:

- Composite PK first-row bootstrap keeps the original records and verifies the physical data-file row count does not include dropped duplicates.
- Composite clustering columns verify the rewritten file's clustering range is ascending; this test fails if the `ClusteringFileRewriter` sort order is reverted.

Validation run locally:

- `mvn -pl paimon-core -Pfast-build -Dtest=ClusteringTableTest#testFirstRow* test -DfailIfNoTests=false` — passed, 14 tests.
- `mvn -pl paimon-core -Pfast-build -Dtest=ClusteringTableTest test -DfailIfNoTests=false` — passed, 38 tests.
- `mvn -pl paimon-core spotless:check -DskipTests` — passed.

No further blocking comments from my side.

leaves12138 (approved):
Approved. The two sort-order fixes are correct, and the added regression tests cover both `ClusteringKeyIndex.bootstrap` and `ClusteringFileRewriter.sortAndRewriteFile`.

PR #7796, opened 2026-05-09
[python] Add nested-field projection support (append-only paths)

TheR1sing3un:
## Purpose

`ReadBuilder.with_projection` only accepts top-level column names today; dotted forms like `'struct.subfield'` are silently dropped. Reading just one leaf out of a struct ends up materialising the whole parent and discarding the rest client-side.

This PR ports nested projection on **append-only** read paths.

## Commits

1. **`Projection` utility + `ReadBuilder` API** — `pypaimon/utils/projection.py` (top-level / nested / empty subclasses). `with_projection(['a.b.c'])` resolves dotted names; `with_nested_projection(int[][])` is the low-level entry point. Flat output names are `_`-joined with `_$N` suffix on collisions; leaf field IDs are preserved for schema-evolution remap.

2. **Parquet/ORC pushdown** — `FormatPyArrowReader` switches to `dataset.scanner(columns={flat: ds.field(*path)})` when any path has length > 1; missing leaves return NULL. `SplitRead` threads the paths through with small bypasses on `_construct_partition_mapping`, `_get_fields_and_predicate`, `_get_final_read_data_fields`, and `create_index_mapping`. Avro / Lance / Vortex / Blob raise `NotImplementedError` if a nested path reaches them. PK tables that need a merge read also raise; raw-convertible PK splits work.

3. **Avro fallback** — fastavro has no native nested pruning; the reader walks each record dict by path. Top-level-only stays on the `record.get(name)` fast path.

## Tests

* `test_projection_utility.py` (21) + `test_read_builder_nested_projection.py` (10) cover the API and the projection utility.
* `test_nested_projection_e2e.py` (8) covers Parquet dotted projection, mixed nested + top-level, low-level integer paths, partitioned tables, Avro fallback, top-level fast-path regression, PK + merge `NotImplementedError`.
* Lint (`flake8 --config=dev/cfg.ini`) clean. Existing `reader_*_test` regression clean.

## Out of scope (follow-up)

* PK tables that go through `MergeFileSplitRead` — needs an `OuterProjectionRecordReader` so the merge function still sees full parent structs. Tracked separately; raises `NotImplementedError` until then.
* `ARRAY<ROW>` / `MAP` nested paths.
* Sub-field schema evolution by field ID (nested currently walks by name → NULL on rename).

## API / format

Backward compatible: top-level-only `with_projection` callers see no change. New: `with_nested_projection`. No file format change.

## Generative AI disclosure

Drafted with AI assistance; every behavioural guarantee is exercised by a test in one of the three new test files.

### DIFF
--- a/paimon-python/pypaimon/read/read_builder.py
+++ b/paimon-python/pypaimon/read/read_builder.py
@@ -80,8 +80,21 @@ class ReadBuilder:
         return TableRead(
             table=self.table,
             predicate=self._predicate,
-            read_type=self.read_type()
+            read_type=self.read_type(),
+            nested_name_paths=self._nested_name_paths(),
         )
+
+    def _nested_name_paths(self) -> Optional[List[List[str]]]:
+        """Resolve the current nested-projection state into a parallel list
+        of name paths against the underlying table schema. Returns ``None``
+        if the user only requested top-level projection (or no projection).
+        """
+        if not self._nested_paths:
+            return None
+        table_fields = self.table.fields
+        if self.table.options.row_tracking_enabled():
+            table_fields = SpecialFields.row_type_with_row_tracking(table_fields)
+        return Projection.of(self._nested_paths).to_name_paths(table_fields)
 
     def new_predicate_builder(self) -> PredicateBuilder:
         return PredicateBuilder(self.read_type())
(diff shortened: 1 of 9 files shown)


JingsongLi (approved):
+1

PR #7798, opened 2026-05-09
[core] Add Flink blob null-on-missing-file option

leaves12138:
## What
- Add `blob-write-null-on-missing-file` to let Flink writes return NULL for descriptor BLOB values whose referenced file does not exist.
- Move the existence check into `FlinkRowWrapper` before `BlobRecord.fromDescriptor(...)`, so missing descriptor files are handled at row conversion time with a WARN log.
- Reuse `UriReaderFactory` for descriptor existence checks, so file readers and their FileIO are cached by URI scheme/authority.

## Validation
- `mvn -pl paimon-common,paimon-flink/paimon-flink-common -am -Pfast-build -DfailIfNoTests=false -Dtest=UriReaderFactoryTest,FlinkRowWrapperTest test`
- `git diff --check`

### DIFF
--- a/docs/content/append-table/blob.md
+++ b/docs/content/append-table/blob.md
@@ -127,7 +127,7 @@ This allows one table to mix raw-data BLOB fields, descriptor-only BLOB fields, 
       </td>
     </tr>
     <tr>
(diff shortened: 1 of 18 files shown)


JingsongLi:
+1

PR #7799, opened 2026-05-09
StrictModeChecker don't throw exception if the partitions of checked snapshots don't overlapped

yuzelin:
### Purpose
Too many exception when committing will cause orphan files, and I found we can improve the check of StrictModeChecker to reduce exception.

### Tests
TableCommitTest

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/commit/StrictModeChecker.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/commit/StrictModeChecker.java
@@ -21,12 +21,16 @@ package org.apache.paimon.operation.commit;
 import org.apache.paimon.CoreOptions;
 import org.apache.paimon.Snapshot;
 import org.apache.paimon.Snapshot.CommitKind;
+import org.apache.paimon.data.BinaryRow;
 import org.apache.paimon.manifest.ManifestEntry;
 import org.apache.paimon.operation.FileStoreScan;
 import org.apache.paimon.table.source.ScanMode;
 import org.apache.paimon.utils.SnapshotManager;
 
+import java.util.HashSet;
 import java.util.Iterator;
+import java.util.List;
+import java.util.Set;
 
 /** A checker to check strict mode based on last safe snapshot. */
 public class StrictModeChecker {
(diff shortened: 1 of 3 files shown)


paimon-core/src/main/java/org/apache/paimon/operation/commit/StrictModeChecker.java
```diff
@@ -92,6 +101,29 @@ public void check(long newSnapshotId, CommitKind newCommitKind) {
         }
     }
 
+    private boolean hasOverlappedPartition(Snapshot snapshot, Set<BinaryRow> newPartitions) {
+        if (newPartitions.isEmpty()) {
+            return false;
+        }
+        Iterator<ManifestEntry> entries =
+                scan.withSnapshot(snapshot).withKind(ScanMode.DELTA).dropStats().readFileIterator();
```
leaves12138: `FileStoreScan` is mutable: `onlyReadRealBuckets()` sets state on both `AbstractFileStoreScan` and its `ManifestsReader`, and there is no reset in this checker. During an OVERWRITE commit, once the loop sees an earlier APPEND snapshot, the APPEND branch calls `scan.withSnapshot(...).withKind(DELTA).onlyReadRealBuckets()...`. The same `scan` instance is then reused here for later COMPACT/OVERWRITE snapshots, so those scans also inherit `onlyReadRealBuckets=true`.

That can make strict mode miss an overlapping later COMPACT/OVERWRITE whose files are not real-bucket files (for example bucket-unaware / postpone-bucket entries). A sequence like APPEND snapshot -> OVERWRITE/COMPACT snapshot on the same target partition -> new OVERWRITE can pass when it should fail.

Please avoid sharing this mutated scan state between checks, e.g. create a fresh `FileStoreScan` for each snapshot/check (pass a supplier into `StrictModeChecker`) or otherwise reset the bucket filter before `hasOverlappedPartition(snapshot, ...)`. It would also be good to add a regression test where an APPEND snapshot is checked before a later overlapping COMPACT/OVERWRITE snapshot.

leaves12138 (changes requested):
Thanks for the PR. The partition-overlap direction makes sense, but I found one correctness issue caused by reusing the mutable FileStoreScan inside StrictModeChecker. This can make strict mode miss a later overlapping COMPACT/OVERWRITE snapshot after checking an APPEND snapshot.

JingsongLi (approved):
+1

Issue #7781, opened 2026-05-08
[Bug] FileSystemBranchManager requires at least one tag on branch in fastForward


nryanov:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Paimon version

- 1.4.1

### Compute Engine

N/A
tested on Java API

### Minimal reproduce step

It's hard to give a minimal reproducible sample. Currently i'm implementing custom rest catalog and in pseudo-code it looks like this:
```
catalog.createBranch(Identifier(database, table), branchName)
... // produce some events in branch to trigger snapshot creation
catalog.forwardBranch(Identifier(database, table))
```

Working example in this case would be:
```
catalog.createBranch(...)
... // produce some events in branch to trigger snapshot creation
catalog.createTag(new Identifier(database, table, branch), tagName, ...)
catalog.forwardBranch()
```

Without tag i got an error:
```
Caused by: java.lang.RuntimeException: Exception occurs when fast forward 'branch_63db5f27e1ca4162a8cc7f79a5addc8c' (directory in s3a://warehouse/db_6e4bd5b663ad4dc496ebe99118156b27.db/table_b6de925dbf49482b8fdc313bb74797e1/branch/branch-branch_63db5f27e1ca4162a8cc7f79a5addc8c).
	at org.apache.paimon.utils.FileSystemBranchManager.fastForward(FileSystemBranchManager.java:205)
...
Caused by: java.io.FileNotFoundException: No such file or directory: s3a://warehouse/db_6e4bd5b663ad4dc496ebe99118156b27.db/table_b6de925dbf49482b8fdc313bb74797e1/branch/branch-branch_63db5f27e1ca4162a8cc7f79a5addc8c/tag

```

### What doesn't meet your expectations?

According to current spec of catalog#fastForward:
```
Fast-forward a branch to main branch.
Params:
identifier – path of the table, cannot be system or branch name.
branch – the branch name
void fastForward(Identifier identifier, String branch) throws BranchNotExistException;
```

i expect that tag(s) is not required for fast-forward.

### Anything else?

It seems, that a minor change in the FileSystemBranchManager#fastForward will fix the problem: check if any tags on branch exist and if yes, then copy them, otherwise just skip this step

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #7800, opened 2026-05-09
[core] Skip tag copy in fastForward when branch has no tags

ArnavBalyan:
### Purpose
  - fastForward copies the tag directory to main, even when the branch did not have a tag created.
  - On object store,`listStatus` on the missing directory throws FNF exception, causing failure for fastforward for branch without tags.
  - Add a guard to check that the tag dir exists before the copy operation.
  - Closes user reported bug #7781
### Tests
 - UT


JingsongLi (approved):
+1

PR #7801, opened 2026-05-09
[python] Add nested-field projection on primary-key merge-read path

TheR1sing3un:
## Purpose

Complete #7796: PK tables now serve nested-field projections through the merge reader.
Before this PR, ``with_projection(['mv.latest_version'])`` on a PK table whose split
went through the merge path raised ``NotImplementedError``.

## Approach

Inner reader keeps the full ROW sub-structure (so deduplicate / partial-update /
aggregation merge functions still see the original row). The new
``OuterProjectionRecordReader`` sits above the merge unwrap and walks each
configured name path on the inner row to emit a flat OffsetRow matching the
user-visible read schema. This mirrors Java's NestedProjectedRowData split.

## Commits

1. ``Add OuterProjectionRecordReader for nested-field PK reads`` — the wrapper +
   16-case unit test.
2. ``Fix primary-key read when value projection drops PK columns`` — pre-existing
   bug surfaced by this work: the PARQUET/ORC/Lance/Vortex format readers used
   merge-internal aliases (``_KEY_id``) to look up DataFields, but the file
   stores PK columns under their bare name (``id``). When projection narrowed
   value fields enough to drop the bare PK from ``self.read_fields``, the
   format reader silently skipped the PK column and the merge KeyValue layout
   raised ``Offset N plus arity M is out of row length L``. Build the lookup
   from both name spaces.
3. ``Support nested-field projection on primary-key tables`` — wire commit 1
   into ``MergeFileSplitRead`` and remove the guard in ``TableRead``.

## Tests

- 16 new unit tests for ``OuterProjectionRecordReader`` (None handling, path
  walks, batch reuse contract, row-kind inheritance, empty-input rejection).
- 3 new e2e tests on PK tables: single sub-path, multiple sub-paths under the
  same struct, mixed top-level and nested ordering.
- 1 new regression test on PK + value-only projection.
- All existing append-only nested + reader_primary_key + reader_append_only +
  projection-utility + read-builder tests still pass.

## Out of scope

- DataEvolution + nested projection (still raises ``NotImplementedError``);
  needs the same widen-then-extract treatment, follow-up PR.
- ARRAY<ROW> / MAP<X, ROW> nested projection (Java side also keeps NestedProjection
  on ROW only).
- Aggregation merge engine + nested combo regression — depends on
  PR-MERGE-AGGREGATION landing first.

### DIFF
--- a/paimon-python/pypaimon/read/split_read.py
+++ b/paimon-python/pypaimon/read/split_read.py
@@ -189,6 +189,19 @@ class SplitRead(ABC):
         nested_path_by_name = self._nested_path_by_name()
         has_nested = nested_path_by_name is not None
 
+        # Look DataFields up by every name the format reader may receive.
+        # ``self.read_fields`` covers nested-projection flat names and
+        # merge-internal aliases (``_KEY_id``); the trimmed-fields list
+        # adds the user-facing PK name (``id``) the file actually stores
+        # so PK columns stay reachable when value projection narrows the
+        # user-facing read_type and drops the bare PK field.
+        name_to_field: Dict[str, DataField] = {f.name: f for f in self.read_fields}
+        _, _trimmed_lookup_fields = self._get_trimmed_fields(
+            self._get_read_data_fields(), self._get_all_data_fields()
+        )
+        for f in _trimmed_lookup_fields:
+            name_to_field.setdefault(f.name, f)
+
         format_reader: RecordBatchReader
         if file_format == CoreOptions.FILE_FORMAT_AVRO:
             avro_nested_paths = (
@@ -210,7 +223,6 @@ class SplitRead(ABC):
             if has_nested:
                 raise NotImplementedError(
                     "Nested-field projection is not supported on Lance files")
-            name_to_field = {f.name: f for f in self.read_fields}
             ordered_read_fields = [name_to_field[n] for n in read_file_fields if n in name_to_field]
             format_reader = FormatLanceReader(self.table.file_io, file_path, ordered_read_fields,
                                               read_arrow_predicate, batch_size=batch_size,
@@ -219,7 +231,6 @@ class SplitRead(ABC):
             if has_nested:
                 raise NotImplementedError(
                     "Nested-field projection is not supported on Vortex files")
-            name_to_field = {f.name: f for f in self.read_fields}
             ordered_read_fields = [name_to_field[n] for n in read_file_fields if n in name_to_field]
             predicate_fields = _get_all_fields(self.push_down_predicate) if self.push_down_predicate else set()
             format_reader = FormatVortexReader(self.table.file_io, file_path, ordered_read_fields,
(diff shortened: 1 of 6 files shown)


paimon-python/pypaimon/read/split_read.py
```diff
@@ -189,6 +189,15 @@ def file_reader_supplier(self, file: DataFileMeta, for_merge_read: bool,
         nested_path_by_name = self._nested_path_by_name()
         has_nested = nested_path_by_name is not None
 
+        # Cover both the merge-internal aliases (``_KEY_id``) and the
+        # bare user-facing PK name (``id``) the file actually stores.
+        name_to_field: Dict[str, DataField] = {f.name: f for f in self.read_fields}
```
leaves12138, line 194: This alias-safe lookup is used below for Lance/Vortex/Parquet/ORC, but the Avro branch still passes `self.read_fields` as `full_fields`. For a PK merge read with `with_projection(['behavior'])` (or `with_projection(['mv.latest_version'])` on an Avro PK table), `_get_final_read_data_fields()` asks the file reader for the bare PK name such as `user_id`, while `FormatAvroReader` builds `full_fields_map` from `self.read_fields`, which only contains the merge alias `_KEY_user_id` when the PK was dropped from the value projection. That raises `KeyError: 'user_id'` before the merge reader can run.

Please apply the same bare-PK alias handling to the Avro path too (for example by passing a full field list that also contains the trimmed bare PK fields, or by letting `FormatAvroReader` receive resolved `DataField`s) and add an Avro regression test for PK value-only projection / PK nested projection.

leaves12138 (changes requested):
Thanks for the PR. The merge-read widening/extraction approach looks reasonable, and the new Parquet/ORC-style paths passed my targeted checks. I found one blocking gap in the Avro path: the alias-safe PK field lookup added for the other format readers is not applied to FormatAvroReader, so PK merge reads still fail when the user projection drops the PK column (and the new nested PK projection fails for Avro as well).

paimon-python/pypaimon/read/split_read.py
TheR1sing3un: > This alias-safe lookup is used below for Lance/Vortex/Parquet/ORC, but the Avro branch still passes `self.read_fields` as `full_fields`. For a PK merge read with `with_projection(['behavior'])` (or `with_projection(['mv.latest_version'])` on an Avro PK table), `_get_final_read_data_fields()` asks the file reader for the bare PK name such as `user_id`, while `FormatAvroReader` builds `full_fields_map` from `self.read_fields`, which only contains the merge alias `_KEY_user_id` when the PK was dropped from the value projection. That raises `KeyError: 'user_id'` before the merge reader can run.
> 
> Please apply the same bare-PK alias handling to the Avro path too (for example by passing a full field list that also contains the trimmed bare PK fields, or by letting `FormatAvroReader` receive resolved `DataField`s) and add an Avro regression test for PK value-only projection / PK nested projection.

Nice catch! Already fixed. Please review it again.

JingsongLi:
+1

PR #7803, opened 2026-05-10
[core] Skip stats from schema in DataEvolution when column type has changed

ArnavBalyan:
### Purpose
  - Stats merges today ignores type changes during schema evolution. After `alter table` changes (eg Int to String), stats from older files written under the prior schema are merged in under the new type.
  - When types mismatch, this causes garbage stats to be accumulated which causing wrong pruning/silently data drop.
  - Ensure we can detect this mismatch and skip the stats for such cases.

### Tests
 - UT


leaves12138 (approved):
LGTM. The change is conservative and correct for data-evolution stats pruning: if a file contains the same field id but its stats were written with a different physical type than the current table field, using those min/max bytes with the current predicate type can produce invalid comparisons and wrong pruning. Skipping that file's stats for the field makes the predicate evaluation fall back to unknown stats instead of dropping data incorrectly. The added regression test covers the mixed old/new type case.

JingsongLi:
+1

ArnavBalyan:
Thanks @leaves12138 @JingsongLi!

PR #7804, opened 2026-05-10
[python] Add per-partition bucket pruning for HASH_FIXED tables

TheR1sing3un:
## Background

PR-5.4 (#7744) added bucket pruning for HASH_FIXED tables but only on
the bucket-key dimension. Predicates that mix a partition column and
a bucket column under a top-level OR — e.g.
``(part='a' AND id=1) OR (part='b' AND id=2)`` — couldn't be pruned:
the OR mixes two dimensions, so the existing logic gave up and read
every bucket in both partitions. PR-5.4 left this as a TODO in the
module docstring.

## Effect

Same query now reads exactly one bucket per partition (the bucket
holding ``id=1`` in ``part='a'``, the bucket holding ``id=2`` in
``part='b'``). The selector evaluates the predicate per partition
value first — the OR collapses to a single AND inside each partition
— and bucket selection runs on that simplified form.

Soundness contract is unchanged: the bucket set remains a superset
of the buckets that contain matching rows; any error falls open to
"all buckets accept", never drops a bucket with matches.

Two commits — helper + ``FileScanner`` wiring. 9 unit tests cover the
predicate-fold walker and the per-partition cache; one e2e test on a
2-partition × 4-bucket table proves the mixed-OR query reads ≤ 2
splits instead of one per (partition, bucket).

### DIFF
--- a/paimon-python/pypaimon/read/scanner/bucket_select_converter.py
+++ b/paimon-python/pypaimon/read/scanner/bucket_select_converter.py
@@ -51,38 +51,26 @@ Conservative scope (deliberately narrower than Java's general flexibility):
   * Total cartesian product capped at MAX_VALUES (1000), again matching
     Java; above that, fall back to a full scan.
 
-Returns a callable ``selector(bucket: int, total_buckets: int) -> bool``.
-The callable is cached per ``total_buckets`` to handle the rare case
-where bucket count varies across snapshots (rescale).
+Returns a callable ``selector(partition, bucket: int, total_buckets: int)
+-> bool``. The callable is cached per ``(partition, total_buckets)`` to
+handle (a) bucket count variation across snapshots (rescale) and (b)
+per-partition predicate specialisation: predicates of the form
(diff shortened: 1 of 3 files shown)


leaves12138 (approved):
Reviewed the change against the Java `BucketSelectConverter` / `BucketSelector` / `PartitionValuePredicateVisitor` flow.

The important semantics line up: the selector specializes the predicate with the concrete partition value before extracting finite Equal/In bucket-key constraints, keeps the existing MAX_VALUES / fail-open behavior for unsupported bucket predicates, and keeps the early manifest filter conservative when the partition is still unknown.

One difference from Java is intentional and correctness-safe: when partition specialization folds the predicate to false, Python returns an empty bucket set, while Java's `BucketSelector` itself falls open and relies on the scan-level partition filter to drop the entry. Since no row in that concrete partition can satisfy the complete predicate, the stricter Python pruning does not introduce false negatives.

The added tests cover the partition predicate folding, per-partition cache keying, unknown-partition fallback, and the mixed partition/bucket OR integration case. Looks good to me.

JingsongLi:
+1

PR #7805, opened 2026-05-10
[python] Add three-layer predicate pushdown correctness regression tests

TheR1sing3un:
## Background

Predicate pushdown in pypaimon spans three correctness-sensitive layers
(filter / file index / value stats) plus a couple of subtle invariants
around PK tables and post-merge value filtering. The correctness
obligations have so far been enforced by spot tests in a handful of
unrelated files; a regression that re-introduces a
``value_stats``-vs-``key_stats`` confusion or a per-file value filter
on PK rows would be hard to detect without targeted coverage.

## Effect

Adds a single test file that locks in the invariants as a directed
regression suite:

- **PK manifest pruning consults ``key_stats`` only** — covered by both
  the value-only-predicate early-return path and the compound
  (``id`` + ``val``) path, in each case using ``value_stats`` that
  would drop the file if consulted.
- **Reader-level value predicate is post-merge** — writing a newer
  value that fails ``< 50`` must not resurrect the older matching
  value.
- **Append-only ``value_stats`` pruning** is applied and correct.

Three layers of coverage:
1. **Unit** — synthetic ``ManifestEntry`` through
   ``FileScanner._filter_manifest_entry`` to assert the manifest
   gate's exact behaviour without relying on full I/O.
2. **Round-trip** — real catalog writes/reads with a post-merge oracle.
3. **Property** — seeded random datasets + 12 random predicate ops
   (incl. ``is_null``, re-covering a historical bug where missing
   ``null_counts`` caused ``isNull`` to drop every file). No
   ``hypothesis`` dependency, keeps the Python 3.6 compatibility
   contract intact.

No production code changes — pure regression coverage. 11 new test
cases; full suite passes locally.

### DIFF
--- a/paimon-python/pypaimon/tests/pushdown_correctness_test.py
+++ b/paimon-python/pypaimon/tests/pushdown_correctness_test.py
@@ -1,0 +1,681 @@
+################################################################################
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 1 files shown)


JingsongLi:
+1

Issue #7806, opened 2026-05-10
[Feature] Support snapshot-based sequence ordering for primary-key tables


JunRuiLee:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

## Motivation

  We run a dataset management platform on top of Paimon for LLM training data. Multiple teams and pipelines (cleaning, annotation, deduplication, etc.) write to the same primary-key tables concurrently.

  Each writer assigns sequence numbers independently, so **sequence numbers across writers are incomparable**. The only reliable cross-writer ordering signal is the commit order (snapshot id). Today Paimon resolves primary-key conflicts solely by sequence number (or `sequence.field`), which cannot express this.

  **What we need: records committed in a later snapshot always win**, regardless of per-record sequence numbers.

## Solution

  Add a table option `sequence.snapshot-ordering`. When enabled, merge uses the commit snapshot id as the primary tiebreaker for primary-key conflicts, with sequence number as the secondary tiebreaker within the same snapshot. This reuses `DataFileMeta.minSequenceNumber` to carry the snapshot id at file level — same pattern as row-tracking tables repurposing `minSequenceNumber` for `firstRowId`.

  ### Design Details

  1. **File-level stamp at commit time.** Repurpose `minSequenceNumber` to carry the commit snapshot id, same injection point as row-id assignment.
  2. **Propagate through compaction.** Compaction output inherits `max(input snapshot ids)`, so it never shadows concurrent writes.
  3. **Inflate to record at read time.** `KeyValueFileReaderFactory` stamps the file-level snapshot id onto each `KeyValue`.
  4. **Conditional comparator injection.** Sort-merge readers use snapshot id as primary tiebreaker only when enabled.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

JingsongLi:
Is this a deduplicated pk table?

JunRuiLee:
> Is this a deduplicated pk table?

 Not limited to deduplicate. This feature aims to work with all merge engines — snapshot-ordering only defines record precedence, the same way sequence number does.

JingsongLi:
Can we use current time as a sequence field? Or is the core requirement of your business to differentiate based on snapshots?

JingsongLi:
I am more concerned about the modifications to DataFileMeta because it is the most basic structure, unless we really introduce a major feature.

JunRuiLee:
[...]

PR #7807, opened 2026-05-10
[core] Add commit-snapshot-id field to DataFileMeta

JunRuiLee:
### Purpose

  Part 1/3 of https://github.com/apache/paimon/issues/7806

  Add the infrastructure for snapshot-based sequence ordering:
  - Reserve `_COMMIT_SNAPSHOT_ID` special field and add `snapshotId` attribute on `KeyValue`.
  - Add nullable `commitSnapshotId` field to `DataFileMeta` with builder-based copy methods.
  - Bump `DataSplit` version 8→9 and `CommitMessageSerializer` version 11→12 with legacy serializer for backward-compatible deserialization.

  Pure plumbing — no behavior change. The field is always null until the write-path PR stamps it.

  ### Tests
  - `DataFileMetaCommitSnapshotIdTest`: field default null, assign/preserve across copy methods, serializer round-trip.
  - `DataFileMetaSerializerCompatibilityTest`: legacy 20-field serialize → deserialize produces `commitSnapshotId == null`.

### DIFF
--- a/paimon-api/src/main/java/org/apache/paimon/table/SpecialFields.java
+++ b/paimon-api/src/main/java/org/apache/paimon/table/SpecialFields.java
@@ -93,13 +93,17 @@ public class SpecialFields {
     public static final DataField ROW_ID =
             new DataField(Integer.MAX_VALUE - 5, "_ROW_ID", DataTypes.BIGINT().notNull());
 
+    public static final DataField COMMIT_SNAPSHOT_ID =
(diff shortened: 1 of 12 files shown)


JingsongLi:
Is this a deduplicated pk table?

JunRuiLee:
> Is this a deduplicated pk table?

Not limited to deduplicate. This feature aims to work with all merge engines — snapshot-ordering only defines record precedence, the same way sequence number does.

JingsongLi:
@JunRuiLee Let's discuss this in issue.

PR #7809, opened 2026-05-11
[flink] Add PARTITION_DYNAMIC strategy for partitioned append-only tables

JingsongLi:
### Purpose

This adds a new partition sink strategy that dynamically adjusts shuffle based on per-partition traffic statistics collected at each checkpoint. Unlike HASH which statically maps each partition to one subtask, PARTITION_DYNAMIC monitors data distribution and rebalances load across downstream subtasks proportionally to partition traffic weight.

This PR refers to the implementation of https://github.com/apache/fluss/pull/1784

### Tests

UTs and ITCase in AppendTableITCase


paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/partition/DataStatisticsOperator.java
```diff
@@ -0,0 +1,128 @@
...
+        if (globalStatistics != null) {
+            output.collect(new StreamRecord<>(StatisticsOrRecord.fromStatistics(globalStatistics)));
+        }
+    }
+
+    @Override
+    public void processElement(StreamRecord<InternalRow> streamRecord) throws Exception {
+        InternalRow row = streamRecord.getValue();
+        BinaryRow partition = extractor.partition(row);
+        String partitionKey = partition.toString();
```
leaves12138: `BinaryRow.toString()` is not a lossless partition key. `BinaryRow`/`BinarySection` does not override `toString()`, so this falls back to `Object.toString()`, which only embeds the content-based `hashCode()`. Different partition values can collide and then be merged into the same statistics entry, so the dynamic assignment can be built from incorrect weights. Could we use a canonical key here instead, e.g. copied/serialized `BinaryRow` bytes (with an appropriate serializer) or a partition-path/value string, and use the same key on the lookup side?

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/AppendTableITCase.java
```diff
@@ -562,7 +567,179 @@ public void testPartitionStrategyForPartitionedTable(CoreOptions.PartitionSinkSt
...
+        for (int i = 1; i <= 100; i++) {
+            values.append(String.format("(%d, 'data%d', '20250301'),", i, i));
+        }
+        for (int i = 101; i <= 110; i++) {
+            values.append(String.format("(%d, 'data%d', '20250302'),", i, i));
+        }
+        for (int i = 111; i <= 115; i++) {
+            values.append(String.format("(%d, 'data%d', '20250303'),", i, i));
+        }
+
```
leaves12138, line 626: Most of the new IT coverage runs through batch `INSERT`, so it mainly verifies data correctness and the no-statistics fallback (`min(parallelism, 4)`). Since the new dynamic behavior is driven by checkpoint statistics and coordinator events, could we add an end-to-end streaming/checkpointed test that writes a skewed partition before/after at least one checkpoint and verifies that the hot partition is actually spread according to the updated statistics? That would cover the coordinator -> operator event -> partitioner update path, not only the direct unit-test path.

leaves12138 (changes requested):
Thanks for the PR. The checkpoint/coordinator/partitioner flow looks generally consistent, but I found one correctness issue in how partition keys are encoded for statistics and assignment lookup. I also left a test coverage suggestion for the checkpoint-driven dynamic path.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/sink/partition/StatisticsOrRecordChannelComputer.java
```diff
@@ -0,0 +1,202 @@
...
+                        Math.min(numChannels, DEFAULT_SUBTASK_COUNT_FOR_UNKNOWN_PARTITION);
+                int startChannel = Math.abs(partitionKey.hashCode()) % numChannels;
+                List<Integer> subtasks = new ArrayList<>(defaultSubtaskCount);
+                List<Long> weights = new ArrayList<>(defaultSubtaskCount);
+                for (int i = 0; i < defaultSubtaskCount; i++) {
+                    subtasks.add((startChannel + i) % numChannels);
+                    weights.add(1L);
+                }
+                assignment = new WeightedRandomAssignment(subtasks, weights, random);
+                assignments.put(partitionKey, assignment);
```
leaves12138: Please copy the partition key before storing it in `assignments`. `RowPartitionKeyExtractor` uses the generated `Projection`, whose output `BinaryRow` is reused by default, so this `partitionKey` is mutable and can be overwritten by the next record. Storing it directly as a `HashMap` key can corrupt the fallback cache for unknown/no-statistics partitions (duplicate mutable keys or reusing another partition's cached assignment). `assignments.put(partitionKey.copy(), assignment)` should avoid that; a small test with multiple unknown partitions would cover it.

leaves12138 (changes requested):
Thanks for the update. The previous `BinaryRow.toString()` issue is fixed: statistics now use copied `BinaryRow` keys and the `DataStatisticsSerializer` preserves them across coordinator events. The new unit coverage for aggregation, serialization, operator events, and weighted assignment also looks useful.

I found one remaining mutable-key issue in the unknown/no-statistics fallback path. After that is fixed, I think this should be good.

leaves12138 (approved):
Thanks for the update. The mutable `BinaryRow` key in the unknown/no-statistics fallback path is fixed by storing `partitionKey.copy()`, and the added multi-unknown-partition test covers the cache-corruption case I was concerned about.

The earlier `BinaryRow.toString()` statistics-key issue was already addressed with copied `BinaryRow` keys plus `DataStatisticsSerializer`, and the checkpoint/coordinator/operator-event path now has reasonable unit and integration coverage. Looks good to me.

PR #7812, opened 2026-05-11
[core] fix NPE in BTreeFileMetaSelector when btree index contains empty string key.

### LINKED ISSUE #7811: [Bug] NPE in BTreeFileMetaSelector when btree global index contains empty string key
### Search before asking

### Paimon version

1.14

### Compute Engine

flink 1.18

### Minimal reproduce step

```java
@Test
    public void testMetaSelectorWithEmptyStringKey() {
        // Simulate the real NPE scenario: btree index file with empty string as firstKey
        KeySerializer stringSerializer = KeySerializer.create(new VarCharType());
        byte[] emptyKey = stringSerializer.serialize(BinaryString.EMPTY_UTF8);
        byte[] normalKey = stringSerializer.serialize(BinaryString.fromString("www.example.com"));

        BTreeIndexMeta metaWithEmptyFirstKey = new BTreeIndexMeta(emptyKey, normalKey, false);
        BTreeIndexMeta metaWithNormalKeys =
                new BTreeIndexMeta(
                        stringSerializer.serialize(BinaryString.fromString("aaa.com")),
                        stringSerializer.serialize(BinaryString.fromString("zzz.com")),
                        false);
        BTreeIndexMeta metaOnlyNulls = new BTreeIndexMeta(null, null, true);

        List<GlobalIndexIOMeta> files =
                Arrays.asList(
                        new GlobalIndexIOMeta(
                                new Path("file_empty"), 1, metaWithEmptyFirstKey.serialize()),
                        new GlobalIndexIOMeta(
                                new Path("file_normal"), 1, metaWithNormalKeys.serialize()),
                        new GlobalIndexIOMeta(
                                new Path("file_nulls"), 1, metaOnlyNulls.serialize()));

        FieldRef ref = new FieldRef(1, "page_host", new VarCharType());
        BTreeFileMetaSelector selector = new BTreeFileMetaSelector(files, stringSerializer);

        // visitEqual should not throw NPE
        Optional<List<GlobalIndexIOMeta>> result =
                selector.visitEqual(ref, BinaryString.fromString("www.example.com"));
        Assertions.assertThat(result).isNotEmpty();
        assertFiles(result.get(), Arrays.asList("file_empty", "file_normal"));

        // visitLessThan should not throw NPE
        result = selector.visitLessThan(ref, BinaryString.fromString("bbb.com"));
        Assertions.assertThat(result).isNotEmpty();
        assertFiles(result.get(), Arrays.asList("file_empty", "file_normal"));

        // visitGreaterThan should not throw NPE
        result = selector.visitGreaterThan(ref, BinaryString.fromString("www.example.com"));
        Assertions.assertThat(result).isNotEmpty();

        // visitIn should not throw NPE
        result =
                selector.visitIn(
                        ref,
                        Arrays.asList(
                                BinaryString.fromString("www.example.com"),
                                BinaryString.fromString("zzz.com")));
        Assertions.assertThat(result).isNotEmpty();

        // visitBetween should not throw NPE
        result =
                selector.visitBetween(
                        ref, BinaryString.EMPTY_UTF8, BinaryString.fromString("zzz.com"));
        Assertions.assertThat(result).isNotEmpty();
    }
```

When querying a table with a btree global index on a STRING/VARCHAR column, a NullPointerException is thrown: 
Image

### Anything else?

_No response_

zhoulii:
### Purpose

close #7811

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/globalindex/btree/BTreeIndexMeta.java
+++ b/paimon-common/src/main/java/org/apache/paimon/globalindex/btree/BTreeIndexMeta.java
@@ -70,13 +70,13 @@ public class BTreeIndexMeta {
             sliceOutput.writeInt(firstKey.length);
             sliceOutput.writeBytes(firstKey);
         } else {
-            sliceOutput.writeInt(0);
+            sliceOutput.writeInt(-1);
         }
         if (lastKey != null) {
             sliceOutput.writeInt(lastKey.length);
             sliceOutput.writeBytes(lastKey);
         } else {
-            sliceOutput.writeInt(0);
+            sliceOutput.writeInt(-1);
         }
         sliceOutput.writeByte(hasNulls ? 1 : 0);
         return sliceOutput.toSlice().getHeapMemory();
@@ -86,10 +86,18 @@ public class BTreeIndexMeta {
         MemorySliceInput sliceInput = MemorySlice.wrap(data).toInput();
         int firstKeyLength = sliceInput.readInt();
         byte[] firstKey =
-                firstKeyLength == 0 ? null : sliceInput.readSlice(firstKeyLength).copyBytes();
+                firstKeyLength < 0
+                        ? null
+                        : firstKeyLength == 0
+                                ? new byte[0]
+                                : sliceInput.readSlice(firstKeyLength).copyBytes();
         int lastKeyLength = sliceInput.readInt();
         byte[] lastKey =
-                lastKeyLength == 0 ? null : sliceInput.readSlice(lastKeyLength).copyBytes();
+                lastKeyLength < 0
+                        ? null
+                        : lastKeyLength == 0
+                                ? new byte[0]
+                                : sliceInput.readSlice(lastKeyLength).copyBytes();
         boolean hasNulls = sliceInput.readByte() == 1;
         return new BTreeIndexMeta(firstKey, lastKey, hasNulls);
     }
--- a/paimon-common/src/main/java/org/apache/paimon/globalindex/btree/BTreeIndexMeta.java
+++ b/paimon-common/src/main/java/org/apache/paimon/globalindex/btree/BTreeIndexMeta.java
@@ -108,7 +108,7 @@ public class BTreeIndexMeta {
         // V0 format ends with hasNulls (0 or 1), V1 format ends with VERSION (>= 2).
         byte lastByte = data[data.length - 1];
 
-        if (lastByte >= VERSION) {
+        if (lastByte == VERSION) {
             // V1 format: -1 means null, 0 means empty byte array, version byte at the end
             MemorySliceInput sliceInput = MemorySlice.wrap(data).toInput();
             int firstKeyLength = sliceInput.readInt();
--- a/paimon-common/src/main/java/org/apache/paimon/globalindex/btree/BTreeIndexMeta.java
+++ b/paimon-common/src/main/java/org/apache/paimon/globalindex/btree/BTreeIndexMeta.java
@@ -27,8 +27,25 @@ import javax.annotation.Nullable;
 /**
  * Index Meta of each BTree index file. The first key and last key of this meta could be null if the
  * entire btree index file only contains nulls.
+ *
+ * <h3>Serialization format versions:</h3>
+ *
+ * <p>V0 (legacy): {@code [firstKeyLength(4)] [firstKey] [lastKeyLength(4)] [lastKey] [hasNulls(1)]}
+ *
+ * <p>V0 uses 0 to represent null key length, which cannot distinguish null from empty byte array
+ * (e.g. empty string key). The last byte is hasNulls (0 or 1).
+ *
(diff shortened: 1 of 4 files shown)


paimon-common/src/main/java/org/apache/paimon/globalindex/btree/BTreeIndexMeta.java
```diff
@@ -70,13 +70,13 @@ public byte[] serialize() {
...
             sliceOutput.writeBytes(firstKey);
         } else {
-            sliceOutput.writeInt(0);
+            sliceOutput.writeInt(-1);
         }
         if (lastKey != null) {
             sliceOutput.writeInt(lastKey.length);
             sliceOutput.writeBytes(lastKey);
         } else {
-            sliceOutput.writeInt(0);
```
steFaiz, line 79: Dose this have incompatibility risks? For example if an old version file is all-null, a new version meta will set the firstKey and lastKey as an empty byte[].

zhoulii: Good catch! You're right — there is a backward compatibility risk. When an old-version file has all-null keys (serialized as int(0) for both keys), the new deserializer would incorrectly read them as byte[0] instead of null. Fixed already.

paimon-common/src/main/java/org/apache/paimon/globalindex/btree/BTreeIndexMeta.java
```diff
@@ -70,27 +87,59 @@ public byte[] serialize() {
...
-        int lastKeyLength = sliceInput.readInt();
-        byte[] lastKey =
-                lastKeyLength == 0 ? null : sliceInput.readSlice(lastKeyLength).copyBytes();
-        boolean hasNulls = sliceInput.readByte() == 1;
-        return new BTreeIndexMeta(firstKey, lastKey, hasNulls);
+        // Detect format version by reading the last byte.
+        // V0 format ends with hasNulls (0 or 1), V1 format ends with VERSION (>= 2).
+        byte lastByte = data[data.length - 1];
+
+        if (lastByte >= VERSION) {
```
JingsongLi: Use == VERSION instead of >= VERSION for safer forward compatibility

JingsongLi (changes requested):
And add defensive null handling in BTreeFileMetaSelector to protect against legacy V0 data that already has the corruption. should consider adding a defensive null check in BTreeFileMetaSelector.deserialize() (line 235):
```
  private Object deserialize(byte[] valueBytes) {
      if (valueBytes == null) {
          return null;  // defensive: V0 data may have null for empty string
      }
      return keySerializer.deserialize(MemorySlice.wrap(valueBytes));
  }
```

paimon-common/src/main/java/org/apache/paimon/globalindex/btree/BTreeIndexMeta.java
zhoulii: Comments addressed. Thanks!

JingsongLi (changes requested):
This may cause significant compatibility issues, and I am trying to see if this problem can be bypassed.

JingsongLi:
We can do this modification:

- BTreeIndexMeta remains unchanged (compatible with old data)
- In BTreeFileMetaSelector, all comparison operations check whether getFirstKey()/getLastKey() is null before calling deserialize():
     - For one-sided comparison (<,<=,>,>=): If the required key is null, conservatively return true (including the file)
     - For range operations (equal, in, between): if any key is null, conservatively return true

This not only avoids NPE, but also avoids missing potentially matching files (at most, reading unnecessary files will not result in missing results), and is fully compatible with old serialization formats.

steFaiz:
> We can do this modification:
> 
> * BTreeIndexMeta remains unchanged (compatible with old data)
> * In BTreeFileMetaSelector, all comparison operations check whether getFirstKey()/getLastKey() is null before calling deserialize():
>   

@JingsongLi  Thanks for your advise! I think the main complications lie within BTreeIndexReader. Currently, BTreeIndexReader stores minKey and maxKey to simplify many operations. Due to filtering by BTreeMetaSelector, BTreeIndexReader currently assumes that either both minKey and maxKey are null (containing only the nullBitmap) or neither is null. It then uses minKey and maxKey to simplify operations such as lessThan and greaterThan.

For example, if a btree file accidentally only stores keys of empty string, it's minKey and maxKey will be deserialized as null from BTreeFileMeta.

I think the solution may be do not store minKey and maxKey anymore in BTreeIndexReader, and change the rangeQuery implementation, so that null `from` means starting from the beginning, null `to` means reading to the end. 

How do you think? I could submit an another PR to fix this in both java, python and rust.

steFaiz:
@JingsongLi  I found another case that may be difficult to solve without modifying BTreeIndexMeta:
image

visitIsNotNull will filter out all metas which only contain nulls. If a file only contains empty string, it will be filtered. We couldn't distinguish it from a real only-null file.

JingsongLi:
@steFaiz I think it is a good direction, let's go ahead.

PR #7813, opened 2026-05-11
[python] Pre-repartition Ray writes by (partition, bucket) for fixed-bucket tables

TheR1sing3un:
## Purpose

When `write_paimon` is given a Ray Dataset, Ray's default round-robin
block distribution scatters rows that share the same `(partition,
bucket)` across many Ray tasks. Each task opens its own writer and emits
its own data file, so the write produces
`partitions × buckets × ray_tasks` files instead of the
`partitions × buckets` the writer would naturally produce.

Spark and Flink already cluster rows by `(partition, bucket)` before
writing — see `PaimonSparkWriter.repartitionByPartitionsAndBucket` and
the `RowAssignerChannelComputer` / `RowWithBucketChannelComputer` chain.
This PR brings the same pre-clustering to the Ray path.

## Linked Issue

N/A

## Effect

Two new keyword-only parameters on `write_paimon`:

- **`shuffle: bool = False`** — for HASH_FIXED tables, group rows by
  `(partition_keys..., bucket)` via Ray's `groupby` / `map_groups` so
  each `(partition, bucket)` lands in one Ray task. Bucket assignment
  is computed with `FixedBucketRowKeyExtractor`, the same extractor the
  writer uses, so the shuffle-time bucket is byte-equivalent to the
  writer's. Non-HASH_FIXED tables log a warning and write as before.

- **`num_blocks: Optional[int] = None`** — optional Ray output block
  count. With `shuffle=True` it is a parallelism hint for the groupby;
  with `shuffle=False` it triggers a plain Ray block rebalance.

Defaults preserve the previous behaviour, so no existing caller is
affected.

## Tests

- `pypaimon/tests/test_ray_shuffle_helper.py` — 8 unit tests covering
  the bucket-key UDF (column type, empty input, multi-chunk combine)
  and every no-op / soft-fallback branch.

- `pypaimon/tests/ray_repartition_test.py` — 7 end-to-end tests:
  - default `shuffle=False` roundtrip equality
  - `shuffle=True` roundtrip on a HASH_FIXED PK table
  - `shuffle=True` on a partitioned HASH_FIXED PK table (post-groupby
    schema integrity check)
  - file-count reduction on a multi-block HASH_FIXED write
  - soft fallback for BUCKET_UNAWARE + warning emitted
  - `num_blocks=0` raises `ValueError`
  - `num_blocks`-only plain block rebalance

Existing Ray tests (`ray_integration_test.py`, `ray_data_test.py`)
remain green.

## API & Format Impact

- Public API: adds two new keyword-only parameters with safe defaults
  to `pypaimon.ray.write_paimon`. No signature break.
- File format: unchanged. The transient `__paimon_bucket__` column is
  stripped before the sink sees the dataset, so on-disk layout is
  unaffected.

## Documentation

- `docs/content/pypaimon/ray-data.md` — new section explaining the
  small-file problem and the `shuffle` / `num_blocks` options.

### DIFF
--- a/paimon-python/pypaimon/ray/ray_paimon.py
+++ b/paimon-python/pypaimon/ray/ray_paimon.py
@@ -107,30 +107,22 @@ def write_paimon(
     catalog_options: Dict[str, str],
     *,
     overwrite: bool = False,
-    shuffle: bool = False,
-    override_num_blocks: Optional[int] = None,
     concurrency: Optional[int] = None,
     ray_remote_args: Optional[Dict[str, Any]] = None,
 ) -> None:
     """Write a Ray Dataset to a Paimon table.
+
+    For HASH_FIXED tables, rows are automatically clustered by
+    ``(partition_keys..., bucket)`` before writing so that each
+    (partition, bucket) lands in a single Ray task. This avoids the
+    small-file storm that Ray's default round-robin distribution would
+    otherwise produce. No user configuration is required.
 
     Args:
         dataset: The Ray Dataset to write.
         table_identifier: Full table name, e.g. ``"db_name.table_name"``.
         catalog_options: Options passed to ``CatalogFactory.create()``.
         overwrite: If ``True``, overwrite existing data in the table.
-        shuffle: When ``True`` and the target is a HASH_FIXED table, cluster
-            rows by ``(partition_keys..., bucket)`` so each (partition,
-            bucket) lands in one Ray task — reduces the small-file count
-            for distributed writes. Non-HASH_FIXED tables log a warning
-            and fall back to no-shuffle. Defaults to ``False`` (Ray's
-            default round-robin distribution).
-        override_num_blocks: Optional Ray output block count. Must be
-            ``>= 1`` when set. With ``shuffle=True``, used as a
-            parallelism hint for the groupby shuffle; with
-            ``shuffle=False``, triggers a plain block rebalance to that
-            count. ``None`` (default) skips the rebalance. Mirrors the
-            ``override_num_blocks`` parameter on :func:`read_paimon`.
         concurrency: Optional max number of Ray write tasks to run concurrently.
         ray_remote_args: Optional kwargs passed to ``ray.remote`` in write tasks.
     """
@@ -138,19 +130,10 @@ def write_paimon(
     from pypaimon.ray.shuffle import maybe_apply_repartition
     from pypaimon.write.ray_datasink import PaimonDatasink
 
-    if override_num_blocks is not None and override_num_blocks < 1:
-        raise ValueError(
-            "override_num_blocks must be at least 1, got {}".format(
-                override_num_blocks)
-        )
-
     catalog = CatalogFactory.create(catalog_options)
     table = catalog.get_table(table_identifier)
 
-    dataset, _ = maybe_apply_repartition(
-        dataset, table,
-        shuffle=shuffle, override_num_blocks=override_num_blocks,
-    )
+    dataset = maybe_apply_repartition(dataset, table)
 
     datasink = PaimonDatasink(table, overwrite=overwrite)
 
--- a/paimon-python/pypaimon/ray/ray_paimon.py
+++ b/paimon-python/pypaimon/ray/ray_paimon.py
@@ -107,6 +107,8 @@ def write_paimon(
     catalog_options: Dict[str, str],
     *,
     overwrite: bool = False,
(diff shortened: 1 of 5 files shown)


paimon-python/pypaimon/ray/shuffle.py
```diff
@@ -0,0 +1,157 @@
...
+if TYPE_CHECKING:
+    import ray.data
+
+    from pypaimon.table.table import Table
+
+logger = logging.getLogger(__name__)
+
+# Transient column the helper appends before the groupby and strips
+# afterwards. Sink-side schema is identical to caller-provided schema.
+BUCKET_KEY_COL = "__paimon_bucket__"
```
leaves12138, line 53: `__paimon_bucket__` can be a valid user/table column name. With a fixed transient name, `batch.append_column()` creates a duplicate Arrow field when the input already has this column; in PyArrow this produces duplicate names and `drop_columns([BUCKET_KEY_COL])` then raises `KeyError` (or, depending on the dataset backend/version, could remove the user's real column). As a result, `write_paimon(..., shuffle=True)` breaks for an otherwise valid table schema. Could we generate a temporary bucket column name that is guaranteed not to collide with the table/dataset columns (or fail early with a clear error), and add a regression test for a table containing `__paimon_bucket__`?

leaves12138 (changes requested):
Thanks for the patch. The overall approach looks reasonable and the existing CI is green, but I found one correctness issue in the transient bucket column handling. Please fix this before merge.

paimon-python/pypaimon/ray/shuffle.py
TheR1sing3un: > `__paimon_bucket__` can be a valid user/table column name. With a fixed transient name, `batch.append_column()` creates a duplicate Arrow field when the input already has this column; in PyArrow this produces duplicate names and `drop_columns([BUCKET_KEY_COL])` then raises `KeyError` (or, depending on the dataset backend/version, could remove the user's real column). As a result, `write_paimon(..., shuffle=True)` breaks for an otherwise valid table schema. Could we generate a temporary bucket column name that is guaranteed not to collide with the table/dataset columns (or fail early with a clear error), and add a regression test for a table containing `__paimon_bucket__`?

nice catch! thanks! fixed it already.

TheR1sing3un:
> Thanks for the patch. The overall approach looks reasonable and the existing CI is green, but I found one correctness issue in the transient bucket column handling. Please fix this before merge.

I have fixed it. Please review it again. Thank you

paimon-python/pypaimon/ray/ray_paimon.py
```diff
@@ -117,15 +119,39 @@ def write_paimon(
         table_identifier: Full table name, e.g. ``"db_name.table_name"``.
         catalog_options: Options passed to ``CatalogFactory.create()``.
         overwrite: If ``True``, overwrite existing data in the table.
+        shuffle: When ``True`` and the target is a HASH_FIXED table, cluster
```
JingsongLi: Why we need to add this option? Just shuffle for bucketed table?

paimon-python/pypaimon/ray/ray_paimon.py
```diff
@@ -117,15 +119,39 @@ def write_paimon(
...
+            rows by ``(partition_keys..., bucket)`` so each (partition,
+            bucket) lands in one Ray task — reduces the small-file count
+            for distributed writes. Non-HASH_FIXED tables log a warning
+            and fall back to no-shuffle. Defaults to ``False`` (Ray's
+            default round-robin distribution).
+        override_num_blocks: Optional Ray output block count. Must be
```
JingsongLi: It can be directly derived from the number of buckets, it is not necessarily necessary

JingsongLi (changes requested):

paimon-python/pypaimon/ray/ray_paimon.py
TheR1sing3un: > Why we need to add this option? Just shuffle for bucketed table?

done~

paimon-python/pypaimon/ray/ray_paimon.py
TheR1sing3un: > It can be directly derived from the number of buckets, it is not necessarily necessary

done~

TheR1sing3un:
@JingsongLi Thanks for your review, all comments fixed. Please take a look again!

JingsongLi (approved):
+1

PR #7814, opened 2026-05-11
[core] Make ConflictDetection error message more readable

yuzelin:
### Purpose
Current error message cannot get the readable partition info.


JingsongLi:
+1

PR #7815, opened 2026-05-11
[python] Support StreamTableUpdate of data evolution for StreamWriteBuilder

SteNicholas:
### Purpose                                                               

Support data evolution for `StreamWriteBuilder` by introducing `StreamTableUpdate` that exposes `update_by_arrow_with_row_id` and `upsert_by_arrow_with_key` APIs with `commit_identifier` support. Previously, `StreamWriteBuilder.new_update()` raised `ValueError("StreamWriteBuilder.new_update() not supported.")`, making data-evolution table-update operations unavailable in streaming mode.

Key changes:

- `TableUpdate` → `BatchTableUpdate` / `StreamTableUpdate` split: Refactored `TableUpdate` into a shared base with `_update_by_arrow_with_row_id` and `_upsert_by_arrow_with_key` internals that accept a `commit_identifier`. `BatchTableUpdate` delegates to `BATCH_COMMIT_IDENTIFIER` (no commit_identifier parameter); `StreamTableUpdate` requires an explicit commit_identifier per call — mirroring the existing `TableWrite` / `TableCommit` batch/stream pattern.
- `TableCommit` cleanup: Moved `_check_committed` / `batch_committed` guard down from the shared `TableCommit` base into `BatchTableCommit`, since stream-mode commits are reusable by design. Improved log message and docstrings.
- `TableUpdateByRowId` / `TableUpsertByKey`: Both now accept `commit_identifier` via constructor instead of hard-coding `BATCH_COMMIT_IDENTIFIER` internally. `TableUpsertByKey._do_appends` switched from `BatchTableWrite` to `StreamTableWrite` so that appended commit messages carry the correct identifier.
- `WriteBuilder`: `StreamWriteBuilder.new_update()` returns `StreamTableUpdate` instead of raising.                                                                                          

### Tests

Stream-mode tests in `table_update_test.py` and `table_upsert_by_key_test.py` cover `update_by_arrow_with_row_id` and `upsert_by_arrow_with_key` under streaming, including: single-column update, multi-column update, partial-row update, partitioned tables, consecutive commits with correct snapshot identifiers, and concurrent stream commits.

### DIFF
--- a/paimon-python/pypaimon/write/table_commit.py
+++ b/paimon-python/pypaimon/write/table_commit.py
@@ -27,7 +27,17 @@ from pypaimon.write.file_store_commit import FileStoreCommit
 
 
 class TableCommit:
-    """Python implementation of BatchTableCommit for batch writing scenarios."""
+    """Common base for batch and stream table commits.
+
+    Owns the underlying :class:`FileStoreCommit` and provides the shared
+    :meth:`_commit` implementation. The concrete subclasses differ only in
+    their public ``commit`` signature and lifecycle constraints:
+
+    * :class:`BatchTableCommit` accepts no ``commit_identifier`` and may be
+      committed at most once.
+    * :class:`StreamTableCommit` requires an explicit ``commit_identifier``
+      on every call and may be reused for many commits.
+    """
 
     def __init__(self, table, commit_user: str, static_partition: Optional[dict]):
         from pypaimon.table.file_store_table import FileStoreTable
@@ -42,11 +52,8 @@ class TableCommit:
             raise RuntimeError("Table does not provide a SnapshotCommit instance")
 
         self.file_store_commit = FileStoreCommit(snapshot_commit, table, commit_user)
-        self.batch_committed = False
 
     def _commit(self, commit_messages: List[CommitMessage], commit_identifier: int = BATCH_COMMIT_IDENTIFIER):
-        self._check_committed()
-
         non_empty_messages = [msg for msg in commit_messages if not msg.is_empty()]
 
         if self.overwrite_partition is not None:
@@ -66,7 +73,7 @@ class TableCommit:
             if not non_empty_messages:
                 return
             logger.info(
-                "Committing batch table %s, %d non-empty messages",
+                "Committing table %s, %d non-empty messages",
                 self.table.identifier, len(non_empty_messages)
             )
             self.file_store_commit.commit(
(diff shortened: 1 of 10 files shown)


paimon-python/pypaimon/write/table_update_by_row_id.py
```diff
@@ -40,53 +42,53 @@ class TableUpdateByRowId:
...
-
-        total_row_count = sum(first_row_id_to_row_count_map.values())
+                if file.first_row_id is None or file.file_name.endswith('.blob'):
+                    continue
+                entry = index.get(file.first_row_id)
+                if entry is None:
+                    index[file.first_row_id] = (split, [file])
+                else:
+                    entry[1].append(file)
+                total_row_count += file.row_count
```
XiaoHongbo-Hope: This over-counts rows after data evolution. Multiple physical files can share the same first_row_id, so
  summing every file makes total_row_count too large and lets invalid _ROW_ID values pass

SteNicholas: @XiaoHongbo-Hope, yes, it's right. I have aligned with the Java implementation and fixed this summary.

leaves12138 (approved):
Reviewed the StreamTableUpdate data-evolution changes. The batch/stream split is consistent with the existing write/commit API shape, the update/upsert paths now propagate the caller-provided stream commit identifier, and moving the one-shot guard to BatchTableCommit makes StreamTableCommit reusable as intended.

I also applied the PR diff locally and verified the changed Python tests:

- `python3.8 -m py_compile` on the changed implementation and test files
- `PYTHONPATH=$PWD python -m pytest -q pypaimon/tests/table_commit_test.py` → 8 passed
- `PYTHONPATH=$PWD python -m pytest -q pypaimon/tests/table_update_test.py pypaimon/tests/table_upsert_by_key_test.py` → 82 passed

LGTM.

JingsongLi (commented):
Can you also update documentation?

paimon-python/pypaimon/write/table_commit.py
```diff
@@ -99,8 +108,19 @@ def truncate_partitions(self, partitions: List[Dict[str, str]]) -> None:
...
 
 class StreamTableCommit(TableCommit):
+    """Stream-mode commit; reusable across many commit rounds.
+
+    Each call must be tagged with a monotonically increasing
+    ``commit_identifier`` — analogous to
+    :meth:`StreamTableWrite.prepare_commit`.
+    """
 
     def commit(self, commit_messages: List[CommitMessage], commit_identifier: int = BATCH_COMMIT_IDENTIFIER):
```
XiaoHongbo-Hope: Should we update the default value of `commit_identifier` too?

SteNicholas: @XiaoHongbo-Hope, removed the default value.

JingsongLi (approved):
+1 cc @XiaoHongbo-Hope to take a look again.

XiaoHongbo-Hope:
+1

PR #7817, opened 2026-05-11
[core] Optimize Like for wildcard patterns via chained indexOf

ArnavBalyan:
### Purpose
 - Today all Like patterns with `%%` go through the Java regex matcher which is inefficient and can be improved by string matching algorithms. We only cater to basic cases of startwith/endwith, everything else falls through java regex. 
 - Introduce new chainedpatternmatcher which can tokenize %% into segments and handle with byte level Memorysegmentutils per row. 
 - 13x improvement for 10k input row for `xyz%xyz` type strings over java regex, Can be enabled with `paimon.predicate.like.chain-checker.enabled`. 
 - In the future we can further improve with Boyer Moore Sunday algo similar to Duckdb pattern matching. Similar matching is used by Flink and other analytical engines
### Tests
 - UT

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/predicate/Like.java
+++ b/paimon-common/src/main/java/org/apache/paimon/predicate/Like.java
@@ -69,8 +69,10 @@ public class Like extends LeafBinaryFunction {
             Object literal = optimized.get().getValue();
             return field -> func.test(type, field, literal);
         }
-        // TODO optimize for chain checkers when there is no '_'
-        // TODO for example: "abc%def%","%abc%def","%abc%def%","abc%def"
+        LikeChainChecker chainChecker = LikeChainChecker.compile(patternLiteral.toString());
+        if (chainChecker != null) {
+            return chainChecker::test;
+        }
         String regex = sqlToRegexLike(patternLiteral.toString(), null);
         Pattern pattern = Pattern.compile(regex);
         return input -> pattern.matcher(input.toString()).matches();
--- a/paimon-common/src/main/java/org/apache/paimon/predicate/LikeChainChecker.java
+++ b/paimon-common/src/main/java/org/apache/paimon/predicate/LikeChainChecker.java
@@ -1,0 +1,166 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.paimon.predicate;
+
+import org.apache.paimon.data.BinaryString;
+import org.apache.paimon.memory.MemorySegmentUtils;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Like matcher for patterns containing only literal text and % wildcards and no escape characters.
+ * Avoids the regex engine for %% clauses.
+ *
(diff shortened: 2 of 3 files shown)


ArnavBalyan:
cc @JingsongLi thanks! :)

leaves12138 (approved):
Thanks for the update. I reviewed the chain-checker path against the existing regex fallback, including anchoring, consecutive `%`, empty patterns, escaped-pattern fallback, UTF-8 byte matching through `MemorySegmentUtils.find`, and the `Like` cache wiring.

I also ran the focused tests locally:

```
mvn -pl paimon-common -Dtest=LikeChainCheckerTest,LikeWiringTest -DfailIfNoTests=false test
mvn -pl paimon-common -Dtest=PredicateTest#executeLike test
```

Both passed. I don't see blockers from my side.

paimon-common/src/main/java/org/apache/paimon/predicate/Like.java
```diff
@@ -40,12 +41,31 @@ public class Like extends LeafBinaryFunction {
 
     public static final Like INSTANCE = new Like();
 
+    static final String CHAIN_CHECKER_PROPERTY = "paimon.predicate.like.chain-checker.enabled";
+
     private static final Cache<BinaryString, Filter<BinaryString>> CACHE =
             Caffeine.newBuilder().softValues().executor(Runnable::run).build();
 
+    private static volatile boolean chainCheckerEnabled =
```
JingsongLi: Why add this option?

paimon-common/src/main/java/org/apache/paimon/predicate/Like.java
```diff
@@ -40,12 +41,31 @@ public class Like extends LeafBinaryFunction {
...
+            Boolean.parseBoolean(System.getProperty(CHAIN_CHECKER_PROPERTY, "false"));
+
+    private static final AtomicLong chainCheckerInvocations = new AtomicLong();
```
JingsongLi: Why add this? Just for test? It is better to not bother official codes.

paimon-common/src/main/java/org/apache/paimon/predicate/Like.java
ArnavBalyan: Had added this to be conservative and let the user control the behaviour. Tbh this is not needed, we can centrally control it, removed.

paimon-common/src/main/java/org/apache/paimon/predicate/Like.java
ArnavBalyan: Removed thanks, this was temporary to check the wiring

ArnavBalyan:
> Thanks for the update. I reviewed the chain-checker path against the existing regex fallback, including anchoring, consecutive `%`, empty patterns, escaped-pattern fallback, UTF-8 byte matching through `MemorySegmentUtils.find`, and the `Like` cache wiring.
> 
> I also ran the focused tests locally:
> 
> ```
> mvn -pl paimon-common -Dtest=LikeChainCheckerTest,LikeWiringTest -DfailIfNoTests=false test
> mvn -pl paimon-common -Dtest=PredicateTest#executeLike test
> ```
> 
> Both passed. I don't see blockers from my side.

Thanks for taking a look @leaves12138 !

leaves12138 (approved):
I did a second pass against Flink's `SqlLikeChainChecker` and the planner-side `LikeCallGen` selection logic.

For the pure literal + `%` cases where the chain checker is applicable, the Paimon implementation is semantically aligned with Flink: minimum byte length check, left/right anchoring, consecutive `%`, empty pattern / `%` / `%%`, ordered middle segment search, and UTF-8 byte-level matching all look correct. The implementation is structured differently from Flink (`segments + anchors` instead of `begin/end/middle`), but the matching condition is equivalent.

The main intentional difference is escape handling: Flink's two-argument LIKE treats `\` as a normal literal and only rewrites patterns for an explicit `ESCAPE`; Paimon's existing LIKE regex path uses `\` as the default escape character, so this PR safely rejects any pattern containing `\` and falls back to the existing regex path. That difference is expected and avoids changing Paimon's escape semantics.

I also re-ran the focused tests on the latest head:

```
mvn -pl paimon-common -Dtest=LikeChainCheckerTest,PredicateTest#executeLike test
```

Additionally, I used a temporary scratch parity test with Flink-applicable quoted cases, regex-special literals, Unicode/emoji, and 5k random no-escape patterns against SQL-LIKE regex semantics; it passed. No new blocker from my side.

JingsongLi:
+1

PR #7818, opened 2026-05-11
[spark] Convert exceptions to NoSuchTable for file-format namespace

Zouxxyy:
### Purpose

When using Paimon as the Spark catalog, `parquet.\`path\`` queries may fail with non-existence
exceptions (e.g., permission denied). This change converts all exceptions for file-format
namespaces to `NoSuchTableException`, allowing Spark's `ResolveSQLOnFile` to handle them.

### Tests

Added test in `PaimonQueryTest.scala`.


JingsongLi (approved):
+1

PR #7819, opened 2026-05-12
[core] Drop global index external path with table

leaves12138:
### Purpose

When dropping a table, Paimon already collects and deletes paths from `data-file.external-paths`. However, it missed `global-index.external-path`, so global index files could be left behind after `dropTable`.

### Changes

- Include `GLOBAL_INDEX_EXTERNAL_PATH` when collecting schema external paths during table drop.
- Add catalog coverage verifying both data file external paths and global index external paths are removed.

### Tests

- `mvn -pl paimon-core -Pfast-build -Dtest=FileSystemCatalogTest#testDropTableWithExternalPaths,CachingCatalogTest#testDropTableWithExternalPaths test`
- `mvn -pl paimon-core -DskipTests spotless:check`


JingsongLi:
+1

PR #7823, opened 2026-05-12
[python] Fix RecursionError in vector search with many splits

XiaoHongbo-Hope:
### Purpose
#7752 fixed the closure chain recursion in BTree path, but Lumina / Tantivy  path still nests score_getter closures, causing the same issue.  This PR fixes it.


leaves12138 (approved):
LGTM. I checked the updated dict-based merge against the previous recursive `or_` semantics:

- duplicate row ids still keep the earlier/left result, matching the old chained merge behavior;
- `top_k` is still applied after merging all split results, so high-scored rows from later splits are retained;
- vector search and full-text search now use the same non-recursive merge strategy, which avoids the deep `score_getter` chain for many splits.

I also ran the focused `vector_search_filter_test.py` suite locally with a temporary pure-Python bitmap stub because this container cannot use the native `pyroaring.BitMap64` dependency; all 11 tests passed under that setup.

JingsongLi:
+1

Issue #7064, opened 2026-01-16
[Bug] When executing an insert operation on a Hive partitioned table, an array out-of-bounds error occurs


3216670078:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Paimon version

1.3.1

### Compute Engine

hive3.1.3

### Minimal reproduce step

CREATE TABLE test_paimon2 (
    id INT,
    name STRING,
    salary DOUBLE,
    department STRING
)
partitioned by (dt string)
STORED BY 'org.apache.paimon.hive.PaimonStorageHandler'
TBLPROPERTIES (
    'primary-key' = 'id,dt',
    'bucket' = '-1',
    'file.format' = 'parquet'
);

INSERT INTO TABLE test_paimon2
partition (dt='2026')
VALUES 
(1, 'Alice', 5000.0, 'IT'),  
(2, 'Bob', 6000.0, 'HR'),
(3, 'Charlie', 5500.0, 'IT');

### What doesn't meet your expectations?

Diagnostic Messages for this Task:
Error: java.lang.RuntimeException: org.apache.hadoop.hive.ql.metadata.HiveException: Hive Runtime Error while processing writable (null)
        at org.apache.hadoop.hive.ql.exec.mr.ExecMapper.map(ExecMapper.java:163)
        at org.apache.hadoop.mapred.MapRunner.run(MapRunner.java:54)
        at org.apache.hadoop.hive.ql.exec.mr.ExecMapRunner.run(ExecMapRunner.java:37)
        at org.apache.hadoop.mapred.MapTask.runOldMapper(MapTask.java:465)
        at org.apache.hadoop.mapred.MapTask.run(MapTask.java:349)
        at org.apache.hadoop.mapred.YarnChild$2.run(YarnChild.java:174)
        at java.security.AccessController.doPrivileged(Native Method)
        at javax.security.auth.Subject.doAs(Subject.java:422)
        at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1729)
        at org.apache.hadoop.mapred.YarnChild.main(YarnChild.java:168)
Caused by: org.apache.hadoop.hive.ql.metadata.HiveException: Hive Runtime Error while processing writable (null)
        at org.apache.hadoop.hive.ql.exec.MapOperator.process(MapOperator.java:568)
        at org.apache.hadoop.hive.ql.exec.mr.ExecMapper.map(ExecMapper.java:154)
[...]

PR #7824, opened 2026-05-12
[hive] Fix insert into static partitions on managed Paimon tables

ArnavBalyan:
### Purpose
 - Insert into static partition against a Hive managed table fails with `ArrayIndexOutOfBoundsException` in `TableWriteImpl.checkNullability`. 
 - This is because Hive sends data rows without the static partition columns and instead passes their values in `path`/`partition_columns` serde property.
 - PaimonOutputFormat.getHiveRecordWriter does not respect these properties, and row arrives incomplete.
 -  Read the partition column names, parse the values and wrap the writer which is aware of the extra properies and can construct the row properly. 
 - Closes user reported bug #7064.
### Tests
 - UT

### DIFF
--- a/paimon-hive/paimon-hive-connector-common/src/main/java/org/apache/paimon/hive/mapred/PaimonOutputFormat.java
+++ b/paimon-hive/paimon-hive-connector-common/src/main/java/org/apache/paimon/hive/mapred/PaimonOutputFormat.java
@@ -19,10 +19,15 @@
 package org.apache.paimon.hive.mapred;
 
 import org.apache.paimon.CoreOptions;
+import org.apache.paimon.data.GenericRow;
 import org.apache.paimon.hive.RowDataContainer;
+import org.apache.paimon.schema.TableSchema;
 import org.apache.paimon.table.FileStoreTable;
 import org.apache.paimon.table.sink.BatchTableWrite;
 import org.apache.paimon.table.sink.BatchWriteBuilder;
+import org.apache.paimon.types.DataField;
+import org.apache.paimon.utils.PartitionPathUtils;
+import org.apache.paimon.utils.TypeUtils;
 
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
@@ -37,7 +42,10 @@ import org.apache.hadoop.mapred.TaskAttemptID;
 import org.apache.hadoop.util.Progressable;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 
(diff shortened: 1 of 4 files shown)


ArnavBalyan:
cc @JingsongLi thanks!

ArnavBalyan:
cc @JingsongLi gentle reminder if you could PTAL thanks!

ArnavBalyan:
cc @leaves12138 @JingsongLi gentle reminder thanks ! :)

JingsongLi (commented):
## Review: [hive] Fix insert into static partitions on managed Paimon tables

Overall this is a well-structured fix for a legitimate bug (#7064). The approach of detecting static partition columns via serde properties and wrapping the writer to reconstruct full rows is sound. A few observations:

### Correctness

1. **Partition column ordering assumption**: `PartitionedRecordWriter` uses `JoinedRow(source, partitionRow)` which assumes partition columns are always appended after data columns. This is correct for Hive-managed tables (Hive always places `PARTITIONED BY` columns at the end of the schema), but if a table were originally created via Flink/Spark with partition columns in the middle and later accessed from Hive, this could silently produce incorrect rows. Consider adding a defensive check that validates the partition column indices are actually at the tail of the schema in `buildStaticPartitionRow`.

2. **Mixed static/dynamic partitions**: If a user writes `PARTITION (region='us', year)` (one static, one dynamic), `META_TABLE_PARTITION_COLUMNS` will list both columns, but the path will only contain `region=us`. The `lookupCaseInsensitive` for `year` returns null, causing `buildStaticPartitionRow` to return null, which falls through to the unwrapped writer. That writer will then also fail because Hive still strips the static partition column from the data. This is not a regression (it was already broken), but it might be worth a code comment noting this limitation.

3. **Separator in `META_TABLE_PARTITION_COLUMNS`**: The split uses `org.apache.paimon.fs.Path.SEPARATOR` which is `"/"`. This matches the Hive convention for this property. Correct, but somewhat fragile -- a named constant or comment clarifying why `/` is the right delimiter here would improve readability.

### Design

4. **Direct access to `inner.batchTableWrite().write(toWrite)`**: `PartitionedRecordWriter` bypasses `PaimonRecordWriter.write()` and calls the underlying `BatchTableWrite` directly. Currently this is fine since `PaimonRecordWriter.write()` does nothing beyond unwrapping `RowDataContainer` and forwarding. However, this creates tight coupling -- if `PaimonRecordWriter.write()` later gains additional logic (metrics, validation, etc.), `PartitionedRecordWriter` would silently miss it. A safer pattern would be to accept an `InternalRow` transformer in `PaimonRecordWriter`, or at minimum add a package-private `writeRow(InternalRow)` method that both paths invoke.

5. **`forWriteOnly` extraction**: Nice refactoring to separate the `WRITE_ONLY` option application from writer creation. Clean.

### Minor

6. The exception wrapping differs: `PaimonRecordWriter.write()` throws `RuntimeException` on write failure, while `PartitionedRecordWriter.write()` wraps in `IOException`. The inconsistency is minor but could confuse error handling upstream.

7. Test coverage is good -- unit tests for `buildStaticPartitionRow` cover the key scenarios (typed values, unpartitioned tables, missing path segments), and the integration test validates end-to-end correctness.

### Summary

The fix correctly addresses the reported bug for the common case (fully static partition inserts on Hive-managed tables). The main suggestion is to add a defensive assertion that partition columns are indeed at the schema tail, and to reduce the coupling between `PartitionedRecordWriter` and `PaimonRecordWriter`'s internals.

JingsongLi (changes requested):
-1 It seems that there are really issues with the testing

ArnavBalyan:
Thanks for the review! Have addressed all comments

  **1. Partition column ordering assumption**
  Added a check that throws a clear error when partition columns aren't the trailing columns of the schema. Covered by a new unit test.

  **2. Mixed static/dynamic partitions**
  Documented why we fall back to the unwrapped writer when a partition key has no value in the path, and flags this as a known unsupported case.

  **3. Separator in the partition columns property**
  Extracted the forward slash into a constant with a comment explaining it's the Hive metastore convention, not the OS path separator.

  **4. Direct access to the underlying batch write**
  Added a write method on the inner writer. Both the wrapped and unwrapped paths now go through this, so any future logic added to the write path won't be silently bypassed.

  **5. Inconsistent exception wrapping**
  Aligned both write paths to throw the same exception type, so error handling upstream is consistent.

leaves12138 (changes requested):
I re-reviewed the latest head (`13980ce`) after your update. I still think the current patch needs changes before merge.

The main issue from the previous review is still present: `PartitionedRecordWriter` always reconstructs the row as `JoinedRow(source, partitionRow)`, so it assumes the static partition columns are exactly the trailing columns of the Paimon table schema and in the same order as `META_TABLE_PARTITION_COLUMNS`. However, `buildStaticPartitionRow()` currently only looks up partition values and field types; it does not validate that the schema tail actually matches those partition columns. This can silently write incorrect rows for a table whose partition columns are not at the tail, or whose order differs from the Hive property.

Please either:

1. add a defensive validation before wrapping the writer that the last N fields of `schema.fields()` match the partition keys case-insensitively and in order, and fail with a clear exception otherwise; or
2. reconstruct a full row by schema field index instead of blindly appending the partition row.

Please also add a unit test for the non-tail / wrong-order case. The response says this check was added, but I do not see it in the current code.

There is also still an edge case for mixed static/dynamic partitions: if Hive provides only some partition values in the path, `buildStaticPartitionRow()` returns `null`, which falls back to the unwrapped writer even though the incoming row may still be missing the static partition columns. If mixed static/dynamic partitions are not supported in this patch, it would be better to fail explicitly instead of silently falling back to the old path.

Finally, the current status rollup still has a failed `UTCase and ITCase Others` check and a cancelled JDK 11 Others check, so this should not be approved yet.

ArnavBalyan:
Hi @leaves12138 thanks for the re-review, have addressed both comments and the CI is now green. Could you please take a look again thanks!

JingsongLi (commented):
LGTM, thanks.

Issue #6547, opened 2025-11-06
[Bug] lastOptimizedCompaction in UniversalCompaction doesn't work as expected


qingfei1994:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Paimon version

master

### Compute Engine

Flink 1.18

### Minimal reproduce step

1.  create a paimon table with optimization interval and enable debug log
```
CREATE TABLE IF NOT EXISTS table_a (
    task_id varchar,
   `value` bigint,
    metric_name varchar,
     PRIMARY KEY (`torrent_task_id`) NOT ENFORCED
) WITH (
   'merge-engine' = 'partial-update',
   'changelog-producer' = 'lookup',
   'bucket' = '2',
   'compaction.optimization-interval' = '24 h'
 );

   insert into table_a
  select metric_name,`value`,task_id from kafka_table_b;
```

3. producing records to kafka 
- produce records to kafka when the first checkpoint is triggered.
- don't produce any records for the next several checkpoints
- reproduce records to  kafka 

4.we'll see logs like that

```
 DEBUG org.apache.paimon.operation.AbstractFileStoreWrite           [] - Closing writer for partition org.apache.paimon.data.BinaryRow@9c67b85d, bucket 32. Writer's last modified identifier is 3300, while current commit identifier is 3302.
 DEBUG org.apache.paimon.operation.AbstractFileStoreWrite           [] - Creating writer for partition org.apache.paimon.data.BinaryRow@9c67b85d, bucket 32
 DEBUG org.apache.paimon.mergetree.compact.UniversalCompaction      [] - Universal compaction due to optimized compaction interval
```

### What doesn't meet your expectations?

Whenever MergeTreeWriter is recreated, as lastOptimizedCompaction is null in UniversalCompaction instance, full compaction will be triggered even if it is triggered before.

I think lastOptimizedCompaction may need to be persisted or retrieved from the file list, instead of being null all the time when UniversalCompaction instance is recreated.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #7825, opened 2026-05-12
[core] Honor the compaction optimization interval across Merge tree writer re-creation

ArnavBalyan:
### Purpose
   - Today `compaction.optimization-interval` is not respected. When this interval set, the full compaction is supposed to run at max once per time interval. But this actually runs on every writer recreate much more frequently than the time interval, which can silently inflate i/o and CPU
   - The last compaction ts is an inmemory field that starts as `null`. The `MergeTreeWriter` is recreated many times, causing the compaction fires more frequently.
   - This can be avoided by maintaining the ts of last compaction. Today this ts is lost across writer recreations of MergeTreeWriter, but can be tracked using restoreFiles. 
 - Fix this by computing the most recent creationTimeEpochMillis of files at the max LSM level from restoreFiles and passing it to full compaction create.
 - Closes user reported bug #6547.
### Tests
 - UT


ArnavBalyan:
cc @leaves12138 @JingsongLi thanks! :)

ArnavBalyan:
cc @JingsongLi gentle reminder thanks!

JingsongLi (approved):
+1

PR #7827, opened 2026-05-12
[core] Allow blob inline fields without blob-field

leaves12138:
### Purpose

`blob-descriptor-field` and `blob-view-field` already fully describe descriptor/view BLOB semantics, so users should not need to repeat those fields in `blob-field`.

### Changes

- Relax schema validation so `blob-descriptor-field` and `blob-view-field` no longer need to be subsets of `blob-field`.
- Treat fields listed in `blob-descriptor-field` or `blob-view-field` as BLOB fields during Flink and Spark table creation.
- Keep the old style compatible: configuring the same field in `blob-field` still works.
- Update docs and generated core option descriptions.
- Replace old failure tests with coverage for descriptor/view fields without `blob-field`.

### Tests

- `mvn -pl paimon-core -am -Pfast-build -DfailIfNoTests=false -Dtest=BlobTableTest#testBlobInlineFieldCanDeclareBlobWithoutBlobField test`
- `mvn -pl paimon-flink/paimon-flink-common -am -Pfast-build -DfailIfNoTests=false -Dtest=BlobTableITCase#testWriteBlobViewWithBuiltInFunction+testBlobInlineFieldCanDeclareBlobWithoutBlobField test`
- `mvn -Pspark3 -pl paimon-spark/paimon-spark-common,paimon-spark/paimon-spark3-common,paimon-spark/paimon-spark-ut -am -Pfast-build -DfailIfNoTests=false -DwildcardSuites=org.apache.paimon.spark.sql.BlobTestWithV2Write -Dtest=none test`
- `hugo -b "" --quiet`
- `git diff --check`
- `JAVA_HOME=/opt/jdk-11.0.18.jdk/Contents/Home mvn -pl paimon-flink/paimon-flink-common -am -Pflink-2.2 -Pfast-build -DskipTests compile`

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/SparkCatalog.java
+++ b/paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/SparkCatalog.java
@@ -83,6 +83,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 import static org.apache.paimon.CoreOptions.FILE_FORMAT;
@@ -457,6 +458,7 @@ public class SparkCatalog extends SparkBaseCatalog
             StructType schema, Transform[] partitions, Map<String, String> properties) {
         Map<String, String> normalizedProperties = new HashMap<>(properties);
         List<String> blobFields = CoreOptions.blobField(properties);
+        Set<String> blobDescriptorFields = new CoreOptions(properties).blobDescriptorField();
         List<String> blobViewFields = CoreOptions.blobViewField(properties);
         String provider = properties.get(TableCatalog.PROP_PROVIDER);
         if (!usePaimon(provider)) {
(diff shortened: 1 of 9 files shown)


paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/SparkCatalog.java
```diff
@@ -491,7 +493,9 @@ private Schema toInitialSchema(
         for (StructField field : schema.fields()) {
             String name = field.name();
             DataType type;
-            if (blobFields.contains(name) || blobViewFields.contains(name)) {
+            if (blobFields.contains(name)
+                    || blobDescriptorFields.contains(name)
+                    || blobViewFields.contains(name)) {
```
XiaoHongbo-Hope, line 498: nit: Could we extract a shared blobTypeFields() helper like FlinkCatalog

XiaoHongbo-Hope:
+1

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/SparkCatalog.java
leaves12138: Yes， we can

PR #7830, opened 2026-05-12
[core] Support writing empty partitions

### LINKED ISSUE #7682: [Feature] Chained tables should allow set fallback rules for non-existent partitions.
### Search before asking

### Motivation

when using chain tables:
my partitions : 
branch_snapshot: 20260101
branch_delta: 20260102、20260104

so i using: `select count(1) from table where dt=20260103` is 0
but in a scenario, It should return the most recent partition from the beginning of the timeline, 
like above, should return data in 20260102
not 0.

### Solution

add a configuration, allow user to decide whether to fallback to the most recent partition from the beginning of the timeline

### Anything else?

_No response_

Aitozi:
### Purpose

This change adds support for writing an empty data file for static partition overwrite when the input produces no records and `write.empty.partition.enable` is enabled.

### Empty Partition Semantics

This behavior is consistent with Hive `INSERT OVERWRITE` semantics for static partitions. When overwriting a fully specified static partition, Hive materializes the target partition even if the input query produces no rows. In contrast, for dynamic partitions, only partitions actually produced by the query result are materialized.

Paimon needs to preserve the same semantic distinction. A partition is not only a by-product of existing data files; in some scenarios, the existence of the partition itself carries business meaning.

For example, in chain table scenarios, whether a Delta partition exists is determined by the partition metadata itself. If Paimon hides or skips a partition only because it contains no data rows, the semantic meaning of the Delta partition would be broken. An empty Delta partition still indicates that the partition has been generated and should be visible to downstream logic.

Therefore, when `write.empty.partition.enable` is enabled, Paimon writes an empty file for fully static partition overwrites whose input is empty, so that the partition can be materialized and discovered. For partial static or dynamic partition overwrites, Paimon does not synthesize empty partitions, because the final partition values are determined by the query result.

### Tests

close #7682


Aitozi:
CC @JingsongLi @Stefanietry

JingsongLi (changes requested):
This is a bit too tricky and has affected too much code. We should think of a simpler method.

Aitozi:
> This is a bit too tricky and has affected too much code. We should think of a simpler method.

@JingsongLi Do you have some good suggestions for this? 
I feel that expressing this type of empty partition within Paimon's existing architecture is a bit difficult. I think the biggest issue here is that Paimon doesn't have independent partition metadata, so it can only represent its partitions through files. Therefore, to indicate whether a partition exists, we need to represent it by creating empty files.

JingsongLi (commented):
It seems that the requirement is only to notify CommitListener of the partition information when Overwriting? I feel there is a simpler way to modify it.

Aitozi:
> It seems that the requirement is only to notify CommitListener of the partition information when Overwriting? I feel there is a simpler way to modify it.

I have updated the implementation:

1. Added a Writer interface to support writing EmptyFiles.
2. In the Flink and Spark commit phases, empty partitions are now committed directly during static overwrite.

Could you please help review it again? @JingsongLi

JingsongLi:
Thanks @Aitozi for the update. But I still think something like `_SUCCESS` file helps this scenario. Empty file is too dangerous to Paimon.

Aitozi:
> Thanks @Aitozi for the update. But I still think something like `_SUCCESS` file helps this scenario. Empty file is too dangerous to Paimon.

Thanks for your suggestion, closing it

PR #7832, opened 2026-05-12
[core] Support snapshot-based sequence ordering for primary-key tables

### LINKED ISSUE #7806: [Feature] Support snapshot-based sequence ordering for primary-key tables

JunRuiLee:
### Purpose
close https://github.com/apache/paimon/issues/7806
### Tests
  - SchemaValidationTest#testSnapshotSequenceOrderingHappyPath
  - SchemaValidationTest#testSnapshotSequenceOrderingRejectsSequenceField
  - SchemaValidationTest#testSnapshotSequenceOrderingRejectsNonPkTable
  - KeyValueWithLevelNoReusingSerializerSnapshotIdTest#testRoundTripWithSnapshotId
  - KeyValueWithLevelNoReusingSerializerSnapshotIdTest#testRoundTripWithoutSnapshotId
  - SortMergeSnapshotOrderingTest#testLaterSnapshotWinsOverHigherSequence
  - SortMergeSnapshotOrderingTest#testFallsBackToSequenceWhenSnapshotMissing
  - SortMergeSnapshotOrderingTest#testSameSnapshotFallsBackToSequence
  - SortMergeSnapshotOrderingTest#testStampedAlwaysBeatsUnstamped
  - PrimaryKeySimpleTableTest#testSnapshotSequenceOrdering
  - PrimaryKeySimpleTableTest#testSnapshotSequenceOrderingFallsBackToSequenceWithinSnapshot
  - PrimaryKeySimpleTableTest#testSnapshotSequenceOrderingCompactionPreservesInputSnapshotId
  - PrimaryKeySimpleTableTest#testSnapshotSequenceOrderingWithChangelogInput
  - PrimaryKeySimpleTableTest#testSnapshotSequenceOrderingWithChangelogLookup
  - PrimaryKeySimpleTableTest#testSnapshotSequenceOrderingDeleteFromLaterSnapshot

### DIFF
--- a/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
@@ -968,6 +968,24 @@ public class CoreOptions implements Serializable {
                     .withDescription("Specify the order of sequence.field.");
 
     @Immutable
+    public static final ConfigOption<Boolean> SEQUENCE_SNAPSHOT_ORDERING =
+            key("sequence.snapshot-ordering")
+                    .booleanType()
+                    .defaultValue(false)
+                    .withDescription(
+                            "When enabled, merge uses the commit snapshot id as the ordering key "
+                                    + "for primary-key conflicts: records from later snapshots "
+                                    + "always win. Designed for multi-writer scenarios on the same "
+                                    + "primary-key table where wall-clock sequence numbers cannot "
+                                    + "be globally ordered. The order of records within the same "
+                                    + "snapshot is not guaranteed. Mutually exclusive with "
+                                    + "sequence.field. Requires a primary-key table with "
+                                    + "write-only=true. Inline compaction is not allowed because "
+                                    + "snapshot ids are assigned only after commit. To compact such "
+                                    + "tables, run a dedicated compaction job/action with "
+                                    + "write-only=false.");
+
+    @Immutable
     public static final ConfigOption<Boolean> AGGREGATION_REMOVE_RECORD_ON_DELETE =
             key("aggregation.remove-record-on-delete")
                     .booleanType()
@@ -3413,6 +3431,10 @@ public class CoreOptions implements Serializable {
 
     public boolean sequenceFieldSortOrderIsAscending() {
         return options.get(SEQUENCE_FIELD_SORT_ORDER) == SortOrder.ASCENDING;
+    }
+
+    public boolean snapshotSequenceOrdering() {
+        return options.get(SEQUENCE_SNAPSHOT_ORDERING);
     }
 
     public Optional<String> rowkindField() {
--- a/paimon-core/src/main/java/org/apache/paimon/io/KeyValueDataFileRecordReader.java
+++ b/paimon-core/src/main/java/org/apache/paimon/io/KeyValueDataFileRecordReader.java
@@ -72,19 +72,10 @@ public class KeyValueDataFileRecordReader implements FileRecordReader<KeyValue> 
                         return null;
                     }
                     KeyValue kv = serializer.fromRow(internalRow).setLevel(level);
-                    // When snapshot-ordering is enabled, an APPEND file's per-record sequence
-                    // numbers are replaced with the commit snapshot id so that later snapshots
-                    // always win during merge. COMPACT files are left untouched: their records
-                    // already carry the snapshot id in the per-record _SEQUENCE_NUMBER column.
-                    //
-                    // CAUTION: in this mode the per-record sequence number physically stored in
-                    // an APPEND file is NOT trustworthy and must not be relied upon. The writer
-                    // seeds its sequence counter from the file-level maxSequenceNumber (which has
-                    // been stamped to the snapshot id), so the on-disk per-record values can be
-                    // small and non-monotonic across snapshots. Correct ordering is established
-                    // here, by this override; any future read path that bypasses this override
-                    // would order APPEND records by the stale on-disk values and break snapshot
-                    // ordering.
+                    // In snapshot-ordering mode, an APPEND file's on-disk per-record sequence
+                    // numbers are stale; we override them with the commit snapshot id so later
+                    // snapshots win during merge. Any read path bypassing this override would
+                    // order APPEND records incorrectly.
                     if (overrideSequenceWithSnapshotId) {
                         kv.setSequenceNumber(snapshotId);
                     }
--- a/paimon-core/src/main/java/org/apache/paimon/io/KeyValueDataFileRecordReader.java
+++ b/paimon-core/src/main/java/org/apache/paimon/io/KeyValueDataFileRecordReader.java
@@ -36,12 +36,26 @@ public class KeyValueDataFileRecordReader implements FileRecordReader<KeyValue> 
     private final FileRecordReader<InternalRow> reader;
     private final KeyValueSerializer serializer;
     private final int level;
+    private final boolean overrideSequenceWithSnapshotId;
+    private final long snapshotId;
 
     public KeyValueDataFileRecordReader(
             FileRecordReader<InternalRow> reader, RowType keyType, RowType valueType, int level) {
+        this(reader, keyType, valueType, level, false, KeyValue.UNKNOWN_SEQUENCE);
+    }
+
+    public KeyValueDataFileRecordReader(
+            FileRecordReader<InternalRow> reader,
+            RowType keyType,
+            RowType valueType,
+            int level,
+            boolean overrideSequenceWithSnapshotId,
+            long snapshotId) {
         this.reader = reader;
         this.serializer = new KeyValueSerializer(keyType, valueType);
         this.level = level;
+        this.overrideSequenceWithSnapshotId = overrideSequenceWithSnapshotId;
+        this.snapshotId = snapshotId;
     }
 
     @Nullable
@@ -53,10 +67,29 @@ public class KeyValueDataFileRecordReader implements FileRecordReader<KeyValue> 
         }
 
         return iterator.transform(
-                internalRow ->
-                        internalRow == null
-                                ? null
-                                : serializer.fromRow(internalRow).setLevel(level));
+                internalRow -> {
+                    if (internalRow == null) {
+                        return null;
+                    }
+                    KeyValue kv = serializer.fromRow(internalRow).setLevel(level);
+                    // When snapshot-ordering is enabled, an APPEND file's per-record sequence
+                    // numbers are replaced with the commit snapshot id so that later snapshots
+                    // always win during merge. COMPACT files are left untouched: their records
+                    // already carry the snapshot id in the per-record _SEQUENCE_NUMBER column.
+                    //
+                    // CAUTION: in this mode the per-record sequence number physically stored in
+                    // an APPEND file is NOT trustworthy and must not be relied upon. The writer
+                    // seeds its sequence counter from the file-level maxSequenceNumber (which has
+                    // been stamped to the snapshot id), so the on-disk per-record values can be
+                    // small and non-monotonic across snapshots. Correct ordering is established
+                    // here, by this override; any future read path that bypasses this override
+                    // would order APPEND records by the stale on-disk values and break snapshot
+                    // ordering.
+                    if (overrideSequenceWithSnapshotId) {
+                        kv.setSequenceNumber(snapshotId);
+                    }
+                    return kv;
+                });
     }
 
     @Override
--- a/paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreWrite.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreWrite.java
@@ -95,6 +95,7 @@ public abstract class AbstractFileStoreWrite<T> implements FileStoreWrite<T> {
     private boolean closeCompactExecutorWhenLeaving = true;
     private boolean ignorePreviousFiles = false;
     private boolean ignoreNumBucketCheck = false;
+    private final boolean compactOnly;
 
     protected CompactionMetrics compactionMetrics = null;
     protected final String tableName;
@@ -123,6 +124,7 @@ public abstract class AbstractFileStoreWrite<T> implements FileStoreWrite<T> {
         this.tableName = tableName;
         this.writerNumberMax = options.writeMaxWritersToSpill();
         this.legacyPartitionName = options.legacyPartitionName();
+        this.compactOnly = options.snapshotSequenceOrdering() && !options.writeOnly();
         this.partitionTimestampValidator =
                 PartitionTimestampValidator.create(options, partitionType);
     }
@@ -167,6 +169,15 @@ public abstract class AbstractFileStoreWrite<T> implements FileStoreWrite<T> {
 
     @Override
     public void write(BinaryRow partition, int bucket, T data) throws Exception {
+        if (compactOnly) {
+            throw new IllegalStateException(
+                    CoreOptions.SEQUENCE_SNAPSHOT_ORDERING.key()
+                            + " is enabled with "
+                            + CoreOptions.WRITE_ONLY.key()
+                            + " = false. This writer only allows compaction, not data writes. "
+                            + "Inline compaction is not supported because snapshot ids are "
+                            + "assigned only after commit.");
+        }
         WriterContainer<T> container = getWriterWrapper(partition, bucket);
         container.writer.write(data);
         if (container.dynamicBucketMaintainer != null) {
--- a/paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreWrite.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreWrite.java
@@ -95,7 +95,6 @@ public abstract class AbstractFileStoreWrite<T> implements FileStoreWrite<T> {
     private boolean closeCompactExecutorWhenLeaving = true;
     private boolean ignorePreviousFiles = false;
     private boolean ignoreNumBucketCheck = false;
-    private final boolean compactOnly;
 
     protected CompactionMetrics compactionMetrics = null;
     protected final String tableName;
@@ -124,7 +123,6 @@ public abstract class AbstractFileStoreWrite<T> implements FileStoreWrite<T> {
         this.tableName = tableName;
         this.writerNumberMax = options.writeMaxWritersToSpill();
         this.legacyPartitionName = options.legacyPartitionName();
-        this.compactOnly = options.snapshotSequenceOrdering() && !options.writeOnly();
         this.partitionTimestampValidator =
                 PartitionTimestampValidator.create(options, partitionType);
     }
@@ -169,15 +167,6 @@ public abstract class AbstractFileStoreWrite<T> implements FileStoreWrite<T> {
 
     @Override
     public void write(BinaryRow partition, int bucket, T data) throws Exception {
-        if (compactOnly) {
-            throw new IllegalStateException(
-                    CoreOptions.SEQUENCE_SNAPSHOT_ORDERING.key()
-                            + " is enabled with "
-                            + CoreOptions.WRITE_ONLY.key()
-                            + " = false. This writer only allows compaction, not data writes. "
-                            + "Inline compaction is not supported because snapshot ids are "
-                            + "assigned only after commit.");
-        }
         WriterContainer<T> container = getWriterWrapper(partition, bucket);
         container.writer.write(data);
         if (container.dynamicBucketMaintainer != null) {
--- a/paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
@@ -989,6 +989,10 @@ public class FileStoreCommitImpl implements FileStoreCommit {
                 deltaFiles = assigned.assignedEntries;
             }
 
+            if (options.snapshotSequenceOrdering()) {
+                deltaFiles = stampSequenceWithSnapshotId(newSnapshotId, commitKind, deltaFiles);
+            }
+
             // the added records subtract the deleted records from
             long deltaRecordCount = recordCountAdd(deltaFiles) - recordCountDelete(deltaFiles);
             long totalRecordCount = previousTotalRecordCount + deltaRecordCount;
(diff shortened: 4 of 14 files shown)


JunRuiLee:
Hi @JingsongLi, could you help take a look? Many thanks.

paimon-core/src/main/java/org/apache/paimon/io/KeyValueDataFileRecordReader.java
```diff
@@ -36,12 +36,23 @@ public class KeyValueDataFileRecordReader implements FileRecordReader<KeyValue>
     private final FileRecordReader<InternalRow> reader;
     private final KeyValueSerializer serializer;
     private final int level;
+    private final long snapshotId;
 
     public KeyValueDataFileRecordReader(
             FileRecordReader<InternalRow> reader, RowType keyType, RowType valueType, int level) {
+        this(reader, keyType, valueType, level, KeyValue.UNKNOWN_SNAPSHOT_ID);
```
leaves12138: Remote this constructor

paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
```diff
@@ -1193,4 +1197,57 @@ public void close() {
...
+     */
+    private static List<ManifestEntry> assignSnapshotSequenceOrdering(
+            long snapshotId, CommitKind commitKind, List<ManifestEntry> files) {
+        long stamp = snapshotId;
+        if (commitKind == CommitKind.COMPACT) {
+            boolean found = false;
+            stamp = 0;
+            for (ManifestEntry entry : files) {
+                if (entry.kind() == FileKind.DELETE) {
+                    stamp = Math.max(stamp, entry.file().minSequenceNumber());
```
leaves12138: Can you double check here?
If there is a compact between file a (snapshot = 1) and file c (snapshot = 3). Then we get file d (snapshot = 3). But the file b (snapshot = 2) does not participate in the compaction. So file b remains snapshot = 2. File b and File a have key range overlapping. So the may merge. 

Before compaction, file a sequence number(snapshot id) is smaller than file b.
After compaction, file a (which now is file c) sequence number is larger than file b?

leaves12138: It too dangerous here, after compaction file may have a sequence number far exceed the real snapshot it they write.

JunRuiLee:
Thanks @leaves12138 for the review! Fixed the compaction ordering issue by persisting per-record snapshotId through _SEQUENCE_NUMBER column. Added tests for the scenario you described. Old constructor removed. 

PTAL, Thanks!

paimon-core/src/main/java/org/apache/paimon/io/KeyValueDataFileRecordReader.java
JunRuiLee: fixed

paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
JunRuiLee: fixed

JunRuiLee: fixed

paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
```diff
@@ -578,6 +582,19 @@ private static void validateFileIndex(TableSchema schema) {
...
+                !schema.primaryKeys().isEmpty(),
+                "%s = true requires a primary-key table; append-only tables cannot use "
+                        + "snapshot-based sequence ordering.",
+                CoreOptions.SEQUENCE_SNAPSHOT_ORDERING.key());
+        checkArgument(
+                options.sequenceField().isEmpty(),
+                "%s = true is mutually exclusive with %s; the snapshot id is the sole tiebreaker.",
+                CoreOptions.SEQUENCE_SNAPSHOT_ORDERING.key(),
+                CoreOptions.SEQUENCE_FIELD.key());
+    }
```
leaves12138, line 655: This option is currently accepted for every primary-key merge engine, but the implementation only preserves `snapshotId` for merge functions that return an input `KeyValue`. For example, `PartialUpdateMergeFunction` and `AggregateMergeFunction` build a new `KeyValue` via `replace(...)`, which resets `snapshotId` to `UNKNOWN_SNAPSHOT_ID`. During compaction, `stampSequenceWithSnapshotId` then writes `-1` into `_SEQUENCE_NUMBER` / file sequence metadata, so later reads can order compacted records incorrectly. Could you either restrict `sequence.snapshot-ordering` to the supported merge engine(s) here, or propagate the winning snapshot id through all merge functions and add tests for partial-update / aggregation?

paimon-core/src/main/java/org/apache/paimon/mergetree/compact/LookupMergeTreeCompactRewriter.java
```diff
@@ -217,7 +221,8 @@ public MergeFunctionWrapper<ChangelogResult> create(
                     valueEqualiser,
                     lookupStrategy,
                     deletionVectorsMaintainer,
-                    userDefinedSeqComparator);
+                    userDefinedSeqComparator,
+                    snapshotSequenceOrdering);
```
leaves12138: The lookup changelog path can still lose the snapshot id when `LookupMergeFunction` spills its `KeyValueBuffer` to the binary buffer. `KeyValueBuffer.createBinaryBuffer` still constructs `new KeyValueWithLevelNoReusingSerializer(keyType, valueType)` without `includeSnapshotId`, so after `lookup.merge-records-threshold` is exceeded, deserialized candidates have `UNKNOWN_SNAPSHOT_ID` and this comparator falls back to sequence-only ordering. Please thread `snapshotSequenceOrdering` into `KeyValueBuffer`'s serializer and add a test that forces lookup-buffer spill, for example with a very small `lookup.merge-records-threshold` and an `IOManager`.

paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -965,6 +965,20 @@ public InlineElement getDescription() {
                     .defaultValue(SortOrder.ASCENDING)
                     .withDescription("Specify the order of sequence.field.");
 
+    @Immutable
+    public static final ConfigOption<Boolean> SEQUENCE_SNAPSHOT_ORDERING =
+            key("sequence.snapshot-ordering")
+                    .booleanType()
+                    .defaultValue(false)
+                    .withDescription(
+                            "When enabled, merge uses the commit snapshot id as the primary "
```
leaves12138: This option also looks unsafe to enable on a table that already has data written without the feature. Existing APPEND files have `minSequenceNumber` as the old sequence range, and existing COMPACT files have `_SEQUENCE_NUMBER` as the old per-record sequence number; after toggling this option on, readers will interpret those values as snapshot ids. Could this be documented and/or rejected for `ALTER TABLE` as a creation-only option? Otherwise an existing table can silently reorder old records.

leaves12138 (changes requested):
Thanks for the update. I took another careful pass over the snapshot-ordering implementation. I think there are still a few correctness issues to address before this can be safely merged.

JunRuiLee:
Thanks @leaves12138 for the careful review.

  I fixed the first two correctness issues:

  1. PartialUpdateMergeFunction and AggregateMergeFunction now preserve the winning input record’s snapshotId when returning a newly built KeyValue, so compaction no longer stamps UNKNOWN_SNAPSHOT_ID into
     _SEQUENCE_NUMBER.
  2. KeyValueBuffer now preserves snapshotId when snapshot ordering is enabled, so lookup compaction buffer spill does not lose it during binary serialization/deserialization.

  I also added regression coverage for merge-function snapshotId preservation, partial-update compaction, aggregate compaction, and lookup buffer spill.

  For the ALTER TABLE concern: this option is annotated as immutable, so enabling it via ALTER on a table with existing snapshots is rejected; empty-table ALTER remains allowed.

paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
JunRuiLee: fixed

paimon-core/src/main/java/org/apache/paimon/mergetree/compact/LookupMergeTreeCompactRewriter.java
JunRuiLee: fixed

paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
JunRuiLee: this option is annotated as immutable, so enabling it via ALTER on a table with existing snapshots is rejected; empty-table ALTER remains allowed.

paimon-core/src/main/java/org/apache/paimon/mergetree/compact/PartialUpdateMergeFunction.java
```diff
@@ -358,7 +362,8 @@ public KeyValue getResult() {
         }
 
         RowKind rowKind = currentDeleteRow || !meetInsert ? RowKind.DELETE : RowKind.INSERT;
-        return reused.replace(currentKey, latestSequenceNumber, rowKind, row);
+        return reused.replace(currentKey, latestSequenceNumber, rowKind, row)
```
leaves12138: Thanks for fixing the missing snapshotId propagation. One more partial-update edge case may need either an explanation or a regression test: a compacted partial-update result can contain fields coming from different input snapshots, but this code assigns one row-level snapshotId (`latestSnapshotId`) to the whole merged row. If snapshot 1 writes `b=100`, snapshot 2 writes `b=200` and does not participate in this compaction, and snapshot 3 writes `c=300` and is compacted together with snapshot 1, the compacted row may carry snapshotId=3 for both fields and later override snapshot 2's `b=200` with snapshot 1's old `b=100`. Could you add a `partial-update` version of the non-contiguous compaction test (similar to the deduplicate ordering-reversal case) to prove this cannot happen?

leaves12138 (commented):
One remaining edge case around partial-update snapshot ordering.

paimon-core/src/main/java/org/apache/paimon/mergetree/compact/PartialUpdateMergeFunction.java
JunRuiLee: I don’t think this case is reachable in normal compaction. 
Partial-update without sequence group relies on the merge-tree invariant that compaction picks a run-based prefix, not arbitrary files; so it cannot compact snapshot 1 and snapshot 3 while leaving snapshot 2 for the same key outside. 
If such a non-contiguous compaction were possible, regular row-level sequence number ordering would already have the same correctness issue, not only snapshot ordering.

paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java
```diff
@@ -969,6 +969,10 @@ CommitResult tryCommitOnce(
                 deltaFiles = assigned.assignedEntries;
             }
 
+            if (options.snapshotSequenceOrdering()) {
+                deltaFiles = stampSequenceWithSnapshotId(newSnapshotId, commitKind, deltaFiles);
```
leaves12138, line 993: I think this still has a correctness issue around where the snapshot-id stamp is applied.

`stampSequenceWithSnapshotId` only rewrites the `ManifestEntry`s that are written to the snapshot. The writer-side `DataFileMeta`s kept in `MergeTreeWriter` / `Levels` are still the original file metas produced by the rolling writer, whose min/max sequence numbers are local writer sequence numbers. If the same writer later compacts those already-committed APPEND files without being restored from the committed manifest, `KeyValueFileReaderFactory` will recover `snapshotId` from `file.minSequenceNumber()` for APPEND files and may treat that local sequence as the snapshot id.

A failing shape would be:
1. write enough records so the local sequence number becomes much larger than the snapshot id;
2. commit an old version of key K (the manifest is stamped correctly, but the writer-side `Levels` still has the local sequence in `DataFileMeta`);
3. trigger compaction in the same writer, which reads the APPEND file and writes that local sequence into the compacted row's per-record sequence/snapshot id;
4. write a later snapshot for key K. The older compacted row can now compare as newer and win.

Could we either update the writer-side file metas after commit (or otherwise avoid using unstamped in-memory metas for snapshot-ordering compaction), and add a regression test where local sequence numbers are intentionally larger than subsequent snapshot ids before compaction?

JunRuiLee: Thanks for catching this!

  The root cause is: snapshot ordering relies on snapshot id to determine record order, but inline compaction happens **before** snapshot creation — files have not been stamped with the correct snapshot id yet, so compaction would propagate incorrect values to KV records.

  **Fix**: require `write-only=true` for snapshot-ordering tables, disabling inline compaction. Compaction must be done by dedicated compact jobs (which read from committed manifests where files are already correctly stamped).

leaves12138:
This feature is more complex than it is expected. I can't figure out more corner case, but there must exist some cases we have not catch. This looks good to me, but it need more experts to review this feature. @JingsongLi

JunRuiLee:
@leaves12138 Thanks for the review! I've reorganized the commits to make the progression clearer.

JunRuiLee:
Hi @JingsongLi, could you please take a look? Thanks so much.

JingsongLi (commented):
## Review: snapshot-based sequence ordering for primary-key tables

Well-structured feature with comprehensive test coverage. A few observations and potential issues:

### Design

1. **Repurposing `minSequenceNumber`/`maxSequenceNumber` in `DataFileMeta`** -- Clever hack that avoids a schema evolution in the manifest format, but it makes these fields semantically overloaded. Any existing or future code that reads `minSequenceNumber()` with its original semantics (e.g., snapshot expiration, incremental reads) could silently misinterpret the value when snapshot-ordering is enabled. The Javadoc addition helps, but consider adding a helper like `DataFileMeta.commitSnapshotId(CoreOptions)` to make call-sites self-documenting and less error-prone.

2. **Threading `boolean snapshotSequenceOrdering` through ~10 constructors** -- This increases coupling and makes signature changes expensive. A lighter alternative would be to encapsulate the ordering strategy in a `Comparator<KeyValue>` factory (or reuse/extend `UserDefinedSeqComparator`) created once in `MergeTreeCompactManagerFactory` and passed downstream, rather than a bare boolean. Not blocking, but worth considering for maintainability.

### Correctness

3. **Loss of within-snapshot secondary ordering after compaction** -- In `MergeTreeCompactRewriter.stampSequenceWithSnapshotId`, you overwrite per-record `sequenceNumber` with `snapshotId`. After a compacted file is read back (`recoverSnapshotIdFromSequence = true`), `sequenceNumber` still holds the snapshotId value (line ~289 of `KeyValueDataFileRecordReader`), so the original within-snapshot sequence is lost. This is safe today because compaction already resolved same-key conflicts, but if a future merge path re-merges records from the same snapshot across compacted and un-compacted files with different keys sharing a snapshot, the tiebreaker is effectively random (all records show the same "sequence"). Worth documenting this invariant explicitly.

4. **`PartialUpdateMergeFunction` snapshot propagation with field-level sequence groups** -- `latestSnapshotId` always tracks the globally-latest input's snapshotId, but when `fieldSeqComparators` is non-empty, individual fields may be retained from older inputs. The result's snapshotId could overstate recency for some columns. This is analogous to the existing `latestSequenceNumber` behavior so it is consistent, but worth a comment acknowledging the limitation.

5. **`FileSource` absent case** -- In `KeyValueFileReaderFactory.createRecordReader`:
   ```java
   boolean recoverSnapshotIdFromSequence =
       snapshotSequenceOrdering
           && file.fileSource().isPresent()
           && file.fileSource().get() == FileSource.COMPACT;
   ```
   If `fileSource()` is empty (e.g., files from older table versions before file-source tracking was added), the code falls through to the uniform file-level stamp path. If those legacy files were written before snapshot-ordering was enabled, `minSequenceNumber` still carries the original sequence range, not a snapshot ID. The reader would then interpret that as a snapshotId, which could cause incorrect merge ordering. Is there a guard against enabling this option on a table that already has existing data files without file-source metadata?

### Nits

6. In `SortMergeReaderWithMinHeap`, the snapshot comparison is placed inside the `else` branch of `if (userDefinedSeqComparator != null)`. This means `snapshotSequenceOrdering` is silently ignored when a user-defined seq comparator is present. The schema validation rejects `sequence.field` + `snapshot-ordering` together, so it should be unreachable, but an `assert` or early-fail would make this safer against future relaxation of validation.

7. The `SchemaValidation` overload `validateTableSchema(schema, dynamicOptionKeys)` is public with no Javadoc. Since it gates security-sensitive behavior (allowing write-only bypass for compact jobs), documenting its contract would help future maintainers.

### Testing

The test matrix is thorough -- covering deduplicate, partial-update, aggregate, changelog-input, changelog-lookup, multi-round compaction, and ordering reversal scenarios. The `MergeFunctionSnapshotIdTest` is a great regression safety net. One gap: there is no test with `SortEngine.LOSER_TREE` specifically exercising the snapshot ordering path in an integration test (though `SortMergeSnapshotOrderingTest` covers both engines at the unit level via `@EnumSource`).

Overall: solid implementation for the targeted use case. The main risk area is the semantic overloading of `minSequenceNumber`/`maxSequenceNumber` and potential interaction with legacy files.

JunRuiLee:
Thanks @JingsongLi for the detailed review.

  I addressed the actionable safety/documentation/test items in this update:

  1. Added documentation for snapshot-ordering compaction: compacted records intentionally store snapshot id in the sequence field, so within-snapshot secondary ordering is not preserved after compaction. This is
  safe under the current invariant because compaction has already resolved same-key conflicts.
  2. Added a comment in `PartialUpdateMergeFunction` for field-level sequence groups: the output row has one row-level snapshot id, which may overstate recency for fields retained from older inputs. This matches
  the existing `latestSequenceNumber` behavior.
  3. Added a fail-fast guard when snapshot-ordering reads a file without `fileSource` metadata. `sequence.snapshot-ordering` is immutable and should only be enabled for newly-created/empty tables, so legacy files
  should not reach this path.
  4. Added defensive checks that `sequence.snapshot-ordering` is not used together with a user-defined sequence comparator near the sort/lookup comparator creation paths.
  5. Added Javadoc for `validateTableSchema(schema, dynamicOptionKeys)` to document the dynamic `write-only` override contract for dedicated compaction jobs.
  6. The existing integration tests use the default `LOSER_TREE`; I added a `MIN_HEAP` variant so both sort engines are covered.


  I did not include the broader comparator-strategy refactor in this PR because it would touch several merge paths and make the correctness-focused review larger. I added fail-fast checks and documentation around the current paths instead, and I think the strategy refactor is better handled as a follow-up cleanup.

leaves12138 (commented):
Thanks for the update. I took another pass over the latest snapshot-ordering implementation.

The current approach looks internally consistent to me now: APPEND files carry the commit snapshot id via the file-level sequence range, COMPACT files recover per-record snapshot ids from `_SEQUENCE_NUMBER`, and the sort-merge readers use snapshot id before the existing sequence tiebreaker. The added guards around `write-only`, legacy files without `fileSource`, and `sequence.field` are important and address the major safety concerns from the previous reviews.

I also checked the partial-update / aggregation snapshot propagation, lookup buffer spill serialization, dedicated compact-writer path, and the multi-round compaction tests. The test coverage is strong for the tricky ordering reversal cases.

One non-blocking follow-up suggestion: `LookupChangelogMergeFunctionWrapper.replace(...)` still creates changelog `KeyValue`s without preserving `snapshotId`. Those changelog files do not appear to feed back into the main data merge ordering, so I do not think this blocks the PR, but preserving it there too would make `_SEQUENCE_NUMBER` in compact-produced changelog files more consistent and easier to reason about.

I also verified locally with:

```
mvn -pl paimon-core -am -DskipTests -Dspotless.check.skip=true -Drat.skip=true -Dcheckstyle.skip=true -Dscala-2.12 -DskipITs compile
```

No blocking issues from my side.

JingsongLi (changes requested):
In my understanding, only when reading data should the sequenceNumbers in the KeyValue be modified to the sequenceNumbers of the file, but the current changes are much larger.

JunRuiLee:
> In my understanding, only when reading data should the sequenceNumbers in the KeyValue be modified to the sequenceNumbers of the file, but the current changes are much larger.


  @JingsongLi Thanks for taking another look.

  My understanding is that we may need to preserve the original KeyValue.sequenceNumber as a secondary tiebreaker within the same snapshot. If we replace it directly with the file-level sequence number on read, records from the same snapshot would lose their original ordering.

  Does it make sense to keep both values here: the file-level sequence number as snapshotId, and the original KeyValue.sequenceNumber as the secondary ordering key? Or do you think snapshot ordering can replace the original sequence tiebreaker completely?

JingsongLi:
@JunRuiLee 
I think we can use the following strategies:
1. For the new files generated by Append, we use the Snapshot ID as the Sequence Number.
2. For Compact files, the sequenceNumbers in the actual data remain unchanged.

JunRuiLee:
Aligned offline with @JingsongLi : we reuse the existing per-record _SEQUENCE_NUMBER to carry the commit snapshot id (APPEND files are stamped at read time, COMPACT files pass through) rather than introducing a new
KeyValue.snapshotId field. 
Same-key ordering within a single snapshot is intentionally not preserved — the default spillable writer collapses such records through the merge function first, so it's practically unreachable — so we don't force write-buffer-spillable=true and just document the limitation.

JingsongLi (commented):
Overall LGTM, still some comments.

paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreWrite.java
```diff
@@ -167,6 +169,15 @@ public void withCompactExecutor(ExecutorService compactExecutor) {
 
     @Override
     public void write(BinaryRow partition, int bucket, T data) throws Exception {
+        if (compactOnly) {
```
JingsongLi: Why do we need to intercept here?

JingsongLi (commented):
Another comment is: comments in codes are too more...

paimon-core/src/main/java/org/apache/paimon/operation/AbstractFileStoreWrite.java
JunRuiLee: Removed the guard. It was a defensive runtime check for the dedicated-compact path. Agree it's too intrusive on the hot path.

JunRuiLee:
> Another comment is: comments in codes are too more...

@JingsongLi , thanks for the reminder. Shortened the comments across the read/commit paths. PTAL, thanks!

JingsongLi (approved):
+1

PR #7836, opened 2026-05-13
[spark] Resolve blob_view field name to field id at analysis time

leaves12138:
### Purpose
Add Spark SQL coverage for writing blob view values with the built-in `sys.blob_view` function.

### Changes
- Add an E2E case in `BlobTestBase` that writes upstream BLOB data, constructs downstream view references from the row-tracking table, and verifies the read path resolves the original bytes.

### Tests
- `mvn -pl paimon-spark/paimon-spark-ut -am -Pfast-build -DfailIfNoTests=false -DwildcardSuites=org.apache.paimon.spark.sql.BlobTestBase -Dtest=none test`


JingsongLi (approved):
+1

PR #7839, opened 2026-05-13
[common] Avoid secured Hadoop FileSystem without Kerberos config

### LINKED ISSUE #7146: [Bug] SecurityConfiguration.isLegal() incorrectly returns true when no security configuration is provided
### Search before asking

### Paimon version

1.3.1

### Compute Engine

Hadoop: 2.7.5
Flink: 1.19.1

### Minimal reproduce step

1. Create a Paimon catalog **without** Kerberos configuration:
`
   CREATE CATALOG paimon_catalog WITH (
       'type' = 'paimon',
       'metastore' = 'hive',
       'uri' = 'thrift://xxx:9083',
       'warehouse' = 'hdfs://nameservice1/user/hive/warehouse/paimon.db'
       -- No security.kerberos.login.keytab or security.kerberos.login.principal
   );
   `
2. Write data using Flink (with checkpoint enabled):
`   INSERT INTO paimon_catalog.database.table
   SELECT * FROM source_table;`
3. Wait for checkpoint to trigger (or manually trigger checkpoint)
4. Observe errors:
TaskManager logs show InterruptedIOException during checkpoint
Flink Web UI shows "Unable to close file because the last block does not have enough number of replicas"
Checkpoint fails
Code to reproduce:
`Options options = new Options();
// No security configuration provided
SecurityConfiguration config = new SecurityConfiguration(options);
boolean isLegal = config.isLegal();  // Returns true (incorrect)`

- `isLegal()` should return `false` when no security configuration is provided
- Security wrappers should only be created when Kerberos is actually configured
- Non-Kerberos environments should use the native `FileSystem` directly without `doAs()` overhead
- **Checkpoints should succeed without `InterruptedIOException` errors**
- **HDFS files should close properly without block replica issues**
`
	at org.apache.paimon.fs.hadoop.HadoopSecuredFileSystem.lambda$delete$5(HadoopSecuredFileSystem.java:109) ~[zrc_2635_1.0.59.jar:?]
	at java.security.AccessController.doPrivileged(Native Method) ~[?:1.8.0_201]
	at javax.security.auth.Subject.doAs(Subject.java:422) ~[?:1.8.0_201]
	at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1937) ~[flink-shaded-hadoop-2-uber-3.3.5-infra-v2-10.0.jar:3.3.5-infra-v2-10.0]
	at org.apache.paimon.fs.hadoop.HadoopSecuredFileSystem.runSecuredWithIOException(HadoopSecuredFileSystem.java:155) ~[zrc_2635_1.0.59.jar:?]
	at org.apache.paimon.fs.hadoop.HadoopSecuredFileSystem.delete(HadoopSecuredFileSystem.java:109) ~[zrc_2635_1.0.59.jar:?]
	at org.apache.paimon.fs.hadoop.HadoopFileIO.delete(HadoopFileIO.java:209) ~[zrc_2635_1.0.59.jar:?]
	at org.apache.paimon.fs.FileIO.deleteQuietly(FileIO.java:246) ~[zrc_2635_1.0.59.jar:?]
	at org.apache.paimon.io.SingleFileWriter.abort(SingleFileWriter.java:161) ~[zrc_2635_1.0.59.jar:?]
	at org.apache.paimon.io.RollingFileWriter.abort(RollingFileWriter.java:147) ~[zrc_2635_1.0.59.jar:?]
	at org.apache.paimon.io.RollingFileWriter.close(RollingFileWriter.java:172) ~[zrc_2635_1.0.59.jar:?]
	at org.apache.paimon.mergetree.MergeTreeWriter.flushWriteBuffer(MergeTreeWriter.java:234) ~[zrc_2635_1.0.59.jar:?]
	at org.apache.paimon.mergetree.MergeTreeWriter.prepareCommit(MergeTreeWriter.java:253) ~[zrc_2635_1.0.59.jar:?]
	at org.apache.paimon.operation.AbstractFileStoreWrite.prepareCommit(AbstractFileStoreWrite.java:215) ~[zrc_2635_1.0.59.jar:?]
	at org.apache.paimon.operation.MemoryFileStoreWrite.prepareCommit(MemoryFileStoreWrite.java:152) ~[zrc_2635_1.0.59.jar:?]
	at org.apache.paimon.table.sink.TableWriteImpl.prepareCommit(TableWriteImpl.java:262) ~[zrc_2635_1.0.59.jar:?]
	at org.apache.paimon.flink.sink.StoreSinkWriteImpl.prepareCommit(StoreSinkWriteImpl.java:150) ~[zrc_2635_1.0.59.jar:?]
	at org.apache.paimon.flink.sink.TableWriteOperator.prepareCommit(TableWriteOperator.java:151) ~[zrc_2635_1.0.59.jar:?]
	at org.apache.paimon.flink.sink.RowDataStoreWriteOperator.prepareCommit(RowDataStoreWriteOperator.java:205) ~[zrc_2635_1.0.59.jar:?]
	at org.apache.paimon.flink.sink.PrepareCommitOperator.emitCommittables(PrepareCommitOperator.java:115) ~[zrc_2635_1.0.59.jar:?]
	at org.apache.paimon.flink.sink.PrepareCommitOperator.prepareSnapshotPreBarrier(PrepareCommitOperator.java:95) ~[zrc_2635_1.0.59.jar:?]
	at org.apache.flink.streaming.runtime.tasks.RegularOperatorChain.prepareSnapshotPreBarrier(RegularOperatorChain.java:89) ~[flink-dist-1.19.1.jar:1.19.1]`

### Anything else?

**My Understanding:**

1. The `isLegal()` method should indicate whether a valid security configuration exists
2. When no configuration is provided, it should return `false` to avoid unnecessary security wrappers
3. The `ugi.doAs()` wrapper in `HadoopSecuredFileSystem` causes thread interruption handling issues in Hadoop 2.7.5
4. During Flink checkpoints, thread interruption combined with `doAs()` leads to:
   - HDFS I/O operations timing out
   - Files not being closed properly
   - HDFS reporting insufficient block replicas
   - Checkpoint failures

**My Design:**

- Change the return value from `true` to `false` when no keytab/principal is configured
- Add a comment explaining why we return `false`
- This ensures backward compatibility (Kerberos environments still work correctly)
- This is a minimal, safe change that fixes the root cause

**POC Code:**

I have tested this fix locally and confirmed:
- Non-Kerberos environments no longer create unnecessary wrappers
- Kerberos environments continue to work correctly
- Checkpoint operations succeed in Hadoop 2.7.5
- No more `InterruptedIOException` errors
- HDFS files close properly without block replica issues

**Proposed Solution:**

Change line 96 in `SecurityConfiguration.java` from `return true;` to `return false;`:

`
public boolean isLegal() {
    if (StringUtils.isNullOrWhitespaceOnly(keytab)
            != StringUtils.isNullOrWhitespaceOnly(principal)) {
        return false;
    }

    if (!StringUtils.isNullOrWhitespaceOnly(keytab)) {
        File keytabFile = new File(keytab);
        return keytabFile.exists() && keytabFile.isFile() && keytabFile.canRead();
    }

    // Return false when no security configuration is provided
    // This prevents unnecessary security wrapper creation
    return false;  // ✅ Fixed
}
`
Workaround:
As a temporary workaround, I rolled back to Paimon 1.0 behavior by modifying `HadoopFileIO.createFileSystem()`:
`
protected FileSystem createFileSystem(org.apache.hadoop.fs.Path path) throws IOException {
    return path.getFileSystem(hadoopConf.get());
    // Removed: HadoopSecuredFileSystem.trySecureFileSystem() call
}
`
This workaround resolves the issue but loses the security features. The proper fix is to correct `isLegal()` logic.

**Understanding:** SecurityConfiguration.isLegal() logic flaw causes unnecessary security wrapper creation, leading to checkpoint failures in Hadoop 2.7.5.

**Solution:** Change line 96 from `return true;` to `return false;` when no keytab/principal configured.

**POC:** Tested locally - checkpoints succeed, no more errors.

officialasishkumar:
### Purpose

Fixes #7146.

`SecurityConfiguration#isLegal` now returns false when either Kerberos keytab or principal is missing. This prevents `HadoopFileIO` from wrapping the native Hadoop `FileSystem` in `HadoopSecuredFileSystem` for catalogs without Kerberos configuration, while keeping valid keytab/principal configurations on the secured path.

### Tests

- `mvn -pl paimon-common -Pfast-build -Dtest=HadoopSecuredFileSystemTest test`
- `mvn -pl paimon-common spotless:check`

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/security/SecurityConfiguration.java
+++ b/paimon-common/src/main/java/org/apache/paimon/security/SecurityConfiguration.java
@@ -84,15 +84,11 @@ public class SecurityConfiguration {
 
     public boolean isLegal() {
         if (StringUtils.isNullOrWhitespaceOnly(keytab)
-                != StringUtils.isNullOrWhitespaceOnly(principal)) {
+                || StringUtils.isNullOrWhitespaceOnly(principal)) {
             return false;
         }
 
-        if (!StringUtils.isNullOrWhitespaceOnly(keytab)) {
-            File keytabFile = new File(keytab);
-            return keytabFile.exists() && keytabFile.isFile() && keytabFile.canRead();
-        }
-
-        return true;
+        File keytabFile = new File(keytab);
+        return keytabFile.exists() && keytabFile.isFile() && keytabFile.canRead();
     }
 }
--- a/paimon-common/src/test/java/org/apache/paimon/fs/hadoop/HadoopSecuredFileSystemTest.java
+++ b/paimon-common/src/test/java/org/apache/paimon/fs/hadoop/HadoopSecuredFileSystemTest.java
@@ -34,7 +34,15 @@ public class HadoopSecuredFileSystemTest {
     @TempDir private java.nio.file.Path tmp;
 
     @Test
-    public void test() throws Exception {
+    public void testEmptySecurityConfigurationDoesNotWrapFileSystem() throws Exception {
+        HadoopFileIO fileIO = createFileIO(new Options());
+
+        assertThat(fileIO.getFileSystem(new org.apache.hadoop.fs.Path("file:///tmp/test")))
+                .isNotInstanceOf(HadoopSecuredFileSystem.class);
+    }
+
+    @Test
+    public void testValidKeytabAndPrincipalWrapsFileSystem() throws Exception {
         File keytabFile = new File(tmp.toFile(), "test-keytab.keytab");
         assertThat(keytabFile.createNewFile()).isTrue();
 
@@ -42,8 +50,8 @@ public class HadoopSecuredFileSystemTest {
         options.set("security.kerberos.login.principal", "test-user");
         options.set("security.kerberos.login.keytab", keytabFile.getAbsolutePath());
 
-        HadoopFileIO fileIO = new HadoopFileIO(new Path("file:///tmp/test"));
-        fileIO.configure(CatalogContext.create(options));
+        HadoopFileIO fileIO = createFileIO(options);
+
         assertThat(fileIO.getFileSystem(new org.apache.hadoop.fs.Path("file:///tmp/test")))
                 .isInstanceOf(HadoopSecuredFileSystem.class);
     }
@@ -52,8 +60,7 @@ public class HadoopSecuredFileSystemTest {
     public void testPreserveExternalUgiWhenNoKerberosCredentials() throws Exception {
         Options options = new Options();
 
-        HadoopFileIO fileIO = new HadoopFileIO(new Path("file:///tmp/test"));
-        fileIO.configure(CatalogContext.create(options));
+        HadoopFileIO fileIO = createFileIO(options);
         assertThat(fileIO.getFileSystem(new org.apache.hadoop.fs.Path("file:///tmp/test")))
                 .isNotInstanceOf(HadoopSecuredFileSystem.class);
     }
@@ -66,9 +73,14 @@ public class HadoopSecuredFileSystemTest {
         Options options = new Options();
         options.set("security.kerberos.login.keytab", keytabFile.getAbsolutePath());
 
-        HadoopFileIO fileIO = new HadoopFileIO(new Path("file:///tmp/test"));
-        fileIO.configure(CatalogContext.create(options));
+        HadoopFileIO fileIO = createFileIO(options);
         assertThat(fileIO.getFileSystem(new org.apache.hadoop.fs.Path("file:///tmp/test")))
                 .isNotInstanceOf(HadoopSecuredFileSystem.class);
+    }
+
+    private HadoopFileIO createFileIO(Options options) {
+        HadoopFileIO fileIO = new HadoopFileIO(new Path("file:///tmp/test"));
+        fileIO.configure(CatalogContext.create(options));
+        return fileIO;
     }
 }


JingsongLi (commented):
## Review of SecurityConfiguration.isLegal() change

### Correctness

The logic change from `!=` (XOR) to `||` (OR) is correct. Here is how the behavior changes:

| keytab | principal | Old result | New result |
|--------|-----------|-----------|-----------|
| empty  | empty     | `true`    | `false`   |
| set    | empty     | `false`   | `false`   |
| empty  | set       | `false`   | `false`   |
| set    | set       | check file| check file|

The fix correctly prevents `HadoopSecuredFileSystem` wrapping and `SecurityContext.install()` from running when no Kerberos config is present. The root cause of the `InterruptedIOException` during checkpoints is properly addressed.

### Interaction with PR #7862 (already merged to master)

Commit `1d0f005f78` ("Preserve external UGI when paimon has no kerberos credentials") added a guard inside `HadoopSecuredFileSystem.trySecureFileSystem()` at lines 204-211:

```java
if (config.isLegal()) {
    if (StringUtils.isNullOrWhitespaceOnly(config.getKeytab())
            && StringUtils.isNullOrWhitespaceOnly(config.getPrincipal())) {
        LOG.info("No paimon Kerberos credentials configured ...; skip ...");
        return fileSystem;
    }
    // ...wrap in secured FS...
}
```

With this PR applied, `isLegal()` already returns `false` when both keytab and principal are empty, so the inner guard (lines 204-211) becomes **dead code** — that branch is unreachable. Consider removing it in this PR or a follow-up to avoid confusing future readers.

That said, this PR is still valuable because it also fixes `SecurityContext.install()`, which lacks the #7862 guard and would still incorrectly attempt `HadoopModule.install()` with null keytab on the old logic.

### Design

The semantic shift of `isLegal()` is an improvement: it now means "a complete and valid security configuration is present" rather than "the security configuration is internally consistent." This aligns better with the method name and caller expectations.

### Test feedback

The two tests cover the happy path well. Consider adding a test for the **partial configuration** case (keytab set but principal missing, or vice versa) to guard against regressions in that validation path — especially since the `!=` to `||` change collapses two previously distinct failure modes into one.

### Minor nit

After the `||` change, the null-check on `keytab` before `new File(keytab)` is implicitly guaranteed by the early return. This is fine, but a brief inline comment (e.g., `// Both keytab and principal are non-empty at this point`) would make the invariant explicit for future readers.

Overall: the fix is correct, minimal, and addresses a real production issue. Good contribution.

JingsongLi:
Please rebase master.

officialasishkumar:
Rebased onto latest `master` in 0c24eb4; the focused `HadoopSecuredFileSystemTest` run passed.

JingsongLi:
+1

PR #7840, opened 2026-05-13
[core] Coalesce concurrent CachingCatalog table loads

### LINKED ISSUE #7484: [Feature] Optimize CachingCatalog#getTable
### Search before asking

### Motivation

In `CachingCatalog`, when multi thread calls the `getTable` with the same `Identifier` ,if them all find cache miss, they will call the `wrapped.getTable(identifier)` multiple times concurrently, which is very expensive. And actually, we just need to call the `wrapped.getTable(identifier)` once, and return the result for all calls.

### Solution

Only allow the first cache-missing thread to call `wrapped.getTable(identifier)`. Other concurrent threads requesting the same Identifier should wait for and reuse the result. This can be implemented by introducing a lock or a Future-based coordination mechanism.

### Anything else?

_No response_

officialasishkumar:
### Purpose

Fixes #7484.

`CachingCatalog#getTable` now uses Caffeine's atomic cache loading path for regular tables, so concurrent cache misses for the same identifier share one `wrapped.getTable` call. System tables continue to wrap the cached origin table instead of being cached directly.

### Tests

- `mvn -pl paimon-core -Pfast-build -Dtest=CachingCatalogTest#testConcurrentGetTableLoadsTableOnce test`
- `mvn -pl paimon-core -Pfast-build -Dtest=CachingCatalogTest test`
- `mvn -pl paimon-core spotless:check`

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/catalog/CachingCatalog.java
+++ b/paimon-core/src/main/java/org/apache/paimon/catalog/CachingCatalog.java
@@ -226,11 +226,6 @@ public class CachingCatalog extends DelegateCatalog {
 
     @Override
     public Table getTable(Identifier identifier) throws TableNotExistException {
-        Table table = tableCache.getIfPresent(identifier);
-        if (table != null) {
-            return table;
-        }
-
         // For system table, do not cache it directly. Instead, cache the origin table and then wrap
         // it to generate the system table.
         if (identifier.isSystemTable()) {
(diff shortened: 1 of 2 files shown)


leaves12138 (approved):
LGTM. The approach looks right: using Caffeine's atomic `Cache.get(key, mappingFunction)` coalesces concurrent misses per table identifier while preserving the existing system-table behavior (only the origin table is cached). Wrapping/unwrapping `TableNotExistException` keeps the public API semantics unchanged, and the added concurrent test covers the intended race.

JingsongLi:
+1

PR #7843, opened 2026-05-13
[core] Diff nested schema changes inside ARRAY and MAP value types

Zouxxyy:
### Purpose

Follow-up of #7789. Extend `SchemaMergingUtils.diffSchemaChanges` to emit precise `AddColumn` for new struct fields nested inside `ARRAY` element / `MAP` value, instead of a coarse `UpdateColumnType` rewriting the whole nested type.

Falls back to `UpdateColumnType` when a nested field is removed or a map key type changes.

### Tests

- `SchemaMergingUtilsTest`
- `MergeIntoTableTestBase`


JingsongLi:
+1

PR #7844, opened 2026-05-13
[python] initialize daft connector for paimon

chenghuichen:
### Purpose
Initialize `paimon-daft` module in the Paimon main repository.

Based on discussion from https://github.com/Eventual-Inc/Daft/pull/6839, the community decided to maintain the Daft-Paimon connector in the Paimon repository going forward — for more efficient maintenance and stronger ownership by the Paimon community.

This is the init PR that migrates the connector code from the Daft repository and adds basic project infrastructure.

### Tests
paimon-daft/tests/*

### DIFF
--- a/paimon-python/dev/lint-python.sh
+++ b/paimon-python/dev/lint-python.sh
@@ -136,7 +136,12 @@ function check_stage() {
 #########################
 # Flake8 check
 function flake8_check() {
-    local PYTHON_SOURCE="$(find . \( -path ./dev -o -path ./.tox -o -path ./.venv \) -prune -o -type f -name "*.py" -print )"
+    local PRUNE_PATHS="\( -path ./dev -o -path ./.tox -o -path ./.venv"
+    if python -c "import sys; sys.exit(0 if sys.version_info < (3, 7) else 1)" 2>/dev/null; then
+        PRUNE_PATHS="$PRUNE_PATHS -o -path ./pypaimon/daft"
+    fi
+    PRUNE_PATHS="$PRUNE_PATHS \)"
+    local PYTHON_SOURCE="$(eval "find . $PRUNE_PATHS -prune -o -type f -name '*.py' -print")"
 
     print_function "STAGE" "flake8 checks"
     if [ ! -f "$FLAKE8_PATH" ]; then
--- a/paimon-python/dev/lint-python.sh
+++ b/paimon-python/dev/lint-python.sh
@@ -138,7 +138,7 @@ function check_stage() {
 function flake8_check() {
     local PRUNE_PATHS="\( -path ./dev -o -path ./.tox -o -path ./.venv"
     if python -c "import sys; sys.exit(0 if sys.version_info < (3, 7) else 1)" 2>/dev/null; then
-        PRUNE_PATHS="$PRUNE_PATHS -o -path ./pypaimon/daft"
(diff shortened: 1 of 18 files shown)


JingsongLi:
Maybe just in PyPaimon like ray-data?

TheR1sing3un:
> Maybe just in PyPaimon like ray-data?

+1

I'm really looking forward to the return of this feature to pypaimon itself. Currently, the integration of ray data has been closely combined with pypaimon. I also suggest integrating in the same way as ray data. In addition, I also found that in the current implementation, there are some monkey patches for pypaimon. I think the capabilities belonging to pypaimon should be implemented within pypaimon instead of having the engine layer provide upper-level support.

chenghuichen:
Thanks @JingsongLi @TheR1sing3un for the suggestion! I looked into the ray-data integration style and it fits the daft scenario quite well. Refactored to follow the same approach within pypaimon.

paimon-python/pypaimon/daft/daft_predicate_visitor.py
```diff
@@ -0,0 +1,243 @@
...
+
+    # -- Comparison operators --
+
+    def _cmp(self, left: Expression, right: Expression, fn: Any) -> Predicate | None:
+        """Fold a binary comparison: extract col ref and literal value, then apply fn."""
+        lhs, rhs = self.visit(left), self.visit(right)
+        if isinstance(lhs, _ColRef) and not isinstance(rhs, _ColRef):
+            return fn(lhs.name, rhs)
+        if isinstance(rhs, _ColRef) and not isinstance(lhs, _ColRef):
+            return fn(rhs.name, lhs)
```
JingsongLi: For visit_less_than (left, right), the semantics are left<right. When rhs is _SlRef (i.e. 3<col), the code calls less_than (col, 3), which means col<3, but the actual semantics should be col>3. For asymmetric operators (<,<=,>,>=), reversing the position of column/literal requires reversing the operator at the same time. ==And! =Coincidentally, there's no problem because they're symmetrical.

paimon-python/dev/lint-python.sh
```diff
@@ -134,25 +134,14 @@ function check_stage() {
...
-    else
-        print_function "STAGE" "license header checks... [SUCCESS]"
-    fi
-}
-
 # Flake8 check
 function flake8_check() {
-    local PYTHON_SOURCE="$(find . \( -path ./dev -o -path ./.tox -o -path ./.venv \) -prune -o -type f -name "*.py" -print )"
+    local PRUNE_PATHS="\( -path ./dev -o -path ./.tox -o -path ./.venv"
+    if python -c "import sys; sys.exit(0 if sys.version_info < (3, 7) else 1)" 2>/dev/null; then
```
JingsongLi: The meaning of this code is to exclude the daft directory only when the Python version is below 3.7. But daft>=0.7.6 requires Python>=3.10 (defined in setup. py), so the judgment condition should be sys. version info<(3, 10). Moreover, it is unlikely that Python below 3.7 is currently in use.

paimon-python/pypaimon/daft/daft_predicate_visitor.py
chenghuichen: Fixed, thanks.

JingsongLi:
+1

PR #7845, opened 2026-05-13
[format] Support reading Parquet TIMESTAMP(NANOS) as timestamp(9)

leaves12138:
### Purpose

Native Parquet writers such as Arrow can encode `TIMESTAMP(9)` as `INT64` with the `TIMESTAMP(NANOS)` logical annotation. This is valid Parquet, but Paimon's vectorized Parquet reader did not use the logical timestamp unit when decoding `INT64` timestamp columns, so nanosecond timestamps could not be read as Paimon `timestamp(9)` correctly.

### Changes

- Decode `INT64` Parquet timestamps according to their logical time unit (`MILLIS`, `MICROS`, or `NANOS`).
- Convert `TIMESTAMP(NANOS)` schema annotations to Paimon timestamp precision 9.
- Add regression coverage for top-level `timestamp(9)` and `array<timestamp(9)>` written as Parquet `INT64 TIMESTAMP(NANOS)`.

### Tests

- `mvn -pl paimon-format -DskipTests compile`
- `mvn -pl paimon-format -Pfast-build -Dtest=ParquetReadWriteTest,ParquetSchemaConverterTest test`


JingsongLi (approved):
+1

PR #7846, opened 2026-05-13
[core] Validate file index related options

tsreaper:
### Purpose

Add schema-level validation for the file-index.<index-type>.columns options (bloom-filter / bitmap / bsi / range-bitmap).

Currently, if a user configures a column name that does not exist in the table schema (or uses the nested col[k] syntax on a non-map column / a map with a   non-string key), the table can be created and written to successfully, but the write job will fail much later — when MergeTreeWriter.flushWriteBuffer      triggers DataFileIndexWriter.<init> during checkpoint, throwing IllegalArgumentException: xxx does not exist in column fields (or is not map type / Only    support map data type with key field of CHAR、VARCHAR、STRING.). The same misconfiguration also blocks compaction.

This PR moves these three checks up-front into SchemaValidation so the error is surfaced at create-table / alter-table time:

1. Every column listed in any file-index.<type>.columns must exist in the schema.
2. If nested syntax col[k] is used, col must be of MAP type.
3. A MAP referenced via nested syntax must have CHAR / VARCHAR key type (matching MapFileIndexMaintainer's runtime requirement).

### Tests

Added two tests in SchemaValidationTest, each looping over all four index types (bloom-filter, bitmap, bsi, range-bitmap) to cover the configuration uniformly:

- testFileIndexColumns — valid columns pass; an unknown column throws with a clear "does not exist in table schema" message.
- testFileIndexNestedColumn — m[k] on MAP<STRING, INT> passes; f3[k] on a non-map column throws "is not a map type"; mi[k] on MAP<INT, INT> throws "Only CHAR/VARCHAR is supported".

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
+++ b/paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java
@@ -23,6 +23,7 @@ import org.apache.paimon.CoreOptions.ChangelogProducer;
 import org.apache.paimon.CoreOptions.MergeEngine;
 import org.apache.paimon.TableType;
 import org.apache.paimon.factories.FactoryUtil;
+import org.apache.paimon.fileindex.FileIndexOptions;
 import org.apache.paimon.format.FileFormat;
 import org.apache.paimon.mergetree.compact.aggregate.FieldAggregator;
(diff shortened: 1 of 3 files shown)


JingsongLi (approved):
+1

PR #7847, opened 2026-05-14
[python] Support null blob

XiaoHongbo-Hope:

### DIFF
--- a/paimon-python/pypaimon/filesystem/local_file_io.py
+++ b/paimon-python/pypaimon/filesystem/local_file_io.py
@@ -420,7 +420,6 @@ class LocalFileIO(FileIO):
             if data.num_columns != 1:
                 raise RuntimeError(f"Blob format only supports a single column, got {data.num_columns} columns")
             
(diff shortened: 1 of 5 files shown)


leaves12138 (approved):
+1

leaves12138:
Thanks for the patch. The null encoding/decoding is aligned with the Java blob format.

One edge case: `BlobFormatWriter.write_value` accepts `col_data is None` before validating that the field is actually a BLOB field. This means a non-BLOB single-column Arrow table with all-null values, e.g. `pa.int64()` or `pa.string()`, can now be written as a valid blob file, while non-null values of the same type are still rejected. Could we validate the field type before accepting nulls, or add a `write_blob`-level check that only `large_binary` is supported?

Also, the `file_io.write_blob` null cases currently only assert that the file exists. It would be better to add a public API round-trip test, e.g. write `[b"data", None, b"more"]` through `write_blob` and read it back with `FormatBlobReader` to verify the null value is preserved.

PR #7848, opened 2026-05-14
[core] Fix BranchesTable returning empty result for non Equal/IN predicates

heye1005:
### Purpose

`BranchesRead#withFilter` is a `// TODO`, and `BranchesRead#branches()`
only handles `Equal` / OR-IN on `branch_name` inside
`if (predicate != null)` with no else. Pushing any other predicate
into `BranchesRead` returns an empty result:

```java
rb.withFilter(builder.notEqual(0, BinaryString.fromString("b2")));
// [] on master, expected: all branches except b2


JingsongLi:
+1

PR #7850, opened 2026-05-14
[python] Fix DataBlobWriter KeyError for partial writes with blob columns

### LINKED ISSUE #7849: [Bug] DataBlobWriter._split_data raise KeyError when a partial write included blob columns
### Search before asking

### Paimon version

1.4.1

### Compute Engine

PyPaimon

### Minimal reproduce step

DataBlobWriter._split_data always selected the full normal and blob column lists from the table schema. With TableWrite.with_write_type, batches only contain the narrowed columns, so pa.RecordBatch.select(...) could reference missing names and raise KeyError when a partial write included blob columns.

Pass write_cols from FileStoreWrite into DataBlobWriter, narrow normal_column_names and blob_file_column_names to that subset (and only open blob writers for blob columns in the subset), so splits only select columns present in the batch.

### Anything else?

_No response_

SteNicholas:
### Purpose

**Problem**: For tables with BLOB columns, pypaimon uses `DataBlobWriter`, which splits each `pyarrow.RecordBatch` into “normal” columns (written to Parquet/ORC/…) and blob-file columns (written via `BlobWriter`). `_split_data` used the full table lists of normal and blob-file column names when calling `RecordBatch.select(...)`.

**Regression**: When `TableWrite.with_write_type(...)` narrows the write to a partial column list, validation ensures incoming batches only contain those `columns. _split_data` still tried to select columns not present in the batch (e.g. a normal column omitted from the partial write), which caused PyArrow to raise KeyError.

**Fix**:

- Pass `write_cols` from `FileStoreWrite` into `DataBlobWriter` (same as for `AppendOnlyDataWriter`), so the blob writer sees the narrowed column set from with_write_type.
- In `DataBlobWriter.__init__`, derive `normal_column_names` and `blob_file_column_names` from that subset when `write_cols` is set: only blob-file columns that appear in write_cols, and normal columns = write_cols minus blob-file columns (order preserved from `write_cols`). Only instantiate `BlobWriter` for blob-file columns in that narrowed set.
- Keep full-table behavior when `write_cols` is None (full schema write).
- This keeps `_split_data` consistent with the actual batch schema and matches the intent of partial / data-evolution writes.

Close #7849.

### Tests

Added in `paimon-python/pypaimon/tests/blob_table_test.py`(`DataBlobWriterTest`):

1. Partial normal + one blob — `with_write_type(['id', 'blob_data'])` with a batch that only has those columns; asserts one Parquet file (`write_cols == ['id']`), blob file(s) (write_cols == ['blob_data']), row counts, commit + read-back (unwritten name is null).
2. Partial normal only — `with_write_type(['id', 'name'])` without blob columns in the batch; asserts no .blob files, Parquet write_cols == ['id', 'name'], read-back with blob column null.
3. Two blob columns, write one — schema with blob1 and blob2, `with_write_type(['id', 'blob1'])`; asserts exactly one blob file and write_cols == ['blob1'].

### DIFF
--- a/paimon-python/pypaimon/tests/blob_table_test.py
+++ b/paimon-python/pypaimon/tests/blob_table_test.py
@@ -242,6 +242,133 @@ class DataBlobWriterTest(unittest.TestCase):
         result = table.new_read_builder().new_read().to_arrow(table.new_read_builder().new_scan().plan().splits())
         self.assertEqual(result.num_rows, 3)
 
+    def test_data_blob_writer_partial_write_with_write_type(self):
+        """Partial write (normal + blob subset) via with_write_type: split must match batch columns."""
+        from pypaimon import Schema
+
+        pa_schema = pa.schema([
+            ('id', pa.int32()),
+            ('name', pa.string()),
+            ('blob_data', pa.large_binary()),
+        ])
+        schema = Schema.from_pyarrow_schema(
+            pa_schema,
+            options={
+                'row-tracking.enabled': 'true',
+                'data-evolution.enabled': 'true'
+            }
+        )
+        self.catalog.create_table('test_db.blob_partial_write_type', schema, False)
+        table = self.catalog.get_table('test_db.blob_partial_write_type')
+
+        partial_schema = pa.schema([('id', pa.int32()), ('blob_data', pa.large_binary())])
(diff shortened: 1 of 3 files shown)


JingsongLi (approved):
+1

PR #7855, opened 2026-05-14
[codex] Support replacing index manifest in overwrite commit

leaves12138:
### Purpose

Allow overwrite-style commits that already replace manifest lists to also provide an explicit index manifest and next row id.

The existing replaceManifestList behavior stays compatible, while a same-name overload lets metadata rewrite workflows preserve or replace snapshot-level index metadata and row-id state explicitly.

### Changes

- Add a replaceManifestList overload that accepts indexManifest and nextRowId.
- Keep the existing replaceManifestList signature by delegating to the overload with the latest snapshot values.

### Verification

- mvn -pl paimon-core -DskipTests compile


yuzelin (approved):
+1

PR #7856, opened 2026-05-14
[core] Optimize String format in CacheManager

JingsongLi:


leaves12138 (approved):
Looks good to me.

This change is a small cleanup in `CacheManager`: it switches from eager `String.format(...)` to the existing `Preconditions.checkNotNull(reference, template, args...)` overload, so the error message is only formatted on the null path. The `checkNotNull` overload supports `%s` templates, so the behavior remains equivalent when the cache lookup unexpectedly returns null.

I reviewed the diff and checked the corresponding `Preconditions` API. The existing failing CI job looks unrelated to this one-line change.

PR #7857, opened 2026-05-14
[core] Support batch vector search

XiaoHongbo-Hope:

### DIFF
--- a/docs/docs/multimodal-table/global-index/vector.mdx
+++ b/docs/docs/multimodal-table/global-index/vector.mdx
@@ -182,6 +182,21 @@ try (RecordReader<InternalRow> reader = readBuilder.newRead().createReader(plan)
 }
 ```
 
+Batch results keep the same order as input vectors.
+
+```java
+float[][] queryVectors = {
+    {1.0f, 2.0f, 3.0f},
+    {3.0f, 2.0f, 1.0f}
+};
+List<GlobalIndexResult> batchResults = table.newBatchVectorSearchBuilder()
+        .withVectors(queryVectors)
+        .withLimit(5)
+        .withVectorColumn("embedding")
+        .executeBatchLocal();
+// batchResults.get(i) corresponds to queryVectors[i].
+```
+
 For Java, use `Table.newVectorSearchBuilder()` to produce a global index result, then pass
 the result to `TableScan.withGlobalIndexResult`.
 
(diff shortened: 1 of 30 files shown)


JingsongLi (changes requested):
This batch query is not meaningful for performance, we need to ensure that the local file cache is enabled. See ce38b2ccd7cbaabcfc1d5e90f68586ebecc66265

XiaoHongbo-Hope:
> This batch query is not meaningful for performance, we need to ensure that the local file cache is enabled. See [ce38b2c](https://github.com/apache/paimon/commit/ce38b2ccd7cbaabcfc1d5e90f68586ebecc66265)

Do you mean we should add a validation for `local-cache.enabled=true`?

XiaoHongbo-Hope:
@JingsongLi   Sorry, I accidentally deleted your comment while trying to reply. 

I want to confirm if I understood your point correctly.

Batch query vectors share one scan and one reader/native index loading per split within one invocation. Do you mean that local file cache for meta/global-index files is sufficient for repeated single-vector queries, so the reader/native-index reuse provided by this batch API is not necessary?

JingsongLi (commented):
Maybe we can introduce a new `BatchVectorSearch`?

JingsongLi:
Also, how is the return value handled? How do users know which input corresponds to which one?

XiaoHongbo-Hope:
> Also, how is the return value handled? How do users know which input corresponds to which one?

Results keep the same order as input vectors — results.get(i) corresponds to queryVectors[i]

XiaoHongbo-Hope:
> This batch query is not meaningful for performance, we need to ensure that the local file cache is enabled. See [ce38b2c](https://github.com/apache/paimon/commit/ce38b2ccd7cbaabcfc1d5e90f68586ebecc66265)

I ran an OSS benchmark with a 529MB Lumina index, 100 queries, dim=1024.
```
  Without Paimon local cache:
  - user-level loop: 188.6s
  - batch: 39.7s

  With warm local memory cache:
  - user-level loop: 41.4s
  - batch: 39.7s
```
  So without cache, batch is effective. With warm cache, the benefit is very small, which matches your
  point.

  Since users need a batch API, for the Spark/Flink batch API, which direction do you prefer?
  1. implement it by looping single-vector search internally and relying on local cache;
  2. continue this PR and add a batch vector search/readBatch path.

XiaoHongbo-Hope:
@JingsongLi @yunfengzhou-hub Could you help take a look again when you are free

JingsongLi (changes requested):
My suggestion is that the API design should not be mixed with previous single search, and there should also be a corresponding BatchVectorSearchBuilder.

JingsongLi:
And we should also support Python api.

JingsongLi:
+1

PR #7860, opened 2026-05-15
[core][spark] Support history-preserving replace table

Zouxxyy:
### Purpose

Support Spark replace-table flows for Paimon tables while preserving table history where possible.

This change adds staged replace support for `REPLACE TABLE` and `CREATE OR REPLACE TABLE ... AS SELECT`, using truncate plus schema replacement for compatible FileStore tables so old snapshots remain available for time travel. It also falls back to drop-and-create when the source or target is not a Paimon table, handles SparkGenericCatalog provider checks, and documents the Spark SQL behavior.

Follow-up TODOs:

- Support replace-table flows with RESTCatalog.
- Support replacing a table with different partitioning. This is tied to the underlying truncate mechanism and likely needs a separate PR.

### Tests

Added coverage for catalog-level replace table, Spark SQL replace table, create-or-replace-as-select, incompatible schema replacement, table existence errors, provider/table type fallback, and time-travel reads after replace.


JingsongLi:
We can implement REST Catalog in this PR too. This way we can have a global perspective.

JingsongLi (commented):
See below comment.

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/SparkCatalog.java
```diff
@@ -389,6 +387,65 @@ public boolean dropTable(Identifier ident) {
...
+            Identifier ident,
+            StructType schema,
+            Transform[] partitions,
+            Map<String, String> properties)
+            throws NoSuchNamespaceException, NoSuchTableException {
+        org.apache.paimon.catalog.Identifier tableIdent = toIdentifier(ident, catalogName);
+        Schema targetSchema = toInitialSchema(schema, partitions, properties);
+
+        try {
+            catalog.replaceTable(tableIdent, targetSchema, false);
```
JingsongLi, line 438: We need to be more secure. If the RESTCatalog returns an unimplemented exception, we should revert back to a working implementation.

JingsongLi (approved):
+1

PR #7866, opened 2026-05-15
feat(core): support sort manifest entries by partition

baiyangtx:
### Purpose

Manifest entries are currently written in arrival order, which scatters entries belonging
to the same partition across multiple manifest files. This leads to:

- Inefficient partition-pruning during reads — the reader must scan many manifest files
  to find all entries for a single partition.
- Suboptimal data locality when compaction reorganizes files.

This PR introduces manifest entry sorting by partition, so that entries for the same
partition are clustered together within manifest files.

### Changes

**New configuration options:**

| Option | Default | Description |
|--------|---------|-------------|
| `manifest.merge.sorted` | `true` | Sort entries by partition during manifest full compaction |
| `manifest.merge.sort-on-commit` | `false` | Sort entries by partition during manifest full merge in commit |
| `manifest.delta.sorted` | `true` | Sort entries by partition when writing manifest delta |

**Core implementation:**

- `ManifestFileMerger`: Introduces `mergeSortedByPartition()` and `mergeUnsorted()`.
  When `manifest.merge.sorted` is enabled, entries are collected into a
  `BinaryExternalSortBuffer` (with spill-to-disk support), then written in
  partition-major order: `(partition, bucket, level)`.
- `ManifestFileMerger.createManifestEntryComparator()`: Comparator used for
  sorting delta manifests, falling back to pure-Java comparison when codegen
  is unavailable.
- `FileStoreCommitImpl`: Wires sort parameters into all three paths —
  commit manifest merge, delta file writing, and manifest compaction.

### Tests

```
# Manifest merge with sorting
mvn -pl paimon-core -am -DfailIfNoTests=false \
    -Dtest=ManifestFileMetaTest test

# No-partition edge case
mvn -pl paimon-core -am -DfailIfNoTests=false \
    -Dtest=NoPartitionManifestFileMetaTest test

# Spark compact procedure
mvn -pl paimon-spark/paimon-spark-ut -am -DfailIfNoTests=false \
    -Dtest=CompactProcedureTestBase test
```

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/ManifestFileMerger.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/ManifestFileMerger.java
@@ -18,15 +18,34 @@
 
 package org.apache.paimon.operation;
 
+import org.apache.paimon.CoreOptions;
+import org.apache.paimon.codegen.NormalizedKeyComputer;
+import org.apache.paimon.codegen.RecordComparator;
+import org.apache.paimon.compression.CompressOptions;
 import org.apache.paimon.data.BinaryRow;
+import org.apache.paimon.data.GenericRow;
+import org.apache.paimon.data.InternalRow;
+import org.apache.paimon.data.serializer.BinaryRowSerializer;
+import org.apache.paimon.data.serializer.InternalRowSerializer;
+import org.apache.paimon.disk.IOManager;
 import org.apache.paimon.io.RollingFileWriter;
 import org.apache.paimon.manifest.FileEntry;
 import org.apache.paimon.manifest.ManifestEntry;
+import org.apache.paimon.manifest.ManifestEntrySerializer;
 import org.apache.paimon.manifest.ManifestFile;
 import org.apache.paimon.manifest.ManifestFileMeta;
+import org.apache.paimon.memory.HeapMemorySegmentPool;
+import org.apache.paimon.memory.MemorySegment;
+import org.apache.paimon.memory.MemorySegmentPool;
 import org.apache.paimon.partition.PartitionPredicate;
+import org.apache.paimon.sort.BinaryExternalSortBuffer;
+import org.apache.paimon.sort.BinaryInMemorySortBuffer;
+import org.apache.paimon.types.IntType;
 import org.apache.paimon.types.RowType;
 import org.apache.paimon.utils.Filter;
+import org.apache.paimon.utils.InternalRowUtils;
+import org.apache.paimon.utils.MutableObjectIterator;
+import org.apache.paimon.utils.SerializationUtils;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -34,6 +53,7 @@ import org.slf4j.LoggerFactory;
 import javax.annotation.Nullable;
 
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
@@ -45,6 +65,7 @@ import java.util.Set;
 import java.util.function.Function;
 
 import static java.util.Collections.singletonList;
+import static org.apache.paimon.codegen.CodeGenUtils.newRecordComparator;
 import static org.apache.paimon.utils.ManifestReadThreadPool.sequentialBatchedExecute;
 import static org.apache.paimon.utils.Preconditions.checkArgument;
 
@@ -66,7 +87,30 @@ public class ManifestFileMerger {
             int suggestedMinMetaCount,
             long manifestFullCompactionSize,
             RowType partitionType,
-            @Nullable Integer manifestReadParallelism) {
+            @Nullable Integer manifestReadParallelism,
+            boolean manifestMergeSorted) {
+        return merge(
+                input,
+                manifestFile,
+                suggestedMetaSize,
+                suggestedMinMetaCount,
+                manifestFullCompactionSize,
+                partitionType,
+                manifestReadParallelism,
+                manifestMergeSorted,
+                CoreOptions.MANIFEST_MERGE_SORT_BUFFER.defaultValue().getBytes());
+    }
+
+    public static List<ManifestFileMeta> merge(
+            List<ManifestFileMeta> input,
+            ManifestFile manifestFile,
+            long suggestedMetaSize,
+            int suggestedMinMetaCount,
+            long manifestFullCompactionSize,
+            RowType partitionType,
+            @Nullable Integer manifestReadParallelism,
+            boolean manifestMergeSorted,
+            long manifestMergeSortBufferSize) {
         // these are the newly created manifest files, clean them up if exception occurs
         List<ManifestFileMeta> newFilesForAbort = new ArrayList<>();
 
@@ -79,7 +123,9 @@ public class ManifestFileMerger {
                             suggestedMetaSize,
                             manifestFullCompactionSize,
                             partitionType,
-                            manifestReadParallelism);
+                            manifestReadParallelism,
+                            manifestMergeSorted,
+                            manifestMergeSortBufferSize);
             return fullCompacted.orElseGet(
                     () ->
                             tryMinorCompaction(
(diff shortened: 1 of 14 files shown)


Aitozi (approved):
LGTM +1, Please resolve the conflict

Aitozi:
CC @JingsongLi for another look

JingsongLi:
Hi @baiyangtx @Aitozi , please take a look to https://github.com/apache/paimon/pull/7842 , I didn't look closely, maybe the abilities of these two PR implementations are the same.

baiyangtx:
> Hi @baiyangtx @Aitozi , please take a look to #7842 , I didn't look closely, maybe the abilities of these two PR implementations are the same.

Thanks @discivigour for working on this too. My PR #7866 takes a different approach — sorting inline during the existing full compaction path, rather than a separate post-commit sort pipeline. Key differences:

- **No new compact type** — sorting happens inside `ManifestFileMerger.tryFullCompaction`, reusing the existing full compaction trigger and lifecycle. No additional manifest scans or rewrite passes.
- **Memory-safe with external sort** — uses `BinaryExternalSortBuffer` (already battle-tested across Paimon). When entries exceed `manifest.merge.sort.buffer`, it automatically spills to disk. No risk of OOM regardless of manifest count.
- **Finer-grained config** — separate toggles for merge compaction, commit-time merge, and delta writing, instead of a single switch.

Happy to align with the community's direction, but I believe the inline approach is simpler and less invasive.

paimon-core/src/main/java/org/apache/paimon/operation/ManifestFileMerger.java
```diff
@@ -272,27 +350,276 @@ public static Optional<List<ManifestFileMeta>> tryFullCompaction(
...
             } else {
-                entries.add(entry);
+                result.add(file);
             }
         }
+        return actualRewriteCount;
+    }
+
+    private static int mergeSortedByPartition(
+            List<ManifestFileMeta> toBeMerged,
```
discivigour, line 430: Thanks @baiyangtx for working on this too. Can this code sort the out-of-order large file manifest that already exists in the table? If the user wants to enable this parameter halfway, it seems that the feature won't work.

JingsongLi (commented):
## Review: feat(core): support sort manifest entries by partition

Good motivation -- clustering manifest entries by partition will improve partition-pruning scan performance. Below are several issues I found while reviewing.

---

### 1. Loss of ADD filter during manifest reads (performance regression)

In the original `readForFullCompaction`, the code called:
```java
manifestFile.read(file.fileName(), file.fileSize(), FileEntry.addFilter(), Filter.alwaysTrue())
```

Both `mergeUnsorted()` and `mergeSortedByPartition()` now call:
```java
manifestFile.read(file.fileName(), file.fileSize())
```

...and then filter `FileKind.DELETE` in Java. This means DELETE entries are now deserialized from disk only to be discarded. For manifest files with many DELETE entries (e.g., after bulk deletes or partition drops), this is a regression. Please push the add-filter down into the reader call.

---

### 2. Parallel manifest reading was silently removed

The original code used `sequentialBatchedExecute(reader, toBeMerged, manifestReadParallelism)` to read manifest files in parallel. The refactored `mergeUnsorted()` and `mergeSortedByPartition()` now iterate sequentially over `toBeMerged`. For large tables with hundreds of manifest files needing compaction, this could significantly increase compaction latency. The `manifestReadParallelism` parameter is now effectively unused in the full compaction path.

The test `testFullCompactionReadManifestsInParallel` was modified to remove the parallelism assertions (`maxConcurrentManifestReads`), which confirms the feature was dropped rather than preserved.

---

### 3. IOManager created from `System.getProperty("java.io.tmpdir")` 

In `mergeSortedByPartition()`:
```java
ioManager = IOManager.create(System.getProperty("java.io.tmpdir"));
```

This bypasses any user-configured temp directory or spill path. The commit code already has access to an IOManager (or its configuration) through the FileStore. Please pass the IOManager (or temp directories) from the caller rather than hard-coding the system temp directory. This also avoids creating a new IOManager per compaction.

---

### 4. Test assertions weakened to `containsExactlyInAnyOrderElementsOf`

Many tests (especially clustering/sort-related ones like `testLocalSortClusterUnpartitionedTable`) were changed from `containsExactlyElementsOf` to `containsExactlyInAnyOrderElementsOf`. This fundamentally weakens the test -- e.g., the comment in `IncrementalClusterActionITCase` says "verify internal order: within the single output file, rows must be sorted ascending by (a, b)" but then uses an assertion that does not verify ordering.

If the manifest sorting changes the read order for these tests, the correct fix is to add an `ORDER BY` clause to the query (as done in `SparkWriteITCase`) rather than dropping the ordering guarantee from the assertion. Otherwise these tests no longer validate that clustering produces correctly sorted data.

---

### 5. Inconsistent sort key between delta sort and full compaction sort

`createManifestEntryComparator` (used for delta sorting) orders by `(partition, bucket, level, fileName)`, while `mergeSortedByPartition` (used for full compaction) orders by `(partition, bucket, level)`. Without a stable tiebreaker in the full compaction sort, entries with the same (partition, bucket, level) have non-deterministic order. Consider adding fileName to the full compaction comparator as well for consistency and deterministic output.

---

### 6. Default values: `manifest.delta.sorted=true` changes write-path behavior silently

`manifest.delta.sorted` defaults to `true`, meaning every commit will now incur sorting overhead on the delta file list. Given that `manifest.merge.sort-on-commit` defaults to `false` explicitly to avoid commit-path latency impact, it seems inconsistent that delta sorting is on by default. Consider defaulting to `false` or at least calling this out in migration/upgrade notes since it changes existing behavior.

---

### Minor

- The variable name `partitionRmpr` in `createManifestEntrySortBuffer` seems like a typo. Consider `partitionCmpr` or `partitionComparator`.
- The overload chain for `merge()` and `tryFullCompaction()` (3 overloads each) is getting long. A parameter object or builder might improve readability in the future.

---

Overall the feature is valuable and the use of `BinaryExternalSortBuffer` for spill-to-disk is a sound approach. Addressing the performance regressions (filter pushdown, parallel reads) and the weakened test assertions would strengthen this PR significantly.

paimon-core/src/main/java/org/apache/paimon/operation/ManifestFileMerger.java
baiyangtx: @discivigour Good point — my approach indeed doesn't handle existing out-of-order manifests. I think our two PRs are actually complementary: yours handles the bootstrap/rewrite path for historical manifests, mine maintains sorted order going forward via inline compaction. I'd like to collaborate and align the configuration surface first so users see one coherent "manifest sort" feature rather than two separate mechanisms. WDYT?

baiyangtx:
Thanks @JingsongLi for the thorough review. I've addressed most of the feedback:

1. **ADD filter regression** — Restored `FileEntry.addFilter()` in both `mergeUnsorted()` and `mergeSortedByPartition()`. DELETE entries are now filtered at the IO level as before.

2. **IOManager hard-coded tmpdir** — Added `@Nullable IOManager` parameter to `tryFullCompaction` and `merge`. Falls back to `System.getProperty("java.io.tmpdir")` when null is passed, so existing callers are unaffected.

3. **Inconsistent sort key** — Added `fileName` as the 5th field in the sort row, serving as a deterministic tiebreaker after (partition, bucket, level).

4. **manifest.delta.sorted default** — Changed from `true` to `false`, consistent with `manifest.merge.sort-on-commit`.

5. **Test assertion for internal order** — Reverted `containsExactlyInAnyOrderElementsOf` back to `containsExactlyElementsOf` for the `testLocalSortClusterUnpartitionedTable` internal-order assertion. With parallelism=1 and a single output file, the order is deterministic.

6. **Minor** — Renamed `partitionRmpr` → `partitionCmp`.

**For discussion — parallel manifest reads (2)**: The original `sequentialBatchedExecute` was removed because the sorting path needs all entries collected before spilling to external sort. I'd appreciate guidance on whether to restore parallel reads for the non-sorted path only, or if the simplicity of unified sequential reads is acceptable here.

paimon-core/src/main/java/org/apache/paimon/operation/ManifestFileMerger.java
discivigour: What does your 'inline compaction' mean? My method has done the following things:
1. Incrementally sort the newly arrived data
2. Sort the existing data

discivigour:
Thank you for sharing this feature. I believe my feature can already support the following functions well: 
1. The LSM Tree was introduced to achieve incremental sorting, uniformly sorting existing files and newly arrived files. 
2. Integrate the three functions of merging small files, eliminating add and delete pairs, and manifest sorting into this feature. The original full compaction and minor compaction would undermine the sorting effect. 

Moreover, I think the current PR has the following issues: 
1. It does not reorder the large number of existing manifests. 
2. The manifest sort function is limited by the trigger frequency of full compaction; if only minor compaction is performed, manifest sort cannot be triggered. 
3. Manifest sort only sorts small files and delete-related files, so the sorting is not sufficient. 
4. Minor compaction will destroy the sorting effect.

To avoid conflicts and duplicated implementation, I’ll continue to improve my PR and port any useful parts if needed, with credit. -1 for this pr, thanks for understanding! 

However, thank you very much for your contribution! If you have any suggestions or find areas that need improvement in my PR #7842 , please feel free to continue providing review comments. Your feedback is very helpful for my work

JingsongLi:
https://github.com/apache/paimon/pull/7842 has been merged, if you feel there is still room for improvement, please submit a separate PR. Thanks @baiyangtx and @Aitozi

PR #7868, opened 2026-05-15
[spark] Fix merge into update columns detection

leaves12138:
## What changed

This updates Spark merge-into data evolution update column detection so target self-assignments are not treated as modified columns when Spark adds or changes qualifiers. The logic now treats matching AttributeReference exprIds as the same target column, while still including source-side assignments such as target_col = source_col.

## Why

Spark AttributeReference.equals also compares qualifiers, so the same target field can compare unequal as file_name#2 versus t.file_name#2. That made updateColumns include unchanged fields and caused partial updates to rewrite more columns than expected.

## Validation

- mvn -pl paimon-spark/paimon-spark-common -DskipTests -Dcheckstyle.skip -Drat.skip -Dspotless.check.skip=true compile
- mvn -pl paimon-spark/paimon-spark-common -DskipTests -Dcheckstyle.skip -Drat.skip spotless:check


JingsongLi (approved):
+1

PR #7869, opened 2026-05-15
[python] Add ReadBuilder.explain() for scan-plan visibility

TheR1sing3un:
## What

Add `ReadBuilder.explain()` returning a structured `ExplainResult` so users
can see what a PyPaimon read will actually do — target snapshot, pushed-down
predicate / projection / limit, partition / bucket / file-stats pruning
funnel, and split-level execution signals (raw-convertible ratio, deletion-
vector ratio, level histogram, split-size skew).

The default `__str__` is a compact debug layout; `verbose=True` lists every
split. Reads manifest list + manifests only — data files are never opened.

## Why

`Plan` exposes only `splits` and `snapshot_id` today; `FileScanner` already
does partition / bucket / file-stats pruning but none of that is visible to
users. The only way to inspect cost is reading INFO logs or walking
`plan().splits()` by hand. Apache Paimon Java has no SQL EXPLAIN of its own
either (that comes from Flink / Spark); this PR is scoped to scan-plan
visibility, not query planning.

## Sample output

PK + partition + HASH_FIXED bucket, predicate `dt = '2026-05-12' AND id = 7`:

```
== PyPaimon Scan Plan ==
Table:              default.demo (PK, HASH_FIXED)
Snapshot:           5  (schema 0)
Predicate:          (dt = '2026-05-12') AND (id = 7)
Projection:         [dt, id, val]
Limit:              100

Partition pruning:  20 -> 4  (pruned 16)
Bucket pruning:     4 -> 1  (pruned 3)
File skipping:      1 -> 1  (pruned 0)

Splits:             1
  raw-convertible:  1 / 1
  with DV:          0 / 1
  all-above-L0:     0 / 1
  files/split:      min=1  max=1  avg=1.00
  size/split:       min=2.6 KiB  p50=2.6 KiB  p95=2.6 KiB  max=2.6 KiB

Files:              1
Total size:         2.6 KiB
Estimated rows:     10   (merged: 10)
Level histogram:    L0=1
Deletion files:     0
```

## Tests

`pypaimon/tests/read_builder_explain_test.py` covers 7 scenarios:
append-only baseline, PK/partition/bucket pruning funnel, predicate
rendering, verbose splits, empty snapshot, split-level signals, pretty-print
smoke. Full read regression is clean.

## API / format impact

New API only: `ReadBuilder.explain(verbose=False) -> ExplainResult`. Hot
read path untouched — `ScanStats` is opt-in and only enabled by `explain()`.
No data / wire format change. No Java-side change.

## Follow-up

A follow-up patch will surface `explain` through the pypaimon CLI
(alongside `cli_sql` / `cli_table`) so users can inspect a query plan from
the command line without writing any Python. A `# TODO` next to
`ReadBuilder.explain` marks the entry point.

## Generative AI usage

Drafted with the help of Claude Code; reviewed and tested locally by the
author.

### DIFF
--- a/paimon-python/README.md
+++ b/paimon-python/README.md
@@ -31,3 +31,36 @@ pip3 install dist/*.tar.gz
 
 The command will install the package and core dependencies to your local Python environment.
 
+# Inspecting query plans
+
+`ReadBuilder.explain()` returns a structured scan plan so you can see what
+a query will actually do — which snapshot it targets, what predicate /
+projection / limit were pushed down, how partition / bucket / file-stats
+pruning trimmed the scan, and split-level execution signals such as the
+raw-convertible ratio, deletion-vector ratio, and split-size skew.
+
+```python
+read_builder = (
+    table.new_read_builder()
+         .with_filter(predicate)
+         .with_projection(['dt', 'user_id'])
+         .with_limit(1000)
+)
+
+# Compact layout
+print(read_builder.explain())
+
+# Verbose layout: also lists every split
+print(read_builder.explain(verbose=True))
+
+# Programmatic access
+result = read_builder.explain()
+result.split_count
+result.partition_pruning            # PruningStat(before=..., after=...) or None
+result.splits_raw_convertible
+result.split_size_p95
+```
+
+`explain()` runs one planning pass (manifest list + manifests only — data
+files are never opened).
+
--- a/paimon-python/README.md
+++ b/paimon-python/README.md
@@ -31,36 +31,3 @@ pip3 install dist/*.tar.gz
 
 The command will install the package and core dependencies to your local Python environment.
 
-# Inspecting query plans
-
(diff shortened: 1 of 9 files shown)


paimon-python/README.md
```diff
@@ -31,3 +31,36 @@ pip3 install dist/*.tar.gz
 
 The command will install the package and core dependencies to your local Python environment.
 
+# Inspecting query plans
```
JingsongLi: Do not doc here, doc in docs.

TheR1sing3un: > Do not doc here, doc in docs.

done

JingsongLi:
@TheR1sing3un You can add doc to docs.

TheR1sing3un:
> @TheR1sing3un You can add doc to docs.

done~

JingsongLi:
+1

PR #7870, opened 2026-05-15
[python] Add parallel split reading to to_pandas / to_arrow

TheR1sing3un:
## Purpose

Today `TableRead.to_pandas` / `to_arrow` iterate splits serially in `_arrow_batch_generator`, so wall time scales linearly with the number of splits even though PyArrow's parquet/orc readers release the GIL during decode. Unlike Java, where Flink/Spark fan splits out across TaskManagers/Executors, PyPaimon has no external framework above the SDK; split-level parallelism therefore has to live inside the SDK.

This PR adds an opt-in `max_workers` parameter to `to_pandas` / `to_arrow`. Default behavior is unchanged.

## Linked issue

N/A — direct contribution.

## API

```python
read.to_arrow(splits, max_workers=4)
read.to_pandas(splits, max_workers=4)
```

- `max_workers=None` (default) or `1` → original serial path, no thread pool created.
- `>= 2` with at least 2 splits → `ThreadPoolExecutor` runs splits concurrently; the final `Table` is assembled in the input splits' order (results collected by submission index).
- `< 1` → `ValueError`.

Other `to_*` methods (`to_arrow_batch_reader`, `to_iterator`, `to_duckdb`, `to_ray`, `to_torch`) are intentionally untouched — their order-preserving / streaming semantics deserve a separate look.

## Correctness under `limit`

`_RemainingRows` is a thread-safe row-quota counter shared by all workers. Quota is *pre-debited* under a single lock so the combined output never exceeds `self.limit`, even if individual readers decode one extra batch after the quota is gone (the surplus batch is simply dropped, never emitted).

## Resource handling

Each worker uses `try/finally: reader.close()`. `ThreadPoolExecutor`'s wait-on-exit guarantees every started reader is closed before `to_arrow` returns, even when one worker raises and propagates its exception.

## Tests

Added `paimon-python/pypaimon/tests/reader_parallel_test.py` (16 tests):

- `_RemainingRows`: unbounded, bounded pre-debit, zero-request, 8-thread contention.
- Append-only multi-partition: parallel result is byte-equal to serial.
- PK merge-on-read multi-bucket: parallel + serial produce the same merged rows.
- `limit` + parallel: 10 repeated runs return exactly the configured row count.
- Edge cases: empty splits with `max_workers=4`, `max_workers` exceeding split count, `max_workers=0/-1` rejected, `max_workers=1` matches serial, `include_row_kind=True` parity.
- Reader error propagation: when one split's `create_reader` raises, the exception surfaces from `to_pandas` and sibling readers are cleaned up.

## API / format impact

- API: additive only (one new optional parameter; default preserves existing behavior).
- Storage / on-disk format: no change.
- `CoreOptions`: no new option introduced in this round.

## Documentation impact

Docstrings on `to_arrow` / `to_pandas` updated. Design doc added at `paimon-python/docs/design/2026-05-15-pypaimon-parallel-to-pandas.md`. README untouched.

## Generative AI disclosure

Yes — the implementation, tests, and design doc were drafted with Claude Code assistance under my direction and review.

### DIFF
--- a/paimon-python/pypaimon/read/table_read.py
+++ b/paimon-python/pypaimon/read/table_read.py
@@ -15,6 +15,8 @@
 # specific language governing permissions and limitations
 # under the License.
 
+import threading
+from concurrent.futures import ThreadPoolExecutor, as_completed
 from typing import Any, Dict, Iterator, List, Optional
 
 import pandas
@@ -30,6 +32,41 @@ from pypaimon.schema.data_types import DataField, PyarrowFieldParser
 from pypaimon.table.row.offset_row import OffsetRow
 
 ROW_KIND_COLUMN = "_row_kind"
+
+
+class _RemainingRows:
+    """Thread-safe remaining-rows counter for parallel reads.
+
+    Row quota is pre-debited under a single lock so that any rows that
+    threads commit to emit are guaranteed not to overshoot the limit, even
+    if individual readers keep decoding one extra batch after the quota is
+    exhausted.
+
+    When ``limit`` is None the counter is unbounded and ``try_consume``
+    always returns the requested row count.
+    """
+
+    def __init__(self, limit: Optional[int]):
+        self._lock = threading.Lock()
+        self._remaining = limit  # None == unlimited
+
+    def try_consume(self, requested: int) -> int:
+        if self._remaining is None:
+            return requested
+        if requested <= 0:
+            return 0
+        with self._lock:
+            if self._remaining <= 0:
+                return 0
+            allowed = min(requested, self._remaining)
+            self._remaining -= allowed
+            return allowed
+
+    def exhausted(self) -> bool:
+        if self._remaining is None:
+            return False
+        with self._lock:
+            return self._remaining <= 0
 
 
 class TableRead:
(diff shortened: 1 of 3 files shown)


paimon-python/pypaimon/read/table_read.py
```diff
@@ -104,13 +141,31 @@ def _try_to_pad_batch_by_schema(batch: pyarrow.RecordBatch, target_schema):
 
         return pyarrow.RecordBatch.from_arrays(columns, schema=target_schema)
 
-    def to_arrow(self, splits: List[Split]) -> Optional[pyarrow.Table]:
-        batch_reader = self.to_arrow_batch_reader(splits)
+    def to_arrow(
+        self,
+        splits: List[Split],
+        max_workers: Optional[int] = None,
```
JingsongLi: Maybe it is better to provide an table option `read.parallelism`?

TheR1sing3un: > Maybe it is better to provide an table option `read.parallelism`?

A very good suggestion. I added this option. 
By the way, since this option is a table option, and for read apis like `to_pandas`/`to_arrow`, is it still necessary to retain a parameter to override the parallelism of the table parameter? I have currently provided this parameter. If you think it's not necessary, I can remove it again.

JingsongLi (approved):
+1

Issue #7859, opened 2026-05-15
[Bug] Kafka Debezium Avro deserialization uses wrong schema subject for multi-topic sources


vikasap:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Paimon version

1.4.1

### Compute Engine

Flink

### Minimal reproduce step

Minimal repro:

1. Create two Kafka topics with Debezium Avro records registered under different Confluent Schema Registry subjects, for example:

```text
db.public.orders
db.public.users
```

2. Run one Paimon `kafka_sync_database` job with both topics and Debezium Avro:

```bash
kafka_sync_database \
  --warehouse /tmp/paimon \
  --database test_db \
  --kafka_conf properties.bootstrap.servers=<broker> \
  --kafka_conf properties.group.id=paimon-multi-topic-avro-repro \
  --kafka_conf value.format=debezium-avro \
  --kafka_conf schema.registry.url=<schema-registry-url> \
  --kafka_conf 'topic=db.public.orders;db.public.users'
```

Equivalent repro with `topic-pattern`:

```bash
--kafka_conf 'topic-pattern=db\.public\.(orders|users)'
```

3. Produce/read records from both topics.

Expected: each record is deserialized using its own topic subject, e.g. `db.public.orders-key` or `db.public.users-key`.

Actual: Paimon resolves one topic once and uses that subject for records from both topics, causing Schema Registry lookup/deserialization failures for records from the other topic.

### What doesn't meet your expectations?

We see the following stacktrace when using the kafka_sync_database

```text

java.io.IOException: Failed to deserialize consumer record due to
    at org.apache.flink.connector.kafka.source.reader.KafkaRecordEmitter.emitRecord(KafkaRecordEmitter.java:56)
    at org.apache.flink.connector.kafka.source.reader.KafkaRecordEmitter.emitRecord(KafkaRecordEmitter.java:33)
    at org.apache.flink.connector.base.source.reader.SourceReaderBase.pollNext(SourceReaderBase.java:164)
    at org.apache.flink.streaming.api.operators.SourceOperator.emitNext(SourceOperator.java:493)
    at org.apache.flink.streaming.runtime.io.StreamTaskSourceInput.emitNext(StreamTaskSourceInput.java:68)
    at org.apache.flink.streaming.runtime.tasks.StreamTask.processInput(StreamTask.java:638)
[...]

PR #7871, opened 2026-05-15
[flink] Fix Debezium Avro deserialization for multi topic source

ArnavBalyan:
### Purpose
   - KafkaDebeziumAvroDeserializationSchema returns 1 topic via `findOneTopic` at init time, this gets reused for every incoming record.
   - With multi topic configs (topic=a;b), any message coming from another topic leads to a crash with `SerializationException: The given schema does not match`. The same cached topic is also passed into `CdcSourceRecord`, misrouting records to the wrong table.
   - Read topic from the actual message for deserialize instead of using the cached value at init time.
   - Closes user reported bug #7859.
 
### Tests
 - Existing UTs

### DIFF
--- a/paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/action/cdc/kafka/KafkaDebeziumAvroDeserializationSchema.java
+++ b/paimon-flink/paimon-flink-cdc/src/main/java/org/apache/paimon/flink/action/cdc/kafka/KafkaDebeziumAvroDeserializationSchema.java
@@ -41,14 +41,12 @@ public class KafkaDebeziumAvroDeserializationSchema
 
     private static final long serialVersionUID = 1L;
 
-    private final String topic;
     private final String schemaRegistryUrl;
 
     /** The deserializer to deserialize Debezium Avro data. */
     private ConfluentAvroDeserializationSchema avroDeserializer;
 
     public KafkaDebeziumAvroDeserializationSchema(Configuration cdcSourceConfig) {
-        this.topic = KafkaActionUtils.findOneTopic(cdcSourceConfig);
         this.schemaRegistryUrl = cdcSourceConfig.get(SCHEMA_REGISTRY_URL);
     }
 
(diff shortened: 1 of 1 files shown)


ArnavBalyan:
cc @JingsongLi thanks! :)

ArnavBalyan:
cc @XiaoHongbo-Hope could you PTAL thanks!

XiaoHongbo-Hope:
> cc @XiaoHongbo-Hope could you PTAL thanks!

Sure.

ArnavBalyan:
thanks!

ArnavBalyan:
cc @JingsongLi @XiaoHongbo-Hope gentle reminder could PTAL minor bug fix thanks!

ArnavBalyan:
Gentle reminder thanks!

JingsongLi (commented):
**Review of the fix**

The change is correct and minimal. The root cause is clear: `findOneTopic()` picks the first topic from the config at construction time and caches it as a field, but in a multi-topic source (semicolon-separated list or topic-pattern), messages from other topics arrive and get deserialized/routed using the wrong topic name. This causes both a Schema Registry subject mismatch (crash) and incorrect `CdcSourceRecord` routing.

**Correctness**: Using `message.topic()` per record is the right fix. It resolves both problems:
1. The Confluent Schema Registry subject lookup now uses the actual topic the message originated from.
2. The `CdcSourceRecord` constructed downstream already received the local `topic` variable (which previously held the cached field), so it will now carry the correct per-record topic for table routing.

**Consistency with existing code**: The JSON counterpart (`KafkaDebeziumJsonDeserializationSchema`) already uses `message.topic()` directly, so this brings the Avro schema in line with the established pattern.

**One minor observation**: The `topic` field removal also removes the only usage of `KafkaActionUtils.findOneTopic` in the deserialization path. The import of `KafkaActionUtils` is still needed (for `extractKafkaMetadata`), so no dead-import issue here. However, it may be worth confirming that CI still passes given `findOneTopic` in the topic-pattern case used to eagerly connect to the Kafka admin client at construction time -- removing that call means the admin connection no longer happens at deserialization-schema instantiation, which is actually a positive side effect (no unnecessary admin call for Avro schemas).

**Suggestion for future improvement (non-blocking)**: Consider adding a targeted unit test that exercises multi-topic deserialization with distinct schemas to prevent regression. The PR description mentions "Existing UTs" pass, which is sufficient for merge, but a dedicated test for this scenario would strengthen coverage.

Overall this is a clean, well-scoped bug fix. LGTM.

JingsongLi:
+1

PR #7873, opened 2026-05-16
[python] Extend commit protocol for compaction (DataIncrement/CompactIncrement)

TheR1sing3un:
## Purpose

Align `CommitMessage` with Java's `CommitMessageImpl` shape and add a JSON-safe wire format, so later compaction work has somewhere to plug `compact_before` / `compact_after` files and a serializer to ship them through Ray workers.

Foundation only — read / write / commit produce the same snapshots.

Split from #7771.

## Changes

- `DataIncrement` / `CompactIncrement` value objects; `CommitMessage` now holds `(partition, bucket, total_buckets, data_increment, compact_increment, check_from_snapshot)`. Convenience properties preserve `msg.new_files` / `msg.compact_before` ergonomics.
- `FileStoreCommit` emits ADD for `compact_after`, DELETE for `compact_before`, auto-picks `commit_kind=COMPACT` when only compact increments are present. New `commit_compact()` skips row-id assignment.
- `DataFileMeta.to_dict` / `from_dict` with tagged encoding for `bytes` / `Decimal` / `datetime` / `date` / `time` / `Timestamp`; `encode_value` / `decode_value` public for `CommitMessage.partition` round-trip.
- `CommitMessageSerializer` (VERSION=1) covers the full `DataIncrement` + `CompactIncrement` shape including `IndexFileMeta`.

## Tests

- `commit_message_serializer_test` — round-trip with non-JSON-native partition values + index files + version rejection.
- `file_store_commit_compact_test` — `compact_before` → DELETE, `compact_after` → ADD, auto-COMPACT kind.
- Existing `file_store_commit_test` / `partition_predicate_test` / `table_commit_test` adapted to the new `CommitMessage` signature.

### DIFF
--- a/paimon-python/pypaimon/write/file_store_commit.py
+++ b/paimon-python/pypaimon/write/file_store_commit.py
@@ -168,11 +168,16 @@ class FileStoreCommit:
 
     def _commit_data(self, commit_messages: List[CommitMessage], commit_identifier: int,
                      index_adds: List, index_deletes: List) -> None:
-        # Extract the minimum check_from_snapshot from commit messages
+        # Only data-side messages (those contributing new/deleted files) anchor
+        # row-id conflict detection; compaction-only messages must not. Assign
+        # unconditionally so a prior commit's anchor cannot leak into this
+        # APPEND/OVERWRITE commit via the reusable ConflictDetection instance
+        # (StreamTableCommit reuses the same FileStoreCommit across commits).
         valid_snapshots = [msg.check_from_snapshot for msg in commit_messages
-                           if msg.check_from_snapshot != -1]
-        if valid_snapshots:
-            self.conflict_detection._row_id_check_from_snapshot = min(valid_snapshots)
+                           if msg.check_from_snapshot != -1
+                           and not msg.data_increment.is_empty()]
+        self.conflict_detection._row_id_check_from_snapshot = (
+            min(valid_snapshots) if valid_snapshots else None)
 
         commit_entries = self._build_data_entries(commit_messages)
         changelog_entries = self._collect_changelog_entries(commit_messages)
--- a/paimon-python/pypaimon/write/file_store_commit.py
+++ b/paimon-python/pypaimon/write/file_store_commit.py
@@ -124,32 +124,61 @@ class FileStoreCommit:
         self.rollback = CommitRollback(table_rollback) if table_rollback is not None else None
 
     def commit(self, commit_messages: List[CommitMessage], commit_identifier: int):
-        """Commit the given commit messages in normal append mode."""
+        """Commit the given messages, splitting write and compaction results.
+
+        Each CommitMessage may carry a data_increment (normal write), a
+        compact_increment (compaction result), or both. Data increments feed an
+        APPEND (or OVERWRITE) snapshot; compact increments feed a separate
+        COMPACT snapshot, so a single commit() call produces up to two
+        snapshots, mirroring Java's FileStoreCommitImpl. Compaction never
+        assigns row-ids or runs conflict detection.
+        """
         if not commit_messages:
             return
 
-        # Extract the minimum check_from_snapshot from commit messages
-        valid_snapshots = [msg.check_from_snapshot for msg in commit_messages
-                           if msg.check_from_snapshot != -1]
-        if valid_snapshots:
-            self.conflict_detection._row_id_check_from_snapshot = min(valid_snapshots)
+        self._reject_unwired_slots(commit_messages)
 
         logger.info(
             "Ready to commit to table %s, number of commit messages: %d",
             self.table.identifier,
             len(commit_messages),
         )
-        commit_entries = self._collect_manifest_entries(commit_messages)
+
+        # index_adds / index_deletes ride with the snapshot their payload
+        # belongs to: messages carrying a compact increment go to the COMPACT
+        # snapshot, everything else (writes and pure index updates) goes to the
+        # APPEND snapshot. Attributing them to exactly one side keeps the same
+        # index manifest change from being committed twice.
+        data_index_adds, data_index_deletes = [], []
+        compact_index_adds, compact_index_deletes = [], []
+        for msg in commit_messages:
+            if msg.compact_increment.is_empty():
+                data_index_adds.extend(msg.index_adds)
+                data_index_deletes.extend(msg.index_deletes)
+            else:
+                compact_index_adds.extend(msg.index_adds)
+                compact_index_deletes.extend(msg.index_deletes)
+
+        # APPEND/OVERWRITE snapshot first, COMPACT second, mirroring Java's
+        # FileStoreCommitImpl.commit ordering.
+        self._commit_data(commit_messages, commit_identifier,
+                          data_index_adds, data_index_deletes)
+        self._commit_compact(commit_messages, commit_identifier,
+                             compact_index_adds, compact_index_deletes)
+
+    def _commit_data(self, commit_messages: List[CommitMessage], commit_identifier: int,
+                     index_adds: List, index_deletes: List) -> None:
+        # Extract the minimum check_from_snapshot from commit messages
+        valid_snapshots = [msg.check_from_snapshot for msg in commit_messages
+                           if msg.check_from_snapshot != -1]
+        if valid_snapshots:
+            self.conflict_detection._row_id_check_from_snapshot = min(valid_snapshots)
+
+        commit_entries = self._build_data_entries(commit_messages)
         changelog_entries = self._collect_changelog_entries(commit_messages)
 
         logger.info("Finished collecting changes, including: %d entries, %d changelog entries",
                     len(commit_entries), len(changelog_entries))
-
-        index_deletes = []
-        index_adds = []
-        for msg in commit_messages:
-            index_deletes.extend(msg.index_deletes)
-            index_adds.extend(msg.index_adds)
 
         if not index_deletes:
             from pypaimon.write.global_index_update_checker import (
@@ -186,6 +215,12 @@ class FileStoreCommit:
         if self.conflict_detection.has_global_index_additions(index_adds):
             detect_conflicts = True
 
+        # Nothing on the data side to commit (e.g. a compaction-only batch):
+        # skip so we do not emit an empty APPEND snapshot. Mirrors the
+        # short-circuit inside _try_commit.
+        if not commit_entries and not index_deletes and not index_adds:
+            return
+
         self._try_commit(commit_kind=commit_kind,
                          commit_identifier=commit_identifier,
                          commit_entries_plan=lambda snapshot: commit_entries,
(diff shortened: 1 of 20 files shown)


TheR1sing3un:
Ready for review, 1st pr of the entire compaction feature.

JingsongLi (commented):
## Review: [python] Extend commit protocol for compaction (DataIncrement/CompactIncrement)

Overall this is a well-structured PR that cleanly separates write-side and compaction-side semantics by introducing `DataIncrement` / `CompactIncrement` value objects, aligning pypaimon's `CommitMessage` with Java's `CommitMessageImpl`. The code is clean, well-documented, and has good test coverage. A few observations:

### Correctness

1. **`encode_value` does not handle `list` / `tuple` values.** If a GenericRow or partition column ever carries an ARRAY-typed value, serialization will raise `TypeError`. This may be fine for the current scope (partition keys are typically scalar), but worth a brief comment or a graceful error message mentioning "ARRAY" if it's intentionally out of scope.

2. **`_index_file_to_dict` drops `dv_ranges` and `global_index_meta`** — the comment explains this is deferred to Phase 6/7. Just want to flag: if any code path populates those fields *before* Phase 6/7 lands, a round-trip through the serializer will silently lose them. The `IndexFileMeta.__eq__` only checks the four scalar fields so tests won't catch this. Consider adding an assertion/warning in `_index_file_to_dict` if those fields are non-None, similar to how `_build_commit_entries` rejects un-wired increment slots.

### Design

3. **`_build_commit_entries` entry ordering.** For a message with both `new_files` and `compact_before`/`compact_after` (currently rejected at the API level but structurally possible), the method emits ADD(new_files), then DELETE(compact_before), then ADD(compact_after). The explicit separation of `commit()` vs `commit_compact()` makes this a non-issue today, but the ordering assumption should be documented in a comment on `_build_commit_entries` for future maintainers who may relax the validation.

4. **`total_buckets` fallback logic is good.** Using the message's `total_buckets` (captured at plan time) over the table's current value correctly handles bucket rescale races. The test `test_build_entries_uses_message_total_buckets_when_set` explicitly covers this, which is great.

5. **Convenience properties on `CommitMessage`** (`new_files`, `compact_before`, etc.) preserve ergonomics nicely. These delegate cleanly and avoid breaking existing call sites.

### Minor / Nits

6. In `_generic_row_to_dict`, the fallback path `[row.get_field(i) for i in range(len(row))]` assumes the row implements `__len__`. A brief type annotation or protocol check (e.g., `InternalRow`) would improve clarity for readers unfamiliar with the row hierarchy.

7. The `CommitMessageSerializer` uses `separators=(",", ":")` for compact JSON output — good choice for a wire format since it minimizes payload size without sacrificing readability when pretty-printed for debugging.

8. `DataIncrement.empty_increment()` and `CompactIncrement.empty_increment()` class methods duplicate what `DataIncrement()` / `CompactIncrement()` already do with default factories. Unless these are intended as semantic markers for readability in call sites, they could be removed to reduce API surface.

### Summary

Solid foundation work. The separation of concerns between write and compaction paths is well-motivated, the serializer is defensively versioned, and the fail-loud approach for un-wired slots (`NotImplementedError`) is the right call for incremental development. The test suite covers the key round-trip and rejection scenarios thoroughly.

JingsongLi:
I think the new `CommitMessageSerializer` drops one existing part of `CommitMessage`.

`CommitMessage` has `index_deletes`, and `is_empty()` treats it as meaningful content. The global-index update path can build messages that only contain `index_deletes` (`build_index_delete_msgs(...)`). But `CommitMessageSerializer.to_dict/from_dict` only round-trips partition, bucket, total_buckets, data/compact increments, and `check_from_snapshot`; it does not serialize `index_deletes`.

If such a message is shipped through this serializer, the deserialized message will lose the index manifest deletions and the commit can leave stale global-index entries behind. Could we include `index_deletes` in the wire format and add a round-trip test for a message whose only content is `index_deletes`?

TheR1sing3un:
> I think the new `CommitMessageSerializer` drops one existing part of `CommitMessage`.
> 
> `CommitMessage` has `index_deletes`, and `is_empty()` treats it as meaningful content. The global-index update path can build messages that only contain `index_deletes` (`build_index_delete_msgs(...)`). But `CommitMessageSerializer.to_dict/from_dict` only round-trips partition, bucket, total_buckets, data/compact increments, and `check_from_snapshot`; it does not serialize `index_deletes`.
> 
> If such a message is shipped through this serializer, the deserialized message will lose the index manifest deletions and the commit can leave stale global-index entries behind. Could we include `index_deletes` in the wire format and add a round-trip test for a message whose only content is `index_deletes`?

Nice catch, fixed~

TheR1sing3un:
@JingsongLi @XiaoHongbo-Hope Hi, could me help to review it? Hope to land it!

paimon-python/pypaimon/write/file_store_commit.py
```diff
@@ -189,6 +191,98 @@ def commit(self, commit_messages: List[CommitMessage], commit_identifier: int):
...
+            self.table.identifier,
+            len(commit_messages),
+        )
+        commit_entries = self._build_commit_entries(commit_messages)
+        if not commit_entries:
+            return
+
+        logger.info("Finished collecting compact changes: %d entries", len(commit_entries))
+
+        self._try_commit(
```
JingsongLi, line 254: `commit_compact()` drops `CommitMessage.index_deletes`. The normal `commit()` path collects these deletions and passes them into `_try_commit`, and the serializer now explicitly preserves `index_deletes` because a message can be non-empty only due to index manifest deletions. Here, however, compact messages are neither rejected when `msg.index_deletes` is set nor forwarded to `_try_commit`, so any compaction task that emits index manifest deletions would silently leave the old index manifest entries behind. Could we either collect/pass `index_deletes` here like `commit()` does, or explicitly reject them until compact-side index deletion is wired, with a regression test?

TheR1sing3un: fixed

paimon-python/pypaimon/write/file_store_commit.py
```diff
@@ -189,6 +191,116 @@ def commit(self, commit_messages: List[CommitMessage], commit_identifier: int):
...
+
+        Each message must carry no new_files; compact_before → DELETE entries,
+        compact_after → ADD entries; snapshot kind = COMPACT. Skips row-id
+        assignment and conflict detection (a compaction never produces new rows).
+        """
+        if not commit_messages:
+            return
+
+        for msg in commit_messages:
+            if msg.new_files:
```
JingsongLi: `commit_compact()` should reject any non-empty `data_increment`, not only `new_files`. Today a message such as `CommitMessage(..., data_increment=DataIncrement(changelog_files=[...]))` passes this check, `_build_commit_entries()` deliberately produces no delta entries for data-side changelog files, and the `if not commit_entries and not index_deletes` branch below returns without committing or raising. That silently drops a non-empty commit message. Please change this guard to `if not msg.data_increment.is_empty(): raise ...` (or explicitly route data-side changelog entries through `commit()`), and add a regression test for a data-increment changelog-only compact message.

paimon-python/pypaimon/write/file_store_commit.py
```diff
@@ -189,6 +191,116 @@ def commit(self, commit_messages: List[CommitMessage], commit_identifier: int):
                          allow_rollback=allow_rollback,
                          index_deletes=index_deletes)
 
+    def commit_compact(self, commit_messages: List[CommitMessage], commit_identifier: int):
+        """Commit compaction-only messages.
+
+        Each message must carry no new_files; compact_before → DELETE entries,
+        compact_after → ADD entries; snapshot kind = COMPACT. Skips row-id
```
JingsongLi: Once `commit_compact()` accepts `compact_after` files, the abort path needs to clean them as well. `FileStoreCommit.abort()` still iterates only over `message.new_files + message.changelog_files`, so a failed/cancelled compaction would leave the rewritten `compact_after` files (and any future `compact_changelog_files`) behind, unlike the Java abort path which includes `compactIncrement().compactAfter()` and compact changelog files. Please extend `abort()` and add a regression test for aborting a compact message.

TheR1sing3un: fixed

paimon-python/pypaimon/write/table_update_by_row_id.py
```diff
@@ -382,7 +383,7 @@ def _write_group(self, partition: GenericRow, first_row_id: int,
                     CommitMessage(
                         partition=partition_tuple,
                         bucket=0,
-                        new_files=new_files,
+                        data_increment=DataIncrement(new_files=new_files),
```
JingsongLi, line 576: This constructor is the one write-side path that still does not carry the planned bucket count into the new protocol. `FileStoreWrite.prepare_commit()` and `ShardTableUpdator.prepare_commit()` both set `total_buckets=self.table.total_buckets`, and `_build_commit_entries()` now relies on `msg.total_buckets` to avoid rewriting a stale plan with the table value observed at commit time. Row-id updates created here will fall back to the current table total, so the new protection is not applied to this path. Please set `total_buckets` here as well (and ideally add a small regression alongside the existing total_buckets test).

TheR1sing3un: fixed

paimon-python/pypaimon/write/file_store_commit.py
```diff
@@ -189,6 +191,118 @@ def commit(self, commit_messages: List[CommitMessage], commit_identifier: int):
                          allow_rollback=allow_rollback,
                          index_deletes=index_deletes)
 
+    def commit_compact(self, commit_messages: List[CommitMessage], commit_identifier: int):
+        """Commit compaction-only messages.
+
+        Each message must carry no data increment; compact_before → DELETE
+        entries, compact_after → ADD entries; snapshot kind = COMPACT. Skips
```
JingsongLi: This says COMPACT skips row-id assignment, but `commit_compact()` still calls `_try_commit()` with `commit_kind="COMPACT"`, and `_try_commit()` applies `_assign_snapshot_id()` / `_assign_row_tracking_meta()` whenever row tracking is enabled. That means a `compact_after` file with no `first_row_id` is treated like a new append file, gets a fresh row-id range from `latest_snapshot.next_row_id`, and the COMPACT snapshot advances `next_row_id`. Compaction should preserve the logical row ids of the compacted input files, otherwise row-id based updates/conflict checks can no longer locate those rows after Python commits the compaction. Please gate the row-tracking assignment to write/overwrite commits (or explicitly skip it for COMPACT) and add a row-tracking compact test that verifies `compact_after.first_row_id` and `next_row_id` are not rewritten.

paimon-python/pypaimon/write/file_store_commit.py
```diff
@@ -189,6 +191,118 @@ def commit(self, commit_messages: List[CommitMessage], commit_identifier: int):
...
+                    "the feature that first needs each one."
+                )
+
+        entries: List[ManifestEntry] = []
+        for msg in commit_messages:
+            partition = GenericRow(list(msg.partition), self.table.partition_keys_fields)
+            # Prefer the message's total_buckets (captured when the plan was
+            # built) over the current table value, so a plan that survived a
+            # bucket rescale is not silently rewritten with the new count.
+            total_buckets = msg.total_buckets if msg.total_buckets is not None \
```
JingsongLi: This fixes `total_buckets` for the delta manifest entries, but `_collect_changelog_entries()` still writes changelog manifest entries with `self.table.total_buckets`. For a write plan prepared before a bucket-count change, the data entries now keep `msg.total_buckets` while the changelog entries are silently rewritten with the current table count, so the same commit can publish inconsistent bucket metadata. Please apply the same `msg.total_buckets if msg.total_buckets is not None else self.table.total_buckets` logic in `_collect_changelog_entries()` (and ideally cover a message with both `new_files` and `changelog_files`).

TheR1sing3un: fixed

paimon-python/pypaimon/write/file_store_commit.py
TheR1sing3un: fixed

paimon-python/pypaimon/write/file_store_commit.py
```diff
@@ -114,10 +114,22 @@ def __init__(self, snapshot_commit: SnapshotCommit, table, commit_user: str,
...
+        and skip row-id assignment.
+        """
         if not commit_messages:
             return
 
+        for msg in commit_messages:
+            if not msg.compact_increment.is_empty():
+                raise ValueError(
+                    "commit() rejects messages carrying compact_increment; "
+                    "use commit_compact() for compaction results."
```
JingsongLi: This new path is not reachable through the public table commit API yet. `WriteBuilder.new_commit()` returns `BatchTableCommit` / `StreamTableCommit`, and both `commit()` implementations still delegate to `FileStoreCommit.commit(...)`; with this guard, any caller that receives a `CommitMessage` carrying `compact_increment` through the normal writer/driver flow will just get this `ValueError`. Please expose a matching `commit_compact(...)` on `BatchTableCommit` and `StreamTableCommit` (or make the public `commit()` dispatch compact-only messages to `FileStoreCommit.commit_compact`) so the new compact protocol can actually be committed without reaching into `table_commit.file_store_commit` internals.

paimon-python/pypaimon/write/file_store_commit.py
```diff
@@ -124,33 +124,62 @@ def __init__(self, snapshot_commit: SnapshotCommit, table, commit_user: str,
...
+        # FileStoreCommitImpl.commit ordering.
+        self._commit_data(commit_messages, commit_identifier,
+                          data_index_adds, data_index_deletes)
+        self._commit_compact(commit_messages, commit_identifier,
+                             compact_index_adds, compact_index_deletes)
+
+    def _commit_data(self, commit_messages: List[CommitMessage], commit_identifier: int,
+                     index_adds: List, index_deletes: List) -> None:
+        # Extract the minimum check_from_snapshot from commit messages
+        valid_snapshots = [msg.check_from_snapshot for msg in commit_messages
```
JingsongLi, line 176: `check_from_snapshot` is collected from every message, including compaction-only messages, and the value is stored on the reusable `ConflictDetection` instance. A compact message can already round-trip this field through `CommitMessageSerializer`, but compaction is supposed to skip row-id conflict detection. After such a compact-only commit returns at the data side, `_row_id_check_from_snapshot` remains set; the next normal `StreamTableCommit` using the same `FileStoreCommit` will run `detect_conflicts=True`/`allow_rollback=True` even if its own data messages have `check_from_snapshot=-1`. Please filter this to messages that actually contribute data-side entries, or reset the row-id check state around each commit phase so compact metadata cannot leak into later APPEND/OVERWRITE commits.

paimon-python/pypaimon/write/file_store_commit.py
TheR1sing3un: fixed

paimon-python/pypaimon/write/file_store_commit.py
TheR1sing3un: fixed

JingsongLi (commented):
Thank you for your continuous updates, but I still feel concerned about this ability because Compaction is a very inefficient task in Python. I would prefer Compaction to be completed by the Rust layer.

TheR1sing3un:
> Thank you for your continuous updates, but I still feel concerned about this ability because Compaction is a very inefficient task in Python. I would prefer Compaction to be completed by the Rust layer.

Agreed！ Lets close it，thanks for your kindly reviews

PR #7874, opened 2026-05-16
[core] Implement withFilter for TagsTable and FileKeyRangesTable

heye1005:
### Purpose

Implement `withFilter` for `TagsRead` / `FileKeyRangesRead` (was `TODO`, predicates were silently ignored).

`FileKeyRangesRead` excludes `partition` / `bucket` / `level` from the post-filter — already pushed down by scan, and re-evaluating them on the cast-to-string row breaks range predicates. Same pattern as `FilesTable` (#7791).

### Tests

- `TagsTableTest`: `=` / `IN` / `<>` on `tag_name`, filter on `snapshot_id`.
- `FileKeyRangesTableTest`: `record_count` / `schema_id`, combined predicate, `partition < '{11}'` regression.


JingsongLi:
+1

PR #7875, opened 2026-05-16
[core] Support IsNaN predicate pushdown with Parquet

ArnavBalyan:
### Purpose
 - Paimon has no IsNaN predicate today, add support for the same with Paimon evaluation and parquet pushdown
 - Add IsNaN as a new LeafUnaryFunction, which can evaluate rows and pushdown to Parquet for double and float
 - ParquetFilters pushes down to Parquet via new UserDefinedPredicate.

### Tests
 - UT


ArnavBalyan:
cc @JingsongLi thanks!

JingsongLi (approved):
+1

Issue #7876, opened 2026-05-17
[Feature] Support Parquet map shredding for hot key columns


Aitozi:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

## Motivation

Many tables store semi-structured attributes as `MAP<STRING, STRING>`, for example HTTP headers,
request parameters, or extensible entity properties. This model is flexible, but it can be
inefficient when a small set of keys is present in most rows and is frequently used by queries:

- common keys such as `content-type`, `user-agent`, `locale`, or `campaign` are repeatedly stored in
  the map branch;
- filters and projections frequently access a small set of these keys;
- a standard Parquet map stores keys and values in the repeated `key_value` branch, so readers must
  scan the map payload even when a query only needs one common key.

This proposal adds a Paimon-aware Parquet encoding for selected map columns. The logical table
schema remains `MAP<STRING, STRING>`, while the physical Parquet file stores configured hot keys as
independent sidecar columns.

## Goals

- Keep the Paimon logical type unchanged. Users still read and write a normal map column.
- Avoid introducing a custom Parquet logical type.
- Reduce storage by removing promoted hot keys from the residual map branch and improve the compression ratio for independent columns as well..
- Enable column pruning and future predicate pushdown for expressions such as
  `headers['content-type']`.
- Preserve compatibility for files written without map shredding.

## Non-Goals

- This proposal does not define a new Parquet standard logical type.
- The first implementation only supports top-level `MAP<STRING, STRING>` columns.
- Dynamic key discovery is left for a follow-up. The initial version uses static keys configured in
  table options.

## Physical Layout

For a logical column:

```text
headers MAP<STRING, STRING>
```

Paimon may write the following Parquet physical layout:

```text
headers                         // standard Parquet MAP, containing only residual keys
__paimon_map_shred_headers_0    // value for configured hot key 0
__paimon_map_shred_headers_1    // value for configured hot key 1
...
```

For example, if the configured keys are `content-type,user-agent`, a row:

```text
headers = {
  "content-type": "application/json",
  "user-agent": "mobile-app",
  "x-request-id": "req-001"
}
```

is written as:

```text
headers = {
  "x-request-id": "req-001"
}
__paimon_map_shred_headers_0 = "application/json"
__paimon_map_shred_headers_1 = "mobile-app"
```

When Paimon reads the file, it reconstructs the logical map by merging the residual map with the
sidecar values.

## Metadata

[...]

PR #7877, opened 2026-05-17
[parquet] Add map shredding for hot keys

Aitozi:
### Purpose

Add Parquet map shredding support for `MAP<STRING, T>` columns.

This allows selected map columns to extract hot keys into independent physical Parquet columns while preserving the original logical map schema for readers. The feature is controlled by `map.shredding.*` options, aligned with the existing `variant.shredding.*` naming style. It also adds a focused round-trip test and a storage benchmark to validate the storage benefit.

### Tests

- `mvn -pl paimon-api,paimon-format -Pfast-build -DskipTests compile`
- `mvn -pl paimon-format -am -Pfast-build -DfailIfNoTests=false -Dtest=ParquetFormatReadWriteTest#testMapShreddingRoundTrip,MapShreddingStorageBenchmark test`
- `git diff --check`

### Physical Layout

This change does not introduce a new Parquet logical type and does not modify the standard Parquet `MAP` encoding. A shredded map is still written with the regular Parquet map group as the residual map. Hot keys are promoted into additional sibling sidecar columns in the parent Parquet group.

For example, a logical field:

```sql
headers MAP<STRING, STRING>
```

is normally written as:

```text
message paimon_schema {
  optional group headers (MAP) {
    repeated group key_value {
      required binary key (STRING);
      optional binary value (STRING);
    }
  }
}
```

With map shredding enabled, if `user-agent` and `host` are selected as hot keys, the physical Parquet schema becomes:

```text
message paimon_schema {
  optional group headers (MAP) {
    repeated group key_value {
      required binary key (STRING);
      optional binary value (STRING);
    }
  }

  optional binary dynamic_column_headers_value_0 (STRING);
  optional binary dynamic_column_headers_value_1 (STRING);
}
```

The footer metadata records the mapping from sidecar columns to map keys:

```text
parquet.meta.dynamic.column.map.keys.of.headers = user-agent,host
```

During writing, entries for promoted hot keys are omitted from the residual map when their values are non-null, and their values are written into the corresponding sidecar columns. During reading, Paimon reads both the residual map and the sidecar columns, then reconstructs the original logical `MAP<STRING, T>` value.

For nested maps, the same rule applies within the containing row group. For example, for `payload.headers`, sidecar columns are added as siblings of the `headers` map inside the `payload` group, and the footer metadata uses the full logical path:

```text
parquet.meta.dynamic.column.map.keys.of.payload.headers = user-agent,host
```

#7876

### DIFF
--- a/docs/layouts/shortcodes/generated/core_configuration.html
+++ b/docs/layouts/shortcodes/generated/core_configuration.html
@@ -868,6 +868,30 @@ Mainly to resolve data skew on primary keys. We recommend starting with 64 mb wh
             <td>Level threshold of lookup to generate remote lookup files. Level files below this threshold will not generate remote lookup files.</td>
         </tr>
         <tr>
+            <td><h5>map.shredding.columns</h5></td>
+            <td style="word-wrap: break-word;">(none)</td>
+            <td>String</td>
+            <td>Comma-separated MAP&lt;STRING, T&gt; column paths to write with map shredding.</td>
+        </tr>
+        <tr>
+            <td><h5>map.shredding.maxInferBufferMemory</h5></td>
+            <td style="word-wrap: break-word;">32 mb</td>
+            <td>MemorySize</td>
+            <td>Maximum memory used to buffer rows while inferring map shredding keys.</td>
+        </tr>
+        <tr>
+            <td><h5>map.shredding.maxInferBufferRow</h5></td>
+            <td style="word-wrap: break-word;">10000</td>
+            <td>Integer</td>
+            <td>Maximum number of rows to buffer for map shredding key inference.</td>
+        </tr>
+        <tr>
+            <td><h5>map.shredding.maxKeys</h5></td>
+            <td style="word-wrap: break-word;">64</td>
+            <td>Integer</td>
+            <td>Maximum number of hot keys extracted for each map.</td>
+        </tr>
+        <tr>
             <td><h5>manifest.compression</h5></td>
             <td style="word-wrap: break-word;">"zstd"</td>
             <td>String</td>
--- a/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
@@ -436,6 +436,33 @@ public class CoreOptions implements Serializable {
                     .defaultValue(4096)
                     .withDescription("Maximum number of rows to buffer for schema inference.");
 
+    public static final ConfigOption<String> MAP_SHREDDING_COLUMNS =
+            key("map.shredding.columns")
+                    .stringType()
+                    .noDefaultValue()
+                    .withDescription(
+                            "Comma-separated MAP<STRING, T> column paths to write with map shredding.");
+
+    public static final ConfigOption<Integer> MAP_SHREDDING_MAX_KEYS =
+            key("map.shredding.maxKeys")
+                    .intType()
+                    .defaultValue(64)
+                    .withDescription("Maximum number of hot keys extracted for each map.");
+
+    public static final ConfigOption<MemorySize> MAP_SHREDDING_MAX_INFER_BUFFER_MEMORY =
+            key("map.shredding.maxInferBufferMemory")
+                    .memoryType()
+                    .defaultValue(MemorySize.ofMebiBytes(32))
+                    .withDescription(
+                            "Maximum memory used to buffer rows while inferring map shredding keys.");
+
+    public static final ConfigOption<Integer> MAP_SHREDDING_MAX_INFER_BUFFER_ROW =
+            key("map.shredding.maxInferBufferRow")
+                    .intType()
+                    .defaultValue(10000)
+                    .withDescription(
+                            "Maximum number of rows to buffer for map shredding key inference.");
+
     public static final ConfigOption<String> MANIFEST_FORMAT =
             key("manifest.format")
                     .stringType()
--- a/paimon-format/src/main/java/org/apache/paimon/format/parquet/MapShreddingKeyExtractor.java
+++ b/paimon-format/src/main/java/org/apache/paimon/format/parquet/MapShreddingKeyExtractor.java
@@ -1,0 +1,164 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.paimon.format.parquet;
+
+import org.apache.paimon.CoreOptions;
+import org.apache.paimon.data.BinaryString;
+import org.apache.paimon.data.DataGetters;
+import org.apache.paimon.data.InternalArray;
+import org.apache.paimon.data.InternalMap;
+import org.apache.paimon.data.InternalRow;
+import org.apache.paimon.data.serializer.InternalRowSerializer;
+import org.apache.paimon.options.Options;
+import org.apache.paimon.types.DataType;
+import org.apache.paimon.types.InternalRowToSizeVisitor;
+import org.apache.paimon.types.RowType;
+
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.BiFunction;
+import java.util.stream.Collectors;
+
+/** Extract file-level hot keys for map shredding columns. */
+public class MapShreddingKeyExtractor {
+
+    private final List<MapShreddingUtils.ResolvedMapShreddingColumn> columns;
+    private final InternalRowSerializer serializer;
+    private final int maxKeys;
+    private final long memoryLimit;
+    private final int rowLimit;
+    private final List<InternalRow> bufferedRows;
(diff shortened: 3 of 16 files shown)


Aitozi:
Benchmark command:

```bash
mvn -s ~/.m2/apache-community.xml -pl paimon-format -am -Pfast-build \
  -DfailIfNoTests=false -Dtest=MapShreddingStorageBenchmark test
```

Benchmark file: [MapShreddingStorageBenchmark.java]

**Common Setup**
- Schema: `id INT, headers MAP<STRING, STRING>`
- Rows: `100,000`
- Hot keys: `32`
- Value length: `16`
- Compression: `snappy`
- Compared layouts:
  - regular: normal Parquet map encoding
  - mapShredding: promotes 32 hot keys from `headers` into sidecar columns
- Map shredding options:
  - `map.shredding.columns=headers`
  - `map.shredding.maxKeys=32`
  - `map.shredding.maxInferBufferRow=10000`
  - `map.shredding.maxInferBufferMemory=64 mb`

**Results**

| Scenario | Regular | Map Shredding | Saved | Saving |
|---|---:|---:|---:|---:|
| Columnar value storage | 708,012 bytes | 431,637 bytes | 276,375 bytes | 39.04% |
| Long hot key storage | 40,845,943 bytes | 16,365,106 bytes | 24,480,837 bytes | 59.93% |

**Scenario Details**
- **Columnar value storage**: key names are short, values follow a repeated pattern with `valueRunLength=128` and `valueCardinality=4`, dictionary encoding enabled. This measures whether promoted hot-key values benefit from columnar and dictionary encoding.
- **Long hot key storage**: hot key names include `128` bytes of padding, dictionary encoding disabled. This measures the benefit of avoiding repeated long map-key strings in every row.

Conclusion: in this synthetic storage benchmark, map shredding reduces file size in both cases. The biggest gain appears when hot map keys are long and repeated across many rows, saving about `59.93%`.

JingsongLi:
This looks very suitable to be solved using Variant, why not?

Aitozi:
> This looks very suitable to be solved using Variant, why not?

Hi @JingsongLi  here are two reason we considering to introduce the shredding to the map

1. The map type is widely used in the internal usage, the migration to variant involves a large number of downstream function changes.
2. Regarding storage savings, the variant type does not offer a significant advantage. Based on the benchmark results, the actual reduction in storage usage with variant is very minimal compared to the dynamic column shredding of the map.

JingsongLi:
> > This looks very suitable to be solved using Variant, why not?
> 
> Hi @JingsongLi here are two reason we considering to introduce the shredding to the map
> 
> 1. The map type is widely used in the internal usage, the migration to variant involves a large number of downstream function changes.
> 2. Regarding storage savings, the variant type does not offer a significant advantage. Based on the benchmark results, the actual reduction in storage usage with variant is very minimal compared to the dynamic column shredding of the map.

Can you demonstrate some benchmarks? As you said, the difference in storage between map and variant?

JingsongLi (commented):
## Review: [parquet] Add map shredding for hot keys

Overall this is a well-structured feature that extends the Parquet format layer to promote frequently-occurring map keys into independent columnar sidecar columns. The physical layout is clean, the metadata round-trip through footer key-value pairs is self-contained, and the tests cover multiple value types (STRING, INT, VARIANT) plus nested maps. A few observations follow.

---

### Correctness Concerns

**1. Comma in hot key names breaks metadata parsing**

Hot keys are serialized into footer metadata as a comma-separated string (`String.join(",", keys)` in `toFooterMetadata`), and deserialized using the same `parseColumnPaths` that splits on commas. If a map key legitimately contains a comma (e.g. `"Content-Type, charset"`), it will be incorrectly split during reads. Consider using a delimiter that is less likely to appear in map keys or adding proper escaping (e.g. backslash-escape or JSON array encoding).

**2. `MapSidecarWriter.valueIndex` mutable state coupling**

In `ParquetRowDataWriter`, `MapSidecarWriter.findValueIndex()` stores the found index in a mutable field `valueIndex`, which is later consumed by `write()`. The correctness depends on `RowFieldWriter.shouldWrite()` always being called immediately before `write()` for the same row. If this invariant is ever violated (e.g. a future refactor), the wrong value or a stale index would be written silently. A safer pattern would be to pass the found index explicitly, or recompute it in `write()`.

**3. `InternalRowToSizeVisitor` allocated per-row in the extractor**

In `MapShreddingKeyExtractor.add()`:
```java
BiFunction<DataGetters, Integer, Integer> valueSizer =
        column.valueType().accept(new InternalRowToSizeVisitor());
```
This creates a new visitor object on every call to `add()` for every column. Since the visitor is stateless and depends only on `column.valueType()` (which is fixed), it should be pre-computed once per column during construction and cached in `ResolvedMapShreddingColumn` or in the extractor itself. For 10,000 buffered rows x N columns, this is significant unnecessary allocation.

---

### Design Observations

**4. Hot key selection metric: total byte size vs. frequency**

The extractor ranks keys by total accumulated value size (`sizes.merge(key.toString(), valueSize, Long::sum)`). This means a rare key with a single 10 KB value can outrank a key that appears in 99% of rows but has small values. The storage benefit of shredding comes primarily from eliminating repeated key strings and improving columnar encoding. A frequency-based or hybrid metric (frequency * average_value_size) might give better compression in practice. Worth documenting the trade-off in the config description at minimum.

**5. Sidecar field ID collision risk**

The sidecar column field IDs are computed as:
```java
SpecialFields.getMapValueFieldId(field.id(), 1) + 1024 + i
```
The magic `1024` offset is not validated against the existing field ID space. For schemas with many fields or deeply nested types that already use high IDs, this could silently collide. Consider deriving IDs from a dedicated namespace or adding a collision check during schema conversion.

**6. `reachTargetSize` during the buffering phase**

In `MapShreddingFormatWriter.reachTargetSize()`, when the delegate is null (still buffering), it returns `stream.getPos() >= targetSize`. But during buffering no data has been written to the stream yet (position is 0), so this always returns false. This is likely fine in practice since the buffer phase is bounded by `maxInferBufferRow`/`maxInferBufferMemory`, but the semantics are misleading. A comment explaining this would help future readers.

**7. Schema cache bypass for shredded files**

When `dynamicMapKeys` is non-empty, `getOrCreateRequestedSchema` skips the `requestedSchemaCache` and recomputes the schema every time. Since each file can have different hot keys, this is correct, but it could become a performance concern if many small files are read. Consider keying the cache on `(fileSchema, dynamicMapKeys)` if profiling shows this matters.

---

### Minor / Style

- `MapShreddingStorageBenchmark` duplicates the option key strings as constants (`MAP_SHREDDING_COLUMNS`, etc.) instead of referencing `CoreOptions.MAP_SHREDDING_COLUMNS.key()`. Using the canonical keys avoids silent drift.
- The `hasResidualEntry` method in `MapWriter` iterates all entries to check whether any should not be skipped. In the common case where the map is mostly composed of hot keys (which are skipped), this results in O(n) per row to determine if the repeated group should be opened at all. This is functionally correct but could be noted as a potential optimization point for large maps.
- The option naming uses camelCase (`maxInferBufferRow`, `maxInferBufferMemory`) while the existing variant shredding options and most other Paimon options use dot-separated lowercase (`max-infer-buffer-row`). Aligning with the project convention would be more consistent.

---

Good work overall. The feature is well-isolated, the test coverage is thorough (especially testing Variant values inside shredded maps), and the benchmark demonstrates clear storage savings.

lxy-9602:
Hi @Aitozi, thanks for the great work on this PR! The map shredding approach is clean and well-tested.

I've been working on a related problem and recently drafted a PIP for a storage optimization on `MAP<STRING, T>` as well — would love to get your thoughts.

**The PIP**: [Columnar-Extend Storage Optimization for MAP Type](https://cwiki.apache.org/confluence/display/PAIMON/PIP-43%3A+Columnar-Extend+Storage+Optimization+for+MAP+Type+in+Paimon)

Both approaches aim to give MAP keys true columnar treatment (pushdown, pruning, statistics). The main difference is **whether physical columns are dedicated per-key or reused across keys**:

| | This PR (map shredding) | PIP (columnar-extend) |
|---|---|---|
| Column assignment | One sidecar column per hot key | Fixed `K` reusable columns shared across all keys |
| Key-to-column mapping | Static (inferred once at write time) | Dynamic per-row (`__field_mapping`) |
| Best for | Stable hot keys, moderate key count | High key cardinality with locally repetitive patterns |

Interestingly, the shredding mode can be seen as a **special case** of the extend mode: if we disable `__field_mapping` (i.e. each column is permanently bound to one key, no cross-row reuse), the extend layout degrades exactly into a per-key shredding layout. This means both modes could potentially live under a **single physical framework** — the Struct-based layout from the PIP — with a configuration switch controlling whether columns are reusable or dedicated:

- **`extend` mode** (with `__field_mapping`): columns are reused across keys, best for high-cardinality dynamic scenarios.
- **`shredding` mode** (without `__field_mapping`): each column is bound to a fixed key, essentially what this PR does, but wrapped in a Struct per MAP column.

The Struct wrapping also naturally addresses the **isolation concern**: in the current PR, sidecar columns are placed as flat siblings in the parent Parquet group, so multiple MAP columns' sidecars share the same namespace (with a `+1024` field ID offset that could collide in large schemas, as noted in the review). Wrapping each MAP's physical columns inside a Struct gives clean namespace isolation without changing the logical type.

Would you mind taking a look at the PIP doc? I'd love to hear whether you think unifying these two approaches under one framework makes sense, or if there are practical reasons to keep them separate.

Aitozi:
Hi @lxy-9602 thank you for your proposal, I will take a look

Aitozi:
Hi @lxy-9602, when you start on your PIP, i would like help to migrate the current shredding implementation to the new framework.

lxy-9602:
@Aitozi Hi, sounds good, thank you. We’re currently working on a basic read/write implementation in paimon-cpp. Once that is in place, we’ll start migrating it into the Java side.

You can trace the related C++ work from this issue: https://github.com/alibaba/paimon-cpp/issues/342

Feel free to take a look when you have time, and let me know if you notice any issues or have suggestions.

JingsongLi:
Let's follow PIP's plan, @Aitozi If you have advance plans for the Java side, you can also start implementing the Java side directly, following the general framework of PIP.

lxy-9602:
@Aitozi @JingsongLi Around next week, we plan to sync the PIP framework, along with some utility helpers and general parameter settings to the Java side. I’ll let everyone know once it’s ready, and I’d appreciate it if you could help review it then. Thanks :)

PR #7878, opened 2026-05-17
[core] Fix main table's CatalogEnvironment used when switching branch on fallback table

XiaoHongbo-Hope:
### Purpose

  When a table has `scan.fallback-branch` or `scan.primary-branch` set, `switchToBranch()` passed the main
  table's `CatalogEnvironment` to the branch table. This caused REST snapshot loading to query the main
  table instead of the branch, returning the wrong snapshot ID.

  Fix: copy the `CatalogEnvironment` with a branch-aware `Identifier`.

  ### Tests

  `FallbackReadFileStoreTableTest.testSwitchToBranch`


JingsongLi (approved):
+1

PR #7879, opened 2026-05-17
[core] Track NaN counts in column statistics

ArnavBalyan:
### Purpose
 - Spark, Flink and Iceberg all support IsNaN as a first class predicate and use NaN counts in file/partition pruning.
 - Paimon's SimpleColStats only tracks min, max and null count today, there is no signal at the manifest layer to skip files.
 - Add a nanCount field to SimpleColStats and update the collectors to count the nans, and further be used for engine level predicate pushdown.

### Tests
 - UT


ArnavBalyan:
cc @JingsongLi could you PTAL thanks! :)

JingsongLi (changes requested):
This has a significant impact. I want to determine what changes have been made to the underlying storage, is it necessary?

ArnavBalyan:
Will do small poc on the improvement of isnan pushdown and perf for stat collection, will revisit on mailing list for larger consensus

Issue #7863, opened 2026-05-15
[Feature] Support branch merge for append-only tables


JunRuiLee:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

## Motivation

  Paimon branch management supports create, delete, rename, and fast-forward, but cannot incrementally merge one branch into another. Users need a non-destructive way to bring new data from a source branch into a
  target branch while keeping both branches intact.

  ## Solution

  Add `merge_branch` for append-only tables. It merges data files that exist only in the source branch into the target branch and skips files already present in the target.

  Correctness depends on source and target branches sharing a pure-append history. If compaction or `INSERT OVERWRITE` is allowed, old files may be rewritten or removed from the target branch, and file-level diff can no longer reliably tell whether a source file represents new data or data that has already been rewritten in the target. To make this invariant explicit, branch merge is guarded by the immutable table option `'branch-merge.enabled' = 'true'`. Tables created with this option reject compaction and `INSERT OVERWRITE` from the beginning, and deletion vectors are not supported.

  ## PR Plan

  - PR1: Core APIs, merge implementation, REST endpoint, and unit tests.
  - PR2: Flink/Spark procedures, Flink Action, integration tests, and docs.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #7882, opened 2026-05-18
[core][rest] Add branch merge support for append-only tables

JunRuiLee:
### Purpose
This PR is the first part of #7863, focusing on the core APIs, implementation, and REST support for branch merge.

  It adds `mergeBranch`, which incrementally adds data files that exist only in the source branch to the target branch, without replacing either branch. Correctness is guarded by the immutable table option `'branch-merge.enabled' = 'true'`, which enforces the pure-append invariant required by file-level branch merge: compaction and `INSERT OVERWRITE` are rejected, and deletion vectors are not supported.

### Tests
Added core tests:

  - `AppendOnlySimpleTableTest#testMergeBranch`
  - `AppendOnlySimpleTableTest#testMergeBranchMultipleTimes`
  - `AppendOnlySimpleTableTest#testMergeBranchFailsOnStaleDuplicateCommit`
  - `AppendOnlySimpleTableTest#testMergeBranchBidirectional`
  - `AppendOnlySimpleTableTest#testMergeBranchEmptyDiff`
  - `AppendOnlySimpleTableTest#testMergeBranchSchemaConflict`
  - `AppendOnlySimpleTableTest#testMergeBranchSchemaHistoryConflict`
  - `AppendOnlySimpleTableTest#testMergeBranchRowTrackingTable`
  - `AppendOnlySimpleTableTest#testMergeBranchRowTrackingMultipleTimes`
  - `AppendOnlySimpleTableTest#testMergeBranchRowTrackingAfterTargetWrites`
  - `AppendOnlySimpleTableTest#testMergeBranchRowTrackingBetweenNonMainBranches`
  - `AppendOnlySimpleTableTest#testMergeBranchRowTrackingMismatch`
  - `AppendOnlySimpleTableTest#testMergeBranchRowTrackingStaleMerge`
  - `AppendOnlySimpleTableTest#testMergeBranchSameBranch`
  - `AppendOnlySimpleTableTest#testMergeBranchSamePartition`
  - `AppendOnlySimpleTableTest#testMergeBranchNonExistentBranch`
  - `AppendOnlySimpleTableTest#testMergeBranchMultiBucket`
  - `AppendOnlySimpleTableTest#testMergeBranchNonExistentTargetBranch`
  - `AppendOnlySimpleTableTest#testMergeBranchBetweenNonMainBranches`
  - `AppendOnlySimpleTableTest#testMergeBranchWithoutBranchMergeEnabled`
  - `AppendOnlySimpleTableTest#testMergeBranchAfterSnapshotExpiration`
  - `AppendOnlySimpleTableTest#testBranchMergeEnabledRejectsOverwrite`
  - `PrimaryKeySimpleTableTest#testMergeBranchPrimaryKeyTable`

  Added REST tests:

  - `RESTCatalogTest#testMergeBranch`
  - branch-not-exist coverage in `RESTCatalogTest#testBranches`


paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -2215,6 +2215,17 @@ public InlineElement getDescription() {
                     .defaultValue(false)
                     .withDescription("Whether enable data evolution for row tracking table.");
 
+    @Immutable
+    public static final ConfigOption<Boolean> BRANCH_MERGE_ENABLED =
+            key("branch-merge.enabled")
```
JingsongLi: It looks like we don't need this option, just throw exception in branch merging is oK.

paimon-core/src/main/java/org/apache/paimon/io/PojoDataFileMeta.java
```diff
@@ -379,6 +379,31 @@ public PojoDataFileMeta assignFirstRowId(long firstRowId) {
                 writeCols);
     }
 
+    @Override
+    public PojoDataFileMeta clearFirstRowId() {
```
JingsongLi: `newFirstRowId(@Nullable Long newFirstRowId)`

paimon-api/src/main/java/org/apache/paimon/rest/RESTApi.java
```diff
@@ -1006,6 +1007,25 @@ public void fastForward(Identifier identifier, String branch) {
...
+     *
+     * @param identifier database name and table name.
+     * @param sourceBranch source branch name
+     * @param targetBranch target branch name
+     * @throws NoSuchResourceException Exception thrown on HTTP 404 means the branch or table not
+     *     exists
+     * @throws ForbiddenException Exception thrown on HTTP 403 means don't have the permission for
+     *     this table
+     */
+    public void mergeBranch(Identifier identifier, String sourceBranch, String targetBranch) {
```
JingsongLi: Why do we need it as a REST API? Does it feel like just creating a new commit?

JingsongLi (changes requested):
Thanks @JunRuiLee , left comments.

paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
JunRuiLee: Thanks for the review. I added this option mainly to keep a clear invariant for branch merge: the table history must be pure-append.

My concern is that checking this only when mergeBranch is called may not be reliable, because old snapshots can expire. In that case, we may no longer be able to tell whether compaction or INSERT OVERWRITE happened before, and file-level merge could become unsafe.

So the option is intended to make this an explicit opt-in behavior and keep the table merge-safe from the beginning by rejecting compaction / INSERT OVERWRITE.

Another possible approach could be to persist some state indicating whether unsafe operations have ever happened, but that seems a bit heavier to me. 

Please let me know what you think.

paimon-api/src/main/java/org/apache/paimon/rest/RESTApi.java
JunRuiLee: Thanks for the question. Yes, it creates a new commit on the target branch.

I added it for REST catalog because branch merge needs to read and validate metadata from both branches before committing. It is also aligned with fastForward, which is already a REST branch operation.

paimon-core/src/main/java/org/apache/paimon/io/PojoDataFileMeta.java
JunRuiLee: done

paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
JingsongLi: I think enabling it directly is OK, after all, this is a brand new operation that users will not be involved in by default.

paimon-api/src/main/java/org/apache/paimon/rest/RESTApi.java
JingsongLi: Unlike fastforward, merge has a higher overhead and involves many file operations, which is not suitable for being placed on a REST server to complete

JunRuiLee:
Thanks @JingsongLi, I addressed both comments.

  1. Removed branch-merge.enabled. Branch merge now validates at merge time. To avoid introducing branch lineage / fork metadata in this PR, it conservatively requires both source and target branches to retain complete append-only snapshot history from the first snapshot. If snapshots have expired or non-append commits are found, mergeBranch fails.

  2. Removed branch merge from the REST API path.

paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
```diff
@@ -901,6 +901,19 @@ void renameBranch(Identifier identifier, String fromBranch, String toBranch)
...
+     * Merge source branch into target branch.
+     *
+     * @param identifier path of the table, cannot be system or branch name.
+     * @param sourceBranch the source branch name
+     * @param targetBranch the target branch name
+     * @throws BranchNotExistException if the source or target branch doesn't exist
+     * @throws UnsupportedOperationException if the catalog does not {@link
+     *     #supportsVersionManagement()}
+     */
+    void mergeBranch(Identifier identifier, String sourceBranch, String targetBranch)
```
JingsongLi: Do not need to add this method to catalog.

paimon-core/src/main/java/org/apache/paimon/utils/BranchMergeHandler.java
```diff
@@ -0,0 +1,35 @@
...
+package org.apache.paimon.utils;
+
+import org.apache.paimon.manifest.FileEntry;
+import org.apache.paimon.manifest.ManifestEntry;
+
+import java.util.List;
+import java.util.Map;
+
+/** Handler for branch merge data operations (manifest reading, committing). */
+public interface BranchMergeHandler {
```
JingsongLi: Do not need to add this interface. Now there is only one implementation.

JunRuiLee: fixed

paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java
JunRuiLee: fixed

JunRuiLee:
Thanks @JingsongLi for the review. I’ve addressed all comments. PTAL.

JingsongLi (approved):
+1

PR #7886, opened 2026-05-18
[Core] Support to order data by columns in append only writer

### LINKED ISSUE #7885: [Feature] Support to order data by keys in append only writer
### Search before asking

### Motivation

realted to #7253 , order data in append only writer

### Solution

_No response_

### Anything else?

_No response_

FangYongs:
### Purpose
Order data by specific columns in single file which is written by append only writer

### Tests
AppendOnlyWriterTest#testSortedBufferedSinkWriter

Close #7885

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/append/AppendOnlyWriter.java
+++ b/paimon-core/src/main/java/org/apache/paimon/append/AppendOnlyWriter.java
@@ -18,6 +18,7 @@
 
 package org.apache.paimon.append;
 
+import org.apache.paimon.CoreOptions;
 import org.apache.paimon.annotation.VisibleForTesting;
 import org.apache.paimon.compact.CompactDeletionFile;
 import org.apache.paimon.compact.CompactManager;
@@ -51,6 +52,7 @@ import org.apache.paimon.utils.RecordWriter;
 import org.apache.paimon.utils.SinkWriter;
 import org.apache.paimon.utils.SinkWriter.BufferedSinkWriter;
 import org.apache.paimon.utils.SinkWriter.DirectSinkWriter;
+import org.apache.paimon.utils.SinkWriter.SortedBufferedSinkWriter;
 import org.apache.paimon.utils.StatsCollectorFactories;
 
 import javax.annotation.Nullable;
@@ -97,6 +99,8 @@ public class AppendOnlyWriter implements BatchRecordWriter, MemoryOwner {
     @Nullable private final IOManager ioManager;
     private final FileIndexOptions fileIndexOptions;
     private final MemorySize maxDiskSize;
+    private final CoreOptions coreOptions;
+    private final boolean sortEnabled;
 
     @Nullable private CompactDeletionFile compactDeletionFile;
     private SinkWriter<InternalRow> sinkWriter;
@@ -129,7 +133,8 @@ public class AppendOnlyWriter implements BatchRecordWriter, MemoryOwner {
             boolean asyncFileWrite,
             boolean statsDenseStore,
             boolean dataEvolutionEnabled,
-            @Nullable BlobFileContext blobContext) {
+            @Nullable BlobFileContext blobContext,
+            CoreOptions coreOptions) {
         this.fileIO = fileIO;
         this.schemaId = schemaId;
         this.fileFormat = fileFormat;
@@ -159,11 +164,18 @@ public class AppendOnlyWriter implements BatchRecordWriter, MemoryOwner {
         this.statsCollectorFactories = statsCollectorFactories;
         this.maxDiskSize = maxDiskSize;
         this.fileIndexOptions = fileIndexOptions;
+        this.coreOptions = coreOptions;
 
-        this.sinkWriter =
-                useWriteBuffer
-                        ? createBufferedSinkWriter(spillable)
-                        : new DirectSinkWriter<>(this::createRollingRowWriter);
+        // Determine if we need to enable sorting based on clustering configuration
+        List<String> clusteringColumns = coreOptions.clusteringColumns();
+        this.sortEnabled =
+                coreOptions.clusteringIncrementalEnabled()
+                        && coreOptions.clusteringIncrementalOptimizeWrite()
+                        && coreOptions.clusteringIncrementalMode()
+                                == CoreOptions.ClusteringIncrementalMode.LOCAL_SORT
+                        && !clusteringColumns.isEmpty();
+
+        this.sinkWriter = createSinkWriter(useWriteBuffer, spillable);
 
         if (increment != null) {
             newFiles.addAll(increment.newFilesIncrement().newFiles());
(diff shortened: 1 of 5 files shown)


FangYongs:
@Aitozi @shidayang Have a look when you're free

paimon-core/src/main/java/org/apache/paimon/append/AppendOnlyWriter.java
```diff
@@ -159,11 +164,18 @@ public AppendOnlyWriter(
...
-                useWriteBuffer
-                        ? createBufferedSinkWriter(spillable)
-                        : new DirectSinkWriter<>(this::createRollingRowWriter);
+        // Determine if we need to enable sorting based on clustering configuration
+        List<String> clusteringColumns = coreOptions.clusteringColumns();
+        this.sortEnabled =
+                coreOptions.clusteringIncrementalEnabled()
+                        && coreOptions.clusteringIncrementalOptimizeWrite()
+                        && coreOptions.clusteringIncrementalMode()
+                                == CoreOptions.ClusteringIncrementalMode.LOCAL_SORT
```
Aitozi, line 175: The `LOCAL_SORT` is described as Task-Level sorting, but what we have actually implemented is File-Level sorting.

Do we need to introduce a mode similar to "file_local" to represent this specific granularity of File-Level sorting functionality?

```
 /**
         * Sort rows only within each compaction task (no global shuffle). Every output file is
         * internally ordered by the clustering columns, which is sufficient for per-file Parquet
         * lookup optimizations.
         */
        LOCAL_SORT(
                "local-sort",
                "Sort rows only within each compaction task without global shuffle. Every output file is internally ordered.");
```

FangYongs: @JingsongLi What do you think of LOCAL_SORT? In our previous discussion, this was meant for local sorting in a single file. However, judging from the current situation, it is used for data sorting at the task level.

paimon-core/src/main/java/org/apache/paimon/append/AppendOnlyWriter.java
```diff
@@ -174,16 +186,31 @@ public AppendOnlyWriter(
...
-                this::createRollingRowWriter,
-                t -> t,
-                t -> t,
-                ioManager,
-                writeSchema,
-                spillable,
-                maxDiskSize,
-                spillCompression);
+    private SinkWriter<InternalRow> createSinkWriter(boolean useWriteBuffer, boolean spillable) {
+        if (useWriteBuffer) {
```
shidayang, line 190: When sortEnabled is enabled, should we throw an error if useWriteBuffer is false ? Otherwise, it is hard for users to notice this behavior.

paimon-core/src/main/java/org/apache/paimon/append/AppendOnlyWriter.java
JingsongLi: I feel it's okay. Local sorting can just do its best to sort, whether at the task level or file level. We can add comments to explain.

JingsongLi (commented):
## Review

The overall idea — enabling local sort within each data file for incremental clustering in the append-only writer — is reasonable. However there are several design and correctness concerns.

### Design Issues

**1. `SortedBufferedSinkWriter` flushes ALL buffered rows into a single `RollingFileWriter` call.**

```java
public List<DataFileMeta> flush() throws IOException {
    RollingFileWriter<T, DataFileMeta> writer = writerSupplier.get();
    MutableObjectIterator<BinaryRow> sorted = sortBuffer.sortedIterator();
    while ((reuse = sorted.next(row)) != null) {
        writeToWriter(writer, reuse);
    }
    writer.close();
    flushedFiles.addAll(writer.result());
    sortBuffer.clear();
}
```

The `RollingFileWriter` will roll files based on target file size. But the sorted data is written sequentially — it will be split at arbitrary points based on byte size, NOT on sort-key boundaries. This means after rolling, the value ranges in adjacent files may overlap. The sort is only meaningful within each individual output file, which is the stated goal. So this is correct but worth documenting.

**2. Passing the entire `CoreOptions` to `AppendOnlyWriter` and `SortedBufferedSinkWriter` is heavy-handed.**

The writer now stores `CoreOptions coreOptions` as a field. It only uses it to extract clustering columns, mode, and a few sort config values. Consider passing just the needed parameters (clustering columns, sort config) instead of the entire options bag. This avoids coupling the writer to all of `CoreOptions` and makes testing easier.

**3. The `sortEnabled` condition requires all four checks simultaneously.**

```java
this.sortEnabled =
    coreOptions.clusteringIncrementalEnabled()
        && coreOptions.clusteringIncrementalOptimizeWrite()
        && coreOptions.clusteringIncrementalMode() == LOCAL_SORT
        && !clusteringColumns.isEmpty();
```

This is very specific — users must set exactly the right combination. The option name `clustering.incremental.optimize-write` is a separate boolean on top of `clustering.incremental=true` + `clustering.incremental.mode=local-sort`. That's 3 options to enable a single behavior. Would a simpler option like `write.sort-by-columns` (independent of clustering) be more user-friendly?

**4. `instanceof` chain in `writeBundle` is fragile.**

```java
if (sinkWriter instanceof BufferedSinkWriter) {
    ((BufferedSinkWriter<InternalRow>) sinkWriter).writeBundle(bundle);
} else if (sinkWriter instanceof SortedBufferedSinkWriter) {
    ((SortedBufferedSinkWriter<InternalRow>) sinkWriter).writeBundle(bundle);
} else {
    ((DirectSinkWriter<?>) sinkWriter).writeBundle(bundle);
}
```

Since `SortedBufferedSinkWriter extends BaseBufferedSinkWriter` and `BufferedSinkWriter extends BaseBufferedSinkWriter`, and `writeBundle` is defined on `BaseBufferedSinkWriter`, you could just check `instanceof BaseBufferedSinkWriter`. Or better: make `writeBundle` part of the `SinkWriter<T>` interface with a default implementation, eliminating the instanceof entirely.

### Correctness Issues

**5. `SortedBufferedSinkWriter.write()` doesn't return a meaningful value.**

```java
public boolean write(T data) throws IOException {
    return sortBuffer.write(toRow.apply(data));
}
```

`SortBuffer.write()` returns `false` when the buffer is full and cannot accept more data. When this returns `false`, the caller (`AppendOnlyWriter.write()`) should trigger a flush. But looking at `AppendOnlyWriter.write()`:

```java
public void write(InternalRow rowData) throws Exception {
    boolean success = sinkWriter.write(rowData);
    if (!success) {
        flush(false, false);
        success = sinkWriter.write(rowData);
        if (!success) {
            flush(false, true);
        }
    }
}
```

This flush-and-retry logic works correctly for `BufferedSinkWriter` (which clears the write buffer). For `SortedBufferedSinkWriter`, `flush()` calls `sortBuffer.clear()` — so after flush, the retry write should succeed. Good.

**6. `setMemoryPool` resolves clustering column indices by name — no validation.**

```java
int[] keyFields = clusteringColumns.stream()
    .mapToInt(rowType.getFieldNames()::indexOf).toArray();
```

If a clustering column name doesn't exist in `rowType`, `indexOf` returns `-1`. This will cause `ArrayIndexOutOfBoundsException` later when the comparator tries to access field `-1`. Should validate and throw a clear error if a clustering column is missing from the schema.

**7. The `sortBuffer` is initialized in `setMemoryPool`, but `write()` may be called before `setMemoryPool`.**

If `write()` is called before `setMemoryPool()`, `sortBuffer` is `null` → NPE. The `BufferedSinkWriter` has the same pattern (write buffer created in `setMemoryPool`), so this is consistent but risky. At minimum, add a null check with a clear error.

### Minor

- `BaseBufferedSinkWriter` is a good refactoring to reduce duplication
- Removing two comment lines (`// cleanup code that might throw another exception`, `// reuse writeBuffer`) is fine — they were noise
- The test covers the basic happy path but doesn't test: buffer spill to disk, write more data than memory allows, clustering column not in schema (negative test)

paimon-core/src/main/java/org/apache/paimon/append/AppendOnlyWriter.java
```diff
@@ -159,11 +164,18 @@ public AppendOnlyWriter(
...
+        // Determine if we need to enable sorting based on clustering configuration
+        List<String> clusteringColumns = coreOptions.clusteringColumns();
+        this.sortEnabled =
+                coreOptions.clusteringIncrementalEnabled()
+                        && coreOptions.clusteringIncrementalOptimizeWrite()
+                        && coreOptions.clusteringIncrementalMode()
+                                == CoreOptions.ClusteringIncrementalMode.LOCAL_SORT
+                        && !clusteringColumns.isEmpty();
+
+        this.sinkWriter = createSinkWriter(useWriteBuffer, spillable);
```
JingsongLi, line 178: This makes local sort silently depend on `write-buffer-for-append`, whose default is `false`. Even when all clustering-local-sort options above make `sortEnabled=true`, the default append path passes `useWriteBuffer=false`, so `createSinkWriter` falls through to `DirectSinkWriter` and no sorting happens. The new test sets `write-buffer-for-append=true`, but the incremental clustering configuration/docs do not require users to set that extra option, so the feature will appear enabled while being bypassed in the common path. Could local-sort force the buffered writer here (or fail fast/document and test the required `write-buffer-for-append=true` setting)?

JingsongLi:

Issue #7887, opened 2026-05-18
[Feature] Support compact for chain table


Stefanietry:
### Motivation

Support compact for chain table on spark

### Solution

Implement a new procedure for chain table : ChainMergeProcedure

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #7888, opened 2026-05-18
[core] Support compact for chain table

Stefanietry:
Purpose
Linked issue: https://github.com/apache/paimon/issues/7887

### Purpose
Support compact for chain table

### Tests
org.apache.paimon.spark.SparkChainTableITCase#testChainMerge

### DIFF
--- a/docs/content/primary-key-table/chain-table.md
+++ b/docs/content/primary-key-table/chain-table.md
@@ -87,7 +87,6 @@ Notice that:
 - Chain table is only supported for primary key table, which means you should define `bucket` and `bucket-key` for the table.
 - Chain table should ensure that the schema of each branch is consistent.
 - Only spark support now, flink will be supported later.
-- Chain compact is not supported for now, and it will be supported later.
 - Deletion vector is not supported for chain table.
 
 After creating a chain table, you can read and write data in the following ways.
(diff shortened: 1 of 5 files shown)


JingsongLi (commented):
## Review: [core] Support compact for chain table

### Critical Bug: Resource lifecycle in mapPartitions closure

In `ChainMergeProcedure.java`, the `BatchTableWrite` and `IOManager` are created before the `while (splitIterator.hasNext())` loop, but `write.close()` and `ioManager.close()` are called inside a `finally` block within the loop iteration. If a Spark partition receives more than one split, the second iteration will attempt to use already-closed resources and fail. Additionally, `prepareCommit()` is called per-split rather than once after all data has been written.

**Fix**: move the `try/finally` to wrap the entire while-loop, and call `prepareCommit()` only after all splits are processed.

### Hardcoded version in pom.xml

The new dependency uses `1.5-SNAPSHOT` instead of `${project.version}`. This will break when the project version changes.

### Typo: `validataChainMerge` should be `validateChainMerge`

### Inconsistent Javadoc

The class-level doc shows `CALL sys.compact(...)` but the actual registered procedure name is `chain_merge`.

### Incorrect PR title prefix

Title says `[core]` but all changes are in `paimon-spark`. Should be `[spark]`.

### Test coverage suggestion

The test should also assert that the un-merged partition in the snapshot branch remains unaffected, to guard against accidental overwrites.

juntaozhang:
Hi, it looks like this feature duplicates what was implemented in #7313.

Stefanietry:
> Hi, it looks like this feature duplicates what was implemented in #7313.

OK，if you have already supported this feature, I will turn it off.

Issue #7883, opened 2026-05-18
[Bug] TIMESTAMP(9) write to PostponeBucketTable fails with 'file.format.per.level' = '0:avro'


lvyanquan:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Paimon version

1.4

### Compute Engine

Flink

### Minimal reproduce step

CREATE TABLE T (k INT, v TIMESTAMP(9), PRIMARY KEY (k) NOT ENFORCED)
 WITH ('bucket' = '-2',
'file.format.per.level' = '0:avro',
'postpone.batch-write-fixed-bucket' = 'false');

INSERT INTO T VALUES (1, TIMESTAMP '2025-06-11 16:35:45.123456789'), (2, CAST(NULL AS TIMESTAMP(9)));

### What doesn't meet your expectations?

Exception happened.

### Anything else?

Error message:
```

Caused by: java.lang.IllegalArgumentException: Avro does not support TIMESTAMP type with precision: 9, it only supports precision less than 6.
        at org.apache.paimon.format.avro.AvroSchemaConverter.convertToSchema(AvroSchemaConverter.java:121)
        at org.apache.paimon.format.avro.AvroSchemaConverter.convertToSchema(AvroSchemaConverter.java:190)
        at org.apache.paimon.format.avro.AvroSchemaConverter.convertToSchema(AvroSchemaConverter.java:63)
        at org.apache.paimon.format.avro.AvroFileFormat$RowAvroWriterFactory.lambda$new$433a2ff1$1(AvroFileFormat.java:123)
        at org.apache.paimon.format.avro.AvroWriterFactory.create(AvroWriterFactory.java:46)
        at org.apache.paimon.format.avro.AvroFileFormat$RowAvroWriterFactory.create(AvroFileFormat.java:138)
        at org.apache.paimon.io.SingleFileWriter.<init>(SingleFileWriter.java:87)
        at org.apache.paimon.io.StatsCollectingSingleFileWriter.<init>(StatsCollectingSingleFileWriter.java:52)
        at org.apache.paimon.io.KeyValueDataFileWriter.<init>(KeyValueDataFileWriter.java:91)
        at org.apache.paimon.io.KeyValueDataFileWriterImpl.<init>(KeyValueDataFileWriterImpl.java:51)
        at org.apache.paimon.io.KeyValueFileWriterFactory.createDataFileWriter(KeyValueFileWriterFactory.java:176)
[...]

PR #7892, opened 2026-05-18
[core] Reject per level avro format with high precision timestamps

ArnavBalyan:
### Purpose
 - Avro cannot store timestamps with precision over six.
 - Per level avro setting not checked and was allowing invalid table creation with incompatible precision.
 - Validate the precision upfront and fail with clear message, if configuration cannot be satisfied.
 - Closes user reported bug #7883

### Tests
 - UT


JingsongLi (approved):
+1

PR #7893, opened 2026-05-18
[core] Add tests and implement withFilter for StatisticTable

heye1005:
### Purpose

Remove the two `// TODO` markers on `StatisticTable`

### Tests

Adds `StatisticTableTest`

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/system/StatisticTable.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/system/StatisticTable.java
@@ -47,6 +47,8 @@ import org.apache.paimon.utils.SerializationUtils;
 
 import org.apache.paimon.shade.guava30.com.google.common.collect.Iterators;
 
+import javax.annotation.Nullable;
+
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Iterator;
@@ -119,7 +121,6 @@ public class StatisticTable implements ReadonlyTable {
 
         @Override
         public InnerTableScan withFilter(Predicate predicate) {
-            // TODO
             return this;
         }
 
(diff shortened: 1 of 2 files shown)


heye1005:
## CI Failures

These appear to be unrelated flaky tests, not caused by this PR's changes:
- `MySqlSyncTableActionITCase.testAllTypes` (timeout)
- `KafkaCanalSyncDatabaseActionITCase.testExpressionNow` (timeout)
- Maven Central network timeout
Could you help re-run the failed jobs?

I also want to separate new pr to address the CDC ITCase flakiness.

JingsongLi (commented):
Review comment test

JingsongLi (commented):
## Review: StatisticTable withFilter implementation

Overall this is a clean, well-structured change that resolves the TODO markers and adds solid test coverage. A few observations:

### Correctness

**Filter-before-projection ordering is correct.** The `postFilter` is applied to the full-schema row produced by `toRow()`, and projection via `ProjectedRow` happens afterward. This ensures predicate column indices align with `TABLE_TYPE` rather than the projected `readType`.

**Scan-level `withFilter` is intentionally a no-op** -- since `StatisticTable` always produces at most one split, there is nothing to prune at the scan layer. The actual row-level filtering is delegated to `StatisticRead.withFilter`, which is the right approach for this table.

### Minor suggestion

The `Iterators.filter(rows, postFilter::test)` call: since `Statistics` yields at most **one row**, the practical effect of the filter is a simple keep-or-discard decision. The current implementation is fine and consistent with how other system tables handle filtering.

### Tests

Good coverage: empty statistics, basic read, predicate hit/miss on `snapshot_id`, and range predicate on `merged_record_count`. One optional enhancement for the future: a test combining `withFilter` and `withProjection` together to validate the filter-then-project ordering invariant.

### Summary

The change is correct and well-tested. LGTM.

JingsongLi:
+1

PR #7897, opened 2026-05-19
[spark] Introduce plan-auto-tag-for-read for Spark read

JingsongLi:
### Purpose

Introduce `'scan.plan-auto-tag-for-read.time-retained'` option:

When set, a temporary tag will be auto-created during batch scan planning to protect the read snapshot from expiration. The value specifies the tag's TTL. Should be longer than the longest expected batch read duration.


paimon-core/src/main/java/org/apache/paimon/tag/BatchReadTagCreator.java
```diff
@@ -0,0 +1,101 @@
...
+                    snapshotId,
+                    e);
+            return null;
+        }
+    }
+
+    public void deleteReadTag(String tagName) {
+        try {
+            if (tagManager.tagExists(tagName)) {
+                snapshotManager.fileIO().deleteQuietly(tagManager.tagPath(tagName));
```
XiaoHongbo-Hope, line 85: Is this intentionally deleting only the tag meta file instead of calling table.deleteTag(tagName) to
  keep read cleanup lightweight

JingsongLi: Directly delete the tag metadata file instead of using TagManager.deleteTag(), which would also scan and delete unreferenced data files — too heavyweight for a read-path cleanup. Any orphan data files left behind will be reclaimed by OrphanFilesClean.

Commented in https://github.com/apache/paimon/pull/7897/commits/1d4a46123ea8a4e7835d4a801a90524eb96011ed

XiaoHongbo-Hope: > 

Got it.

XiaoHongbo-Hope:
+1

Zouxxyy (approved):
+1

PR #7900, opened 2026-05-19
[python] Add StartupMode enum and scan.mode option to CoreOptions

JunRuiLee:
### Purpose
Introduce StartupMode enum (default, latest-full, latest, from-timestamp, from-snapshot, etc.) and the scan.mode config option, aligning with Java CoreOptions.SCAN_MODE for explicit scan startup mode control.

### DIFF
--- a/paimon-python/pypaimon/common/options/core_options.py
+++ b/paimon-python/pypaimon/common/options/core_options.py
@@ -67,6 +67,23 @@ class SortOrder(str, Enum):
     DESCENDING = "descending"
 
 
+class StartupMode(str, Enum):
+    """
+    Startup mode for scan operations.
+    """
+    DEFAULT = "default"
+    LATEST_FULL = "latest-full"
+    FULL = "full"
+    LATEST = "latest"
+    COMPACTED_FULL = "compacted-full"
+    FROM_TIMESTAMP = "from-timestamp"
+    FROM_SNAPSHOT = "from-snapshot"
+    FROM_SNAPSHOT_FULL = "from-snapshot-full"
+    FROM_CREATION_TIMESTAMP = "from-creation-timestamp"
+    FROM_FILE_CREATION_TIME = "from-file-creation-time"
+    INCREMENTAL = "incremental"
+
+
 class GlobalIndexColumnUpdateAction(str, Enum):
     THROW_ERROR = "THROW_ERROR"
     DROP_PARTITION_INDEX = "DROP_PARTITION_INDEX"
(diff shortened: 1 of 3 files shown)


JingsongLi (commented):
Good alignment with Java's CoreOptions.SCAN_MODE. A few comments:

1. **`startup_mode()` resolution logic**: The fallback from `DEFAULT` to `LATEST_FULL` when no scan options are present matches Java behavior. However, the `contains_key` checks use raw string literals like `"scan.timestamp-millis"` and `"scan.watermark"` instead of `CoreOptions` constants. This makes it fragile if those option keys are ever refactored. Please use the constant-based approach (e.g., `CoreOptions.SCAN_TIMESTAMP_MILLIS`) if they exist, or define them as constants.

2. **Validation whitelist**: The `_validate_scan_mode()` method in `table_scan.py` is well-structured, but it raises `NotImplementedError` for `COMPACTED_FULL`, `FROM_CREATION_TIMESTAMP`, `FROM_FILE_CREATION_TIME`. These should probably log a warning or raise a more specific exception type that doesn't suggest a code bug (since `NotImplementedError` typically means the method should have been overridden).

3. **Missing test coverage**: I don't see test files in this PR. The validation logic has many branches — at minimum there should be tests for:
   - `scan.mode=from-timestamp` without `scan.timestamp-millis` → error
   - Conflicting options (e.g., `scan.mode=latest` + `scan.snapshot-id`) → error
   - `scan.mode=default` with `scan.timestamp-millis` → resolves to `FROM_TIMESTAMP`

4. **`FULL` deprecation**: The mapping from `FULL` to `LATEST_FULL` is correct but should there be a deprecation warning emitted?

Please add unit tests covering the validation paths before merging.

JunRuiLee:
> Good alignment with Java's CoreOptions.SCAN_MODE. A few comments:
> 
> 1. **`startup_mode()` resolution logic**: The fallback from `DEFAULT` to `LATEST_FULL` when no scan options are present matches Java behavior. However, the `contains_key` checks use raw string literals like `"scan.timestamp-millis"` and `"scan.watermark"` instead of `CoreOptions` constants. This makes it fragile if those option keys are ever refactored. Please use the constant-based approach (e.g., `CoreOptions.SCAN_TIMESTAMP_MILLIS`) if they exist, or define them as constants.
> 2. **Validation whitelist**: The `_validate_scan_mode()` method in `table_scan.py` is well-structured, but it raises `NotImplementedError` for `COMPACTED_FULL`, `FROM_CREATION_TIMESTAMP`, `FROM_FILE_CREATION_TIME`. These should probably log a warning or raise a more specific exception type that doesn't suggest a code bug (since `NotImplementedError` typically means the method should have been overridden).
> 3. **Missing test coverage**: I don't see test files in this PR. The validation logic has many branches — at minimum there should be tests for:
>    
>    * `scan.mode=from-timestamp` without `scan.timestamp-millis` → error
>    * Conflicting options (e.g., `scan.mode=latest` + `scan.snapshot-id`) → error
>    * `scan.mode=default` with `scan.timestamp-millis` → resolves to `FROM_TIMESTAMP`
> 4. **`FULL` deprecation**: The mapping from `FULL` to `LATEST_FULL` is correct but should there be a deprecation warning emitted?
> 
> Please add unit tests covering the validation paths before merging.

Thanks for the detailed review. I addressed the comments in the latest update. PTAL.

JunRuiLee:
Hi @JingsongLi , I've addressed all comments, PTAL. Thanks~

JingsongLi (commented):
LGTM, thanks.

JingsongLi:
Please rebase master.

JunRuiLee:
@JingsongLi Thanks for remind, rebased. PTAL.

PR #7901, opened 2026-05-19
[python] Fix null blob write

XiaoHongbo-Hope:
### Purpose

PR #7847 added null support to `BlobFormatWriter.add_element` / `write_value` and `FormatBlobReader`, but the public `write_arrow` path (`DataBlobWriter` → `BlobWriter` → `BlobFileWriter._to_blob`) still rejects `None` and raises `ValueError`, so writing a batch with a NULL inline blob fails end-to-end:

```
ValueError: Blob field value must be bytes/blob or serialized BlobDescriptor bytes, got <class 'NoneType'>.
```

This PR makes `BlobFileWriter._to_blob` return `None` for `None` input. 

### Tests

- New e2e test `DataBlobWriterTest.test_null_blob`

### DIFF
--- a/paimon-python/pypaimon/tests/blob_table_test.py
+++ b/paimon-python/pypaimon/tests/blob_table_test.py
@@ -1015,13 +1015,6 @@ class DataBlobWriterTest(unittest.TestCase):
             f"✅ End-to-end blob write/read test passed: wrote and read back {len(blob_data)} blob records correctly")  # noqa: E501
 
     def test_blob_write_read_end_to_end_with_null_values(self):
-        """Reproduce: writing NULL into an inline blob column via the public
-        write API (write_arrow) currently raises in BlobFileWriter._to_blob,
-        even though BlobFormatWriter.add_element / FormatBlobReader already
-        handle the NULL marker (-1) correctly. This test asserts the desired
-        behaviour: NULL blob values must round-trip through the standard
-        write -> commit -> read path.
-        """
         from pypaimon import Schema
 
         pa_schema = pa.schema([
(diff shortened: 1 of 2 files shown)


JingsongLi:
+1

JingsongLi (commented):
## Review: [python] Fix null blob write

**Summary**: This PR fixes a gap in the null-blob write path where `BlobFileWriter._to_blob()` raised `ValueError` on `None` input, even though both the underlying `BlobFormatWriter.add_element()` (appends `-1` to lengths index) and the reader side (`FormatBlobReader` / `BlobRecordIterator`) already handle nulls correctly since PR #7847.

### Correctness

The fix is correct and minimal. The control flow for a null blob value is now:

1. `BlobFileWriter._to_blob(None)` returns `None`
2. `GenericRow([None], fields, RowKind.INSERT)` is created
3. `BlobFormatWriter.add_element(row)` detects `row.values[0] is None`, appends `-1` to the lengths index, and returns without writing any data bytes
4. `DeltaVarintCompressor` correctly handles negative values via ZigZag encoding
5. On the read side, `BlobRecordIterator.__next__()` detects `blob_lengths[pos] == -1` and returns `GenericRow([None], ...)`

All layers are consistent.

### Design

- The `Optional[Blob]` return type annotation on `_to_blob` is correct and communicates the contract clearly.
- The import reordering (stdlib before third-party) follows PEP 8 conventions -- good housekeeping.
- The early-return pattern (`if col_data is None: return None`) placed after `as_py()` extraction is the right location, since PyArrow scalars representing null will have `as_py()` return `None`.

### Test

The test covers the key scenario (interleaved null and non-null blobs) and validates full round-trip correctness. Good coverage.

### Minor observation (pre-existing, not blocking)

In `BlobWriter._add_file_metadata()` (line 170), when called from `close_current_writer()` with an integer `row_count`, `value_null_counts` is hardcoded to `[0]`. This means blob file metadata stats will under-report null counts when the file contains null blob entries. This is a pre-existing issue outside the scope of this PR, but worth tracking if stats accuracy becomes important for blob columns in the future (e.g., for query planning or compaction decisions). The `BlobFileWriter` could track a null counter and expose it alongside `row_count`.

### Verdict

LGTM -- the fix is correct, well-scoped, and properly tested.

PR #7902, opened 2026-05-19
[python] PVFS supports jindo backend for OSS reads/writes

shyjsarah:
### Purpose

  `PaimonVirtualFileSystem` (PVFS) backed every OSS `open()` with `ossfs`, whose fsspec buffered file flushes each block via OSS `AppendObject`. When a result exceeds one block, a later `AppendObject` can fail with `PositionNotEqualToLength` (409) on the OSS data-acceleration endpoint due to file-length cache lag.

  This PR makes the PVFS OSS backend selectable via `fs.oss.impl` — the same option `PyArrowFileIO` already honors:

  - `jindo` (default): native JindoSDK, writes via `PutObject` / multipart upload, so `AppendObject` is never used.
  - `legacy`: `ossfs` (the previous behavior).

  When `fs.oss.impl=jindo` but `pyjindosdk` is not installed, PVFS falls back to `ossfs`, consistent with `PyArrowFileIO`.

  Notes:

  - `build_jindo_config` is now shared by `JindoFileSystemHandler` (the PyArrow FileIO path) and the new `create_jindo_oss_filesystem` (the PVFS path), so both jindo entry points
  consume identical OSS credential options.
  - `pyjindo`'s fsspec filesystem needs the `oss://` scheme on paths while `ossfs` needs it stripped, so `_strip_storage_protocol` keeps or strips the scheme according to the active backend.
  - No storage format change; no user-facing API change — `fs.oss.impl` already exists, this PR only makes PVFS honor it.

  ### Tests

  - New `pvfs_oss_filesystem_test.py` — CI unit tests, no DLF / OSS / pyjindosdk required (all stubbed):
    - `fs.oss.impl` dispatch: `jindo` / `legacy` / default / fallback when pyjindosdk is absent / invalid value
    - `_get_filesystem` forwards the table's OSS storage path to the backend
    - `_strip_storage_protocol` keeps `oss://` for jindo, strips it for ossfs
    - `_extract_oss_bucket` parsing and error cases
  - Verified end-to-end against a real DLF REST catalog with both backends (`legacy` and `jindo`) on pyjindo 6.10.2: open/read/write (including an 8 MB multipart write), `info`, `ls`, `cat_file`, `cp_file`, `mv`, `rm`, `get_file` all succeed on the jindo backend.

### DIFF
--- a/paimon-python/pypaimon/filesystem/jindo_file_system_handler.py
+++ b/paimon-python/pypaimon/filesystem/jindo_file_system_handler.py
@@ -24,15 +24,75 @@ from pyarrow.fs import FileInfo, FileSelector, FileType
 
 try:
     import pyjindo.fs as jfs
+    import pyjindo.ossfs as jossfs
     import pyjindo.util as jutil
     JINDO_AVAILABLE = True
 except ImportError:
     JINDO_AVAILABLE = False
     jfs = None
+    jossfs = None
     jutil = None
 
 from pypaimon.common.options import Options
 from pypaimon.common.options.config import OssOptions
+
+
+def build_jindo_config(catalog_options: Options):
+    """Build a pyjindo ``Config`` from OSS catalog options.
+
+    Shared by ``JindoFileSystemHandler`` (the PyArrow FileIO path) and
+    ``create_jindo_oss_filesystem`` (the PVFS fsspec path) so both jindo entry
+    points consume exactly the same credential / endpoint options.
+    """
+    if not JINDO_AVAILABLE:
+        raise ImportError("Module pyjindo is not available. Please install pyjindosdk.")
+
+    config = jutil.Config()
+
+    access_key_id = catalog_options.get(OssOptions.OSS_ACCESS_KEY_ID)
+    access_key_secret = catalog_options.get(OssOptions.OSS_ACCESS_KEY_SECRET)
+    security_token = catalog_options.get(OssOptions.OSS_SECURITY_TOKEN)
+    endpoint = catalog_options.get(OssOptions.OSS_ENDPOINT)
+    region = catalog_options.get(OssOptions.OSS_REGION)
+
+    if access_key_id:
+        config.set("fs.oss.accessKeyId", access_key_id)
+    if access_key_secret:
+        config.set("fs.oss.accessKeySecret", access_key_secret)
+    if security_token:
+        config.set("fs.oss.securityToken", security_token)
+    if endpoint:
+        endpoint_clean = endpoint.replace('http://', '').replace('https://', '')
+        config.set("fs.oss.endpoint", endpoint_clean)
+    if region:
+        config.set("fs.oss.region", region)
+    config.set("fs.oss.user.agent.features", "pypaimon")
+    return config
+
+
+def create_jindo_oss_filesystem(root_uri: str, catalog_options: Options):
+    """Create an fsspec-compatible ``JindoOssFileSystem`` for an OSS bucket.
+
+    ``PaimonVirtualFileSystem`` uses this to back OSS reads/writes with the
+    native JindoSDK instead of ``ossfs``. JindoSDK writes objects via
+    PutObject / multipart upload, so it never issues OSS ``AppendObject`` --
+    the call that fails with ``PositionNotEqualToLength`` (409) on the OSS
+    data-acceleration endpoint when ``ossfs`` flushes a multi-chunk write.
+
+    ``root_uri`` is the bucket root, e.g. ``oss://my-bucket/``; it must carry
+    the bucket so ``JindoOssFileSystem`` can re-attach the ``oss://`` scheme to
+    the bucket-relative paths that ``PaimonVirtualFileSystem`` passes in.
+    """
+    if not JINDO_AVAILABLE:
+        raise ImportError("Module pyjindo is not available. Please install pyjindosdk.")
+
+    return jossfs.JindoOssFileSystem(
+        uri=root_uri,
+        config=build_jindo_config(catalog_options),
+        # PaimonVirtualFileSystem owns directory semantics for the virtual FS;
+        # the backing object-store fs must not auto-create dir-marker objects.
+        auto_mkdir=False,
+    )
 
 
 class JindoInputFile:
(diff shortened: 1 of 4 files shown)


paimon-python/pypaimon/filesystem/jindo_file_system_handler.py
```diff
@@ -24,17 +24,83 @@
 
 try:
     import pyjindo.fs as jfs
+    import pyjindo.ossfs as jossfs
```
jerry-024: Compatibility regression: `pyjindo.ossfs` should not gate `JINDO_AVAILABLE`.

`pyjindo.ossfs` now shares this try-block with `pyjindo.fs` / `pyjindo.util`, so it gates `JINDO_AVAILABLE`. But `JindoFileSystemHandler` (the pre-existing PyArrow FileIO path) only needs `pyjindo.fs` + `pyjindo.util`. On a `pyjindosdk` version that ships `fs`/`util` but not `ossfs`, this `ImportError` flips `JINDO_AVAILABLE` to `False` and silently disables the previously-working PyArrow jindo path — a regression for existing users.

Suggest splitting into two independent flags:

```python
try:
    import pyjindo.fs as jfs
    import pyjindo.util as jutil
    JINDO_AVAILABLE = True
except ImportError:
    JINDO_AVAILABLE = False
    jfs = None
    jutil = None

try:
    import pyjindo.ossfs as jossfs
    JINDO_OSSFS_AVAILABLE = True
except ImportError:
    jossfs = None
    JINDO_OSSFS_AVAILABLE = False
```

Then:
- `build_jindo_config` stays gated on `JINDO_AVAILABLE` (it only needs `jutil`).
- `create_jindo_oss_filesystem` and `_use_jindo_oss_backend` should gate on `JINDO_AVAILABLE and JINDO_OSSFS_AVAILABLE` — the PVFS jindo backend needs both `jossfs` and, via `build_jindo_config`, `jutil`.
- Export `JINDO_OSSFS_AVAILABLE` so `pvfs.py` can import it.
- `test_jindo_falls_back_to_ossfs_when_unavailable` should then patch `JINDO_OSSFS_AVAILABLE` to exercise the real fallback branch.

paimon-python/pypaimon/filesystem/jindo_file_system_handler.py
```diff
@@ -24,17 +24,83 @@
...
+        config=build_jindo_config(catalog_options),
+        # PaimonVirtualFileSystem owns directory semantics for the virtual FS;
+        # the backing object-store fs must not auto-create dir-marker objects.
+        auto_mkdir=False,
+        # Bypass fsspec's _Cached metaclass instance cache, so the only
+        # reference to this filesystem -- and to its underlying native jindo
+        # connection -- is the PaimonRealStorage cache in PVFS. On token
+        # refresh PVFS replaces that entry and the native resources can be
+        # released, instead of being pinned forever by fsspec's global cache.
+        skip_instance_cache=True,
```
jerry-024, line 112: This pins PVFS to a specific `pyjindosdk` API surface: `pyjindo.ossfs.JindoOssFileSystem` must exist and accept the `uri` / `config` / `auto_mkdir` / `skip_instance_cache` keyword arguments. The PR description mentions verification on pyjindo 6.10.2, but nothing declares a minimum supported version.

If an older supported `pyjindosdk` version lacks the `ossfs` module or this constructor signature, users would hit an `ImportError` / `TypeError` rather than the intended clean fallback to `ossfs`.

Could you confirm the minimum `pyjindosdk` version that provides this API, and declare it explicitly (packaging metadata / docs / a comment here)? Otherwise a version/signature guard around the `ossfs` import and this constructor call would be needed.

jerry-024 (commented):
Thanks for the fix — the `fs.oss.impl` dispatch and the shared `build_jindo_config` refactor are clean, and the unit-test coverage is solid.

Two compatibility concerns on the jindo import surface, left inline. One other thing worth calling out in the PR description: with `fs.oss.impl` defaulting to `jindo`, the effective OSS backend for PVFS silently changes from `ossfs` to JindoSDK in any environment that already has `pyjindosdk` installed — so it is a behavior change for existing deployments, not strictly a no-op as "no user-facing API change" might suggest.

paimon-python/pypaimon/filesystem/jindo_file_system_handler.py
shyjsarah: Thanks — declared the minimum explicitly in e989d453.

  Added a new jindo extra to setup.py:

  'jindo': [
      'pyjindosdk>=6.10.4',
  ],

  so pip install pypaimon[jindo] pulls the right version.

  Notes on the choice of 6.10.4 and the ctor:

  - PyPI currently ships only pyjindosdk==6.10.4 (https://pypi.org/simple/pyjindosdk/) — there is no older PyPI release a user could resolve to.
  - pyjindo.ossfs.JindoOssFileSystem.__init__ on 6.10.4 is (self, uri=None, config=None, auto_mkdir=True, **kwargs). skip_instance_cache flows through **kwargs, so the call site stays forward-compatible if a future
  6.x adds further fsspec kwargs.
  - The two-flag split above plus the precise ImportError("pyjindo.ossfs is not available. Please install pyjindosdk>=6.10.4.") give a clean signal when a user is on an older internal build that lacks pyjindo.ossfs,
  instead of an opaque attribute error later.

  Verified end-to-end on a fresh pip install pyjindosdk==6.10.4: PVFS _use_jindo_oss_backend()=True, 8 MB multipart write through JindoOssFile, and the token-refresh close path runs after the write lock is released.

paimon-python/pypaimon/filesystem/jindo_file_system_handler.py
shyjsarah: Thanks — agreed, fixed in e989d453.

  Split into two independent flags so the two pyjindosdk surfaces are tracked separately:

  ```python
  try:
      import pyjindo.fs as jfs
      import pyjindo.util as jutil
      JINDO_AVAILABLE = True
  except ImportError:
      JINDO_AVAILABLE = False
      jfs = jutil = None

  try:
      import pyjindo.ossfs as jossfs
      JINDO_OSSFS_AVAILABLE = True
  except ImportError:
      jossfs = None
      JINDO_OSSFS_AVAILABLE = False

  - build_jindo_config keeps the JINDO_AVAILABLE gate (only needs jutil).
  - create_jindo_oss_filesystem and PaimonVirtualFileSystem._use_jindo_oss_backend now require both flags, so a pyjindosdk build that ships fs/util but not ossfs no longer flips the PyArrow JindoFileSystemHandler path
   to unavailable.
  - pvfs.py imports JINDO_OSSFS_AVAILABLE from the handler module.
  - test_jindo_falls_back_to_ossfs_when_pyjindo_ossfs_missing (renamed) now patches JINDO_OSSFS_AVAILABLE and pins JINDO_AVAILABLE=True, so the result is independent of whether pyjindosdk is installed in the CI image.

jerry-024:
+1

PR #7903, opened 2026-05-19
[hotfix] pin vortex version = 0.70.0 to avoid test failure

steFaiz:
### Purpose
So many PRs are blocked by the same test failure:

```text
pyarrow.lib.ArrowInvalid: External error: Reduced array dtype mismatch from vortex.between to vortex.constant
```

It seems that in vortex, the PR https://github.com/vortex-data/vortex/pull/7948/changes change the optimize logic.
When trying to reduce the between predicate to constant, the returned type losses its nullability.

I tested that 0.70.0 is fine. I'll try to fix this issue in Vortex, and temporarily, pin the vortex version to 0.70.0.


JingsongLi:
+1

PR #7904, opened 2026-05-19
[python] Add replace_tag support to TagManager and FileStoreTable

JunRuiLee:
### Purpose
Java supports replacing an existing tag with a new snapshot via Table.replaceTag() and Spark's ReplaceTagProcedure. pypaimon currently lacks this capability. 
Users who want to update a tag (e.g. periodically refreshing a latest-prod tag to point to the newest snapshot) must delete and re-create it, leaving a window where the tag does not exist. replace_tag provides atomic replacement semantics.

### Tests
- Replace an existing tag to a specified snapshot_id
- Replace with snapshot_id=None uses the latest snapshot
- Replacing a non-existent tag raises ValueError
- Replacing with a non-existent snapshot_id raises ValueError

### DIFF
--- a/paimon-python/pypaimon/table/file_store_table.py
+++ b/paimon-python/pypaimon/table/file_store_table.py
@@ -313,6 +313,28 @@ class FileStoreTable(Table):
         tag_mgr = self.tag_manager()
         tag_mgr.rename_tag(old_name, new_name)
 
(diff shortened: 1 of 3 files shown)


JingsongLi (approved):
+1

PR #7910, opened 2026-05-20
[core] Null out supplier in LazyField after evaluation to prevent memory leak

### LINKED ISSUE #7909: [Bug] LazyField retains supplier reference after evaluation, causing memory leak in chained lazy computations
### Search before asking

### Paimon version

1.4

### Compute Engine

flink 1.18

### Minimal reproduce step

LazyField keeps a strong reference to the supplier even after the lazy value has been evaluated and cached. This causes a memory leak when LazyField instances are chained together, such as in GlobalIndexResult.or() / GlobalIndexResult.and() operations.

should null out the supplier reference after evaluation to break the closure chain and allow GC to reclaim intermediate objects.

### Anything else?

_No response_

zhoulii:
### Purpose

close #7909


JingsongLi:
+1

PR #7911, opened 2026-05-20
[python] Add drop_partitions support to FileSystem catalog and CLI

JunRuiLee:
### Purpose
Add drop_partitions support for the Python FileSystem catalog and expose it through the PyPaimon CLI, aligned with the Java implementation by using the table batch commit path and truncate_partitions semantics to generate an overwrite commit that removes the specified partition data.

### Tests
Added coverage for:
  - Dropping a single partition through FileSystemCatalog.drop_partitions.
  - Dropping multiple partitions through FileSystemCatalog.drop_partitions.
  - Rejecting an empty partition list.
  - Dropping a single partition through the CLI.
  - Dropping multiple partitions through repeated CLI --partition options.

### DIFF
--- a/paimon-python/pypaimon/catalog/filesystem_catalog.py
+++ b/paimon-python/pypaimon/catalog/filesystem_catalog.py
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-from typing import List, Optional, Union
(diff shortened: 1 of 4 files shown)


JingsongLi (approved):
+1

PR #7913, opened 2026-05-20
[spark] Support variant pushdown via SupportsPushDownVariantExtractions for Spark 4.1+

Zouxxyy:
### Purpose

Integrate Spark 4.1+ `SupportsPushDownVariantExtractions` so that `variant_get(col, '$.path', type)` is pushed into the Paimon read path as a shredded variant struct. Same trait-shadowing pattern as `PaimonSupportsRuntimeFiltering` — no-op trait in `paimon-spark-common`, 4.1 binding overrides at the same FQN. Nested variant columns supported; column pruning preserved; full-Variant access on a column rejects per-column pushdown.

### Tests

Six new pushdown checks in `VariantTestBase` (gated `assume(gteqSpark4_1)`), running across 4 `sparkConf` subclasses. Existing 84 functional Variant tests cover correctness unchanged.


JingsongLi (approved):
+1

Issue #7905, opened 2026-05-19
[Bug][Flink] 'scan.parallelism' cannot be scoped to source because PaimonDataStreamScanProvider does not override ParallelismProvider#getParallelism()


hackergin:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Paimon version

1.3.1

### Compute Engine

Flink

### Minimal reproduce step

 -- Flink 1.19+
  SET 'parallelism.default' = '4';

  CREATE TABLE t (
    id BIGINT,
    v  STRING
  ) WITH (
    'connector'        = 'paimon',
    'bucket'           = '16',
    'scan.parallelism' = '16'
  );

  EXPLAIN PLAN FOR
  SELECT id, UPPER(v) FROM t;

  Inspect the resulting job graph in the Flink Web UI / EXPLAIN.

### What doesn't meet your expectations?

Expected

  The scan.parallelism option, by its name and by Flink's standard Table connector contract, is supposed to control only the source operator's parallelism. Downstream operators should
   run at parallelism.default, with a rebalance edge inserted between them and the source:

  Source[t] (parallelism=16)  --rebalance-->  Calc(UPPER(v)) (parallelism=4)

  Actual

  The configured scan.parallelism = 16 propagates to every downstream operator in the forward chain. The whole pipeline runs at parallelism 16, ignoring parallelism.default:

  Source[t] (parallelism=16)  --forward-->  Calc(UPPER(v)) (parallelism=16)

  There is no way through SQL today to read a Paimon table with high source parallelism while keeping downstream computation at the global default.

### Anything else?

Root cause

  PaimonDataStreamScanProvider already inherits the ParallelismProvider interface (Flink's DataStreamScanProvider extends ParallelismProvider), but it never overrides
  getParallelism(), so the planner always receives the default Optional.empty(). Paimon then applies the configured parallelism by mutating the produced DataStream directly

References

  - Flink — [Table connector ParallelismProvider](https://nightlies.apache.org/flink/flink-docs-stable/docs/dev/table/sourcessinks/#parallelism)
  - [FLIP-367: Support Setting Parallelism for Table/SQL Sources](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=263429150)

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

Akash3121:
Hi @JingsongLi, This issue can be closed as the fix PR #9803 is merged to master
Thanks

PR #7914, opened 2026-05-20
[flink] Fix scan parallelism propagating to downstream operators

ArnavBalyan:
### Purpose
 - PaimonDataStreamScanProvider did not override ParallelismProvider getParallelism.
 - As a result whole pipeline runs at the configured scan parallelism.
 - Update scan provider to override and provide getParallelism.
 - Closes user reported bug #7905
### Tests
 - UT

### DIFF
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/BaseDataTableSource.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/BaseDataTableSource.java
@@ -19,6 +19,7 @@
 package org.apache.paimon.flink.source;
 
 import org.apache.paimon.CoreOptions;
+import org.apache.paimon.flink.FlinkConnectorOptions;
 import org.apache.paimon.flink.FlinkConnectorOptions.WatermarkEmitStrategy;
 import org.apache.paimon.flink.PaimonDataStreamScanProvider;
 import org.apache.paimon.flink.lookup.FileStoreLookupFunction;
@@ -206,7 +207,8 @@ public abstract class BaseDataTableSource extends FlinkTableSource
                                 .env(env)
                                 .build(),
                 tableIdentifier.asSummaryString(),
-                table);
+                table,
+                options.get(FlinkConnectorOptions.SCAN_PARALLELISM));
     }
 
     private ScanRuntimeProvider createPushedAggregateScan() {
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/PaimonDataStreamScanProvider.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/PaimonDataStreamScanProvider.java
@@ -29,6 +29,9 @@ import org.apache.flink.table.connector.ProviderContext;
 import org.apache.flink.table.connector.source.DataStreamScanProvider;
 import org.apache.flink.table.data.RowData;
 
+import javax.annotation.Nullable;
+
+import java.util.Optional;
 import java.util.function.Function;
 
 /**
@@ -41,16 +44,27 @@ public class PaimonDataStreamScanProvider implements DataStreamScanProvider, Lin
     private final Function<StreamExecutionEnvironment, DataStream<RowData>> producer;
     private final String name;
     private final Table table;
+    @Nullable private final Integer parallelism;
 
     public PaimonDataStreamScanProvider(
             boolean isBounded,
             Function<StreamExecutionEnvironment, DataStream<RowData>> producer,
             String name,
             Table table) {
+        this(isBounded, producer, name, table, null);
+    }
+
+    public PaimonDataStreamScanProvider(
+            boolean isBounded,
+            Function<StreamExecutionEnvironment, DataStream<RowData>> producer,
+            String name,
+            Table table,
+            @Nullable Integer parallelism) {
         this.isBounded = isBounded;
         this.producer = producer;
         this.name = name;
         this.table = table;
+        this.parallelism = parallelism;
     }
 
     @Override
@@ -62,6 +76,11 @@ public class PaimonDataStreamScanProvider implements DataStreamScanProvider, Lin
     @Override
     public boolean isBounded() {
         return isBounded;
+    }
+
+    @Override
+    public Optional<Integer> getParallelism() {
+        return Optional.ofNullable(parallelism);
     }
 
     @Override
--- a/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/SystemTableSource.java
+++ b/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/SystemTableSource.java
@@ -132,7 +132,8 @@ public class SystemTableSource extends FlinkTableSource {
                     return dataStreamSource;
                 },
                 tableIdentifier.asSummaryString(),
-                table);
+                table,
+                options.get(FlinkConnectorOptions.SCAN_PARALLELISM));
     }
 
     @Override
--- a/paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/lineage/LineageUtilsTest.java
+++ b/paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/lineage/LineageUtilsTest.java
@@ -193,6 +193,24 @@ class LineageUtilsTest {
     }
 
     @Test
+    void testScanProviderGetParallelism() throws Exception {
+        FileStoreTable table =
+                createTable(new HashMap<>(), Collections.emptyList(), Arrays.asList("f0"));
+
+        PaimonDataStreamScanProvider noOverride =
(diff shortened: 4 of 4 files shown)


ArnavBalyan:
cc @JingsongLi thanks!

hackergin:
@ArnavBalyan  Thanks for picking this up! The direction is right, but I think there are a few gaps worth addressing before merging:

  1. Only handles explicitly configured scan.parallelism, misses the inferred path

  The PR only propagates options.get(SCAN_PARALLELISM). However, Paimon's parallelism resolution has two sources — see FlinkTableSource#inferSourceParallelism:

  1. User-configured scan.parallelism (explicit)
  2. infer-scan.parallelism = true (enabled by default), which infers from bucket count / split count

  This PR only covers case (1). For case (2) — which is the default behavior most users hit — the inferred value is still applied via DataStream#setParallelism() inside the
  producer lambda, so ParallelismProvider#getParallelism() returns Optional.empty() and the planner-side isParallelismConfigured=true flag is never set. The forward-edge
  propagation described in #7905 will still happen for any user who relies on inferred parallelism.
  
  2. Parallelism is now set twice — provider + producer

  After this PR, parallelism is applied in two places:
  - The planner reads ParallelismProvider#getParallelism() and calls setParallelism(p, true) on the Source Transformation
  - The producer lambda still calls sourceBuilder.sourceParallelism(inferSourceParallelism(env)) / dataStreamSource.setParallelism(parallelism) internally
  
  The DataStream#setParallelism(int) overload sets isParallelismConfigured=false, while the planner path sets it to true. Depending on execution order, the planner's
  "configured" flag may be overridden back to false, which would silently re-introduce the forward-edge propagation this PR is trying to fix.

  3. Test placement and coverage
  
  - testScanProviderGetParallelism is placed in LineageUtilsTest, but it has nothing to do with lineage.
  - The test only asserts Optional.ofNullable behavior on the field. It doesn't actually verify the bug being fixed — there's no assertion that downstream operators no longer
  inherit the source parallelism, and no assertion on Source Transformation#isParallelismConfigured(). An end-to-end SQL test would be much more convincing, and the inferred
  path from (1) should also be covered.

  4. Flink version note

  ParallelismProvider#getParallelism() only fully solves the forward-edge issue from Flink 1.19+ (FLIP-367), where the planner explicitly calls setParallelism(p, true) on the
  Source Transformation. On older Flink versions the behavior is unchanged. Worth mentioning this in the PR description / a Javadoc note so users know what to expect.

hackergin:
State compatibility / upgrade risk
  
  This fix changes the resulting job graph topology for any existing user who has scan.parallelism set or relies on inferred parallelism:

  - Before: source and downstream operators are chained via forward edges, all running at source parallelism.
  - After: the planner inserts a rebalance edge between source and downstream, and operator chaining is broken.

  For users upgrading from an older Paimon version with an existing Flink savepoint/checkpoint, this is a breaking topology change:

  - The operator IDs / chain hashes of downstream operators will change, so state restore from savepoint will fail (or silently drop state) unless users explicitly set uids on
  every operator — which most SQL users don't.
  - Even watermark/state alignment behavior of downstream stateful operators may differ once the chain is broken.

  This needs to be called out explicitly in the PR description / release notes as a breaking change for stateful streaming jobs.

JingsongLi (commented):
## Code Review

The direction is correct — overriding `ParallelismProvider#getParallelism()` is the right way to inform Flink's planner about source-specific parallelism so it can insert rebalance edges. However, I see a concrete correctness issue in the current implementation and a coverage gap.

### 1. Double-set conflict: `getParallelism()` vs `produceDataStream` will fight over `isParallelismConfigured`

In `BaseDataTableSource.getScanRuntimeProvider()`, the `producer` lambda still calls:
```java
sourceBuilder.sourceParallelism(inferSourceParallelism(env)).env(env).build()
```

Inside `FlinkSourceBuilder.build()` (line 257-258):
```java
if (parallelism != null) {
    dataStream.setParallelism(parallelism);  // sets isParallelismConfigured = false
}
```

Meanwhile, when `getParallelism()` returns a non-empty Optional, the Flink planner calls `Transformation#setParallelism(p, true)` which sets `isParallelismConfigured = true`. The issue is that `produceDataStream()` is invoked **after** the planner configures the transformation — so `FlinkSourceBuilder.build()` will **override** `isParallelismConfigured` back to `false`, silently re-introducing the forward-edge problem this PR aims to fix.

**Suggested fix**: When `getParallelism()` returns a value, the producer lambda should **not** also set parallelism on the DataStream. One approach:

```java
// In BaseDataTableSource.getScanRuntimeProvider():
Integer scanParallelism = options.get(FlinkConnectorOptions.SCAN_PARALLELISM);
return new PaimonDataStreamScanProvider(
        !unbounded,
        env -> sourceBuilder
                .sourceParallelism(scanParallelism != null ? null : inferSourceParallelism(env))
                .env(env)
                .build(),
        tableIdentifier.asSummaryString(),
        table,
        scanParallelism);
```

Or more cleanly, remove the `setParallelism` call from the producer when `getParallelism()` will handle it at the planner level.

### 2. Inferred parallelism is not propagated to `getParallelism()`

`inferSourceParallelism(env)` can compute a non-null parallelism even when `SCAN_PARALLELISM` is null (via bucket count or split count inference). However, the PR only passes the raw `options.get(SCAN_PARALLELISM)` to `getParallelism()`. This means the inferred-parallelism path — which is the **default** behavior for most users — still suffers from forward-edge propagation.

The challenge is that inference requires access to `StreamExecutionEnvironment` (which is only available inside `produceDataStream`), while `getParallelism()` is called by the planner before that. For a complete fix, you may need to eagerly compute the parallelism outside the lambda (or accept this as a known limitation and document it).

### 3. Same issue exists in `SystemTableSource`

`SystemTableSource` (line 123) calls `inferSourceParallelism(env)` inside the lambda and sets it on the DataStreamSource directly:
```java
if (parallelism != null) {
    dataStreamSource.setParallelism(parallelism);
}
```
This will also override `isParallelismConfigured` to false. Same conflict as point 1 applies here.

### 4. Test does not verify the actual fix

`testScanProviderGetParallelism` only tests `Optional.ofNullable` wrapping — it does not assert that `isParallelismConfigured` is set correctly on the resulting Transformation, nor does it verify the downstream operator parallelism behavior. A test that creates a mini pipeline and inspects `transformation.isParallelismConfigured()` would give confidence the fix actually works end-to-end.

Also, the test is placed in `LineageUtilsTest` but is unrelated to lineage. Consider placing it in a dedicated test class or in an existing scan-provider test.

### Minor

- The new 5-arg constructor is fine for backward compatibility. Consider adding a `@VisibleForTesting` annotation on the 4-arg constructor if it's only kept for tests.

ArnavBalyan:
Thanks for review @JingsongLi @hackergin  I'm working on the change could not get to it, will raise shortly!

leaves12138 (changes requested):
I re-checked the latest head (`0d9a08e`). The previous blocking concerns still apply, so I do not think this is ready to merge yet.

The implementation still passes only the explicit `SCAN_PARALLELISM` value to `PaimonDataStreamScanProvider#getParallelism()`, while the default inferred-parallelism path is still applied inside the producer (`sourceBuilder.sourceParallelism(inferSourceParallelism(env))` in `BaseDataTableSource`, and `dataStreamSource.setParallelism(parallelism)` in `SystemTableSource`). That means the default inferred path still does not report a planner-level configured parallelism, and the producer can still overwrite the planner's configured flag when a parallelism is set.

The test also still only verifies `Optional.ofNullable` behavior on the provider field. It does not exercise the planner/runtime behavior that this PR is meant to fix, such as `Transformation#isParallelismConfigured()` or downstream operators no longer inheriting the source parallelism.

Please address the explicit + inferred parallelism paths, avoid setting source parallelism twice, and add an end-to-end/planner-level test for the actual bug. Also, the current status rollup still has failing required checks.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/BaseDataTableSource.java
```diff
@@ -206,7 +207,8 @@ public ScanRuntimeProvider getScanRuntimeProvider(ScanContext scanContext) {
                                 .env(env)
                                 .build(),
                 tableIdentifier.asSummaryString(),
-                table);
+                table,
+                options.get(FlinkConnectorOptions.SCAN_PARALLELISM));
```
JingsongLi, line 211: This only exposes the explicitly configured `scan.parallelism` to `ParallelismProvider`. The default path still calls `inferSourceParallelism(env)` inside the producer (for example bucket/split based inference when `infer-scan.parallelism` is enabled and the env parallelism is default), but `getParallelism()` remains empty for that inferred value. In that common case the planner never gets a configured parallelism to mark with `setParallelism(..., true)`, while the producer still sets the source parallelism through `DataStream#setParallelism` / transformation parallelism. That leaves the forward-edge propagation bug unresolved for the default inferred path. Please compute/reuse the same resolved source parallelism for the provider, and cover the inferred path with a job-graph assertion, not just the explicit option.

PR #7915, opened 2026-05-20
[common] Serialize BlobDescriptor with current version

leaves12138:
### Purpose

`BlobDescriptor.serialize()` should always write the current descriptor format version. Otherwise, a descriptor deserialized from an older version may be serialized with the old version byte while still using the current layout with the magic header, making the bytes inconsistent.

### Changes

- Write `CURRENT_VERSION` in `BlobDescriptor.serialize()`.
- Add a regression test for deserializing a v1 blob descriptor and serializing it back with the current version.

### Tests

- `mvn -U -pl paimon-common -am -Dtest=BlobDescriptorTest -DfailIfNoTests=false test`


JingsongLi:
+1

PR #7916, opened 2026-05-20
 [python] Add statistics infrastructure and ANALYZE commit path

JunRuiLee:
## Purpose

  Internal infrastructure for ANALYZE TABLE:
  - `ColStats` / `Statistics` / `StatsFileHandler` — Java-compatible stats JSON format
  - `FileStoreCommit.commit_statistics()` — creates ANALYZE snapshot, preserves watermark/next_row_id, inherits stats on subsequent commits
  - `StatisticsCollector` — reads merged data via read path, computes distinctCount/nullCount/min/max/avgLen/maxLen with correct type gating and Java-compatible serialization (DATE→epoch days, TIME→millis, TIMESTAMP→formatted string,
  string/binary→no min/max)

  ## Tests

  - ColStats/Statistics serialization roundtrip + Java key format
  - Empty colStats always emitted (avoids Java NPE)
  - StatsFileHandler write/read via snapshot reference

### DIFF
--- a/paimon-python/pypaimon/stats/__init__.py
+++ b/paimon-python/pypaimon/stats/__init__.py
@@ -1,0 +1,28 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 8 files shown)


JingsongLi:
I want to ask if this is used in the production environment? I haven't seen it yet.

JunRuiLee:
> I want to ask if this is used in the production environment? I haven't seen it yet.

Yes, there are scenarios where users access table/statistics information from Python, and that was the original motivation for this PR.

After reconsidering it, I think this requirement may be mostly covered by system tables, especially for reading existing metadata or statistics. The ANALYZE commit support may be a larger scope than necessary for PyPaimon at this stage.

If the community does not think this is useful enough or worth the maintenance cost, I am fine with closing this PR.

JingsongLi:
@JunRuiLee Let's close this PR.

PR #7918, opened 2026-05-20
[python] Implement partition expiration for pypaimon

JunRuiLee:
## Purpose

  Add pypaimon support for expiring partitions based on table options and CLI overrides, matching the Java partition expiration behavior for `values-time` and `update-time` strategies.

  This includes:
  - Partition expiration orchestration for `FileStoreTable`.
  - Partition entry discovery from manifest metadata.
  - Time extraction from partition values.
  - Java `DateTimeFormatter`-style timestamp formatter support.
  - `table expire-partitions` CLI command.
  - Partition expiration options in Python `CoreOptions`.
  - Null partition handling via `partition.default-name`.
  - Default `partition.expiration-max-num=100`.

  ## Tests

  Added `paimon-python/pypaimon/tests/partition_expire_test.py` covering:
  - Extracting partition time from simple date and timestamp values.
  - Extracting time with timestamp patterns across multiple partition fields.
  - Java formatter conversion cases like `yyyyMMdd`.
  - Selecting expired partitions with `values-time`.
  - Selecting expired partitions with `update-time`.
  - Skipping unparseable partition values.
  - Respecting expiration check interval.
  - Respecting max expire count.
  - Empty partition lists.
  - Unconfigured expiration.
  - Non-partitioned tables.
  - Duration parsing for days, hours, minutes, seconds, milliseconds, and combined durations.

### DIFF
--- a/paimon-python/pypaimon/cli/cli_table.py
+++ b/paimon-python/pypaimon/cli/cli_table.py
@@ -818,6 +818,68 @@ def cmd_table_list_partitions(args):
         sys.exit(1)
 
 
+def cmd_table_expire_partitions(args):
+    """
+    Execute the 'table expire-partitions' command.
+
+    Expires (drops) partitions based on time strategies.
+
+    Args:
+        args: Parsed command line arguments.
+    """
+    from pypaimon.cli.cli import load_catalog_config, create_catalog
+    from pypaimon.table.file_store_table import FileStoreTable
+
+    config_path = args.config
+    config = load_catalog_config(config_path)
+    catalog = create_catalog(config)
+
+    table_identifier = args.table
+    parts = table_identifier.split('.')
+    if len(parts) != 2:
+        print(f"Error: Invalid table identifier '{table_identifier}'. "
+              f"Expected format: 'database.table'", file=sys.stderr)
+        sys.exit(1)
+
+    database_name, table_name = parts
+
+    try:
+        table = catalog.get_table(f"{database_name}.{table_name}")
+    except Exception as e:
+        print(f"Error: Failed to get table '{table_identifier}': {e}", file=sys.stderr)
+        sys.exit(1)
+
+    if not isinstance(table, FileStoreTable):
+        print(f"Error: Table '{table_identifier}' is not a FileStoreTable.", file=sys.stderr)
+        sys.exit(1)
+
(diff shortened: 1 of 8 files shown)


JingsongLi (commented):
## Review: [python] Implement partition expiration for pypaimon

Good work porting the partition expiration logic to pypaimon. The architecture is clean and the strategy pattern mirrors the Java implementation well. A few issues I noticed:

---

### Bug: Dead code in `_parse_with_formatter` (partition_time_extractor.py)

The `except` block in `_parse_with_formatter` calls `datetime.strptime(timestamp_string, formatter)` again — the exact same call that just raised `ValueError`. This will always re-raise. The Java version uses a **different** formatter in the fallback (`LocalDate.parse` vs `LocalDateTime.parse`). To match the Java behavior, the fallback should attempt date-only parsing with a separate format.

---

### Design: Java formatter conversion (`_JAVA_TO_PYTHON_PATTERNS`) is fragile

1. **`SSS`/`SS` -> `%f`**: Python's `%f` expects 6 digits (microseconds), while Java's `SSS` is 3-digit millis. This is a subtle semantic mismatch.
2. **Ordering risk**: Sequential `str.replace()` can produce incorrect results for overlapping tokens.

Consider a regex-based tokenizer or documenting the constraints.

---

### Correctness: `PartitionUpdateTimeExpireStrategy` with `last_file_creation_time=0`

When `last_file_creation_time` is 0, the partition will **always** be considered expired. Consider adding a guard or log warning.

---

### Scope: Unrelated methods in file_store_table.py

The `statistics()` and `analyze()` methods appear unrelated to partition expiration and should be in their own PR.

---

### Minor issues

1. `_parse_duration` type hint says `str` but tests pass `None` — add `Optional[str]`.
2. No `expire-batch-size` support (Java has this to prevent OOM on large drops).
3. Verify `truncate_partitions` has same semantics as Java's `commit.dropPartitions(...)`.
4. Missing `.done` partition handling for metastore-managed tables.

---

### Tests

Thorough coverage. Suggest adding a test for `last_file_creation_time=0` to document the "always expired" behavior.

Overall solid work. The main actionable fix is the `_parse_with_formatter` bug.

JunRuiLee:
Thanks @JingsongLi for the review, I've addressed all comments. PTAL~

PR #7295, opened 2026-02-20
[core] Support empty dirs cleaning without bucket sub dir in orphan files cleanup

XiaoHongbo-Hope:
### Purpose/Problem

Currently, OrphanFilesClean does not clean empty partition directories with no bucket subdirectories, causing a lot of empty dirs left.

**Example Scenario:**
 **Note: All belows are empty dirs.**

**Before:**
```
table_root/
├── part1=0/
│   ├── part2=a/
│   │   ├── bucket-0/      ← cleaned
│   │   └── bucket-1/      ← cleaned
│   └── part2=b/           ← NOT cleaned
└── part1=1/               ← NOT cleaned
```

**After:**
```
table_root/
├── part1=0/
│   ├── part2=a/
│   │   ├── bucket-0/      ← cleaned
│   │   └── bucket-1/      ← cleaned
│   └── part2=b/           ← cleaned
└── part1=1/               ← cleaned
```

### Tests
`OrphanFilesCleanTest`
`RemoveOrphanFilesActionITCaseBase`
`LocalOrphanFilesCleanTest`

### Generative AI tooling

UT is coauthored by Cursor

PR #7920, opened 2026-05-21
[core] Avoid deleting directories in orphan files cleanup

XiaoHongbo-Hope:
### Purpose
This PR fixes a potential data loss risk in orphan files cleanup.

  The issue was introduced by #7295, which added support for cleaning empty partition directories without
  bucket subdirectories. In that change, `listFileDirs` may add a partition directory to the directories to
  be scanned when its child listing is empty.

  This is safe when the partition is truly empty. However, if a transient `listStatus` failure returns an
  empty result, a partition directory that still contains data may be treated as empty and later scanned
  again. **When the later listing succeeds**, bucket or partition directories could be collected as orphan file
  candidates.

  Before this fix, `cleanFile` recursively deleted directory candidates via `deleteDirectoryQuietly`, so a
  directory incorrectly collected as an orphan candidate could delete still-referenced data files under
  that directory.

### Changes
This PR adds multiple safeguards:

  1. Filter directories from all orphan file candidate collection paths, including Local, Flink, Spark, and
  snapshot/changelog special cleanup.
  2. Make `cleanFile` refuse directories by removing recursive directory deletion from orphan file cleanup.
  3. Harden Flink empty directory cleanup: respect `dryRun`, stop at the table location, and keep deletion
  non-recursive via `delete(path, false)`.
### Tests
  - `testDirectoriesNotTreatedAsOrphanCandidates`
  - `testDirectoryInSnapshotDirNotTreatedAsCandidate`

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/operation/LocalOrphanFilesClean.java
+++ b/paimon-core/src/main/java/org/apache/paimon/operation/LocalOrphanFilesClean.java
@@ -258,7 +258,7 @@ public class LocalOrphanFilesClean extends OrphanFilesClean {
             }
 
             return files.stream()
-                    .filter(status -> !isDataStructureDirectory(status))
+                    .filter(status -> !status.isDir())
                     .filter(this::oldEnough)
                     .map(status -> Pair.of(status.getPath(), status.getLen()))
                     .collect(Collectors.toList());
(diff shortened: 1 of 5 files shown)


JingsongLi:
+1

PR #7922, opened 2026-05-21
[docs] Migrate documentation from Hugo to Docusaurus

JingsongLi:
Replace the Hugo (hugo-book theme) documentation site with Docusaurus 3, providing a modern UI with dark mode, better search, tab synchronization, and built-in admonition support.

Key changes:
- Convert 145 markdown files from Hugo shortcode syntax to MDX/Docusaurus
- Add custom components: Stable, Unstable, ConfigTable, Label, AllVersions
- Add remark-variable-interpolation plugin for @@VERSION@@ token replacement
- Auto-generate sidebars.js from Hugo frontmatter weights
- Collect 126 URL aliases into client-redirects plugin config
- Move generated config tables to docs/generated/ with raw HTML loading
- Update paimon-docs Maven module to output HTML comments instead of Hugo template comments, and point generated.docs.dir to new location

image


JingsongLi (commented):
The PR is huge because of `docs/yarn.lock`, it is a dependency lock file automatically generated by yarn install. This file needs to be committed - it guarantees that others will receive the exact same dependent version as you when installing yarn, ensuring that the build is reproducible.

docs/docusaurus.config.js
```diff
@@ -0,0 +1,180 @@
...
+        docsRouteBasePath: '/',
+      },
+    ],
+  ],
+
+  plugins: [
+    [
+      '@docusaurus/plugin-client-redirects',
+      {
+        redirects: [],
```
leaves12138: `docs/redirects.js` is generated with the Hugo aliases, but it is never imported here; the plugin is configured with an empty array. As a result, the build does not create the expected old `.html` redirect pages. I verified that after `yarn build`, paths such as `build/append-table/blob-storage.html`, `build/cdc-ingestion/flink-cdc.html`, and `build/concepts/basic-concepts.html` are missing while the new canonical pages exist. Please wire this to `require('./redirects')` (or inline the generated redirects) so existing external links keep working after the migration.

docs/docusaurus.config.js
```diff
@@ -0,0 +1,180 @@
...
+  tagline: 'Data Lake Platform',
+  favicon: 'favicon.png',
+
+  url: 'https://paimon.apache.org',
+  baseUrl: '/docs/master/',
+
+  organizationName: 'apache',
+  projectName: 'paimon',
+
+  onBrokenLinks: 'warn',
```
leaves12138, line 15: `yarn build` currently succeeds but emits broken-anchor warnings for several migrated links, for example `../flink/procedures#procedures`, `./#options-table`, `../maintenance/manage-snapshots#expiring-snapshots`, and the self links in `project/contributing`. These become broken in-page navigations in the generated site, and CI will not catch regressions while they are warnings. Please fix the current broken anchors and consider setting `onBrokenAnchors: 'throw'` for this migration.

leaves12138 (changes requested):
I reviewed the Docusaurus migration and ran the main validations: `yarn build` completes, and `mvn -pl paimon-docs -DskipITs -Dcheckstyle.skip -Drat.skip=true test` passes. However, I found two migration issues that should be fixed before merging.

docs/redirects.js
```diff
@@ -0,0 +1,507 @@
...
+  {
+    "from": "/flink/consumer-id.html",
+    "to": "/flink/consumer-id"
+  },
+  {
+    "from": "/flink/default-value.html",
+    "to": "/flink/default-value"
+  },
+  {
+    "from": "/flink/procedures.html",
```
leaves12138, line 180: There is still a duplicate redirect source for `/flink/procedures.html`. Docusaurus warns during `yarn build`: the same `from` path maps to both `/flink/action-jars` and `/flink/procedures`, and the generated redirect page currently sends `/flink/procedures.html` to `/docs/master/flink/action-jars`. Please deduplicate this alias so the old URL has a single intended target and the docs build is warning-free.

leaves12138 (changes requested):
Thanks for the update. The previous two issues are addressed: `redirects.js` is now wired into Docusaurus, and `onBrokenAnchors: 'throw'` plus the link fixes remove the broken-anchor warnings. `yarn build` now succeeds, but it still emits one redirect warning that should be resolved before merging.

leaves12138 (approved):
Looks good to me now. I re-ran `yarn build`; it completes without Docusaurus warnings, the duplicate redirect source is gone, the old `.html` redirect pages are generated, and `/flink/procedures.html` now has a single redirect target to `/docs/master/flink/procedures`.

PR #7923, opened 2026-05-21
[core] Changelog expiration skip expired changes to avoid FileNotFoundException

yuzelin:
### Purpose
If some changelogs are deleted by other job, we should ensure the commit process won't be interrupted by expiration (like ExpireSnapshotsImpl).


paimon-core/src/main/java/org/apache/paimon/table/ExpireChangelogImpl.java
```diff
@@ -130,10 +130,20 @@ public int expire() {
         // Only clean the snapshot in changelog dir
         maxExclusive = Math.min(maxExclusive, latestChangelogId);
 
+        if (maxExclusive <= earliestChangelogId) {
+            // This happens when retainMin >= total changelog count
+            // (e.g. latestSnapshotId - retainMin + 1 <= earliestChangelogId),
```
LsomeYeah: nit: trailing comma

LsomeYeah (approved):
LGTM with minor changes.

PR #7927, opened 2026-05-21
[common] Introduce N-gram file index for query

xuzifu666:
### Purpose
Currently Paimon not support N-gram file index, so there is room for improvement in scenarios involving prefix and suffix queries.
Let me briefly explain the principles and workflow of the n-gram file index within this PR：

```
┌─────────────────────────────────────────────────────────────────────────────────┐
   │  1. Overall Architecture (Integration with Paimon FileIndexer Framework)        │
   └─────────────────────────────────────────────────────────────────────────────────┘

                               FileIndexer Interface
                                       │
                       ┌───────────────┼───────────────┐
                       │               │               │
               BloomFilter         Bitmap          N-gram ⭐
                  Index             Index           Index
                (equality)        (equality)    (prefix/suffix)

                       N-gram File Index
                              │
           ┌──────────────────┼──────────────────┐
           │                  │                  │
       Writer           Factory            Reader
      (Build)        (SPI Creation)       (Query Filter)
           │                  │                  │
           ▼                  ▼                  ▼

      Write Data  →   NgramFileIndex    →   Query Filter
      Generate N-gram  (Core Impl)       Apply Predicates
      Store HashSet    gram_size param    REMAIN/SKIP


   ┌─────────────────────────────────────────────────────────────────────────────────┐
   │  2. Index Build Process (Writing Phase)                                         │
   └─────────────────────────────────────────────────────────────────────────────────┘

              Input Rows
                   │
                   ▼
       ┌──────────────────────┐
       │ write("hello")       │
       │ write("world")       │
       └────────────┬─────────┘
                    │
                    ▼
       ┌──────────────────────────────────────┐
       │ 1. BinaryString → String conversion  │
       │ 2. Extract N-grams                   │
       │    "hello" → {he, el, ll, lo}        │
       │    "world" → {wo, or, rl, ld}        │
       │ 3. Add to HashSet                    │
       └────────────┬─────────────────────────┘
                    │
                    ▼
       ┌──────────────────────────────────────┐
       │ Final N-gram Set:                    │
       │ {he, el, ll, lo, wo, or, rl, ld}     │
       │                                      │
       │ Size: 680 bytes (100K records)       │
       │ Compression ratio: 0.03%             │
       └────────────┬─────────────────────────┘
                    │
                    ▼
       ┌──────────────────────────────────────┐
       │ Serialization Format:                │
       │ [4B gramSize][4B setSize]            │
       │ [2B len1][N bytes token1]            │
       │ [2B len2][N bytes token2]            │
       │ ...                                  │
       └────────────┬─────────────────────────┘
                    │
                    ▼
              Index Bytes
           (Written to file)


   ┌─────────────────────────────────────────────────────────────────────────────────┐
   │  3. Query Execution Flow (Reading & Filtering Phase)                            │
   └─────────────────────────────────────────────────────────────────────────────────┘

       SQL Query
         │
         ├─ LIKE 'he%'
         ├─ LIKE '%lo'
         ├─ LIKE '%ll%'
         └─ = 'hello'

            ▼
       ┌────────────────────────────────┐
       │ Predicate Optimization         │
       │ LIKE 'prefix%'                 │
       │   → StartsWith("prefix")       │
       │ LIKE '%suffix'                 │
       │   → EndsWith("suffix")         │
       │ LIKE '%middle%'                │
       │   → Contains("middle")         │
       └────────────┬───────────────────┘
                    │
                    ▼
       ┌──────────────────────────────────────┐
       │ FileIndexPredicate.evaluate()        │
       │ Iterate over each data file          │
       └────────────┬─────────────────────--──┘
                    │
                    ▼
       ┌──────────────────────────────────────┐
       │ visitStartsWith(fieldRef, "he")      │
       │                                      │
       │ 1. Get query pattern: "he"           │
       │ 2. Generate N-grams: {he}            │
       │ 3. Check each against index set      │
       │                                      │
       │ Check "he" ∈ {he,el,ll,lo,...}?      │
       └────────────┬─────────────────────--──┘
                    │
           ┌────────┴────────┐
           │                 │
           ▼                 ▼
          YES               NO
           │                 │
           ▼                 ▼
       ┌──────────────┐  ┌──────────────┐
       │ REMAIN       │  │ SKIP         │
       │ File might   │  │ File cannot  │
       │ contain data │  │ contain data │
       │ (scan rows)  │  │ (skip file)  │
       └──────────────┘  └──────────────┘
           │                 │
           └────────┬────────┘
                    │
                    ▼
           Merge results & row-level scan


   ┌─────────────────────────────────────────────────────────────────────────────────┐
   │  4. Filter Decision Logic (Decision Tree)                                       │
   └─────────────────────────────────────────────────────────────────────────────────┘

       Query pattern: pattern
               │
               ▼
       ┌──────────────────────────────┐
       │ pattern == null?             │ YES ──► REMAIN (conservative)
       │ pattern.isEmpty()?           │ YES ──► REMAIN (conservative)
       │ pattern.length < gramSize?   │ YES ──► REMAIN (cannot judge)
       └──────────────┬───────────────┘
                      │ NO
                      ▼
       ┌──────────────────────────────┐
       │ FOR i = 0 TO pattern.length  │
       │     ngram = pattern[i:i+g]   │
       │     IF ngram ∉ ngramSet:     │
       │         RETURN SKIP          │ Early exit (99% case)
       │ RETURN REMAIN               │
       └──────────────────────────────┘


   ┌─────────────────────────────────────────────────────────────────────────────────┐
   │  5. Data Flow Diagram (From Data to Filter Result)                              │
   └─────────────────────────────────────────────────────────────────────────────────┘

       Input Data (100K rows)
               │
               ▼
       ┌──────────────────────────┐
       │ NgramFileIndex.Writer    │ (38 ms)
       │ Build index              │ 2,631 rows/ms
       └──────────────┬───────────┘
                      │
                      ▼
       ┌──────────────────────────┐
       │ Index Bytes (680 bytes)  │
       │ {N-gram set serialized}  │
       └──────────────┬───────────┘
                      │
           ┌──────────┴──────────┐
           │                     │
           ▼                     ▼
       File 1              File 1000
       Index segment       Index segment
           │                     │
           ▼ (25 µs)            ▼ (25 µs)
       ┌─────────────┐    ┌─────────────┐
       │ visitXxx()  │    │ visitXxx()  │
       │ REMAIN/SKIP │    │ REMAIN/SKIP │
       └─────────────┘    └─────────────┘
           │                     │
           └──────────┬──────────┘
                      │
                      ▼
       ┌──────────────────────────────┐
       │ File-level filter result     │
       │ - REMAIN: 100 files          │
       │ - SKIP: 900 files            │
       │ Skipped 900/1000 files       │
       └──────────────┬───────────────┘
                      │
                      ▼
       ┌──────────────────────────────┐
       │ Row-level scan (REMAIN only) │
       │ 100 files × 100K rows        │
       │ = 10M rows (vs 100M without) │
       │ Reduced 90%                  │
       └──────────────────────────────┘

```

benchmark test result：
```
   ┌────────────────────────────────────────────────────────────────────────────────┐
   │ REAL-WORLD PERFORMANCE GAINS (Scenario: Query 1,000 files, 100K rows each)   │
   ├────────────────────────────────────────────────────────────────────────────────┤
   │                                                                                │
   │  No Index                                 With N-gram Index                   │
   │  ─────────────────────────────────────    ─────────────────────────────────   │
   │  • Scan: 1,000 files × 100K rows         • Index: 1,000 × 25µs = 25ms       │
   │  • Total: 100 million rows               • Scan: 100 files × 100K rows      │
   │  • Latency: ~100 ms                      • Total: 10 million rows           │
   │  • Files Scanned: 100%                   • Latency: ~26 ms                 │
   │                                          • Files Scanned: 10%               │
   │                                                                                │
   │  IMPROVEMENT: 74% faster | 90% fewer rows scanned | 99% I/O reduction       │
   │                                                                                │
   └────────────────────────────────────────────────────────────────────────────────┘
```

**The current solution does not employ a Bloom filter, primarily to avoid the issue of false positives.**

### Tests
NgramFileIndexSimpleTest.java
NgramFileIndexTest.java

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/fileindex/ngram/NgramFileIndex.java
+++ b/paimon-common/src/main/java/org/apache/paimon/fileindex/ngram/NgramFileIndex.java
@@ -1,0 +1,206 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.paimon.fileindex.ngram;
+
+import org.apache.paimon.data.BinaryString;
+import org.apache.paimon.fileindex.FileIndexReader;
+import org.apache.paimon.fileindex.FileIndexResult;
+import org.apache.paimon.fileindex.FileIndexWriter;
+import org.apache.paimon.fileindex.FileIndexer;
+import org.apache.paimon.fs.SeekableInputStream;
+import org.apache.paimon.options.Options;
+import org.apache.paimon.predicate.FieldRef;
+import org.apache.paimon.utils.IOUtils;
+
+import java.io.ByteArrayOutputStream;
+import java.io.DataInputStream;
+import java.io.DataOutputStream;
+import java.io.IOException;
+import java.util.HashSet;
+import java.util.Set;
+
+import static org.apache.paimon.fileindex.FileIndexResult.REMAIN;
+import static org.apache.paimon.fileindex.FileIndexResult.SKIP;
+
+/** N-gram file index for string prefix/suffix/contains queries. */
+public class NgramFileIndex implements FileIndexer {
+
+    private static final int DEFAULT_GRAM_SIZE = 2;
+    private static final String GRAM_SIZE = "gram_size";
+
+    private final int gramSize;
+
+    public NgramFileIndex(Options options) {
+        this.gramSize = options.getInteger(GRAM_SIZE, DEFAULT_GRAM_SIZE);
+    }
+
+    @Override
+    public FileIndexWriter createWriter() {
+        return new Writer(gramSize);
+    }
+
+    @Override
+    public FileIndexReader createReader(SeekableInputStream inputStream, int start, int length) {
+        try {
+            inputStream.seek(start);
+            byte[] serializedBytes = new byte[length];
+            IOUtils.readFully(inputStream, serializedBytes);
+            return new Reader(serializedBytes);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    /** Writer for N-gram index. */
+    private static class Writer extends FileIndexWriter {
+
+        private final int gramSize;
+        private final Set<String> ngramSet = new HashSet<>();
+
+        Writer(int gramSize) {
+            this.gramSize = gramSize;
+        }
+
+        @Override
+        public void write(Object key) {
+            if (key == null) {
+                return;
+            }
+            String value = ((BinaryString) key).toString();
+            addNgrams(value);
+        }
+
+        private void addNgrams(String value) {
+            if (value.length() < gramSize) {
+                ngramSet.add(value);
+            } else {
+                for (int i = 0; i <= value.length() - gramSize; i++) {
+                    ngramSet.add(value.substring(i, i + gramSize));
+                }
+            }
+        }
+
+        @Override
+        public byte[] serializedBytes() {
+            try {
+                ByteArrayOutputStream byteOut = new ByteArrayOutputStream();
+                DataOutputStream dataOut = new DataOutputStream(byteOut);
+
+                dataOut.writeInt(gramSize);
+                dataOut.writeInt(ngramSet.size());
+
+                for (String token : ngramSet) {
+                    byte[] tokenBytes = token.getBytes("UTF-8");
+                    dataOut.writeShort(tokenBytes.length);
+                    dataOut.write(tokenBytes);
+                }
+
+                dataOut.close();
+                return byteOut.toByteArray();
+            } catch (IOException e) {
+                throw new RuntimeException(e);
+            }
+        }
+    }
+
+    /** Reader for N-gram index. */
+    private static class Reader extends FileIndexReader {
+
+        private final int gramSize;
+        private final Set<String> ngramSet;
+
+        Reader(byte[] serializedBytes) {
+            try {
+                DataInputStream dataIn =
+                        new DataInputStream(new java.io.ByteArrayInputStream(serializedBytes));
+                this.gramSize = dataIn.readInt();
+                int setSize = dataIn.readInt();
+                this.ngramSet = new HashSet<>(setSize);
+
+                for (int i = 0; i < setSize; i++) {
+                    int tokenLength = dataIn.readShort();
+                    byte[] tokenBytes = new byte[tokenLength];
+                    dataIn.readFully(tokenBytes);
+                    ngramSet.add(new String(tokenBytes, "UTF-8"));
+                }
+
+                dataIn.close();
+            } catch (IOException e) {
+                throw new RuntimeException(e);
+            }
+        }
+
+        @Override
+        public FileIndexResult visitEqual(FieldRef fieldRef, Object literal) {
+            return checkPattern(literalToString(literal));
+        }
+
+        @Override
+        public FileIndexResult visitStartsWith(FieldRef fieldRef, Object literal) {
+            return checkPattern(literalToString(literal));
+        }
+
+        @Override
+        public FileIndexResult visitEndsWith(FieldRef fieldRef, Object literal) {
+            return checkPattern(literalToString(literal));
+        }
+
+        @Override
+        public FileIndexResult visitContains(FieldRef fieldRef, Object literal) {
+            return checkPattern(literalToString(literal));
+        }
+
+        @Override
+        public FileIndexResult visitLike(FieldRef fieldRef, Object literal) {
+            String pattern = literalToString(literal);
+            String[] parts = pattern.split("%");
+            String longestPart = "";
+            for (String part : parts) {
+                if (part.length() > longestPart.length()) {
+                    longestPart = part;
+                }
+            }
+            return checkPattern(longestPart);
+        }
+
+        private FileIndexResult checkPattern(String pattern) {
+            if (pattern == null || pattern.isEmpty() || pattern.length() < gramSize) {
+                return REMAIN;
+            }
+
+            for (int i = 0; i <= pattern.length() - gramSize; i++) {
+                String ngram = pattern.substring(i, i + gramSize);
+                if (!ngramSet.contains(ngram)) {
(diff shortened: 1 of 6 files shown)


JingsongLi (commented):
## Review

The concept is simple and effective — store all n-grams from a file's string column as a HashSet, then at query time check if the query pattern's n-grams exist in the set. If any n-gram from the query is missing, the file cannot contain a match → SKIP.

### Critical Issues

**1. The index size grows unboundedly with data cardinality — no upper bound.**

The n-gram set is a `HashSet<String>`. For a 2-gram index over diverse string data, the set is bounded by the alphabet squared (e.g., ~700 unique 2-grams for ASCII lowercase+digits+common chars). But for larger gram sizes or Unicode data, the set can grow unboundedly. A file with 1M rows of UUIDs will produce a massive index.

The PR description claims "680 bytes for 100K records" — that's because the benchmark uses a small alphabet (6 prefixes). Real-world data (UUIDs, URLs, free text) will produce much larger indexes. There's no size cap or fallback to a bloom filter when the set exceeds a threshold.

Consider either: (a) add a max-size config that degrades to REMAIN when exceeded, or (b) use a bloom filter when the n-gram count exceeds a threshold (the PR states "no bloom filter to avoid false positives" — but an unbounded HashSet serialized as strings can be worse than a bounded bloom filter in practice).

**2. `visitLike` pattern parsing is incorrect.**

```java
public FileIndexResult visitLike(FieldRef fieldRef, Object literal) {
    String pattern = literalToString(literal);
    String[] parts = pattern.split("%");
    String longestPart = "";
    for (String part : parts) {
        if (part.length() > longestPart.length()) {
            longestPart = part;
        }
    }
    return checkPattern(longestPart);
}
```

Problems:
- `split("%")` doesn't handle LIKE escape characters (`_` wildcard, `\%` escaped percent)
- For pattern `%hello%world%`, splitting gives `["", "hello", "world"]`. It picks "hello" (5 chars) or "world" (5 chars) — only checks ONE part. But correct logic should check ALL non-wildcard parts: if ANY part's n-grams are missing, we can SKIP
- For pattern `hello%` → splits to `["hello"]` — works. But for `%` alone → splits to `["", ""]` → longestPart is `""` → REMAIN. OK but fragile.

Should check all parts, not just the longest:
```java
for (String part : parts) {
    FileIndexResult result = checkPattern(part);
    if (result == SKIP) return SKIP;
}
return REMAIN;
```

**3. `visitEqual` semantics are wrong for equality.**

```java
public FileIndexResult visitEqual(FieldRef fieldRef, Object literal) {
    return checkPattern(literalToString(literal));
}
```

If the file contains "hello" and "world", the n-gram set is {he,el,ll,lo,wo,or,rl,ld}. Query `visitEqual("helo")` would check n-grams {he,el,lo} — all present in the set! So it returns REMAIN, but "helo" is NOT in the file. This is expected (false positive), but the PR description says "avoid the issue of false positives" — it doesn't, it just reduces them compared to bloom filters.

More importantly, for `visitEqual` on strings, a bloom filter on the full string value (not n-grams) would be more effective since it's an exact-match check. The n-gram approach is specifically designed for substring/prefix/suffix — `visitEqual` should probably just delegate to the base class (return REMAIN) or use a separate bloom filter.

**4. `writeShort(tokenBytes.length)` — token length limited to 65535 bytes.**

With `gramSize=2` this is safe. But if someone sets `gramSize=100000` (no validation), a single n-gram could exceed the short limit. Add validation that `gramSize` is reasonable (e.g., 2-10).

**5. Strings shorter than `gramSize` are stored as-is in the n-gram set.**

```java
private void addNgrams(String value) {
    if (value.length() < gramSize) {
        ngramSet.add(value);  // stored whole
    } else { ... }
}
```

But `checkPattern` requires `pattern.length() >= gramSize` to not early-return REMAIN:
```java
if (pattern == null || pattern.isEmpty() || pattern.length() < gramSize) {
    return REMAIN;
}
```

So these short values can never be matched by the index — they're stored but never queried. This wastes space. Either don't store them, or handle short patterns differently (direct set lookup for short patterns).

### Minor Issues

- `NgramFileIndexFactory.create()` ignores `DataType` — doesn't pass it to the index. Fine for now but inconsistent with other factories.
- `NgramFileIndexSimpleTest.testNgramGeneration` tests the expected set directly but doesn't actually verify the serialized bytes contain these n-grams (just checks a local HashSet).
- Benchmark tests should not be `@Test` methods that run in CI — they produce console output and have loose assertions (`isLessThan(5000)` ms). Move to a separate benchmark class or exclude from CI.
- `"UTF-8"` string should use `StandardCharsets.UTF_8` to avoid the checked `UnsupportedEncodingException`.

### Summary

The core idea works for `startsWith`/`endsWith`/`contains` on limited-alphabet data. Main issues: (1) unbounded index size for diverse data, (2) `visitLike` should check all parts not just the longest, (3) `visitEqual` gives a false sense of effectiveness.

paimon-common/src/main/java/org/apache/paimon/fileindex/ngram/NgramFileIndex.java
```diff
@@ -0,0 +1,206 @@
...
+
+        @Override
+        public FileIndexResult visitContains(FieldRef fieldRef, Object literal) {
+            return checkPattern(literalToString(literal));
+        }
+
+        @Override
+        public FileIndexResult visitLike(FieldRef fieldRef, Object literal) {
+            String pattern = literalToString(literal);
+            String[] parts = pattern.split("%");
```
JingsongLi, line 172: This LIKE handling is unsafe for SQL wildcard semantics. `_` is a single-character wildcard in Paimon LIKE (for example, the existing predicate test expects `LIKE "st_rt%"` to match `startxx`), but this code treats `_` as a literal n-gram. I reproduced the new index returning SKIP for a file containing only `startxx` with `reader.visitLike(..., "st_rt%").remain() == false`, which would incorrectly drop a file containing matching rows. Escaped `%` has a similar risk because `split("%")` does not distinguish wildcard percent from literal percent. Could we either parse LIKE patterns with the same wildcard/escape rules as `Like.sqlToRegexLike` and only check literal runs, or conservatively return REMAIN whenever the pattern contains `_` or escaped wildcards?

paimon-common/src/main/java/org/apache/paimon/fileindex/ngram/NgramFileIndex.java
```diff
@@ -0,0 +1,206 @@
...
+/** N-gram file index for string prefix/suffix/contains queries. */
+public class NgramFileIndex implements FileIndexer {
+
+    private static final int DEFAULT_GRAM_SIZE = 2;
+    private static final String GRAM_SIZE = "gram_size";
+
+    private final int gramSize;
+
+    public NgramFileIndex(Options options) {
+        this.gramSize = options.getInteger(GRAM_SIZE, DEFAULT_GRAM_SIZE);
```
JingsongLi, line 50: `gram_size` should be validated before the writer is created. This option is user-provided via `file-index.ngram.<column>.gram_size`; with `gram_size=-1` I can reproduce the first write failing with `StringIndexOutOfBoundsException` from `substring(i, i + gramSize)`, and `gram_size=0` builds an index containing only the empty token so every query effectively returns REMAIN. Could we reject non-positive values here with a clear `IllegalArgumentException` (and preferably cover it with a small factory/index test)?

JingsongLi:

PR #7928, opened 2026-05-22
[spark] Unify Spark write column alignment via PaimonOutputResolver

Zouxxyy:
### Purpose

Replace the ad-hoc column-alignment logic in `PaimonAnalysis` (a parallel,
top-level-only implementation of `TableOutputResolver`) with a dedicated
`PaimonOutputResolver` — a trimmed fork of `TableOutputResolver` with
`mergeSchemaEnabled` support that applies uniformly at every nesting depth.

- Missing target fields → NULL-fill at any depth
- Extra source fields → reject when `mergeSchemaEnabled=false`, else kept
  inline so `SchemaHelper.mergeSchema` evolves the table at write time
- Nullable downcast → `AssertNotNull`; type mismatch → tagged `Cast`

`paimonWriteResolved` now mirrors Spark's V2 `outputResolved` strictness
(name + type + nullability, including nested). A `PAIMON_WRITE_RESOLVED`
`TreeNodeTag` prevents re-entry when extras are kept inline.

### Tests

Added recursive nested by-name/by-position alignment cases and merge-schema
conflict cases (deep nested array/struct, top-level same-name type mismatch)
in `DataFrameWriteTestBase`, `DDLTestBase`, `WriteMergeSchemaTest`.


JingsongLi:
+1

PR #7929, opened 2026-05-22
[flink] Support blob compaction for data evolution tables

leaves12138:
### Purpose
Enable Flink data-evolution compaction to plan and execute blob compaction tasks.

This implements `DataEvolutionCompactTask` support for `blobTask` by rewriting dedicated blob files with `MultipleBlobFileWriter`. It also enables the Flink data-evolution compact source to plan blob tasks.

The reader path now also supports blob-only merge groups, which is required when a data-evolution blob task contains multiple blob fields without the corresponding normal data file in the same split.

### Tests
- `mvn -pl paimon-core -DfailIfNoTests=false -Dtest=MultipleBlobTableTest#testDataEvolutionBlobCompaction test`
- `mvn -pl paimon-flink/paimon-flink-common -DfailIfNoTests=false -Dtest=BlobTableITCase#testBlobCompaction+testMultipleBlobCompaction test`
- `mvn -pl paimon-core -DfailIfNoTests=false -Dtest=MultipleBlobTableTest test`
- `mvn -pl paimon-flink/paimon-flink-common -DfailIfNoTests=false -Dtest=BlobTableITCase test`

### DIFF
--- a/paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/BlobTableITCase.java
+++ b/paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/BlobTableITCase.java
@@ -108,6 +108,44 @@ public class BlobTableITCase extends CatalogITCaseBase {
     }
 
     @Test
+    public void testBlobCompaction() throws Exception {
+        for (int i = 1; i <= 10; i++) {
+            batchSql("INSERT INTO blob_table VALUES (%s, 'paimon', X'48656C6C6F')", i);
+        }
+        batchSql("INSERT INTO blob_table VALUES (1, 'paimon', X'48656C6C6F')");
+
+        assertThat(batchSql("SELECT COUNT(*) FROM `blob_table$files`"))
+                .containsExactly(Row.of(22L));
+        assertThat(
+                        batchSql(
+                                "SELECT COUNT(*) FROM `blob_table$files` WHERE file_path LIKE '%%.blob'"))
+                .containsExactly(Row.of(11L));
+        assertThat(
+                        batchSql(
+                                "SELECT COUNT(*) FROM `blob_table$files` WHERE file_path NOT LIKE '%%.blob'"))
+                .containsExactly(Row.of(11L));
+
+        tEnv.getConfig().set("table.dml-sync", "true");
+        tEnv.executeSql("CALL sys.compact(`table` => 'default.blob_table')").await();
+
+        assertThat(batchSql("SELECT COUNT(*) FROM `blob_table$files`"))
+                .containsExactly(Row.of(12L));
+        assertThat(
+                        batchSql(
+                                "SELECT COUNT(*) FROM `blob_table$files` WHERE file_path LIKE '%%.blob'"))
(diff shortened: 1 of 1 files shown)


JingsongLi (commented):
## Review: [flink] Support blob compaction for data evolution tables

### Summary

This PR adds a single integration test `testBlobCompaction` to `BlobTableITCase.java` (38 lines added, 0 deleted). No production code changes are included in the diff.

### Issues Found

**1. Contradictory assertions vs. stated purpose**

The PR title claims to "Support blob compaction for data evolution tables," but the test assertions show that blob files are NOT compacted after calling `sys.compact`:

```java
// Before compaction: 11 blob files
// After compaction: still 11 blob files  <-- blob files are untouched
assertThat(batchSql("SELECT COUNT(*) FROM `blob_table$files` WHERE file_path LIKE '%%.blob'"))
        .containsExactly(Row.of(11L));
```

Only the normal data files are reduced (from 11 to 1). This test validates that data-evolution compaction does NOT break in the presence of blob files, but it does not demonstrate blob compaction support. If the intent is to compact blob files, the expected blob file count after compaction should be 1, not 11.

**2. Missing `blob-compaction.enabled` table property**

Looking at how `DataEvolutionTableCompactSource` passes `table.coreOptions().blobCompactionEnabled()` to the `DataEvolutionCompactCoordinator`, blob compaction is gated by this option. The DDL in this PR for `blob_table` does not set `'blob-compaction.enabled'='true'`, so the compact procedure will never plan blob tasks for this table. Either:
- The DDL should include `'blob-compaction.enabled'='true'` if the test intends to verify blob compaction, or
- The PR title/description should be updated to reflect that this only tests data-file compaction coexistence with blob files.

**3. Missing production code**

The PR description references implementing `DataEvolutionCompactTask` support for `blobTask` with `MultipleBlobFileWriter` and enabling the Flink compact source to plan blob tasks, but the diff contains no production code. If those changes are in a dependent PR, this should be noted as a stacked PR with a dependency link.

**4. Incomplete test coverage per the stated test plan**

The PR body references running `BlobTableITCase#testBlobCompaction+testMultipleBlobCompaction`, but only `testBlobCompaction` is added. A `testMultipleBlobCompaction` test (exercising the `multiple_blob_table` with two blob fields) would strengthen the coverage, especially since the description mentions "multiple blob fields without the corresponding normal data file in the same split."

### Minor Observations

- The test inserts `id=1` twice (once in the loop, once after), which is good for verifying duplicate handling. The final assertion correctly expects 2 rows for `id=1`.
- The use of `tEnv.getConfig().set("table.dml-sync", "true")` combined with `.await()` is redundant but harmless; `.await()` alone should suffice for `CALL` statements.
- The hex literal `X'48656C6C6F'` decodes to "Hello" (bytes `{72, 101, 108, 108, 111}`), which is consistent across assertions.

### Recommendation

This PR in its current form does not achieve the goal stated in the title. It should either (a) include the production code enabling blob compaction in the data-evolution path, or (b) be re-scoped as a test-only PR that verifies normal compaction still works correctly when blob files are present, with the title updated accordingly.

PR #7932, opened 2026-05-22
[flink] Support blob compaction for data evolution tables

leaves12138:
### Purpose
Enable Flink data-evolution compaction to plan and execute blob compaction tasks.

This implements `DataEvolutionCompactTask` support for `blobTask` by rewriting dedicated blob files with `MultipleBlobFileWriter`. It also enables the Flink data-evolution compact source to plan blob tasks.

The reader path now also supports blob-only merge groups, which is required when a data-evolution blob task contains multiple blob fields without the corresponding normal data file in the same split.

### Tests
- `mvn -pl paimon-core -DfailIfNoTests=false -Dtest=MultipleBlobTableTest#testDataEvolutionBlobCompaction test`
- `mvn -pl paimon-flink/paimon-flink-common -DfailIfNoTests=false -Dtest=BlobTableITCase#testBlobCompaction+testMultipleBlobCompaction test`
- `mvn -pl paimon-core -DfailIfNoTests=false -Dtest=MultipleBlobTableTest test`
- `mvn -pl paimon-flink/paimon-flink-common -DfailIfNoTests=false -Dtest=BlobTableITCase test`

### DIFF
--- a/docs/generated/core_configuration.html
+++ b/docs/generated/core_configuration.html
@@ -435,7 +435,7 @@
             <td>The TTL in rocksdb index for cross partition upsert (primary keys not contain all partition fields), this can avoid maintaining too many indexes and lead to worse and worse performance, but please note that this may also cause data duplication.</td>
         </tr>
         <tr>
-            <td><h5>data-evolution.compaction.blob.enabled</h5></td>
+            <td><h5>blob-compaction.enabled</h5></td>
             <td style="word-wrap: break-word;">false</td>
             <td>Boolean</td>
             <td>Whether to compact blob files when compacting a data evolution table.</td>
--- a/docs/generated/core_configuration.html
+++ b/docs/generated/core_configuration.html
@@ -435,6 +435,12 @@
             <td>The TTL in rocksdb index for cross partition upsert (primary keys not contain all partition fields), this can avoid maintaining too many indexes and lead to worse and worse performance, but please note that this may also cause data duplication.</td>
         </tr>
         <tr>
+            <td><h5>data-evolution.compaction.blob.enabled</h5></td>
+            <td style="word-wrap: break-word;">false</td>
+            <td>Boolean</td>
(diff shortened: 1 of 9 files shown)


JingsongLi (changes requested):
1. Should blob compaction be controlled by option
2. In CompactPlanner and DataEvolution CompactTask, the logic for parsing the field ID of blob files is basically the same.

paimon-core/src/main/java/org/apache/paimon/append/dataevolution/DataEvolutionCompactTask.java
```diff
@@ -165,6 +184,207 @@ public CommitMessage doCompact(FileStoreTable table, String commitUser) throws E
...
+        DataSplit dataSplit =
+                DataSplit.builder()
+                        .withPartition(partition)
+                        .withBucket(0)
+                        .withDataFiles(sortedCompactBefore)
+                        .withBucketPath(pathFactory.parent().toString())
+                        .rawConvertible(false)
+                        .build();
+        RecordReader<InternalRow> reader =
+                store.newDataEvolutionRead().withReadType(blobWriteType).createReader(dataSplit);
```
JingsongLi, line 216: You should use `blob-as-descriptor` to read blob.

docs/generated/core_configuration.html
```diff
@@ -434,6 +434,12 @@
             <td>Duration</td>
             <td>The TTL in rocksdb index for cross partition upsert (primary keys not contain all partition fields), this can avoid maintaining too many indexes and lead to worse and worse performance, but please note that this may also cause data duplication.</td>
         </tr>
+        <tr>
+            <td><h5>data-evolution.compaction.blob.enabled</h5></td>
```
JingsongLi: blob-compaction.enabled

leaves12138: Fixed

JingsongLi:
+1

PR #7933, opened 2026-05-22
[globalindex] Support multi-column GlobalIndex framework

CrownChu:
Extend the GlobalIndex SPI, build path, and query path to support one index builder handling multiple columns (e.g. Lucene indexing title + content + tags together). Key changes:
  
  - GlobalIndexerFactory/GlobalIndexer: add List<DataField> create overloads
  - GlobalIndexMultiColumnWriter: new interface for multi-column writes
  - GlobalIndexBuilderUtils: toIndexFileMetas/createIndexWriter accept List<DataField>
  - GlobalIndexScanner: route extraFieldIds to same reader group
  - VectorScanImpl/FullTextScanImpl: match against extraFieldIds
  - GenericIndexTopoBuilder (Flink): multi-column projection and writer dispatch
  - DefaultGlobalIndexBuilder/TopoBuilder (Spark): multi-column support
  - All single-column APIs preserved for backward compatibility

  Purpose

  Some index engines (e.g. Lucene) can build a single index over multiple columns — full-text on title and vector on embedding in the same index file. Previously the GlobalIndex SPI only supported one column
  per indexer: GlobalIndexerFactory.create(DataField, Options) and GlobalIndexSingletonWriter.write(Object). This meant multi-column engines had to create separate index files per column, losing co-located
  search benefits and doubling I/O.
  
  This PR adds a multi-column path through the entire stack:

  1. SPI layer (paimon-common): GlobalIndexerFactory.create(List<DataField>, Options) default method (falls back to single-column for existing implementations). New GlobalIndexMultiColumnWriter interface
  accepts InternalRow with all indexed columns projected in field order.
  2. Core index metadata (paimon-core): GlobalIndexBuilderUtils.toIndexFileMetas populates GlobalIndexMeta.extraFieldIds from the field list beyond the first. GlobalIndexScanner.createReaders resolves the full
   field list from metadata and passes it to the factory. Extra field IDs are registered in the indexMetas map so queries against any column in the group find the same reader.
  3. Query path (paimon-core): VectorScanImpl and FullTextScanImpl check extraFieldIds when matching index files to query columns, so a vector query on embedding finds an index whose primary field is title but
   includes embedding as an extra field.
  4. Flink build (paimon-flink): GenericIndexTopoBuilder accepts List<String> indexColumns, projects all columns + _ROW_ID, and dispatches to GlobalIndexMultiColumnWriter.write(InternalRow) when multi-column.
  findMinNonIndexableRowId checks containsAll(indexColumns) for schema evolution safety.
  5. Spark build (paimon-spark): DefaultGlobalIndexBuilder and DefaultGlobalIndexTopoBuilder gain List<DataField> overloads, with single-column constructors delegating to the multi-column path.

  All existing single-column callers are unchanged — new APIs have default implementations that delegate to the original single-column methods.

  Tests
  
  - GenericIndexTopoBuilderTest: Updated findMinNonIndexableRowId call to use List<String> signature.
  - paimon-lucene module (on feature-globalindex-support-multi-test branch): A full Lucene 9.12.0 implementation exercising the multi-column framework end-to-end:
    - LuceneGlobalIndexTest — single-column vector write/search, top-K score ordering (2 tests)
    - LuceneFullTextIndexTest — single-column full-text write/search, score verification, no-results case (3 tests)
    - LuceneMultiColumnIndexTest — multi-column (text + vector) via GlobalIndexerFactory.create(List<DataField>, Options) → GlobalIndexMultiColumnWriter.write(InternalRow), verifies both full-text and vector
  queries on same index file; also tests SPI discovery path via GlobalIndexer.create("lucene", ...) (2 tests)
    - LuceneGlobalIndexScanTest — end-to-end Paimon table tests: creates FileStoreTable, writes data, builds Lucene index, commits via DataIncrement.indexIncrement, queries via
  VectorSearchBuilder/FullTextSearchBuilder → ReadBuilder.newScan().withGlobalIndexResult(), reads back rows and asserts correctness (3 tests)

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/globalindex/GlobalIndexerFactory.java
+++ b/paimon-common/src/main/java/org/apache/paimon/globalindex/GlobalIndexerFactory.java
@@ -29,7 +29,7 @@ public interface GlobalIndexerFactory {
 
     String identifier();
 
-    GlobalIndexer create(DataField dataField, Options options);
+    GlobalIndexer create(DataField indexField, Options options);
 
     /**
      * Whether this index type supports multi-column indexes. A factory that returns {@code true}
@@ -40,18 +40,18 @@ public interface GlobalIndexerFactory {
     }
 
     /**
-     * Creates an indexer over a primary column plus optional extra columns. {@code dataField} is
+     * Creates an indexer over a primary column plus optional extra columns. {@code indexField} is
      * the primary column; {@code extraFields} holds the remaining columns and is empty for a
      * single-column index.
      */
     default GlobalIndexer create(
-            DataField dataField, List<DataField> extraFields, Options options) {
+            DataField indexField, List<DataField> extraFields, Options options) {
         if (extraFields != null && !extraFields.isEmpty()) {
             throw new UnsupportedOperationException(
                     String.format(
                             "Index type '%s' does not support multi-column index, got extra columns: %s",
                             identifier(), extraFields));
         }
-        return create(dataField, options);
+        return create(indexField, options);
     }
 }
--- a/paimon-common/src/main/java/org/apache/paimon/globalindex/GlobalIndexerFactory.java
+++ b/paimon-common/src/main/java/org/apache/paimon/globalindex/GlobalIndexerFactory.java
@@ -32,6 +32,12 @@ public interface GlobalIndexerFactory {
     GlobalIndexer create(DataField dataField, Options options);
 
     default GlobalIndexer create(List<DataField> fields, Options options) {
+        if (fields.size() > 1) {
+            throw new UnsupportedOperationException(
+                    String.format(
+                            "Index type '%s' does not support multi-column index, got columns: %s",
+                            identifier(), fields));
+        }
         return create(fields.get(0), options);
     }
 }
--- a/paimon-common/src/main/java/org/apache/paimon/globalindex/GlobalIndexerFactory.java
+++ b/paimon-common/src/main/java/org/apache/paimon/globalindex/GlobalIndexerFactory.java
@@ -32,14 +32,6 @@ public interface GlobalIndexerFactory {
     GlobalIndexer create(DataField indexField, Options options);
 
     /**
-     * Whether this index type supports multi-column indexes. A factory that returns {@code true}
-     * must override {@link #create(DataField, List, Options)} to handle extra columns.
-     */
-    default boolean supportsMultiColumn() {
-        return false;
-    }
-
-    /**
      * Creates an indexer over a primary column plus optional extra columns. {@code indexField} is
      * the primary column; {@code extraFields} holds the remaining columns and is empty for a
      * single-column index.
--- a/paimon-common/src/main/java/org/apache/paimon/globalindex/GlobalIndexerFactory.java
+++ b/paimon-common/src/main/java/org/apache/paimon/globalindex/GlobalIndexerFactory.java
@@ -31,6 +31,14 @@ public interface GlobalIndexerFactory {
 
     GlobalIndexer create(DataField dataField, Options options);
 
+    /**
+     * Whether this index type supports multi-column indexes. A factory that returns {@code true}
+     * must override {@link #create(List, Options)} to handle more than one column.
+     */
+    default boolean supportsMultiColumn() {
+        return false;
+    }
+
     default GlobalIndexer create(List<DataField> fields, Options options) {
         if (fields.size() > 1) {
             throw new UnsupportedOperationException(
--- a/paimon-common/src/main/java/org/apache/paimon/globalindex/GlobalIndexerFactory.java
+++ b/paimon-common/src/main/java/org/apache/paimon/globalindex/GlobalIndexerFactory.java
@@ -33,19 +33,25 @@ public interface GlobalIndexerFactory {
 
     /**
      * Whether this index type supports multi-column indexes. A factory that returns {@code true}
-     * must override {@link #create(List, Options)} to handle more than one column.
+     * must override {@link #create(DataField, List, Options)} to handle extra columns.
      */
     default boolean supportsMultiColumn() {
         return false;
     }
 
-    default GlobalIndexer create(List<DataField> fields, Options options) {
-        if (fields.size() > 1) {
+    /**
+     * Creates an indexer over a primary column plus optional extra columns. {@code dataField} is
+     * the primary column; {@code extraFields} holds the remaining columns and is empty for a
+     * single-column index.
+     */
+    default GlobalIndexer create(
+            DataField dataField, List<DataField> extraFields, Options options) {
+        if (extraFields != null && !extraFields.isEmpty()) {
             throw new UnsupportedOperationException(
                     String.format(
-                            "Index type '%s' does not support multi-column index, got columns: %s",
-                            identifier(), fields));
+                            "Index type '%s' does not support multi-column index, got extra columns: %s",
+                            identifier(), extraFields));
         }
-        return create(fields.get(0), options);
+        return create(dataField, options);
     }
 }
--- a/paimon-core/src/main/java/org/apache/paimon/globalindex/GlobalIndexBuilderUtils.java
+++ b/paimon-core/src/main/java/org/apache/paimon/globalindex/GlobalIndexBuilderUtils.java
@@ -29,6 +29,8 @@ import org.apache.paimon.table.FileStoreTable;
 import org.apache.paimon.types.DataField;
 import org.apache.paimon.utils.Range;
 
+import javax.annotation.Nullable;
+
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
@@ -45,12 +47,54 @@ public class GlobalIndexBuilderUtils {
             String indexType,
             List<ResultEntry> entries)
             throws IOException {
+        return toIndexFileMetas(
+                fileIO, indexPathFactory, options, range, indexFieldId, null, indexType, entries);
+    }
+
+    public static List<IndexFileMeta> toIndexFileMetas(
+            FileIO fileIO,
+            IndexPathFactory indexPathFactory,
+            CoreOptions options,
+            Range range,
+            List<DataField> fields,
+            String indexType,
+            List<ResultEntry> entries)
+            throws IOException {
+        int indexFieldId = fields.get(0).id();
+        int[] extraFieldIds =
+                fields.size() > 1
+                        ? fields.subList(1, fields.size()).stream()
+                                .mapToInt(DataField::id)
+                                .toArray()
+                        : null;
+        return toIndexFileMetas(
+                fileIO,
+                indexPathFactory,
+                options,
+                range,
+                indexFieldId,
+                extraFieldIds,
+                indexType,
+                entries);
+    }
+
+    private static List<IndexFileMeta> toIndexFileMetas(
+            FileIO fileIO,
+            IndexPathFactory indexPathFactory,
+            CoreOptions options,
+            Range range,
+            int indexFieldId,
+            @Nullable int[] extraFieldIds,
+            String indexType,
+            List<ResultEntry> entries)
+            throws IOException {
         List<IndexFileMeta> results = new ArrayList<>();
         for (ResultEntry entry : entries) {
             String fileName = entry.fileName();
             long fileSize = fileIO.getFileSize(indexPathFactory.toPath(fileName));
             GlobalIndexMeta globalIndexMeta =
-                    new GlobalIndexMeta(range.from, range.to, indexFieldId, null, entry.meta());
+                    new GlobalIndexMeta(
+                            range.from, range.to, indexFieldId, extraFieldIds, entry.meta());
 
             Path externalPathDir = options.globalIndexExternalPath();
             String externalPathString = null;
@@ -75,6 +119,13 @@ public class GlobalIndexBuilderUtils {
             FileStoreTable table, String indexType, DataField indexField, Options options)
             throws IOException {
         GlobalIndexer globalIndexer = GlobalIndexer.create(indexType, indexField, options);
+        return globalIndexer.createWriter(createGlobalIndexFileReadWrite(table));
+    }
+
+    public static GlobalIndexWriter createIndexWriter(
+            FileStoreTable table, String indexType, List<DataField> fields, Options options)
+            throws IOException {
+        GlobalIndexer globalIndexer = GlobalIndexer.create(indexType, fields, options);
         return globalIndexer.createWriter(createGlobalIndexFileReadWrite(table));
     }
 
--- a/paimon-core/src/main/java/org/apache/paimon/globalindex/GlobalIndexBuilderUtils.java
+++ b/paimon-core/src/main/java/org/apache/paimon/globalindex/GlobalIndexBuilderUtils.java
@@ -31,6 +31,9 @@ import org.apache.paimon.table.FileStoreTable;
 import org.apache.paimon.types.DataField;
 import org.apache.paimon.utils.Range;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import javax.annotation.Nullable;
 
 import java.io.IOException;
@@ -41,6 +44,8 @@ import java.util.Map;
 
 /** Utils for global index build. */
 public class GlobalIndexBuilderUtils {
+
+    private static final Logger LOG = LoggerFactory.getLogger(GlobalIndexBuilderUtils.class);
 
     public static final int MULTI_COLUMN_INDEX_FIELD_ID = -1;
 
@@ -148,6 +153,7 @@ public class GlobalIndexBuilderUtils {
             SchemaManager schemaManager, List<ManifestEntry> entries, List<String> indexColumns) {
         Map<Long, Boolean> schemaContainsColumns = new HashMap<>();
         long minRowId = Long.MAX_VALUE;
+        long minSchemaId = -1;
         for (ManifestEntry entry : entries) {
             long sid = entry.file().schemaId();
             boolean contains =
@@ -155,8 +161,26 @@ public class GlobalIndexBuilderUtils {
                             sid,
                             id -> schemaManager.schema(id).fieldNames().containsAll(indexColumns));
             if (!contains && entry.file().firstRowId() != null) {
-                minRowId = Math.min(minRowId, entry.file().nonNullFirstRowId());
+                long rowId = entry.file().nonNullFirstRowId();
+                if (rowId < minRowId) {
+                    minRowId = rowId;
+                    minSchemaId = sid;
+                }
             }
+        }
+        if (minRowId != Long.MAX_VALUE) {
+            List<String> schemaFields = schemaManager.schema(minSchemaId).fieldNames();
+            List<String> missingColumns = new ArrayList<>();
+            for (String col : indexColumns) {
+                if (!schemaFields.contains(col)) {
+                    missingColumns.add(col);
+                }
+            }
+            LOG.info(
+                    "Found non-indexable files: schemaId={} missing columns {}, boundaryRowId={}.",
+                    minSchemaId,
+                    missingColumns,
+                    minRowId);
         }
         return minRowId;
     }
@@ -174,6 +198,11 @@ public class GlobalIndexBuilderUtils {
                 result.add(entry);
             }
         }
+        LOG.info(
+                "Filtered {} files to {} indexable files (boundaryRowId={}).",
+                entries.size(),
+                result.size(),
+                boundaryRowId);
         return result;
     }
 
--- a/paimon-core/src/main/java/org/apache/paimon/globalindex/GlobalIndexBuilderUtils.java
+++ b/paimon-core/src/main/java/org/apache/paimon/globalindex/GlobalIndexBuilderUtils.java
@@ -38,6 +38,8 @@ import java.util.List;
 /** Utils for global index build. */
 public class GlobalIndexBuilderUtils {
 
+    public static final int MULTI_COLUMN_INDEX_FIELD_ID = -1;
+
     public static List<IndexFileMeta> toIndexFileMetas(
             FileIO fileIO,
             IndexPathFactory indexPathFactory,
@@ -60,13 +62,15 @@ public class GlobalIndexBuilderUtils {
             String indexType,
             List<ResultEntry> entries)
             throws IOException {
-        int indexFieldId = fields.get(0).id();
-        int[] extraFieldIds =
-                fields.size() > 1
-                        ? fields.subList(1, fields.size()).stream()
-                                .mapToInt(DataField::id)
-                                .toArray()
-                        : null;
+        int indexFieldId;
+        int[] extraFieldIds;
+        if (fields.size() > 1) {
+            indexFieldId = MULTI_COLUMN_INDEX_FIELD_ID;
+            extraFieldIds = fields.stream().mapToInt(DataField::id).toArray();
+        } else {
+            indexFieldId = fields.get(0).id();
+            extraFieldIds = null;
+        }
         return toIndexFileMetas(
                 fileIO,
                 indexPathFactory,
--- a/paimon-core/src/main/java/org/apache/paimon/globalindex/GlobalIndexBuilderUtils.java
+++ b/paimon-core/src/main/java/org/apache/paimon/globalindex/GlobalIndexBuilderUtils.java
@@ -142,12 +142,12 @@ public class GlobalIndexBuilderUtils {
     public static GlobalIndexWriter createIndexWriter(
             FileStoreTable table,
             String indexType,
-            DataField dataField,
+            DataField indexField,
             List<DataField> extraFields,
             Options options)
             throws IOException {
         GlobalIndexer globalIndexer =
-                GlobalIndexer.create(indexType, dataField, extraFields, options);
+                GlobalIndexer.create(indexType, indexField, extraFields, options);
         return globalIndexer.createWriter(createGlobalIndexFileReadWrite(table));
     }
 
--- a/paimon-core/src/main/java/org/apache/paimon/globalindex/GlobalIndexBuilderUtils.java
+++ b/paimon-core/src/main/java/org/apache/paimon/globalindex/GlobalIndexBuilderUtils.java
@@ -140,9 +140,14 @@ public class GlobalIndexBuilderUtils {
     }
 
     public static GlobalIndexWriter createIndexWriter(
-            FileStoreTable table, String indexType, List<DataField> fields, Options options)
+            FileStoreTable table,
+            String indexType,
+            DataField dataField,
+            List<DataField> extraFields,
+            Options options)
             throws IOException {
-        GlobalIndexer globalIndexer = GlobalIndexer.create(indexType, fields, options);
+        GlobalIndexer globalIndexer =
+                GlobalIndexer.create(indexType, dataField, extraFields, options);
         return globalIndexer.createWriter(createGlobalIndexFileReadWrite(table));
     }
 
--- a/paimon-core/src/main/java/org/apache/paimon/globalindex/GlobalIndexBuilderUtils.java
+++ b/paimon-core/src/main/java/org/apache/paimon/globalindex/GlobalIndexBuilderUtils.java
@@ -47,8 +47,6 @@ public class GlobalIndexBuilderUtils {
 
     private static final Logger LOG = LoggerFactory.getLogger(GlobalIndexBuilderUtils.class);
 
-    public static final int MULTI_COLUMN_INDEX_FIELD_ID = -1;
-
     public static List<IndexFileMeta> toIndexFileMetas(
             FileIO fileIO,
             IndexPathFactory indexPathFactory,
@@ -62,6 +60,12 @@ public class GlobalIndexBuilderUtils {
                 fileIO, indexPathFactory, options, range, indexFieldId, null, indexType, entries);
     }
 
+    /**
+     * Builds the index file metas. The first column in {@code fields} is treated as the primary
+     * index column (e.g. the first column in {@code CREATE ... INDEX ON (a, b, c)}) and is stored
+     * as {@code indexFieldId}; the remaining columns go into {@code extraFieldIds}. Callers must
+     * pass {@code fields} in the intended column order.
+     */
     public static List<IndexFileMeta> toIndexFileMetas(
             FileIO fileIO,
             IndexPathFactory indexPathFactory,
@@ -71,15 +75,15 @@ public class GlobalIndexBuilderUtils {
             String indexType,
             List<ResultEntry> entries)
             throws IOException {
-        int indexFieldId;
-        int[] extraFieldIds;
-        if (fields.size() > 1) {
-            indexFieldId = MULTI_COLUMN_INDEX_FIELD_ID;
-            extraFieldIds = fields.stream().mapToInt(DataField::id).toArray();
-        } else {
-            indexFieldId = fields.get(0).id();
-            extraFieldIds = null;
-        }
+        // The first column is the primary index column and is stored as indexFieldId; the
+        // remaining columns (if any) go into extraFieldIds.
+        int indexFieldId = fields.get(0).id();
+        int[] extraFieldIds =
+                fields.size() > 1
+                        ? fields.subList(1, fields.size()).stream()
+                                .mapToInt(DataField::id)
+                                .toArray()
+                        : null;
         return toIndexFileMetas(
                 fileIO,
                 indexPathFactory,
--- a/paimon-core/src/main/java/org/apache/paimon/globalindex/GlobalIndexBuilderUtils.java
+++ b/paimon-core/src/main/java/org/apache/paimon/globalindex/GlobalIndexBuilderUtils.java
@@ -24,7 +24,9 @@ import org.apache.paimon.fs.Path;
 import org.apache.paimon.index.GlobalIndexMeta;
 import org.apache.paimon.index.IndexFileMeta;
 import org.apache.paimon.index.IndexPathFactory;
+import org.apache.paimon.manifest.ManifestEntry;
 import org.apache.paimon.options.Options;
+import org.apache.paimon.schema.SchemaManager;
 import org.apache.paimon.table.FileStoreTable;
 import org.apache.paimon.types.DataField;
 import org.apache.paimon.utils.Range;
@@ -33,7 +35,9 @@ import javax.annotation.Nullable;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 /** Utils for global index build. */
 public class GlobalIndexBuilderUtils {
@@ -131,6 +135,46 @@ public class GlobalIndexBuilderUtils {
             throws IOException {
         GlobalIndexer globalIndexer = GlobalIndexer.create(indexType, fields, options);
         return globalIndexer.createWriter(createGlobalIndexFileReadWrite(table));
+    }
+
+    /**
+     * Find the minimum firstRowId among files whose schema does not contain all index columns.
+     * Files at or beyond this rowId cannot be indexed because the column was added later via ALTER
+     * TABLE.
+     *
+     * @return the boundary rowId, or {@link Long#MAX_VALUE} if all files contain the columns
+     */
+    public static long findMinNonIndexableRowId(
+            SchemaManager schemaManager, List<ManifestEntry> entries, List<String> indexColumns) {
+        Map<Long, Boolean> schemaContainsColumns = new HashMap<>();
+        long minRowId = Long.MAX_VALUE;
+        for (ManifestEntry entry : entries) {
+            long sid = entry.file().schemaId();
+            boolean contains =
+                    schemaContainsColumns.computeIfAbsent(
+                            sid,
+                            id -> schemaManager.schema(id).fieldNames().containsAll(indexColumns));
+            if (!contains && entry.file().firstRowId() != null) {
+                minRowId = Math.min(minRowId, entry.file().nonNullFirstRowId());
+            }
+        }
+        return minRowId;
+    }
+
+    /** Keep only entries whose firstRowId is strictly less than the given boundary. */
+    public static List<ManifestEntry> filterEntriesBefore(
+            List<ManifestEntry> entries, long boundaryRowId) {
+        if (boundaryRowId == Long.MAX_VALUE) {
+            return entries;
+        }
+        List<ManifestEntry> result = new ArrayList<>();
+        for (ManifestEntry entry : entries) {
+            if (entry.file().firstRowId() != null
+                    && entry.file().nonNullFirstRowId() < boundaryRowId) {
+                result.add(entry);
+            }
+        }
+        return result;
     }
 
     private static GlobalIndexFileReadWrite createGlobalIndexFileReadWrite(FileStoreTable table) {
--- a/paimon-core/src/main/java/org/apache/paimon/globalindex/GlobalIndexScanner.java
+++ b/paimon-core/src/main/java/org/apache/paimon/globalindex/GlobalIndexScanner.java
@@ -20,6 +20,7 @@ package org.apache.paimon.globalindex;
 
 import org.apache.paimon.fs.FileIO;
 import org.apache.paimon.fs.Path;
+import org.apache.paimon.globalindex.GlobalIndexBuilderUtils;
 import org.apache.paimon.globalindex.io.GlobalIndexFileReader;
 import org.apache.paimon.index.GlobalIndexMeta;
 import org.apache.paimon.index.IndexFileMeta;
@@ -37,6 +38,7 @@ import org.apache.paimon.utils.Range;
 import java.io.Closeable;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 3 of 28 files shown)


JingsongLi (commented):
## Review: [globalindex] Support multi-column GlobalIndex framework

Overall this is a well-structured change that cleanly extends the single-column SPI to support multi-column indexes. The backward compatibility is handled well via default methods. I have a few concerns about correctness and robustness:

---

### 1. [Bug] Flink `BuildIndexOperator`: Multi-column writer receives row with extra `_ROW_ID` field

**File:** `paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/globalindex/GenericIndexTopoBuilder.java`

The projected row type is built as:
```java
List<String> readColumns = new ArrayList<>(indexColumns);
readColumns.add(SpecialFields.ROW_ID.name());
RowType projectedRowType = SpecialFields.rowTypeWithRowId(rowType).project(readColumns);
```

So the `InternalRow` read from data has the layout `[indexCol1, indexCol2, ..., _ROW_ID]`. In the multi-column branch:
```java
if (multiColumn) {
    ((GlobalIndexMultiColumnWriter) indexWriter).write(row);
}
```

The entire row (including the trailing `_ROW_ID` field) is passed to the writer. But the writer was created with only the index fields via `createIndexWriter(table, indexType, indexFields, mergedOptions)`, and the Javadoc on `GlobalIndexMultiColumnWriter.write()` states: *"The row layout matches the fields order passed to GlobalIndexerFactory#create(List, Options)"*.

This is a contract violation. Writer implementations that iterate over all row fields or check `row.getFieldCount()` will see an unexpected extra column. You should either:
- Create a sub-projection of the row that excludes `_ROW_ID` before passing it to the writer, or
- Clearly document that the row may contain trailing fields beyond the indexed columns (and update the Javadoc accordingly).

---

### 2. [Correctness] No null-field handling in multi-column mode (Flink)

In single-column mode, if the indexed field is null, the shard stops early:
```java
Object fieldData = indexFieldGetters[0].getFieldOrNull(row);
if (fieldData == null) {
    LOG.info("Null value at rowId={}, stopping shard [{}, {}].", ...);
    break;
}
```

In multi-column mode, the row is passed directly without any null check on individual fields. If one of the indexed columns is null in a multi-column row, the behavior depends entirely on the writer implementation. This asymmetry could lead to:
- Writer failures (NPE inside Lucene, for example)
- Silent corruption of the index

Consider at minimum documenting this contract difference, or adding validation that checks indexed columns for null before passing to the multi-column writer.

---

### 3. [Robustness] Unsafe cast without `instanceof` check

**Files:** `GenericIndexTopoBuilder.java` (Flink), `DefaultGlobalIndexBuilder.java` (Spark)

Both paths cast the writer based solely on the `multiColumn` flag:
```java
((GlobalIndexMultiColumnWriter) indexWriter).write(row);
```

However, `GlobalIndexerFactory.create(List<DataField>, Options)` has a default implementation that falls back to single-column:
```java
default GlobalIndexer create(List<DataField> fields, Options options) {
    return create(fields.get(0), options);
}
```

If an existing factory has not been updated to support multi-column (and relies on this default), it will return a `GlobalIndexSingletonWriter`, and the cast will fail with a `ClassCastException` at runtime.

Suggestion: Add a validation check after writer creation:
```java
if (multiColumn && !(indexWriter instanceof GlobalIndexMultiColumnWriter)) {
        "Index type '" + indexType + "' does not support multi-column indexing. " +
        "The factory must override create(List<DataField>, Options) and return a GlobalIndexMultiColumnWriter.");
}
```

---

### 4. [Design] Interface default method creates silent data-loss path

**File:** `paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/globalindex/GlobalIndexTopologyBuilder.java`

The new default method on the interface:
```java
default List<CommitMessage> buildIndex(..., List<DataField> indexFields, ...) {
    return buildIndex(..., indexFields.get(0), options);
}
```

This silently discards all fields beyond the first for any implementation that has not been updated to override the multi-field method. Combined with `DefaultGlobalIndexTopoBuilder` overriding the single-field method to delegate to multi-field, this creates an unusual delegation pattern that is correct for `DefaultGlobalIndexTopoBuilder` but could be a trap for other implementations.

Consider adding a log warning in the default method when `indexFields.size() > 1`, or throwing `UnsupportedOperationException` instead of silently dropping fields.

---

### 5. [Minor] `resolveFields` assumes metadata consistency across range groups

**File:** `paimon-core/src/main/java/org/apache/paimon/globalindex/GlobalIndexScanner.java`

```java
private List<DataField> resolveFields(Map<Range, List<IndexFileMeta>> metas, RowType rowType) {
    GlobalIndexMeta firstMeta =
            checkNotNull(metas.values().iterator().next().get(0).globalIndexMeta());
    ...
}
```

This takes the first index file's metadata from an arbitrary range to determine the field list. If different ranges have inconsistent metadata (e.g., during a transition where an old single-column index coexists with a new multi-column index for the same type), this could resolve incorrect fields. A defensive check that all ranges share the same field list would prevent hard-to-diagnose query errors.

---

### Summary

The most critical issue is #1 (the `_ROW_ID` field leaking into the multi-column writer). Issues #2 and #3 are important for production robustness. Issues #4 and #5 are lower priority but worth addressing for long-term maintainability.

CrownChu:
### Fixes and Additions in This Round
  
  #### 1. [Bug] Flink BuildIndexOperator: Multi-column writer receives row with extra _ROW_ID field

  **Fix:** The read projection includes `indexColumns + _ROW_ID` (_ROW_ID is needed for shard boundary positioning). Before passing to the writer, a secondary projection via
  `ProjectedRow.from(indexOnlyMapping)` strips the trailing `_ROW_ID`, ensuring `GlobalIndexMultiColumnWriter.write()` receives a row containing only the index fields.

  #### 2. [Correctness] No null-field handling in multi-column mode (Flink)

  **Fix:** In multi-column mode, each index field is checked individually. If any field is null, the current shard stops writing immediately (break). Only rows where all indexed columns are non-null are
  written to the index.

  #### 3. [Robustness] Unsafe cast without instanceof check

  **Fix:** Added `instanceof GlobalIndexMultiColumnWriter` check before casting. If the check fails, an `UnsupportedOperationException` is thrown with a clear message indicating the factory must override
  `create(List<DataField>, Options)` and return a `GlobalIndexMultiColumnWriter`.

  #### 4. [Design] Interface default method creates silent data-loss path

  **Fix:** Runtime instanceof check + exception ensures that when a factory does not properly implement multi-column support, it fails fast rather than silently falling through to SingletonWriter and losing
  data.

  #### 5. [Minor] resolveFields assumes metadata consistency across range groups

  **Fix:** For multi-column indexes (`indexFieldId == -1`), added validation that iterates all entries across range groups and verifies `indexFieldId` and `extraFieldIds` are consistent. Throws an exception on
   mismatch. Single-column case retains the original logic unchanged.

  #### 6. Multi-column minimum rowId for schema evolution

  `findMinNonIndexableRowId` accepts `List<String> indexColumns` and uses `containsAll(indexColumns)` to check. If any index column is missing from a file's schema, that file is considered non-indexable and
  its `firstRowId` is used as the filter boundary.

  #### 7. Multi-column GlobalIndexMeta storage convention

  In multi-column mode, `indexFieldId = -1` (`MULTI_COLUMN_INDEX_FIELD_ID` constant) and all actual field IDs are stored in `extraFieldIds`. Single-column remains unchanged (`indexFieldId` = actual field ID,
  `extraFieldIds = null`). Scanner skips registering under -1 and routes predicates to the correct reader solely through `extraFieldIds`.

  #### 8. New ES Topology Builder (es-multi-index)

  - Added `ESIndexTopoBuilder` (Flink) — standalone topology with shard-level parallelism, schema evolution check, and transparent indexType pass-through
  - Added `ESGlobalIndexTopoBuilder` (Spark) — registered via SPI, `identifier()` returns `"es-multi-index"`
  - Flink `CreateGlobalIndexProcedure` routing: `startsWith("es-multi-index")` → `ESIndexTopoBuilder`
  - Spark `GlobalIndexTopologyBuilderUtils`: exact match first, then prefix match

  #### 9. Multi-condition reader predicate push-down

  - `GlobalIndexEvaluator` refactored to async parallel execution (`CompletableFuture`), no longer implements `PredicateVisitor`
  - Added `flattenChildren()` to flatten nested CompoundPredicates of the same type
  - `UnionGlobalIndexReader` removed internal `ExecutorService`; parallelism managed by the upper-level evaluator
  - `GlobalIndexScanner` registers `extraFieldIds` into the index map, ensuring multi-column predicates hit the correct reader

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/globalindex/GenericIndexTopoBuilder.java
```diff
@@ -626,16 +693,20 @@ public void processElement(StreamRecord<ShardTask> element) throws Exception {
...
-                            Object fieldData = indexFieldGetter.getFieldOrNull(row);
-                            if (fieldData == null) {
-                                LOG.info(
-                                        "Null vector at rowId={}, stopping shard [{}, {}].",
-                                        currentRowId,
-                                        task.shardRange.from,
-                                        task.shardRange.to);
-                                break;
+                            if (multiColumn) {
+                                ((GlobalIndexMultiColumnWriter) indexWriter).write(row);
```
jerry-024: Bug: missing null check in multi-column path.

The single-column path below checks for null and breaks the shard, and `ESIndexTopoBuilder.BuildESIndexOperator.processElement()` in this same PR also checks all columns for null before writing. But this multi-column path writes `row` directly with no null check.

Known implementations like `LuminaVectorGlobalIndexWriter.write()` throw `IllegalArgumentException` on null input — so a multi-column index containing a vector column will crash the Flink job if any row has a null value in the indexed columns.

Suggested fix — add the same null-field guard that `ESIndexTopoBuilder` has:

```java
if (multiColumn) {
    boolean hasNull = false;
    for (InternalRow.FieldGetter getter : indexFieldGetters) {
        if (getter.getFieldOrNull(row) == null) {
            hasNull = true;
            break;
        }
    }
    if (hasNull) {
        LOG.info(
                "Null value in indexed columns at rowId={}, stopping shard [{}, {}].",
                currentRowId,
                task.shardRange.from,
                task.shardRange.to);
        break;
    }
    ((GlobalIndexMultiColumnWriter) indexWriter).write(row);
}
```

CrownChu: fixed

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/globalindex/ESIndexTopoBuilder.java
```diff
@@ -0,0 +1,667 @@
...
+import static org.apache.paimon.io.DataIncrement.indexIncrement;
+import static org.apache.paimon.utils.Preconditions.checkArgument;
+
+/**
+ * Builds a Flink topology for creating ES-based global indexes with parallelism. Follows the same
+ * shard-based approach as {@link GenericIndexTopoBuilder} but specialized for ES vector indexes
+ * with multi-column support.
+ */
+public class ESIndexTopoBuilder {
+
```
jerry-024: This entire file is an ES-specific topology builder, unrelated to the multi-column GlobalIndex SPI framework. Consider moving it to a separate PR to keep this one focused on the framework changes.

Also, `findMinNonIndexableRowId`, `filterEntriesBefore`, `computeShardTasks`, and `closeWriterQuietly` are copy-pasted from `GenericIndexTopoBuilder` — consider extracting them into a shared utility (e.g. `GlobalIndexBuilderUtils`).

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/globalindex/ESGlobalIndexTopoBuilder.java
```diff
@@ -0,0 +1,341 @@
...
+
+import static org.apache.paimon.CoreOptions.GLOBAL_INDEX_ROW_COUNT_PER_SHARD;
+import static org.apache.paimon.utils.Preconditions.checkArgument;
+
+/**
+ * ES-specific topology builder for Spark. Builds es-vector-diskbbq global indexes using Spark's
+ * parallelism. Supports both single-column and multi-column indexing.
+ */
+public class ESGlobalIndexTopoBuilder implements GlobalIndexTopologyBuilder {
+
```
jerry-024: Same as `ESIndexTopoBuilder` on the Flink side — this ES-specific Spark topology builder is not part of the multi-column framework and should go in a separate PR.

`findMinNonIndexableRowId` and `filterEntriesBefore` are again duplicated here (third copy). Please extract into a shared utility.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/procedure/CreateGlobalIndexProcedure.java
```diff
@@ -97,26 +107,34 @@ public String[] call(
...
                         procedureContext.getExecutionEnvironment(),
                         table,
-                        indexColumn,
+                        indexColumns.get(0),
                         partitionPredicate,
                         userOptions);
                 return new String[] {
                     "BTree global index created successfully for table: " + table.name()
                 };
+            } else if (indexType.startsWith(ESIndexTopoBuilder.ES_INDEX_TYPE_PREFIX)) {
```
jerry-024: This ES routing branch (`ESIndexTopoBuilder.ES_INDEX_TYPE_PREFIX`) is ES-specific and not related to multi-column support. Consider moving it to the ES-specific PR.

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/globalindex/GlobalIndexTopologyBuilderUtils.java
```diff
@@ -54,7 +54,17 @@ public class GlobalIndexTopologyBuilderUtils {
     }
 
     public static GlobalIndexTopologyBuilder createTopoBuilder(String indexType) {
+        // Exact match first
         GlobalIndexTopologyBuilder builder = FACTORIES.get(indexType);
-        return builder == null ? new DefaultGlobalIndexTopoBuilder() : builder;
+        if (builder != null) {
+            return builder;
+        }
+        // Prefix match: e.g. "es-multi-index-diskbbq" matches registered "es-multi-index"
```
jerry-024: The prefix matching logic here is ES-specific — the comment even says `"es-multi-index-diskbbq" matches registered "es-multi-index"`. Consider moving this to the ES-specific PR.

paimon-spark/paimon-spark-common/src/main/resources/META-INF/services/org.apache.paimon.spark.globalindex.GlobalIndexTopologyBuilder
```diff
@@ -14,3 +14,4 @@
 # limitations under the License.
 
 org.apache.paimon.spark.globalindex.btree.BTreeIndexTopoBuilder
+org.apache.paimon.spark.globalindex.ESGlobalIndexTopoBuilder
```
jerry-024: Registration of `ESGlobalIndexTopoBuilder` should go with the ES-specific PR.

jerry-024 (commented):
The multi-column GlobalIndex SPI framework changes look good. However, the ES-specific code (ESIndexTopoBuilder, ESGlobalIndexTopoBuilder, ES routing in procedures, prefix matching in utils, and SPI registration) is a separate feature that happens to use the multi-column framework. Suggest splitting it into its own PR to keep this one focused.

Also, `findMinNonIndexableRowId` and `filterEntriesBefore` are copy-pasted across `GenericIndexTopoBuilder`, `ESIndexTopoBuilder`, and `ESGlobalIndexTopoBuilder` — consider extracting them into a shared utility.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/globalindex/ESIndexTopoBuilder.java
CrownChu: done

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/globalindex/ESGlobalIndexTopoBuilder.java
CrownChu: done

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/procedure/CreateGlobalIndexProcedure.java
CrownChu: done

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/globalindex/GlobalIndexTopologyBuilderUtils.java
CrownChu: done

paimon-spark/paimon-spark-common/src/main/resources/META-INF/services/org.apache.paimon.spark.globalindex.GlobalIndexTopologyBuilder
CrownChu: done

jerry-024 (commented):
Thanks for splitting out the ES code and extracting the shared utilities — much cleaner now. Found a few issues in the latest version:

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/globalindex/GenericIndexTopoBuilderTest.java
```diff
@@ -475,7 +475,7 @@ void testAppendFilterOldFilesBeforeNewFiles() {
                 GenericIndexTopoBuilder.filterEntriesBefore(
                         entries,
                         GenericIndexTopoBuilder.findMinNonIndexableRowId(
```
jerry-024: **Compile error**: `findMinNonIndexableRowId` and `filterEntriesBefore` were moved from `GenericIndexTopoBuilder` to `GlobalIndexBuilderUtils` in commit 0cfc7ef, but this test still references them via `GenericIndexTopoBuilder.findMinNonIndexableRowId(...)` and `GenericIndexTopoBuilder.filterEntriesBefore(...)`. This will fail to compile.

Should be:
```java
GlobalIndexBuilderUtils.filterEntriesBefore(
    entries,
    GlobalIndexBuilderUtils.findMinNonIndexableRowId(
        schemaManager, entries, Collections.singletonList("vec")));
```

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/globalindex/GenericIndexTopoBuilder.java
```diff
@@ -626,16 +648,35 @@ public void processElement(StreamRecord<ShardTask> element) throws Exception {
...
+                                            currentRowId,
+                                            task.shardRange.from,
+                                            task.shardRange.to);
+                                    break;
+                                }
+                                ((GlobalIndexMultiColumnWriter) indexWriter).write(row);
+                            } else {
+                                Object fieldData = indexFieldGetters[0].getFieldOrNull(row);
+                                if (fieldData == null) {
+                                    LOG.info(
```
jerry-024: **Multi-column writer receives extra `_ROW_ID` column**: In the multi-column path, `row` is passed directly to `GlobalIndexMultiColumnWriter.write(row)`, but this row comes from `projectedRowType` which is `indexColumns + _ROW_ID`. The writer's contract (javadoc on `GlobalIndexMultiColumnWriter.write`) says the row layout should match the fields passed to `GlobalIndexerFactory.create(List<DataField>, Options)` — which doesn't include `_ROW_ID`.

The previous ES-specific code handled this with a `ProjectedRow` that stripped `_ROW_ID` before writing. Consider adding a similar projection here, or clarifying the writer contract.

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/globalindex/DefaultGlobalIndexBuilder.java
```diff
@@ -99,27 +120,38 @@ public CommitMessage build(CloseableIterator<InternalRow> data) throws IOExcepti
...
-                    row -> {
-                        Object indexO = getter.getFieldOrNull(row);
-                        indexWriter.write(indexO);
-                        rowCounter.add(1);
-                    });
+            if (multiColumn) {
+                GlobalIndexMultiColumnWriter multiWriter =
+                        (GlobalIndexMultiColumnWriter) indexWriter;
+                rows.forEachRemaining(
+                        row -> {
```
jerry-024: Same `_ROW_ID` issue as the Flink side: `rows` come from a reader using `readType = indexColumns + _ROW_ID`, but the multi-column writer expects only index columns. The row passed to `multiWriter.write(row)` includes the extra `_ROW_ID` field.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/procedure/CreateGlobalIndexProcedure.java
```diff
@@ -97,7 +106,7 @@ public String[] call(
                 BTreeIndexTopoBuilder.buildIndexAndExecute(
                         procedureContext.getExecutionEnvironment(),
                         table,
```
jerry-024, line 140: BTree silently drops extra columns: when a user passes `"col1,col2"` with index type `btree`, only `indexColumns.get(0)` is used — no error, no warning. Consider adding a validation:
```java
if ("btree".equalsIgnoreCase(indexType)) {
    checkArgument(indexColumns.size() == 1,
        "BTree index only supports single column, got: %s", indexColumns);
}
```

paimon-core/src/main/java/org/apache/paimon/globalindex/GlobalIndexBuilderUtils.java
```diff
@@ -78,6 +130,53 @@ public static GlobalIndexWriter createIndexWriter(
...
+                    schemaContainsColumns.computeIfAbsent(
+                            sid,
+                            id -> schemaManager.schema(id).fieldNames().containsAll(indexColumns));
+            if (!contains && entry.file().firstRowId() != null) {
+                minRowId = Math.min(minRowId, entry.file().nonNullFirstRowId());
+            }
+        }
+        return minRowId;
+    }
+
```
jerry-024, line 196: Minor: the old `filterEntriesBefore` in `GenericIndexTopoBuilder` had a `LOG.info("Filtered {} files ...")` line for observability. This was lost during extraction since `GlobalIndexBuilderUtils` has no logger. Consider adding one — this log is useful for debugging index build issues in production.

paimon-core/src/main/java/org/apache/paimon/globalindex/GlobalIndexBuilderUtils.java
```diff
@@ -24,18 +24,26 @@
...
 import org.apache.paimon.types.DataField;
 import org.apache.paimon.utils.Range;
 
+import javax.annotation.Nullable;
+
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
```
jerry-024, line 43: **High risk — MERGE path crash**: `MULTI_COLUMN_INDEX_FIELD_ID = -1` breaks existing code that calls `rowType.getField(globalIndexMeta.indexFieldId())` without guarding against -1:

1. `MergeIntoUpdateChecker.java:104` (Flink): scans index manifest entries and does `rowType.getField(globalIndexMeta.indexFieldId())` — will throw when encountering a multi-column index.
2. `MergeIntoPaimonDataEvolutionTable.scala:514` (Spark): same pattern — `rowType.getField(globalIndexMeta.indexFieldId()).name()`.

Once a table has a multi-column global index, any `MERGE INTO` that touches indexed columns will crash with "Cannot find field by field id: -1".

Fix: these callers need to handle `MULTI_COLUMN_INDEX_FIELD_ID` by reading `extraFieldIds()` to get the actual column list.

paimon-core/src/main/java/org/apache/paimon/table/source/FullTextScanImpl.java
```diff
@@ -61,7 +61,17 @@ public Plan scan() {
...
-                    return textColumn.id() == globalIndex.indexFieldId();
+                    if (textColumn.id() == globalIndex.indexFieldId()) {
+                        return true;
+                    }
+                    if (globalIndex.extraFieldIds() != null) {
+                        for (int id : globalIndex.extraFieldIds()) {
+                            if (textColumn.id() == id) {
+                                return true;
+                            }
+                        }
```
jerry-024: **High risk — read path mismatch**: `FullTextScanImpl` now correctly selects multi-column index files via the `extraFieldIds` check added in this PR. However, the corresponding **read** path in `FullTextReadImpl.java:72-74` still creates the reader with:
```java
GlobalIndexerFactoryUtils.load(indexType).create(textColumn, options)
```

This uses the single-column factory method. When the selected index file was built with `create(List<DataField>[text, vector], options)`, reading it with a single-column reader will either fail to decode the file or produce incorrect results.

The scan/filter path and the read path are now inconsistent — scan discovers multi-column files, but read doesn't know how to open them. Need to use `resolveFields`-style metadata lookup + `create(List<DataField>, Options)` in the read path as well.

paimon-core/src/main/java/org/apache/paimon/table/source/VectorScanImpl.java
```diff
@@ -82,7 +82,17 @@ public Plan scan() {
...
                     int fieldId = globalIndex.indexFieldId();
-                    return vectorColumn.id() == fieldId || filterFieldIds.contains(fieldId);
+                    if (vectorColumn.id() == fieldId || filterFieldIds.contains(fieldId)) {
+                        return true;
+                    }
+                    if (globalIndex.extraFieldIds() != null) {
+                        for (int id : globalIndex.extraFieldIds()) {
+                            if (vectorColumn.id() == id || filterFieldIds.contains(id)) {
+                                return true;
+                            }
```
jerry-024, line 93: Same read path issue as FullText: `VectorReadImpl.java:87-89` creates the reader with `create(vectorColumn, options)` (single-column), but this scan filter now discovers multi-column index files. The read side needs the same multi-column awareness.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/globalindex/GenericIndexTopoBuilder.java
jerry-024: Still unresolved from last review: multi-column writer receives the full projected row including `_ROW_ID`. Need to project down to index-only columns before `write(row)`. See `ProjectedRow` approach.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/procedure/CreateGlobalIndexProcedure.java
jerry-024, line 140: Still unresolved: BTree and other index types that don't support multi-column will silently drop extra columns. When `indexColumns.size() > 1` and the factory doesn't override `create(List<DataField>, Options)`, the default implementation falls back to `fields.get(0)` without any error.

Suggest: throw `UnsupportedOperationException` in the default `GlobalIndexerFactory.create(List<DataField>, Options)` when `fields.size() > 1`, instead of silently falling back to `create(fields.get(0), options)`. This forces implementations to explicitly opt-in to multi-column support.

jerry-024 (commented):
Found additional high-risk issues in code paths NOT modified by this PR but broken by the introduction of `MULTI_COLUMN_INDEX_FIELD_ID = -1`:

paimon-flink/paimon-flink-common/src/test/java/org/apache/paimon/flink/globalindex/GenericIndexTopoBuilderTest.java
CrownChu: done

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/globalindex/GenericIndexTopoBuilder.java
CrownChu: fixed

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/globalindex/DefaultGlobalIndexBuilder.java
CrownChu: fixed

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/procedure/CreateGlobalIndexProcedure.java
CrownChu: fixed

paimon-core/src/main/java/org/apache/paimon/globalindex/GlobalIndexBuilderUtils.java
CrownChu: add log done

paimon-core/src/main/java/org/apache/paimon/globalindex/GlobalIndexBuilderUtils.java
CrownChu: Fix:

  Added getIndexedFieldNames helper in both Flink and Spark paths:
  - When indexFieldId == MULTI_COLUMN_INDEX_FIELD_ID (-1): resolve column names from extraFieldIds()
  - Otherwise: use the original single-column logic (rowType.getField(indexFieldId) + optional extraFieldIds)
  
  Both the index filter (which entries are affected) and the error reporting (conflicted column names) now correctly handle multi-column indexes.

  Affected files:
  - paimon-flink/.../dataevolution/MergeIntoUpdateChecker.java
  - paimon-spark/paimon-spark-common/.../MergeIntoPaimonDataEvolutionTable.scala
  - paimon-spark/paimon-spark-4.0/.../MergeIntoPaimonDataEvolutionTable.scala

paimon-core/src/main/java/org/apache/paimon/table/source/FullTextScanImpl.java
CrownChu: Before creating the GlobalIndexer, inspect the first index file's GlobalIndexMeta to determine the file's column layout:
  - indexFieldId == -1 (multi-column): resolve all field IDs from extraFieldIds(), create reader with factory.create(List<DataField>, options)
  - Otherwise (single-column): use original factory.create(textColumn, options)
  
  This ensures the reader matches the format used when the index file was built.

  Affected file:
  - paimon-core/src/main/java/org/apache/paimon/table/source/FullTextReadImpl.java

CrownChu: Same approach as FullTextReadImpl — inspect the first index file's GlobalIndexMeta before creating the reader:
  - indexFieldId == -1 (multi-column): resolve fields from extraFieldIds(), use factory.create(List<DataField>, options)
  - Otherwise (single-column): use original factory.create(vectorColumn, options)
  
  Affected file:
  - paimon-core/src/main/java/org/apache/paimon/table/source/VectorReadImpl.java

paimon-core/src/main/java/org/apache/paimon/table/source/VectorScanImpl.java
CrownChu: same as last comment,fixed

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/globalindex/GenericIndexTopoBuilder.java
CrownChu: commit no push，now already push

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/procedure/CreateGlobalIndexProcedure.java
CrownChu: done

jerry-024:
+1

paimon-core/src/main/java/org/apache/paimon/globalindex/GlobalIndexBuilderUtils.java
```diff
@@ -24,18 +24,31 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nullable;
+
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
```
leaves12138, line 43: **Risk — `IndexManifestFileHandler` overlap detection false positive with multi-column indexes**

`IndexManifestFileHandler.java:243` uses `indexFieldId` to determine whether two index files belong to the same field:
```java
retainedMeta.indexFieldId() != addedMeta.indexFieldId()
```

With `MULTI_COLUMN_INDEX_FIELD_ID = -1`, **all** multi-column indexes share the same sentinel value. Two indexes on different column sets (e.g. `[title, vec]` vs `[content, embedding]`) will both have `indexFieldId == -1`, so the handler treats them as "same field". If their row ranges overlap, it throws `IllegalStateException` and rejects the commit — even though they are logically independent indexes.

Suggestion: add `extraFieldIds` comparison (e.g. `Arrays.equals`) to the overlap check, or compare `indexType` as well.

leaves12138, line 43: **Minor — `TableIndexesTable` shows null for multi-column index field name**

`TableIndexesTable.java:238` does `logicalRowType.getField(globalMeta.indexFieldId())` which throws when `indexFieldId == -1`. The exception is caught, but `index_field_name` silently displays `null` to users.

Suggestion: when `indexFieldId == MULTI_COLUMN_INDEX_FIELD_ID`, resolve names from `extraFieldIds()` and join them with commas.

leaves12138, line 43: **Suggestion — add `isMultiColumn()` helper to `GlobalIndexMeta`**

The sentinel check `indexFieldId == MULTI_COLUMN_INDEX_FIELD_ID` is now scattered across many modules (MergeIntoUpdateChecker, MergeIntoPaimonDataEvolutionTable x2, FullTextReadImpl, VectorReadImpl, GlobalIndexScanner, GlobalIndexBuilderUtils, etc.). This is fragile — new code that touches `indexFieldId()` can easily forget the guard and crash on -1.

Consider adding a convenience method to `GlobalIndexMeta`:
```java
public boolean isMultiColumn() {
    return indexFieldId == MULTI_COLUMN_INDEX_FIELD_ID;
}
```
Then all call sites replace `meta.indexFieldId() == MULTI_COLUMN_INDEX_FIELD_ID` with `meta.isMultiColumn()`, which is more readable and harder to miss.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/dataevolution/MergeIntoUpdateChecker.java
```diff
@@ -159,4 +165,23 @@ private void checkUpdatedColumns() {
...
+        List<String> names = new ArrayList<>();
+        names.add(rowType.getField(fieldId).name());
+        if (meta.extraFieldIds() != null) {
+            for (int id : meta.extraFieldIds()) {
+                names.add(rowType.getField(id).name());
+            }
+        }
+        return names;
+    }
 }
```
leaves12138, line 164: **Code duplication — `getIndexedFieldNames` is copy-pasted 3 times**

This same helper logic exists in:
1. Here — `MergeIntoUpdateChecker.java` (Flink, Java)
2. `MergeIntoPaimonDataEvolutionTable.scala` (Spark common, Scala)
3. `MergeIntoPaimonDataEvolutionTable.scala` (Spark 4.0, Scala)

Consider extracting it as a static method in `GlobalIndexMeta` or `GlobalIndexBuilderUtils` so all three call sites can reuse a single implementation. This also reduces the risk of future inconsistencies when the logic evolves.

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/procedure/CreateGlobalIndexProcedure.java
```diff
@@ -170,7 +198,7 @@ public InternalRow[] call(InternalRow args) {
                     } catch (Exception e) {
```
leaves12138, line 221: Nit: this error message uses `column` (the raw comma-separated input string) instead of `indexColumns` (the parsed `List<String>`). The Flink procedure uses `indexColumns` in its error message — should be consistent.

```java
// current
String.format("Failed to create %s index for columns '%s' on table '%s'.", indexType, column, tableIdent)
// suggested
String.format("Failed to create %s index for columns '%s' on table '%s'.", indexType, indexColumns, tableIdent)
```

paimon-core/src/main/java/org/apache/paimon/globalindex/GlobalIndexBuilderUtils.java
CrownChu: Thanks for catching this!

Fixed in IndexManifestFileHandler.validateRetainedIndexFiles().
Split the overlap detection into two branches:
- Single-column: keep original indexFieldId comparison
- Multi-column (indexFieldId == -1): use Arrays.equals(extraFieldIds) to distinguish different column groups
This way two multi-column indexes on different column sets (e.g. [title, vec] vs [content, embedding]) won't trigger false positive overlap errors.

CrownChu: Thanks! Fixed in TableIndexesTable.toRow(). When indexFieldId == MULTI_COLUMN_INDEX_FIELD_ID, resolve field names from extraFieldIds() and join with commas (e.g. "title,vec"). Single-column path unchanged.

CrownChu: Done. Added GlobalIndexMeta.isMultiColumn() and replaced all sentinel checks across the following classes:
 - GlobalIndexScanner
 - FullTextReadImpl
 - VectorReadImpl
 - MergeIntoUpdateChecker
 - TableIndexesTable

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/dataevolution/MergeIntoUpdateChecker.java
CrownChu: Done. Extracted as an instance method GlobalIndexMeta.getIndexedFieldNames(RowType) and replaced all three call sites.
following classes:
GlobalIndexMeta、MergeIntoUpdateChecker、MergeIntoPaimonDataEvolutionTable.scala、MergeIntoPaimonDataEvolutionTable.scala

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/procedure/CreateGlobalIndexProcedure.java
CrownChu: fixed

JingsongLi:
I reviewed the latest change and found two issues that look worth fixing before merge:

- **[P1] Null values now stop the whole shard during generic index build.** In `GenericIndexTopoBuilder`, both the single-column path and multi-column path `break` when an indexed value is null. The Spark multi-column path does the same in `DefaultGlobalIndexBuilder`. Existing real writers such as Tantivy/Lumina treat null as “advance logical row id but do not add an indexed value”, so later non-null rows in the same shard still get indexed. With the new `break`, any rows after the first null in that shard are missing from the global index, which can make search/filter results incomplete. This should probably keep writing through the shard and let the writer preserve row-id alignment, or introduce a multi-column writer contract that can skip one logical row without ending the shard.

- **[P2] `GlobalIndexScanner` only maps each field id to one multi-column group.** The PR allows creating multi-column indexes like `(a,b)` and `(a,c)`, but scanner construction stores a single `fieldId -> group` mapping. If the same field participates in multiple multi-column indexes, it can either throw `Inconsistent extraFieldIds across index files` or overwrite the previous group, so predicate pushdown/pre-filtering may miss one set of readers. This should be modeled as `fieldId -> multiple groups/readers` and merged with any single-column readers for the field.

JingsongLi (changes requested):
Our current design is a primary secondary concept, requiring a primary field to indicate the ownership of the index, followed by auxiliary fields to assist with the index. It is best not to change the current design.
It can be seen that current APIs such as GlobalIndexMeta, VectorSearchBuilder, FullTextSearchBuilder, etc. are designed around this concept.

JingsongLi (commented):
I found one issue that should be addressed before this is considered ready.

`create_global_index` now accepts multi-column global indexes for every index type except `btree`, and then passes the multi-column `indexFields` into the topology builder. However, the default `GlobalIndexerFactory#create(List<DataField>, Options)` still throws `UnsupportedOperationException` when more than one field is provided, and the existing real factories such as `lumina` and `tantivy-fulltext` only implement the single-column `create(DataField, Options)` method.

That means a user can submit a multi-column `lumina` / `tantivy-fulltext` global index creation request, but the job will fail later at runtime when `DefaultGlobalIndexBuilder` calls `createIndexWriter(..., indexFields, ...)` and the factory rejects the list. The procedure should either reject unsupported multi-column index types up front, or the intended index types need to implement the multi-column factory/writer path fully.

JingsongLi (commented):
Thanks for pushing this forward. I took another careful pass and found a few correctness issues that I think should be fixed before merge:

1. `GlobalIndexScanner` can drop results when a filtered field participates as an extra column in multiple multi-column index groups. `extraIndexMetas` is a `Map<Integer, List<IndexMetaFileGroup>>`, but the lookup path currently selects only `extraGroups.get(0)`. If a table has indexes such as `(a, b)` and `(c, b)` covering different ranges/partitions, a predicate on `b` will read only the first group. Since `DataEvolutionBatchScan` converts the returned bitmap directly into row-range pushdown, this is not just a missed optimization: rows covered only by the skipped group can disappear from query results. Please union readers/results from all matching groups, or reject such metadata consistently.

2. `VectorScanImpl` still matches scalar pre-filter indexes only by `globalIndex.indexFieldId()`. Normal global-index scan now checks `extraFieldIds`, but vector search split planning does not. A vector query with a pushed filter on an extra column of a multi-column scalar index will not attach that scalar index, so top-k is computed over unfiltered rows; after the remaining table filter is applied, the query may return fewer or wrong top-k rows for the filtered dataset. The manifest filter and the vector/scalar classification should use the same “contains field” logic as the general scanner.

3. The drop procedures still expose `index_column` as a single column and match only `indexFieldId()`. After creating an index with `index_column => 'a,b'`, dropping with `'a,b'` fails validation, and dropping with `'b'` reports no index found; only dropping by the primary column works. That is surprising for a multi-column index and makes lifecycle management inconsistent with creation/display. Please either parse the same comma-separated column list and match the full indexed field set, or document/enforce that drop is by primary column only.

paimon-common/src/main/java/org/apache/paimon/globalindex/GlobalIndexerFactory.java
```diff
@@ -22,10 +22,30 @@
...
+
+    /**
+     * Whether this index type supports multi-column indexes. A factory that returns {@code true}
+     * must override {@link #create(List, Options)} to handle more than one column.
+     */
+    default boolean supportsMultiColumn() {
+        return false;
+    }
+
+    default GlobalIndexer create(List<DataField> fields, Options options) {
```
JingsongLi: Can you modify these methods to `extraFields`? Just like:
`default GlobalIndexer create(DataField dataField, List<DataField> extraFields, Options options)`

CrownChu: Done. Refactored all create / createIndexWriter methods in the call chain to use explicit (DataField dataField, List<DataField> extraFields, Options options) signature, pushed in bfc87104a.

JingsongLi:
Can you adjust **all** methods interfaces to `indexField` and `extraFields`? All interfaces should match this just like `GlobalIndexMeta`.

paimon-common/src/main/java/org/apache/paimon/globalindex/GlobalIndexerFactory.java
```diff
@@ -22,10 +22,36 @@
...
     String identifier();
 
-    GlobalIndexer create(DataField dataField, Options options);
+    GlobalIndexer create(DataField indexField, Options options);
+
+    /**
+     * Whether this index type supports multi-column indexes. A factory that returns {@code true}
+     * must override {@link #create(DataField, List, Options)} to handle extra columns.
+     */
+    default boolean supportsMultiColumn() {
```
JingsongLi: Remove this, I saw this is just for checkArguement.

paimon-core/src/main/java/org/apache/paimon/index/GlobalIndexMeta.java
```diff
@@ -78,6 +80,15 @@ public int indexFieldId() {
         return indexFieldId;
     }
 
+    /**
+     * Whether this index covers more than one column. {@link #indexFieldId} is always the primary
+     * column; {@link #extraFieldIds} holds the remaining columns and is null/empty for a
+     * single-column index.
+     */
+    public boolean isMultiColumn() {
```
JingsongLi: Remove this, you can just use `extraFieldIds != null`

paimon-core/src/main/java/org/apache/paimon/globalindex/GlobalIndexScanner.java
```diff
@@ -74,29 +75,86 @@ public GlobalIndexScanner(
...
+                    IndexMetaFileGroup group = indexMetas.get(fId);
+                    if (group != null) {
+                        return createReaders(indexFileReader, group, rowType);
+                    }
+                    List<IndexMetaFileGroup> extraGroups = extraIndexMetas.get(fId);
+                    if (extraGroups == null || extraGroups.isEmpty()) {
+                        return Collections.emptyList();
+                    }
+                    // Union readers from all groups that share this extra column
+                    List<GlobalIndexReader> allReaders = new ArrayList<>();
```
leaves12138, line 121: This still does not actually union the extra-column groups. `GlobalIndexEvaluator.visitLeafAsync` combines every reader returned for the same leaf predicate with `and(...)`, while `UnionGlobalIndexReader` is the layer that performs the OR/union. So returning one reader per extra group here can turn `(a, b)` and `(c, b)` into `result(a,b) AND result(c,b)` for a predicate on `b`, which may produce an empty/partial bitmap and then `DataEvolutionBatchScan` pushes that bitmap down as row ranges. We should either wrap the readers from all `extraGroups` into a single `UnionGlobalIndexReader`, or otherwise make the evaluator combine these alternative groups with OR instead of AND.

leaves12138 (approved):
+1

PR #7936, opened 2026-05-22
[core] Fix blob-only multi-field data evolution read

leaves12138:
### Purpose

Fix a data evolution read failure when a projection contains only multiple blob/vector-store fields and no normal field.

In this case, there is no normal data file in the split, so the expected row count remains unknown (`-1`). The special-field bunch should not validate its accumulated row count against this sentinel value.

### Changes

- Skip the special-field row-count upper-bound check when the expected row count is unknown.
- Add a regression test for reading multiple blob fields without normal fields.

### Tests

- `mvn -pl paimon-core -Dtest=DataEvolutionReadTest test`


JingsongLi:
+1

PR #7939, opened 2026-05-23
[core] Fix chain table sequence field validation

QuakeWang:
### Purpose

`CoreOptions.sequenceField()` returns an empty list when `sequence.field` is not configured, but chain table validation only checked it against `null`.

This allowed tables with `'chain-table.enabled' = 'true'` to pass schema validation without `sequence.field`, while chain table requires sequence semantics.

### Tests

- `mvn -pl paimon-core -Pfast-build -Dtest=SchemaValidationTest test`


JingsongLi (approved):
+1

PR #7941, opened 2026-05-23
[python] Add rollback_to_timestamp to FileStoreTable

JunRuiLee:
### Purpose

  Add `rollback_to_timestamp` to `FileStoreTable`, allowing users to rollback a table to the latest snapshot with commit time <= a given timestamp in milliseconds.

  **Note:** The timestamp lookup (`earlier_or_equal_time_mills`) uses a binary search that breaks on exact match. If multiple snapshots share the same millisecond, it may not return the highest matching snapshot ID. This matches Java's
  `earlierOrEqualTimeMills` behavior — a coordinated fix across both languages is a follow-up.

  ### Tests

  - `test_table_rollback_to_timestamp` — writes 5 commits, rolls back using snapshot 3's timestamp, verifies snapshots 4 and 5 are removed.
  - `test_table_rollback_to_timestamp_no_match` — verifies `ValueError` is raised when timestamp is before the earliest snapshot.

### DIFF
--- a/paimon-python/pypaimon/table/file_store_table.py
+++ b/paimon-python/pypaimon/table/file_store_table.py
@@ -298,6 +298,23 @@ class FileStoreTable(Table):
         return RollbackHelper(
             self.snapshot_manager(), self.tag_manager(), self.file_io)
 
+    def rollback_to_timestamp(self, timestamp_millis: int) -> None:
+        """Rollback table to the latest snapshot with commit time <= the given timestamp.
+
+        Args:
+            timestamp_millis: The timestamp in milliseconds to rollback to.
+
+        Raises:
+            ValueError: If no snapshot exists at or before the given timestamp.
+        """
+        snapshot_mgr = self.snapshot_manager()
+        snapshot = snapshot_mgr.earlier_or_equal_time_mills(timestamp_millis)
+        if snapshot is None:
+            raise ValueError(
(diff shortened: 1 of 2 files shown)


JingsongLi (commented):
Thanks for the PR!

The implementation is clean and straightforward. A few suggestions:

1. **Thread safety concern**: `rollback_to_timestamp` calls `snapshot_manager().earlier_or_equal_time_mills()` followed by `self.rollback_to(snapshot.id)`. Between these two calls, another process could commit new snapshots. This is a low-risk window but worth noting in the docstring.

2. **Error message**: The error message says "earlier than or equal to" but the scenario is that the timestamp is *before the earliest snapshot*. Consider being more specific: `f"No snapshot found at or before timestamp {timestamp_millis}ms. Earliest snapshot is at {snapshot_mgr.try_get_earliest_snapshot().time_millis}ms."`

3. **Test timing**: In `test_table_rollback_to_timestamp`, if snapshots 3 and 4 get the same `time_millis` (fast CI machine), the test could behave unexpectedly since binary search may not return the highest matching ID (as noted in the PR description). Consider adding a small `time.sleep(0.01)` between commits, or at least a comment acknowledging this risk.

Overall LGTM - the alignment with Java's `earlierOrEqualTimeMills` behavior is the right approach.

JingsongLi:
+1

PR #7943, opened 2026-05-24
[python] Add table repair verification logic for pypaimon (Part 1/3)

JunRuiLee:
## Summary
- Implement read-only metadata consistency verification for Paimon tables
- Verifies the chain: LATEST → snapshot → manifest list → manifest files → data files
- Reports broken links, corrupted files, and dangling references
- Supports branch-qualified tables and custom `partition.default-name`
- Progress logging every 1000 data files; time complexity: O(total_data_files)

## Context

- **Part 1 (this PR)**: Read-only verification logic (`TableRepair.verify()`)
- **Part 2**: Fix mode + catalog integration (depends on Part 1)
- **Part 3**: CLI command (depends on Part 2)

## Tests added
- `TestRepairReport`: 5 cases covering RepairReport/RepairIssue data classes
- `TestTableRepairVerify`: 11 cases covering verify() — healthy table, missing manifest list, dangling LATEST, corrupted snapshot, data file checks (missing/existing/DELETE entries/partitioned/custom default-name), branch support


JingsongLi (commented):
Thanks for the well-structured PR. The metadata chain verification logic looks solid overall. A few comments:

1. **Performance concern with `check_data_files`**: When verifying all snapshots, each snapshot's manifest entries are checked independently. For tables with many snapshots sharing the same base manifest files, the same data files will be checked repeatedly. Consider deduplicating the set of (manifest_file, data_file) pairs across snapshots, or at minimum adding a `checked_files` set to skip already-verified paths.

2. **Exception handling in `_verify_manifest_file`**: The blanket `except Exception` that swallows partition deserialization failures (line ~396 in repair.py) means if `GenericRowDeserializer.from_bytes` fails, we silently fall through to construct a wrong data file path (without partition dirs). This could produce false-positive "data file missing" errors. Consider logging a warning or reporting a specific issue category when partition deserialization fails.

3. **`_list_snapshot_ids` — `hasattr` check**: 
   ```python
   name = status.base_name if hasattr(status, 'base_name') else ""
   ```
   If `FileIO.list_status` has a well-defined return type (e.g., `FileStatus`), this `hasattr` check shouldn't be needed. If there are multiple implementations with different return types, it would be clearer to define an interface.

4. **Test coverage**: Good integration tests. Consider adding a test for a table with multiple snapshots sharing manifest files to verify no double-counting in the report stats.

Overall the design is clean — looking forward to seeing Parts 2/3.

JunRuiLee (commented):
@JingsongLi Thanks for the detailed review! All 4 points have been addressed:

1. **Performance**: Added a `checked_files` set that deduplicates data file existence checks across snapshots. Files already verified are skipped.

2. **Exception handling**: Partition deserialization failures now report a warning issue (category: `"partition"`) instead of being silently swallowed, so users can see when data file paths may be incorrect.

3. **`hasattr` check**: Removed — both `LocalFileStatus` and PyArrow `FileInfo` define `base_name`.

4. **Test coverage**: Added `test_check_data_files_shared_manifests_no_double_count` — two snapshots referencing the same manifest file, asserting `data_files_checked == 1`.

PR #7945, opened 2026-05-24
[python] Add CLI command for table repair (Part 3/3)

JunRuiLee:
## Summary
- Add `pypaimon table repair <db.table> [--fix]` CLI command
- Default is dry-run (report only); `--fix` applies available repairs

## Context

Depends on Part 2: #7944
- **Part 1**: Read-only verification logic ✅
- **Part 2**: Fix mode + catalog integration ✅
- **Part 3 (this PR)**: CLI command

## Tests added
- No new test cases in this PR (CLI integration tested via existing catalog-level tests)


JingsongLi (commented):
This PR seems to contain the full diff from Parts 1+2+3 combined (the `operation/repair.py` and `repair_test.py` appear in full). If this is meant to be reviewed as a standalone "Part 3/3" (CLI addition only), the diff should only contain the CLI changes — the repair module changes belong in Parts 1 and 2.

If the intent is that this PR is self-contained (all 3 parts in one), then Parts 1 and 2 (#7943, #7944) should probably be closed.

Regarding the CLI-specific additions:

1. **`cmd_table_repair`** looks good — simple and follows the pattern of other commands.

2. **Missing `--check-data-files` flag**: Same comment as on Part 2 — consider exposing this option in the CLI.

3. **Exit code**: When `report.has_errors` and it's a dry run, you print a hint but exit 0. Consider exit(1) for unhealthy tables even in dry-run mode, so scripting/CI can detect issues:
   ```python
   if report.has_errors:
       sys.exit(1)
   ```

Please clarify the PR dependency/structure between #7943, #7944, and this PR.

JunRuiLee (commented):
Thanks for the review! This PR depends on Part 1 (#7943) and Part 2 (#7944). The current diff includes all three parts because the base hasn't been rebased yet. Marking as Draft — once #7943 and #7944 are merged in order, I'll rebase onto master so this PR only contains the CLI-specific changes. Will address your feedback points then.

PR #7946, opened 2026-05-24
[python] Add create_tag_from_timestamp to FileStoreTable

JunRuiLee:
## Summary

- Add `create_tag_from_timestamp(tag_name, timestamp_millis, ignore_if_exists)` method to `FileStoreTable`
- Uses the existing `SnapshotManager.earlier_or_equal_time_mills()` binary search to find the latest snapshot at or before the given timestamp
- Aligns with Java's `CreateTagFromTimestampProcedure` behavior

## Motivation

PyPaimon currently supports creating tags from a specific snapshot ID or the latest snapshot, but lacks the ability to create a tag from a timestamp — a feature available in the Java API via `CreateTagFromTimestampProcedure`. This is useful for batch workflows that need to pin a snapshot based on wall-clock time.

## Tests

- No new tests added in this PR (existing `earlier_or_equal_time_mills` is already covered by `SnapshotManager` tests)

### DIFF
--- a/paimon-python/pypaimon/table/file_store_table.py
+++ b/paimon-python/pypaimon/table/file_store_table.py
@@ -183,6 +183,31 @@ class FileStoreTable(Table):
         tag_mgr = self.tag_manager()
         tag_mgr.create_tag(snapshot, tag_name, ignore_if_exists)
 
+    def create_tag_from_timestamp(self, tag_name: str, timestamp_millis: int,
+                                  ignore_if_exists: bool = False) -> None:
+        """
+        Create a tag from the latest snapshot at or before the given timestamp.
+
+        Args:
+            tag_name: Name for the tag
+            timestamp_millis: The timestamp in milliseconds. The tag will be
+                created from the latest snapshot with commit time <= this value.
+            ignore_if_exists: If True, don't raise error if tag already exists
+
+        Raises:
+            ValueError: If no snapshot exists at or before the given timestamp,
+                or tag already exists (when ignore_if_exists=False)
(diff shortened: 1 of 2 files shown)


JingsongLi (commented):
Clean and concise implementation. A couple of minor points:

1. **Consistency with Java API**: In the Java Paimon `TagManager`, `create_tag_from_timestamp` typically uses `SnapshotManager.earlierOrEqualTimeMills()` which returns a `Snapshot` or `null`. Your implementation follows the same pattern. LGTM.

2. **Error message**: The ValueError message says "earlier than or equal to" but the method name says `earlier_or_equal_time_mills`. Consider aligning: either say "at or before" or "earlier than or equal to" — just pick one style.

3. **Suggest adding a test**: Even though this is a thin wrapper, a unit test that creates a couple of snapshots with known timestamps and then calls `create_tag_from_timestamp` would protect against regressions (e.g., if `earlier_or_equal_time_mills` semantics change).

Otherwise LGTM — this is a useful addition for time-based tag creation from the Python API.

JunRuiLee (commented):
@JingsongLi Thanks for the review!

1. **Consistency with Java API**: Thanks, glad the pattern aligns.

2. **Error message**: Updated to use "at or before" consistently (matching the docstring).

3. **Test**: Added `test_create_tag_from_timestamp` (creates 2 snapshots, tags from a future timestamp, asserts correct snapshot) and `test_create_tag_from_timestamp_no_snapshot_raises` (timestamp 0, asserts ValueError).

PR #7947, opened 2026-05-24
[python] Fix Daft Paimon write column alignment

QuakeWang:
### What
- Align Daft write batches to the Paimon target schema by column name before casting.
- Reject missing, extra, or duplicate fields instead of silently writing wrong data.
- Add regression coverage with native pypaimon reader verification.

### Tests
- `DAFT_RUNNER=native /tmp/paimon-daft-pr6951-venv/bin/pytest -q paimon-python/pypaimon/tests/daft/daft_sink_test.py`

Related: Eventual-Inc/Daft#6951

### DIFF
--- a/paimon-python/pypaimon/daft/daft_datasink.py
+++ b/paimon-python/pypaimon/daft/daft_datasink.py
@@ -68,28 +68,53 @@ class PaimonDataSink(DataSink[list[Any]]):
             ]
         )
 
+    def _validate_input_schema(self, input_schema: pa.Schema) -> None:
+        target_names = self._target_schema.names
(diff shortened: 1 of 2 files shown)


QuakeWang:
PTAL @chenghuichen

chenghuichen:
+1. 
This is a PR migrated from Daft, and the content is solid — I've already reviewed it there. Daft is order-agnostic on DataFrame columns, while pypaimon requires strict schema match including column order, so aligning by name in the sink is the necessary fix.

JingsongLi:
Re-trigger tests.

JingsongLi:
+1

PR #7950, opened 2026-05-25
[core] Minor refactor partition predicates in FallbackReadScan

juntaozhang:
### Purpose

  This PR is a prerequisite for [`[spark] Support compact_chain_table procedure`](https://github.com/apache/paimon/pull/7313).

  `FallbackReadScan` currently uses a single `partitionPredicate` for both main and fallback scans. However, in chain table compaction scenarios, we need to apply different
  partition filters to the main (snapshot) branch and the fallback (delta) branch. For example, when overwriting an existing partition, we need to:
  - Exclude the target partition from the main scan
  - Include only the target partition in the fallback scan

  This PR refactors `FallbackReadScan` to support separate partition predicates for main and fallback scans.

  ### Tests

  - `FallbackReadFileStoreTableTest.testMainAndFallbackPartitionPredicates`

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/FallbackReadFileStoreTable.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/FallbackReadFileStoreTable.java
@@ -592,7 +592,7 @@ public class FallbackReadFileStoreTable extends DelegatedFileStoreTable {
             DataTableScan scan = scanCreator.apply(isMain ? wrappedTable : fallbackTable);
             if (isMain && getMainPartitionPredicate() != null) {
                 scan.withPartitionFilter(getMainPartitionPredicate());
-            } else if (getFallbackPartitionPredicate() != null) {
+            } else if (!isMain && getFallbackPartitionPredicate() != null) {
                 scan.withPartitionFilter(getFallbackPartitionPredicate());
             }
             return scan;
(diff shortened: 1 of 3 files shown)


paimon-core/src/main/java/org/apache/paimon/table/FallbackReadFileStoreTable.java
```diff
@@ -560,18 +569,31 @@ public List<PartitionEntry> listPartitionEntries() {
...
 
-        private DataTableScan newPartitionListingScan(
-                boolean isMain, PartitionPredicate scanPartitionPredicate) {
+        private DataTableScan newPartitionListingScan(boolean isMain) {
             DataTableScan scan = scanCreator.apply(isMain ? wrappedTable : fallbackTable);
-            if (scanPartitionPredicate != null) {
-                scan.withPartitionFilter(scanPartitionPredicate);
+            if (isMain && getMainPartitionPredicate() != null) {
+                scan.withPartitionFilter(getMainPartitionPredicate());
+            } else if (getFallbackPartitionPredicate() != null) {
```
JingsongLi: No `!isMain &&`? Please add test for this.

juntaozhang: Thanks a lot for your review. The `testMainAndFallbackPartitionPredicates` has already been added.

JingsongLi (approved):
+1

PR #7952, opened 2026-05-25
[python] Implement aggregation merge engine in pypaimon

TheR1sing3un:
### Purpose

Port Java's `AggregateMergeFunction` to pypaimon, following the shape of #7745.

Ships the `FieldAggregator` framework, 9 value aggregators (`sum` / `max` / `min` / `last_value` / `last_non_null_value` / `first_value` / `first_non_null_value` / `bool_or` / `bool_and`) plus the `primary_key` placeholder, the `AggregateMergeFunction` wired into `MergeFileSplitRead._build_merge_function`, and a `merge_engine_support` guard that rejects retract opt-ins, sequence fields and out-of-scope aggregator identifiers (`collect` / `nested_update` / `theta_sketch` / `roaring_bitmap_*` / ...).

Retract handling (DELETE / UPDATE_BEFORE) and the remaining 14 Java aggregators are intentionally deferred to follow-up PRs, mirroring #7745's scoping.

### Tests

- Unit: `test_field_aggregator_registry.py`, `test_field_aggregators.py`, `test_aggregation_merge_function.py`
- E2E: `test_aggregation_e2e.py` covers positive merges (sum across commits, mixed aggregators, null semantics, disjoint keys, default aggregator) plus the unsupported-option guard.


JingsongLi (approved):
LGTM. Solid port of the Java `AggregateMergeFunction` to pypaimon. 

Highlights:
- The `FieldAggregator` framework with registry pattern is extensible and mirrors the Java architecture well
- The 9 built-in aggregators cover the most common use cases
- `merge_engine_support` guard correctly rejects unsupported configurations (retract, sequence-field, out-of-scope aggregators) with clear error messages
- Good test coverage: unit tests for individual aggregators + e2e tests for the full merge pipeline

The scoping decision to defer retract handling and less-common aggregators is pragmatic and well-documented.

+1

paimon-python/pypaimon/read/split_read.py
```diff
@@ -650,6 +652,17 @@ def _build_merge_function(self):
...
+            field_aggregators = build_field_aggregators(
+                self.value_fields,
+                self.trimmed_primary_key,
+                self.table.options,
+            )
+            return AggregateMergeFunction(
+                key_arity=len(self.trimmed_primary_key),
+                value_arity=self.value_arity,
+                field_aggregators=field_aggregators,
+            )
```
leaves12138, line 716: This should use the full table primary key list, not `trimmed_primary_key`. `read_type`/`value_fields` still contains partition columns, while `trimmed_primary_key` removes partition columns from the heap key. If a primary key column is also a partition column and a table-wide default aggregator is configured, that PK value is treated as a normal value field and gets aggregated.

Minimal repro on this PR:

```python
schema = pa.schema([
    pa.field('p', pa.int64(), nullable=False),
    pa.field('id', pa.int64(), nullable=False),
    pa.field('v', pa.int64()),
])
Schema.from_pyarrow_schema(
    schema,
    primary_keys=['p', 'id'],
    partition_keys=['p'],
    options={
        'bucket': '1',
        'merge-engine': 'aggregation',
        'fields.default-aggregate-function': 'sum',
    },
)
# write {'p': 1, 'id': 1, 'v': 10}, then {'p': 1, 'id': 1, 'v': 20}
```

Expected read result is `{'p': 1, 'id': 1, 'v': 30}`, but pypaimon currently returns `{'p': 2, 'id': 1, 'v': 30}` because field `p` falls through to the default `sum` aggregator. Java builds `AggregateMergeFunction` with `tableSchema.primaryKeys()` for the same reason. Please pass `self.table.primary_keys` here and add an e2e test covering a partition key that is also part of the primary key with `fields.default-aggregate-function=sum`.

leaves12138 (changes requested):
Thanks for the implementation. I found one correctness issue in the read-path wiring for partitioned primary-key tables; please fix before merge.

Validation I ran locally:
- `PYTHONPATH=. python -m pytest -q pypaimon/tests/test_field_aggregators.py pypaimon/tests/test_field_aggregator_registry.py pypaimon/tests/test_aggregation_merge_function.py` -> 47 passed
- `PYTHONPATH=. python -m pytest -q pypaimon/tests/test_aggregation_e2e.py pypaimon/tests/test_partial_update_e2e.py` -> 28 passed

JingsongLi (changes requested):
See above review comments.

paimon-python/pypaimon/read/split_read.py
TheR1sing3un: done

TheR1sing3un:
@JingsongLi @leaves12138 Thanks for your review, please review again!

TheR1sing3un:
@JingsongLi @leaves12138 Hi, any suggestions? Lets land it! Thanks!

JingsongLi:
+1

PR #7959, opened 2026-05-25
[python] [daft] Inject REST catalog OSS credentials into Daft IOConfig

XiaoHongbo-Hope:
### Purpose

  Daft reads against an Apache Paimon REST catalog (DLF) currently 401 on OSS: `_convert_paimon_catalog_options_to_io_config` builds Daft's `IOConfig` from static `fs.oss.*` keys, but DLF delivers per-table STS tokens through `table.file_io` rather than `catalog_options`.

  `_read_table` now folds `table.file_io.token.token` into the options before building `IOConfig`, so Daft's native parquet reader gets the same credentials pypaimon uses. It also stops trimming `catalog_options` to `{warehouse: ...}` so the enrichment can read the
  `metastore` field.

  ### Tests

  `daft_catalog_rest_test.py::DaftRestReadTest`:
  - `test_read_table_forwards_full_catalog_options_to_datasource`
  - `test_read_table_enriches_io_config_with_rest_token`

### DIFF
--- a/paimon-python/pypaimon/daft/daft_datasource.py
+++ b/paimon-python/pypaimon/daft/daft_datasource.py
@@ -166,9 +166,6 @@ class PaimonDataSource(DataSource):
 
         self._file_format = table.options.file_format().lower()
         self._is_parquet = self._file_format == PAIMON_FILE_FORMAT_PARQUET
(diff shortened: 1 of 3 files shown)


JingsongLi:
+1

PR #7960, opened 2026-05-25
[python] Add requests as dependency

aisk:
### Purpose

Currently requests is required to using Paimon, otherwise it will raise error when importing `paimon` on current main branch:

```
>>> import pypaimon
Traceback (most recent call last):
  File "<python-input-0>", line 1, in <module>
    import pypaimon
  File "/home/asaka/paimon/paimon-python/pypaimon/__init__.py", line 26, in <module>
    from pypaimon.catalog.catalog_factory import CatalogFactory
  File "/home/asaka/paimon/paimon-python/pypaimon/catalog/catalog_factory.py", line 23, in <module>
    from pypaimon.catalog.filesystem_catalog import FileSystemCatalog
  File "/home/asaka/paimon/paimon-python/pypaimon/catalog/filesystem_catalog.py", line 45, in <module>
    from pypaimon.table.file_store_table import FileStoreTable
  File "/home/asaka/paimon/paimon-python/pypaimon/table/file_store_table.py", line 25, in <module>
    from pypaimon.read.read_builder import ReadBuilder
  File "/home/asaka/paimon/paimon-python/pypaimon/read/read_builder.py", line 20, in <module>
    from pypaimon.common.predicate import Predicate
  File "/home/asaka/paimon/paimon-python/pypaimon/common/predicate.py", line 29, in <module>
    from pypaimon.manifest.schema.simple_stats import SimpleStats
  File "/home/asaka/paimon/paimon-python/pypaimon/manifest/schema/simple_stats.py", line 22, in <module>
    from pypaimon.table.row.generic_row import GenericRow
  File "/home/asaka/paimon/paimon-python/pypaimon/table/row/generic_row.py", line 28, in <module>
    from pypaimon.table.row.blob import BlobData
  File "/home/asaka/paimon/paimon-python/pypaimon/table/row/blob.py", line 24, in <module>
    from pypaimon.common.uri_reader import UriReader, FileUriReader
  File "/home/asaka/paimon/paimon-python/pypaimon/common/uri_reader.py", line 23, in <module>
    import requests
ModuleNotFoundError: No module named 'requests'
```

With the latest release of `pypaimon` on PyPI, we can import `pypaimon`, but `import CatalogFactory from pypaimon` will raise the same error.

Adding `requests` to `requirements.txt` can resolve this error. `requests 2.21.0` was released in 2018, I think it's old enough to be used as the minimum version range.

### DIFF
--- a/paimon-python/dev/requirements.txt
+++ b/paimon-python/dev/requirements.txt
@@ -37,3 +37,4 @@ zstandard>=0.19,<1
 backports.zstd>=1.0.0,<1.4.0; python_version >= "3.9" and python_version < "3.14"
 cramjam>=1.3.0,<3; python_version>="3.7"
 pyyaml>=5.4,<7
+requests>=2.21.0,<3


JingsongLi (commented):
The fix itself is correct — `requests` is imported unconditionally by `uri_reader.py` so it must be declared as a dependency.

However, the file being modified is `dev/requirements.txt`. Is this actually the runtime dependency spec that PyPI packaging picks up? I'd expect it to be in `setup.py`, `pyproject.toml`, or a dedicated `requirements.txt` at the package root (not under `dev/`). If this is only used for development/CI environments, the import-time failure on the released wheel would remain unsolved.

Could you check whether this is the correct file for declaring runtime dependencies? If there's a separate packaging config (e.g., `pyproject.toml` or `setup.cfg`), `requests` should be added there as well.

aisk:
Currently, pypaimon is using a legacy `setup.py` and reading `dev/requirements.txt` as dependencies in it, so this PR could resolve the issue:

https://github.com/apache/paimon/blob/d2ce7ae5d3c853ed523f87fab0244aed531c1896/paimon-python/setup.py#L121-L134

I think we should upgrade to a modern way of managing dependencies (e.g., using `pyproject.toml` with or without uv), but I think that could happen in another PR.

XiaoHongbo-Hope:
+1

XiaoHongbo-Hope (approved):

XiaoHongbo-Hope:
Thanks for raising and fixing the issue. I will include this fix in 1.4.2.

PR #7961, opened 2026-05-25
[python] Support Vector dedicated file write/read and row-level access

JingsongLi:
Add full Python support for Paimon's VECTOR data type with dedicated vector file storage (.vector.<format>), row-level get_vector() access, and Data Evolution read support.


leaves12138 (changes requested):
Thanks for the update. The multi-vector-column Java/Python layout is now aligned, but I found one remaining correctness issue.

When a table contains only VECTOR columns and `vector.file.format` is enabled, Python silently drops all written rows. `DataVectorWriter.write()` sets `processed_normal` to `None` because `normal_column_names` is empty, while it still writes the vector batch to `VectorWriter`. Later `prepare_commit()` calls `_close_current_writers()`, but that method returns immediately when `pending_normal_data is None`, so the vector writer is never closed and no `DataFileMeta` is returned/committed.

A minimal Python reproduction on this PR head:

```python
pa_schema = pa.schema([
    ('embed1', pa.list_(pa.float32(), 3)),
    ('embed2', pa.list_(pa.float32(), 2)),
])
schema = Schema.from_pyarrow_schema(pa_schema, options={
    'vector.file.format': 'parquet',
    'row-tracking.enabled': 'true',
    'data-evolution.enabled': 'true',
    'bucket': '-1',
})
# write two rows...
msgs = writer.prepare_commit()
print(len(msgs), [f.file_name for m in msgs for f in m.new_files])
# => 0, []
```

Reading the table afterwards returns zero rows. Java does not have this problem because it still creates a normal data file with an empty row type as the row-id anchor and then adds the vector-store file. Python should do the same (or otherwise create equivalent anchor metadata) before committing vector files. Please add a regression test for a vector-only table as well.

leaves12138 (changes requested):
Thanks for the update. The vector-only case still fails when row tracking is enabled, which is the configuration required by the dedicated vector/data-evolution path in the E2E tests.

The new unit test creates the vector-only table with only `vector.file.format` and `bucket`, so `_assign_row_tracking_meta` is skipped. With the same options used by the dedicated vector path (`row-tracking.enabled=true`, `data-evolution.enabled=true`, `bucket=-1`), Python now produces one `.vector.parquet` file, but commit fails before assigning row ids:

```text
RuntimeError: This is a bug, vectorStoreStart 0 should be less than start 0 when assigning a vector-store entry file.
```

Minimal reproduction on the latest head:

```python
pa_schema = pa.schema([
    ('embed1', pa.list_(pa.float32(), 3)),
    ('embed2', pa.list_(pa.float32(), 2)),
])
schema = Schema.from_pyarrow_schema(pa_schema, options={
    'vector.file.format': 'parquet',
    'row-tracking.enabled': 'true',
    'data-evolution.enabled': 'true',
    'bucket': '-1',
})
# write two rows, then commit
```

`writer.prepare_commit()` returns one vector file with `first_row_id=None`, then `new_commit().commit(...)` raises because there is no normal data file to advance `start` before the vector file is processed. Please update the fix/test to cover the row-tracking-enabled vector-only case as well.

leaves12138 (changes requested):
I found one more Python-side validation gap around the dedicated vector path.

Java schema validation already rejects `data-evolution.enabled=true` unless `row-tracking.enabled=true`, but `Schema.from_pyarrow_schema` only checks `data-evolution.enabled` for `vector.file.format`. As a result Python can create and commit a dedicated-vector table with data evolution enabled but row tracking disabled:

```python
    pa.schema([
        ('id', pa.int32()),
        ('embed', pa.list_(pa.float32(), 3)),
    ]),
    options={
        'vector.file.format': 'parquet',
        'data-evolution.enabled': 'true',
        'bucket': '-1',
    },
)
```

The write produces normal/vector files with `first_row_id=None`; then reading fails in `DataEvolutionSplitGenerator._split_by_row_id` because it tries to sort `None` row-id ranges:

```text
AttributeError: 'NoneType' object has no attribute 'from_'
```

Please align Python validation with Java and reject data-evolution tables when `row-tracking.enabled` is not true (or at least add that requirement for `vector.file.format`). A regression test with `vector.file.format + data-evolution.enabled=true` but missing `row-tracking.enabled` would cover this.

leaves12138 (approved):
Thanks for the updates. I rechecked the vector-only and missing row-tracking cases, and the latest validation fixes address the previous blockers. I also verified the normal + multi-vector dedicated write/read path locally.

PR #7965, opened 2026-05-26
[python] Fix Daft fallback filter limit pushdown

QuakeWang:
### Purpose

The Daft Paimon datasource used the configured read builder for scan planning, but fallback split tasks rebuilt a bare `TableRead`. As a result, fallback reads for PK merge, non-Parquet formats, BLOB columns, and deletion-vector paths could miss pushed predicate/projection/limit state.

This was correctness-sensitive because Daft may treat pushed filters as already handled by the source. A query filtering a fallback table could plan the right split but still emit unfiltered rows from that split. There was also a related limit-ordering issue: applying `limit` inside the fallback reader while Daft still had remaining row or partition filters could truncate rows before those filters were evaluated.

This patch makes fallback tasks use a configured `TableRead`, keeps the required filter columns available for fallback execution, and centralizes the source-side limit decision so `limit` is only pushed when it is safe for the source to apply it before returning rows.

### Tests

```shell
python -m py_compile \
  paimon-python/pypaimon/daft/daft_datasource.py \
  paimon-python/pypaimon/tests/daft/daft_data_test.py

python -m pytest paimon-python/pypaimon/tests/daft/daft_data_test.py -q

python -m pytest \
  paimon-python/pypaimon/tests/daft/daft_sink_test.py::TestBlobType::test_write_read_blob_type -q

### DIFF
--- a/paimon-python/pypaimon/daft/daft_datasource.py
+++ b/paimon-python/pypaimon/daft/daft_datasource.py
@@ -18,6 +18,7 @@
 
 from __future__ import annotations
 
+from dataclasses import dataclass
 import logging
 from typing import TYPE_CHECKING, Any
 from urllib.parse import urlparse
@@ -37,6 +38,7 @@ if TYPE_CHECKING:
     from collections.abc import AsyncIterator
 
     from pypaimon.common.predicate import Predicate
+    from pypaimon.read.table_read import TableRead
     from pypaimon.read.split import Split
     from pypaimon.table.file_store_table import FileStoreTable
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: paimon-python/pypaimon/daft/daft_datasource.py [hunk 1]
```diff
@@ -213,18 +219,29 @@ async def get_tasks(self, pushdowns: Pushdowns) -> AsyncIterator[DataSourceTask]
             read_table = self._table.copy({"blob-as-descriptor": "true"})
 
         read_builder = read_table.new_read_builder()
+        reader_predicate, filters_consumed = self._pushdown_filter_state(pushdowns)
+        planning_predicate = self._planning_predicate(reader_predicate)
+        requested_columns = self._valid_output_columns(pushdowns.columns)
+        task_columns = self._task_columns(read_table, requested_columns, pushdowns)
+        read_columns = self._fallback_read_columns(read_table, task_columns, reader_predicate)
```
XiaoHongbo-Hope: nit: these locals (+ `source_limit` below) must stay in lockstep across
  `get_tasks`'s planner and `_fallback_read_builder`. Consider grouping
  them into a small shared dataclass. Non-blocking.

XiaoHongbo-Hope:
Looks good to me with little comment.

paimon-python/pypaimon/daft/daft_datasource.py
QuakeWang: @XiaoHongbo-Hope I have addressed this : )

JingsongLi:
+1

PR #7968, opened 2026-05-26
[python] Implement first-row merge engine for read path

JunRuiLee:
### Purpose

Add read-path support for the `first-row` merge engine in pypaimon. The first-row engine keeps only the earliest row per primary key, which is the opposite of the default `deduplicate` engine that keeps the latest.

Previously, reading a table configured with `merge-engine: first-row` raised `NotImplementedError`. This PR implements the merge function and wires it into the read pipeline.

### Changes

- Add `FirstRowMergeFunction` that keeps the first add-type row per key, with `ignore-delete` support matching Java semantics
- Add `CoreOptions.ignore_delete()` with fallback key resolution (`ignore-delete`, `first-row.ignore-delete`, `deduplicate.ignore-delete`, `partial-update.ignore-delete`), reusing `OptionsUtils.convert_to_boolean()` for type-safe parsing
- Update `merge_engine_support.check_supported()` and `MergeFileSplitRead._build_merge_function()` to accept and dispatch `FIRST_ROW`
- Update existing `test_first_row_engine_raises_not_implemented` to verify correct first-row behavior instead

### Scope

This PR implements first-row merge function support in the merge-read path (cross-file/cross-commit duplicate key merging). It intentionally does **not** address duplicate primary keys that may be written within a single Python writer flush — that writer-side parity issue is tracked separately by https://github.com/apache/paimon/pull/7759.

### Testing

- 10 unit tests covering: first-row-wins semantics, reset/reuse, ignore-delete on/off, retract rejection
- 4 end-to-end tests: single/multi-key writes, three-write override, single-write passthrough

### DIFF
--- a/paimon-python/pypaimon/common/options/core_options.py
+++ b/paimon-python/pypaimon/common/options/core_options.py
@@ -23,6 +23,7 @@ from typing import Dict, Optional
 from pypaimon.common.memory_size import MemorySize
 from pypaimon.common.options import Options
 from pypaimon.common.options.config_option import ConfigOption
+from pypaimon.common.options.options_utils import OptionsUtils
 from pypaimon.common.options.config_options import ConfigOptions
 
 
@@ -375,6 +376,14 @@ class CoreOptions:
         .with_description("Specify the merge engine for table with primary key. "
                           "Options: deduplicate, partial-update, aggregation, first-row.")
     )
+
+    IGNORE_DELETE: ConfigOption[bool] = (
+        ConfigOptions.key("ignore-delete")
+        .boolean_type()
+        .default_value(False)
+        .with_description("Whether to ignore delete records.")
+    )
+
     # Commit options
     COMMIT_USER_PREFIX: ConfigOption[str] = (
         ConfigOptions.key("commit.user-prefix")
@@ -762,6 +771,19 @@ class CoreOptions:
 
     def merge_engine(self, default=None):
         return self.options.get(CoreOptions.MERGE_ENGINE, default)
+
+    def ignore_delete(self) -> bool:
+        raw = self.options.to_map()
+        fallback_keys = (
+            "ignore-delete", "first-row.ignore-delete",
+            "deduplicate.ignore-delete",
+            "partial-update.ignore-delete",
+        )
+        for key in fallback_keys:
+            val = raw.get(key)
+            if val is not None:
+                return OptionsUtils.convert_to_boolean(val)
+        return False
 
     def data_file_external_paths(self, default=None):
         external_paths_str = self.options.get(CoreOptions.DATA_FILE_EXTERNAL_PATHS, default)
--- a/paimon-python/pypaimon/read/merge_engine_support.py
+++ b/paimon-python/pypaimon/read/merge_engine_support.py
@@ -62,6 +62,8 @@ def check_supported(table) -> None:
     engine = table.options.merge_engine()
     if engine == MergeEngine.DEDUPLICATE:
         return
+    if engine == MergeEngine.FIRST_ROW:
+        return
     if engine == MergeEngine.PARTIAL_UPDATE:
         unsupported = partial_update_unsupported_options(table)
         if unsupported:
@@ -79,7 +81,7 @@ def check_supported(table) -> None:
         return
     raise NotImplementedError(
         "merge-engine '{}' is not implemented in pypaimon yet "
-        "(supported: deduplicate, partial-update). Use the Java "
+        "(supported: deduplicate, first-row, partial-update). Use the Java "
         "client or open an issue to track support.".format(engine.value)
     )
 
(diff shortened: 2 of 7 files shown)


leaves12138 (changes requested):
Thanks for the patch. I reviewed the latest version and ran the focused tests:

```
PYTHONPATH=. python -m unittest \
  pypaimon.tests.test_first_row_merge_function \
  pypaimon.tests.test_first_row_e2e \
  pypaimon.tests.test_partial_update_e2e
```

They pass, but I think there is still a correctness gap before enabling `first-row` generally in pypaimon.

The new `FirstRowMergeFunction` is only used by `MergeFileSplitRead`, i.e. when a PK split is not `raw_convertible`. However, a single-file PK split is still raw-convertible, so `TableRead` dispatches it through `RawFileSplitRead` and bypasses the merge function completely. This means duplicate keys inside one data file are returned as duplicate rows instead of applying first-row semantics.

A minimal reproduction on this PR is:

```python
    pa.schema([pa.field('id', pa.int64(), nullable=False), ('a', pa.string())]),
    primary_keys=['id'],
    options={'bucket': '1', 'merge-engine': 'first-row'},
)
# one commit / one data file
w.write_arrow(pa.Table.from_pylist([
    {'id': 1, 'a': 'first'},
    {'id': 1, 'a': 'second'},
], schema=pa_schema))
```

The planned split is `raw_convertible=True`, and reading returns both rows:

```text
[{'id': 1, 'a': 'first'}, {'id': 1, 'a': 'second'}]
```

For Java writers this is normally reduced before writing the file, but pypaimon can create such a file today because the Python PK writer only sorts by PK/sequence and does not merge duplicate keys before writing. So enabling `first-row` in the read support check can expose incorrect results for data produced by pypaimon itself.

I think we should either make pypaimon writes reduce duplicate PKs for `first-row`/`deduplicate`, or ensure the read path does not mark such splits raw-convertible when the data file may contain duplicate keys. At minimum, this case needs a regression test before considering the engine supported.

JunRuiLee:
> Thanks for the patch. I reviewed the latest version and ran the focused tests:
> 
> ```
> PYTHONPATH=. python -m unittest \
>   pypaimon.tests.test_first_row_merge_function \
>   pypaimon.tests.test_first_row_e2e \
>   pypaimon.tests.test_partial_update_e2e
> ```
> 
> They pass, but I think there is still a correctness gap before enabling `first-row` generally in pypaimon.
> 
> The new `FirstRowMergeFunction` is only used by `MergeFileSplitRead`, i.e. when a PK split is not `raw_convertible`. However, a single-file PK split is still raw-convertible, so `TableRead` dispatches it through `RawFileSplitRead` and bypasses the merge function completely. This means duplicate keys inside one data file are returned as duplicate rows instead of applying first-row semantics.
> 
> A minimal reproduction on this PR is:
> 
> ```python
> schema = Schema.from_pyarrow_schema(
>     pa.schema([pa.field('id', pa.int64(), nullable=False), ('a', pa.string())]),
>     primary_keys=['id'],
>     options={'bucket': '1', 'merge-engine': 'first-row'},
> )
> # one commit / one data file
> w.write_arrow(pa.Table.from_pylist([
>     {'id': 1, 'a': 'first'},
>     {'id': 1, 'a': 'second'},
> ], schema=pa_schema))
> ```
> 
> The planned split is `raw_convertible=True`, and reading returns both rows:
> 
> ```
> [{'id': 1, 'a': 'first'}, {'id': 1, 'a': 'second'}]
> ```
> 
> For Java writers this is normally reduced before writing the file, but pypaimon can create such a file today because the Python PK writer only sorts by PK/sequence and does not merge duplicate keys before writing. So enabling `first-row` in the read support check can expose incorrect results for data produced by pypaimon itself.
> 
> I think we should either make pypaimon writes reduce duplicate PKs for `first-row`/`deduplicate`, or ensure the read path does not mark such splits raw-convertible when the data file may contain duplicate keys. At minimum, this case needs a regression test before considering the engine supported.

Thanks for pointing this out.

I mentioned this issue already in the PR description as an out-of-scope writer-side invariant issue: pypaimon can currently write duplicate PKs into a single data file, while `raw_convertible` reads assume each PK file has already been folded according to the merge engine.

This is the same class of issue discussed in #7745: the same-file duplicate-PK case was identified as a general pypaimon PK writer invariant problem, not as a blocker for that read-side merge-engine PR, and was left to a separate follow-up.

  So my intended scope for this PR is only:
  1. implement `FirstRowMergeFunction`;
  2. allow `first-row` in the Python merge-read path when `MergeFileSplitRead` is used;
  3. keep the single-file duplicate-PK case tracked by #7759.

JingsongLi (approved):
LGTM. Clean implementation of the first-row merge engine for pypaimon's read path.

The `FirstRowMergeFunction` correctly keeps only the first add-type row per key, rejects DELETE/UPDATE_BEFORE by default with a clear error message, and respects the `ignore-delete` option with fallback key resolution matching Java semantics.

Good scoping note about writer-side parity being tracked separately.

One minor observation: the `ignore_delete()` method in `CoreOptions` checks multiple fallback keys (`ignore-delete`, `first-row.ignore-delete`, `deduplicate.ignore-delete`, `partial-update.ignore-delete`). This matches Java behavior, but make sure the iteration order is intentional — currently the generic `ignore-delete` takes precedence over engine-specific keys.

+1

JingsongLi:
Please rebase master.

JunRuiLee:
> Please rebase master.

Thanks for remind, rebased.

Issue #7964, opened 2026-05-26
[Bug] Partition expiration stuck in infinite failure loop when HMS drop succeeds but Paimon drop fails due to unhandled MetaException


GangYang-HX:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Paimon version

Paimon-1.1.1

### Compute Engine

Spark-3.5.0

### Minimal reproduce step

Executing the "CALL sys.expire_partitions" operation via the Spark engine:

1. Executing `partition expire` throws an exception after the HMS Drop Partition is completed, preventing subsequent Paimon partition metadata deletion processing.

2. Executing `partition expire` again will throw an exception.

### What doesn't meet your expectations?

Failed to correctly identify the partition not found exception

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

GangYang-HX:
**Complete exception log**
```
2026/05/25 11:32:13  java.lang.RuntimeException: MetaException(message:java.lang.NullPointerException)
	at org.apache.paimon.hive.HiveCatalog.dropPartitions(HiveCatalog.java:428)
	at org.apache.paimon.table.PartitionHandler$1.dropPartitions(PartitionHandler.java:54)
	at org.apache.paimon.operation.PartitionExpire.doExpire(PartitionExpire.java:153)
	at org.apache.paimon.operation.PartitionExpire.expire(PartitionExpire.java:129)
	at org.apache.paimon.operation.PartitionExpire.expire(PartitionExpire.java:99)
	at org.apache.paimon.spark.procedure.ExpirePartitionsProcedure.lambda$call$2(ExpirePartitionsProcedure.java:105)
	at org.apache.paimon.spark.procedure.BaseProcedure.execute(BaseProcedure.java:88)
	at org.apache.paimon.spark.procedure.BaseProcedure.modifyPaimonTable(BaseProcedure.java:78)
	at org.apache.paimon.spark.procedure.ExpirePartitionsProcedure.call(ExpirePartitionsProcedure.java:85)
	at org.apache.paimon.spark.execution.PaimonCallExec.run(PaimonCallExec.scala:32)
Caused by: MetaException(message:java.lang.NullPointerException)
	at org.apache.hadoop.hive.metastore.api.ThriftHiveMetastore$drop_partition_with_environment_context_result$drop_partition_with_environment_context_resultStandardScheme.read(ThriftHiveMetastore.java)
[...]

PR #7969, opened 2026-05-26
Add the parameter: metastore.drop-partition-ignore-nonexistent to ignore exceptions when a Hive partition does not exist.

GangYang-HX:
Fixed a bug that caused the Partition Expire operation to fail due to a non-existent Hive partition.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,5 @@ Cargo.lock
 *libvortex_jni*
 ### Tantivy lib ###
 *libtantivy_jni*
+
+.cursor-svg-check/
--- a/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
+++ b/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
@@ -1673,6 +1673,20 @@ public class CoreOptions implements Serializable {
                                     + "you need to create this table as a partitioned table in Hive metastore.\n"
                                     + "This config option does not affect the default filesystem metastore.");
 
+    public static final ConfigOption<Boolean> METASTORE_DROP_PARTITION_IGNORE_NONEXISTENT =
+            key("metastore.drop-partition-ignore-nonexistent")
+                    .booleanType()
+                    .defaultValue(false)
+                    .withDescription(
+                            "Whether to ignore HMS drop-partition failures when the partition "
+                                    + "is confirmed not to exist in Hive Metastore. "
+                                    + "Useful when HMS has been manually cleaned up or returns "
+                                    + "non-standard exceptions (e.g., MetaException caused by "
+                                    + "NullPointerException in some HMS versions) instead of "
+                                    + "NoSuchObjectException for missing partitions. "
+                                    + "When enabled, Paimon will verify the partition's existence "
+                                    + "before deciding to ignore the error.");
+
     public static final ConfigOption<String> METASTORE_TAG_TO_PARTITION =
             key("metastore.tag-to-partition")
                     .stringType()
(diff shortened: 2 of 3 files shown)


JingsongLi (commented):
A few comments on this PR:

1. **`.gitignore` change**: The `.cursor-svg-check/` entry looks unrelated to this PR — likely IDE-specific. Please remove it or put it into a separate commit.

2. **Design question**: When `MetaException` is caught with `ignorePartitionNonexistent=true`, the code calls `partitionExistsInHms()` which issues a second HMS RPC (`getPartition`). If HMS is unhealthy (which is likely the cause of the original `MetaException`), this verification call may also fail — and the fallback in that case conservatively returns `true`, causing the original exception to be rethrown anyway. This is a reasonable safety choice, but it means the option won't help when HMS is in a bad state.

   Consider documenting this behavior (i.e., "the option only helps when the specific partition is genuinely absent, not when HMS is having connectivity issues").

3. **Title format**: The PR title doesn't follow the project convention of `[module] description`. Should be something like: `[hive] Add metastore.drop-partition-ignore-nonexistent option`.

4. **Option naming**: The Paimon project typically uses shorter, more consistent option names. `metastore.drop-partition-ignore-nonexistent` is quite long. Have you considered `metastore.drop-partition.ignore-not-exists` to be more consistent with other Paimon options?

.gitignore
```diff
@@ -51,3 +50,5 @@ Cargo.lock
 *libvortex_jni*
 ### Tantivy lib ###
 *libtantivy_jni*
+
+.cursor-svg-check/
```
lilei1128, line 54: remove this file from the PR

paimon-hive/paimon-hive-catalog/src/main/java/org/apache/paimon/hive/HiveCatalog.java
```diff
@@ -461,6 +463,20 @@ public void dropPartitions(Identifier identifier, List<Map<String, String>> part
                                                     false));
                 } catch (NoSuchObjectException e) {
                     // do nothing if the partition not exists
+                } catch (MetaException e) {
+                    if (ignorePartitionNonexistent
+                            && !partitionExistsInHms(identifier, partitionValues)) {
```
lilei1128, line 468: I think the MetaException returned by HMS is very likely still remaining in the HMS.
Check the processing logic of HMS:
1.ms.getPartition()   → OK
2.ms.getDatabase()    → NPE，MetaException 
3.ms.dropPartition()

lilei1128: Therefore, I suggest that when this configuration is enabled, we ignore HMS errors, log a warning, and then delete the Paimon data.

lilei1128:
close  #7964

paimon-api/src/main/java/org/apache/paimon/CoreOptions.java
```diff
@@ -1673,6 +1673,20 @@ public InlineElement getDescription() {
                                     + "you need to create this table as a partitioned table in Hive metastore.\n"
                                     + "This config option does not affect the default filesystem metastore.");
 
+    public static final ConfigOption<Boolean> METASTORE_DROP_PARTITION_IGNORE_NONEXISTENT =
+            key("metastore.drop-partition-ignore-nonexistent")
```
JingsongLi, line 1677: This new option also needs to be added to the generated configuration docs. The PR CI is currently failing in `ConfigOptionsDocsCompletenessITCase` with `Option metastore.drop-partition-ignore-nonexistent in class org.apache.paimon.CoreOptions is not documented.` Please regenerate the docs following `paimon-docs/README.md` (or otherwise add the generated doc update) so the docs completeness check passes.

PR #7970, opened 2026-05-26
[core] [flink] [spark] Preserve blob view references during forwarding

leaves12138:
### Purpose

Blob view currently resolves `BlobViewStruct` to upstream BLOB content on reads. This makes it hard to forward one blob view table into another blob view table while preserving the original upstream reference.

This PR adds a dynamic read option to preserve blob view references during such forwarding:

- `blob-view.resolve.enabled=true` keeps the existing default behavior and resolves blob view fields at read time.
- `blob-view.resolve.enabled=false` skips read-time resolution and returns the original `BlobViewStruct` as a `BlobView`, so writing it into another `blob-view-field` stores the same upstream reference.

### Changes

- Add `CoreOptions.BLOB_VIEW_RESOLVE_ENABLED`.
- Make `DataEvolutionTableRead` skip blob view resolving when this option is disabled.
- Add a test that forwards `T1` blob view rows into `T2` while preserving references to `T0`.
- Document the option and the forwarding pattern.

### Tests

- `mvn -pl paimon-format -am -DskipTests package`
- `mvn -pl paimon-api,paimon-common,paimon-format -am -DskipTests install`
- `mvn -pl paimon-core -Pfast-build -Dtest=BlobTableTest#testForwardBlobViewReference test`
- `mvn -pl paimon-core -Pfast-build -DskipTests compile`
- `mvn -pl paimon-api,paimon-core spotless:check -DskipTests`
- `git diff --check`

### DIFF
--- a/docs/docs/append-table/blob.mdx
+++ b/docs/docs/append-table/blob.mdx
@@ -148,6 +148,18 @@ This allows one table to mix raw-data BLOB fields, descriptor-only BLOB fields, 
       </td>
     </tr>
     <tr>
+      <td><h5>blob-view.resolve.enabled</h5></td>
+      <td>No</td>
+      <td style={{wordWrap: "break-word"}}>true</td>
+      <td>Boolean</td>
+      <td>
+        Controls whether <code>blob-view-field</code> values are resolved to the upstream BLOB
+        content at read time. Set this dynamic option to <code>false</code> when forwarding blob view
+        references from one view table to another view table and you want the target table to keep
+        referencing the original upstream BLOB.
+      </td>
+    </tr>
+    <tr>
       <td><h5>blob-external-storage-field</h5></td>
       <td>No</td>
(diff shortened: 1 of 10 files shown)


leaves12138:
Thanks for the update. I reviewed the latest patch and the blob-view forwarding path looks good to me.

The new `blob-view.resolve.enabled=false` read option cleanly bypasses the blob-view resolving reader, so unresolved `BlobViewStruct` values can be forwarded without triggering upstream lookup or materializing the actual blob bytes. The Flink and Spark row wrappers also preserve unresolved `BlobView` values via `Blob.serializeBlob`, while keeping the existing resolved/blob-as-descriptor behavior unchanged.

The added core/Flink/Spark coverage exercises the important end-to-end forwarding case, and the current CI is green. I do not see remaining blocking issues.

JingsongLi (approved):
+1

PR #7972, opened 2026-05-26
[core][flink] Detect overwrite conflicts for index commits

leaves12138:
### Purpose

Prevent stale global index commits after later `OVERWRITE` snapshots change row-id/index safety.

### Changes

- Add index commit overwrite conflict checking via `commit.overwrite-conflict-with-index.last-safe-snapshot`.
- Mark row-id reassignment `OVERWRITE` snapshots with transient `overwrite-barrier`.
- Abort index commits on later overwrite barrier snapshots or overlapping ordinary overwrite snapshots.

### Validation

- `git diff --check`
- `mvn -pl paimon-api,paimon-docs -Pfast-build -DfailIfNoTests=false -Dtest=ConfigOptionsDocsCompletenessITCase#testCompleteness test`


paimon-core/src/main/java/org/apache/paimon/operation/commit/ConflictDetection.java
```diff
@@ -131,6 +132,14 @@ public boolean hasRowIdCheckFromSnapshot() {
         return rowIdCheckFromSnapshot != null;
     }
 
+    public void setRowIdReassignCheckFromSnapshot(@Nullable Long rowIdReassignCheckFromSnapshot) {
```
JingsongLi: Can we just reuse setRowIdCheckFromSnapshot?

jerry-024 (commented):
Overall LGTM. Three minor optimization suggestions below — none are blockers.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/globalindex/GenericIndexTopoBuilder.java
```diff
@@ -179,6 +179,10 @@ public static boolean buildIndex(
 
         List<ManifestEntry> entries = indexBuilder.scan();
         List<IndexManifestEntry> deletedIndexEntries = indexBuilder.deletedIndexEntries();
+        Long rowIdReassignCheckFromSnapshot =
+                indexBuilder.scanSnapshotId().isPresent()
+                        ? indexBuilder.scanSnapshotId().get()
+                        : null;
```
jerry-024: nit: Can be simplified to:
```java
Long rowIdReassignCheckFromSnapshot = indexBuilder.scanSnapshotId().orElse(null);
```
`Optional.orElse(null)` is idiomatic and avoids the double method call.

paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/globalindex/GenericIndexTopoBuilder.java
```diff
@@ -506,22 +512,39 @@ private static List<Committable> createDeleteCommittables(
...
+            FileStoreTable table, Long rowIdReassignCheckFromSnapshot) {
+        if (rowIdReassignCheckFromSnapshot == null) {
+            return table;
+        }
+        return table.copy(
+                Collections.singletonMap(
+                        CoreOptions.COMMIT_ROW_ID_REASSIGN_LAST_SAFE_SNAPSHOT.key(),
+                        String.valueOf(rowIdReassignCheckFromSnapshot)));
+    }
+
```
jerry-024: nit: This method is identical to `BTreeIndexTopoBuilder.withRowIdReassignCheck`. Consider extracting to a shared utility to avoid drift if the option key or logic changes later. Not blocking — the method is small.

paimon-core/src/test/java/org/apache/paimon/operation/FileStoreCommitTest.java
```diff
@@ -1008,6 +1009,100 @@ public void testCommitManifestWithProperties() throws Exception {
...
+        }
+
+        Snapshot normalSnapshot = checkNotNull(store.snapshotManager().latestSnapshot());
+        assertThat(normalSnapshot.properties())
+                .containsEntry("keep", "v1")
+                .doesNotContainKey(Snapshot.ROW_ID_REASSIGN_PROPERTY);
+    }
+
+    @Test
+    public void testRowIdReassignConflictFromOptions() throws Exception {
```
jerry-024: suggestion: Consider adding a test that verifies `compactManifestOnce` also strips `ROW_ID_REASSIGN_PROPERTY`. Current tests cover `replaceManifestList` but not the compact path.

paimon-core/src/main/java/org/apache/paimon/operation/commit/ConflictDetection.java
leaves12138: Thanks for the suggestion. I think these two snapshot markers should stay separate because they guard different invariants. `rowIdCheckFromSnapshot` is used by row-id range / column-level conflict checks, while this one is driven by `commit.row-id-reassign.last-safe-snapshot` and scans later OVERWRITE snapshots for row-id reassign or overlapped partition overwrite conflicts during index build/compact commits. Reusing the existing field would couple two independent checks and make the COMPACT/index path harder to reason about.

JingsongLi:
Reviewed commit 36f6f1a. The overwrite-conflict check looks scoped to index commits via the scan snapshot from the index builders, the overwrite barrier is treated as a one-shot marker and stripped from derived snapshots, and the added tests cover barrier, partition-overlap, non-overwrite, and data-only cases. I didn't find blocking issues from my side.

JingsongLi (changes requested):
I cannot get why introducing `OVERWRITE_BARRIER_PROPERTY`. Can you explain more?

JingsongLi (changes requested):
After discussing, you should only need to check whether the RowID data exists

paimon-core/src/main/java/org/apache/paimon/globalindex/btree/BTreeGlobalIndexBuilder.java
```diff
@@ -133,6 +134,10 @@ public BTreeGlobalIndexBuilder withSnapshot(Snapshot snapshot) {
         return this;
     }
 
+    public Optional<Long> scanSnapshotId() {
```
JingsongLi: Why providing this method?

leaves12138: Good catch. This method was a leftover from an earlier approach where the BTree index builder exposed its scan snapshot to the commit conflict checker. The current implementation no longer consumes this stateful getter; index commits are validated by checking added global-index row ranges against current data-file row ranges during commit. I removed the unused field/getter and the related assignments in 9c2752005.

JingsongLi:
+1

PR #7973, opened 2026-05-26
[core] Add origin table option for blob view lookup

leaves12138:
### What changed
- Pass the current table identifier from append-only data evolution reads into BlobViewLookup.
- Create a lookup-only CatalogContext carrying origin-view-table without mutating the original catalog context.
- Add unit coverage for the context option injection path.

### Why
BlobViewLookup may need to recreate a catalog while resolving referenced blob descriptors. Carrying the origin table in catalog options lets downstream catalog/auth integrations keep the original query table context.

### Validation
- mvn -pl paimon-core -Pfast-build -DskipTests compile
- mvn -pl paimon-core -Pfast-build -Dtest=BlobViewLookupTest test
- mvn -pl paimon-core -Pfast-build -Dtest=BlobTableTest#testBlobViewE2E test


JingsongLi (changes requested):
-1 We should not contaminate the Catalog configuration

PR #7975, opened 2026-05-26
[cli] paimon cli

hqlalala:
### Purpose
 - Add a standalone Java CLI (`paimon-cli`) for Apache Paimon, providing a native command-line tool that works without Flink/Spark
  - Full SQL support via Apache Calcite planner (JOIN, GROUP BY, window functions, subqueries, UNION, etc.)
  - Supports catalog operations: create/drop/alter database and table, schema evolution, snapshot/tag management, branch management, orphan file cleanup
  - Includes data read/write, EXPLAIN plan, and an interactive SQL REPL
  - Installs as a single `paimon` command with fat jar packaging

  ## Comparison with Python CLI

  | Feature | Java CLI | Python CLI |
  |---------|----------|------------|
  | SQL Engine | Apache Calcite (full planner) | DataFusion (Rust) |
  | SQL Capabilities | JOIN, window functions, subqueries, UNION | Same |
  | Filter/Projection Pushdown | Yes (via ProjectableFilterableTable) | Yes (native) |
  | Deployment | Single fat jar, JVM only | Requires Python + pypaimon wheel |
  | Catalog Operations | Full (create/alter/drop DB & table) | Full |
  | Schema Evolution | Yes (add/drop/rename column, set options) | Limited |
  | Snapshot Management | Yes (view, tag, expire, rollback) | No |
  | Branch Management | Yes (create, list, rename, delete) | No |
  | Orphan File Cleanup | Yes | No |
  | Interactive REPL | Yes | Yes |

### Tests
  - [x] 98 unit/integration tests passing (`mvn clean test -pl paimon-cli`)
  - [x] SQL tests cover: SELECT, WHERE, LIMIT, GROUP BY, HAVING, DISTINCT, ORDER BY, JOIN, subquery, CASE WHEN, UNION, window functions (ROW_NUMBER)
  - [x] Manual verification via `paimon sql` against local filesystem warehouse

### DIFF
--- a/docs/docs/program-api/java-cli.md
+++ b/docs/docs/program-api/java-cli.md
@@ -1,0 +1,886 @@
+---
+title: "Java CLI"
+weight: 4
+type: docs
+aliases:
+- /program-api/java-cli.html
+---
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+# Java CLI
+
+Paimon provides a Java-based command-line interface (CLI) for interacting with Paimon catalogs and tables.
+The CLI supports reading, writing, and managing tables directly from the command line without requiring a compute engine like Flink or Spark.
+
+## Installation
+
+### Quick Install (from source)
+
+```shell
+cd paimon-cli
+./bin/install.sh
+```
+
+This builds the fat jar and installs the `paimon` command to `/usr/local/bin`. You can specify a custom prefix:
+
+```shell
+./bin/install.sh --prefix ~/.local
+```
+
+### Development Mode
+
+After building, use the bundled launcher directly without installing:
+
+```shell
+cd paimon-cli
+mvn clean package -DskipTests
+./bin/paimon --help
+```
+
+## Basic Usage
+
+Before using the CLI, create a catalog configuration file. By default, the CLI looks for a `paimon.yaml` file in the current directory.
+
+**Filesystem Catalog:**
+
+```yaml
+metastore: filesystem
+warehouse: /path/to/warehouse
+```
+
+**REST Catalog:**
+
+```yaml
+metastore: rest
+uri: http://localhost:8080
+warehouse: catalog_name
+```
+
+**Hive Catalog:**
+
+```yaml
+metastore: hive
+warehouse: /path/to/warehouse
+uri: thrift://localhost:9083
+```
+
+Run a command:
+
+```shell
+paimon <command> [options]
+```
+
+**Global Options:**
+
+- `-c, --config PATH`: Path to catalog configuration file (default: `paimon.yaml`)
+- `-h, --help`: Show help message and exit
+- `-v, --version`: Show version and exit
+
+## Catalog Commands
+
+### list-databases
+
+List all databases in the catalog.
+
+```shell
+paimon list-databases
+```
+
+Output:
+```
+default
+mydb
+analytics
+```
+
+### list-tables
+
+List all tables in a database.
+
+```shell
+paimon list-tables mydb
+```
+
+Output:
+```
+orders
+products
+users
+```
+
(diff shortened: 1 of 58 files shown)


JingsongLi (commented):
This is a big addition (9000+ lines, new module). A few high-level thoughts:

**Scope & Direction Questions:**

1. **Overlap with Python CLI**: The PR description includes a comparison table, but the fundamental question is: do we want to maintain **two** standalone CLIs (Java + Python) long-term? Each new feature would need to be implemented in both. What's the intended audience segmentation?

2. **Calcite dependency**: Adding Apache Calcite as a SQL engine brings significant dependency weight. The fat jar will be large. Have you measured the final jar size? Calcite also has its own version compatibility concerns with different Hadoop/Hive environments.

3. **Module structure**: Adding a top-level `paimon-cli` module to the root `pom.xml` means it's part of the main release artifacts. Should this be an experimental/optional module first?

**Technical Concerns:**

4. **SQL engine completeness**: Using Calcite's planner is powerful but means you need to handle Calcite's SQL dialect differences from Flink/Spark SQL. Users may expect the same SQL syntax to work across Flink SQL, Spark SQL, and this CLI. What's the compatibility story?

5. **Write path**: The CLI includes a `WriteCommand` — what's the commit protocol? Does it use the same `FileStoreCommit` as Flink/Spark? How are conflicts handled for concurrent writers?

6. **No PIP reference**: A new module of this scope typically goes through a PIP (Paimon Improvement Proposal) for community discussion. Has this been discussed on the mailing list?

I'd suggest:
- Start with a PIP discussion to align on scope and long-term maintenance plan
- Consider starting as a separate repository or `paimon-extras` module to avoid blocking releases
- Focus the first version on read-only operations (SQL queries, catalog inspection) before adding write support

cc @apache/paimon-committers for broader input on whether we want a Java CLI in the main repo.

hqlalala:
Thanks for the thorough review! Addressing each point:

  ### 1. Overlap with Python CLI

  The two CLIs serve different user segments, not a duplication problem. The Python CLI targets data scientists with `pip` environments; the Java CLI targets DevOps/SREs on production clusters where JDK is always available but Python often isn't.

  The feature sets are also not 1:1 — the Java CLI includes ops-oriented commands (expire-snapshots, orphan-clean, tag management, rollback) that the  Python CLI doesn't have. This mirrors how Flink, Spark, and Hive each maintain their own CLI despite overlapping SQL capabilities.

  ### 2. Calcite dependency

  Calcite 1.37.0 itself adds ~8MB — the cost for a full SQL planner with JOIN, subquery, window functions, and UNION support. This parallels the Python  CLI's use of DataFusion as its SQL engine.

  Final jar size: **70MB** (core only). S3/OSS cloud storage plugins are **not bundled** — they're loaded as optional plugins via install script  (`./install.sh --with oss`), same model as Python CLI's `pip install paimon[oss]`.

  Calcite runs embedded in the CLI process with its own classloader. It never touches the user's Hadoop/Hive classpaths, so there are no version  compatibility concerns with the cluster environment.

  ### 3. Module structure

  `paimon-python/` is already a top-level module in the main repo and part of releases. `paimon-cli` follows the same pattern. It has zero dependency on  Flink/Spark and won't affect their release cycles.

  I'm open to adding `<maven.deploy.skip>true</maven.deploy.skip>` initially so it's built and tested in CI but not published to Maven Central until it  stabilizes.

  ### 4. SQL engine completeness

  The CLI is a standalone diagnostic/management tool, not a Flink SQL or Spark SQL replacement. The `sql` subcommand explicitly states it uses Calcite's SQL dialect (configured with `"lex": "MYSQL"` mode, same as Python CLI's DataFusion default).

  Calcite covers standard SQL which is sufficient for the CLI's use case: ad-hoc queries for table inspection and debugging. Users who need  engine-specific syntax (e.g., Flink's temporal joins, Spark's TRANSFORM) should use those engines directly.

  ### 5. Write path

  `WriteCommand` uses the standard Paimon Java API: `table.newBatchWriteBuilder()` → `BatchTableWrite` → `prepareCommit()` → `commit()`. This goes  through the same `FileStoreCommit` with snapshot-based atomic commit that Flink and Spark use internally.

  The CLI is designed for single-user batch imports (CSV/JSON → table), not concurrent streaming writes — same scope as Python CLI's `paimon table  import` command, which also uses `write_builder.new_commit().commit()`.

  ### 6. PIP

  I wasn't aware this required a PIP — `paimon-python/` (including its CLI, DataFusion SQL engine, and write path) was added to the main repo without  one. If the community feels a PIP is needed for this module, I'm happy to draft one and start a discussion.

leaves12138:
Thanks for the large CLI contribution. I did a high-level pass over the module structure, command coverage, parser/printing utilities, and the test set. Since this is a 9k+ line new module with many commands and public UX decisions, I am not comfortable approving it in this broad form yet. I would suggest splitting follow-ups or getting explicit agreement on the CLI scope/UX first, especially around overlap with existing Python/SQL entry points, command stability, packaging, and which operations should be included in the first commit.

JingsongLi:
Such big 10000 lines PR, we should discuss in dev first.

PR #7977, opened 2026-05-26
[core] Log when scan uses global index

leaves12138:
### Purpose

Add an INFO log when data evolution scan evaluates filters through the global index and gets an index result.

### Changes

- Add a logger to `DataEvolutionBatchScan`.
- Log `Scan table '{}' with global index.` when global-index scan is actually used.

### Tests

- `mvn -pl paimon-core spotless:apply -DskipTests -DskipITs -Dscala-2.12`
- `mvn -pl paimon-core -DskipTests -DskipITs -Dscala-2.12 -Dcheckstyle.skip=false compile`


LsomeYeah (approved):
+1

PR #7979, opened 2026-05-26
[python][daft] Use external paths for native Parquet reads

QuakeWang:
### Purpose

Daft native Parquet reads always used `DataFileMeta.file_path` when creating Parquet tasks. This is incorrect for tables configured with `data-file.external-paths`, where the physical data file is stored in `DataFileMeta.external_path` while  `file_path` may point to the warehouse-derived table path.

### Tests

  - `pytest paimon-python/pypaimon/tests/daft/daft_datasource_test.py -q -rs`
  - `python -m pytest paimon-python/pypaimon/tests/daft/daft_datasource_test.py -q -rs`


JingsongLi:
+1

PR #7980, opened 2026-05-26
[python][daft] Fix predicate conversion for unsupported expressions

QuakeWang:
### Purpose

Daft predicate conversion used `None` to represent both unsupported expressions and literal values. As a result, comparison pushdown could treat unsupported expressions as literals, for example converting a cast expression into `id = None`.

String predicates also called `str(...)` on visited arguments, so column or unsupported arguments could be pushed as literal strings. This could mark filters as pushed and remove them from Daft remaining filters, changing query semantics.

This patch adds explicit internal markers for literals and unsupported expressions. Predicates are only pushed when the non-column side is a real scalar literal, and string predicates now require real string literals. Unsupported or mixed expressions stay in Daft remaining filters.

  ### Tests

  - `python -m pytest paimon-python/pypaimon/tests/daft/daft_data_test.py -k 'FilterPushdown'`
  - `python -m pytest paimon-python/pypaimon/tests/daft/daft_data_test.py`

### DIFF
--- a/paimon-python/pypaimon/daft/daft_predicate_visitor.py
+++ b/paimon-python/pypaimon/daft/daft_predicate_visitor.py
@@ -49,11 +49,26 @@ class _ColRef:
     name: str
 
 
+@dataclass(frozen=True, slots=True)
+class _Literal:
+    """Literal marker to keep real literal values separate from unsupported expressions."""
+
+    value: Any
+
+
+class _Unsupported:
+    pass
+
+
+_UNSUPPORTED = _Unsupported()
+
+
 class PaimonPredicateVisitor(PredicateVisitor[Any]):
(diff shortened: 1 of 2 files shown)


JingsongLi (approved):
LGTM. This is a well-designed safety fix. The core insight is correct: using `None` as a sentinel for both "unsupported expression" and "literal value" was a type confusion that could silently push incorrect predicates and remove them from Daft's remaining filter list — changing query semantics.

The `_Literal` / `_Unsupported` marker approach is clean and makes the type distinction explicit throughout the visitor. The `_is_pushable_scalar` guard also correctly rejects `None` literals and collection types from pushdown.

Test coverage for the mixed-expression scenarios is good.

+1

PR #7981, opened 2026-05-26
[python] Fix local-cache silently bypassed in REST catalog

XiaoHongbo-Hope:
### Purpose

  Two issues prevent the local cache from working safely in REST catalog mode:

  1. With `data-token.enabled=true` (DLF mode), enabling `local-cache.enabled=true`
     had no effect — reads silently went uncached.
  2. When `local-cache.max-size` was unset, the cache could grow without bound
     (default 2^63-1 bytes), risking disk full / OOM.

  This PR fixes them by:
  - Routing `RESTTokenFileIO` through `CachingFileIO.wrap_with_caching_if_needed`.
  - Defaulting `local-cache.max-size` to 256 MB (memory) / 10 GB (disk) when unset.

  ### Tests
  - `test_file_io_for_data_wraps_cache_when_data_token_enabled`
  - `test_default_memory_cache_max_size_capped`
  - `test_default_disk_cache_max_size_capped`

### DIFF
--- a/paimon-python/pypaimon/catalog/rest/rest_catalog.py
+++ b/paimon-python/pypaimon/catalog/rest/rest_catalog.py
@@ -646,16 +646,18 @@ class RESTCatalog(Catalog):
         Get FileIO for data access, supporting FUSE local path mapping.
         """
         if self._fuse_resolver is not None:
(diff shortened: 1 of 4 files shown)


JingsongLi:
+1

PR #7983, opened 2026-05-26
[python] Fix CachingFileIO bypassing delegate for several FileIO methods

XiaoHongbo-Hope:
### Purpose

  CachingFileIO does not forward several FileIO methods to the delegate,
  so the wrapper hits FileIO base's no-op / raising defaults instead.

  Observed failures:
  - PyArrow data reads fail with
    "Expected an S3 object path of the form 'bucket/key', got 'oss://...'"
    (to_filesystem_path returns the input unchanged)
  - Writes fail with
    "NotImplementedError: write_parquet must be implemented by FileIO subclasses"
    (write_parquet/orc/avro/lance/blob/vortex and try_to_write_atomic are not forwarded)

  ### Tests
  - test_to_filesystem_path_forwarded_to_delegate
  - test_write_parquet_when_enable_local_cache

### DIFF
--- a/paimon-python/pypaimon/filesystem/caching_file_io.py
+++ b/paimon-python/pypaimon/filesystem/caching_file_io.py
@@ -382,9 +382,30 @@ class CachingFileIO(FileIO):
     def is_dir(self, path: str) -> bool:
         return self._delegate.is_dir(path)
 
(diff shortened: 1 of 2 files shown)


JingsongLi:
+1

PR #7984, opened 2026-05-26
[python] Remove tests in the build artifact

aisk:
### Purpose
Currently, the `tests` folder is included in the generated package uploaded to PyPI server:

{82B2247E-71C9-42C7-B588-706665221CDD}

It contains more than 100 extra files and increases the package size by about 1.5 MB. These files are also installed on users' machines, but they are not needed. We should exclude the `tests` folder from the package.


JingsongLi:
+1

PR #7985, opened 2026-05-26
[core] Add access context for blob view fallback

leaves12138:
### Purpose

Blob view resolution may need to access an upstream blob table while users are querying a downstream blob-view table. Permission systems need to distinguish this fallback access from a direct upstream table query.

### Changes

- Add `TableAccessContext` to describe table access origins.
- Add context-aware `Catalog#getTable` and `Catalog#authTableQuery` default overloads.
- Carry table access context through `CatalogEnvironment` so scans can pass the same context to query auth.
- Pass `BLOB_VIEW_FALLBACK` context from `DataEvolutionTableRead` / `BlobViewLookup` when resolving upstream blob descriptors and URI readers.
- Add a focused test to verify access context reaches query auth.

### Tests

- `mvn -pl paimon-core -DskipITs -Dspotless.check.skip=true -Dcheckstyle.skip=true -Dtest=CatalogAccessContextTest test`
- `mvn -pl paimon-core -DskipITs -DskipTests -Dspotless.check.skip=true -Dcheckstyle.skip=true test-compile`

### DIFF
--- a/docs/static/rest-catalog-open-api.yaml
+++ b/docs/static/rest-catalog-open-api.yaml
@@ -3167,6 +3167,12 @@ components:
           type: array
           items:
             type: string
+        accessType:
+          type: string
+          description: Type of table access, omitted for direct table access.
+        fallbackFrom:
+          type: string
+          description: Source table for fallback access, omitted for direct table access.
     AuthTableQueryResponse:
       type: object
       properties:
--- a/paimon-api/src/main/java/org/apache/paimon/rest/RESTApi.java
+++ b/paimon-api/src/main/java/org/apache/paimon/rest/RESTApi.java
@@ -809,7 +809,27 @@ public class RESTApi {
      */
     public AuthTableQueryResponse authTableQuery(
             Identifier identifier, @Nullable List<String> select) {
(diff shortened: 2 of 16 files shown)


JingsongLi (approved):
LGTM. Clean design for distinguishing blob-view fallback access from direct table queries. This is important for permission systems that need to know the access origin.

The approach is well-structured:
- `TableAccessContext` is a simple, serializable value type with factory methods
- Default method overloads on `Catalog` interface maintain backwards compatibility
- `CatalogEnvironment` carries the context through to scans and blob resolution
- `DelegateCatalog` and `PrivilegedCatalog` properly propagate the new overloads

One minor thought: the `TableAccessContext.direct()` factory creates a new instance each call. Since it's immutable and always the same, you could cache it as a static constant to avoid allocation in hot paths (e.g., `private static final TableAccessContext DIRECT = new TableAccessContext(AccessType.DIRECT, null)`). Not a blocker.

+1

JingsongLi (changes requested):
This PR needs more discussion about REST integration.

leaves12138:
I re-checked the latest state. This still needs the REST integration discussion from the existing requested-changes review before I would consider it ready.

PR #7987, opened 2026-05-27
[common] Fix SubstringTransform null check for field refs

QuakeWang:
### Purpose

`SubstringTransform` checked source field nullability with column `0`, but read the value from `sourceFieldRef.index()`. When column `0` was null and the actual source field was non-null, substring masking incorrectly returned null.

This PR uses the source field index for both the null check and value access, and adds a regression test for that case.

### Tests

`mvn -pl paimon-common -am -Pfast-build -DfailIfNoTests=false -Dtest=SubstringTransformTest test`


JingsongLi (approved):
LGTM. The bug is clear: `row.isNullAt(0)` was always checking column 0 instead of the actual source field index, causing incorrect null returns when column 0 happened to be null. The fix correctly uses `sourceFieldRef.index()` for both the null check and value access. Regression test covers the exact scenario.

+1

PR #7988, opened 2026-05-27
[lumina] Support null vectors in Lumina index writer

jerry-024:
Skip null vectors during index build instead of throwing or stopping the shard. Non-null vectors continue to be written with correct shard-relative IDs (containing gaps at null positions).

- Writer: write(null) advances logicalRowId without writing, tracks contiguous non-null segments for correct ID mapping in getNextBatch
- Writer: add NaN/Infinity validation for vector elements
- Flink caller: remove null break, use resultEntries.isEmpty() for commit skip, rename rowsWritten to rowsSeen
- Tests: null gap IDs, all-null shard, pre-filter with gap IDs, null at start/end boundaries, NaN/Infinity rejection

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/globalindex/ResultEntry.java
+++ b/paimon-common/src/main/java/org/apache/paimon/globalindex/ResultEntry.java
@@ -24,7 +24,10 @@ import javax.annotation.Nullable;
 public class ResultEntry {
 
     private final String fileName;
+
+    /** Total logical rows processed by the writer, including null/skipped rows. */
     private final long rowCount;
+
     @Nullable private final byte[] meta;
 
     public ResultEntry(String fileName, long rowCount, @Nullable byte[] meta) {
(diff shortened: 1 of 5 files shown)


JingsongLi (approved):
LGTM. Well-designed change to handle null vectors gracefully.

**Key design points that look correct:**

1. **Segment-based ID mapping**: The `segments` list (tracking contiguous non-null runs with their start offset) is an efficient way to map physical vector indices back to logical row IDs with gaps. This avoids needing a full ID array in memory.

2. **`ResultEntry.rowCount()` now reports `logicalRowId`** (total rows seen including nulls) rather than just non-null count — this is important for correct downstream join with the data rows.

3. **NaN/Infinity validation**: Good addition. These values would silently corrupt the HNSW index structure, so rejecting them early with a clear error message is the right call.

4. **Flink caller change**: Replacing the null-break with pass-through is correct. The old behavior (stop the shard on first null) was too aggressive and would lose data. Using `resultEntries.isEmpty()` for the skip decision is the right boundary check.

5. **Test coverage**: Comprehensive — null gaps, all-null shard, boundary nulls (start/end), pre-filter with gap IDs, NaN/Infinity rejection. All the edge cases are covered.

One minor note: in `getNextBatch`, the segment exhaustion check throws `IllegalStateException` — this should never happen in practice since `cursor < totalCount` guards the loop. Good defensive check.

+1

PR #7989, opened 2026-05-27
[python] Add $buckets system table

JunRuiLee:
## Summary

pypaimon currently implements 8 system tables: `$snapshots`, `$schemas`, `$options`, `$manifests`, `$files`, `$partitions`, `$tags`, `$branches`. Compared to the Java side, it still lacks `$buckets`, `$audit_log`, `$read_optimized`, `$consumers`, `$statistics`, `$aggregation_fields`, `$file_key_ranges`, `$table_indexes`, etc.

This PR adds `$buckets`, which is one of the more commonly used system tables for diagnosing data skew. It aggregates manifest entries by (partition, bucket) and exposes per-bucket record_count, file_size, file_count, and last_update_time.

## Changes
- New: `buckets_table.py` — `BucketsTable` implementation
- New: `buckets_table_test.py` — 5 end-to-end tests (schema validation, empty snapshot, aggregation correctness, sort order, catalog dispatch)
- Modified: `system_table_loader.py` — register `"buckets"`
- Modified: `system_table_loader_test.py` — update expected table list

### DIFF
--- a/paimon-python/pypaimon/table/system/buckets_table.py
+++ b/paimon-python/pypaimon/table/system/buckets_table.py
@@ -1,0 +1,165 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+"""The ``$buckets`` system table — per-bucket aggregated stats."""
+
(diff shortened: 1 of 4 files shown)


leaves12138 (changes requested):
Thanks for adding the Python `$buckets` system table. The Python-side implementation and tests look reasonable, but this PR currently reverts the recent `SubstringTransform` fix from #7987.

In `paimon-common/src/main/java/org/apache/paimon/predicate/SubstringTransform.java`, the diff changes the null check from the referenced source field index back to column `0`:

```java
sourceString = row.isNullAt(0) ? null : row.getString(sourceFieldRef.index());
```

and it also removes `testSubstringRefInputUsesSourceFieldNullability`. This reintroduces the bug where `SUBSTRING(FieldRef(index > 0), ...)` returns `null` whenever column 0 is null, even if the actual referenced source column is non-null. That is unrelated to `$buckets` and would regress existing transform behavior.

Please rebase/merge current master and keep #7987's `sourceIndex` fix and test, or remove the Java `SubstringTransform` changes from this PR. I ran the new Python tests locally and they passed:

```bash
PYTHONPATH=. python -m pytest -q pypaimon/tests/system/buckets_table_test.py pypaimon/tests/system/system_table_loader_test.py
# 10 passed
```

JunRuiLee:
@leaves12138 Thanks for catching this! Rebased onto latest master — the SubstringTransform changes are no longer included in this PR.

JingsongLi:
+1

PR #7990, opened 2026-05-27
[python][daft] Add public API integration tests

QuakeWang:
### Purpose

The Daft public facade was not covered by integration tests, so regressions in `pypaimon.daft.read_paimon()` / `write_paimon()` could be missed even though the internal `_read_table()` and `_write_table()` paths were tested.

This PR adds public API integration coverage for read/write round trips, projection, filter, limit, snapshot/tag time travel, overwrite, and mutual exclusion of `snapshot_id` and `tag_name`. It also makes the public `read_paimon()` entry point reject conflicting time-travel parameters before loading the catalog/table.

  ### Tests

  - `pytest pypaimon/tests/daft/daft_integration_test.py`
  - `pytest pypaimon/tests/daft/daft_integration_test.py pypaimon/tests/daft/daft_data_test.py pypaimon/tests/daft/daft_sink_test.py`

### DIFF
--- a/paimon-python/pypaimon/daft/daft_paimon.py
+++ b/paimon-python/pypaimon/daft/daft_paimon.py
@@ -143,6 +143,11 @@ def read_paimon(
     Returns:
         A lazy ``daft.DataFrame`` backed by this Paimon table.
     """
+    if snapshot_id is not None and tag_name is not None:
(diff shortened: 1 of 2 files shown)


XiaoHongbo-Hope:
+1

PR #7995, opened 2026-05-27
[core] Improve sort buffer overflow message

huangxiaopingRD:
### Purpose
Improve the error message thrown when a single record exceeds the available sort buffer in `BinaryExternalSortBuffer`, so users can more easily understand that the failure is caused by an oversized row and know to check the input data or increase `write-buffer-size`.
### Tests
None


JingsongLi (approved):
+1

PR #7996, opened 2026-05-27
[core] support adding blob column through comments

steFaiz:
### Purpose
This PR is about to replace https://github.com/apache/paimon/pull/7921

This PR introduces a comment format to mark an adding column as BlobType. So that these engines which do not have BlobType can also add blob columns. 
The rule is: 
1. define 2 directives: `__BLOB_FIELD`, `__BLOB_DESCRIPTOR_FIELD`
2. user comments are seperated by `;`: e.g. `__BLOB_FIELD; this is a blob field`, The starting directives will be trimmed.
3. blob-field and blob-descriptor-field are immutable, and will be atomically modified during `Alter Table Add Column` and `Drop Column`.

**NOTE**:
1. we do not permit updating any existing non-blob fields to blob type nor updating any blob field to non-blob type.
2. we do not permit adding an blobType field without directives, since we don't know how to store them.

### Tests
ITCases and UnitTests

### DIFF
--- a/docs/docs/append-table/blob.mdx
+++ b/docs/docs/append-table/blob.mdx
@@ -268,6 +268,56 @@ CREATE TABLE image_table (
 
 </Tabs>
 
+### Adding a Blob Column to an Existing Table
+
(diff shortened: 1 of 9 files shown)


JingsongLi (approved):
+1

PR #7997, opened 2026-05-27
[python] Fix duplicate /snapshot REST call in plan flow

XiaoHongbo-Hope:
### Purpose

  A single plan issues 2 GET /snapshot round-trips to the metastore because
  file_scanner.manifest_scanner and GlobalIndexScanner.create independently
  fetch latest snapshot. 

**This PR fixes:**
  - perf: eliminates the duplicate REST round-trip (~150ms saved per plan)
  - correctness: under time-travel scan, global index was evaluated
    against latest snapshot while manifest used the time-travel snapshot.

  ### Backward compatible — no design break

  `GlobalIndexScanner.create()` stays self-contained for ad-hoc / standalone
  use. The new `snapshot` argument is **optional**.

  Tests

`PlanSnapshotFetchRegressionTest.test_plan_fetches_latest_snapshot_only_once`

### DIFF
--- a/paimon-python/pypaimon/globalindex/global_index_scanner.py
+++ b/paimon-python/pypaimon/globalindex/global_index_scanner.py
@@ -82,12 +82,16 @@ class GlobalIndexScanner:
         return GlobalIndexEvaluator(fields, readers_function, self._executor)
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


JingsongLi:
+1

PR #7999, opened 2026-05-27
[core] Don't allow full compaction with lookup changelog producer

ArnavBalyan:
### Purpose
 - Currently there is no check against full compaction on lookup changelog producer. 
 - User reported unexpected behaviour in #6183
 - Disable setting both the properties together instead of the silent behaviour.

### Tests
 - UT
 - CI


ArnavBalyan:
cc @JingsongLi thanks!

leaves12138:
Thanks for the fix. I am holding off on approval for now because the current CI status has failing build/build_test jobs. Please fix or rerun the failures first.

ArnavBalyan:
Thank you so much for the review! The failures seem unrelated, would it be possible to rerun them

JingsongLi:

PR #8000, opened 2026-05-27
[global-index-eslib]  integrates the Elasticsearch (Lucene) index engine into Paimon's global index system

CrownChu:
Summary
  
  This PR integrates the Elasticsearch (Lucene) index engine into Paimon's global index system via the paimon-eslib module, enabling Paimon tables to directly leverage the ES engine's vector search (DiskBBQ / HNSW) and scalar/text
  filtering capabilities. Key changes:

  - Parallel cluster search: Inject a shared ExecutorService from ESIndexGlobalIndexerFactory for DiskBBQ search, replacing per-reader thread pool creation
  - Concurrent close safety: Add a volatile closed flag with checkNotClosed() to fix a race condition in the reader lifecycle
  - Full scalar/text filtering: Implement all GlobalIndexReader visitor methods (visitEqual, visitLessThan, visitStartsWith, visitLike, visitIsNull, etc.), dispatching to ESLib's unified IndexFilter / ScalarPredicate API
  - Dependency coordinates: Change the eslib dependency groupId from org.elasticsearch to io.github.crownchu, version from 1.0.0-SNAPSHOT to 1.0.0
  - Public dependency repository: Add a GitHub-hosted Maven repository for public CI dependency resolution

  Details

  Paimon Integration with the ES Index Engine

  Paimon's global index is bridged to the ES (Lucene) engine through paimon-eslib: on the write side, Flink uses this engine to build vector/scalar indexes (producing ESLib archive files); on the query side, ES's paimon-store mounts
  and reuses the same engine. This PR completes the query-side engine's parallel search, concurrency safety, and filter operators.

  Parallel Search

  The search thread pool lifecycle is owned by the factory layer (ESIndexGlobalIndexerFactory), lazily initialized and shared across all readers. Configurable via global-index.es-index.read-search-threads:

  - -1 (default): auto = CPU/2, min 2
  - 0: disable parallel search (serial only)
  - >0: use the specified thread count

  The executor is injected through the full chain: Factory → Indexer → Reader → ESIndexSearcher → Lucene Codec (via the SearchExecutorHolder ThreadLocal bridge, to work around Lucene SPI's no-arg constructor constraint).

  Scalar Filter
  
  ESIndexGlobalIndexReader now implements all visitor methods, dispatching to ESLib's unified filter API:

  - Numeric comparisons → ScalarPredicate.eq/lt/lte/gt/gte/in/notIn
  - Text matching → IndexFilter.TextFilter with TERM / PREFIX / WILDCARD ops
  - Null checks → IndexFilter.exists() / notExists()

  Dependency Publishing
  
  ESLib jars are published to a GitHub raw repository (CrownChu/es-paimon-lib-releases) with full Maven metadata and checksums. paimon-eslib/pom.xml declares the repository so CI can resolve dependencies without manual local
  installation.


leaves12138:
Thanks for the contribution. This is a large new global-index integration and the current CI status has many failing jobs, so I am holding off on approval for now. Please get the build/test matrix green first, then it will be easier to do a meaningful code review.

paimon-eslib/src/main/java/org/apache/paimon/eslib/index/ESIndexGlobalIndexer.java
```diff
@@ -0,0 +1,67 @@
...
+    public GlobalIndexWriter createWriter(GlobalIndexFileWriter fileWriter) throws IOException {
+        return new ESIndexGlobalIndexWriter(fileWriter, fields, indexOptions);
+    }
+
+    @Override
+    public GlobalIndexReader createReader(
+            GlobalIndexFileReader fileReader,
+            List<GlobalIndexIOMeta> files,
+            ExecutorService executor) {
+        return new ESIndexGlobalIndexReader(fileReader, files, fields, indexOptions, executor);
```
JingsongLi, line 60: The configured read/search executor is dropped here. `ESIndexGlobalIndexerFactory` creates and passes the pool controlled by `global-index.es-index.read-search-threads`, but `ESIndexGlobalIndexer` only stores it and then constructs the reader with the caller-provided `executor` instead. As a result the new option is silently ignored: setting it to `0` does not disable async/searcher execution, and setting a custom size does not affect the ESLib `searcher.load(..., searchExecutor)` or the reader `async` path. Could we pass the stored `searchExecutor` into `ESIndexGlobalIndexReader` here (or otherwise merge the two executors deliberately) and add a small test so the option is exercised?

CrownChu: fixed in 4f39eafcda.
  createReader now passes the factory-resolved searchExecutor straight to ESIndexGlobalIndexReader instead of the caller-provided executor, so global-index.es-index.read-search-threads is authoritative:
  - 0 → factory returns null → reader runs serially and searcher.load(..., null) / the async path are both disabled;
  - N (>0) → an N-thread pool is used;
  - unset / -1 → the default CPU/2 pool.
  The caller-supplied executor is intentionally not substituted when the configured one is null, otherwise 0 could not actually disable async execution (the previous behavior silently fell back to the caller's pool).
  Added ESIndexGlobalIndexerExecutorTest to exercise the option:
  - the configured executor (not the caller's) reaches the reader;
  - read-search-threads=0 leaves the reader serial (null executor).

paimon-eslib/src/main/java/org/apache/paimon/eslib/index/ESIndexGlobalIndexReader.java
```diff
@@ -0,0 +1,797 @@
...
+import org.apache.paimon.predicate.FieldRef;
+import org.apache.paimon.predicate.FullTextSearch;
+import org.apache.paimon.predicate.VectorSearch;
+import org.apache.paimon.types.DataField;
+import org.apache.paimon.utils.RoaringNavigableMap64;
+
+import org.elasticsearch.eslib.api.ArchiveDataProvider;
+import org.elasticsearch.eslib.api.ESIndexSearcher;
+import org.elasticsearch.eslib.api.model.FieldIndexConfig;
+import org.elasticsearch.eslib.api.model.FullTextParams;
```
JingsongLi, line 46: This class is not present in the declared `io.github.paimon.eslib:eslib-core:1.0.0:lucene9` artifact, so the new module does not compile. With JDK 11 and the profile enabled, `mvn -pl paimon-eslib -Ppaimon-eslib,fast-build -DfailIfNoTests=false test` fails with `cannot find symbol: class FullTextParams`; `jar tf eslib-core-1.0.0-lucene9.jar` only contains `FieldIndexConfig`, `IndexFilter`, `ScalarPredicate`, `SearchResult`, etc. Please either publish/use an eslib version that contains `FullTextParams` and the matching `fullTextSearch` overload, or adapt this code to the API available in `1.0.0` so the optional module can compile.

CrownChu: Thanks — root cause was that the earlier change reused version 1.0.0 for eslib-core, so consumers (and CI) kept resolving a stale cached 1.0.0 jar that predates FullTextParams. Republishing over the same version is unreliable:
  Maven caches release versions in ~/.m2 and never re-downloads them.
  Fixed by cutting a new version instead of overwriting:
  - Bumped io.github.paimon.eslib:eslib-core / eslib-simdvec to 1.0.1, which now contains FullTextParams and the matching fullTextSearch(String, String, int, FullTextParams) overload.
  - Published 1.0.1 (jar + pom + checksums) to the OSS repo; verified the artifact contains FullTextParams (jar tf … | grep FullTextParams → present, sha1 4e2e2be3…).
  - Updated paimon-eslib/pom.xml to depend on 1.0.1 for both eslib-core and eslib-simdvec.
  - Removed the dead eslib-releases GitHub-raw repository entry (it was empty / 404), so the module now resolves from a single, populated source.
  Verified locally with the profile enabled:
  mvn -pl paimon-eslib -Ppaimon-eslib,fast-build -DfailIfNoTests=false test → compiles and passes (9 tests).
  A new version number forces every environment (CI, local .m2) to pull the fresh jar, so the cannot find symbol: FullTextParams failure no longer reproduces.
  (Commits: paimon 3fcbefb bump to 1.0.1, 2badffe drop dead repo; es-paimon-lib d0de40e release 1.0.1.)

paimon-eslib/src/main/java/org/apache/paimon/eslib/index/ESIndexGlobalIndexReader.java
```diff
@@ -0,0 +1,797 @@
...
+     * serialized to JSON and fed to the text parser, which would silently search for the literal
+     * JSON tokens.
+     */
+    private static org.apache.paimon.predicate.FullTextQuery.Match requireMatch(
+            FullTextSearch fullTextSearch) {
+        org.apache.paimon.predicate.FullTextQuery ftq = fullTextSearch.query();
+        if (ftq instanceof org.apache.paimon.predicate.FullTextQuery.Match) {
+            return (org.apache.paimon.predicate.FullTextQuery.Match) ftq;
+        }
+        throw new UnsupportedOperationException(
```
jerry-024: This makes `es-index` fail for structured full-text queries. `FullTextReadImpl` can pass `Phrase`, `BooleanQuery`, `Boost`, and `MultiMatch` through to any backend that advertises `supportsFullTextSearch()`, but this reader throws for everything except `Match`. That means valid full-text reads against ES splits fail at runtime instead of being evaluated or falling back. Either implement these query shapes, avoid advertising support for queries this backend cannot serve, or make the planner/read path fall back before reaching this exception.

paimon-eslib/src/main/java/org/apache/paimon/eslib/index/ESIndexGlobalIndexerFactory.java
```diff
@@ -0,0 +1,123 @@
...
+    private static volatile ExecutorService readSearchExecutor;
+    private static final Object LOCK = new Object();
+
+    @Override
+    public String identifier() {
+        return IDENTIFIER;
+    }
+
+    @Override
+    public boolean supportsFullTextSearch() {
```
jerry-024, line 42: The comment says FULLTEXT extra fields are supported, but the current full-text scan path only selects indexes whose queried text column is the primary `indexFieldId`. A hybrid index with vector as primary and text as an extra field will not produce `IndexFullTextSearchSplit`s for that text column, so full-text search on the extra field returns no indexed results. The scanner needs to include matching `extraFieldIds` as well, or this support claim should be narrowed to primary full-text fields only.

paimon-eslib/src/main/java/org/apache/paimon/eslib/index/ESIndexGlobalIndexerFactory.java
```diff
@@ -0,0 +1,123 @@
...
+
+    /**
+     * Returns the shared read/search thread pool. Default (unset or -1) creates a pool sized to
+     * CPU/2. Set to 0 to disable parallel search (returns null → serial only).
+     */
+    private static ExecutorService getOrCreateReadSearchExecutor(Options options) {
+        int threads = options.getInteger(READ_SEARCH_THREADS_KEY, DEFAULT_READ_SEARCH_THREADS);
+        if (threads == 0) {
+            return null;
+        }
```
jerry-024: This static executor is first-use-wins. Once one reader creates `readSearchExecutor`, later ES indexers with a different `global-index.es-index.read-search-threads` value reuse the old pool, and even a later `0` cannot disable async search if the pool already exists. Since this option is table/read configuration, the factory should either scope/cache executors by resolved thread count or make the first-use global behavior explicit and reject conflicting values.

jerry-024 (commented):
Inline review findings.

paimon-eslib/src/main/java/org/apache/paimon/eslib/index/ESIndexGlobalIndexerFactory.java
```diff
@@ -0,0 +1,123 @@
...
+        return IDENTIFIER;
+    }
+
+    @Override
+    public boolean supportsFullTextSearch() {
+        // The ES index is a hybrid backend: a FULLTEXT-typed column (whether it is the primary
+        // index field or an extra field of a multi-column index) is served by Lucene full-text
+        // search. Without this override FullTextScanImpl filters out every es-index file and the
+        // planner full-text path returns nothing.
+        return true;
```
jerry-024, line 43: This is broader than the reader can safely support. String fields default to KEYWORD in `ESIndexOptions` unless an analyzer/type=fulltext is configured, but `FullTextScanImpl` only knows `supportsFullTextSearch()` at the index-type level. For an `es-index` on a KEYWORD string column, full-text scan will treat the indexed range as covered, produce index splits, and then `visitFullTextSearch()` returns `Optional.empty()` for that field; because the range is considered indexed, no raw split is added, so the full-text query can produce false negatives. This should be gated on actual field config/metadata or avoid claiming full-text support for all ES index files.

paimon-eslib/src/main/java/org/apache/paimon/eslib/index/ESIndexOptions.java
```diff
@@ -0,0 +1,228 @@
...
+
+/**
+ * Parses Paimon Options into ESLib FieldIndexConfig for each field.
+ *
+ * <p>Options format (in table properties):
+ *
+ * <pre>
+ *   global-index.es-index.fields.vector_field.algorithm = diskbbq
+ *   global-index.es-index.fields.vector_field.dimension = 128
+ *   global-index.es-index.fields.text_field.analyzer = standard
```
jerry-024, line 55: The documented option keys here do not match what the parser reads below. The examples use `global-index.es-index.fields.<field>...`, but `parseFieldConfig()` builds keys from `FIELDS_PREFIX = "fields."`, so users following this comment will have analyzer/algorithm/dimension/type silently ignored. This also differs from the native vector parser, which accepts index-type-level keys separately and then field-level `fields.<field>...` overrides.

jerry-024 (commented):
Additional compatibility/behavior consistency findings.

paimon-eslib/src/main/java/org/apache/paimon/eslib/index/ESIndexGlobalIndexReader.java
CrownChu: Fixed in 9212f48b
  visitFullTextSearch now recursively maps the paimon FullTextQuery tree to an eslib FullTextQuerySpec and executes it via searcher.fullTextSearch(spec, topK). All shapes are implemented, not rejected:
  - Match → analyzed boolean of term/fuzzy queries (operator/boost/fuzziness honoured)
  - Phrase → QueryBuilder.createPhraseQuery (analyzer-aware, uses indexed positions)
  - BooleanQuery → Lucene BooleanQuery (must/should/must_not, recursive)
  - Boost → FunctionScoreQuery.boostByQuery (Lucene 9 replacement for BoostingQuery)
  - MultiMatch → expanded to a SHOULD-bool of per-column matches (per-column boost)
  The only remaining throw is a defensive guard for an unknown FullTextQuery subtype, which the framework never produces. This required eslib-core 1.0.2 (adds FullTextQuerySpec + the structured fullTextSearch overload +
  lucene-queries for the boost query). 
  Covered by E2E tests (e.g. Phrase order semantics, operator AND vs OR, fuzziness).

paimon-eslib/src/main/java/org/apache/paimon/eslib/index/ESIndexGlobalIndexerFactory.java
CrownChu: Fixed in 9212f48b.
  FullTextScanImpl now matches a searched text column against both the primary indexFieldId and the extraFieldIds (via matchedTextColumnIds), in both the index-file filter and the per-column grouping. So a hybrid index (vector primary + text extra) now produces FullTextSearchSplits for the extra text column, and full-text search on it returns indexed results. Single-column indexes are unaffected (extraFieldIds == null → only the primary matches).
  The read path already reconstructs a multi-column indexer (FullTextReadImpl → factory.create(indexField, extraFields)), which requires a multi-column-capable backend; TestFullTextGlobalIndexerFactory was extended to support
  multi-column so the framework test covers it.
  Tests: FullTextSearchBuilderTest.testFullTextSearchServesTextColumnAsExtraField (extra text column is searchable) and testFullTextSearchSkipsIndexNotCoveringTextColumn (an index not covering the text column still returns empty).
  The previous testFullTextSearchRequiresTextColumnAsPrimaryField is replaced accordingly — the "primary only" contract is intentionally relaxed (discussed and agreed with the reviewer).

paimon-eslib/src/main/java/org/apache/paimon/eslib/index/ESIndexGlobalIndexerFactory.java
CrownChu: Thanks for pushing on this — I traced the full-text path end-to-end and this does not cause a false negative; the behavior is correct.

Full-text is index-only: FullTextScanImpl adds no raw/unindexed split (unlike the vector path's RawVectorSearchSplit/unindexedRanges), so there is no raw split that "coverage" could suppress. FullTextReadImpl unions per-column  results (or) and skips Optional.empty(). So an es-index whose queried column has no analyzed (full-text) index just contributes an empty split to the union — it can't hide matches from a real full-text index on the same column/range, and full-text search on a non-analyzed column returns empty, which is the only correct answer. The "no raw split → false negative" reasoning applies to the vector path, not here.

  Separately, with the multi-field change a FULLTEXT column now also carries a <col>.keyword sub-field, so the same column serves full-text match (analyzed field) and exact =/IN/prefix (keyword sub-field).

paimon-eslib/src/main/java/org/apache/paimon/eslib/index/ESIndexOptions.java
CrownChu: thanks, the javadoc was wrong, the parser convention is right. The canonical field-config convention (native vector parser, and Morax/Lumina's AbstractIncrementalGlobalIndexBuilder/LuminaIndexOptions — e.g.
  fields.vec.distance.metric, lumina.index.dimension) is field-level fields.<field>.<key> overriding index-type-level <indexType>.<key>, not global-index.es-index.fields.... Fixed in 393a0e8dd0:
  - corrected the javadoc to fields.<field>.<key>;
  - added index-type-level es-index.<key> keys as defaults with field-level override (mirrors the native vector parser), via a resolve() helper;
  - added ESIndexOptionsTest (field-level read, type-level default + field override, long global-index.* keys are not read, keyword_subfield toggle).

paimon-core/src/main/java/org/apache/paimon/table/source/FullTextScanImpl.java
```diff
@@ -101,16 +101,20 @@ public Plan scan() {
...
-            byColumnAndRange
-                    .computeIfAbsent(columnName, k -> new HashMap<>())
-                    .computeIfAbsent(range, k -> new ArrayList<>())
-                    .add(indexFile);
+            for (int columnId : matchedTextColumnIds(meta, textColumnIds)) {
+                String columnName = checkNotNull(idToColumn.get(columnId));
+                byColumnAndRange
+                        .computeIfAbsent(columnName, k -> new HashMap<>())
+                        .computeIfAbsent(range, k -> new ArrayList<>())
+                        .add(indexFile);
```
JingsongLi: This grouping can put multiple unrelated full-text-capable index definitions into the same split when they cover the same text column and row range. For example, a table can have a normal full-text index whose primary field is `content` and a hybrid/vector ES index whose primary field is another column but has `content` in `extraFieldIds`; both match `matchedTextColumnIds(...)` and end up in the same `FullTextSearchSplit`. `FullTextReadImpl` then picks the first file index type/meta and passes the whole list to one backend reader, but `ESIndexGlobalIndexReader` still requires `files.size() == 1` and will fail for this valid layout. Please keep separate splits/readers per index identity (for example index type + primary field + extra field set) or otherwise avoid merging files from different global index definitions into one reader input.

CrownChu: Thanks，this was a real defect. Fixed by making the grouping identity-aware instead of grouping purely on (column, range):                                                                                                                                                                                                                                                                                                                       
 - Added an index-definition identity = indexType | primary indexFieldId | sorted extraFieldIds (IndexIdentity).                                                                                                                      
 - chooseIndexPerColumn now selects exactly one index definition to serve each searched text column, preferring the index where the column is the primary indexFieldId (a dedicated full-text index) over one that merely carries it  
 as an extra field, with a deterministic tie-break for stability across planning runs.                                                                                                                                                
 - The grouping loop only adds a file to a column's split when its identity matches the chosen one, so every IndexFullTextSearchSplit — and therefore the file list handed to a single backend reader — contains files from a single index identity. This keeps ESIndexGlobalIndexReader's files.size() == 1 invariant intact for the exact layout you described (a content full-text index alongside a vector index that has content in extraFieldIds).                                                                                                                                                                                                                                    
 On "separate splits per identity vs. pick one": I went with picking one per column deliberately, because serving the same text column through two different indexes would double-count / return overlapping hits for that column. If 
 you'd rather emit both as independent splits and dedup downstream, I'm happy to switch — but for single-column full-text semantics choosing one index seems correct. PTAL.

CrownChu:
@leaves12138 @jerry-024 @JingsongLi 
[Discussion] Global Index: how should es-index declare per-column index type / vector algorithm — per-type identifiers vs. single type + options?
  ---
Context
  
The es-index global index (added on feature-globalindex-support-multi-eslib, docs in the commit below) is a Lucene-backed, multi-column, multi-modal index. A single es-index file indexes a primary column (typically a vector) together with optional companion columns — full-text, keyword, scalar, and date — so vector search and full-text search can be served from one index definition.
This differs structurally from the existing single-purpose global indexes:
  - paimon-vector (ivf-flat, ivf-pq, ivf-hnsw-flat, ivf-hnsw-sq, lumina): one column, one algorithm per index. The index-type identifier already is the algorithm.                                                                       
  - tantivy-fulltext: one text column, one engine.
    Today es-index registers a single GlobalIndexerFactory (identifier es-index). The vector algorithm and every column's sub-index type are chosen through options under the global-index.es-index. prefix (field-level                   
  global-index.es-index.fields.<field>.<key> overriding index-type-level global-index.es-index.<key>).
  Docs / reference commit: <link to commit b081205bfb / PR>
  ---
  The question  
  How should users declare the index type and vector algorithm for es-index? Two shapes are on the table.
  Design A — per-type identifiers (aligned with paimon-vector)
  Encode the algorithm in the index-type identifier: es-hnsw, es-diskbbq, es-native, … Each is its own SPI identifier; adding an algorithm means adding a placeholder type. No option is needed to pick the algorithm.
  Pros
  - Self-describing and consistent with paimon-vector's ivf-* naming.
  - Simpler mental model for the common single-vector-column case.
  
  Open problems for a multi-modal, multi-column index                                                                                                                                                                                    
  - How do companion columns (full-text / keyword / date) get their sub-index type when the identifier only names the vector algorithm?                                                                                                  
  - How do you express two vector columns wanting different algorithms inside one index, when a single identifier can name only one?

    Design B — single es-index type + options (current implementation)                                                                                                                                                                                                                                                                                                                                                                                                              
  index_type = 'es-index'; the algorithm and each column's behavior come from options.                                                                                                                                                                                                                                                                                                                                                                          
  Pros                                                                                                                                                                                                                                   
  - Naturally supports the multi-column, multi-modal container.                                                                                                                                                                          
  - Companion columns are configured independently per field.                                                                                                                                                                                                                                                                                                                                                                                                       
  Trade-off                                                                                                                                                                                                                              
  - In the single-column case, letting the type directly imply the algorithm would reduce user friction — but Design B always routes through options.                                                                                    
  - In the multi-column case, companion-column types must be given via options; the type alone cannot express them.
    What we'd like the community to weigh in on                                                                                                                                                                                                                                                                                                                                                                                                                              
  1. Should es-index follow Design A, Design B, or a hybrid — type-implied algorithm for the single-column case, options required for multi-column?                                                                                      
  2. Option-prefix consistency: paimon-vector uses bare prefixes (ivf-pq.*, fields.<field>.*), while es-index uses global-index.es-index.*. Should these be unified, and if so, on which convention?
  Feedback welcome — especially on the multi-column vs. per-type tension in (1), since that determines whether the multi-modal "one index, many columns" capability stays first-class.

pom.xml
```diff
@@ -519,6 +519,19 @@ under the License.
...
+        <profile>
+            <!-- paimon-eslib depends on Java-11 eslib-api/eslib-core (Lucene 9.12), so it must be
+                 built with JDK 11+. Activated only on JDK 11+ (like paimon-iceberg) so the default
+                 JDK-8 build skips it instead of failing the RequireJavaVersion enforcer rule. -->
+            <id>paimon-eslib</id>
+            <modules>
+                <module>paimon-eslib</module>
+            </modules>
+            <activation>
+                <jdk>11</jdk>
```
JingsongLi: This only auto-activates the module on JDK 11, not on later JDKs. I verified `mvn help:active-profiles -DskipTests` locally: with JAVA_HOME set to JDK 11 the `paimon-eslib` profile is active, but with JDK 17 or JDK 21 it is not, so the new module is silently skipped in those builds despite the comment saying JDK 11+. Please use a range such as `[11,)` (and keep the Spark-side profile aligned) so Java 17/21 builds also compile and test `paimon-eslib`.

paimon-core/src/main/java/org/apache/paimon/table/source/FullTextScanImpl.java
```diff
@@ -102,16 +103,34 @@ public Plan scan() {
...
-            String columnName = checkNotNull(idToColumn.get(meta.indexFieldId()));
+            IndexIdentity identity = IndexIdentity.of(indexFile);
             Range range = new Range(meta.rowRangeStart(), meta.rowRangeEnd());
-            byColumnAndRange
-                    .computeIfAbsent(columnName, k -> new HashMap<>())
-                    .computeIfAbsent(range, k -> new ArrayList<>())
-                    .add(indexFile);
+            for (int columnId : matchedTextColumnIds(meta, textColumnIds)) {
+                String columnName = checkNotNull(idToColumn.get(columnId));
+                if (!identity.equals(chosenByColumn.get(columnName))) {
```
JingsongLi: This selection is global for the whole column, so it can drop valid ranges from the non-chosen identity. For example, suppose a dedicated full-text index on `content` covers rows [0, 9], and an ES/vector index with `content` in `extraFieldIds` covers rows [10, 19]. `chooseIndexPerColumn` picks the dedicated identity, this branch skips the ES file for [10, 19], and the raw fallback below still builds `GlobalIndexCoverage` from `allIndexFiles`; because coverage counts `extraFieldIds`, [10, 19] is considered indexed and is not added as a raw range. A full-text query then misses matching rows in [10, 19]. Please choose the serving identity per column/range, or compute raw coverage from the same files that are actually emitted as searchable splits.

paimon-eslib/src/main/java/org/apache/paimon/eslib/index/ESIndexOptions.java
```diff
@@ -0,0 +1,284 @@
...
+            case TINYINT:
+                return ScalarFieldType.INT;
+            case BIGINT:
+                return ScalarFieldType.LONG;
+            case FLOAT:
+                return ScalarFieldType.FLOAT;
+            case DOUBLE:
+                return ScalarFieldType.DOUBLE;
+            default:
+                return ScalarFieldType.KEYWORD;
```
JingsongLi, line 565: This fallback makes unsupported scalar types fail later in the writer. For example, a table can include a BOOLEAN or DECIMAL companion column in an `es-index`; `ESIndexOptions` maps it to `ScalarFieldType.KEYWORD`, but `ESIndexGlobalIndexWriter.extractScalar(...)` falls through to `row.getString(pos).toString()` for those type roots, and `GenericRow.getString` casts the stored value to `BinaryString`. The index build then throws a `ClassCastException` instead of either indexing the value correctly or rejecting the schema up front. Please either add type-specific extraction/support for the non-numeric scalar roots you want to allow, or throw an `IllegalArgumentException` here like `mapArrayScalarType` does for unsupported array element types.

pom.xml
CrownChu: Thanks. <jdk>11</jdk> is a prefix match that only fires on JDK 11, so 17/21 builds silently skipped the module. Changed both the root paimon-eslib profile and the paimon-spark-3.5 paimon-eslib profile to activate on          
 <jdk>[11,)</jdk> — the Spark-side one had no <activation> at all, so it now auto-activates in sync with the root, and JDK 8 still skips both. Verified with mvn help:active-profiles that paimon-eslib is now active on JDK 11 and 17. Fixed in c2e1f85.

paimon-core/src/main/java/org/apache/paimon/table/source/FullTextScanImpl.java
CrownChu: thanks. The selection was per-column, so a range covered only by the non-chosen identity was dropped while the raw-coverage fallback (computed from all index files) still counted it as indexed, leaving those rows unsearchable. Switched to selecting the serving identity per (column, row range) in chooseIndexPerColumnAndRange, so each range's covering index is emitted as its own split. This also required building the reader per split in FullTextReadImpl, since a column can now legitimately carry splits from multiple identities (reusing the first split's config would misread the others). Added a regression test that places a dedicated full-text index over rows  [0,1] and an extra-field index over [2,3] and asserts a query matches rows in both ranges. Fixed in c2e1f85.

paimon-eslib/src/main/java/org/apache/paimon/eslib/index/ESIndexOptions.java
CrownChu: Fixed by failing fast, mirroring mapArrayScalarType: mapScalarType now maps CHAR/VARCHAR to KEYWORD explicitly and throws IllegalArgumentException for unsupported scalar roots (BOOLEAN, DECIMAL, ...), instead of defaulting them  to KEYWORD and blowing up later in ESIndexGlobalIndexWriter.extractScalar with a ClassCastException. Added a test asserting a BOOLEAN companion column is rejected up front at option-parse time. Fixed in c2e1f85.

paimon-eslib/src/main/java/org/apache/paimon/eslib/index/ESIndexGlobalIndexerFactory.java
jerry-024: I think there is still a correctness gap here.

`FullTextScanImpl` now treats an `es-index` file as full-text coverage whenever the searched text column appears in either `indexFieldId` or `extraFieldIds`, because `ESIndexGlobalIndexerFactory.supportsFullTextSearch()` returns true at the index-type level. But `es-index` field capability is per field: a STRING field defaults to KEYWORD in `ESIndexOptions` unless an analyzer/type=fulltext is configured.

A concrete case:

1. Rows [0, 99] are covered by a real full-text index on `title`.
2. Rows [100, 199] are covered only by a hybrid `es-index` whose primary field is `embedding` and whose extra field is `title`.
3. That `title` extra field has no analyzer configured, so inside the ES index it is KEYWORD, not FULLTEXT.
4. A query runs `full_text_search(title, {"match":{"query":"paimon"}})`.

During planning, the `es-index` file for rows [100, 199] is still counted as full-text coverage for `title`, so `GlobalIndexCoverage.unindexedRanges(title)` will not add a raw fallback split for that range. During reading, `ESIndexGlobalIndexReader.visitFullTextSearch()` cannot serve FULLTEXT on the KEYWORD `title` field and returns empty/no-match. The result is that rows [100, 199] are skipped entirely, even though raw full-text fallback should have searched them.

So I think the scan/coverage decision needs to be field-capability-aware, not only index-type-aware. Only ranges where the queried field is actually configured as FULLTEXT in that `es-index` should count as full-text indexed coverage. Otherwise those ranges should remain eligible for raw fallback.

Could we add a regression test for this mixed-range case: one range covered by a real FULLTEXT index, another range covered only by an `es-index` where the searched text column is a KEYWORD extra field, and verify the latter range is still searched via raw fallback?

CrownChu: Thanks — fixed at the root on the write side rather than in the scan. String columns now default to FULLTEXT (with a .keyword sub-field) instead of KEYWORD (commit c417d6d677), so a text column carried as an extra field of a  hybrid index is genuinely full-text-searchable and the planning/reading paths agree — no KEYWORD-vs-FULLTEXT mismatch. Exact predicates (=, IN, ranges, prefix) keep exact semantics by routing to the <field>.keyword sub-field; type=keyword opts a column out of full-text. Let me know if you'd still prefer capability-aware coverage as defense-in-depth for the explicit type=keyword case.

paimon-core/src/main/java/org/apache/paimon/table/source/FullTextScanImpl.java
```diff
@@ -93,7 +94,7 @@ public Plan scan() {
                     if (globalIndex == null) {
                         return false;
                     }
-                    return textColumnIds.contains(globalIndex.indexFieldId())
+                    return !matchedTextColumnIds(globalIndex, textColumnIds).isEmpty()
                             && supportsFullTextSearch(entry.indexFile().indexType());
```
JingsongLi, line 97: Blocking: this still treats every `es-index` file as full-text coverage at the index-type level. The default STRING change fixes the implicit case, but a user can still explicitly configure `global-index.es-index.fields.title.type=keyword`. In that valid layout, `FullTextScanImpl` includes the ES file because `supportsFullTextSearch()` is true and the title field is covered by `indexFieldId`/`extraFieldIds`; `GlobalIndexCoverage` then marks that row range as indexed, so no `RawFullTextSearchSplit` is emitted. At read time `ESIndexGlobalIndexReader` no-matches the KEYWORD field for full-text search, so matching rows in that range are silently skipped. Please make the coverage decision field-capability-aware, or exclude explicit keyword fields from full-text coverage, and add a regression for an `es-index` range where the searched text column is configured as `type=keyword`.

CrownChu: Thanks for pointing this out. This has been addressed in 3b2f57abc4 without adding a field-level capability API to GlobalIndexerFactory.

We no longer allow a STRING field in es-index to be keyword-only.`type=keyword` now selects KEYWORD as the primary representation, while ESIndexOptions automatically adds a `<field>.fulltext` FULLTEXT multi-field.
The writer indexes the same value into both fields, and ESIndexGlobalIndexReader routes full-text queries on a KEYWORD-primary field to `<field>.fulltext`.

Therefore, an es-index file covering a text column is now genuinely capable of serving full-text search, so the existing index-type-level coverage decision no longer suppresses a required raw fallback.

JingsongLi:
ESLib's SQL LIKE conversion does not follow the default slash escape, resulting in incorrect matching for 'admin \ _%'; The supplementary regression test failed stably (expected 1, actual 0). All 22 tests have passed * *. Additionally, there is a potential concurrent resource leakage window between 'ensueLoaded()' and 'close()'

CrownChu:
Thanks — both issues have now been fixed.

  1. SQL LIKE escaping (`5e6135ccc2`)
     - `\` is now handled as Paimon's default SQL LIKE escape.
     - Escaped `_`, `%`, and `\` are treated as literals, while unescaped `_` and `%` remain wildcards.
     - The regression test `likeHonorsSqlEscapeAndLuceneLiterals` covers `admin\_%`, `admin\%%`, and `admin\\%`. The `admin\_%` case now consistently returns exactly one match.

  2. Concurrent reader lifecycle (`5e6135ccc2`)
     - The complete lazy-load/search operation now holds the lifecycle read lock.
     - `close()` holds the write lock, so it cannot close resources while `ensureLoaded()` is opening or publishing them.
     - `closeWaitsForInFlightLoadAndClosesEveryStream` deterministically verifies that close waits for an in-flight load and releases every opened stream.
     - `failedLazyLoadClosesStreamsBeforeRetry` covers cleanup after a failed lazy load.

  The branch has also been upgraded to ESLib 1.0.4 in `36bc216e26`.

  Local verification:
  `mvn -pl paimon-eslib clean -DwildcardSuites=none test`
  Result: 32 tests passed, 0 failures.

  `mvn -pl paimon-eslib -DskipTests package`
  Result: BUILD SUCCESS.

  回复 executor 的旧 review 线程：

  Sorry for missing the reply on this thread. This was fixed in `31f1758c56`.

  `READ_SEARCH_EXECUTORS` is now a `ConcurrentMap<Integer, ExecutorService>` keyed by the resolved thread count:

  - different configured thread counts receive different pools;
  - the same resolved count reuses the same pool;
  - `read-search-threads=0` always returns `null` and cannot reuse a previously created pool.

  Therefore the executor configuration is no longer first-use-wins. The executor regression tests also cover configured/caller executor separation, serial mode, and the nested-executor deadlock guard.

JingsongLi:
+1

PR #8004, opened 2026-05-28
[spark] Improve unsupported provider error message with catalog details

Zouxxyy:
### Purpose

Improve the "Provider is not supported" error message in SparkCatalog to include the catalog name, implementation class, and supported providers list.

Before: `Provider is not supported: paimon1`
After: `Provider 'paimon1' is not supported by catalog 'my_catalog' (implementation: SparkCatalog). Supported providers: [paimon, orc, parquet, csv, text, json]`

### Tests

Update existing test in `DDLTestBase` to match the new error message format.


JingsongLi:
+1

PR #8008, opened 2026-05-28
[spark] Avoid serializing known splits in read builder

wzx140:
### Purpose

Avoid capturing `KnownSplitsTable` in the read builder. The known split list can be large and Spark already carries per-task splits through input partitions, so the read builder can delegate to the origin table.

### Tests

Not run locally in this split step.


wzx140:
Closing in favor of preview PRs in the fork.

PR #8010, opened 2026-05-28
[python][DISCUSS] supports chunk shuffle in file meta layer.

steFaiz:
#### Background
This PR originates from our inner cases: when using paimon table as dataloaders, engine training always needs deterministically shuffled data rather than sequential data.
It's highly expensive to shuffle the entire dataset for each training, so a common way is:
1. When loading data into paimon, perform a global row-level shuffle
2. During training, adapt chunk shuffle rather than row-level shuffle.
3. More sophisticated shuffle, e.g. read several chunks and do row-level shuffle among them.

We can provide 1 & 2. This PR introduces a chunk shuffle for pypaimon. The mechanism can be illustrated as below:

image

1. we logically divide data files into chunks. The most simple case:
   ```python
   Chunk 1 {
       file: file1,
      range: [0, 100]
   }
   ```
   This means the chunk is in file1, covering [0, 100] rows
2. deterministically shuffle chunks
3. wraps each chunk to a single `SliceSplit`

#### Usage
The usage is simple:
```python
readBuilder.new_scan().with_chunk_shuffle(chunk_size, seed).with_shard(idx, total_worker);
```

#### More
1. A random-access optimized file format. I think the ROW format introduced in java module is nice. I've tested the random access of several file formats （On local Disk）:
   | Chunk Size | row (ms) | parquet (ms) | lance (ms) |
   |:----------:|:--------:|:------------:|:----------:|
   | 100        | 42       | 440          | 41         |
   | 500        | 94       | 689          | 29         |
   | 1000       | 145      | 944          | 36         |

   Note that lance can only be stored on ObjectStore. DFS is not supported now.
2. During training, callers can pre-fetch next several chunks to read. So the read latency actually only influence the first batch.

### Tests
UnitTests

### DIFF
--- a/paimon-python/pypaimon/read/scanner/chunk_shuffle_split_generator.py
+++ b/paimon-python/pypaimon/read/scanner/chunk_shuffle_split_generator.py
@@ -1,0 +1,374 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+import random
+from abc import abstractmethod
+from collections import defaultdict
+from dataclasses import dataclass
+from typing import Any, List, Optional, Tuple
+
+from pypaimon.globalindex.indexed_split import IndexedSplit
+from pypaimon.manifest.schema.data_file_meta import DataFileMeta
+from pypaimon.manifest.schema.manifest_entry import ManifestEntry
+from pypaimon.read.scanner.split_generator import AbstractSplitGenerator
+from pypaimon.read.sliced_split import SlicedSplit
+from pypaimon.read.split import DataSplit, Split
+from pypaimon.table.row.generic_row import GenericRow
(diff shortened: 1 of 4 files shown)


steFaiz:
@JingsongLi  @XiaoHongbo-Hope  This PR is more of a discussion. I'd really appreciate any feedback or suggestions you might have. Thanks!

JingsongLi:
I am okay with this feature. Is there any other system with a similar design for API hierarchy?

steFaiz:
>  Is there any other system with a similar design for API hierarchy?

Yes. The closest precedent I found is **Petastorm**, an ML data reader for Parquet datasets. Its reader API exposes shuffle and distributed sharding at the same reader-construction level: seed, shuffle_row_groups, shuffle_rows, cur_shard, and shard_count. The main difference is that Petastorm shuffles existing Parquet row groups, while this PR derives logical row-count chunks from Paimon manifest entries/files and then converts them back to Splits.

Similar API hierarchies also exist in ML input systems such as Hugging Face IterableDataset, NVIDIA DALI readers, WebDataset, Mosaic Streaming, Ray Data, TensorFlow tf.data, and PyTorch/TorchData. They commonly expose deterministic shuffle options and distributed shard/rank options in the dataset/reader/input-pipeline layer rather than rewriting the physical dataset.

I think the main advantage of paimon is:
1. In our current implementation, each chunk is a single file, including blobs and structured cols(metadatas). Considering a 500,000,000 image dataset, chunk size is 100, there would be 5 million files. In paimon, we set the target blob size as 1G, the total file num is less than 30000
2. The table arch makes it much easier to manage.
3. We could easily deal with multiple datasets by importing each dataset to a single partition.

JingsongLi:
I found Hugging Face IterableDataset.shuffle：
```
  def shuffle(
      self,
      seed: Optional[int] = None,
      generator: Optional[np.random.Generator] = None,
      buffer_size: int = 1000
  ) -> "IterableDataset"
```
You can continue, I think this is a good direction.

leaves12138:
Thanks for the proposal. Since this PR is marked as DISCUSS and the current Python lint job is failing, I am leaving it unapproved for now. Please resolve the CI failure and clarify when the design is ready for final review.

PR #8011, opened 2026-05-28
[core] Fix format table date partition pruning

leaves12138:
### Purpose

Fix `FormatTableScan` equality partition path optimization for partition columns whose internal literal representation differs from the partition directory string.

For example, a DATE partition predicate `dt = DATE '2026-05-01'` is represented internally as epoch days (`20574`). The previous code used `literal.toString()` when building the optimized partition path, producing `dt=20574/` instead of the actual `dt=2026-05-01/`, which caused zero splits to be returned.

### Changes

- Convert equality partition literals to their partition string representation via Paimon's cast framework before generating the partition path.
- Pass `partitionType` into `extractLeadingEqualityPartitionSpecWhenOnlyAnd` so the literal can be formatted according to its field type.
- Add a regression test covering DATE partition equality pruning for both normal partition paths and value-only partition paths.

### Tests

- `mvn -pl paimon-core -Pfast-build -DfailIfNoTests=false -Dtest=FormatTableScanTest#testComputeScanPathWithDateEqualityFilter test`
- `mvn -pl paimon-core -Pfast-build -DfailIfNoTests=false -Dtest=FormatTableScanTest test`
- `mvn -pl paimon-core spotless:apply`

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/format/FormatTableScan.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/format/FormatTableScan.java
@@ -19,7 +19,10 @@
 package org.apache.paimon.table.format;
 
 import org.apache.paimon.CoreOptions;
+import org.apache.paimon.casting.CastExecutor;
+import org.apache.paimon.casting.CastExecutors;
 import org.apache.paimon.data.BinaryRow;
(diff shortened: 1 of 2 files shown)


JingsongLi (approved):
+1

miomiocat (approved):

PR #8013, opened 2026-05-28
[core] Return null for blob view referencing null value

leaves12138:
### Purpose

When a blob view references an upstream row whose blob value is SQL null, the reader should faithfully return null instead of treating the reference as missing.

### Changes

- Track preloaded blob view references as descriptor values or null values.
- Make `BlobViewResolvingRow` report `isNullAt` and `getBlob` consistently for blob views resolved to null.
- Keep missing references fail-fast by distinguishing them from references that resolve to null.
- Extend `BlobTableTest#testBlobViewE2E` to cover upstream null blob values.

### Tests

- `mvn -pl paimon-common,paimon-core -DskipTests compile`
- `mvn -pl paimon-common,paimon-core spotless:check -DskipTests`
- `mvn -pl paimon-core -am -Pfast-build -DfailIfNoTests=false -Dtest=BlobTableTest#testBlobViewE2E+testBlobViewResolutionFailsOnMissingUpstream test`

This pull request was created by codex robot.

### DIFF
--- a/paimon-common/src/main/java/org/apache/paimon/data/BlobViewResolver.java
+++ b/paimon-common/src/main/java/org/apache/paimon/data/BlobViewResolver.java
@@ -24,4 +24,8 @@ import java.io.Serializable;
 public interface BlobViewResolver extends Serializable {
 
     void resolve(BlobView blobView);
+
+    default boolean resolvesToNull(BlobView blobView) {
+        return false;
+    }
 }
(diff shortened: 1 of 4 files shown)


leaves12138:
LGTM from my side.

I rechecked the null-upstream blob view flow. The change now distinguishes upstream null values from missing references, and `isNullAt`/`getBlob` expose the resolved null consistently. I also ran the focused `BlobTableTest` cases for null upstream, forwarded `BlobView` references, and missing upstream failure; all passed.

JingsongLi:
+1

PR #8014, opened 2026-05-28
[python] Add ROW file format read and write support

JingsongLi:
Implement the ROW binary file format in the Python SDK, matching the Java implementation for cross-language interoperability. The format uses block-level ZSTD compression with delta-varint block indexing and a 32-byte footer.

### DIFF
--- a/paimon-python/pypaimon/read/split_read.py
+++ b/paimon-python/pypaimon/read/split_read.py
@@ -45,6 +45,7 @@ from pypaimon.read.reader.row_range_filter_record_reader import RowIdFilterRecor
 from pypaimon.read.reader.format_blob_reader import FormatBlobReader
 from pypaimon.read.reader.format_lance_reader import FormatLanceReader
 from pypaimon.read.reader.format_pyarrow_reader import FormatPyArrowReader
+from pypaimon.read.reader.format_row_reader import FormatRowReader
 from pypaimon.read.reader.format_vortex_reader import FormatVortexReader
 from pypaimon.read.reader.iface.record_batch_reader import (RecordBatchReader,
                                                             RowPositionReader, EmptyRecordBatchReader)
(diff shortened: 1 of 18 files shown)


paimon-python/pypaimon/read/split_read.py
```diff
@@ -268,10 +272,19 @@ def file_reader_supplier(self, file: DataFileMeta, for_merge_read: bool,
                 ordered_read_fields, read_arrow_predicate, batch_size=batch_size,
                 options=self.table.options,
                 nested_name_paths=ordered_nested_paths)
+        elif file_format == CoreOptions.FILE_FORMAT_ROW:
+            if has_nested:
+                raise NotImplementedError(
+                    "Nested-field projection is not supported on ROW files")
+            format_reader = FormatRowReader(
+                self.table.file_io, file_path, read_file_fields,
+                list(name_to_field.values()),
```
leaves12138: `FormatRowReader` decodes each row by physical field order, so it needs the complete physical ROW schema, not only the requested fields. Here `full_fields` is built from `name_to_field`, which only contains the read/projection fields plus trimmed lookup fields. For example, with a ROW append table `(id INT, name STRING, val DOUBLE)`, `with_projection(["id", "val"])` passes only `[id, val]`; `_decode_block` then decodes the bytes of `name` as `val` and returns corrupted doubles. Please pass the full file schema in physical order to `FormatRowReader` (while keeping `read_file_fields` as the projected columns).

leaves12138 (changes requested):
Thanks for adding ROW format support. I found one correctness issue around column projection, so requesting changes.

leaves12138 (approved):
The projection issue from my previous review is fixed now. I rechecked the ROW reader schema path and ran the focused ROW tests (, ) plus the projection reproducer; all passed. Looks good to me.

Issue #8015, opened 2026-05-28
[Bug] [spark] Align spark.paimon.branch option with explicit branch syntax for chain tables


juntaozhang:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Paimon version

master

### Compute Engine

spark

### Minimal reproduce step

```sql
CREATE DATABASE IF NOT EXISTS my_db1;
USE spark_catalog.my_db1;
DROP TABLE IF EXISTS `my_db1`.`chain_test`;
CREATE TABLE IF NOT EXISTS `chain_test` (
  `t1` BIGINT,
  `t2` BIGINT,
  `t3` STRING
) PARTITIONED BY (`dt` STRING)
TBLPROPERTIES (
  'bucket-key' = 't1',
  'primary-key' = 'dt,t1',
  'partition.timestamp-pattern' = '$dt',
  'partition.timestamp-formatter' = 'yyyyMMdd',
  'chain-table.enabled' = 'true',
  'bucket' = '1',
  'merge-engine' = 'deduplicate',
  'sequence.field' = 't2'
);

SET spark.paimon.branch = delta;
INSERT OVERWRITE TABLE `my_db1`.`chain_test` PARTITION (dt = '20250810') VALUES (1, 2, '1');
INSERT OVERWRITE TABLE `my_db1`.`chain_test$branch_delta` PARTITION (dt = '20250811') VALUES (2, 2, '1');
SELECT * FROM `my_db1`.`chain_test$snapshots`;

RESET spark.paimon.branch;
// ①
SELECT * FROM `my_db1`.`chain_test$branch_snapshot` WHERE dt = '20250811';

// ②
SET spark.paimon.branch = snapshot;
SELECT * FROM `my_db1`.`chain_test` where dt = '20250811';

DROP TABLE IF EXISTS `my_db1`.`chain_test`;
```

### What doesn't meet your expectations?

①  ② should be empty.
but currently ② is not empty.

### Anything else?

#5460 deprecated the branch option.
#7084 discussed the priority between Paimon dynamic options and table identifiers.

Before the branch option is fully removed, we should unify their behaviors first.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #8016, opened 2026-05-28
[spark] Align spark.paimon.branch option with explicit branch syntax for chain tables

juntaozhang:
### Purpose
#8015

### Tests
SparkChainTableITCase.testChainTableWithBranchOption


leaves12138:
Thanks for the fix. I am holding off on approval for now because the current CI status still has a failing `build_test` job. Please fix or rerun it, then I can take another pass.

juntaozhang:
Hi @leaves12138, appreciate your feedback,  PTAL.
Thanks

JingsongLi:
Reviewed with focus on alignment between the deprecated `spark.paimon.branch` option and the explicit Java/Paimon `table$branch_xxx` identifier path.

Applying the option before `catalog.getTable(...)` makes the Spark catalog resolve the same branched `Identifier` that explicit branch syntax would use, while leaving an explicit branch name untouched. I also verified the common Spark module compiles with:

```shell
mvn -pl paimon-spark/paimon-spark-common -am -Pfast-build -DskipTests compile
```

No additional blocker from my side.

Issue #8018, opened 2026-05-28
[Feature] Support vector type on spark


Stefanietry:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Motivation

Support vector type on spark

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #8019, opened 2026-05-28
[core] support vector on spark

Stefanietry:
### Purpose
Linked issue:  https://github.com/apache/paimon/issues/8018

### Tests
org.apache.paimon.spark.SparkMultimodalITCase#testVector

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/SparkCatalog.java
+++ b/paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/SparkCatalog.java
@@ -44,7 +44,9 @@ import org.apache.paimon.table.object.ObjectTable;
 import org.apache.paimon.types.BlobType;
 import org.apache.paimon.types.DataField;
 import org.apache.paimon.types.DataType;
+import org.apache.paimon.types.DataTypes;
 import org.apache.paimon.utils.ExceptionUtils;
+import org.apache.paimon.utils.Preconditions;
 
 import org.apache.spark.sql.PaimonSparkSession$;
 import org.apache.spark.sql.SparkSession;
@@ -73,6 +75,7 @@ import org.apache.spark.sql.connector.expressions.Transform;
 import org.apache.spark.sql.execution.datasources.DataSource;
 import org.apache.spark.sql.execution.datasources.FileFormat;
 import org.apache.spark.sql.execution.datasources.v2.FileDataSourceV2;
+import org.apache.spark.sql.types.ArrayType;
 import org.apache.spark.sql.types.StructField;
 import org.apache.spark.sql.types.StructType;
 import org.apache.spark.sql.util.CaseInsensitiveStringMap;
@@ -563,6 +566,7 @@ public class SparkCatalog extends SparkBaseCatalog
         List<String> blobFields = CoreOptions.blobField(properties);
         Set<String> blobDescriptorFields = new CoreOptions(properties).blobDescriptorField();
         List<String> blobViewFields = CoreOptions.blobViewField(properties);
+        Set<String> vectorFields = CoreOptions.fromMap(properties).vectorField();
         String provider = properties.get(TableCatalog.PROP_PROVIDER);
         if (!usePaimon(provider)) {
             if (isFormatTable(provider)) {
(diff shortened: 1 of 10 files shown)


leaves12138:
Thanks for the contribution. I am holding off on approval because the current CI status has multiple failing jobs. Please fix or rerun the failures, then I can take another pass.

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/SparkTypeUtils.java
```diff
@@ -242,6 +251,11 @@ public DataType visit(ArrayType arrayType) {
             return DataTypes.createArrayType(elementType.accept(this), elementType.isNullable());
         }
 
+        @Override
+        public DataType visit(VectorType vectorType) {
+            return DataTypes.createArrayType(vectorType.getElementType().accept(this), vectorType.isNullable());
```
JingsongLi: Bug: Please add tests, visit(VectorType) passes vectorType.isNullable() (column nullability) as containsNull, but containsNull controls whether array elements can be null. The existing visit(ArrayType) correctly passes elementType.isNullable().

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/SparkCatalog.java
```diff
@@ -609,6 +607,17 @@ private Schema toInitialSchema(
...
                         "The type of blob field must be binary");
                 type = new BlobType();
+            } else if (vectorFields.contains(field.name())) {
+                Preconditions.checkArgument(
+                        field.dataType() instanceof ArrayType,
+                        "The type of blob field must be array");
+                ArrayType arrayType = (ArrayType) field.dataType();
+                String dimKey = String.format("field.%s.vector-dim", field.name());
+                type =
+                        new VectorType(
```
JingsongLi: VectorType is constructed with arrayType.containsNull() (element nullability) instead of field.nullable() (column nullability). The general-case code path on line 613 correctly uses field.nullable() — this vector branch is inconsistent.

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/SparkCatalog.java
```diff
@@ -609,6 +607,17 @@ private Schema toInitialSchema(
                         field.dataType() instanceof org.apache.spark.sql.types.BinaryType,
                         "The type of blob field must be binary");
                 type = new BlobType();
+            } else if (vectorFields.contains(field.name())) {
+                Preconditions.checkArgument(
```
JingsongLi, line 611: Integer.parseInt(properties.get(dimKey)) throws NumberFormatException: null when field.<name>.vector-dim is missing. Add Preconditions.checkArgument with a descriptive message.

Stefanietry: Add check:

String dimKey = String.format("field.%s.vector-dim", field.name());
Preconditions.checkArgument(
                        properties.containsKey(dimKey),
                        "When setting '"
                                + CoreOptions.VECTOR_FIELD.key()
                                + "', you must also set 'field.%s.vector-dim',"
                                + " where %s is the name of the vector field.");

paimon-spark/paimon-spark-common/src/main/java/org/apache/paimon/spark/SparkCatalog.java
Stefanietry: Use general api:

DataTypes.VECTOR(
                                Integer.parseInt(properties.get(dimKey)),
                                toPaimonType(arrayType.elementType()));

paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/SparkTypeUtils.java
Stefanietry: Explicitly indicate that the array does not contain null：
 
  public DataType visit(VectorType vectorType) {
            return DataTypes.createArrayType(vectorType.getElementType().accept(this), false);
   }
   
 
 Test: SparkTypeTest.testVectorType

JingsongLi (approved):
+1

PR #8020, opened 2026-05-28
[core] Remove debug print from compact buckets table

leaves12138:
### Purpose

Remove an accidental debug print from `CompactBucketsTable`.

### Tests

- `mvn -pl paimon-core -DskipTests compile`


leaves12138:
I re-checked this PR. It only removes the accidental debug `System.out.println` from `CompactBucketsTable`, and CI is green. I cannot approve it because I am the author of this PR.

PR #8021, opened 2026-05-28
[python] Support BlobView feature

discivigour:
### Summary
- Add Python BlobViewStruct / BlobView wire-format support and the blob-view-field option.
- Store descriptor/view BLOB fields inline, validate bad inline field configuration and payloads, and avoid writing new .blob files for view fields.
- Resolve blob-view fields during reads through catalog-aware lookup, returning bytes by default or upstream BlobDescriptor bytes when blob-as-descriptor=true.

### Tests
- DataBlobWriterTest.test_blob_view_fields_resolve_upstream_blob()
- BlobTest.test_blob_view_struct_roundtrip()

### DIFF
--- a/paimon-python/pypaimon/read/reader/blob_descriptor_convert_reader.py
+++ b/paimon-python/pypaimon/read/reader/blob_descriptor_convert_reader.py
@@ -54,7 +54,6 @@ class BlobInlineConvertReader(RecordBatchReader):
         self._inner = inner
         self._table = table
         self._prescan_reader_factory = prescan_reader_factory
-        self._descriptor_fields = CoreOptions.blob_descriptor_fields(table.options)
         self.file_io = inner.file_io
         self.blob_field_indices = inner.blob_field_indices
         self._view_fields = CoreOptions.blob_view_fields(table.options)
@@ -101,12 +100,14 @@ class BlobInlineConvertReader(RecordBatchReader):
                         continue
                     for value in batch.column(field_name).to_pylist():
                         value = self._normalize_blob_to_bytes(value)
+                        if value is None:
+                            continue
                         if isinstance(value, bytes) and BlobViewStruct.is_blob_view_struct(value):
                             all_view_structs.append(BlobViewStruct.deserialize(value))
                         else:
                             raise ValueError(
                                 f"Expected BlobViewStruct bytes in view field '{field_name}', "
-                                f"but got non-BlobViewStruct bytes (length={len(value)})"
+                                f"but got non-BlobViewStruct bytes."
                             )
         finally:
             prescan_reader.close()
--- a/paimon-python/pypaimon/read/reader/blob_descriptor_convert_reader.py
+++ b/paimon-python/pypaimon/read/reader/blob_descriptor_convert_reader.py
@@ -66,9 +66,16 @@ class BlobInlineConvertReader(RecordBatchReader):
         self._blob_view_lookup = None
 
     def read_arrow_batch(self) -> Optional[RecordBatch]:
-        # Ensure prescan is done before reading (only needed for view fields)
+        # Align with Java: only enter blob view resolution when catalog_loader is available
+        # If catalog_loader is None, skip both Stage 1 (view resolution) and Stage 2 (descriptor resolution)
+        # This matches Java's behavior in DataEvolutionTableRead.createReader where blob view reader
+        # is only created when catalogContext != null
         if self._view_fields and not self._prescan_done:
-            self._prescan_view_structs()
+            if self._table.catalog_environment.catalog_loader is None:
+                # No catalog_loader available, skip view resolution
+                self._prescan_done = True
+            else:
+                self._prescan_view_structs()
 
         batch = self._inner.read_arrow_batch()
         if batch is None:
--- a/paimon-python/pypaimon/read/reader/blob_descriptor_convert_reader.py
+++ b/paimon-python/pypaimon/read/reader/blob_descriptor_convert_reader.py
@@ -15,110 +15,109 @@
 # specific language governing permissions and limitations
 # under the License.
 
-from typing import Optional
+from typing import Callable, Optional, Set
 
+import pyarrow
 from pyarrow import RecordBatch
 
 from pypaimon.common.options.core_options import CoreOptions
 from pypaimon.read.reader.iface.record_batch_reader import RecordBatchReader
-from pypaimon.table.row.blob import Blob, BlobDescriptor, BlobViewStruct
+from pypaimon.table.row.blob import Blob, BlobViewStruct
 
 
 class BlobDescriptorConvertReader(RecordBatchReader):
     """Resolves BlobView and BlobDescriptor fields in record batches.
 
     Processing is split into two clear stages:
-      Stage 1 (BlobView resolution): If view fields exist, prescan all batches,
-               collect BlobViewStructs, bulk-preload their descriptors from
-               upstream tables, and replace view field values with the
-               corresponding BlobDescriptor serialized bytes.
+      Stage 1 (BlobView resolution): If view fields exist, use a lightweight
+               prescan reader (only projecting view columns) to collect
+               BlobViewStructs, bulk-preload their descriptors, then read
+               full data from the main reader and replace view field values
+               with the corresponding BlobDescriptor serialized bytes.
       Stage 2 (BlobData resolution): Controlled by blob-as-descriptor option.
                If false, resolve all BlobDescriptor bytes (from both descriptor
                fields and view fields) into real blob data bytes.
                If true, return as-is.
     """
 
-    def __init__(self, inner: RecordBatchReader, table):
+    def __init__(self, inner: RecordBatchReader, table,
+                 prescan_reader_factory: Optional[Callable[[Set[str]], RecordBatchReader]] = None):
+        """
+        Args:
+            inner: The main data reader (reads all columns).
+            table: The table instance.
+            prescan_reader_factory: Optional factory that creates a lightweight
+                reader projecting only the specified field names. Used for
+                prescan to collect BlobViewStructs without reading all columns.
+                Signature: (field_names: Set[str]) -> RecordBatchReader
+        """
         self._inner = inner
         self._table = table
+        self._prescan_reader_factory = prescan_reader_factory
         self._descriptor_fields = CoreOptions.blob_descriptor_fields(table.options)
         self.file_io = inner.file_io
         self.blob_field_indices = inner.blob_field_indices
         self._view_fields = CoreOptions.blob_view_fields(table.options)
         self._descriptor_fields = CoreOptions.blob_descriptor_fields(table.options)
         self._blob_as_descriptor = CoreOptions.blob_as_descriptor(table.options)
-        self._cached_batches = None
-        self._batch_index = 0
+        self._prescan_done = False
+        self._blob_view_lookup = None
 
     def read_arrow_batch(self) -> Optional[RecordBatch]:
-        import pyarrow
-        # Stage 1: obtain batch (prescan for view fields, or direct read)
-        if self._view_fields:
-            batch = self._read_with_prescan(pyarrow)
-        else:
-            batch = self._inner.read_arrow_batch()
+        # Ensure prescan is done before reading (only needed for view fields)
+        if self._view_fields and not self._prescan_done:
+            self._prescan_view_structs()
+
+        batch = self._inner.read_arrow_batch()
         if batch is None:
             return None
-        # Stage 2: resolve BlobDescriptor -> real bytes (if blob-as-descriptor=false)
-        return self._resolve_blob_data(batch, pyarrow)
+        # Resolve view fields using the preloaded lookup
+        if self._view_fields and self._blob_view_lookup is not None:
+            batch = self._resolve_view_fields(batch, self._blob_view_lookup)
+        # Resolve BlobDescriptor -> real bytes (if blob-as-descriptor=false)
+        return self._resolve_blob_data(batch)
 
     # ------------------------------------------------------------------
-    # Stage 1: BlobView prescan and resolution
+    # Stage 1: BlobView prescan (lightweight, only reads view columns)
     # ------------------------------------------------------------------
 
-    def _read_with_prescan(self, pyarrow):
-        """Return the next batch from cache (view fields already resolved to
-        BlobDescriptor bytes)."""
-        if self._cached_batches is None:
-            self._prescan_and_resolve_views(pyarrow)
-        if self._batch_index >= len(self._cached_batches):
-            return None
-        batch = self._cached_batches[self._batch_index]
-        self._batch_index += 1
-        return batch
-
-    def _prescan_and_resolve_views(self, pyarrow):
-        """Prescan all batches, collect BlobViewStructs, bulk-preload
-        descriptors, then replace view field values with BlobDescriptor bytes."""
+    def _prescan_view_structs(self):
+        """Use a lightweight prescan reader (projecting only view columns) to
+        collect all BlobViewStructs and bulk-preload their descriptors."""
         from pypaimon.table.row.blob import BlobViewStruct
         from pypaimon.utils.blob_view_lookup import BlobViewLookup
 
-        # Step 1: cache all batches and collect BlobViewStructs
-        raw_batches = []
+        self._prescan_done = True
         all_view_structs = []
-        while True:
-            batch = self._inner.read_arrow_batch()
-            if batch is None:
-                break
-            raw_batches.append(batch)
-            for field_name in self._view_fields:
-                if field_name not in batch.schema.names:
-                    continue
-                for value in batch.column(field_name).to_pylist():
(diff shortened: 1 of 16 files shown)


leaves12138:
Thanks for the contribution. This PR is still in draft state, so I am leaving it unapproved for now. Please request review again once it is ready for a full review.

discivigour:
> Thanks for the contribution. This PR is still in draft state, so I am leaving it unapproved for now. Please request review again once it is ready for a full review.

Thanks, it is ready for a full review now. Could you please take a look at this PR when you have a moment?

paimon-python/pypaimon/utils/blob_view_lookup.py
```diff
@@ -0,0 +1,253 @@
...
+        row_id_values: List = result.column(SpecialFields.ROW_ID.name).to_pylist()
+        resolved: Dict[BlobViewStruct, BlobDescriptor] = {}
+        for field in read_fields:
+            if field.name == SpecialFields.ROW_ID.name:
+                continue
+            if field.name not in result.schema.names:
+                continue
+            values = result.column(field.name).to_pylist()
+            for row_id, value in zip(row_id_values, values):
+                if value is None:
```
JingsongLi, line 202: Java keeps null upstream blob values as a separate resolved state (`PreloadedBlobViews.putNull(...)`), and `BlobViewResolvingRow` then returns null for that view. Here nulls are skipped, so a valid view pointing to a null upstream blob will later fail in `resolve_descriptor()` as if the row id was missing. Please track null resolutions separately and make the read path return null for those views.

paimon-python/pypaimon/table/row/blob.py
```diff
@@ -382,3 +492,42 @@ def __eq__(self, other) -> bool:
 
     def __hash__(self) -> int:
         return hash(self._descriptor)
+
+
+class BlobView(Blob):
```
JingsongLi, line 506: This adds the `BlobView` abstraction, but generic deserialization is still not wired up: Java `Blob.fromBytes` detects `BlobViewStruct` before `BlobDescriptor` and returns an unresolved `BlobView`. Python `Blob.from_bytes()` currently only checks descriptors, so serialized `BlobViewStruct` bytes fall through to `BlobData`; `row.get_blob()` / forwarding paths will treat view references as raw bytes instead of BlobView values. Please add `BlobViewStruct` detection there and cover the non-`to_arrow()` path in tests.

paimon-python/pypaimon/read/split_read.py
```diff
@@ -780,6 +780,19 @@ def _push_down_predicate(self) -> Optional[Predicate]:
         return None
 
     def create_reader(self) -> RecordReader:
+        reader = self._create_raw_reader()
+
+        if (CoreOptions.blob_view_fields(self.table.options)
```
JingsongLi: Java has `blob-view.resolve.enabled` (default true) so callers can set it to false and preserve `BlobViewStruct` references when forwarding blob views to another table. This Python path resolves whenever `blob-view-field` is configured, so it loses that mode. Please add the option and gate the `BlobInlineConvertReader` wrapping on it, matching Java semantics.

JingsongLi (commented):
I focused on the Python BlobView implementation against the existing Java semantics. A few behavior gaps below.

paimon-python/pypaimon/utils/blob_view_lookup.py
discivigour: Thanks for the review, I have completed this null feature.

paimon-python/pypaimon/table/row/blob.py
discivigour: Added.

paimon-python/pypaimon/read/split_read.py
discivigour: this param supported.

paimon-python/pypaimon/read/split_read.py
```diff
@@ -792,6 +785,20 @@ def _push_down_predicate(self) -> Optional[Predicate]:
         return None
 
     def create_reader(self) -> RecordReader:
+        reader = self._create_raw_reader()
+
+        if ((CoreOptions.blob_view_fields(self.table.options) and CoreOptions.blob_view_resolve_enabled(
```
JingsongLi, line 806: [P2] Java only enters BlobView resolution in `DataEvolutionTableRead` when `catalogContext != null`, and it gives a clear error if `readFactory` is missing. This path wraps the reader based only on `blob-view-field`, so tables created with `CatalogEnvironment.empty()` such as `FileStoreTable.from_path()` still reach `BlobViewLookup._load_table()` and fail with an `AttributeError` on `catalog_loader.load()`. Please check `catalog_environment.catalog_loader` before enabling resolution: either skip resolving as Java does when no resolver context exists, or raise a clear error, and add coverage for the no-catalog-loader case.

paimon-python/pypaimon/read/reader/blob_descriptor_convert_reader.py
```diff
@@ -15,68 +15,185 @@
...
-        from pypaimon.table.row.blob import Blob, BlobDescriptor
+        all_view_structs = []
 
-        result = batch
-        for field_name in self._descriptor_fields:
-            if field_name not in result.schema.names:
+        prescan_reader = self._prescan_reader_factory(self._view_fields)
+        try:
+            while True:
+                batch = prescan_reader.read_arrow_batch()
```
JingsongLi, line 101: [P2] This prescan exhausts the prescan reader before the first main batch is returned. In Python, `limit` is enforced by outer `TableRead` after a batch is returned or sliced, and `DataEvolutionSplitRead` itself never receives that limit. Java wires `topN/limit` into the prescan reader in `DataEvolutionTableRead.configureBlobViewPrescanRead`. As a result, Python `with_limit(1)` can still prescan and validate every later `BlobViewStruct` in the same split, and can even fail on a bad reference that the user-visible result would never return. Please pass limit/topN into the prescan scope, or at least add a blob-view + limit test to lock down the intended behavior.

paimon-python/pypaimon/schema/schema.py
```diff
@@ -133,3 +102,61 @@ def from_pyarrow_schema(pa_schema: pa.Schema, partition_keys: Optional[List[str]
...
+
+    @staticmethod
+    def _validate_blob_fields(fields, options, primary_keys):
+        """Validate blob field configurations in the schema."""
+        blob_names = [
+            field.name for field in fields
+            if 'blob' in str(field.type).lower()
+        ]
+
+        if not blob_names:
```
JingsongLi: [P2] This early return skips validation for explicitly configured `blob-field` / `blob-descriptor-field` / `blob-view-field` when the schema currently has no BLOB fields. Java `SchemaValidation` validates these options unconditionally against BLOB fields. Today a table can be created with, for example, `blob-view-field=picture` while `picture` remains BYTES/STRING, and the later write/read paths either ignore or misuse that configuration. Please validate the configured fields first, then apply the BLOB-specific table constraints only when BLOB fields exist; `blob-field` should be included in that validation as well.

JingsongLi (commented):
Additional pass focused on parity with the Java BlobView read path and schema validation.

paimon-python/pypaimon/schema/schema.py
discivigour: changed.

paimon-python/pypaimon/read/split_read.py
discivigour: added.

paimon-python/pypaimon/read/reader/blob_descriptor_convert_reader.py
discivigour: In pypaimon, currently, the limit pushdown in the Read stage of the Append table is not supported. I will support this function in the next pr

paimon-python/pypaimon/schema/schema.py
```diff
@@ -133,3 +102,56 @@ def from_pyarrow_schema(pa_schema: pa.Schema, partition_keys: Optional[List[str]
...
+        if len(fields) <= len(blob_field_names):
+            raise ValueError(
+                "Table with BLOB type column must have other normal columns."
+            )
+
+        core_options = CoreOptions.from_dict(options)
+
+        descriptor_fields = core_options.blob_descriptor_fields()
+        view_fields = core_options.blob_view_fields()
+        unknown_inline_fields = descriptor_fields.union(view_fields).difference(blob_field_names)
```
XiaoHongbo-Hope: `blob-field` is not validated against BLOB fields.

discivigour: Thanks, added.

paimon-python/pypaimon/read/split_read.py
XiaoHongbo-Hope: > added.

 This still seems not fully fixed. When `catalog_loader is None`, prescan is skipped, but view fields
  still go through descriptor resolution. `Blob.from_bytes()` returns an unresolved `BlobView`, and
  `to_data()` raises `RuntimeError: BlobView is not resolved.`

  Should we skip `BlobInlineConvertReader` when no `catalog_loader` is available?

discivigour: changed. blob descriptor still needs BlobInlineConvertReader, so we can not skip it.

paimon-python/pypaimon/read/split_read.py
```diff
@@ -823,12 +830,27 @@ def create_reader(self) -> RecordReader:
...
+    def _create_prescan_reader(self, field_names):
+        """Create a prescan reader by constructing a new DataEvolutionSplitRead
+        instance that only projects the specified field names."""
+        from pypaimon.read.reader.iface.record_batch_reader import EmptyRecordBatchReader
+
+        prescan_fields = [f for f in self.read_fields if f.name in field_names]
+        if not prescan_fields:
+            return EmptyRecordBatchReader()
+
+        prescan_read = DataEvolutionSplitRead(
```
JingsongLi, line 871: This prescan reader only projects the blob-view columns, so `SplitRead.__init__` drops any predicate whose fields are outside that reduced projection, and it also does not receive the original read limit. As a result, `BlobInlineConvertReader._prescan_view_structs()` can preload and resolve `BlobViewStruct`s from rows that the final read would filter out or skip. I reproduced this with two target rows where `id = 1` is valid and `id = 2` points at a missing upstream table: reading with `with_filter(id == 1)` or `with_limit(1)` still fails during prescan on row 2. The prescan should preserve the effective filter/limit semantics, or otherwise only scan rows that can be returned.

paimon-python/pypaimon/schema/schema.py
```diff
@@ -62,39 +62,8 @@ def from_pyarrow_schema(pa_schema: pa.Schema, partition_keys: Optional[List[str]
...
-            if missing_options:
-                raise ValueError(
-                    f"Schema contains Blob type but is missing required options: {', '.join(missing_options)}. "
-                    f"Please add these options to the schema."
-                )
-
-            if primary_keys is not None:
-                raise ValueError("Blob type is not supported with primary key.")
+        # Validate Blob type fields in the schema
+        Schema._validate_blob_fields(fields, options, primary_keys)
```
JingsongLi: This validation only runs when callers create schemas through `Schema.from_pyarrow_schema(...)`. Callers can still construct `Schema(fields=..., options=...)` directly and `SchemaManager.create_table` will commit invalid `blob-view-field` / `blob-descriptor-field` settings, for example a `blob-view-field` pointing at a STRING column. Since direct `Schema(...)` construction is a supported/tested path, this validation should also run in `SchemaManager.create_table` after directives are applied, or otherwise be centralized before the schema is committed.

discivigour: In python, currently there is no code similar to java Schema Validation to centrally validate option parameters. In the existing code, blob and vector-related parameters are all validated in schema.from pyarrow schema. If there is a need in the future, I will separately propose a pr to unify the validation of schema parameters.

paimon-python/pypaimon/read/split_read.py
discivigour: I think this case is very rare. If the upstream table does not exist, an error should have been reported. Moreover, I have already introduced the limit pushdown function in another pr.

JingsongLi: Thanks for checking. The missing upstream table was only a minimal way to make the semantic issue visible. The correctness problem is that the prescan observes rows that the final query would filter out or skip. A filtered-out row may contain a stale BlobViewStruct, a permission/network failure, or simply many expensive references, and that should not make `WHERE id = 1` fail or trigger extra upstream reads.

Also, a limit-pushdown change in another PR would not address the predicate case here, because this prescan rebuilds the read type with only blob-view fields, so predicates on non-view columns are trimmed before the prescan reader is built. I think this PR needs to preserve the effective filter/limit in the prescan path, matching the Java path, and add regression tests for both cases.

paimon-python/pypaimon/schema/schema.py
JingsongLi: I agree that Python does not yet have a Java-style centralized validation layer, but this PR is adding new invariants and new read/write behavior that depends on them. Leaving the direct `Schema(...)` path unchecked means users can commit invalid `blob-view-field` / `blob-descriptor-field` options today, and the failure then moves to later writes/reads with much less clear errors.

This does not need a full validation refactor in this PR. A narrow fix would be to call the same blob-field validation from the schema commit path (for example in `SchemaManager.create_table` after the schema is materialized and before `schema-0` is written), so both `Schema.from_pyarrow_schema(...)` and direct `Schema(...)` creation enforce the same invariants.

paimon-python/pypaimon/read/split_read.py
discivigour: When another PR is merged, I will support this feature in BlobView Read

paimon-python/pypaimon/schema/schema.py
discivigour: added

paimon-python/pypaimon/schema/schema_manager.py
```diff
@@ -364,6 +428,7 @@ def create_table(self, schema: Schema) -> TableSchema:
                 comment=schema.comment,
             )
 
+            _validate_blob_fields(schema.fields, schema.options, schema.primary_keys)
```
JingsongLi, line 432: [P2] This closes the direct `create_table(Schema(...))` path, but schema changes still bypass the same Blob validation. `commit_changes()` builds a new `TableSchema` and calls `commit()` without running `_validate_blob_fields()` or `_validate_blob_external_storage_fields()`, whereas Java `SchemaManager.commit()` runs `SchemaValidation.validateTableSchema(newSchema)` for every schema commit. This means users can still `alter_table` with `SetOption('blob-view-field', 'non_blob_col')`, remove `row-tracking.enabled` / `data-evolution.enabled` from a table with BLOB columns, or add a BLOB column via directive without the required options, and the invalid schema is committed. Please centralize this validation before writing schema changes as well, ideally in the Python `commit()` path to match Java.

JingsongLi (commented):
One more pass on the latest revision, focused on schema-change parity with Java.

paimon-python/pypaimon/schema/schema_manager.py
```diff
@@ -166,6 +166,70 @@ def _assert_not_renaming_blob_column(
             )
 
 
+def _validate_blob_fields(fields: List[DataField], options: dict, primary_keys: List[str]):
+    """Validate blob field configurations in the schema."""
+    if options is None:
+        options = {}
+
+    blob_field_names = {
+        field.name for field in fields if 'blob' in str(field.type).lower()
```
JingsongLi: [P2] This detects BLOB columns by substring-matching the rendered type, which is broader than Java's `field.type().getTypeRoot() == DataTypeRoot.BLOB`. A normal top-level ROW column whose nested field is named or typed with `blob` will now be treated as a top-level BLOB column, so create-table validation starts requiring `row-tracking.enabled` and `data-evolution.enabled` even though Java would not. For example `payload ROW<blob_name STRING>` renders with `blob` in the type string and becomes part of `blob_field_names`. Please check only top-level BLOB types here (for example `getattr(field.type, 'type', None) == 'BLOB'` or an `AtomicType`/type-root helper) and add a regression test for nested fields containing the word `blob`.

JingsongLi (commented):
Another schema-validation detail from comparing the Python check with Java's top-level BLOB detection.

paimon-python/pypaimon/read/split_read.py
discivigour: supported.

paimon-python/pypaimon/schema/schema_manager.py
discivigour: added

paimon-python/pypaimon/schema/schema_manager.py
discivigour: changed

JingsongLi (commented):
LGTM, thanks.

discivigour:
> LGTM, thanks.

Thanks, rebased

PR #8023, opened 2026-05-28
[rest] Remove renameBranch REST endpoint

qingwei727:
### Purpose
  The server-side implementation of `renameBranch` ultimately relies on the
  underlying object store's rename (e.g. OSS), which is **not atomic**. While
  the rename is in progress, concurrent writes to the source branch can be
  silently lost, and there is currently no reliable way for the server to
  fence those writes during the operation.

  Because of this consistency risk, we'd rather not offer `renameBranch` over
  the REST catalog at all than offer it with a known data-loss window. This
  PR retires only the REST surface — the `Catalog` interface and the non-REST
  implementations (file-system branch manager, Spark/Flink procedures, etc.)
  are intentionally left untouched, so users on those code paths are not
  affected.

### Tests
 `RESTCatalogTest#testBranches`


sundapeng:
+1. For object storage backends, renameBranch cannot be implemented
atomically given the current branch structure, and silently doing a
non-atomic rename is worse than throwing UnsupportedOperationException.
Removing it from the REST surface prevents implementors from providing a
broken API.

leaves12138 (approved):
Thanks for the cleanup. Removing the unused `renameBranch` REST API surface is consistent across the client API, resource path, request object, REST catalog override, and tests. CI is green. Looks good to me.

PR #8024, opened 2026-05-28
[python] Use pypaimon-rust 0.2.0 from PyPI instead of building from source

JingsongLi:


leaves12138 (approved):
LGTM.

I checked the workflow change and the reduced large-blob test size. `pypaimon-rust==0.2.0` imports `pypaimon_rust.datafusion` successfully without a separate `datafusion` Python package, the changed Python test file compiles, and the PR checks are green.

PR #8025, opened 2026-05-28
[Schema] Fix catalog getTable failing for bucketed append tables created with Paimon 0.7

ArnavBalyan:
### Purpose
 - Paimon 0.7 allowed creating fixed bucket append tables without a bucket-key.
 - The bucket validation added in 1.0 runs on every table read and fails for the legacy tables which dont have a key.
 - The deserializer already handles this with bucket=1 for old schemas, but validation does not check this.
 - Skip this check for older versions.

### Tests
 - UT Added
 - CI


ArnavBalyan:
cc @JingsongLi thanks!

leaves12138 (approved):
Thanks for the compatibility fix. The exemption is narrowly scoped to legacy Paimon 0.7 append tables with `bucket = 1`, while current schemas and other bucket counts still require a bucket key. The added tests cover these cases and CI is green. Looks good to me.

Issue #4957, opened 2025-01-20
[Bug] function copyWithLatestSchema() is not correct


dahai1996:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Paimon version

1.0.0

### Compute Engine

flink

### Minimal reproduce step

no

### What doesn't meet your expectations?

see : https://github.com/apache/paimon/blob/316fa28041bbe4db6c8eb5489d6afd561d82db9f/paimon-core/src/main/java/org/apache/paimon/table/FileStoreTable.java#L100-L101

https://github.com/apache/paimon/blob/316fa28041bbe4db6c8eb5489d6afd561d82db9f/paimon-core/src/main/java/org/apache/paimon/table/AbstractFileStoreTable.java#L356-L367

```java
Map<String, String> options = tableSchema.options(); 
```
here is old old options,Is it because of this reason that we did not get the latest options?

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #8026, opened 2026-05-28
[Table] Fix copyWithLatestSchema option drop after table was opened

ArnavBalyan:
### Purpose
 - copyWithLatestSchema overwrites the latest schema options with the old handle options
 - Options committed via ALTER TABLE after the table was opened are silently dropped
 - Ensure that correct schema options are used when schema is refreshed via copyWithLatestSchema.
 - Closes user reported bug https://github.com/apache/paimon/issues/4957

### Tests
 - UT


ArnavBalyan:
cc @JingsongLi thanks! :)

leaves12138 (approved):
Thanks for the fix. `copyWithLatestSchema` now refreshes the latest schema including altered options, and the added regression test covers the dropped-options case. CI is green. Looks good to me.

PR #8027, opened 2026-05-28
[python] Rename DataBlobWriter to DedicatedFormatWriter and support blob+vector splitting

JingsongLi:
DataBlobWriter now handles normal + blob + vector columns (matching Java's DedicatedFormatRollingFileWriter), so rename it to DedicatedFormatWriter. Also add data evolution format tests covering parquet, blob, and vector paths.

### DIFF
--- a/paimon-python/pypaimon/read/reader/format_vortex_reader.py
+++ b/paimon-python/pypaimon/read/reader/format_vortex_reader.py
@@ -86,29 +86,23 @@ class FormatVortexReader(RecordBatchReader):
             PyarrowFieldParser.from_paimon_schema(read_fields) if read_fields else None
         )
 
-        # Collect predicate-referenced fields for targeted view type casting
-        self._cast_fields = predicate_fields if predicate_fields and vortex_expr is not None else set()
-
     @staticmethod
-    def _cast_view_types(batch: RecordBatch, target_fields: Set[str]) -> RecordBatch:
-        """Cast string_view/binary_view columns to string/binary, only for target fields."""
-        if not target_fields:
-            return batch
+    def _cast_view_types(batch: RecordBatch) -> RecordBatch:
+        """Cast all string_view/binary_view columns to string/binary."""
         columns = []
         fields = []
         changed = False
         for i in range(batch.num_columns):
             col = batch.column(i)
             field = batch.schema.field(i)
-            if field.name in target_fields:
-                if col.type == pa.string_view():
-                    col = col.cast(pa.utf8())
-                    field = field.with_type(pa.utf8())
-                    changed = True
-                elif col.type == pa.binary_view():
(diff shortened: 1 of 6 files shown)


leaves12138 (changes requested):
Thanks for the cleanup and the new data evolution format coverage. I found one blocking edge case in the new `DedicatedFormatWriter` path.

For a table that has blob columns and `vector.file.format` configured, `FileStoreWrite` now always chooses `DedicatedFormatWriter`. If a partial write only contains vector columns, e.g. `new_write().with_write_type(['embed'])`, then `normal_column_names` becomes empty. `write()` still calls `_process_normal_data(None)`, which currently returns `pa.Table.from_batches([])` and raises `ValueError: Must pass schema, or at least one RecordBatch` before the vector writer can commit anything.

Minimal reproduction:

```python
pa_schema = pa.schema([
    ('id', pa.int64()),
    ('doc', pa.large_binary()),
    ('embed', pa.list_(pa.float32(), 3)),
])
schema = Schema.from_pyarrow_schema(pa_schema, options={
    'row-tracking.enabled': 'true',
    'data-evolution.enabled': 'true',
    'vector.file.format': 'parquet',
})
catalog.create_table('default.blob_vec_partial', schema, False)
table = catalog.get_table('default.blob_vec_partial')
writer = table.new_batch_write_builder().new_write().with_write_type(['embed'])
writer.write_arrow(pa.table({
    'embed': pa.FixedSizeListArray.from_arrays(
        pa.array([1.0, 0.0, 0.0, 0.0, 1.0, 0.0], type=pa.float32()), 3),
}))
```

This fails at `DedicatedFormatWriter._process_normal_data`. We need to handle no-normal-column partial writes here, and ensure `_close_current_writers()` can still flush vector/blob-only pending data when there is no normal file in the same batch. A focused test for blob+vector table with `with_write_type(['embed'])` would pin this.

leaves12138 (approved):
LGTM.

I rechecked the latest head. The previous vector-only partial-write issue for blob+vector tables is fixed: no-normal-column writes now skip the empty normal batch path, flush the dedicated vector metadata correctly, and the new regression coverage exercises that scenario. I also reviewed the Vortex view-type casting change and the updated blob/vector data-evolution tests.

Validated locally:
- `python3 -m pytest -q pypaimon/tests/data_evolution_formats_test.py::DataEvolutionFormatsTest::test_blob_vector_partial_write_vector_only pypaimon/tests/data_evolution_formats_test.py::DataEvolutionFormatsTest::test_blob_and_vector_with_vector_file_format`
- `python3 -m py_compile pypaimon/write/writer/dedicated_format_writer.py pypaimon/read/reader/format_vortex_reader.py pypaimon/tests/data_evolution_formats_test.py`

PR #8029, opened 2026-05-29
[python][daft] Make Daft Paimon read source serializable

kerwin-zk:
### Purpose
  Make the Daft Paimon read source serializable when running with Ray.

  Previously, `PaimonDataSource` and fallback read tasks could retain live
  `FileStoreTable`, `FileIO`, `StorageConfig`, or `TableRead` objects. With
  remote filesystems such as OSS/Jindo, Ray failed to serialize the execution
  plan because those objects may contain non-picklable PyArrow filesystem state.

```
RuntimeError: Failed to serialize: OtherString("TypeError: no default __reduce__ due to non-trivial __cinit__")
```

### Tests
CI

### DIFF
--- a/paimon-python/pypaimon/daft/daft_datasource.py
+++ b/paimon-python/pypaimon/daft/daft_datasource.py
@@ -46,7 +46,6 @@ if TYPE_CHECKING:
     from pypaimon.common.predicate import Predicate
     from pypaimon.manifest.schema.data_file_meta import DataFileMeta
     from pypaimon.read.explain import ExplainSplitInfo
-    from pypaimon.read.table_read import TableRead
     from pypaimon.read.split import Split
     from pypaimon.table.file_store_table import FileStoreTable
 
@@ -58,6 +57,8 @@ logger = logging.getLogger(__name__)
 PAIMON_FILE_FORMAT_PARQUET = "parquet"
 PAIMON_FILE_FORMAT_ORC = "orc"
 PAIMON_FILE_FORMAT_AVRO = "avro"
+
+_PaimonIdentifier = tuple[str, str, str | None]
 
 
 @dataclass(frozen=True, slots=True)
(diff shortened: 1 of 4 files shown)


paimon-python/pypaimon/daft/daft_datasource.py
```diff
@@ -63,6 +64,113 @@ class _ReadPushdownState:
...
+        return dict(data)
+
+    return {}
+
+
+def _extract_catalog_options(table: FileStoreTable) -> dict[str, Any]:
+    file_io = getattr(table, "file_io", None)
+    properties = getattr(file_io, "properties", None)
+    if properties is None:
+        properties = getattr(file_io, "catalog_options", None)
```
YannByron: We can all obtain properties information through the `properties` attribute, even for `RESTTokenFileIO`. However, we need to determine whether it is `CachingFileIO` (in which case it needs to be obtained from `_delegate`) to increase code robustness.
I further recommend unifying the abstract `properties` API for `FileIO` on the pypaimon side. Because we currently have too many `getattr` calls.

paimon-python/pypaimon/daft/daft_datasource.py
```diff
@@ -63,6 +64,113 @@ class _ReadPushdownState:
...
+    return _options_to_dict(properties)
+
+
+def _extract_identifier(table: FileStoreTable) -> _PaimonIdentifier | None:
+    identifier = getattr(table, "identifier", None)
+    if identifier is None:
+        return None
+
+    get_database_name = getattr(identifier, "get_database_name", None)
```
YannByron: Maybe it's fine to call `identifier.get_database_name` and `identifier.get_table_name` directly, not though `getattr`.

paimon-python/pypaimon/daft/daft_datasource.py
```diff
@@ -63,6 +64,113 @@ class _ReadPushdownState:
...
+        if callable(get_branch_name)
+        else getattr(identifier, "branch", None)
+    )
+    if database_name is None or table_name is None:
+        return None
+    return database_name, table_name, branch_name
+
+
+def _extract_table_options(table: FileStoreTable) -> dict[str, Any]:
+    table_schema = getattr(table, "table_schema", None)
```
YannByron: Maybe let's define `schema` method in `FileStoreTable`.

paimon-python/pypaimon/daft/daft_datasource.py
```diff
@@ -277,7 +458,12 @@ async def get_tasks(self, pushdowns: Pushdowns) -> AsyncIterator[DataSourceTask]
                 and not has_deletion_vectors
             )
 
-            if can_use_native_reader:
+            use_paimon_reader_task = (
```
YannByron: Please provide detailed notes here regarding the scenarios in which the Daft Native Reader can be used, and those in which the Paimon Reader is necessary.

paimon-python/pypaimon/daft/daft_datasource.py
```diff
@@ -492,6 +665,14 @@ def _source_limit(
     def _requires_fallback_reader(self) -> bool:
         return not self._is_parquet or self._has_blob_columns or self._table.is_primary_key_table
 
+    def _requires_serializable_paimon_reader_task(self) -> bool:
+        if self._warehouse_scheme in ("", "file"):
```
YannByron: In ray environment, to scan a normal append-only paimon table location on aliyun `oss`, we expect this can run in daft native reader way.  But when `_warehouse_scheme` of this case is `oss`, this method will return true, and `use_paimon_reader_task` will return true. This does not meet our expectations. Correct me If i'm wrong please.

YannByron:
@kerwin-zk Thank you for working on this. I left some comments.

paimon-python/pypaimon/daft/daft_datasource.py
kerwin-zk: Done. Added a `properties` property on `CachingFileIO` that delegates to its `_delegate`, so every `FileIO` implementation now exposes `.properties` uniformly. `_extract_catalog_options` reads `table.file_io.properties` directly, with no per-implementation `getattr`.

paimon-python/pypaimon/daft/daft_datasource.py
kerwin-zk: Done, calling `identifier.get_database_name()` / `get_table_name()` / `get_branch_name()` directly now. This also fixes a latent issue: the old `getattr` fallback used `identifier.object`, which is the encoded object name and would round-trip incorrectly for branch tables.

paimon-python/pypaimon/daft/daft_datasource.py
kerwin-zk: Done. Added `FileStoreTable.schema()` returning the `TableSchema`, and `_extract_table_options` now uses `table.schema().options`.

paimon-python/pypaimon/daft/daft_datasource.py
kerwin-zk: Added detailed comments above the reader-selection logic in `get_tasks`, describing when the Daft native Parquet reader is used and when the pypaimon reader task is required (non-Parquet, blob columns, LSM merge, or deletion vectors).

paimon-python/pypaimon/daft/daft_datasource.py
kerwin-zk: You're right, thanks. I removed the gate: a normal append-only Parquet table on OSS now goes through Daft's native reader under the Ray runner. This is safe because both the source and the pypaimon fallback task serialize only rebuildable metadata (catalog options, identifier, table path), and the native task carries Daft's own picklable `StorageConfig`. The pypaimon reader task is now used only for splits that genuinely need it (PK/LSM merge, non-Parquet, blob columns, deletion vectors). I verified end-to-end that community Daft reading an append-only Parquet table on OSS under the Ray runner uses the native reader and returns correct results.

leaves12138 (approved):
Thanks for the fix. Rebuilding the Daft source/task from catalog options, identifier/table path, table options, and pushdown state avoids serializing non-picklable FileIO/storage objects while preserving the native-reader and pypaimon-reader paths. The added serialization/Ray-task tests cover the intended behavior, and CI is green. Looks good to me.

YannByron:
+1

PR #8030, opened 2026-05-29
[python] optimize blob format reader by inline blob data

steFaiz:
### Purpose
I'm testing random access of python's blob. Reading 100 0.5MB blobs cost 30s even blob-as-descriptor = true. This is because each blob will open the input stream once. 

This PR is same as https://github.com/apache/paimon/pull/6989

### Tests
Unit Tests

### DIFF
--- a/paimon-python/pypaimon/read/reader/format_blob_reader.py
+++ b/paimon-python/pypaimon/read/reader/format_blob_reader.py
@@ -16,7 +16,7 @@
 # under the License.
 
 import struct
-from typing import List, Optional, Any, Iterator
+from typing import List, Optional, Any, Iterator, BinaryIO
 
 import pyarrow as pa
 import pyarrow.dataset as ds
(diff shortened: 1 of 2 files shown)


leaves12138 (approved):
LGTM.

I reviewed the blob reader stream reuse change and the descriptor/inline-read behavior. Reusing the already-open stream for inline blob materialization looks correct, while descriptor reads still close the stream after loading the index and keep the existing lazy `BlobRef` behavior.

Validated locally:
- `python3 -m pytest -q pypaimon/tests/blob_test.py::BlobEndToEndTest::test_blob_read_inline_bytes_reuses_reader_stream pypaimon/tests/blob_test.py::BlobEndToEndTest::test_blob_end_to_end pypaimon/tests/blob_test.py::BlobEndToEndTest::test_blob_end_to_end_with_descriptor pypaimon/tests/blob_test.py::BlobEndToEndTest::test_null_blob_read pypaimon/tests/blob_test.py::BlobEndToEndTest::test_null_blob_read_as_descriptor`
- `python3 -m py_compile pypaimon/read/reader/format_blob_reader.py pypaimon/tests/blob_test.py`

JingsongLi:
+1

PR #8031, opened 2026-05-29
[python] Add HDFS native FileIO backend (no Hadoop install required)

TheR1sing3un:
Introduces HdfsNativeFileIO backed by the hdfs-native protocol client (Rust + PyO3)

Default backend for hdfs:// and viewfs:// switches to native; the PyArrow / libhdfs path is kept, with auto-fallback when hdfs-native is unavailable (e.g. on Windows or when the extra is not installed).

Adds: HdfsNativeFileIO, HdfsOptions, _kerberos helpers, unit tests, Docker-based e2e scaffold, native vs pyarrow benchmark, README section.

### DIFF
--- a/paimon-python/pypaimon/filesystem/hdfs_native_file_io.py
+++ b/paimon-python/pypaimon/filesystem/hdfs_native_file_io.py
@@ -1,0 +1,638 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+"""HDFS FileIO backed by the hdfs-native protocol client (no JVM, no libhdfs)."""
+
+import logging
+import os
+import xml.etree.ElementTree as ET
+from datetime import datetime, timezone
+from pathlib import PurePosixPath
+from typing import Dict, Optional
+from urllib.parse import urlparse
+
+import pyarrow
+import pyarrow.fs as pafs
+
+from pypaimon.common.file_io import FileIO
+from pypaimon.common.options import Options
+from pypaimon.common.options.config import HdfsOptions, SecurityOptions
+from pypaimon.common.uri_reader import UriReaderFactory
+from pypaimon.filesystem import _kerberos
+from pypaimon.schema.data_types import AtomicType, DataField, PyarrowFieldParser
+from pypaimon.write.blob_format_writer import BlobFormatWriter
+
+
+class _HdfsFileInfo:
+    """pafs.FileInfo-shaped adapter built from hdfs_native.FileStatus."""
+    __slots__ = ('path', 'size', 'type', 'mtime', 'base_name')
+
+    def __init__(self, path: str, size: Optional[int], file_type, mtime):
+        self.path = path
+        self.size = size
+        self.type = file_type
+        self.mtime = mtime
+        self.base_name = path.rsplit('/', 1)[-1] if path else ''
+
+
+class _HdfsWriterAdapter:
+    """File-like wrapper over hdfs_native.FileWriter."""
+
+    def __init__(self, fw):
+        self._fw = fw
+        self._pos = 0
+        self._closed = False
+
+    def write(self, buf) -> int:
+        n = self._fw.write(buf)
+        if n is None:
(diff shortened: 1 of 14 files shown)


leaves12138:
Thanks for the contribution. This PR is still in draft state, so I am leaving it unapproved for now. Please request review again once it is ready for a full review.

TheR1sing3un:
> Thanks for the contribution. This PR is still in draft state, so I am leaving it unapproved for now. Please request review again once it is ready for a full review.

Thanks, ready for review now!

paimon-python/setup.py
```diff
@@ -181,6 +181,9 @@ def read_requirements():
             'pypaimon-rust; python_version>="3.10"',
             'datafusion>=52; python_version>="3.10"',
         ],
+        'hdfs': [
+            'hdfs-native>=0.13,<1; platform_system!="Windows"',
```
JingsongLi: Could we add a Python version marker here? `pypaimon` still declares `python_requires=">=3.6"` and the CI matrix covers older Python versions, but the published `hdfs-native>=0.13` packages declare `Requires-Python: >=3.10`. With the current extra, `pip install "pypaimon[hdfs]"` on Python 3.6-3.9 will try to resolve this dependency and fail instead of installing `pypaimon` and letting users keep using the legacy pyarrow backend. Something like `hdfs-native>=0.13,<1; python_version >= "3.10" and platform_system != "Windows"` (plus a doc note for older Python) would avoid breaking those environments.

paimon-python/pypaimon/filesystem/hdfs_native_file_io.py
```diff
@@ -0,0 +1,638 @@
...
+        self._setup_kerberos()
+
+        config_dir = (
+            self.properties.get(HdfsOptions.HDFS_CONF_DIR)
+            or os.environ.get("HADOOP_CONF_DIR")
+        )
+        hadoop_xml = self._load_hadoop_xml(config_dir)
+
+        config = self._build_config_dict()
+        self._maybe_inject_viewfs_fallback(scheme, netloc, config, hadoop_xml)
```
JingsongLi, line 202: This fallback only looks at `hadoop_xml`, so it misses the REST/catalog-options path that the PR documents. For a zero-file viewfs config like `fs.viewfs.mounttable.cluster.link./prod=hdfs://ns1/prod` and `dfs.nameservices=ns1` passed in catalog options, `_build_config_dict()` puts those values in `config`, but `_maybe_inject_viewfs_fallback()` searches only `hadoop_xml` for existing links/nameservices. Because the comment says hdfs-native rejects viewfs without `linkFallback`, the documented catalog-only viewfs setup can still fail during `Client(**client_kwargs)`. Could we derive the fallback from the merged config (`{**hadoop_xml, **config}`) while still writing the injected key into `config`?

paimon-python/pypaimon/filesystem/hdfs_native_file_io.py
```diff
@@ -0,0 +1,638 @@
...
+                writer = BlobFormatWriter(output_stream)
+                for i in range(num_rows):
+                    writer.write_value(records_dict[field_name][i],
+                                       fields, self.uri_reader_factory)
+                writer.close()
+        except Exception as e:
+            self.delete_quietly(path)
+            raise RuntimeError(f"Failed to write blob file {path}: {e}") from e
+
+    def close(self):
```
JingsongLi, line 688: Since `FileIO.get()` now makes this class the default for `hdfs://` and `viewfs://`, this needs to keep the same write surface as the previous `PyArrowFileIO` backend. Right now `HdfsNativeFileIO` implements parquet/orc/avro/blob but not `write_lance()` or `write_vortex()`, so any HDFS table configured with `file.format=lance` or `file.format=vortex` will start hitting the abstract `FileIO.write_lance/write_vortex` `NotImplementedError` after this change. The existing writers call these methods directly for those formats. Could you either implement/delegate these two methods here (the fsspec facade may be enough) or avoid routing those formats to native by default?

paimon-python/setup.py
TheR1sing3un: nice suggestion! done!

paimon-python/pypaimon/filesystem/hdfs_native_file_io.py
TheR1sing3un: nice suggestion! done!

paimon-python/pypaimon/filesystem/hdfs_native_file_io.py
TheR1sing3un: nice suggestion! done!

JingsongLi (commented):
Re-reviewed the latest head `96bcecb`. The three issues from my previous pass look addressed: the `hdfs-native` extra is now gated to Python 3.10+, viewfs fallback is derived from merged XML/catalog options, and the native backend now overrides lance/vortex writes. I also ran the targeted HDFS native and Kerberos/FileIO tests locally; no new blocking issues found from my side.

TheR1sing3un:
> Re-reviewed the latest head `96bcecb`. The three issues from my previous pass look addressed: the `hdfs-native` extra is now gated to Python 3.10+, viewfs fallback is derived from merged XML/catalog options, and the native backend now overrides lance/vortex writes. I also ran the targeted HDFS native and Kerberos/FileIO tests locally; no new blocking issues found from my side.

Thanks! Any issue before landing it?

PR #8032, opened 2026-05-29
[spark] Refactor SQL extensions parser into base + per-version wrappers

kerwin-zk:
### Purpose
Refactor SQL extensions parser into base + per-version wrappers

### Tests
CI


leaves12138 (approved):
Thanks for the refactor. I reviewed the parser split into the shared `AbstractPaimonSparkSqlExtensionsParser` plus Spark 3/4 wrappers.

The method split looks consistent with the versioned `ParserInterface` requirements: Spark 3 delegates through the wrapper, and Spark 4 wrappers provide the additional `parseRoutineParam` override while reusing the common parser logic. I also checked the local Spark 4.0.2 / 4.1.2 `ParserInterface` signatures with `javap`.

I tried to compile the Spark 4 modules locally with JDK 17, but the build was blocked before reaching the Spark parser modules by missing cached snapshot artifacts (for example `paimon-core:tests` / `paimon-spark-common_2.13` snapshots) in the local Maven repository. I did not find any code-level blocker in the changed parser files.

JingsongLi:
+1

PR #8033, opened 2026-05-29
[python] supports random access of blob files

steFaiz:
### Purpose
Just like Vortex & Lance & Row, push row indices to blob files.

### Tests
See BlobTests & BlobTableTests


leaves12138 (approved):
Thanks for the patch. I reviewed the blob random-access path and the row-index plumbing from `IndexedSplit` into `FormatBlobReader`.

I also ran the targeted Python checks locally:

`PYTHONPATH=. python3 -m pytest -q pypaimon/tests/blob_test.py pypaimon/tests/blob_table_test.py`

They passed (`107 passed`). The row-index behavior also looks consistent for empty, duplicate, and out-of-order local offsets. Looks good to me.

JingsongLi (approved):
+1

PR #8034, opened 2026-05-29
[spark] Fix schema merge creating duplicate columns with case-mismatched names

Zouxxyy:
### Purpose

When `merge-schema` is enabled and source column names differ only in case from target columns (e.g. source `ID` vs target `id`), `SchemaMergingUtils` treats them as new columns due to case-sensitive `HashMap` lookups. This causes duplicate columns in the schema and makes the table unreadable (`Field names must be unique`).

This PR adds a `caseSensitive` parameter through the schema merge chain (`SchemaMergingUtils` → `SchemaManager` → `FileStore` → Spark `SchemaHelper`), using `TreeMap(String.CASE_INSENSITIVE_ORDER)` for field matching when `caseSensitive=false`. Spark callers pass `spark.sql.caseSensitive` config (default `false`).

Affects both `INSERT ... merge-schema=true` and `MERGE INTO ... merge-schema=true` paths.

### Tests

Added 13 case-sensitivity tests in `WriteMergeSchemaTest` covering:
- INSERT and MERGE INTO with case-mismatched column names
- Nested struct fields with case mismatch
- Schema unchanged when only case differs (no new columns)
- Repeated writes with alternating case
- Mixed case-mismatch with genuinely new columns
- Case-sensitive mode correctly treats different case as new columns

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/AbstractFileStore.java
+++ b/paimon-core/src/main/java/org/apache/paimon/AbstractFileStore.java
@@ -271,9 +271,9 @@ abstract class AbstractFileStore<T> implements FileStore<T> {
     }
 
     @Override
-    public boolean mergeSchema(RowType rowType, boolean allowExplicitCast) {
+    public boolean mergeSchema(RowType rowType, boolean allowExplicitCast, boolean caseSensitive) {
         return schemaManager.mergeSchema(
-                rowType, allowExplicitCast, catalogEnvironment.schemaModification());
+                rowType, allowExplicitCast, caseSensitive, catalogEnvironment.schemaModification());
     }
 
     @Override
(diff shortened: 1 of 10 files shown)


leaves12138 (approved):
Thanks for the patch. I reviewed the case-sensitive/case-insensitive schema merge path through Spark `SchemaHelper`, `MergeSchemaEvolutionHelper`, `FileStore.mergeSchema`, `SchemaManager`, and `SchemaMergingUtils`.

The case-insensitive field map now avoids creating duplicate columns for case-mismatched existing fields, while the case-sensitive path still treats different-case names as distinct. I also checked that the generated schema changes keep the existing table field names for matched columns.

I ran the targeted core test locally:

`mvn -pl paimon-core -DskipITs -Dcheckstyle.skip -Drat.skip=true -Dtest=SchemaMergingUtilsTest test`

It passed. I tried the Spark `WriteMergeSchemaTest` as well, but the local run was blocked by missing snapshot artifacts in the Maven repository before executing tests. The added Spark coverage looks appropriate to me.

JingsongLi:
+1

PR #8035, opened 2026-05-29
[core][python] Extend comment directives for BLOB/VECTOR column creation

JingsongLi:
Support __BLOB_VIEW_FIELD, __BLOB_EXTERNAL_STORAGE_FIELD, and __VECTOR_FIELD;dim comment directives in both CREATE TABLE and ALTER TABLE ADD COLUMN. Rename BlobSchemaUtils to ColumnDirectiveUtils and consolidate all directive logic (parsing, type conversion, option modification, drop cleanup) into it so SchemaManager stays clean with single-method calls.

Key changes:
- ColumnDirectiveUtils: applyAddColumnDirective (one-stop for ADD COLUMN), applyDirectives (Schema-level for CREATE TABLE), removeDroppedDirectiveOptions (clean options on DROP COLUMN by BLOB/VECTOR type)
- SchemaManager: uses ColumnDirectiveUtils for create_table, addColumn, dropColumn
- Python: mirrors Java with column_directive_utils.py and updated schema_manager.py
- Docs: all SQL examples use COMMENT directive as primary approach with Flink SQL / Spark SQL / Java API / Python API tabs; removed blob-field/vector-field option references from docs


leaves12138 (changes requested):
Thanks for the patch. I found two blocking issues.

First, the current PR does not pass the targeted core tests. Running

`mvn -pl paimon-core -DskipITs -Dcheckstyle.skip -Drat.skip=true -Dtest='ColumnDirectiveUtilsTest,SchemaEvolutionTest' test`

fails in `SchemaEvolutionTest.testAddBlobColumnErrors`:

`Expecting code to raise a throwable.`

The failing assertion is the `SchemaChange.setOption(CoreOptions.BLOB_FIELD.key(), "k")` case with `hasSnapshots=true`. Since `blob-field` is still documented/annotated as immutable, this should either continue to be rejected or the test/immutability contract needs to be updated consistently.

Second, the Python implementation is not aligned with Java for legacy fallback keys. Java `ColumnDirectiveUtils.modifyFieldOptions` migrates fallback values such as `blob.stored-descriptor-fields` into the canonical `blob-descriptor-field` before appending, and `removeDroppedDirectiveOptions` also removes dropped fields from fallback keys. Python `_modify_field_options` and `remove_dropped_directive_options` only touch canonical keys. For example:

```python
opts = {"blob.stored-descriptor-fields": "old"}
apply_add_column_directive("__BLOB_DESCRIPTOR_FIELD", "new_col", AtomicType("BYTES"), opts)
```

currently produces both `blob.stored-descriptor-fields=old` and `blob-descriptor-field=new_col`, losing the Java migration behavior. Dropping a BLOB field also leaves `blob.stored-descriptor-fields` unchanged.

Could you fix these and add coverage for the Java failing case plus Python fallback-key migration/removal?

leaves12138 (approved):
Thanks for the update. I rechecked the latest head (`ca3e98d`) and the two issues from my previous review are fixed:

- the Java schema-evolution/directive tests now pass for the immutable option path;
- Python now migrates/removes the legacy `blob.stored-descriptor-fields` fallback key consistently with Java, and has coverage for it.

I also reran the targeted checks locally:

`mvn -pl paimon-core -am -DskipITs -Dcheckstyle.skip -Drat.skip=true -DfailIfNoTests=false -Dtest='ColumnDirectiveUtilsTest,SchemaEvolutionTest' test`

`python3 -m pytest -q pypaimon/tests/column_directive_utils_test.py`

Both passed. Looks good to me.

Issue #8005, opened 2026-05-28
[feature] Enhance COPY INTO with multi-format support and advanced options


JunRuiLee:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon/issues) and found nothing similar.

### Motivation

In production environments, bulk data import and export is a common and critical operation — loading datasets from external storage into Paimon tables, or exporting table data to files for downstream consumption. The `COPY INTO` statement provides a declarative SQL interface for these operations without requiring users to write custom ETL pipelines.

The initial `COPY INTO` implementation was recently introduced with basic CSV support. This umbrella issue tracks the effort to extend it with multi-format support and advanced options, aligning with the commonly adopted capabilities in the industry.

### Planned Features

  We plan to implement the following capabilities as separate PRs:

  #### Format Support
  - [x] **JSON format** — Support `FILE_FORMAT = (TYPE = JSON)` for both import and export. Options: `MULTI_LINE`, `COMPRESSION`, `NULL_IF`, `EMPTY_FIELD_AS_NULL`. JSON import uses column-name matching by default (no positional dependency).
  - [x] **Parquet format** — Support `FILE_FORMAT = (TYPE = PARQUET)` for both import and export. Leverages Spark's native schema inference. Options: `COMPRESSION`.

  #### Error Handling
  - [x] **ON_ERROR = CONTINUE / SKIP_FILE** — Currently only `ABORT_STATEMENT` is supported. `CONTINUE` skips bad rows and reports error counts per file. `SKIP_FILE` skips entire files that fail and reports per-file LOADED/LOAD_FAILED status.

  #### Export Enhancement
  - [ ] **FROM (SELECT ...) query export** — Currently export only supports `FROM table_name`. This adds support for arbitrary SQL queries as the data source for `COPY INTO `.

 
  ### Considered but deprioritized

  - **MATCH_BY_COLUMN_NAME / ENABLE_SCHEMA_EVOLUTION** — Prototyped locally but deprioritized for now; out of scope for this issue.

  ### Proposed SQL Syntax

  #### Import
  ```sql
  COPY INTO table_name [(col1, col2, ...)]
  FROM 'source_path'
  FILE_FORMAT = (TYPE = CSV | JSON | PARQUET [, option = value, ...])
  [PATTERN = 'regex']
  [FORCE = TRUE | FALSE]
  [ON_ERROR = ABORT_STATEMENT | CONTINUE | SKIP_FILE]

  Export

[...]

PR #8037, opened 2026-05-30
[spark] Support Parquet format in COPY INTO

JunRuiLee:
This PR adds Parquet format support for COPY INTO import and export, as part of #8005.

## Changes

**Import** (`COPY INTO table FROM path`):
- Read Parquet files with native typed schema (no string-then-cast like CSV/JSON)
- Column matching by name (case-insensitive), not by position
- Extra source columns are ignored; missing columns become NULL
- Cast validation: detects non-null → null after casting (type incompatibility)
- Supports explicit column list, PATTERN, FORCE, ON_ERROR = ABORT_STATEMENT

**Export** (`COPY INTO path FROM table`):
- Write Parquet files via `df.write.parquet()`
- COMPRESSION option (SNAPPY, GZIP, NONE, etc.)

**Refactoring**:
- Extract `resolveDefaultColumn()` shared helper (was duplicated in Parquet and text paths)
- Unify `recordHistoryAndBuildResults()` to accept a `countDf` parameter (eliminates ~45 lines of copy-paste between Parquet and text paths)
- Add `logWarning` when default value expression parsing fails (was silently swallowed)

## Tests

12 new tests covering: basic import, column name matching, explicit column list, export, export with compression, round-trip, extra fields ignored, missing fields become null, FORCE=FALSE dedup, PATTERN filtering, unsupported option error, rows_loaded count accuracy.


JingsongLi (approved):
+1

PR #8038, opened 2026-05-30
[project] Add security threat model and security docs page

JingsongLi:


leaves12138 (approved):
Thanks for adding this. The threat model clearly documents Paimon's trust boundaries, non-goals, scanner triage guidance, and links it into the project docs. This is docs-only, the generated site page is updated, and CI is green. Looks good to me.

PR #8039, opened 2026-05-30
[python][daft] Fix single-part catalog table lookup

QuakeWang:
### Purpose

Daft Catalog accepts single-part identifiers, and PaimonCatalog previously forwarded them to pypaimon as plain table names. For table APIs, this is invalid because pypaimon's Identifier parser
  requires `db.table`, so `get_table("missing_table")` and `drop_table("missing_table")` leaked `ValueError` instead of Daft `NotFoundError`.

This PR adds table-specific identifier handling in `PaimonCatalog`: single-part table identifiers are handled at the Daft catalog layer as not found, and `has_table("missing_table")` now returns `False`.

### Tests

- `pytest paimon-python/pypaimon/tests/daft/daft_catalog_test.py`
- `pytest paimon-python/pypaimon/tests/daft/daft_catalog_rest_test.py`


JingsongLi:
+1

PR #8040, opened 2026-05-30
[vortex] Refactor JNI layer to use DataSource/Session API replacing proto-based Array/DType

JingsongLi:
Replace the old protobuf-based JNI bindings (Array, DType, File, expressions package)
with a new DataSource/Session/Scan/Partition API that provides a cleaner abstraction
for Vortex file access. Update VortexRecordsReader, VortexRecordsWriter, and
VortexPredicateConverter to work with the new JNI interface.


leaves12138 (changes requested):
Thanks for the refactor. I think the new DataSource/Session shape is the right direction, but I found a few issues that should be fixed before merging:

1. `VortexRecordsReader` no longer pushes the projected columns into the Vortex scan. The old code called `scanBuilder.addAllColumns(physicalReadRowType.getFieldNames())`, while the new `ScanOptions` supports `projection()` but the reader never sets it. This means projected reads will scan the full physical schema and may also hide schema mismatches behind `ArrowBatchReader`'s name-based remapping. Please wire `physicalReadRowType.getFieldNames()` through `Expression.select(...)` (or the equivalent projection API) and add a projection read test.

2. `Partition.close()` intentionally skips `NativePartition.free(pointer)` after `scanArrow()` sets `consumed = true`. That is only safe if `NativePartition.scanArrow` transfers ownership of the native partition to the returned Arrow stream and the stream release callback frees it. I could not see that contract documented/enforced on the Java side, and otherwise every consumed partition leaks the native partition handle. Please either free consumed partitions after the `ArrowReader` is closed, or document and test the ownership transfer clearly.

3. `Scan.close()` does not release a prefetched partition if `hasNext()` primed `nextPartitionPointer` and the scan is closed before `next()` consumes it. This can happen for callers using the standard `Iterator` pattern or if an exception/close happens between `hasNext()` and `next()`. Please free the cached `nextPartitionPointer` before freeing the scan, unless the native scan owns and frees that outstanding partition.

4. `writeBundleVsr` allocates an `ArrowStreamReader`, `ArrowArray`, and `ArrowSchema` but does not close them on the success path, and it only closes `bundleAllocator` on the exception path. Since the native writer is described as async/zero-copy, this path should mirror the retained-writer lifetime used by `flush()` or otherwise make a synchronous copy/close contract explicit. As written, direct `ArrowBundleRecords` writes can leak Arrow/native resources or risk releasing memory too early if we later add success-path closes.

5. `VortexPredicateConverterTest` was weakened from exact expression assertions to only `assertNotNull`. That removes coverage for operator mapping, timestamp units/time zone, string/decimal literal conversion, and compound predicate structure. Please restore semantic assertions, even if the assertion needs to inspect/serialize the new expression representation instead of relying on Java object equality.

leaves12138 (changes requested):
Thanks for the update. The projection pushdown is now wired in `VortexRecordsReader`, and `Scan.close()` now releases a primed partition. I still have two remaining concerns before this can be merged:

1. `writeBundleVsr` now retains `bundleAllocator`, but the `ArrowStreamReader`, `ArrowArray`, and `ArrowSchema` allocated in this method are neither retained nor closed on the success path. Retaining only the allocator does not by itself make the Java wrapper resources lifecycle-safe. If Rust asynchronously uses the exported Arrow C data, please retain all resources that own/release that C data until after `nativeWriter.close()`, or use a clear synchronous-copy contract and close them immediately. As written, this path can still leak wrapper/native resources, and it is hard to reason about which object owns the release callbacks.

2. `VortexPredicateConverterTest` is still much weaker than before. The tests now only assert that the returned native pointer is non-zero, so a wrong operator mapping (for example GT vs GTE), wrong column, wrong timestamp unit, or wrong literal value would still pass. Please add a semantic assertion path for the new native expression representation, e.g. a debug/serialization method from the JNI layer, or keep expression construction testable through an inspectable Java representation. The current tests mainly cover "some native expression was allocated", not predicate conversion correctness.

The newly added integration tests are useful, but they do not replace unit-level checks for the converter mapping and resource ownership in the direct Arrow bundle path.

leaves12138 (approved):
Thanks for the follow-up updates. The issues from my previous reviews are addressed now: scan projection is pushed down, primed partitions are released on close, partition ownership transfer is documented, writer-side resources are retained for the exported Arrow data lifetime, and predicate conversion now has round-trip semantic coverage.

I also pushed a small follow-up commit to fix the two current checkstyle violations. The new CI run is still pending, but the code changes look good to me.

PR #8041, opened 2026-05-30
[python] Fix empty Ray overwrite commit

QuakeWang:
### Purpose

`PaimonDatasink.on_write_complete()` previously returned early when all Ray write tasks produced empty commit messages. This made `write_paimon(empty_ds, table, overwrite=True)` a silent no-op, so existing data was kept even though Paimon overwrite commit semantics require empty overwrite commits to reach `TableCommit`.

This PR keeps the empty-message fast path only for append writes. For overwrite writes, Ray now calls `TableCommit.commit([])`, allowing `FileStoreCommit.overwrite()` to delete the target range for static/unpartitioned overwrite and preserve dynamic-partition empty overwrite no-op semantics.

### Tests

- `python -m pytest pypaimon/tests/ray_sink_test.py pypaimon/tests/ray_integration_test.py::RayIntegrationTest::test_write_paimon_empty_overwrite_unpartitioned`

### DIFF
--- a/paimon-python/pypaimon/tests/ray_integration_test.py
+++ b/paimon-python/pypaimon/tests/ray_integration_test.py
@@ -306,6 +306,33 @@ class RayIntegrationTest(unittest.TestCase):
         df = result.to_pandas()
         self.assertEqual(list(df['id']), [3])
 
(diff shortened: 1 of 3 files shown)


JingsongLi:
+1

PR #8042, opened 2026-05-31
[spark] Decouple type-widening from merge-schema with an explicit switch

Zouxxyy:
### Purpose

Today `write.merge-schema` couples column-addition with *unconditional* type widening. That has two problems: it can attempt unsupported widenings (e.g. `ARRAY<INT>` -> `ARRAY<BIGINT>`) on a plain column-addition write, and the widening behavior is inconsistent between catalog writes and `MERGE INTO`.

This PR decouples the two with an explicit, opt-in switch:

| Option | Default | Effect |
|--------|---------|--------|
| `write.merge-schema` | false | Evolve schema for **new columns only**; existing column types are kept and incoming values are cast to them. |
| `write.merge-schema.type-widening` | false | Additionally widen an existing column type to a wider compatible type (e.g. `INT -> BIGINT`, `DECIMAL` precision increase). |
| `write.merge-schema.explicit-cast` | false | Additionally allow lossy casts (e.g. `BIGINT -> INT`, `STRING -> DATE`). |

Two things worth calling out:

1. **The write flow is `compute -> cast -> commit`.** For catalog writes (`saveAsTable` / `INSERT` / `writeTo`) and `MERGE INTO`, analysis first computes the merged (evolved) schema **in memory** (nothing is persisted yet) and casts the incoming data to it; the evolved schema is persisted only by the final commit, which is **deferred to execution**.

2. **Schema evolution on path-based DataFrame write is intentional.** `df.write.format("paimon").save(path)` and the streaming sink also evolve the schema. They have no analysis-time cast hook, so they commit the evolved schema and write the data as-is (works for column additions and widening, where the incoming data already matches the evolved types). This is by design, and is covered by the added `DataFrameWriteTest` / `PaimonSinkTest` cases.

### Tests

- `SchemaMergingUtilsTest` — typeWidening on/off matrix (core)
- `WriteMergeSchemaTest` / `V2WriteMergeSchemaTest`
- `DataFrameWriteTest` — schema-evolution groups across pk / bucket / format, plus explicit-cast
- `MergeIntoPrimaryKeyBucketedTableTest`, `PaimonSinkTest` (streaming schema evolution)


leaves12138 (approved):
Thanks for the refactor. Decoupling column-add schema evolution from existing-column type widening with `write.merge-schema.type-widening` makes the default behavior safer, and the Spark analysis/execution paths now share the same schema-evolution helper. The docs, generated options, core schema-merge tests, and Spark write/MERGE tests cover the new behavior. CI is green. Looks good to me.

JingsongLi:
+1

PR #8043, opened 2026-05-31
[python] Fix crash when updating all columns via update_by_arrow_with_row_id

XiaoHongbo-Hope:
### Purpose
Updating **all columns** of a row via `update_by_arrow_with_row_id`
  crashes with `ValueError: column_names cannot be empty`. This PR fixes it by resolving update columns from data when update type is unset.

### Tests
`test_update_columns_fall_back_to_data_when_unset`

### DIFF
--- a/paimon-python/pypaimon/tests/table_update_test.py
+++ b/paimon-python/pypaimon/tests/table_update_test.py
@@ -123,6 +123,45 @@ class _TableUpdateTestBase(DataEvolutionTestBase):
             result['city'].to_pylist(),
         )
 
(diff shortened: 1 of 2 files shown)


leaves12138 (approved):
Thanks for the fix. Falling back to all non-row-id columns when update columns are not explicitly set matches the behavior needed by `update_by_arrow_with_row_id`, and the added tests cover both the all-column update path and the row-id-only invalid input. CI is green. Looks good to me.

JingsongLi:
Please rebase master.

JingsongLi:
+1

PR #8045, opened 2026-05-31
[python] Reject or drop stale index when updating globally indexed columns

XiaoHongbo-Hope:
### Purpose
Updating a column covered by a global index via `update_by_arrow_with_row_id`
  caused the index to become outdated — subsequent queries through the index
  returned stale data instead of the updated values.

This PR adds the `global-index.column-update-action` option and enforces it
  during single-machine column updates:

  - **THROW_ERROR** (default): raises when the update touches indexed columns.
  - **DROP_PARTITION_INDEX**: drops affected index entries from the manifest.

  ### Tests

  - e2e: THROW_ERROR raises + DROP_PARTITION_INDEX succeeds with index removed.
  - Unit: index manifest write/read roundtrip and combine_deletes cases.


leaves12138 (approved):
Thanks for the fix. The default behavior now rejects updates that touch globally indexed columns, and the opt-in `DROP_PARTITION_INDEX` path deletes only the affected partition/index entries. The commit-message/index-manifest plumbing and cleanup handling are covered by tests, and CI is green. Looks good to me.

PR #8046, opened 2026-05-31
[python] Make HASH_FIXED Ray pre-clustering opt-in

QuakeWang:
### Purpose

Ray HASH_FIXED writes previously pre-clustered every table with `groupby().map_groups()` by `(partition_keys..., bucket)`. Ray requires each mapped group to fit in memory on one node, so large buckets or hot partitions could OOM during append-only writes.

This patch makes HASH_FIXED pre-clustering explicit:

  - default `auto` writes append-only HASH_FIXED tables directly
  - `map_groups` keeps the existing small-file optimization as an opt-in mode
  - HASH_FIXED primary-key tables fail fast in `auto/off`, because direct Ray writes can split one bucket across multiple writer tasks and allocate overlapping sequence numbers
  - `write_paimon()` and `TableWrite.write_ray()` use the same safety check
  
### Tests

  - `pytest pypaimon/tests/test_ray_shuffle_helper.py pypaimon/tests/ray_repartition_test.py`
  - `pytest pypaimon/tests/ray_integration_test.py pypaimon/tests/ray_data_test.py`

### DIFF
--- a/docs/docs/pypaimon/ray-data.md
+++ b/docs/docs/pypaimon/ray-data.md
@@ -205,19 +205,43 @@ write_paimon(
 )
 ```
 
-**Automatic (partition, bucket) clustering for HASH_FIXED tables:**
+**HASH_FIXED pre-clustering:**
 
-For HASH_FIXED tables, `write_paimon` automatically clusters rows by
-`(partition_keys..., bucket)` before writing so each (partition,
-bucket) lands in a single Ray task — one writer, one file group. This
-avoids the small-file storm that Ray's default round-robin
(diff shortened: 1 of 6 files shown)


leaves12138 (approved):
Thanks for the update. Making HASH_FIXED Ray pre-clustering explicit is a safer default: append-only HASH_FIXED tables now avoid the old memory-heavy `map_groups` path unless requested, while primary-key HASH_FIXED writes still fail fast unless `map_groups` is explicitly selected. The docs and tests cover the default/off/map_groups behavior and invalid modes. CI is green. Looks good to me.

PR #8047, opened 2026-05-31
[core][spark] Use DV-aware tight bounds for Spark MIN/MAX pushdown

kerwin-zk:
### Purpose
Spark currently disables MIN/MAX aggregate pushdown for any table with `deletion-vectors.enabled=true`. This is correct but too conservative: many DV-enabled non-primary-key tables, or many splits inside them, do not actually have deleted rows. In those cases the recorded file min/max stats are still tight and can safely answer MIN/MAX.

This PR makes the decision based on runtime split metadata instead of the table-level DV option. It derives whether a data file still has tight stats from `DataFileMeta.deleteRowCount` and the paired `DeletionFile.cardinality`, then allows Spark MIN/MAX pushdown only when every file in the split is tight.

This keeps the existing safety behavior for files with real deletes or unknown DV cardinality, while recovering MIN/MAX pushdown for DV-enabled tables/splits that have no effective deletions.

### Tests
CI

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/source/DvAwareStats.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/source/DvAwareStats.java
@@ -1,0 +1,41 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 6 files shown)


leaves12138 (approved):
Thanks for the fix. The new tight-bounds check keeps MIN/MAX pushdown disabled for primary-key tables, but allows DV-enabled append cases only when all files still have tight stats (no delete rows and no populated/unknown DV). The core and Spark tests cover the important paths. CI is green. Looks good to me.

PR #8050, opened 2026-05-31
[spark] Avoid serializing known splits in read builder

wzx140:
### Purpose

ReadonlyTable default implementation creates ReadBuilderImpl(this), which makes the ReadBuilder capture KnownSplitsTable. KnownSplitsTable contains all known splits, but the ReadBuilder returned from it should not carry all of those splits. Therefore, KnownSplitsTable delegates newReadBuilder to the origin table.

This issue can happen when merging into a data-evolution table. The Spark reader factory holds the ReadBuilder, ReadBuilderImpl holds the KnownSplitsTable through its table field, and KnownSplitsTable holds the full split list. As a result, each Spark task can become very large because it indirectly serializes all known splits.

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/table/KnownSplitsTable.java
+++ b/paimon-core/src/main/java/org/apache/paimon/table/KnownSplitsTable.java
@@ -21,6 +21,7 @@ package org.apache.paimon.table;
 import org.apache.paimon.fs.FileIO;
 import org.apache.paimon.table.source.InnerTableRead;
 import org.apache.paimon.table.source.InnerTableScan;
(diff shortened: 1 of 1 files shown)


leaves12138:
The code change is small, but I am holding off on approval because the current CI status is not green yet. Please fix or rerun the failing checks first.

wzx140:
@leaves12138 Could you help rerun the failed CI? It looks like a network issue:

Failed to connect to lumina-binary.oss-cn-shanghai.aliyuncs.com:443 (47.102.9.85): connection timed out.

JingsongLi:
Re-open to trigger tests.

PR #8051, opened 2026-05-31
[spark] Broadcast data evolution firstRowIdToPartitionMap

wzx140:
### Purpose

`firstRowIdToPartitionMap` is used by data evolution MERGE INTO writes to locate the original partition and row count for each data file from its `firstRowId`. This mapping is built from scanned `DataSplit`s and then used by each Spark task while writing evolved rows.

For merge plans that affect many files across many partitions, the map can become large. Capturing it directly in every task closure duplicates the serialized map for each task and can make Spark task serialization very large.

### Solution

Broadcast `firstRowIdToPartitionMap` once through Spark instead of embedding the whole map in every task closure.

The map stores partition `BinaryRow`s as serialized bytes. `BinaryRow` keeps its memory segments in transient fields and relies on Java serialization hooks to restore them, while Spark Kryo may serialize the object fields directly. Serializing the partition with `SerializationUtils.serializeBinaryRow` keeps the broadcast value Kryo-safe, and the writer deserializes the bytes before use.

### DIFF
--- a/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/commands/DataEvolutionPaimonWriter.scala
+++ b/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/commands/DataEvolutionPaimonWriter.scala
@@ -19,7 +19,6 @@
 package org.apache.paimon.spark.commands
 
 import org.apache.paimon.CoreOptions
-import org.apache.paimon.data.BinaryRow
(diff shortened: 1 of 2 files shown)


leaves12138:
Thanks for the update. I am holding off on approval for now because two CI jobs are still pending. The change touches Spark data-evolution write behavior, so I would like to see the full Spark/UT matrix finish before approving.

JingsongLi (approved):
+1

PR #8053, opened 2026-05-31
[python] Migrate PyPaimon packaging to pyproject.toml

### LINKED ISSUE #8049: [Feature] Migrate PyPaimon packaging to pyproject.toml
### Search before asking

### Motivation

PyPaimon still keeps most packaging metadata in `setup.py`. Moving the static metadata to `pyproject.toml` would make the packaging configuration easier to maintain, and it also follows the recommended Python packaging standard.

### Solution

Use `setuptools` as the build backend with `pyproject.toml`.

Although `uv` is very popular, it is still a third-party tool. Keeping `setuptools` as the backend gives us more flexibility, and users or developers can still use `uv` to install PyPaimon.

One thing to decide is the Python version requirement. PyPaimon currently declares `python_requires=">=3.6"`, but Python 3.6 reached EOL on 2021-12-23. If we still need to support Python 3.6, the last compatible `setuptools` versions have limited `pyproject.toml` metadata support, so we may only be able to move part of the metadata and keep the rest in `setup.py`.

I think we can raise the minimum Python version to 3.7 and migrate the packaging metadata to `pyproject.toml` more cleanly.

### Anything else?

_No response_

aisk:
### Purpose

Migrate PyPaimon packaging metadata from `setup.py` and requirements files to `pyproject.toml`.

This PR keeps `setuptools` as the build backend, raises the minimum Python version to 3.7, and moves development dependencies into the `dev` dependency group.

Closes #8049.

### Changes

- Add `paimon-python/pyproject.toml` with project metadata, runtime dependencies, optional dependencies, scripts, package discovery, and the `dev` dependency group.
- Remove `setup.py`, `dev/requirements.txt`, and `dev/requirements-dev.txt`.
- Update PyPaimon README and docs to use `python3 -m build --sdist`.
- Update `MANIFEST.in` now that requirements files are removed.

### Tests

- `python3 -c "import tomllib; tomllib.load(open('pyproject.toml','rb')); print('ok')"`
- `UV_CACHE_DIR=/tmp/uv-cache uv build --sdist`


aisk:
I found that there are some test cases failed on Python3.7, which are better to fix in another PR. I'll mark this as draft and will continue the work if the CI cased fixed.

paimon-python/pyproject.toml
```diff
@@ -0,0 +1,129 @@
...
+[build-system]
+requires = [
+    "setuptools>=61",
+    "wheel",
+]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "pypaimon"
+version = "1.5.dev"
```
JingsongLi, line 28: This drops the dynamic dev-version behavior from the old setup.py. Before this PR, `python setup.py sdist` registered an atexit hook that repacked `1.5.dev` into `1.5.devYYYYMMDD` based on the commit date. With the new static metadata I built the PR locally via `python -m build --sdist` and it only produced `pypaimon-1.5.dev0.tar.gz`. If snapshot/dev package publishing relies on monotonically increasing versions, repeated builds will all have the same normalized version and can no longer be uploaded or selected as newer builds. Could we port the dynamic version generation to the pyproject build path (for example via a build hook/setuptools-scm) or explicitly confirm that this behavior is no longer needed?

aisk: Hi, I'm not sure how we currently use nightly builds. AFAIK, Python users can install `pypaimon` directly from git, like:

`pip install "git+https://github.com/apache/paimon.git#subdirectory=paimon-python"`

If we want to publish nightly or snapshot builds, we can tag the related commit and let users install that tag with `pip install "git+https://github.com/apache/paimon.git@<tag>#subdirectory=paimon-python"`.

In principle, nightly or snapshot builds should have a tag or at least a clear revision. So unless an existing publishing workflow depends on the dynamic version, I think we can remove this behavior.

PR #8055, opened 2026-05-31
[python][daft] Split conjunctive filters for pushdown

QuakeWang:
### Purpose

Daft Paimon filter conversion previously treated each filter expression as all-or-nothing. For an expression like `A AND unsupported(B)`, the supported conjunct `A` was not pushed to Paimon, so it could not participate in scan planning or file skipping.

This PR flattens only `AND` conjuncts in the Daft adapter layer and converts each conjunct independently. Supported conjuncts are combined into a Paimon `AND` predicate, while unsupported conjuncts remain as Daft post-scan filters. `OR` expressions are still pushed only when the whole `OR` is supported.

### Tests

- `python3 -m pytest paimon-python/pypaimon/tests/daft/daft_data_test.py paimon-python/pypaimon/tests/daft/daft_explain_test.py -q`

### DIFF
--- a/paimon-python/pypaimon/daft/daft_predicate_visitor.py
+++ b/paimon-python/pypaimon/daft/daft_predicate_visitor.py
@@ -28,6 +28,7 @@ import logging
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any
 
+from daft.expressions import Expression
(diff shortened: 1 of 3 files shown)


XiaoHongbo-Hope:
+1

JingsongLi:
+1

PR #8058, opened 2026-05-31
[core] Disable snapshot expiry in testDiscardDuplicateFilesMultiThread

ArnavBalyan:
### Purpose
 - Fix flaky testDiscardDuplicateFilesMultiThread CI failure.
 - Test runs 10 concurrent writers committing to same table, each writer also runs snapshot expiry.
 - However under load, a writer can read the latest snapshot id from a directory listing, then find that snapshot deleted by another writer's expiry before it can be opened.
 - Ensure that for testing purposes, we can raise num retained snapshots to 1k above to avoid the test race and ensure the test does not fail intermittently. 

### Tests
 - CI


JingsongLi:
+1

PR #8059, opened 2026-06-01
[python][daft] Push down supported NOT predicates

QuakeWang:
### Purpose

Daft predicate conversion previously treated every `NOT` expression as unsupported, so filters such as `NOT (col == lit)`, `NOT col.is_in(...)`, and `NOT col.is_null()` stayed above the datasource even though pypaimon already has matching predicate types.

This change adds a narrow leaf-only rewrite for supported `NOT` predicates:
- `NOT equal` -> `notEqual`
- `NOT in` -> `notIn`
- `NOT between` -> `notBetween`
- `NOT isNull` -> `isNotNull`
- `NOT isNotNull` -> `isNull`

It intentionally does not add De Morgan rewrites for compound expressions.

The change also fixes `notIn` Arrow evaluation to exclude null values. Without this, fallback datasource tasks could return null rows for pushed `NOT IN` filters, and pushed limits could then truncate valid rows.

### Tests

- `pytest paimon-python/pypaimon/tests/daft -q`
- `pytest paimon-python/pypaimon/tests/predicates_test.py -q`

### DIFF
--- a/paimon-python/pypaimon/common/predicate.py
+++ b/paimon-python/pypaimon/common/predicate.py
@@ -299,7 +299,7 @@ class NotIn(Tester):
         return not any(min_v == l == max_v for l in literals)
 
     def test_by_arrow(self, val, literals) -> bool:
-        return ~val.isin(literals)
+        return (~val.isin(literals)) & val.is_valid()
 
 
 class Between(Tester):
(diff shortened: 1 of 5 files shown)


JingsongLi (approved):
+1

PR #8060, opened 2026-06-01
[spark] SparkFilterConverter: support AlwaysTrue/False, fix silent NaN drop

heye1005:
### Purpose

fix two TODOs in `SparkFilterConverter`:
- translate `AlwaysTrue` / `AlwaysFalse`
- rewrite `col = NaN` to `IsNaN(col)` — Spark treats `NaN = NaN` as true, Paimon doesn't, so NaN rows were getting silently dropped on pushdown.

### Tests

`SparkFilterConverterTest#testAlwaysTrueFalse`, `#testEqualToNaN`.


JingsongLi (commented):
LGTM, thanks.

PR #8061, opened 2026-06-01
[python][daft] Preserve pushed filters across source serialization

QuakeWang:
### Purpose

`PaimonDataSource.__setstate__()` restores pushed filter state before reopening the table, but `_init_table()` then reset `_pushed_filters`, `_paimon_predicate`, and `_remaining_filters` to `None`.

This made serialized Daft sources lose filters already accepted by `push_filters()`. In fallback reads, `get_tasks(Pushdowns(filters=None, limit=1))` could then plan an unfiltered limited read and return the wrong row.

This PR moves pushdown state initialization to `__init__()`, keeps `_init_table()` limited to table-derived metadata, and includes `_pushed_filters` in the serialized state for consistent explain/debug output.

### Tests

- `pytest paimon-python/pypaimon/tests/daft/daft_data_test.py`


JingsongLi:
+1

PR #8062, opened 2026-06-01
[spark] Support ON_ERROR = CONTINUE / SKIP_FILE in COPY INTO

JunRuiLee:
## Motivation

This is part of #8005.

COPY INTO previously only supported `ON_ERROR = ABORT_STATEMENT`: any parse or
cast error aborted the entire command. In production data-loading pipelines a
single malformed row or file would then fail the whole batch, which is often
too strict. This adds two error-tolerant modes:

- `CONTINUE` — skip bad rows and load the rest (row-level tolerance).
- `SKIP_FILE` — skip any file that contains an error, all-or-nothing per file.

`ABORT_STATEMENT` remains the default, so existing behavior is unchanged.

## SQL dialect

Paimon's `COPY INTO` is a **Snowflake-style** extension (`FILE_FORMAT = (TYPE = ...)`,
`PATTERN`, `FORCE`, `ON_ERROR`), not the Databricks `COPY INTO` form (`FILEFORMAT` +
`FORMAT_OPTIONS (...)` / `COPY_OPTIONS (...)`). It implements only a **subset** of the
Snowflake syntax. In particular, `ON_ERROR` supports `ABORT_STATEMENT` (default),
`CONTINUE`, and `SKIP_FILE`; the Snowflake variants `SKIP_FILE_<num>` and
`SKIP_FILE_<num>%` are **not** supported. This is documented in `sql-write.md`.

## Changes

- Grammar: `ON_ERROR` now accepts `CONTINUE` and `SKIP_FILE` in addition to
  `ABORT_STATEMENT`.
- Result schema gains two columns:
  - `errors_seen` (BIGINT) — number of error rows per file.
  - `first_error` (STRING) — first error message, NULL when the file is clean.
  - `status` now also reports `PARTIALLY_LOADED` and `LOAD_FAILED`.
- Error detection runs once per batch; both modes write in a single commit.
  Load history is recorded so error-tolerant runs stay idempotent under
  `FORCE = FALSE`.
- Refactor: `CopyIntoTableExec` is split into focused helpers
  (`CopyIntoHelper`, `CopyIntoCastValidator`, `CopyIntoDataFrameBuilder`,
  `CopyIntoErrorHandler`, `CopyIntoResultBuilder`), shared across CSV/JSON/Parquet.
- Docs updated in `sql-write.md`, including a Snowflake-subset dialect note and
  the CSV column-count-mismatch caveat under `CONTINUE`.

Supported for CSV, JSON, and Parquet.


JingsongLi (commented):
Thanks for adding `ON_ERROR` support. I found two things that look worth fixing before merge:

1. The CSV column-count mismatch behavior is inconsistent between the docs and tests. The docs say that with `ON_ERROR = CONTINUE`, Spark PERMISSIVE semantics are followed, so rows with fewer columns are padded with `NULL`, extra columns are ignored, and those rows are loaded without being counted as errors (`docs/docs/spark/sql-write.md`). However, the new tests expect both fewer-column and extra-column CSV rows to be skipped and counted as errors (`CopyIntoOnErrorTest`). Could we align these? Either the implementation/tests should follow the documented PERMISSIVE behavior, or the docs should describe the stricter behavior being introduced.

2. `CONTINUE` and `SKIP_FILE` were added to the grammar as non-reserved identifiers, but they were not added to `AbstractPaimonSparkSqlExtensionsParser.nonReservedIdentifierTokenTypes`. That set is used by the fallback gate for catalog-qualified `CREATE TABLE LIKE`. As a result, a statement such as `CREATE TABLE paimon.db.continue LIKE paimon.db.source_tbl` can fail to fall back to the Paimon parser even though the grammar says `continue` is a valid non-reserved identifier. Please add the new tokens there and cover them with a parser test similar to the existing `tag` case.

docs/docs/spark/sql-write.md
```diff
@@ -597,9 +599,9 @@ By default (`FORCE = FALSE`), COPY INTO tracks which files have been successfull
 
 ### Limitations
 
+- **CSV column-count mismatch with ON_ERROR=CONTINUE**: Paimon follows Spark's PERMISSIVE mode behavior — rows with fewer columns are padded with NULL, and extra columns are silently ignored. These rows are loaded successfully and not counted as errors.
```
JingsongLi: The behavior described here conflicts with the implementation and the new tests below. `runTextImport` reads CSV with `_corrupt_record` in PERMISSIVE mode and `detectTextErrors` filters `corruptCol.isNotNull` as parse errors, while `CopyIntoOnErrorTest` expects fewer/extra columns to become `PARTIALLY_LOADED` with `errors_seen = 1`. So either this limitation should say these rows are skipped/count as errors, or the implementation/tests need to be changed to match the documented Spark PERMISSIVE behavior.

JunRuiLee:
Hi @JingsongLi,


  Thanks for the review! I've addressed all comments, PTAL.

  1. **CSV column-count mismatch**: Updated docs to match the implementation - mismatched rows are treated as errors and skipped with `ON_ERROR = CONTINUE`.

  2. **nonReservedIdentifierTokenTypes**: Added `CONTINUE` and `SKIP_FILE` tokens with parser tests.

JingsongLi:
Thanks for the work here. I have one concern about the SQL dialect alignment before approving.

This syntax looks closer to a Snowflake-style `COPY INTO` subset than Databricks syntax. Databricks `COPY INTO` uses `FILEFORMAT = ...` plus `FORMAT_OPTIONS (...)` / `COPY_OPTIONS (...)`, and the documented copy options are things like `force` and `mergeSchema`; it does not expose `ON_ERROR = ABORT_STATEMENT | CONTINUE | SKIP_FILE` in this top-level form.

If the intent is to follow Databricks syntax, I think this PR is not aligned yet. If the intent is to introduce a Snowflake-style Paimon extension, please clarify that in the PR/docs and document that this is only a subset: `ABORT_STATEMENT`, `CONTINUE`, and `SKIP_FILE` are supported, while Snowflake variants such as `SKIP_FILE_<num>` / `SKIP_FILE_<num>%` and other `COPY INTO` options are not.

JunRuiLee:
Hi @JingsongLi , I've updated the docs that document that Paimon's COPY INTO follows Snowflake syntax (not Databricks) and supports only a subset. PTAL.

JingsongLi:
+1

PR #8063, opened 2026-06-01
[core] Validate schema after applying schema changes

leaves12138:
### Purpose

`SchemaManager.generateTableSchema` can produce an invalid option combination after applying schema changes. For example, resetting `fields.<seq>.sequence-group` may leave a field-specific `aggregate-function` that requires that sequence group, and the failure is only exposed later when the table is loaded.

### Changes

- Validate the generated `TableSchema` before returning it from `SchemaManager.generateTableSchema`.
- Add regression coverage for resetting a sequence group used by `fields.<field>.aggregate-function`.
- Keep `last_non_null_value` behavior unchanged, since it does not require a sequence group.

### Tests

```bash
mvn -U -pl paimon-core -am -DskipITs -Dcheckstyle.skip -Drat.skip=true -Dspotless.check.skip=true -DfailIfNoTests=false -Dtest=SchemaManagerTest#testResetSequenceGroupForAggregateFunction+testResetSequenceGroupForLastNonNullAggregateFunction test
```

### DIFF
--- a/paimon-core/src/main/java/org/apache/paimon/schema/SchemaManager.java
+++ b/paimon-core/src/main/java/org/apache/paimon/schema/SchemaManager.java
@@ -609,14 +609,17 @@ public class SchemaManager implements Serializable {
                         applyRenameColumnsToOptions(newOptions, changes),
                         newComment);
 
-        return new TableSchema(
-                oldTableSchema.id() + 1,
-                newSchema.fields(),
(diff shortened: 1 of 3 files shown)


JingsongLi (commented):
LGTM if tests has been passed.

JingsongLi:
+1

